PR #1632, opened 2025-04-10
Raise an error when rebuilding a PyTree leaf from a ShapedArray

rauletorresc:
**Context:** Trying to reconstruct a PyTree leaf from a ShapedArray fails.

**Description of the Change:** Before unflattening a previously flattened arg list, check if any of the abstracted args is a ShapedArray. If this is true, raise an error.

**Benefits:** Capture the error before JAX tree unflatten function, which might give details to the user which are too low level.

**Related GitHub Issues:** #339

### DIFF
--- a/frontend/catalyst/tracing/type_signatures.py
+++ b/frontend/catalyst/tracing/type_signatures.py
@@ -82,7 +82,10 @@ def get_abstract_signature(args):
 
     abstract_args = [shaped_abstractify(arg) for arg in flat_args]
 
-    return tree_unflatten(treedef, abstract_args)
+    if all(not isinstance(arg, jax.core.ShapedArray) for arg in abstract_args):
+        return tree_unflatten(treedef, abstract_args)
+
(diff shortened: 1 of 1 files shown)


frontend/catalyst/tracing/type_signatures.py
```diff
@@ -82,7 +82,10 @@ def get_abstract_signature(args):
 
     abstract_args = [shaped_abstractify(arg) for arg in flat_args]
 
-    return tree_unflatten(treedef, abstract_args)
+    if all(not isinstance(arg, jax.core.ShapedArray) for arg in abstract_args):
+        return tree_unflatten(treedef, abstract_args)
```
dime10, line 86: I'm confused, aren't these by definition shaped arrays? (`shaped_abstractify(arg)`)

dime10:
@mehrdad2m I think it's safe to remove this one from the milestone

Issue #1611, opened 2025-04-04
Cannot return mid-circuit measurements for `qrack.simulator` within `qjit`


yorgossot:
The following code reproduces the bug:

```python
import pennylane as qml
from catalyst import qjit, measure

noise = 0.01
shots = 10

dev_qrack = qml.device("qrack.simulator", wires=2, noise=noise, shots=shots)

@qjit
@qml.qnode(dev_qrack)
def circuit():
    qml.X(0)
    measurements = [measure(i) for i in range(2)]
    return measurements

circuit()
```

Produces the error message:
```
CompileError: A transformed quantum function must return either a single measurement, or a nonempty sequence of measurements.
```

---

Similarly, the following code also fails:
```python
import pennylane as qml
from catalyst import qjit, measure

noise = 0.01
shots = 10

dev_qrack = qml.device("qrack.simulator", wires=2, noise=noise, shots=shots)

@qjit
@qml.qnode(dev_qrack)
def circuit():
    qml.X(0)
    measurements = [measure(i) for i in range(2)]
    return qml.sample(measurements)


circuit()
```
giving:
```
QuantumFunctionError: Only sequences of single MeasurementValues can be passed with the op argument. MeasurementValues manipulated using arithmetic operators cannot be used when collecting statistics for a sequence of mid-circuit measurements.
```

dime10:
Hi @yorgossot, thanks for reporting this! The second example is using `qml.sample` incorrectly, as it can only accept a single measurement value, so something like this is required:
```py
    return [qml.sample(m) for m in measurements]
```

Note also that to use mcms with shots, you should supply `mcm_method="one-shot"` as otherwise Catalyst will only run a single simulation, and thus generate a single measurement value for all "shots".

dime10:
For the first example you provide, this appears to be a bug in the Catalyst/PennyLane device pre-processing mechanism, we'll push a fix for it asap!

dime10:
Alright, this issue should be fixed now on the latest main branch :)

PR #1634, opened 2025-04-10
Fix measurement handling in JAX tracer and device decomposition

sengthai:
**Context:**
Fix issue when measurements handling in JAX on Qrack device were incorrectly marked as modified when they weren’t.

- Removed validation for transformed quantum function outputs in `is_transform_valid_for_batch_transforms`.
- Enhanced `apply_transform` to return `device_modify_measurements` based on the presence and comparison of measurements in tapes.
- Updated `measurements_from_samples` to check for measurement types and return a postprocessing function if no changes are detected. These changes improve the handling of measurement results and streamline the transformation process in the quantum function tracing.

**Benefits:**
- Fix bug

**Possible Drawbacks:**

**Related GitHub Issues:**
#1611

[[sc-88296]]


frontend/catalyst/jax_tracer.py
```diff
@@ -1072,7 +1066,21 @@ def apply_transform(
     if not is_valid_for_batch and len(tapes) > 1:
         msg = "Multiple tapes are generated, but each run might produce different results."
         raise CompileError(msg)
-    return tapes, post_processing
+
+    # Check if there is exactly one tape and device modification measurements are enabled
+    # TODO: Tapes can be more than one
+    if len(tapes) == 1 and device_modify_measurements:
```
sengthai, line 1072: This is a hack. We should consider this in the case of multiple tapes.

frontend/catalyst/jax_tracer.py
```diff
@@ -1072,7 +1066,21 @@ def apply_transform(
     if not is_valid_for_batch and len(tapes) > 1:
```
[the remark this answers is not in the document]
dime10: I think you're mostly there, but it would be good to refactor this cleanly (e.g. use subroutines for the different conditions) and then raise the appropriate error messages (right now there is only one) and produce the correct tracing mode to use.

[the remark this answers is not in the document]
sengthai: Thanks @dime10 for the explanation and feedback. I’ve refactored the code and slightly adjusted its structure. This improves readability, though we now always perform the transform before checking conditions, which slightly impacts performance.

frontend/catalyst/jax_tracer.py
```diff
@@ -999,47 +999,115 @@ def trace_quantum_measurements(
...
+
+    Returns:
+        int: Tracing mode to use - either 0 (default) or 1 (transform)
+            - "default": Allows mid-circuit measurements and return function's results
+              directly. Used when no transform is applied or when transform doesn't
+              modify measurements or produce multiple tapes.
+            - "transform": Uses tape measurements instead of original function results.
+              Prohibits mid-circuit measurements with multiple tapes and requires function to
+              return only measurements (no classical results). Used when transform produces
+              multiple tapes or modifies measurements.
```
erick-xanadu, line 1085: I think an [enum](https://docs.python.org/3/library/enum.html) could work well here instead of magic numbers.

erick-xanadu (changes requested):
Would it be possible to write some tests for this?

frontend/catalyst/jax_tracer.py
sengthai: Sure thing, well modify to using enum and add some tests as well.

erick-xanadu:
@sengthai remember to rebase your changes to avoid including all changes in the main branch. Thanks!

frontend/catalyst/jax_tracer.py
```diff
@@ -1064,47 +1065,125 @@ def trace_quantum_measurements(
...
+    and Catalyst and discussed more as well."""
+
+    # Check if outputs are valid for transformations (measurements only)
+    is_valid_output = has_valid_measurement_outputs(flat_results)
+    is_wave_function_collapsed = has_midcircuit_measurement(tape)
+
+    are_batch_transforms_valid = is_valid_output and not is_wave_function_collapsed
+    return are_batch_transforms_valid
+
+
```
erick-xanadu, line 1133:
```suggestion
```

It looks like this is just dead code now.

sengthai: Ohhh, that's right. I forgot to remove it. Thanks for catching this one. Erick.

dime10 (changes requested):
After reviewing I tried implementing the proposed changes here: https://github.com/PennyLaneAI/catalyst/pull/1644

Unfortunately I wasn't quite able to make it work, there are some edge cases around the return PyTree shape of qnodes.

I'm happy to delay this PR until post-release until we figure this out :)

frontend/catalyst/jax_tracer.py
```diff
@@ -999,47 +999,115 @@ def trace_quantum_measurements(
...
+    Returns:
+        bool: True if outputs are valid for transformations, False otherwise
+    """
     # Can transforms be applied?
     # Since transforms are a PL feature and PL does not support the same things as
     # Catalyst, transforms may have invariants that rely on PL invariants.
     # For example:
     #   * mid-circuit measurements (for batch-transforms)
     #   * that the output will be only a sequence of `MeasurementProcess`es.
+    class_tracers, meas_tracers = split_tracers_and_measurements(flat_results)
```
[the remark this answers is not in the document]
erick-xanadu: Yes, this is a good change!

frontend/catalyst/jax_tracer.py
```diff
@@ -999,47 +999,115 @@ def trace_quantum_measurements(
...
+
     def is_measurement(op):
         """Only to avoid 100 character per line limit."""
         return isinstance(op, MeasurementProcess)
 
+    has_mixed_classical_quantum_results = class_tracers and meas_tracers
     is_out_measurements = map(is_measurement, meas_tracers)
-    is_all_out_measurements = all(is_out_measurements) and not class_tracers
+    is_all_out_measurements = all(is_out_measurements) and not has_mixed_classical_quantum_results
```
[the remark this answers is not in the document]
sengthai: Yeah, I’m not sure either if only classical tracers break things. Does that mean we wouldn’t be able to return purely classical results? 🤔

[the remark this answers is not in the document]
sengthai: For mixture variable, I thought measurement results shouldn’t mix classical and quantum values, like:
`return [some_number, qml.sample(m)]`

frontend/catalyst/jax_tracer.py
```diff
@@ -999,47 +999,115 @@ def trace_quantum_measurements(
...
+              return only measurements (no classical results). Used when transform produces
+              multiple tapes or modifies measurements.
+
+    Raises:
+        CompileError: If the transform is not legal for the given program
+    """
+    # Before determine the tracing mode, make sure that the transform:
+    # - Not allow mid-circuit measurements on the tapes (wave function collapse)
+    # - Does not produce multiple tapes
+    if not has_valid_measurement_outputs(flat_results):
```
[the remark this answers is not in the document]
sengthai: Ohh I see. Makes sense not to raise the error before knowing the tapes were actually changed.  Thanks for the suggestion!  I’ll give it a shot 😊

frontend/catalyst/jax_tracer.py
```diff
@@ -1064,54 +1065,91 @@ def trace_quantum_measurements(
...
+                function to return only measurements (no classical results).
+                Used when transform produces multiple tapes or modifies measurements.
+    """
+
+    DEFAULT = 0
+    TRANSFORM = 1
+
+
+@debug_logger
+def is_measurement_changed(original_tape, modified_tape):
```
sengthai, line 1132: I'm still working on getting this implementation just right!

frontend/catalyst/jax_tracer.py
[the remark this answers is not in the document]
sengthai: Yes @dime10. Classical results can still be returned and the issue only comes up when there’s a quantum measurement involved.

dime10:
Hi @erick-xanadu, would you mind having another look? It should be working now, although I didn't get as far as I hoped in allowing qnode transforms through that don't modify the tape measurements, it does work with device transforms.

erick-xanadu:
@dime10 , can you remind me again what is a device transform and how that is different from a qnode transform?

erick-xanadu (commented):
Looks good, only one suggested change regarding zip.

frontend/catalyst/jax_tracer.py
```diff
@@ -1064,58 +1059,69 @@ def trace_quantum_measurements(
...
+    # immutability. So we can compare identity directly. Equality comparisons are problematic
+    # when the measurements contain tracers.
+    for original_meas, modified_meas in zip(original_tape.measurements, modified_tape.measurements):
+        if original_meas is not modified_meas:
+            return True
 
-    is_wave_function_collapsed = any(map(is_midcircuit_measurement, tape.operations))
-    are_batch_transforms_valid = is_valid_output and not is_wave_function_collapsed
-    return are_batch_transforms_valid
+    return False
```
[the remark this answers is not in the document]
dime10: Great catch, thanks Erick :)

frontend/catalyst/jax_tracer.py
```diff
@@ -1064,58 +1059,69 @@ def trace_quantum_measurements(
...
+
+    DEFAULT = 0
+    TRANSFORM = 1
+
+
+@debug_logger
+def have_measurements_changed(original_tape, modified_tape):
+    """Check if the measurement has changed."""
+
+    # Copying tapes preserves object identity in the operation and measurement lists, due to
```
dime10, line 1107: Forgot I had this in my other PR:
```suggestion
    if len(original_tape.measurements) != len(modified_tape.measurements):
        return True

    # Copying tapes preserves object identity in the operation and measurement lists, due to
```

frontend/catalyst/jax_tracer.py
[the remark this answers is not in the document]
dime10: https://github.com/PennyLaneAI/catalyst/pull/1634/commits/e6f5de14ee16721cb5fbc5fa866eb422e53149c0

frontend/catalyst/jax_tracer.py
```diff
@@ -1125,19 +1134,35 @@ def apply_transform(
...
+                "Batch transforms are unsupported with MCMs or non-MeasurementProcess QNode "
+                "outputs. The selected device, options, or applied QNode transforms, may be "
+                "attempting to produce multiple tapes."
+            )
+            raise CompileError(msg)
+        tracing_mode = TracingMode.TRANSFORM
+    elif len(qnode_program) or have_measurements_changed(tape, tapes[0]):
+        # TODO: Ideally we should allow qnode transforms that don't modify the measurements to
+        # operate in the permissive tracing mode, but that currently leads to a small number of
+        # test failures due to the different result format produced in trace_quantum_function.
```
erick-xanadu, line 1153: I spent some time here.

The reason for this is likely that the post_processing function is different (?). What is also curious is that the cases that I saw failing was the `broadcast_expand` function but without the need to broadcast it (i.e., it produced only one tape).

erick-xanadu: I wonder if this change in PennyLane would fix it:

```diff
diff --git a/pennylane/transforms/broadcast_expand.py b/pennylane/transforms/broadcast_expand.py
index 2d5596c14..c46555ad3 100644
--- a/pennylane/transforms/broadcast_expand.py
+++ b/pennylane/transforms/broadcast_expand.py
@@ -131,7 +131,7 @@ def broadcast_expand(tape: QuantumScript) -> tuple[QuantumScriptBatch, Postproce
     >>> fn(qml.execute(tapes, qml.device("default.qubit", wires=1), None))
     tensor([0.98006658, 0.82533561, 0.54030231], requires_grad=True)
     """
-    if tape.batch_size is None:
+    if tape.batch_size is None or tape.batch_size == 1:
         return (tape,), null_postprocessing
 
     has_postselect = any(
```

dime10: Hmm, how do you think the post-processing is different?

I think a big difference is this:
https://github.com/PennyLaneAI/catalyst/blob/e6f5de14ee16721cb5fbc5fa866eb422e53149c0/frontend/catalyst/jax_tracer.py#L1380-L1383
That means we only return tuples or a single element when we have qnode transforms.
That was the reason I tentatively updated this: https://github.com/PennyLaneAI/catalyst/pull/1644/commits/76c6b521ada37dcf1fd331190d3e6dd86f121ea2

But there was also another error after that :/

erick-xanadu: I think we are using `len(tapes)` as a proxy for `post_processing != null_postprocessing`.

dime10: I think we can separate that from this PR though, as is at least device transforms won't prevent us from returning classical values.

dime10 (approved):

PR #1643, opened 2025-04-11
Exclude files specifically changed for release from rc sync PR

mehrdad2m:
**Context:**
During feature freeze, we need to modify the `check-catalyst` workflow on the release candidate branch. These changes should not be included in the daily sync that merged into main branch.

- Added the `check-catalyst` workflow which should be modified in feature freeze week to be excluded.

**Benefits:**
Less maintenance of the daily rc sync PR.

**Possible Drawbacks:**

**Related GitHub Issues:**


paul0403:
Is this still relavant?

.github/workflows/rc_sync.yaml
```diff
@@ -61,6 +61,16 @@ jobs:
...
+          git checkout origin/main -- .github/workflows/check-catalyst.yaml
+          git checkout origin/main -- .github/workflows/build-wheel-linux-x86_64.yaml
+          git checkout origin/main -- .github/workflows/build-wheel-linux-arm64.yaml
+          git checkout origin/main -- .github/workflows/build-wheel-macos-x86_64.yaml
+          git checkout origin/main -- .github/workflows/build-wheel-macos-arm64.yaml
+          git checkout origin/main -- .dep-versions
+          git checkout origin/main -- .github/workflows/scripts/linux_arm64/rh8/test_wheels.sh
+          git checkout origin/main -- Makefile
+          git checkout origin/main -- doc/requirements.txt
+          git checkout origin/main -- setup.py
```
dime10, line 73: > @dime10 , My original motivation was to only exclude the check-catalyst.yaml file, but not too sure about the rest. I am not sure how frequently we would have changes in the other files that want in the main too. Do you think it make sense to exclude them?

Thanks Mehrdad, yeah good question, this does look like a lot of files 🤔 I think all the actions are probably fine, since we do inject those patches there during the release process.  The `.dep-versions` and `doc/requirements.txt` are also updated according to the release guide. For the Makefile and setup.py I don't, we could leave those out for now and see later? The linux scripts are gone now (🎉), so that one can be removed :)

mehrdad2m: I only kept the `check-catalyst` script cause that is the only change that happens in the beginning of the week and can cause conflicts everyday during the week. The rest of the files are only modified at the end of the week, so it would cause problem in the last sync which can be fine.

PR #1648, opened 2025-04-14
Remove redundant flatten/unflatten of PyTrees

rauletorresc:
**Context:** When calling `get_abstract_signature` from within `get_decomposed_signature` we are flattening/unflattening the PyTree one more time without need. In this sense `get_abstract_signature` is just a wrapper for the `shaped_abstractify` function that we don't need here.

**Description of the Change:** Avoid the wrapper function call and use `shaped_abstractify` directly.

**Benefits:** Avoid code that does nothing.


dime10:
This change looks good, but if we want to resolve the core issue we have to tackle it here:
https://github.com/PennyLaneAI/catalyst/blob/300091953097ae038b813b8c15ee4c47e97891bc/frontend/catalyst/jit.py#L715
- We generate an abstract signature re-wrapped into pytrees.
- The question is whether we need the abstract values at point, and if so whether we need them in pytrees. We can modify either of those conditions to become compliant.

The abstract signature is used for example to generate jaxpr:
https://github.com/PennyLaneAI/catalyst/blob/300091953097ae038b813b8c15ee4c47e97891bc/frontend/catalyst/jit.py#L740
- but further down we already abstractify: https://github.com/PennyLaneAI/catalyst/blob/300091953097ae038b813b8c15ee4c47e97891bc/frontend/catalyst/jax_extras/tracing.py#L530

Then it is used again during promotion:
https://github.com/PennyLaneAI/catalyst/blob/300091953097ae038b813b8c15ee4c47e97891bc/frontend/catalyst/jit.py#L604
- Here we can probably just do the promotion on a flattened argument list instead.

The last use case, in the compilation cache, is already addressed in this PR.

PR #1663, opened 2025-04-16
Add MBQC dialect

joeycarter:
**Context:** In order to QJIT compile a typical MBQC workload, we require mechanisms to capture and lower the MBQC data structures and functions recently added to the PennyLane [`ftqc`](https://docs.pennylane.ai/en/stable/code/qml_ftqc.html) module. In preparation for this work, we require an `mbqc` dialect in the MLIR layer of Catalyst to contain the corresponding MBQC data types and operations.

**Description of the Change:** This PR adds the framework for the `mbqc` dialect. Currently, it only defines a single operation, `mbqc.measure_in_basis`, which is an instruction representing arbitrary-basis measurements parameterized by a plane along the major axes of the Bloch sphere (either XY, ZX or YZ) and a rotation angle about that plane. Other operations, data types and passes will be added in future PRs as needed.

[[sc-89176](https://app.shortcut.com/xanaduai/story/89176/create-the-mbqc-mlir-dialect)]


mlir/include/MBQC/IR/MBQCDialect.td
```diff
@@ -0,0 +1,72 @@
...
+#define MBQC_DIALECT
+
+include "mlir/IR/DialectBase.td"
+include "mlir/IR/OpBase.td"
+
+//===----------------------------------------------------------------------===//
+// MBQC dialect definition.
+//===----------------------------------------------------------------------===//
+
+def MBQC_Dialect : Dialect {
```
dime10, line 25: Ah, I think we'll want to declare the QuantumDialect as a dependency, since we're reusing the qubit type (forgot to do that in the QEC dialect actually):
https://mlir.llvm.org/docs/DefiningDialects/#dependent-dialects

mlir/include/MBQC/IR/MBQCOps.td
```diff
@@ -0,0 +1,70 @@
...
+        A parametric single-qubit projective measurement is equivalent to the `quantum.measure`
+        operation except that it supports measurements in an arbitrary basis, rather than the
+        computational basis only. Specifically, the measurement basis is parameterized by a plane
+        along the major axes of the Bloch sphere (either XY, ZX or YZ) and a rotation angle about
+        that plane.
+    }];
+
+    let arguments = (ins
+        QubitType:$in_qubit,
+        StrAttr:$plane,
```
dime10, line 55: If wanted to be more declarative you could use an enum attribute for the planes (if understand correctly there are only 3 options?), rather than an opaque string.

dime10 (approved):
Looks great!

mlir/include/MBQC/IR/MBQCDialect.td
joeycarter: d8e50654

mlir/include/MBQC/IR/MBQCOps.td
joeycarter: @paul0403 and I were discussing this yesterday. I agree, I think this would be better since there are only three options.

joeycarter: Updated to use an enum attribute in 16019c19.

Issue #1340, opened 2024-11-28
[Feature] qjit-compatible implementation of `itertools.product`


josh146:
It's quite common to see workflows that make use of itertools related functionality, in particular 

- `itertools.product`,
- `itertools.combinations`, and
- `itertools.permutations`.

Here, I focus only on `product` as this is part of the workflow I am using.

For example, consider:

```python
@qml.qjit(autograph=True)
def fn(x):

    for i, j in itertools.product(range(2), repeat=2):
        x += i + j - i * j * jnp.sin(j)

    return x
```

Here, `product` has the following rough semantics:

```python
def p(*iterables, repeat=1):
    result = [[]]

    if repeat < 0:
        raise ValueError('repeat argument cannot be negative')

    for l in range(repeat):
        for pool in map(tuple, iterables):
            _temp = []

            for i in result:
                for j in pool:
                    _temp.append(i + [j])

            result = _temp

    for prod in result:
        yield tuple(prod)
```

This will fail to compile, since Autograph is not able to recognize and convert `itertools.product` (even though the arguments are all compile-time constant).

I propose adding our own `catalyst.product`, similar to our existing `range`, that AutoGraph can use as the conversion target. This would then lower to a fast implementation of product at the MLIR layer. An advantage here is that this would allow `product` to work even with dynamic arguments.

josh146:
Longer term, we could also do the same thing for `itertools.combinations` and `itertools.permutations`.

dime10:
Hi @josh146, I think this

> This will fail to compile, since Autograph is not able to recognize and convert itertools.product (even though the arguments are all compile-time constant).

was actually a bug because static constructs should always be able to fall-back into their Python implementation. I fixed it in #1665, but it still won't allow you supply dynamic args.

PR #1665, opened 2025-04-16
Fix AutoGraph fallback for certain iteration targets

dime10:
Fix AutoGraph fallback for valid iteration targets with constant data but no length, for example the following program is now supported:

```py
@qml.qjit(autograph=True)
def fn(x):

    for i, j in itertools.product(range(2), repeat=2):
        x += i + j - i * j * jnp.sin(j)

    return x

>>> fn(5)
8.158529015192103
```

Fixes the bug discovered in #1340, but doesn't close it as it also contains a feature request.
[sc-89322]

(Also adjusts the line length `isort` is working with to 100, same as our other tools.)

### DIFF
--- a/frontend/catalyst/autograph/ag_primitives.py
+++ b/frontend/catalyst/autograph/ag_primitives.py
@@ -287,20 +287,18 @@ def for_stmt(
         enum_start = None
         iteration_array = None
     elif isinstance(iteration_target, CEnumerate):
-        start, stop, step = 0, len(iteration_target.iteration_target), 1
-        enum_start = iteration_target.start_idx
         try:
+            start, stop, step = 0, len(iteration_target.iteration_target), 1
+            enum_start = iteration_target.start_idx
             iteration_array = jnp.asarray(iteration_target.iteration_target)
         except:  # pylint: disable=bare-except
-            iteration_array = None
             fallback = True
     else:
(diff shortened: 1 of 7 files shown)


frontend/catalyst/autograph/ag_primitives.py
```diff
@@ -287,20 +287,18 @@ class EmptyResult: ...
...
+            enum_start = iteration_target.start_idx
             iteration_array = jnp.asarray(iteration_target.iteration_target)
         except:  # pylint: disable=bare-except
-            iteration_array = None
             fallback = True
     else:
-        start, stop, step = 0, len(iteration_target), 1
-        enum_start = None
         try:
+            start, stop, step = 0, len(iteration_target), 1
```
paul0403, line 298: Nice! Just so I understand correctly, we move these lines inside the `try` so that for objects coming from itertools (which would fall to the else block here), we can properly go to fallback, instead of encountering "object from itertools does not have `len` method" error? (I assume this is the error, since the issue didn't actually show the error)

paul0403 (approved):
:100:

frontend/catalyst/autograph/ag_primitives.py
dime10: That's right!

PR #1672, opened 2025-04-22
Updating Nightly build download path

runora95:
**Context:**
The nightly build for the standalone plugin is [unable to find the plugin files ](https://github.com/PennyLaneAI/catalyst/actions/runs/14586255969/job/40912963093). This step in the nightly release workflow had [not been run](https://github.com/PennyLaneAI/catalyst/actions/runs/14507591127/job/40701985590) for an unknown amount of time. The if conditions for the cronjobs all of the wheel builds were not becoming `true`.

Adding a pattern to the `actions/download-artifact@v4` job.

**Benefits:**
Nightly builds get uploaded.

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/.github/workflows/build-nightly-release.yaml
+++ b/.github/workflows/build-nightly-release.yaml
@@ -63,6 +63,7 @@ jobs:
       with:
         merge-multiple: true
         path: dist
+        pattern: catalyst-*
 
     - name: Install rename
       run: |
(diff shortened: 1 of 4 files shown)


rashidnhm (commented):
Is there a label to add to test that it's working? 

Mainly need to ensure that standalone plugin tests are executed and passing across the different architectures

dime10:
This is good but I think we should prevent the standalone step from running in the first place

PR #1673, opened 2025-04-22
Fixing demo due to removal of `coeffs` from the `Sum` class

PietropaoloFrisoni:
This demo in Catalyst uses code that was just removed during Deprecations and Removals for PL 0.42.

The reason why a warning message (indicating that the code was deprecated) has not been caught before is that such a demo (and potentially other ones) suppresses all warnings, including deprecation warnings coming from PL. 

I think this needs to change, for example, by suppressing all warnings except for PL deprecation warnings, because a demo using deprecated PL code should not be exposed to users.

[sc-89676]

### DIFF
--- a/demos/adaptive_circuits_demo.ipynb
+++ b/demos/adaptive_circuits_demo.ipynb
@@ -202,7 +202,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 6,
+   "execution_count": null,
    "id": "6a9049c4",
    "metadata": {},
    "outputs": [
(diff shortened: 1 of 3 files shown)


demos/adaptive_circuits_demo.ipynb
```diff
@@ -202,7 +202,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 6,
+   "execution_count": null,
```
dime10, line 205: Minor, but we can revert these to not show "unexecuted" cells
```suggestion
   "execution_count": 6,
```

demos/adaptive_circuits_demo.ipynb
```diff
@@ -796,7 +795,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 18,
+   "execution_count": null,
```
dime10, line 798:
```suggestion
   "execution_count": 18,
```

dime10 (approved):
Thanks Pietro!

PR #1674, opened 2025-04-22
Add runtime C-API for arbitrary-basis measurement operations

joeycarter:
**Context:** A pass to convert the `mbqc.measure_in_basis` MLIR operation (in the `mbqc` dialect) into the LLVM dialect was added in #1679. In order to execute a compiled program containing this operation, we require a runtime function defined in the Catalyst runtime C-API.

**Description of the Change:** Adds a new runtime stub to the Catalyst runtime C-API called `__catalyst__mbqc__measure_in_basis()`. This runtime function is effectively the same as the current `__catalyst__qis__Measure()` runtime function for computational-basis mid-circuit measurements since we will be validating the QJIT-compiled MBQC workloads on the null.qubit device and do not require that it fully simulates arbitrary-basis measurements—rather we treat it as a placeholder for now in order to make the workload executable (see Shortcut story linked below for details).

[[sc-89637](https://app.shortcut.com/xanaduai/story/89637/add-runtime-c-api-for-arbitrary-basis-measurement-operations)]


runtime/lib/capi/RuntimeCAPI.cpp
```diff
@@ -1031,4 +1031,26 @@ int8_t *__catalyst__rt__array_get_element_ptr_1d(QirArray *ptr, int64_t idx)
...
+
+// -------------------------------------------------------------------------- //
+// MBQC Runtime CAPI
+// -------------------------------------------------------------------------- //
+
+// NOTE: Currently this runtime operations is exactly the same as __catalyst__qis__Measure();
+//       we effectively treat it as a no-op for now. When hardware devices that natively support
+//       mid-circuit measurements in an arbitrary basis are available, we will create a new
+//       QuantumDevice to implement this functionality according to the hardware specs.
+RESULT *__catalyst__mbqc__measure_in_basis(QUBIT *wire, uint32_t plane, double angle,
```
paul0403, line 1043: I will look at it more closely next week, but just very quickly, do we just want no-op, or explicit failure, or do we want to silently fall back onto the standard computational basis measurement? The latter could be a bit confusing for users, especially considering the different semantics.

joeycarter: My goal here was to "simulate" the execution pipeline as much as possible (and not to simulate the actual physics). My thinking was that a computational-basis measurement operation will give us a better sense of cpu/memory usage than a complete no-op. Also, this isn't a user-facing feature at the moment, it falls under the umbrella of experimental ftqc work.

paul0403: Makes sense. In that case, should we at least give a warning here, saying "this function does not actually do what the name suggests"? Even this is not user-facing, we might want to make these potential points of confusion explicit.

paul0403: Not sure if we have runtime warning infrastructure, but even just printing a message here would be fine with me.

runtime/include/RuntimeCAPI.h
```diff
@@ -104,6 +104,9 @@ void __catalyst__qis__State(MemRefT_CplxT_double_1d *, int64_t, /*qubits*/...);
 void __catalyst__qis__Gradient(int64_t, /*results*/...);
 void __catalyst__qis__Gradient_params(MemRefT_int64_1d *, int64_t, /*results*/...);
 
+// MBQC operations
```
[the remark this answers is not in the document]
joeycarter: I was thinking the latter, that at this stage in development it would be better to keep all of the MBQC work contained within the `mbqc` namespace and dialect as much as possible.

@dime10 do you have any preference on the naming convention?

runtime/lib/capi/RuntimeCAPI.cpp
mlxd: Let's avoid warnings --- this isn't expected to be supported outside the team, and all failures resulting from incorrect use are fine :)

joeycarter: Sorry, posted my reply before refreshing the page and didn't see your comment! If we're happy **not** emitting warnings that a-okay with me :+1:

paul0403: I see, all good then!

runtime/include/RuntimeCAPI.h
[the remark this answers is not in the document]
dime10: I would agree with Joey here. Either way it's not a big difference, but for now I'm happy with the namespacing and we can always integrate further at a later date :)

runtime/include/RuntimeCAPI.h
```diff
@@ -104,6 +104,9 @@ void __catalyst__qis__State(MemRefT_CplxT_double_1d *, int64_t, /*qubits*/...);
...
+RESULT *__catalyst__mbqc__measure_in_basis(QUBIT *, uint32_t, double, int32_t);
+
```
paul0403, line 109: Nitpick: this organization makes the stub below (`unrecoverable_error()`) look like an MBQC stub.

runtime/tests/Test_NullQubit.cpp
```diff
@@ -558,4 +558,34 @@ TEST_CASE("Test NullQubit device shots methods", "[NullQubit]")
...
\ No newline at end of file
+}
+
+// -------------------------------------------------------------------------- //
+// MBQC Runtime Tests
+// -------------------------------------------------------------------------- //
+
+TEST_CASE("Test __catalyst__mbqc__measure_in_basis, device=null.qubit", "[MBQC]")
+{
+    __catalyst__rt__initialize(nullptr);
```
paul0403, line 569: I am unsure whether this file is suited for this test. 

Looking at all the tests above, we see that this file is testing the `QuantumDevice` interface of the null qubit device itself. In other words, this file is testing the implementation of null qubit.

On the other hand, the test you are adding is a test on a feature, and null qubit is simply a vehicle of that feature. 

I don't actually know why we don't have tests for the capi stubs themselves (@dime10 ) . Although, I see some stub tests in this file too, so I am not so sure. Maybe this is fine?

runtime/include/RuntimeCAPI.h
joeycarter: I've added a header above `__catalyst__host__rt__unrecoverable_error` to make it clear that belongs to another section.

0cd06062

doc/releases/changelog-dev.md
```diff
@@ -68,6 +68,11 @@
   [(#1663)](https://github.com/PennyLaneAI/catalyst/pull/1663)
   [(#1679)](https://github.com/PennyLaneAI/catalyst/pull/1679)
 
+* The Catalyst Runtime C-API now includes a stub for the `mbqc.measure_in_basis` operation,
+  `__catalyst__mbqc__measure_in_basis()`, allowing for execution of MBQC workloads containing
+  parameterized arbitrary-basis measurements.
```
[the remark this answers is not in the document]
joeycarter: I think that's fair. I've updated the changelog to make it clear that `__catalyst__mbqc__measure_in_basis` is for mock execution only and that internally it performs a computational-basis measurement.

94eddb3a

runtime/tests/Test_NullQubit.cpp
joeycarter: I think that's fair. I put the test here mainly for convenience, since, like you say, we don't currently have a dedicated module to test each of the C-API stubs. We also exclusively test some general stubs in this file as well, like you pointed out (`__catalyst__qis__Gradient_params`, for instance). But I agree, now would be a good time grouping related tests of runtime stubs into their own module. I'll make the change :+1:

doc/releases/changelog-dev.md
```diff
@@ -69,6 +69,14 @@
   [(#1663)](https://github.com/PennyLaneAI/catalyst/pull/1663)
   [(#1679)](https://github.com/PennyLaneAI/catalyst/pull/1679)
 
+* The Catalyst Runtime C-API now includes a stub for the experimental `mbqc.measure_in_basis`
+  operation, `__catalyst__mbqc__measure_in_basis()`, allowing for mock execution of MBQC workloads
+  containing parameterized arbitrary-basis measurements.
+  [(#1674)](https://github.com/PennyLaneAI/catalyst/pull/1674)
```
paul0403, line 75: The link should come at the end of the changelog entry no? : )

runtime/tests/Test_MBQC.cpp
```diff
@@ -0,0 +1,49 @@
+
```
paul0403, line 1: There's an unnecessary empty line here

paul0403 (approved):
Nice work! :partying_face: :balloon:

doc/releases/changelog-dev.md
joeycarter: The style we've always used is to put the link after the first block of text, and then supplemental information related to the change after that (see here for examples: https://docs.pennylane.ai/projects/catalyst/en/stable/dev/release_notes.html)

runtime/tests/Test_MBQC.cpp
joeycarter: Agh! Thanks for pointing this out! Fixed: 909a0ef9

(I also fixed the copyright date)

doc/releases/changelog-dev.md
paul0403: Yes you are right, I must have forgotten : )

joeycarter: No worries :wink:

PR #1676, opened 2025-04-23
update wires for HybridOp

albi3ro:
**Context:**

In my PR where I can the default `Operator.num_wires = None`: https://github.com/PennyLaneAI/pennylane/pull/7312

I get failing CI:

https://github.com/PennyLaneAI/pennylane/actions/runs/14602060847/job/40962850893?pr=7312

with errors:
```
TypeError: Must specify a set of wires. None is not a valid wire label.
```

Instead of having a wire label that is `Operator.num_wires`, I'm just giving `HybridOp` empty wires.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-89159]

### DIFF
--- a/frontend/catalyst/_version.py
+++ b/frontend/catalyst/_version.py
@@ -16,4 +16,4 @@
 Version number (major.minor.patch[-label])
 """
 
-__version__ = "0.12.0-dev16"
+__version__ = "0.12.0-dev17"


.dep-versions
```diff
@@ -8,7 +8,7 @@ enzyme=v0.0.149
 
 # For a custom PL version, update the package version here and at
 # 'doc/requirements.txt
-pennylane=0.42.0-dev12
+pennylane=0.42.0-dev14
```
albi3ro, line 11:
```suggestion
pennylane=0.42.0-dev15
```

dime10:
Doesn't look like the braket tests were removed in your last commit, but also there appears to be some braket tests in the lit test suite 🤔 Let me take a look at those

dime10:
Ah, seems like the lit tests doesn't actually need the device to be braket:
```
"""
# TODO: The only reason we are using the braket.local.qubit device here
# is because this test was developed before having support for custom devices.
# We should replace instead create a custom device that has support for ISWAP
# and PSWAP (which I think are unsupported in lightning.qubit and hence why they
# would be decomposed.
"""
```

dime10:
@albi3ro Can you try this for the lit test? https://gist.github.com/dime10/249284a10a9103556ac3555964a13366

dime10:
For the main test suite, we might not be lenient enough upon import:
```py
try:
    qml.device("braket.local.qubit", backend="default", wires=1)
except exceptions.DeviceError:
    pytest.skip(
        "skipping Braket local tests because ``amazon-braket-pennylane-plugin`` is not installed",
        allow_module_level=True,
    )
```
we could just blanket catch instead I suppose? 🤔

dime10 (approved):
Thanks Christina!

PR #1678, opened 2025-04-23
Change logic around *when* standalone plugin is built

rashidnhm:
**Context:**
Currently, the nightly build is failing due to the standalone plugin being built and uploaded, where it shouldn't be built in the first place.

Issue: When a github action is called using `workflow_call` the `github.event_name` is not updated. This is something I was unaware of prior to this.

A good question now would be, why not just align the arm64 workflow if-statements to that of the other workflows, the expression being `github.event.scheulde == '35 4 * * 4'`. The rationale explained in the Benefits section.

This PR updates the logic around when we build the standalone plugins.

A new input is added to the four `build-wheel-` workflows. The input being `build_standalone_plugin`, and it defaults to false.

Then we set an environment variable `BUILD_STANDALONE_PLUGIN` based on the following logic:
- Does the input exist? 
    - If yes, use the value of the input to determine if standalone plugin build needed or not
    - If no, go to next
- Is the event_name 'schedule'?
    - If yes, build standalone plugin
    - If no, return false.

This environment variable is then used in the steps to determine if the step should be run or not.

**Benefits:**
A minor benefit of this is that the steps themselves are more readable, in that, `env.BUILD_STANDALONE_PLUGIN == 'true'` is easy to read and know what the goal is.

The bigger, benefit is that it is best that we do not tie ourselves to what the actual cron schedule is to build the standalone plugin. Though I am sure it is a value that doesn't change too frequently, with the current implementation, a change to the schedule would require re-writing all the if-expressions, moreso, what if we wanted to start building the standalone plugin twice a week? Or more?

The implementation proposed here moves us away from having to rely on the value(s) of the cron schedule.

Another benefit is that we can opt-in to building standalone plugin when we call these workflows from workflow_displatch or workflow_call.

**Possible Drawbacks:**
Possibility of bugs, but I have tested a build [here](https://github.com/PennyLaneAI/catalyst/actions/runs/14619875471) which finished without errors.

**Related GitHub Issues:**
[sc-89680]

### DIFF
--- a/.github/workflows/build-wheel-linux-arm64.yaml
+++ b/.github/workflows/build-wheel-linux-arm64.yaml
@@ -35,7 +35,7 @@ env:
   # If the `inputs.build_standalone_plugin` is set to a value (true/false), then that is used.
   # If the input is empty (meaning this workflow was not triggered by a workflow_call/workflow_dispatch), then check if event_name is schedule.
   # If event_name is NOT schedule, default to false
-  BUILD_STANDALONE_PLUGIN: ${{ inputs.build_standalone_plugin != '' && format('{0}', inputs.build_standalone_plugin) || github.event_name == 'schedule' && 'true' || 'false' }}
+  BUILD_STANDALONE_PLUGIN: ${{ format('{0}', inputs.build_standalone_plugin) || github.event_name == 'schedule' || false }}
 
 jobs:
   check_if_wheel_build_required:
--- a/.github/workflows/build-wheel-linux-arm64.yaml
+++ b/.github/workflows/build-wheel-linux-arm64.yaml
@@ -17,6 +17,10 @@ on:
     inputs:
       build_standalone_plugin:
         description: 'Build the standalone plugin wheel'
+        type: choice
+        options:
+          - 'true'
+          - 'false'
         required: false
         default: 'false'
   workflow_call:
--- a/.github/workflows/build-wheel-linux-arm64.yaml
+++ b/.github/workflows/build-wheel-linux-arm64.yaml
@@ -35,7 +35,7 @@ env:
   # If the `inputs.build_standalone_plugin` is set to a value (true/false), then that is used.
   # If the input is empty (meaning this workflow was not triggered by a workflow_call/workflow_dispatch), then check if event_name is schedule.
   # If event_name is NOT schedule, default to false
-  BUILD_STANDALONE_PLUGIN: ${{ inputs.build_standalone_plugin || github.event_name == 'schedule' || 'false' }}
+  BUILD_STANDALONE_PLUGIN: ${{ (inputs.build_standalone_plugin == true || inputs.build_standalone_plugin == 'true') && 'true' || (github.event_name == 'schedule' && 'true') || 'false' }}
 
 jobs:
   check_if_wheel_build_required:
--- a/.github/workflows/build-wheel-linux-arm64.yaml
+++ b/.github/workflows/build-wheel-linux-arm64.yaml
@@ -39,7 +39,7 @@ env:
   # If the `inputs.build_standalone_plugin` is set to a value (true/false), then that is used.
   # If the input is empty (meaning this workflow was not triggered by a workflow_call/workflow_dispatch), then check if event_name is schedule.
   # If event_name is NOT schedule, default to false
-  BUILD_STANDALONE_PLUGIN: ${{ format('{0}', inputs.build_standalone_plugin) || github.event_name == 'schedule' || false }}
+  BUILD_STANDALONE_PLUGIN: ${{ format('{0}', inputs.build_standalone_plugin) || github.event_name == 'schedule' }}
 
 jobs:
   check_if_wheel_build_required:
--- a/.github/workflows/build-wheel-linux-arm64.yaml
+++ b/.github/workflows/build-wheel-linux-arm64.yaml
@@ -14,11 +14,28 @@ on:
     # Thursdays we test the standalone plugin
     - cron: '35 4 * * 4'
   workflow_dispatch:
+    inputs:
+      build_standalone_plugin:
+        description: 'Build the standalone plugin wheel'
+        required: false
+        default: 'false'
   workflow_call:
+    inputs:
+      build_standalone_plugin:
+        description: 'Build the standalone plugin wheel'
+        type: boolean
+        required: false
+        default: false
 
 concurrency:
   group: Build Catalyst Wheel on Linux (arm64)-${{ github.ref }}
   cancel-in-progress: true
+
+env:
+  # If the `inputs.build_standalone_plugin` is set to a value (true/false), then that is used.
+  # If the input is empty (meaning this workflow was not triggered by a workflow_call/workflow_dispatch), then check if event_name is schedule.
+  # If event_name is NOT schedule, default to false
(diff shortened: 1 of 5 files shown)


.github/workflows/build-wheel-linux-arm64.yaml
```diff
@@ -14,12 +14,29 @@ on:
...
 concurrency:
   group: Build Catalyst Wheel on Linux (arm64)-${{ github.ref }}
   cancel-in-progress: true
 
+env:
+  # If the `inputs.build_standalone_plugin` is set to a value (true/false), then that is used.
+  # If the input is empty (meaning this workflow was not triggered by a workflow_call/workflow_dispatch), then check if event_name is schedule.
+  # If event_name is NOT schedule, default to false
+  BUILD_STANDALONE_PLUGIN: ${{ inputs.build_standalone_plugin || github.event_name == 'schedule' || 'false' }}
+
```
dime10, line 39: I'm confused by this variable, isn't it the same as before and still entirely dependent on whether the event trigger is a schedule?

.github/workflows/build-wheel-linux-arm64.yaml
```diff
@@ -14,12 +14,29 @@ on:
...
     - cron: '35 4 * * 4'
   workflow_dispatch:
+    inputs:
+      build_standalone_plugin:
+        description: 'Build the standalone plugin wheel'
+        required: false
+        default: 'false'
   workflow_call:
+    inputs:
+      build_standalone_plugin:
```
dime10, line 24: We don't seem to set this input in another action, is it just there for the future?

.github/workflows/build-wheel-linux-arm64.yaml
dime10: Oh wait, having a value regardless of true or false uses that value in a github or (`||`) statement?

rashidnhm: > I'm confused by this variable, isn't it the same as before and still entirely dependent on whether the event trigger is a schedule?


It let's us move away from having to rely on the *value* of the schedule, which is what we currently rely on to determine if the standalone plugin should be built or not

rashidnhm: > Oh wait, having a value regardless of true or false uses that value in a github or (||) statement?

The possible values for the input is `true`/`false`/`undefined (or NULL)`, the third value is when this workflow is not called from workflow_dispatch or workflow_call, so then the github or `||` goes to the next value

rashidnhm: Ah I see, we can account enforce a type a little better, no problem!

dime10: I'm still confused, can you confirm if this is true in github or not?
> what I'm wondering is if `false || ...` just evaluates to false

and whether that is what this comment is implying?
> If the `inputs.build_standalone_plugin` is set to a value (true/false), then that is used.

frontend/catalyst/_version.py
```diff
@@ -16,4 +16,4 @@
 Version number (major.minor.patch[-label])
 """
 
-__version__ = "0.12.0-dev12"
+__version__ = "0.12.0-dev13"
```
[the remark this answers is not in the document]
rashidnhm: This was ringo :sweat_smile: ... I can revert it but chances are it will make the changes again

.github/workflows/build-wheel-linux-arm64.yaml
rashidnhm: Yeah if the value is false (as a boolean), the OR would indeed go to the next expression. I have updated the logic so any value for the input is casted to string, where any value would be used as it is truth-y.

Apologies for the mishap there

frontend/catalyst/_version.py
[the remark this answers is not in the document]
erick-xanadu: Did ringo pushed to test pypi when this change was added?

.github/workflows/build-wheel-linux-arm64.yaml
dime10: Sorry I was just wondering if you intended this value to be set by another workflow (on call), but now I understand you primarily want the value to be set at all and contrast it when it is unset.

.github/workflows/build-wheel-macos-arm64.yaml
```diff
@@ -14,11 +14,27 @@ on:
...
+        description: 'Build the standalone plugin wheel'
+        required: false
+        default: 'false'
   workflow_call:
+    inputs:
+      build_standalone_plugin:
+        description: 'Build the standalone plugin wheel'
+        type: boolean
+        required: false
+        default: false
```
erick-xanadu, line 28: Here, one default value is string false, the other is bool false. Just a comment, not sure if it is intended.

.github/workflows/build-wheel-linux-arm64.yaml
rashidnhm: @dime10 I (think) the issue is resolved now.

This is now the behavior as I have tested:
- `true` and `'true'` gets casted to string value `'true'`
- `false` and `'false'` gets casted to string value `'false'`
- The input being missing gets casted to string value `''`

The first two values are truth-y and the third value is false-y.

This yields the expected result stated in PR description and in the comment.

Let me know if you have further concerns.

frontend/catalyst/_version.py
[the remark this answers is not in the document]
rashidnhm: I am not sure what you mean, normally does ringo not auto-bump the version for PRs?

.github/workflows/build-wheel-macos-arm64.yaml
rashidnhm: It is intended, and it's the reason for all the headaches in the above comment.

For workflow_dispatch, the value (as you speicfy within GitHub UI) *always* ends up as string value. Hence why the default is also `'false'`.

For workflow_call, since it is being called by another workflow do support string/boolean as a `type`. However, if we use string, we cannot restrict it to `true` / `false` and it might be confusing in the future from a calling workflow on why it's passing a boolean but it's being treated otherwise here.

frontend/catalyst/_version.py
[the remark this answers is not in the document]
dime10: We only bump versions on nightly uploads. If you ran the nightly upload from your branch, and it did indeed successfully upload, your PR should bump the version on merge to main as well, otherwise the next nightly upload will fail.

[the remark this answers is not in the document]
rashidnhm: Ah yes I did run nightly build on this branch. Yes this is expected then.

.github/workflows/build-wheel-linux-arm64.yaml
dime10: Nice I think that looks pretty sensible! :) 

Would you mind explaining what the `|| false` does in that line?

rashidnhm: Done!

dime10 (approved):
Looks good to me 🎉 Thanks for all the help!

.github/workflows/build-wheel-linux-arm64.yaml
rashidnhm: @dime10 good to mark this as resolved?

erick-xanadu (approved):

PR #1685, opened 2025-04-24
Update LAPACK kernels to account for JAX's XLA FFI migration

mehrdad2m:
**Context:** 
JAX has migrated their LAPACK kernels to use XLA's FFI interface, which relies on XLA internal methods to call external functions (see [this Issue page](https://github.com/jax-ml/jax/issues/23056)). As a result, they have updated the lowering to stableHLO custom_call by removing some of the arguments by either moving them to an Attribute called `backend_config` or removing altogether (args related to matrix dimensions) and deciphering them in the kernel argument using XLA interface.

Note that since Catalyst is not using an XLA runtime, it is not possible to fully use the FFI interface. 

- [x] Update StableHLO custom call lowering to add back deciphered dimensions and `backend_config` fields to arguments.
- [x] Update Cholesky Factorization kernels
- [x] Update LU Decomposition kernels 
- [x] Update Singlar Value Decomposition kernels
- [x] Update QR Factorization kernels
- [x] Update Householder Product kernels
- [x] Update Schur Decomposition kernels
- [x] Update Triangular System Solver kernels
- [x] Update Eigenvalue Decomposition (General) kernels
- [x] Update Eigenvalue Decomposition (Symmetric) kernels
- [x] Update Hessenberg Decomposition kernels
- [x] Remove Tridiagonal Reduction kernels as it is not used by any tests

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-88694]


mlir/lib/Catalyst/Transforms/HloCustomCallPatterns.cpp
```diff
@@ -19,29 +19,120 @@
...
+            }
+
+            // Extract tensor type and dimensions
+            if (auto tensorType = dyn_cast<RankedTensorType>(operands[0].getType())) {
+                auto shape = tensorType.getShape();
+                auto rank = shape.size();
+
+                if (rank < 2 ||
+                    llvm::any_of(shape, [](int64_t d) { return d == ShapedType::kDynamic; }))
+                    return failure(); // Bail out on dynamic shapes
```
mlxd, line 101:
```suggestion
                if (rank < 2 ||
                    llvm::any_of(shape, [](int64_t d) { return d == ShapedType::kDynamic; })){
                    return failure(); // Bail out on dynamic shapes
                }
```
Just a fly-by on brace use :) https://github.com/PennyLaneAI/guidance-docs/blob/master/development/cpp-dev-guidelines.md#brace-style
Never skip on them for one liners as they can be a common source of bugs.

mlir/lib/Catalyst/Transforms/HloCustomCallPatterns.cpp
```diff
@@ -19,29 +19,120 @@
...
+
+                // Add batch count and matrix dimensions as new operands
+                int64_t batch = 1;
+                for (size_t i = 0; i < rank - 2; ++i)
+                    batch *= shape[i];
```
mlxd, line 106: Same here and elsewhere :)

mlir/lib/Catalyst/Transforms/HloCustomCallPatterns.cpp
mehrdad2m: Ah yes good point! thanks for pointing that out! I usually try to respect this but sometime old habits come back! :)

frontend/catalyst/utils/jax_cpu_lapack_kernels/lapack_kernels.cpp
```diff
@@ -63,43 +63,27 @@
...
-    const int m = *reinterpret_cast<int32_t *>(data[4]);
-    const int n = *reinterpret_cast<int32_t *>(data[5]);
-    const int batch = *reinterpret_cast<int32_t *>(data[6]);
-    const T alpha = *reinterpret_cast<T *>(data[7]);
-    const T *a = reinterpret_cast<T *>(data[8]);
-    T *b = reinterpret_cast<T *>(data[9]);
+    const char diag = *(reinterpret_cast<uint8_t *>(data[0]));
+    const char side = *(reinterpret_cast<uint8_t *>(data[1]));
+    const char trans = *(reinterpret_cast<uint8_t *>(data[2]));
+    const char uplo = *(reinterpret_cast<uint8_t *>(data[3]));
```
joeycarter, line 80: Could it be an issue that we're `reinterpret_cast`ing to `uint8_t` types here (and elsewhere), but initializing the variables as `char`s? Or is this just what JAX does now? I bring it up because `uint8_t` is an `unsigned char`, but the signedness of `char` is ambiguous. See for example the Notes section [here](https://en.cppreference.com/w/cpp/types/integer):

> `std::int8_t` may be `signed char` and `std::uint8_t` may be `unsigned char`, but neither can be `char` regardless of its signedness (because `char` is not considered a "signed integer type" or "unsigned integer type").

frontend/catalyst/utils/jax_cpu_lapack_kernels/lapack_kernels.cpp
```diff
@@ -435,32 +406,25 @@ template <typename T> typename ComplexGesdd<T>::FnType *ComplexGesdd<T>::fn = nu
...
 void ComplexGesdd<T>::Kernel(void *out_tuple, void **data, XlaCustomCallStatus *)
 {
-    const int32_t job_opt_full_matrices = *(reinterpret_cast<int32_t *>(data[0]));
-    const int32_t job_opt_compute_uv = *(reinterpret_cast<int32_t *>(data[1]));
-    const int b = *(reinterpret_cast<int32_t *>(data[2]));
-    const int m = *(reinterpret_cast<int32_t *>(data[3]));
-    const int n = *(reinterpret_cast<int32_t *>(data[4]));
-    const int lwork = *(reinterpret_cast<int32_t *>(data[5]));
-    T *a_in = reinterpret_cast<T *>(data[6]);
+    const char jobz = *(reinterpret_cast<int32_t *>(data[0]));
```
joeycarter, line 409: Is this right? `*reinterpret_cast<int32_t *>` to a variable with type `char`?

frontend/catalyst/utils/jax_cpu_lapack_kernels/lapack_kernels.cpp
```diff
@@ -482,33 +446,33 @@ template struct RealGesdd<double>;
...
 
 template <typename T> typename RealSyevd<T>::FnType *RealSyevd<T>::fn = nullptr;
 
 template <typename T> void RealSyevd<T>::Kernel(void *out_tuple, void **data, XlaCustomCallStatus *)
 {
-    const int32_t lower = *(reinterpret_cast<int32_t *>(data[0]));
-    const int b = *(reinterpret_cast<int32_t *>(data[1]));
-    const int n = *(reinterpret_cast<int32_t *>(data[2]));
-    const T *a_in = reinterpret_cast<T *>(data[3]);
+    const char jobz = *(reinterpret_cast<int32_t *>(data[0]));
```
joeycarter, line 456: Same here.

frontend/catalyst/utils/jax_cpu_lapack_kernels/lapack_kernels.cpp
```diff
@@ -523,27 +487,26 @@ template <typename T> typename ComplexHeevd<T>::FnType *ComplexHeevd<T>::fn = nu
 template <typename T>
 void ComplexHeevd<T>::Kernel(void *out_tuple, void **data, XlaCustomCallStatus *)
 {
-    const int32_t lower = *(reinterpret_cast<int32_t *>(data[0]));
-    const int b = *(reinterpret_cast<int32_t *>(data[1]));
-    const int n = *(reinterpret_cast<int32_t *>(data[2]));
-    const T *a_in = reinterpret_cast<T *>(data[3]);
+    const char jobz = *(reinterpret_cast<int32_t *>(data[0]));
```
joeycarter, line 490: And here.

frontend/catalyst/utils/jax_cpu_lapack_kernels/lapack_kernels.cpp
```diff
@@ -889,53 +846,4 @@ template struct Gehrd<double>;
 template struct Gehrd<std::complex<float>>;
 template struct Gehrd<std::complex<double>>;
 
-// Sytrd
-// ~~~~~
```
joeycarter, line 893: Did JAX remove sytrd entirely?

joeycarter (commented):
Thanks @mehrdad2m! Just a few quick questions for now. Let me know when you've got the segfaults in CI fixed and I can take another look.

And don't forget the changelog :wink:

frontend/catalyst/utils/jax_cpu_lapack_kernels/lapack_kernels.cpp
mehrdad2m: No this not exactly how jax does it. They are now using xla ffi interface to take care of attributes. However, I don't see why this cast would cause a problem. I can change this code to be more explicit though.

joeycarter: If that's how JAX is doing it then it's probably not an issue, just wanted to check! Feel free to resolve if you decide that no changes are required.

frontend/catalyst/utils/jax_cpu_lapack_kernels/lapack_kernels.cpp
mehrdad2m: They have not, but I didn't find any test in Catalyst that triggers the use of that function, and also was not able to trigger it with a new test, so I removed it.

frontend/catalyst/utils/jax_cpu_lapack_kernels/lapack_kernels.cpp
mehrdad2m: I made the type casts a bit more explicit now. Also I think I have fixed the issue, which was not related to this but another data layout issue that I was not seeing locally due to jax version mismatch. Hopefully the CI will pass now.

frontend/catalyst/utils/jax_cpu_lapack_kernels/lapack_kernels.cpp
```diff
@@ -63,43 +63,31 @@
 
 namespace {
```
paul0403, line 64: This namespace is not necessary now right? :sweat_smile:

frontend/catalyst/utils/jax_cpu_lapack_kernels/lapack_kernels.cpp
joeycarter: I took a quick look through the JAX repository. It looks like they don't support any high-level linalg functions that call this lapack routine, so yes we should be safe removing it :+1:

joeycarter (approved):
Looks good, thanks @mehrdad2m ! :rocket: 

Just need to remove the empty namespace that @paul0403 pointed out, but then you should be good to merge this in. Otherwise happy to approve.

paul0403 (approved):

PR #1689, opened 2025-04-25
Enable oqd pytests in PR CI

paul0403:
**Context:**
It was discovered that for oqd tests, CIs on PRs only run the cpp runtime device tests, not the pytests. 

We enable oqd pytests in CI. 

**Benefits:**
Better test coverage.

### DIFF
--- a/.github/workflows/check-catalyst.yaml
+++ b/.github/workflows/check-catalyst.yaml
@@ -611,6 +611,58 @@ jobs:
       run: |
         make pytest TEST_BRAKET=LOCAL
 
+  frontend-tests-oqd-device:
+    name: Frontend Tests (backend="oqd")
+    needs: [constants, llvm, runtime, quantum, determine_runner]
+    runs-on: ${{ needs.determine_runner.outputs.runner_group }}
+    strategy:
+      matrix:
+        compiler: ${{ fromJson(needs.constants.outputs.compilers) }}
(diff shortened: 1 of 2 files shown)


paul0403:
Since this is just purely CI and testing, I don't think a changelog entry is needed.

.github/workflows/check-catalyst.yaml
```diff
@@ -611,6 +611,58 @@ jobs:
       run: |
         make pytest TEST_BRAKET=LOCAL
 
+  frontend-tests-oqd-device:
```
dime10, line 614: Can't you just set the appropriate flag? No need for an extra job.

https://github.com/PennyLaneAI/catalyst/blob/2bd942dc9b0f3bb8bbd24927f3e3a7c934e90c1e/Makefile#L175

paul0403: Isn't the job tied to the device? We can sneak the tests into the lightning device job but that feels a bit off.

dime10: Hmm, maybe. I mean the "main" job already runs a whole bunch of tests, not just device-specific ones: async, lit, demos, and even oqc. It's really just kokkos and braket that have their own jobs.

dime10 (approved):
Thanks!

PR #1698, opened 2025-04-29
Exit with success on skipped braket device pytests

paul0403:
**Context:**
We recently skipped all the braket pytests. However, It was discovered that `pytest.skip(reason="...", allow_module_level=True)` will not collect the tests after skipping an entire pytest file, creating a pytest exit code of 5 (which signals a failure).

We force a successful exit code after running the braket pytests in the wheels scripts. 

**Benefits:**
CI does not fail anymore.

### DIFF
--- a/frontend/test/pytest/test_braket_remote_devices.py
+++ b/frontend/test/pytest/test_braket_remote_devices.py
@@ -29,10 +29,7 @@ try:
     )
 except Exception as e:
     # To capture all exceptions associated with AWS credentials or PL plugins
-    pytest.skip(
-        f"skipping Braket remote tests because {e}",
-        allow_module_level=True,
-    )
+    pytestmark = pytest.mark.skip(f"skipping Braket remote tests because {e}")
 
 
 @pytest.mark.braketremote
--- a/frontend/test/pytest/test_braket_remote_devices.py
+++ b/frontend/test/pytest/test_braket_remote_devices.py
@@ -29,7 +29,10 @@ try:
     )
 except Exception as e:
     # To capture all exceptions associated with AWS credentials or PL plugins
-    pytestmark = pytest.mark.skip(f"skipping Braket remote tests because {e}")
+    pytest.skip(
+        f"skipping Braket remote tests because {e}",
+        allow_module_level=True,
+    )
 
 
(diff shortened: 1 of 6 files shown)


frontend/test/pytest/test_braket_remote_devices.py
```diff
@@ -29,10 +29,7 @@
     )
 except Exception as e:
     # To capture all exceptions associated with AWS credentials or PL plugins
-    pytest.skip(
-        f"skipping Braket remote tests because {e}",
-        allow_module_level=True,
-    )
+    pytestmark = pytest.mark.skip(f"skipping Braket remote tests because {e}")
```
dime10, line 32: I think doing module-level here is fine, it will save a small amount of time since we never want to run these on CI anyway.

dime10 (approved):
Thanks Paul!

frontend/test/pytest/test_braket_remote_devices.py
paul0403: https://github.com/PennyLaneAI/catalyst/pull/1698/commits/44fe35e3da70b06fcf97cdc63ba89947e587a77d

dime10 (approved):
Thanks, this should allow the wheels to pass again :)

The only downside I can see is we won't be alerted of future failures of the Braket integration across platforms. @josh146 What do you think about this in the long run?

paul0403:
> The only downside I can see is we won't be alerted of future failures of the Braket integration across platforms. @josh146 What do you think about this in the long run?

While this decision is being made, is it ok if we merge this in first? I think @mehrdad2m told me there's a bug-fix release blocked by this wheels failure.

dime10:
> > The only downside I can see is we won't be alerted of future failures of the Braket integration across platforms. @josh146 What do you think about this in the long run?
> 
> While this decision is being made, is it ok if we merge this in first? I think @mehrdad2m told me there's a bug-fix release blocked by this wheels failure.

That's why I approved ;)

josh146:
> The only downside I can see is we won't be alerted of future failures of the Braket integration across platforms. @josh146 What do you think about this in the long run?

Perhaps we could set up a cron job to run these tests separately from PRs?

PR #1710, opened 2025-05-02
Fix failing wheels due to missing mbqc dialect module

joeycarter:
**Context:** The wheels were failing to build correctly due to the mbqc dialect python module not being available (e.g. https://github.com/PennyLaneAI/catalyst/actions/runs/14797406520)

**Description of the Change:** Adds the mbqc dialect module to list of files to copy when building the wheels.


dime10 (approved):
Thanks!

PR #1712, opened 2025-05-02
Add support for debugging PennyLane (Python) and Catalyst (C++) simultaneously 

mlxd:
**Context:** Debugging multi-language programming environments can be a challenge due to the intermingling and swapping between all participating languages. In the case of PennyLane and Catalyst, this is a mix of Python and C++, through our own libraries all the way through LLVM and back. Interactive debugging of a multi-language environment can require some custom setup to ensure the environment allows for useful interactive data inspection and validation.

VSCode provides support for [mixed-mode debugging](https://learn.microsoft.com/en-us/visualstudio/debugger/how-to-debug-in-mixed-mode?view=vs-2022) where a single editor can support (through the Python and C++ debugging extensions) interactive breakpointing, stepping, and inspection of the running program, where the view can switch between the Python debugger and the C++ debugger, once running. This PR adds preliminary support for this mixed-mode debugging, allowing the user to set breakpoints in their PennyLane program their Catalyst (+/- LLVM) internals, and step all the way through to investigate all aforementioned stages. 

Some notes:
* Use of this functionality requires building Catalyst with debug symbols. This can be achieved via `make all BUILD_TYPE="RelWithDebInfo"`, and may generated approximately 80GB of libraries and binaries. Added internal (`BUILD_TYPE`) and external (`BUILD_TYPE_EXT`) configuration for build-flags to restrict debug symbols to our libs and bins only.
* Launching the C++ debugger requires attaching to a running process. This requires `sudo` privileges.
* The spawned compiler subprocess requires receipt of a `SIGCONT` to continue executing **after** the C++ debugger has attached.
* ~~Setting the active `qml.qjit(debug_compiler=True)` option will not enable the support unless the Python process is called from a debugger (pdb, debugby, etc)~~ This restriction has been removed.
 
**Description of the Change:** As above.

**Benefits:** Enables debugging of mixed mode (Python + C++/LLVM) programs directly through VSCode.

**Possible Drawbacks:**

**Related GitHub Issues:**


frontend/catalyst/debug/__init__.py
```diff
@@ -38,4 +39,5 @@
     "instrumentation",
     "replace_ir",
     "compile_executable",
+    "is_debugger_active",
```
dime10, line 42: This module exposes the user facing API (the stuff that renders in the docs), do we expect this function is something a user would use or an internal function?

dime10 (commented):
This is awesome!!! Thanks Lee :)

mlxd:
> > Use of this functionality requires building Catalyst with debug symbols. This can be achieved via make all BUILD_TYPE="RelWithDebInfo", and may generated approximately 80GB of libraries and binaries.
> 
> 
> 
> To mitigate one could avoid building llvm itself with debug symbols, the catalyst code base itself is not that bad in debug mode 🤔 

This is a good point; I may try to restrict the debug symbols to the Catalyst additions and see what savings we get.

frontend/catalyst/debug/__init__.py
mlxd: Ah, good point yea. I would say we can keep this internal.

.vscode/launch.json
```diff
@@ -0,0 +1,28 @@
...
+            "console": "integratedTerminal",
+            "justMyCode": false 
+        },
+        {
+            "name": "(gdb): Attach To Python Process",
+            "type": "cppdbg",
+            "request": "attach",
+            "program": "${command:python.interpreterPath}",
+            "processId": "${command:pickProcess}",
+            "MIMode": "gdb",
```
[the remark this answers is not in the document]
mlxd: A good point too. There's a few trains of thoughts on this I've read in the wild. For a unified env across a team, many check these configs into the repo so they are immediately ready. Though, as you can imagine it will override custom launch files a user may have, of which we more likely have -- I don't think any unified dev env exists with our dev process.

For now, as I'm setting this up for ftqc I think I'll keep it in the wip, but shifting to docs and/or a generating file with the content could be better long term. Alternatively, this is also another good avenue for the docker vscode dev env integration.

We can chat more about what's best once it's closer to going in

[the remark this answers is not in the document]
dime10: > Alternatively, this is also another good avenue for the docker vscode dev env integration.

Ah yeah, I was just thinking of the same thing. We did have those in place (and apparently already removed it again), [like here](https://github.com/PennyLaneAI/catalyst/blob/v0.7.0/.devcontainer/dev/devcontainer.json). I guess no one was using them and so they weren't maintained, but the nice thing about those is all the vscode config was readily loadable but optional, and wouldn't conflict with existing local configurations. Maybe we could give those another shot? 🤔

frontend/catalyst/compiler.py
```diff
@@ -438,15 +440,31 @@ def run_from_ir(self, ir: str, module_name: str, workspace: Directory):
...
                 print(f"[SYSTEM] {' '.join(cmd)}", file=self.options.logfile)
-            result = subprocess.run(cmd, check=True, capture_output=True, text=True)
+
+            with subprocess.Popen(
+                cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True
+            ) as p:
+                if p.returncode not in {0, None}:
+                    raise subprocess.CalledProcessError(p.returncode, cmd)
+
+                if self.options.debug_compiler and is_debugger_active():
```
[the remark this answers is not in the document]
dime10: So in my mind case 1 could easily be served with case 2 no, and the user just doesn't step into the c++ call? Or is there a scenario where you would get an error (let's say you didn't build the C++ source in debug mode), while only caring about debugging the frontend?

Good point about case 3, that is a little awkward since setting the flag won't do anything without also launching the debugger.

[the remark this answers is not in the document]
mlxd: Came back to this, and removed the `is_debugger_active` check. This way the C++ debugger can be attached within the editor, or external to it if the flag is enabled.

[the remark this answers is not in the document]
mlxd: Given the utility of the python debugger check, I've decided to leave this within the debug module as a free-function.

.vscode/launch.json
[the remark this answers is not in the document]
mlxd: I've decided to add these configurations to the debugger docs instead of checking them in here. That way, we can avoid overwriting any user-defined configurations.

I'll remove the files from this PR

frontend/catalyst/debug/__init__.py
mlxd: Done

frontend/catalyst/jit.py
```diff
@@ -96,6 +96,7 @@ def qjit(
     circuit_transform_pipeline=None,
     pass_plugins=None,
     dialect_plugins=None,
+    debug_compiler=False
```
dime10, line 99: Personally I would have preferred not adding another keyword to this decorator, but it's not the end of the world 😌
The automatic detection based on the already active Python compiler session was actually really neat in that way.

doc/dev/debugging.rst
```diff
@@ -388,3 +388,84 @@ corresponding arguments.
...
+
+Mixed-mode debugging of Python and C++
+======================================
+
+Catalyst supports mixed-mode debugging of Python and/or C++ code when providing the ``debug_compiler=True`` flag to 
+the ``@qjit`` decorator. Enabling this option signals to the compiler to wait for an appropriate user-provided signal
+after launching the compiler process. Some notes about use of this support:
+
+* This functionality requires building Catalyst with debug symbols. This can be achieved via make all 
+  ``BUILD_TYPE="RelWithDebInfo"``. The debug symbols are only within available within the Catalyst-owned targets.
```
dime10, line 400:
```suggestion
* This functionality requires building Catalyst with debug symbols. This can be achieved via
  ``make all BUILD_TYPE="RelWithDebInfo"``. The debug symbols are only within available
  within the Catalyst-owned targets. 
```

doc/dev/debugging.rst
```diff
@@ -388,3 +388,84 @@ corresponding arguments.
...
+where ``<PID>`` is the process-ID. This can also be issued from an active Python debugger session, such as through VSCode's 
+debug terminal as
+
+.. code-block:: python
+
+    import os, signal
+    os.kill(<PID>, signal.SIGCONT)
+
+To enable support from VSCode, the following configuration files can be used to add debugger configurations for Python, and 
+C++.
```
dime10, line 425: I'm trying to test this out in vscode, but I don't actually fully understand how everything is supposed to work 😅 
Maybe I can ask you some questions in person next week? :)

dime10 (commented):
Thanks Lee, I would like to test it before approving, but I'm having some difficulties 😅

frontend/catalyst/jit.py
mlxd: Yea, i'd agree it was nice -- maybe we can chat next week and see which seems better. This way allows attaching from a non-VSCode/non-Python debug session, so if people prefer to use it without, they can. But happy to defer either way.

frontend/catalyst/compiler.py
```diff
@@ -438,15 +439,31 @@ def run_from_ir(self, ir: str, module_name: str, workspace: Directory):
...
+                if p.returncode not in {0, None}:
+                    raise subprocess.CalledProcessError(p.returncode, cmd)
+
+                if self.options.debug_compiler:
+                    print(f"Compiler PID={p.pid}")
+                    print(
+                        f"""Ensure C++ debugger is attached and running before continuing with:
+                        kill -s SIGCONT {p.pid}"""
+                    )
+                    p.send_signal(signal.SIGSTOP)
```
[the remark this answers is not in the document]
mlxd: It's  good question. The assumption (from most places I've read) is that this should be (reasonably) fine. But if we 100% need to be strict on it, we can modify the process launch options to ensure the signal is the first thing that is hit.

doc/dev/debugging.rst
mlxd: Yea, the internal guide is currently the source-of-truth on set-up here. We can discuss next week and see what is portable across to here.

dime10: Reading through it right now, it's very helpful actually, got it to run now :)

doc/dev/debugging.rst
```diff
@@ -388,3 +388,85 @@ corresponding arguments.
...
+
+Mixed-mode debugging of Python and C++
+======================================
+
+Catalyst supports mixed-mode debugging of Python and/or C++ code when providing the ``debug_compiler=True`` flag to 
+the ``@qjit`` decorator. Enabling this option signals to the compiler to wait for an appropriate user-provided signal
+after launching the compiler process. Some notes about use of this support:
+
+* This functionality requires building Catalyst with debug symbols. This can be achieved via
+  ``make all BUILD_TYPE="RelWithDebInfo"``. The debug symbols are only within available
```
dime10:
```suggestion
  ``make all BUILD_TYPE="RelWithDebInfo"``. The debug symbols are only available
```

doc/dev/debugging.rst
```diff
@@ -388,3 +388,85 @@ corresponding arguments.
...
+  within the Catalyst-owned targets. 
+To enable debugging of LLVM and other associated external libraries and binaries, ensure the 
+  ``BUILD_TYPE_EXT="RelWithDebInfo"`` option is also set.
```
dime10:
```suggestion
  ``BUILD_TYPE_EXT="RelWithDebInfo"`` option is also set when building Catalyst.
```

doc/dev/debugging.rst
```diff
@@ -388,3 +388,85 @@ corresponding arguments.
...
+                "console": "integratedTerminal",
+                "justMyCode": false 
+            },
+            {
+                "name": "(C++): Attach To Executing Python Process",
+                "type": "cppdbg",
+                "request": "attach",
+                "program": "${command:python.interpreterPath}",
+                "processId": "${command:pickProcess}",
+                "MIMode": "gdb",
```
dime10, line 448: Mentioned this in the internal guide, but with the given configuration I get
```
Unable to start debugging. Launch options string provided by the project system is invalid. Unable to determine path to debugger.  Please specify the "MIDebuggerPath" option.
```

Not sure what I would put for the path though since I can't find `gdb` on my mac. `"MIMode": "lldb"` works though.

doc/dev/debugging.rst
```diff
@@ -388,3 +388,85 @@ corresponding arguments.
...
+                "request": "launch",
+                "program": "${file}",
+                "console": "integratedTerminal",
+                "justMyCode": false 
+            },
+            {
+                "name": "(C++): Attach To Executing Python Process",
+                "type": "cppdbg",
+                "request": "attach",
+                "program": "${command:python.interpreterPath}",
```
dime10, line 446: I'm confused by this parameter. I tried removing it but then VSCode complains that "program" is not specified. The thing is though, the process we are attaching to is not running Python, it is running the `catalyst` program (the compiler). Further, the value really doesn't seem to matter as long it's a valid path because I can put `/bin/echo` and it still works 😅

frontend/catalyst/compiler.py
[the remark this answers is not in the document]
dime10: Is it just based on hoping that the OS doesn't schedule the subprocess thread until we hit this statement, or is there some other reason it would hold? For instance, does the process only start running once we call `p.communicate` (not sure how it works)?

frontend/catalyst/jit.py
dime10: Good point. I was able to get it running now and the process is pretty straightforward, so I'm happy to go this route if you prefer :)

Although, I just thought of another alternative. If we want to contain this sort of functionality to the debug module, we could add a context manager for emitting the signal, similar to the instrumentation one. A bit more inline with what we already have, and keeps it out of the regular user options. Just an idea though.

frontend/catalyst/debug/debugger.py
```diff
@@ -0,0 +1,26 @@
...
+"""
+This module adds functionality to check if the active Python session
+is being run with an active debugger.
+"""
+
+
+import sys
+
+
+def is_debugger_active() -> bool:
```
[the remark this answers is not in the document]
dime10: Looking at this function again, I don't think we should add it to the Catalyst API, it's completely unrelated to Catalyst code.

doc/dev/debugging.rst
mlxd: Does it influence where the source directory is parsed for the debug information?
If not, all good to put in a no-op here, just to please the VSCode integrator.

doc/dev/debugging.rst
mlxd: Yea, as in the other, I think we need some form of pre-execution script to define a system-dependent argument for this using a [custom task](https://code.visualstudio.com/docs/debugtest/tasks#_custom-tasks)

frontend/catalyst/compiler.py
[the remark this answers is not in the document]
mlxd: So  `communicate` is something akin to a synchronisation point for the child-process --- it allows data to be sent to the process (assuming it is waiting for it) and receives data from the process.

For a quick 'n' dirty example of whether it is fast enough to ensure the process isn't running anything, we can opt for something akin to https://stackoverflow.com/questions/50002804/create-subprocess-in-python-in-suspended-state

In practice, use of the `preexec_fn` argument in `Popen` should work just fine here (as in, we create the PID, and before execution happens we immediately put it into the suspend state). This *should* be fine. Though, if we are ok with letting the child process set itself up first, chances are quite low anything will execute. Taking a simple compiled C binary that outputs a string `test`, we can try out the following to see when the output hits the screen (creating a file should work too):

```python
import os
import signal

def f():
    "pre-execute function to print the new process PID"
    print(os.getpid())

print("Starting process")
p = Popen(["/tmp/test"], preexec_fn=f)
print("Running process")

# Stop process after creating the Process and a print call.
# If the `/tmp/test` binary can output before this line is hit,
# we need to stop within the preexec_fn. Otherwise, all good
# to set up the process.
p.send_signal(signal.SIGSTOP)

# Have some wait-time that we control 
i = input("enter/return to continue")

# Allow direct input/output from communicate
p.communicate()

# It should be done by now
print("Finished process")

# explicitly kill the process
p.kill()
print("Killed process")
```

Reasoning for favouring not relying on the `preexec_fn` is there appears to be some (valid) ongoing attempts to remove this from CPython (see https://github.com/python/cpython/issues/82616).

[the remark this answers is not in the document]
dime10: Thanks for the info! I'll leave it up to you then if we want to use preexec or not, although in the current approach maybe we could at least send the signal before printing? 😅

doc/dev/debugging.rst
dime10: For a basic solution, I think it's ok to mention in the instructions to switch to lldb if on mac.

doc/dev/debugging.rst
dime10: > Does it influence where the source directory is parsed for the debug information?

Hmm, how can I check this? I used the echo path and it still stopped on the breakpoint in my local catalyst installation.

frontend/catalyst/compiler.py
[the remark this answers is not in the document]
mlxd: Fine by me. We can always revert if the above CPython changes make it to a release.

Update: As an aside, since the printing will happen from the subprocess, we may have issues with capturing the stdout and stderr for reporting. I'll see if this is something that can be mitigated, but if not, we just use the existing.

frontend/catalyst/jit.py
mlxd: Actually, not a bad idea. Let me see what is needed to try that approach. If it becomes a pain, we can always support that in a follow-up, and leave this as usable in the current form for now.

dime10:
Coming back to this, I would love to see this merged, as it's quite useful 🤩 

Another thing that would be really useful is being able to debug the runtime as well, since this is limited to the compiler. For instance if errors appear in the runtime, the debugger will only stop on the line that calls the ctypes-loaded library function.

PR #1719, opened 2025-05-05
Add the compiler option for limiting the Pauli size in PPM's passes

sengthai:
**Context:**
The idea is that the size of the Pauli product (PPR weight) increases after commuting or merging. This could lead to the simultaneous use of more qubits, which could negatively affect performance depending on the hardware.

The goal is to add compiler option `max-pauli-size` to the the `commute_ppr` and `merge_ppr_ppm` to limit the size of the Pauli product based on the user's input, or no limit (by default `max-pauli-size=0`).

**Description of the Change:** This commit introduces a new option, `max_pauli_size`, to limit the maximum size of Pauli products in the QEC transformation passes. The option is integrated into the `CommuteCliffordTPPRPass` and `CommuteCliffordPastPPMPass`, allowing for more controlled transformations based on the specified size limit.

**Benefits:** This enhancement allows users to manage the complexity of transformations by setting a maximum size for Pauli products, improving performance and usability in quantum circuit optimizations.

**Example:** 
MLIR pass usage:
```sh
quantum-opt --merge_ppr_ppm="max-pauli-size=3" ./test.mlir
```
Frontend usage:
```python
import pennylane as qml

from catalyst import measure, qjit
from catalyst.passes import commute_ppr, merge_ppr_ppm, to_ppr

pipe = [("pipe", ["enforce-runtime-invariants-pipeline"])]

@qjit(pipelines=pipe, target="mlir")
@to_ppr
@commute_ppr(max_pauli_size=4)
@merge_ppr_ppm(max_pauli_size=3)
@qml.qnode(qml.device("null.qubit", wires=3))
def circuit():
    qml.CNOT([0, 2])
    qml.CNOT([0, 1])
    qml.CNOT([0, 2])
    return measure(0), measure(1), measure(2)

print(circuit.mlir_opt)
```

[[sc-90438]]


doc/releases/changelog-dev.md
```diff
@@ -13,6 +13,10 @@
   ensures that measurement processes on `null.qubit` always return the value 0 or the result
   corresponding to the '0' state, depending on the context.
 
+* The `commute_ppr` and `merge_ppr_ppm` passes now accept an optional `max_pauli_size` argument,
+  which limits the size of the Pauli strings that can be commuted or merged.
+  [(#1719)](https://github.com/PennyLaneAI/catalyst/pull/1719)
```
isaacdevlugt, line 18: Nice!

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -408,6 +408,7 @@ def commute_ppr(qnode):
 
     Args:
         fn (QNode): QNode to apply the pass to.
+        max_pauli_size (uint): The maximum size of the Pauli strings after commuting.
```
isaacdevlugt, line 411: Might be nice to have an example here that uses `max_pauli_size` 👀

sengthai: Thanks @isaacdevlugt. Good idea! Will do!

sengthai: I added examples in `commute_ppr` and `merge_ppr_ppm` in here [fbb9761](https://github.com/PennyLaneAI/catalyst/pull/1719/commits/fbb976131e123c3c2d8e8b851b267453110c560f) and [053b3a4](https://github.com/PennyLaneAI/catalyst/pull/1719/commits/053b3a47e7fc154c877d79c4d0a72560a116f7ca)
🙌

doc/releases/changelog-dev.md
```diff
@@ -37,6 +37,18 @@
   Finally, the `PrintState` and the `One`/`Zero` utility functions have been removed, since they
   did not serve a convincing purpose.
 
+* (Frontend Developers Only) Some Catalyst primitives for JAX have been renamed, and the qubit
```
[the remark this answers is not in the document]
sengthai: We merged this by mistake, this will be undone.

frontend/catalyst/passes/builtin_passes.py
sengthai: New links [e6d3a31](https://github.com/PennyLaneAI/catalyst/pull/1719/commits/e6d3a315e9e81e5c76a99705b0078636cb902cb2) and [fa6aa84](https://github.com/PennyLaneAI/catalyst/pull/1719/commits/fa6aa84181d3af7f5166ce10077dfa9f8919746c)

doc/releases/changelog-dev.md
isaacdevlugt:
```suggestion
* The :func:`~.passes.commute_ppr` and :func:`~.passes.merge_ppr_ppm` passes now accept an optional `max_pauli_size` argument,
  which limits the size of the Pauli strings that can be commuted or merged.
  [(#1719)](https://github.com/PennyLaneAI/catalyst/pull/1719)
```

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -447,11 +448,52 @@ def circuit():
         %mres, %out_qubits = qec.ppm ["Z"] %5 : !quantum.bit
         . . .
 
+    If a commutation resulted in a PPR acting on more than
+    'max_pauli_size' (here, 2) qubits, that commutation would be skipped.
```
isaacdevlugt, line 452:
```suggestion
    `max_pauli_size` qubits (here, `max_pauli_size = 2`), that commutation would be skipped.
```

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -497,5 +540,46 @@ def circuit():
         %mres, %out_qubits = qec.ppm ["X"] %2 : !quantum.bit
         . . .
 
+    If a merging resulted in a PPM acting on more than
+    'max_pauli_size' (here, 2) qubits, that merging would be skipped.
```
isaacdevlugt, line 544:
```suggestion
    If a merging resulted in a PPM acting on more than
    `max_pauli_size` qubits (here, `max_pauli_size = 2`), that merging would be skipped.
```

isaacdevlugt (approved):
Nice!

doc/releases/changelog-dev.md
sengthai: Thanks @isaacdevlugt for suggestions. 🫡

mlir/lib/QEC/Transforms/CommuteCliffordPastPPM.cpp
```diff
@@ -173,18 +173,34 @@ bool shouldRemovePPR(PPRotationOp op)
...
+            // Handle non-commuting case
+            auto commutedResult = normPPROp.computeCommutationRulesWith(normPPMOp);
+
+            // Skip if Pauli size is too large
+            size_t pauliSize = commutedResult.get_pauli_word().size();
+            if (exceedPauliSizeLimit(pauliSize, MAX_PAULI_SIZE)) {
+                return failure();
             }
+
+            moveCliffordPastPPM(normPPROp, normPPMOp, &commutedResult, rewriter);
```
dime10, line 203: Nice 💯

mlir/lib/QEC/Transforms/CommuteCliffordPastPPM.cpp
```diff
@@ -173,18 +173,34 @@ bool shouldRemovePPR(PPRotationOp op)
...
             }
-            else {
-                auto resultStr = normOps.first.computeCommutationRulesWith(normOps.second);
-                moveCliffordPastPPM(normOps.first, normOps.second, &resultStr, rewriter);
+
+            // Handle non-commuting case
+            auto commutedResult = normPPROp.computeCommutationRulesWith(normPPMOp);
+
+            // Skip if Pauli size is too large
+            size_t pauliSize = commutedResult.get_pauli_word().size();
```
dime10, line 198: Does this include Identities or no?

dime10 (changes requested):
Nice work!!

Generally I have two questions regarding this feature:
1. There is a chance that we commute a bunch of gates only for them to not be merged into measurements. This means we may have made the program more expensive (by increasing the Pauli size of some products) without getting any benefit out of it.
    - The problem is I don't see a good way out of this immediately. One approach is to allow for rollback (undoing commutations that weren't beneficial), or to try and lookahead all the way to the end of the circuit to determine whether we will be able to absorb the rotation, but that could be really expensive.
    - Maybe we could restructure the pass to take a single pi/4 rotation and move it all the way back to the end in one change, and if that doesn't work do nothing.
    - Either way the current approach is a good start to allow us to enforce the size limit though.
2. Do we need to consider the final size of the PPMs with this parameter? For instance, with a limit of 10, and a product that due to commutation turns into a product of size 10, upon decomposition to a PPM will require a product of size 11.

frontend/catalyst/passes/builtin_passes.py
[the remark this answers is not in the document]
sengthai: Ohh yeah, there is no built-in uint in Python. 😅

doc/releases/changelog-dev.md
```diff
@@ -13,6 +13,10 @@
   ensures that measurement processes on `null.qubit` always return the value 0 or the result
   corresponding to the '0' state, depending on the context.
 
+* The :func:`~.passes.commute_ppr` and :func:`~.passes.merge_ppr_ppm` passes now accept an optional `max_pauli_size` argument,
+  which limits the size of the Pauli strings that can be commuted or merged.
+  [(#1719)](https://github.com/PennyLaneAI/catalyst/pull/1719)
```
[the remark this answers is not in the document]
dime10: Ah, we have a general line length limit 100 across the board. We set it in a few of our tools (black, isort, clang-format) but we don't have any automatic formatting for markdown or other formats. I would recommend enabling a ruler at column 100 in your editor, that way you always know if you are going over (going over is ok with URLs as there is often no way around it).

frontend/test/pytest/test_peephole_optimizations.py
```diff
@@ -223,6 +223,70 @@ def f():
 
 test_convert_clifford_to_ppr()
 
+#
+# commute_ppr and merge_ppr_ppm with max_pauli_size
+#
```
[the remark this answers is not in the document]
sengthai: This problem happens across my PRs. 😅

mlir/lib/QEC/Transforms/CommuteCliffordPastPPM.cpp
sengthai: Thank you~

sengthai:
> Nice work!!

Thank you, David! 

> Generally I have two questions regarding this feature:
>
> 1. There is a chance that we commute a bunch of gates only for them to not be merged into measurements. This means we may have made the program more expensive (by increasing the Pauli size of some products) without getting any benefit out of it.
>    * The problem is I don't see a good way out of this immediately. One approach is to allow for rollback (undoing commutations that weren't beneficial), or to try and lookahead all the way to the end of the circuit to determine whether we will be able to absorb the rotation, but that could be really expensive.
>    * Maybe we could restructure the pass to take a single pi/4 rotation and move it all the way back to the end in one change, and if that doesn't work do nothing.
>    * Either way the current approach is a good start to allow us to enforce the size limit though.


You are right, we may not get the benefit while hurting the performance.
It seems to me that the last method would work! Moving a single pi/4 rotation to the end in one change.
For now, since `commute_ppr` and `merge_ppr_ppm` are separated, it would be hard for us to roll back and look ahead.

By restructuring, the pass would be simpler to use since we have only one `max-pauli-size`.

> 2. Do we need to consider the final size of the PPMs with this parameter? For instance, with a limit of 10, and a product that due to commutation turns into a product of size 10, upon decomposition to a PPM will require a product of size 11.

Yes, that's a good question. I think we need to consider that because PPM could also be exploited after commuting as well. 🤔 However, back to the first question, if we set the final size of PPM and we can't commute past that, then the bunch of commuting would be of no benefit.

Screenshot 2025-05-12 at 6 39 54 PM

doc/releases/changelog-dev.md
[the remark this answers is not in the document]
sengthai: Ohh, I got it now. It's really helpful. Thank you, David!

dime10:
> > 1. There is a chance that we commute a bunch of gates only for them to not be merged into measurements. This means we may have made the program more expensive (by increasing the Pauli size of some products) without getting any benefit out of it.
> >    
> >    * The problem is I don't see a good way out of this immediately. One approach is to allow for rollback (undoing commutations that weren't beneficial), or to try and lookahead all the way to the end of the circuit to determine whether we will be able to absorb the rotation, but that could be really expensive.
> >    * Maybe we could restructure the pass to take a single pi/4 rotation and move it all the way back to the end in one change, and if that doesn't work do nothing.
> >    * Either way the current approach is a good start to allow us to enforce the size limit though.
> 
> You are right, we may not get the benefit while hurting the performance. It seems to me that the last method would work! Moving a single pi/4 rotation to the end in one change. For now, since `commute_ppr` and `merge_ppr_ppm` are separated, it would be hard for us to roll back and look ahead.
> 
> By restructuring, the pass would be simpler to use since we have only one `max-pauli-size`.


I think it's okay though, we can consider that as a subsequent improvement. For now, the PR enforces the limit so that's a good start :)

doc/releases/changelog-dev.md
```diff
@@ -13,8 +13,9 @@
   ensures that measurement processes on `null.qubit` always return the value 0 or the result
   corresponding to the '0' state, depending on the context.
 
-* The :func:`~.passes.commute_ppr` and :func:`~.passes.merge_ppr_ppm` passes now accept an optional `max_pauli_size` argument,
-  which limits the size of the Pauli strings that can be commuted or merged.
+* The :func:`~.passes.commute_ppr` and :func:`~.passes.merge_ppr_ppm` passes now accept an optional
+  `max_pauli_size` argument, which limits the size of the Pauli strings generated by the passes
+   through can commutation or absorption rules.
```
isaacdevlugt, line 18:
```suggestion
   through commutation or absorption rules.
```

mlir/lib/QEC/Transforms/CommuteCliffordPastPPM.cpp
```diff
@@ -208,9 +224,9 @@ struct RemoveDeadPPR : public OpRewritePattern<PPRotationOp> {
 namespace catalyst {
 namespace qec {
 
-void populateCommuteCliffordPastPPMPatterns(RewritePatternSet &patterns)
+void populateCommuteCliffordPastPPMPatterns(RewritePatternSet &patterns, uint max_pauli_size)
```
dime10, line 227: Is `uint` a thing in C? I thought it was `unsigned` or `unsigned int`

dime10 (approved):
Good 👌

mlir/lib/QEC/Transforms/CommuteCliffordPastPPM.cpp
sengthai: I think `uint` is a typedef of `unsigned int`

dime10: Do you know where? From what I can tell it's not part of the standard

sengthai: My IDE reference to this "/Library/Developer/CommandLineTools/SDKs/MacOSX15.4.sdk/usr/include/sys/types.h" 🤔

`typedef unsigned int            uint;           /* Sys V compatibility */`

dime10: Not sure how portable this is, I can't find that typedef online. Best to use standard types.

sengthai: Yes, it's better to use `unsigned int`. I've just modified here. https://github.com/PennyLaneAI/catalyst/pull/1719/commits/7e1cee2bf9a9c6acf041e18cc828eb4e06dc8075

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -467,6 +508,7 @@ def merge_ppr_ppm(qnode):
 
     Args:
         fn (QNode): QNode to apply the pass to
+        max_pauli_size (uint): The maximum size of the Pauli strings after merging.
```
dime10, line 511:
```suggestion
        max_pauli_size (int): The maximum size of the Pauli strings after merging.
```

mlir/lib/QEC/Transforms/CommuteCliffordPastPPM.cpp
```diff
@@ -173,18 +173,35 @@ bool shouldRemovePPR(PPRotationOp op)
...
-            else {
-                auto resultStr = normOps.first.computeCommutationRulesWith(normOps.second);
-                moveCliffordPastPPM(normOps.first, normOps.second, &resultStr, rewriter);
+
+            // Handle non-commuting case
+            auto commutedResult = normPPROp.computeCommutationRulesWith(normPPMOp);
+
+            // Skip if Pauli size is too large
+            size_t pauliSize = commutedResult.get_pauli_word().size();
+            if (exceedPauliSizeLimit(pauliSize, MAX_PAULI_SIZE)) {
```
dime10, line 200: Looking at the type we compare `MAX_PAULI_SIZE` against, `size_t` might make even more sense than `unsigned` for more uniform types, but this is very minor.

mlir/lib/QEC/Utils/PauliStringWrapper.cpp
```diff
@@ -190,5 +190,16 @@ void updatePauliWordSign(QECOpInterface op, bool isNegated, PatternRewriter &rew
     op.setRotationKind(rotationKind);
 }
 
+bool isNoSizeLimit(uint max_pauli_size) { return max_pauli_size == 0; }
```
dime10, line 193: This name doesn't match what is declared in the header 😅

dime10 (approved):
Looks good 👌

mlir/lib/QEC/Utils/PauliStringWrapper.cpp
sengthai: Ahah, I also prefer consistency. How about them all `size_t`? XD

mlir/lib/QEC/Transforms/CommuteCliffordPastPPM.cpp
sengthai: Ahah, I also prefer consistency. How about them all size_t? XD

PR #1721, opened 2025-05-06
Update clang-format and clang-tidy dependancy

mehrdad2m:
**Context:**
As part of updating dependancy versions for the ecosystem, we want to update formatting tools like `clang-tidy` and `clan-format`.

Bump `clang-tidy` and `clang-format` to v20.

Note that I changed the Formatting CI job to install clang-format from the requirements this makes the job consistent with local development environment and allow us to use the latest version that are not yet available on the ubuntu repositories. 

**Benefits:**
Using the latest formatting tools

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-91016]


mlxd:
Just an FYI, we may want to unify this with Lightning and the ruleset in the guidance repo.

requirements.txt
```diff
@@ -12,8 +12,8 @@ PyYAML
 
 # formatting/linting
 black
-clang-format==14.*
-clang-tidy~=16.0
+clang-format
```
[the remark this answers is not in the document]
mehrdad2m: Oh yes definitely. I will pin it to the latest version. this was just a test lol

mehrdad2m:
> Just an FYI, we may want to unify this with Lightning and the ruleset in the guidance repo.

Made a PR for lightning as well.
https://github.com/PennyLaneAI/pennylane-lightning/pull/1153

.github/workflows/check-formatting.yaml
```diff
@@ -34,13 +34,11 @@ jobs:
...
-      - name: Install dependencies
-        run: |
-          sudo apt-get update
-          sudo apt-get -y install clang-format-14
-
       - name: Checkout code
         uses: actions/checkout@v4
 
+      - name: Install dependencies
+        run: python3 -m pip install -r requirements.txt
```
paul0403, line 41: I know it's unlikely, but installing clang-format through pip scares me a little bit because pip and system goes to different paths. Do we want to enable wheels to make sure this is good on all platforms?

paul0403: Actually ignore this, format is only ever run in main CI on linux arm. All good then!

.github/workflows/check-formatting.yaml
```diff
@@ -34,13 +34,11 @@ jobs:
...
+
       - name: Run clang-format
-        run: python3 ./bin/format.py --check --verbose --cfversion 14 .
```
paul0403, line 46: Why not `--cfversion 20` ?

mlir/lib/QEC/Transforms/CommuteCliffordPastPPM.cpp
```diff
@@ -17,7 +17,7 @@
 #include "mlir/Analysis/SliceAnalysis.h"
 #include "llvm/Support/Casting.h"
 #include "llvm/Support/Debug.h"
-//#include "mlir/Analysis/TopologicalSortUtils.h"  // enable when updating llvm
+// #include "mlir/Analysis/TopologicalSortUtils.h"  // enable when updating llvm
```
paul0403, line 20: :laughing:

runtime/include/QuantumDevice.hpp
```diff
@@ -187,7 +187,7 @@ struct QuantumDevice {
      *
      * @param gen Pointer to a Catalyst-managed Mersenne Twister instance.
      */
-    virtual void SetDevicePRNG(std::mt19937 *gen){};
+    virtual void SetDevicePRNG(std::mt19937 *gen) {};
```
paul0403, line 190: A little scared here about updating the `QuantumDevice` base class. In principle everytime we update this base class, external devices need to rebuild against the new base class. Although I think in this case it should be fine? @dime10 what do you think? :thinking:

.github/workflows/check-formatting.yaml
mehrdad2m: `cfversion` appends the name of the binary from `clang-format` to `clang-format-20`. I could not install `clang-format-20` because it is not available in the ubuntu when doing `apt install clang-format-20` (the last one available is v19 on `ubuntu-24.04`). By removing the cfversion option I am telling the script to simply run `clang-format` which is just installed via `pip install`

mlir/lib/QEC/Transforms/CommuteCliffordPastPPM.cpp
mehrdad2m: I guess this line will be removed in the following week after updating LLVM version :)

runtime/include/QuantumDevice.hpp
dime10: Adding whitespace is fine :)

runtime/Makefile
```diff
@@ -159,6 +159,7 @@ check-tidy:
 		-DCMAKE_BUILD_TYPE=$(BUILD_TYPE) \
 		-DCMAKE_C_COMPILER=$(C_COMPILER) \
 		-DCMAKE_CXX_COMPILER=$(CXX_COMPILER) \
+		-DMLIR_INCLUDE_DIRS="$(abspath $(LLVM_DIR))/mlir/include" \
```
dime10, line 162: How come we didn't need this before?

dime10 (approved):
Looks good!

.github/workflows/check-formatting.yaml
mehrdad2m: Yeah I had that concert as well, but after looking at the runtime of the action, it adds between 20-30 seconds which  is not that bad. The alternative here is to be limited to the latest version available on ubuntu repositories.

paul0403: > The alternative here is to be limited to the latest version available on ubuntu repositories.

Why? I think David was just saying we only need to `pip install clang-format` in the check formatting action, and we don't need all the other dependencies.

dime10: > 20-30 seconds

Half a minute, that can start to add up 😆 When I first wrote the main action I tried my hardest to keep it under 5 minutes, but those times are long gone I think 😞

runtime/Makefile
dime10: Oh I see, yeah good point 😅

PR #1728, opened 2025-05-07
Upgrading Github Action runners to ubuntu24.04

runora95:
### Before submitting

      The latest version of black and `clang-format-14` are used in CI/CD to check formatting.

**Context:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/.github/workflows/check-catalyst.yaml
+++ b/.github/workflows/check-catalyst.yaml
@@ -21,7 +21,7 @@ jobs:
     name: Determine runner type to use
     uses: ./.github/workflows/determine-workflow-runner.yml
     with:
-      default_runner: ubuntu-22.04
+      default_runner: ubuntu-24.04
 
   constants:
     name: "Set build matrix"
--- a/.github/workflows/build-nightly-release.yaml
+++ b/.github/workflows/build-nightly-release.yaml
@@ -9,7 +9,7 @@ on:
 jobs:
   setup:
     name: Setup the release
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     steps:
     - name: Checkout Catalyst repo
       uses: actions/checkout@v4
@@ -53,7 +53,7 @@ jobs:
   upload:
     name: Prepare & Upload wheels to TestPyPI
     needs: [linux-x86, macos-arm, linux-aarch]  #, macos-x86
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     permissions:
       id-token: write  # IMPORTANT: mandatory for trusted publishing
 
--- a/.github/workflows/build-wheel-linux-arm64.yaml
+++ b/.github/workflows/build-wheel-linux-arm64.yaml
@@ -61,7 +61,7 @@ jobs:
         container_img: ["manylinux_2_28_aarch64"]
 
     name: Build Dependencies (Python ${{ matrix.python_version }})
-    runs-on: ubuntu-22.04-arm
+    runs-on: ubuntu-24.04-arm
     container: quay.io/pypa/${{ matrix.container_img }}
 
     if: needs.check_if_wheel_build_required.outputs.build-wheels == 'true'
@@ -241,7 +241,7 @@ jobs:
         container_img: ["manylinux_2_28_aarch64"]
 
     name: Build Wheels (Python ${{ matrix.python_version }})
-    runs-on: ubuntu-22.04-arm
+    runs-on: ubuntu-24.04-arm
     container: quay.io/pypa/${{ matrix.container_img }}
 
     steps:
@@ -405,7 +405,7 @@ jobs:
 
     # To check all wheels for supported python3 versions
     name: Test Wheels (Python ${{ matrix.python_version }}) on Linux Arm
-    runs-on: ubuntu-22.04-arm
+    runs-on: ubuntu-24.04-arm
 
     steps:
     - name: Checkout Catalyst repo
--- a/.github/workflows/build-wheel-linux-x86_64.yaml
+++ b/.github/workflows/build-wheel-linux-x86_64.yaml
@@ -47,7 +47,7 @@ jobs:
     name: Determine runner type to use
     uses: ./.github/workflows/determine-workflow-runner.yml
     with:
-      default_runner: ubuntu-22.04
+      default_runner: ubuntu-24.04
 
   check_if_wheel_build_required:
     if: github.event.pull_request.draft == false
--- a/.github/workflows/changelog-reminder.yaml
+++ b/.github/workflows/changelog-reminder.yaml
@@ -9,7 +9,7 @@ name: Changelog Reminder
 jobs:
   remind:
     name: Changelog Reminder
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     steps:
     - uses: actions/checkout@master
     - name: Changelog Reminder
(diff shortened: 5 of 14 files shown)


.github/workflows/check-catalyst.yaml
```diff
@@ -42,9 +42,15 @@ jobs:
...
         uses: actions/checkout@v4
+      
+      - name: Set up Python # Ensure python3.10 is used
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ needs.constants.outputs.primary_python_version }}
 
       - name: Install dependencies
         run: |
+          python3 --version
```
dime10: Is this just for informational purposes?
I think the setup-python step above already reports that information.

.github/workflows/check-catalyst.yaml
```diff
@@ -159,6 +172,8 @@ jobs:
         CXX_COMPILER=$(which ${{ needs.constants.outputs[format('cxx_compiler.{0}', matrix.compiler)] }}) \
         LLVM_BUILD_DIR="$(pwd)/llvm-build" \
         COMPILER_LAUNCHER="" \
+        PYTHON3_EXECUTABLE=$(which python3) \
+        Python3_NumPy_INCLUDE_DIRS=$$($(PYTHON) -c "import numpy as np; print(np.get_include())") \
```
dime10: I'm not sure these flags do anything, because the variables are going to the make command, not cmake. And make is already relaying this information to CMake:

```
PYTHON ?= $(shell which python3)
...
		-DPython3_EXECUTABLE=$(PYTHON) \
		-DPython3_NumPy_INCLUDE_DIRS=$$($(PYTHON) -c "import numpy as np; print(np.get_include())") \
```

.github/workflows/check-catalyst.yaml
```diff
@@ -649,15 +702,26 @@ jobs:
...
+          python-version: ${{ needs.constants.outputs.primary_python_version }}
+
       - name: Install dependencies
         run: |
           sudo apt-get update
-          sudo apt-get -y -q install cmake ninja-build lcov
+          sudo apt-get -y -q install cmake ninja-build
           python3 -m pip install nanobind pybind11
 
+      - name: Install lcov # Manually install lcov 1.14 (lcov_1.15-1_all.deb)
```
dime10: The comment says 1.14 but we are installing 1.15 no?

.github/workflows/check-catalyst.yaml
```diff
@@ -484,6 +521,7 @@ jobs:
 
     - name: Run Python Lit Tests
       run: |
+        export PYTHONPATH=$PYTHONPATH:$(python3 -c "import site; print(site.getsitepackages()[0])")
```
dime10: This is an odd command, I'm guessing llvm-lit was having trouble finding the Python `lit` package? I think the reason might be that `llvm-lit` is a script that encodes the python executable to use at the time it is generated, maybe something isn't lining up quite right there.

On my system the when I open the `llvm-lit` for example I see:
```py
#!/usr/local/bin/python3
# -*- coding: utf-8 -*-

import os
import sys
```

.github/workflows/check-catalyst.yaml
runora95: yeah the `lcov_1.15-1_all.deb` package installs version 1.14

dime10: 😵‍💫

dime10: Surely that is a bug?

dime10: image

.github/workflows/check-catalyst.yaml
runora95: let me try it without it

.github/workflows/check-catalyst.yaml
runora95: It's probably an issue on their end 🤷‍♀️: `lcov: LCOV version 1.14` 
 https://github.com/PennyLaneAI/catalyst/actions/runs/15399436545/job/43328486730#step:5:89

dime10: Any chance lcov is already installed? What if you run `lcov --version` before the installation / don't install?

Otherwise that seems pretty wild to distribute a package under a false version 😅

.github/workflows/check-catalyst.yaml
runora95: llvm-lit does try to [find the lit package ](https://github.com/llvm/llvm-project/blob/main/llvm/utils/llvm-lit/llvm-lit.in#L26) on it's own, it just wasn't able to find it

dime10: Yeah that's the one it ships within its own source code, but it makes sense for that not to be available since the `frontend-tests` job has no access to the llvm source code (we only restore llvm build artifacts).

It's not a problem though because we install [`lit`](https://github.com/PennyLaneAI/catalyst/blob/4228bf11b578470a7dca5fcd8980633cb7942006/requirements.txt#L22) from [pypi](https://github.com/PennyLaneAI/catalyst/blob/4228bf11b578470a7dca5fcd8980633cb7942006/.github/workflows/check-catalyst.yaml#L440) (the basis for your fix with the Python path). The question is, why doesn't it find the installed package?

The only explanation I can come up with is that two Python environments are being mixed here, since exporting the site-packages location from one interpreter into that same interpreter wouldn't really make any sense.
That's why I was wondering if the Python executable hardcoded into the `llvm-lit` script is different than what we expect, in which case it wouldn't have access to any packages we just installed.

.github/workflows/check-catalyst.yaml
```diff
@@ -432,11 +464,16 @@ jobs:
...
 
+    - name: Set up Python # Ensure python3.10 is used
+      uses: actions/setup-python@v5
+      with:
+        python-version: ${{ needs.constants.outputs.primary_python_version }}
+
     - name: Install Deps
       run: |
         sudo apt-get update
         sudo apt-get install -y python3 python3-pip libomp-dev libasan6 make ninja-build
```
dime10: Do we need these with the setup-python action above?
```suggestion
        sudo apt-get install -y libomp-dev libasan6 make ninja-build
```

.github/workflows/check-catalyst.yaml
```diff
@@ -484,6 +521,7 @@ jobs:
 
     - name: Run Python Lit Tests
       run: |
+        export PYTHONPATH=$PYTHONPATH:$(python3 -c "import site; print(site.getsitepackages()[0])")
         llvm-lit -sv frontend/test/lit -j$(nproc)
```
dime10: If the above is true, I think a slightly more robust approach might just be to force the right python interpreter directly:
```suggestion
        python3 llvm-build/bin/llvm-lit -sv frontend/test/lit -j$(nproc)
```

.github/workflows/check-catalyst.yaml
runora95: lcov isn't already installed, I'll just change the wording of the comment to be less confusing.

dime10 (approved):
🎉

.github/workflows/check-catalyst.yaml
dime10: Yeah must be a bug, the package manager thinks it's 1.15 but the command itself thinks it's 1.14 😆

PR #1734, opened 2025-05-09
Bump docs related packages versions

mehrdad2m:
**Context:**
As part of the work to update PennyLane ecosystem dependencies, we need to update packages related to documentation in Catalyst

Updated docs dependancies listed in `doc/requirements.txt`. We tried to be unified with the respective versions in PennyLane as much as possible. 

**Benefits:**
updated versions in docs

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-91015]


mehrdad2m:
Currently there are some hiccups in rendering these two pages ([1](https://xanaduai-pennylane--1734.com.readthedocs.build/projects/catalyst/en/1734/api/file_runtime_include_RuntimeCAPI.h.html), [2](https://xanaduai-pennylane--1734.com.readthedocs.build/projects/catalyst/en/1734/api/file_runtime_include_QuantumDevice.hpp.html)). However, I am not sure if we even need to include those two pages as they don't look very informative to users. Regardless it would useful to figure out the issue. (probably related to `exhale` or `breathe` packages)

josh146:
The issue seems to be that the in-page contents is generated in HTML using a `<nav>` element:

image

Manually changing this to `<div>` in my web browser fixes this: 

image
image

There might be a way to fix this in the sphinx docs somewhere 🤔 But we can probably do this in a separate PR.

dime10:
> Currently there are some hiccups in rendering these two pages ([1](https://xanaduai-pennylane--1734.com.readthedocs.build/projects/catalyst/en/1734/api/file_runtime_include_RuntimeCAPI.h.html), [2](https://xanaduai-pennylane--1734.com.readthedocs.build/projects/catalyst/en/1734/api/file_runtime_include_QuantumDevice.hpp.html)). However, I am not sure if we even need to include those two pages as they don't look very informative to users. Regardless it would useful to figure out the issue. (probably related to `exhale` or `breathe` packages)

The QuantumDevice sub-page is quite important to render, so as long as we render that one it would be fine: https://xanaduai-pennylane--1734.com.readthedocs.build/projects/catalyst/en/1734/api/structCatalyst_1_1Runtime_1_1QuantumDevice.html#exhale-struct-structcatalyst-1-1runtime-1-1quantumdevice
The "higher-level" page that you're linking to is indeed not too important, although I think that's just how doxygen organizes stuff.

The CAPI page is indeed not very informative because it is undocumented. We included it originally because technically it's another interface that is available to someone developing their own runtime, but we don't really advertise or recommend this anywhere. I would be okay removing it.

mehrdad2m:
> > Currently there are some hiccups in rendering these two pages ([1](https://xanaduai-pennylane--1734.com.readthedocs.build/projects/catalyst/en/1734/api/file_runtime_include_RuntimeCAPI.h.html), [2](https://xanaduai-pennylane--1734.com.readthedocs.build/projects/catalyst/en/1734/api/file_runtime_include_QuantumDevice.hpp.html)). However, I am not sure if we even need to include those two pages as they don't look very informative to users. Regardless it would useful to figure out the issue. (probably related to `exhale` or `breathe` packages)
> 
> The QuantumDevice sub-page is quite important to render, so as long as we render that one it would be fine: https://xanaduai-pennylane--1734.com.readthedocs.build/projects/catalyst/en/1734/api/structCatalyst_1_1Runtime_1_1QuantumDevice.html#exhale-struct-structcatalyst-1-1runtime-1-1quantumdevice The "higher-level" page that you're linking to is indeed not too important, although I think that's just how doxygen organizes stuff.
> 
> The CAPI page is indeed not very informative because it is undocumented. We included it originally because technically it's another interface that is available to someone developing their own runtime, but we don't really advertise or recommend this anywhere. I would be okay removing it.

Thanks @dime10 for the context. That page seems to be rendering fine, but it is not very accessible to users in this format. I propose to fix the docs in another PR to only include the QuantumDevice page under API.

dime10 (approved):
Documentation still looks okay 👌

PR #1741, opened 2025-05-13
Support for ```qml.Snapshot```

### LINKED ISSUE #1132: Add support for `qml.Snapshot`
PennyLane added the [`qml.Snapshot`](https://docs.pennylane.ai/en/stable/code/api/pennylane.Snapshot.html) operation in [Release 0.22.0](https://docs.pennylane.ai/en/stable/development/release_notes.html#release-0-22-0), which saves the internal state of devices at arbitrary points of execution. For example:

```python
import pennylane as qml

NUM_QUBITS = 2
dev = qml.device("default.qubit", wires=NUM_QUBITS)

@qml.qnode(dev)
def circuit():
    wires = list(range(NUM_QUBITS))

    qml.Snapshot("Initial state")

    for wire in wires:
        qml.Hadamard(wires=wire)

    qml.Snapshot("After applying Hadamard gates")

    return qml.probs()


results = qml.snapshots(circuit)()

for k, result in results.items():
    print(f"{k:<30}: {result}")
```

Output:

```pycon
Initial state                 : [1.+0.j 0.+0.j 0.+0.j 0.+0.j]
After applying Hadamard gates : [0.5+0.j 0.5+0.j 0.5+0.j 0.5+0.j]
execution_results             : [0.25 0.25 0.25 0.25]
```

Adding `Snapshot` support to Catalyst would possibly be a nice feature to have for some users.

Note that currently if we try wrapping `circuit()` with a `@qjit` decorator, calling `qml.snapshots(circuit)()` will raise a `TransformError` exception since we cannot apply a PennyLane transform to a `qjit`-compiled function. This also requires changing the device to `"lightning.qubit"` in the example above.

ritu-thombre99:
**Context:** 

**Description of the Change:** add ```qml.Snapshot``` as an operation to Catalyst

**Benefits:** Quantum states at different points in a quantum circuit can be viewed using snapshots

**Possible Drawbacks:** None

**Related GitHub Issues:** Fixes #1132

**TODO**:
1. Add plxpr support and tests: not implemented due to [PennyLane bug](https://github.com/PennyLaneAI/pennylane/issues/7497)
2. Add pytree tests
3. Look if Catalyst's ```qml.Snapshot``` can support ```qml.measurement``` passed as an argument


frontend/catalyst/device/qjit_device.py
```diff
@@ -265,6 +265,9 @@ def get_qjit_device_capabilities(target_capabilities: DeviceCapabilities) -> Dev
             }
         )
 
+    # Optionally enable runtime-powered snapshot of quantum state at any particular instance
+    qjit_capabilities.operations.update({"Snapshot": OperatorProperties()})
```
paul0403, line 269: I know the comment above says optional, but (a) do we want this to be optional? and (b) seems like the addition of snapshot to the capabilities is is unconditional? 

Also, I think you need to set the three properties of the `OperatorProperties`.

frontend/catalyst/jax_tracer.py
```diff
@@ -759,6 +761,22 @@ def bind_native_operation(qrp, op, controlled_wires, controlled_values, adjoint=
             trace_state_prep(op, qrp)
         elif isinstance(op, qml.BasisState):
             trace_basis_state(op, qrp)
+        elif isinstance(op, qml.Snapshot):
+            print("Snapshot")
```
paul0403, line 765: Don't forget to clean up the debug prints! :laughing:

frontend/catalyst/jax_tracer.py
```diff
@@ -1371,7 +1389,10 @@ def is_leaf(obj):
                     postselect_mode=qnode.execute_kwargs["postselect_mode"],
                     mcm_method=qnode.execute_kwargs["mcm_method"],
                 )
-                qrp_out = trace_quantum_operations(tape, device, qreg_in, ctx, trace, mcm_config)
```
paul0403, line 1374: Don't forget to add tests! : )

frontend/catalyst/jax_tracer.py
```diff
@@ -759,6 +761,22 @@ def bind_native_operation(qrp, op, controlled_wires, controlled_values, adjoint=
...
+            )
+            if isinstance(nqubits, DynamicJaxprTracer):
+                shape = (jnp.left_shift(1, nqubits),)
+            else:
+                shape = (2**nqubits,)
+            qreg_out = qrp.actualize()
+            obs_tracers = compbasis_p.bind(qreg_out, qreg_available=True)
+            dyn_dims, static_shape = _extract_tracers_dyn_shape(shape)
+            result = state_p.bind(obs_tracers, *dyn_dims, static_shape=tuple(static_shape))
+            out_snapshot_tracer.append(result)
```
paul0403, line 779: So I know this snippet for adding the `state_p` primitive comes from the snippet in `trace_quantum_measurements`. Is it possible to outline it into a helper function? Since you mentioned you will be adding support for all the other measurements, this could potentially be quite a long duplicate.

frontend/catalyst/jax_tracer.py
```diff
@@ -1383,7 +1404,12 @@ def check_full_raise(arr, func):
                         return func(arr)
 
                 meas_tracers = check_full_raise(meas, trace.to_jaxpr_tracer)
-                meas_results = tree_unflatten(meas_trees, meas_tracers)
+                if len(snapshot_results) > 0:
+                    for s in snapshot_results[::-1]:
+                        meas_tracers.insert(0, s)
+                    meas_results = tree_unflatten(tree_structure(meas_tracers), meas_tracers)
```
paul0403, line 1410: I don't know the supposed output format of the snapshotted results. Are they not supposed to be distinguished from the regular results? 

I guess a pytest would clear up my confusion!

frontend/catalyst/jax_tracer.py
```diff
@@ -1383,7 +1404,12 @@ def check_full_raise(arr, func):
                         return func(arr)
 
                 meas_tracers = check_full_raise(meas, trace.to_jaxpr_tracer)
-                meas_results = tree_unflatten(meas_trees, meas_tracers)
```
paul0403, line 1386: Just a note, catalyst has two tracing mechanism. The one you updated here (`jax_tracer.py`) is currently the main tracing pipeline, but we also have another pipeline called `from_plxpr`. See https://github.com/PennyLaneAI/catalyst/blob/main/frontend/catalyst/from_plxpr.py

from_plxpr is planned to become the default tracing pipeline, so the same updates need to be made there as well.

paul0403 (commented):
Nice work! :tada:

frontend/catalyst/device/qjit_device.py
ritu-thombre99: I'm still unsure what to put in ```OperatorProperties``` of Snapshot

frontend/catalyst/jax_tracer.py
ritu-thombre99: Figured out a way to use ```trace_quantum_measurements``` for this to avoid duplication and handle any arbitrary measurement passed inside ```qml.Snapshot```

frontend/catalyst/jax_tracer.py
ritu-thombre99: Same 😄

frontend/catalyst/jax_tracer.py
ritu-thombre99: Thank you😊. I was unaware of this.

frontend/catalyst/device/qjit_device.py
paul0403: I think the properties can just all be `False`, as snapshot is not an actual operation in the mlir and runtime layers, and will just become measurements.

frontend/test/pytest/test_pytree_args.py
```diff
@@ -276,6 +276,42 @@ def circuit2(params):
...
+                dtype=jnp.complex128,
+            ),
+        ]
+        jitted_fn = qjit(circuit2_snapshot)
+        result = jitted_fn(params)
+        print("Result:", result)
+        assert isinstance(result, tuple)
+        assert isinstance(result[0], list)
+        assert isinstance(result[1], dict)
+        assert isinstance(result[1]["counts"], tuple)
```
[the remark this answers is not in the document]
ritu-thombre99: Yes that's the result format. The string labels wouldn't work since they cannot be traced by JAX.

paul0403:
We discovered that plxpr itself does not work with snapshots in core pennylane yet, so it is ok to not add it for now.

@ritu-thombre99 remember to open a bug report in the pennylane oss qa slack channel!

mehrdad2m (changes requested):
Thanks @ritu-thombre99, Great Work 🚀  Just a few minor comments:

doc/releases/changelog-dev.md
```diff
@@ -23,6 +23,38 @@
   * `decompose_non_clifford_ppr`: Decompose non-Cliford PPR (:math:`\theta = \tfrac{\pi}{8}`)
   into PPMs using a magic state.
 
+* PennyLane supports the [`qml.Snapshot`](https://docs.pennylane.ai/en/stable/code/api/pennylane.Snapshot.html) operation
```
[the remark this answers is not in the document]
mehrdad2m: nvm this. Misread the sentence.

ritu-thombre99:
@mehrdad2m Could you take a look at unresolved changes again? I made a few changes.

doc/releases/changelog-dev.md
```diff
@@ -23,6 +23,44 @@
   * `decompose_non_clifford_ppr`: Decompose non-Cliford PPR (:math:`\theta = \tfrac{\pi}{8}`)
   into PPMs using a magic state.
 
+* PennyLane supports the 
+  [`qml.Snapshot`](https://docs.pennylane.ai/en/stable/code/api/pennylane.Snapshot.html)
+  operation which saves the internal state of devices at arbitrary points of execution. Support for 
+  `qml.Snapshot` to capture quantum states at any point in a circuit has been added to Catalyst.
```
joeycarter, line 29:
```suggestion
*Support for :class:`qml.Snapshot <pennylane.Snapshot>` to capture quantum states at any point in a circuit has been added to Catalyst. 
```

I don't think you need the first sentence, better to start with the changes to Catalyst and add context as needed about anything in PennyLane.

doc/releases/changelog-dev.md
```diff
@@ -23,6 +23,44 @@
...
+
+  results = circuit()
+  print(results)
+  ```
+
+  The output would be a tuple of two elements: 
+    * Array of snapshot states
+    * Tuple of measurements being returned
+
+  ```shell
```
joeycarter, line 57: I would remove the `shell` and just have a plain text block here. I believe `shell` is generally reserved for shell commands, which this is not, so it adds syntax highlighting in funny places.

Another option is to use a `pycon` block, like so:

```pycon
>>> print(results)
([Array([1.+0.j, 0.+0.j, 0.+0.j, 0.+0.j], dtype=complex128), 
Array([0.5+0.j, 0.5+0.j, 0.5+0.j, 0.5+0.j], dtype=complex128)], 
```

joeycarter (commented):
Thanks @ritu-thombre99! Just had a few flyby observations in the docs.

PR #1580, opened 2025-03-21
Implement absorption of pi/4 PPRs into PPMs #2

dime10:
Rebased version of #1577 by @sengthai 

**Context:**
After commuting pi/4 PPR (Pauli Product Rotation) past pi/8 PPM (Pauli Product Measurement) in #1563, this sub-pass, `ppr_to_ppm,` aims to absorb those pi/4 PPRs into PPM. 

**Implementation Details:**
* Clifford (pi/4) PPRs can be absorbed by the final measurements and following two rules:
  * If they are commuting:
     * `P M -> M` 
   * If they are anti-commute:
     * `P P' -> iMP'`
    (`P` Clifford, `M'` PPM, `i` Imaginary number)
* Support gates:
  * `qec.ppr` and `qec.ppm`: Pauli product rotation and Pauli product measurement
  * `qec.ppr` is converted from any Clifford-T gates. 
  * `qec.ppm` is converted from `catalyst.measure` instruction.

**Example:**
```python
  ppm_passes = {
      "to_ppr": {},
      "commute_ppr": {},
      "ppr_to_ppm": {},
  }
  
  @qjit(keep_intermediate=True)
  @pipeline(ppm_passes)
  @qml.qnode(qml.device("null.qubit", wires=2))
  def circuit():
      qml.H(0)
      qml.CNOT([0, 1])
      qml.T(1)
      m1 = catalyst.measure(0)
      m2 = catalyst.measure(1)
      return m1, m2
```
The circuit generated from this pass is not executable on backends; they serve as intermediate representations for analysis and potential execution.

**Output:**
```mlir
. . .
    %3:2 = qec.ppr ["X", "Z"](8) %1, %2 : !quantum.bit, !quantum.bit
    %m1, %o1:2 = qec.ppm ["X", "Z"] %3#0, %3#1 : !quantum.bit, !quantum.bit
    %m2, %o2 = qec.ppm ["X"] %o1#0 : !quantum.bit
. . .
```
[sc-82897]

PR #1486, opened 2025-01-24
Add Pauli Product dialect (QEC)

dime10:
@sengthai We can use this as the target of our Pauli product passes. (feel free to adjust if the need arises)

[sc-82892]

PR #1750, opened 2025-05-16
`ppm_compilation` : Apply all sub-PPM compilation passes in one step. 

sengthai:
**Context:**
Introduced the `ppm_compilation` pass as a top-level composition of existing transformation and optimization passes. This allows users to apply the transformation Clifford+T to Pauli product measurements (PPMs) in a single step without managing sub-pass sequencing manually.
 
The sub-paseses comprise:
  * `to_ppr`: Converts gates into Pauli Product Rotations (PPRs) (Transformation)
       #1486  
       #1499 
  * `commute_ppr`: Commutes PPRs past non-Clifford PPRs (Optimization)
      #1563
  * `merge_ppr_ppm`: Merges PPRs into PPMs (Optimization) 
      #1580 
  * `ppr_to_ppm`: Decomposes PPRs into PPMs (Transformation) 
      #1664 

**Example:**
```python
  import pennylane as qml
  from catalyst import measure, passes, qjit
  
  pipe = [("pipe", ["enforce-runtime-invariants-pipeline"])]
  
  @qjit(pipelines=pipe, target="mlir")
  @passes.ppm_compilation(decompose_method="auto-corrected", avoid_y_measure=True, max_pauli_size=2)
  def circuit():
      qml.CNOT([0, 1])
      qml.CNOT([1, 0])
      qml.adjoint(qml.T)(0)
      qml.T(1)
      return measure(0), measure(1)
  
  print(circuit.mlir_opt)
```

**Benefits:**
- Simplifies the user workflow for compilation.

[[sc-90149]]


frontend/catalyst/passes/builtin_passes.py
```diff
@@ -666,3 +666,110 @@ def circuit():
...
+
+
+def to_ppm(
+    qnode=None, *, decompose_method="auto-corrected", avoid_y_measure=False, max_pauli_size=0
+):
+    R"""
+    Specify that the MLIR compiler pass for transforming
+    clifford+T gates into Pauli Product Measurements (PPM) will be applied.
+
+    This pass combines multiple sub-passes:
```
josh146, line 678:
```suggestion
    This pass combines multiple sub-passes:

```

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -666,3 +666,110 @@ def circuit():
...
+            Defaults to 0 (no limit).
+
+    Returns:
+        ~.QNode or callable: Returns decorated QNode if qnode is provided,
+            otherwise returns a decorator.
+
+    **Example**
+
+    If a merging resulted in a PPM acting on more than
+    `max_pauli_size` qubits (here, `max_pauli_size = 2`), that merging would be skipped.
```
josh146, line 706:
```suggestion
    ``max_pauli_size`` qubits (here, ``max_pauli_size = 2``), that merging would be skipped.
```

doc/releases/changelog-dev.md
```diff
@@ -61,12 +74,12 @@
   deallocation primitive has been split into deallocation and a separate device release primitive.
   [(#1720)](https://github.com/PennyLaneAI/catalyst/pull/1720)
 
-  - `qunitary_p` is now `unitary_p` (unchanged)
-  - `qmeasure_p` is now `measure_p` (unchanged)
-  - `qdevice_p` is now `device_init_p` (unchanged)
-  - `qdealloc_p` no longer releases the device, thus it can be used at any point of a quantum
+  * `qunitary_p` is now `unitary_p` (unchanged)
```
paul0403, line 77: This again? :sweat_smile: Is this some vscode black magic?

sengthai: Oops! 😅 auto formatted...
Let's me reverse.

frontend/test/pytest/test_peephole_optimizations.py
```diff
@@ -386,5 +387,41 @@ def g():
     assert 'qec.ppm ["Y"](-1)' in optimized_ir
 
 
+def test_clifford_to_ppm():
+
+    pipe = [("pipe", ["enforce-runtime-invariants-pipeline"])]
```
paul0403, line 392: Hmm, why do we need this pipeline? Is the new `to_ppm` pass not standalone? e.g. [like cancel_inverses passes here](https://github.com/PennyLaneAI/catalyst/blob/main/frontend/test/pytest/test_peephole_optimizations.py#L76)

frontend/test/pytest/test_peephole_optimizations.py
```diff
@@ -386,5 +387,41 @@ def g():
...
+        def g():
+            for idx in range(5):
+                qml.H(idx)
+                qml.CNOT(wires=[idx, idx + 1])
+                qml.T(idx)
+                qml.T(idx + 1)
+
+        return f(), g()
+
+    assert 'transform.apply_registered_pass "to_ppm"' in test_clifford_to_ppm_workflow.mlir
```
paul0403, line 418: If this test is not actually checking any math/computation, but just checkout the compiled IR, I would say this is more suited to be a lit test.

frontend/test/pytest/test_peephole_optimizations.py
sengthai: Good catch-up! This is because we can't compile those down to lower-level code (e.g LLVM IR), so we want the compiler to run only those PPM passes and return mlir.

paul0403: Thanks! We can keep this for now, and we can revisit about the ppm pipeline once #1758 is done!

mlir/lib/Catalyst/Transforms/RegisterAllPasses.cpp
```diff
@@ -65,4 +65,5 @@ void catalyst::registerAllCatalystPasses()
     mlir::registerPass(catalyst::createDecomposeNonCliffordPPRPass);
     mlir::registerPass(catalyst::createDecomposeCliffordPPRPass);
     mlir::registerPass(catalyst::createMBQCConversionPass);
+    mlir::registerPass(catalyst::createCliffordTToPPMPass);
```
paul0403, line 68: Minor: these are sorted in alphabetical order :) : (

mlir/lib/QEC/Transforms/clifford_t_to_ppm.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+        // Phase 1: Convert Clifford+T to PPR representation
+        {
+            ConversionTarget target(*ctx);
+            target.addIllegalDialect<quantum::QuantumDialect>();
+            target.addLegalOp<quantum::InitializeOp, quantum::FinalizeOp>();
+            target.addLegalOp<quantum::DeviceInitOp, quantum::DeviceReleaseOp>();
+            target.addLegalOp<quantum::AllocOp, quantum::DeallocOp>();
+            target.addLegalOp<quantum::InsertOp, quantum::ExtractOp>();
+            target.addLegalDialect<qec::QECDialect>();
```
paul0403, line 55: Why not in just one call of `addLegalOp`? Just wondering. 

If you are worried about line-too-long, you can let `clang-format` take care of this for you!

mlir/lib/QEC/Transforms/clifford_t_to_ppm.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+            if (failed(applyPatternsAndFoldGreedily(module, std::move(patterns)))) {
+                return signalPassFailure();
+            }
+        }
+
+        // Phase 4: Decompose non-Clifford PPRs into PPMs
+        {
+            RewritePatternSet patterns(ctx);
+            populateDecomposeNonCliffordPPRPatterns(patterns, decomposeMethod, avoidYMeasure);
```
paul0403, line 88: Dose tablegen automatically generate this `avoidYMeasure` pass option variable in cpp? :heart_eyes:

mlir/test/QEC/CliffordTToPPMTest.mlir
```diff
@@ -0,0 +1,110 @@
...
+// limitations under the License.
+
+// RUN: quantum-opt --to_ppm --split-input-file -verify-diagnostics %s > %t.ppm
+// RUN: quantum-opt --to_ppr --commute_ppr --merge_ppr_ppm --decompose_non_clifford_ppr --decompose_clifford_ppr --split-input-file -verify-diagnostics %s > %t.ppr
+// RUN: diff %t.ppm %t.ppr
+
+// With decompose-method=clifford-corrected and avoid-y-measure=false and max-pauli-size=3
+// RUN: quantum-opt --to_ppm="decompose-method=clifford-corrected avoid-y-measure=false max-pauli-size=3" --split-input-file -verify-diagnostics %s > %t.ppm
+// RUN: quantum-opt --to_ppr --commute_ppr="max-pauli-size=3" --merge_ppr_ppm="max-pauli-size=3" --decompose_non_clifford_ppr="decompose-method=clifford-corrected" --decompose_clifford_ppr="avoid-y-measure=false" --split-input-file -verify-diagnostics %s > %t.ppr
+// RUN: diff %t.ppm %t.ppr
```
paul0403, line 22: Same here

paul0403 (approved):
Nice and straightforward! From a technical side all good except the minor things I mentioend :heart_eyes:

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -666,3 +666,110 @@ def circuit():
...
+
+    The ``avoid_y_measure`` and ``decompose_method`` arguments are passed
+    to the :func:`~.passes.ppr_to_ppm` pass.
+    The ``max_pauli_size`` argument is passed to the :func:`~.passes.commute_ppr`
+    and :func:`~.passes.merge_ppr_ppm` passes.
+
+    Args:
+        qnode (QNode, optional): QNode to apply the pass to. If None, returns a decorator.
+        decompose_method (str, optional): The method to use for decomposing non-Clifford PPRs.
+            Options are ``"auto-corrected"`` and ``"clifford-corrected"``.
```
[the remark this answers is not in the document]
dime10: Ah yes this is a good point. Generally force-pushing is fine if you are trying to clean up some commits (a clean and well-written commit history is helpful for reviews and when going back to understand past changes), but you shouldn't rewrite history that has already been reviewed or that has already been referenced in comments, unless you really have to (rare).

mlir/lib/QEC/Transforms/clifford_t_to_ppm.cpp
dime10: I think it's alright, these ops are placed in logical groups so there is some sense to splitting them like this

mlir/test/QEC/CliffordTToPPMTest.mlir
```diff
@@ -0,0 +1,110 @@
...
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// RUN: quantum-opt --to_ppm --split-input-file -verify-diagnostics %s > %t.ppm
+// RUN: quantum-opt --to_ppr --commute_ppr --merge_ppr_ppm --decompose_non_clifford_ppr --decompose_clifford_ppr --split-input-file -verify-diagnostics %s > %t.ppr
+// RUN: diff %t.ppm %t.ppr
```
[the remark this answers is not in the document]
paul0403: > How about using the `tmp` directory?

Agreed!

paul0403, line 17: I also just realized this is just running `diff` without checking anything :sweat_smile: We should check that diff is empty right?

mlir/lib/QEC/Transforms/clifford_t_to_ppm.cpp
sengthai: Yes, tablegen does it for us. XD

mlir/test/QEC/CliffordTToPPMTest.mlir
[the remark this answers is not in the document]
sengthai: Let me try checking the streams directly first!

frontend/catalyst/passes/builtin_passes.py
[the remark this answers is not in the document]
sengthai: Here is the update corresponding to the review. [5fecb89](https://github.com/PennyLaneAI/catalyst/pull/1750/commits/5fecb895a911fe951f68f95a193b7c37a6e26599)

mlir/test/QEC/CliffordTToPPMTest.mlir
sengthai: Good point! 😅

doc/releases/changelog-dev.md
```diff
@@ -19,6 +19,19 @@
   * `decompose_non_clifford_ppr`: Decompose non-Cliford PPR (:math:`\theta = \tfrac{\pi}{8}`)
   into PPMs using a magic state.
 
+* A new compilation pass called :func:`~.passes.to_ppm` has been added to Catalyst to transform
```
dime10, line 22: I'm guessing the naming already came from a decision by @isaacdevlugt or @josh146, but I'm wondering if we don't want to highlight that the intention of this pass (pipeline) is to run the entire ppm compilation infrastructure, including future passes we may have. Assuming that is indeed our intention, something like `ppm_compilation` or similar would be more fitting, whereas *to*_ppm evokes something closer to a simple translation.

The text could also be adjusted to highlight that pass selection and ordering is automatically done for the user in this case.

frontend/test/pytest/test_peephole_optimizations.py
dime10: True, although one advantage this has is that it also doubles to satisfy code-coverage. Maybe is there is a cleaner way to split these concerns up though? 

I guess checking the `transform.apply_registered_pass "to_ppm"` made it into the IR would be the (py)test checking the frontend function works correctly, while looking for gates in the output of the transform is the more lit-like test?

But I don't mind too much either way here.

mlir/include/QEC/Transforms/Passes.td
```diff
@@ -109,4 +109,16 @@ def DecomposeCliffordPPRPass : Pass<"decompose_clifford_ppr"> {
...
+    let summary = "Convert CliffordT operations to Pauli Product Measurement operations.";
+    
+    let dependentDialects = [
+        "catalyst::qec::QECDialect",
+    ];
+
+    let constructor = "catalyst::createCliffordTToPPMPass()";
+
+    let options = [MaxPauliSizeOption, DecomposeMethodOption, AvoidYMeasureOption];
+}
```
dime10, line 123: empty line is in the top instead of bottom
```suggestion
def CliffordTToPPMPass : Pass<"to_ppm"> {
    let summary = "Convert CliffordT operations to Pauli Product Measurement operations.";
    
    let dependentDialects = [
        "catalyst::qec::QECDialect",
    ];

    let constructor = "catalyst::createCliffordTToPPMPass()";

    let options = [MaxPauliSizeOption, DecomposeMethodOption, AvoidYMeasureOption];
}

```

mlir/lib/QEC/Transforms/clifford_t_to_ppm.cpp
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        // Phase 5: Decompose Clifford PPRs into PPMs
+        {
+            RewritePatternSet patterns(ctx);
+            populateDecomposeCliffordPPRPatterns(patterns, avoidYMeasure);
+
+            if (failed(applyPatternsAndFoldGreedily(module, std::move(patterns)))) {
+                return signalPassFailure();
+            }
```
dime10, line 102: This works as a simple sequential combination of all passes, although at some point I was hoping we would do something a bit cleverer to benefit from fix-point iteration of multiple transformations. Maybe once we add optimizations I guess?

dime10 (commented):
Thanks @sengthai, pretty straightforward PR!

frontend/test/pytest/test_peephole_optimizations.py
paul0403: > True, although one advantage this has is that it also doubles to satisfy code-coverage.
Hmm, good point. Although for me, "ir checks should be lit tests" feels more important than codecov :thinking: 

> I guess checking the `transform.apply_registered_pass "to_ppm"` made it into the IR
Yes, I think this is our strategy for all of these frontend passes. Just check the successful registry in the `transform` dialect, no need to check pass effect. Although if @sengthai wants to check pass effect, I'm all for it.

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -669,3 +669,114 @@ def circuit():
...
+    The ``avoid_y_measure`` and ``decompose_method`` arguments are passed
+    to the :func:`~.passes.ppr_to_ppm` pass.
+    The ``max_pauli_size`` argument is passed to the :func:`~.passes.commute_ppr`
+    and :func:`~.passes.merge_ppr_ppm` passes.
+
+    Args:
+        qnode (QNode, optional): QNode to apply the pass to. If None, returns a decorator.
+        decompose_method (str, optional): The method to use for decomposing non-Clifford PPRs.
+            Options are ``"auto-corrected"`` and ``"clifford-corrected"``. Defaults to
+            Defaults to ``"auto-corrected"``.
```
joeycarter, line 697: You have "Defaults to" here twice :slightly_smiling_face: 

```suggestion
            Options are ``"auto-corrected"`` and ``"clifford-corrected"``. Defaults to
            ``"auto-corrected"``.
```

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -669,3 +669,114 @@ def circuit():
...
+            ``"clifford-corrected"`` uses a Clifford rotation for correction.
+        avoid_y_measure (bool): Rather than performing a Pauli-Y measurement for Clifford rotations
+            (sometimes more costly), a :math:`Y` state (:math:`Y\vert 0 \rangle`) is used instead
+            (requires :math:`Y` state preparation). Defaults to ``False``.
+        max_pauli_size (int): The maximum size of the Pauli strings after commuting or merging.
+            Defaults to 0 (no limit).
+
+    Returns:
+        ~.QNode or callable: Returns decorated QNode if qnode is provided,
+            otherwise returns a decorator.
```
joeycarter, line 708: The extra indentation here puts the first and second lines on separate paragraphs.

```suggestion
        ~.QNode or callable: Returns decorated QNode if qnode is provided,
        otherwise returns a decorator.
```

frontend/test/pytest/test_peephole_optimizations.py
sengthai: How about test on both in `lit` (for output correctness) and `pytest` (to ensure pass is registered correctly)?

mlir/lib/QEC/Transforms/clifford_t_to_ppm.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#define DEBUG_TYPE "to_ppm"
+
+#include "llvm/Support/Debug.h"
```
joeycarter, line 17: It looks like we're not using anything from this header. Does it still compile without this `#include`? If so we can remove it.

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -669,3 +669,114 @@ def circuit():
...
+
+    **Example**
+
+    If a merging resulted in a PPM acting on more than
+    ``max_pauli_size`` qubits (here, ``max_pauli_size = 2``), that merging would be skipped.
+    However, when decomposed into PPMs, at least one qubit will be applied, so the final
+    PPMs will act on at least plus one more qubit.
```
joeycarter, line 715: [Grammar]

```suggestion
    However, when decomposed into PPMs, at least one qubit will be applied, so the final
    PPMs will act on at least one additional qubit.
```

joeycarter (commented):
Looks great, thanks @sengthai! :100: 

I only had a few minor suggestions to add to the discussion.

doc/releases/changelog-dev.md
sengthai: > `ppm_compilation` or similar would be more fitting, whereas to_ppm evokes something closer to a simple translation.

Thanks @dime10, I agree! This pass aims to run the PPM's sub-passes, including one optimization (`merge_ppr_ppm`). So if we want to add more optimization to that pipeline, it doesn’t seem straightforward as `to_ppm` suggests.

mlir/lib/QEC/Transforms/clifford_t_to_ppm.cpp
sengthai: You are right! We can optimize this by applying phases 2 and 3 together in a fixed-point iteration, and do the same for phases 4 and 5.

sengthai: Also, when we add more heuristic optimization, such as weight/depth trade-offs, we may need to revisit this as the pipeline evolves.

frontend/catalyst/passes/builtin_passes.py
sengthai: Good catch! 😅

mlir/lib/QEC/Transforms/clifford_t_to_ppm.cpp
sengthai: Thanks Joey for pointing out. Yes, it can be compiled without that being included. I removed it in here: [813b411](https://github.com/PennyLaneAI/catalyst/pull/1750/commits/813b411a723ce6cbfb0f356623d344ed2ff68d73)

mlir/test/QEC/CliffordTToPPMTest.mlir
sengthai: I added additional checks for non-empty results here: [0b60ea4](https://github.com/PennyLaneAI/catalyst/pull/1750/commits/0b60ea4a480ed8543d796a2c59c4720868aed875)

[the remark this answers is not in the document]
paul0403: Actually, we checked and found out that the result mlir files will be dumped into the `mlir/build` folder, so these are git-ignored already.

I'm fine with keeping this as is now, but feel free to delete them if you still want to.

doc/releases/changelog-dev.md
isaacdevlugt: @dime10 yes @josh146 informed me of `ppm_compilation` and I really like it! Let's ship that 🚢

PR #1751, opened 2025-05-20
Migrate the entire bufferization pipeline to one-shot bufferization

paul0403:
**Context:**
This work is based on #1027.

Now that we have migrated all the individual dialects, we should migrate the entire bufferization pipeline.

The Quantum dialect was migrated in https://github.com/PennyLaneAI/catalyst/pull/1686 .
The Catalyst dialect was migrated in https://github.com/PennyLaneAI/catalyst/pull/1708 .
The `Gradient` dialect was migrated in https://github.com/PennyLaneAI/catalyst/pull/1740 .

See more context in #1027. 

**Related GitHub Issues:**
[sc-71487]


frontend/catalyst/pipelines.py
```diff
@@ -215,25 +215,29 @@ def get_quantum_compilation_stage(options: CompileOptions) -> List[str]:
...
-        "arith-bufferize",
+        "gradient-preprocess",
         "empty-tensor-to-alloc-tensor",
-        "func.func(bufferization-bufferize)",
-        "func.func(tensor-bufferize)",
+        ####################
+        "one-shot-bufferize{dialect-filter=gradient unknown-type-conversion=identity-layout-map}",
+        "one-shot-bufferize{dialect-filter=scf " + options + "}",
+        "one-shot-bufferize{dialect-filter=arith " + options + "}",
+        "one-shot-bufferize{dialect-filter=bufferization " + options + "}",
```
erick-xanadu, line 231: Check if this PR is useful to you here to add multiple dialects in the same filter: https://github.com/PennyLaneAI/catalyst/pull/1519

paul0403: (Hence the do-not-merge label, but I need CI so this isn't a draft PR :sweat_smile: )

paul0403:
Unfortunately seems like the patches for funcop interface is still needed, as llvm upstream reverted them :( 
https://github.com/llvm/llvm-project/commit/91c11574e87b5c0f434688edac01e9580ef99a92

paul0403:
Well, we just realized we don't seem to need to patches for bufferization in the first place.

paul0403:
All pytests pass locally for me as of the previous commit https://github.com/PennyLaneAI/catalyst/pull/1751/commits/d3a75ab4ade7676bcc8a2146bd71f6903b861c9c

paul0403:
Also, just a note that **:do-not-merge:** this PR into the base branch PR. I only target the base branch (gradient dialect's individual migration) so I don't have it as a blocker. I will rebase to target main once that PR is merged into main.

frontend/catalyst/pipelines.py
```diff
@@ -213,27 +213,27 @@ def get_quantum_compilation_stage(options: CompileOptions) -> List[str]:
...
     """Returns the list of passes that performs bufferization"""
+
+    bufferization_options = """bufferize-function-boundaries
+        allow-return-allocs-from-loops
+        function-boundary-type-conversion=identity-layout-map
+        unknown-type-conversion=identity-layout-map""".replace(
+        "\n", " "
+    )
+    if options.async_qnodes:
+        bufferization_options += " copy-before-write"
```
paul0403, line 226: https://github.com/PennyLaneAI/catalyst/blob/main/Makefile#L310 async tests do not go through codecov

mlir/lib/Driver/CompilerDriver.cpp
```diff
@@ -738,7 +738,8 @@ LogicalResult QuantumDriverMain(const CompilerOptions &options, CompilerOutput &
         TimingScope translateTiming = timing.nest("Translate");
         llvmModule =
             timer::timer(translateModuleToLLVMIR, "translateModuleToLLVMIR",
-                         /* add_endl */ false, *mlirModule, llvmContext, "LLVMDialectModule");
+                         /* add_endl */ false, *mlirModule, llvmContext, "LLVMDialectModule",
+                         /* disableVerification */ true);
```
dime10, line 742: Were there issues here with verification?

mlir/mlir-hlo
dime10, line 1: Is this right? Doesn't it normally just show the to and from commit hashes?
image

mlir/lib/Driver/Pipelines.cpp
```diff
@@ -67,43 +67,24 @@ void createQuantumCompilationPipeline(OpPassManager &pm)
...
+    options.unknownTypeConverterFn = [=](Value value, Attribute memorySpace,
+                                         const mlir::bufferization::BufferizationOptions &options) {
+        auto tensorType = cast<TensorType>(value.getType());
+        return bufferization::getMemRefTypeWithStaticIdentityLayout(tensorType, memorySpace);
+    };
+    pm.addPass(mlir::bufferization::createOneShotBufferizePass(options));
+    //////////////
     pm.addPass(mlir::createCanonicalizerPass());
     pm.addPass(catalyst::createGradientPostprocessingPass());
     pm.addNestedPass<mlir::func::FuncOp>(mlir::bufferization::createBufferHoistingPass());
```
dime10, line 90: Just looking at the remainder of these passes:
- `createBufferizationToMemRefPass`: Is this still needed in the new pipeline?
- `createArrayListToMemRefPass`: Was looking at this one but the ArrayList is its own data type so lowering it after bufferization is still correct I think 👍 
- After the looping hoisting, the docs recommend these 3 passes before deallocation, is that something we should look at? 
image

mlir/lib/Driver/Pipelines.cpp
```diff
@@ -67,43 +67,24 @@ void createQuantumCompilationPipeline(OpPassManager &pm)
...
+    mlir::bufferization::OneShotBufferizationOptions options;
+    options.bufferizeFunctionBoundaries = true;
+    options.allowReturnAllocsFromLoops = true;
+    options.setFunctionBoundaryTypeConversion(
+        mlir::bufferization::LayoutMapOption::IdentityLayoutMap);
+    options.unknownTypeConverterFn = [=](Value value, Attribute memorySpace,
+                                         const mlir::bufferization::BufferizationOptions &options) {
+        auto tensorType = cast<TensorType>(value.getType());
+        return bufferization::getMemRefTypeWithStaticIdentityLayout(tensorType, memorySpace);
+    };
```
dime10, line 85: Minor: In the textual pipeline we just use `unknown-type-conversion=identity-layout-map`, is there no equivalent C++ option like for the function boundary here?

dime10 (commented):
Great, this looks good as is 👌 Just wondering about the other recommended passes in the bufferization docs.

mlir/lib/Driver/Pipelines.cpp
dime10: Alright no worries :)

mlir/lib/Driver/Pipelines.cpp
paul0403: I tried these 3 recommended passes, the first two gave me errors:
- `buffer-results-to-out-params`: segfaults, unsure why
- `drop-equivalent-buffer-results`: some regular mlir level errors (aka uses after replacement/removal of `Value`s, etc), and jax `KeyError`s. This is probably because the number of results got changed by this pass, from what its name suggests. Also, since this pass deals with "buffer results", I suspect it is related to destination passing style, which we don't have yet.
- `promote-buffers-to-stack`: this one is actually fine to just slot into our pipeline, but with the first two failing, I was afraid that this one is "out-of-place", i.e. it might be designed to be used after the first two. What it actually does is quite simple, [it just promotes heap allocs to stack allocas](https://github.com/llvm/llvm-project/blob/8a198f89bfa0fb8225a3ffc90df8c7b9b722aeff/mlir/lib/Dialect/Bufferization/Transforms/BufferOptimizations.cpp#L442) (the code chain to go to the actual `create<memref::allocaOp>` is a bit long and hard to link here, just follow the things they are calling and you'll find it :) ), to eliminate "new/deletes".

paul0403: `createBufferizationToMemRefPass` is needed. See https://github.com/llvm/llvm-project/blob/8a198f89bfa0fb8225a3ffc90df8c7b9b722aeff/mlir/include/mlir/Conversion/Passes.td#L248, and https://discourse.llvm.org/t/bufferization-error-related-to-memref-clone/4665

TLDR: it's there to clean up something from `buffer-deallocation`

dime10 (approved):
🚀

PR #1752, opened 2025-05-20
Update llvm version

paul0403:
**Context:**
Update llvm version.
I am just trying the latest llvm main branch now.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


paul0403:
@mehrdad2m just a note that in the pass tutorial, we use the separate `match` and `rewrite`. I think we should update those too. https://docs.pennylane.ai/projects/catalyst/en/stable/dev/transforms.html#writing-transformations-on-catalyst-s-ir

We can do this in a separate PR if this one becomes too big.

.github/workflows/build-wheel-linux-x86_64.yaml
```diff
@@ -245,7 +245,7 @@ jobs:
               -DCMAKE_CXX_VISIBILITY_PRESET=default \
               -DCMAKE_CXX_FLAGS="-fuse-ld=lld"
 
-        cmake --build $GITHUB_WORKSPACE/enzyme-build --target EnzymeStatic-19
+        cmake --build $GITHUB_WORKSPACE/enzyme-build --target EnzymeStatic-20
```
paul0403, line 248: 21

mlir/Makefile
```diff
@@ -119,8 +119,14 @@ mhlo:
 	LIT_FILTER_OUT="chlo_legalize_to_mhlo" cmake --build $(MHLO_BUILD_DIR) --target check-mlir-hlo
 
 .PHONY: enzyme
+enzyme: TARGET_FILE := $(MK_DIR)/Enzyme/enzyme/Enzyme/TypeAnalysis/TypeAnalysis.cpp
+enzyme: PATCH_FILE := $(MK_DIR)/patches/enzyme-nvvm-fabs-intrinsics.patch
 enzyme:
 	@echo "build enzyme"
+	# Patch enzyme's dependency on nvidia fabs llvm intrinsics
```
paul0403, line 126: TODO: patch in CI scripts too

.dep-versions
```diff
@@ -1,16 +1,10 @@
...
-
+# To update JAX version alongside compatible dependency tags, run the following script:
+# python3 .github/workflows/set_dep_versions.py {JAX_version}
 jax=0.6.0
-mhlo=25b008569f413d76cfa8f481f3a84e82b89c47f4
-llvm=5f74671c85877e03622e8d308aee15ed73ccee7c
-enzyme=v0.0.149
+mhlo=617a9361d186199480c080c9e8c474a5e30c22d1
+llvm=179d30f8c3fddd3c85056fd2b8e877a4a8513158
+enzyme=v0.0.150 # not 150, just a random number to avoid CI's enzyme cache
```
paul0403, line 7: TODO

paul0403:
As of the previous commit https://github.com/PennyLaneAI/catalyst/pull/1752/commits/32220e5df24200615818e031bc82809f304da864, I can build:
```
mhlo=617a9361d186199480c080c9e8c474a5e30c22d1
llvm=179d30f8c3fddd3c85056fd2b8e877a4a8513158
enzyme=v0.0.180
```

In the above versions, the llvm and mhlo are spit out by the script looking for jax 0.6.0 versions. 
Enzyme is just latest (as of May 22, 2025), with commit db0181320d6e425ee963bd496ed0d8dbb615be18. This is also just enzyme 0.0.180 https://github.com/EnzymeAD/Enzyme/releases/tag/v0.0.180

Patches: 
In mhlo, I remove all shardy-related things.
In enzyme, I remove the nvdia llvm intrinsics.

paul0403:
At the previous commit, only the new buffer deallocation is left. Everything else should be good.

mlir/Makefile
```diff
@@ -95,11 +95,35 @@ mhlo: TARGET_FILE := $(MK_DIR)/mlir-hlo/mhlo/transforms/CMakeLists.txt
...
+	# make clean-mhlo
+	# TODO: probably clean this up
+	# git submodule mlir-hlo checkout -f 617a9361d186199480c080c9e8c474a5e30c22d1
+	#rm -rf $(MK_DIR)/mlir-hlo/mhlo/transforms/legalize_control_flow
+	#rm -rf $(MK_DIR)/mlir-hlo/mhlo/transforms/legalize_sort
+	#rm -rf $(MK_DIR)/mlir-hlo/mhlo/transforms/legalize_to_standard
+
 	# Patch MHLO cmake dependency
 	# TODO: remove once https://github.com/openxla/xla/pull/15446 is merged
 	@if patch --dry-run -p1 -N $(TARGET_FILE) $(PATCH_FILE) > /dev/null 2>&1; then \
```
mehrdad2m, line 108: Do we still need this patch since we have updated mhlo?

mlir/lib/Driver/Pipelines.cpp
```diff
@@ -73,26 +73,23 @@ void createBufferizationPipeline(OpPassManager &pm)
...
+    options.unknownTypeConversion = mlir::bufferization::LayoutMapOption::IdentityLayoutMap;
     pm.addPass(mlir::bufferization::createOneShotBufferizePass(options));
     //////////////
     pm.addPass(mlir::createCanonicalizerPass());
     pm.addPass(catalyst::createGradientPostprocessingPass());
     pm.addNestedPass<mlir::func::FuncOp>(mlir::bufferization::createBufferHoistingPass());
     pm.addNestedPass<mlir::func::FuncOp>(mlir::bufferization::createBufferLoopHoistingPass());
     pm.addNestedPass<mlir::func::FuncOp>(mlir::bufferization::createPromoteBuffersToStackPass());
-    pm.addNestedPass<mlir::func::FuncOp>(mlir::bufferization::createBufferDeallocationPass());
+    // TODO: what's the cpp for buffer-deallocation-pipeline???
```
mehrdad2m, line 89: you can probably use [this function](https://github.com/llvm/llvm-project/blob/02916a432ca6465e2e45f67be94f1c89c9cd425d/mlir/lib/Dialect/Bufferization/Pipelines/BufferizationPipelines.cpp#L21-L36) like:

```
mlir::bufferization::BufferDeallocationPipelineOptions bufferDeallocOptions;
mlir::bufferization::buildBufferDeallocationPipeline(pm, bufferDeallocOptions);
```

mehrdad2m (commented):
🚀🚀 🚀

mlir/test/Quantum/AllocationTest.mlir
```diff
@@ -12,25 +12,27 @@
...
 
-// RUN: quantum-opt --buffer-deallocation --split-input-file %s | FileCheck %s
+// RUN: quantum-opt --buffer-deallocation-pipeline --split-input-file %s | FileCheck %s
 
 // CHECK-LABEL: @existing_deallocation
-func.func @existing_deallocation() {
+func.func @existing_deallocation() -> i64 {
     // CHECK: [[reg:%.+]] = quantum.alloc
     %0 = quantum.alloc( 1) : !quantum.reg
+    %user = builtin.unrealized_conversion_cast %0 : !quantum.reg to i64
```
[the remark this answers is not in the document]
paul0403: I finished the description writeup. See point 3 of lit test updates.

mlir/lib/Quantum/Transforms/BufferizableOpInterfaceImpl.cpp
```diff
@@ -177,7 +177,7 @@ struct SampleOpInterface
         return false;
     }
 
-    bool bufferizesToAllocation(Operation *op, Value value) const { return true; }
+    // bool bufferizesToAllocation(Operation *op, Value value) const { return true; }
```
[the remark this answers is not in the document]
paul0403: It should be removed, thanks!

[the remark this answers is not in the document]
paul0403: See the last item in point 14 of the description:

> The migration guide also suggests that bufferizable ops no longer implement the bufferizesToAllocation method, so we remove them.

.github/workflows/check-catalyst.yaml
```diff
@@ -146,7 +146,7 @@ jobs:
         sudo apt-get update
         sudo apt-get install -y python3 python3-pip cmake ninja-build clang lld
         python3 --version | grep ${{ needs.constants.outputs.primary_python_version }}
-        python3 -m pip install numpy pybind11
+        python3 -m pip install numpy pybind11 nanobind==2.4
```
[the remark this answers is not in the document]
dime10: Definitely should put in in requirements if we now need this :)

Oh nice, so they have moved away from pybind as well? I think that means we could add back the extension modules to our bindings if we ever need to, since we removed them in order to support the ABI3 wheels with nanobind.

frontend/catalyst/pipelines.py
```diff
@@ -240,7 +240,7 @@ def get_bufferization_stage(options: CompileOptions) -> List[str]:
         "func.func(buffer-hoisting)",
         "func.func(buffer-loop-hoisting)",
         "func.func(promote-buffers-to-stack)",
-        "func.func(buffer-deallocation)",
+        "buffer-deallocation-pipeline",
```
dime10, line 243: This is potentially a big change that might have been worth putting in its own PR, like we did for the bufferization, independent of a version upgrade. First and foremost, it would be good to know whether the deallocation behaviour has changed at all compared to the previous pass.

mlir/Makefile
```diff
@@ -95,11 +95,35 @@ mhlo: TARGET_FILE := $(MK_DIR)/mlir-hlo/mhlo/transforms/CMakeLists.txt
...
+	mkdir -p $(MK_DIR)/mlir-hlo/mhlo/transforms/legalize_control_flow
+	mkdir -p $(MK_DIR)/mlir-hlo/mhlo/transforms/legalize_sort
+	mkdir -p $(MK_DIR)/mlir-hlo/mhlo/transforms/legalize_to_standard
+	touch $(MK_DIR)/mlir-hlo/mhlo/transforms/legalize_control_flow/legalize_control_flow.cc
+	touch $(MK_DIR)/mlir-hlo/mhlo/transforms/legalize_sort/legalize_sort.cc
+	touch $(MK_DIR)/mlir-hlo/mhlo/transforms/legalize_to_standard/legalize_to_standard.cc
+	touch $(MK_DIR)/mlir-hlo/mhlo/transforms/legalize_to_standard/legalize_to_standard_patterns.td
+	@if patch -p0 --dry-run -N --verbose < $(MK_DIR)/patches/mhlo-add-back-necessary-passes.patch > /dev/null 2>&1; then \
+		patch -p0 < $(MK_DIR)/patches/mhlo-add-back-necessary-passes.patch; \
+	fi
```
dime10: These are the patches? 😬 That's worse than I thought, I didn't know they just removed everything. Should be high priority on our backlog to address this situation.

mlir/Makefile
```diff
@@ -95,11 +95,35 @@ mhlo: TARGET_FILE := $(MK_DIR)/mlir-hlo/mhlo/transforms/CMakeLists.txt
...
 mhlo:
 	@echo "build MLIR-HLO"
+	# We have too many mhlo patches now, so let's clean it everytime.
+	# make clean-mhlo
+	# TODO: probably clean this up
+	# git submodule mlir-hlo checkout -f 617a9361d186199480c080c9e8c474a5e30c22d1
+	#rm -rf $(MK_DIR)/mlir-hlo/mhlo/transforms/legalize_control_flow
+	#rm -rf $(MK_DIR)/mlir-hlo/mhlo/transforms/legalize_sort
+	#rm -rf $(MK_DIR)/mlir-hlo/mhlo/transforms/legalize_to_standard
+
```
dime10: Are you deleting this? 
```suggestion
```

mlir/include/Catalyst/IR/CatalystOps.td
```diff
@@ -41,7 +41,8 @@ def ListDeallocOp : Catalyst_Op<"list_dealloc"> {
 def ListPushOp : Catalyst_Op<"list_push",
         [TypesMatchWith<"type of 'value' matches element type of 'list'",
                         "list", "value",
-                        "mlir::cast<ArrayListType>($_self).getElementType()">]> {
+                        "mlir::cast<ArrayListType>($_self).getElementType()">,
+         MemoryEffects<[MemWrite, MemAlloc]>]> {
```
dime10, line 45: I was gonna ask why the alloc is here, but I guess you're thinking about the case when the arraylist needs to be extended. In that case, wouldn't you also want a `MemFree`, since resizing involves alloc - copy - free?

The problem I'm thinking about though is that we have to distinguish between the semantics of an operation (its conceptual definition) and its _implementation_. With the reasoning above, you might start attaching memory semantics to all the tensor ops as well because they are implemented as memory reads/writes eventually, but that obviously that defeats the purpose of defining tensor semantics to begin with. 

If this op carries allocation semantics, where is the allocated resource exactly? The op doesn't have any results.

I think a more accurate model of this op is just as a "write" to the list argument. We could even define lists as their own resource separate from the built-in memory resource, to distinguish the actions of these ops from those of memref ops, but I don't actually know what practical difference that would make 🤔

mlir/include/Catalyst/IR/CatalystOps.td
```diff
@@ -71,7 +73,7 @@ def ListLoadDataOp : Catalyst_Op<"list_load_data"> {
     let assemblyFormat = [{ $list attr-dict `:` type($list) `->` type($data) }];
 }
 
-def PrintOp : Catalyst_Op<"print"> {
+def PrintOp : Catalyst_Op<"print", [MemoryEffects<[MemRead, MemAlloc]>]> {
```
dime10, line 76: Why is this an alloc?

mlir/include/Catalyst/IR/CatalystOps.td
```diff
@@ -81,7 +83,7 @@ def PrintOp : Catalyst_Op<"print"> {
     );
 }
 
-def AssertionOp : Catalyst_Op<"assert"> {
+def AssertionOp : Catalyst_Op<"assert", [MemoryEffects<[MemWrite]>]> {
```
dime10, line 86: Assertions are interesting, I think a "write" is the right choice to some abstract computer state.

mlir/lib/Driver/Pipelines.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/1752/commits/e1d3c962584ff3da639955456835d00ef4082165

mlir/include/Gradient/IR/GradientOps.td
dime10, line 1: I don't see a reason to attach alloc/free semantics to these ops, none of them are explicitly modelled to mean those things. Read/Write to the memref arguments is correct, although what's interesting is that in their "tensor" version these ops would not have such semantics and in fact be side-effect free. I think a way to model this (other than defining two sets of ops) might be to attach the effects to the arguments/results themselves, instead of the operation.

.github/workflows/check-catalyst.yaml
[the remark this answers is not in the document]
mehrdad2m: yes I think so. They still have pybind11 as a requirement, but this [PR](https://github.com/llvm/llvm-project/pull/120473/files) essentially ports all the Python core code to nanobind.

mlir/include/Quantum/IR/QuantumDialect.td
```diff
@@ -72,4 +73,24 @@ def ResultType : Quantum_Type<"Result", "res"> {
...
 
+//===----------------------------------------------------------------------===//
+// Quantum resource abstractions.
+//===----------------------------------------------------------------------===//
+
+def QuantumMemory : Resource<"QuantumMemory"> {
+    // This resource represents a generic piece of quantum memory. It can be used to
+    // model the resource bahavior of quantum operations, in order to help make decisions
+    // during various IR analyses.
+    // For example, a gate operation will both read from and write to quantum memory.
```
dime10, line 84: This is not true since we use value-semantics for our gates, although it is an alternative way to model gates.

mlir/include/Quantum/IR/QuantumOps.td
```diff
@@ -63,23 +60,23 @@ def NamedObservableAttr : EnumAttr<QuantumDialect, NamedObservable, "named_obser
 
 class Quantum_Op<string mnemonic, list<Trait> traits = []> : Op<QuantumDialect, mnemonic, traits>;
 
-def InitializeOp : Quantum_Op<"init"> {
+def InitializeOp : Quantum_Op<"init", [MemoryEffects<[MemAlloc<QuantumMemory>]>]> {
```
dime10, line 63: I think the runtime should be a different resource than quantum memory (it's a classical resource), and so should quantum devices (somewhat quantum, but not really quantum memory i.e. qubits) 🤔

mlir/include/Quantum/IR/QuantumOps.td
```diff
@@ -172,7 +169,7 @@ def DeallocQubitOp : Memory_Op<"dealloc_qb"> {
     }];
 }
 
-def DeallocOp : Memory_Op<"dealloc"> {
+def DeallocOp : Memory_Op<"dealloc", [MemoryEffects<[MemFree<QuantumMemory>]>]> {
```
dime10, line 172: The asymmetry between the alloc and dealloc really bothers me, but I guess it was one way achieve what we wanted (which was being able to DCE allocations if they are unused, but maintain deallocs as a valid use that should never be deleted).

mlir/include/Quantum/IR/QuantumOps.td
```diff
@@ -341,7 +338,8 @@ class UnitaryGate_Op<string mnemonic, list<Trait> traits = []> :
     let extraClassDeclaration = extraBaseClassDeclaration;
 }
 
-def SetStateOp : Gate_Op<"set_state", []> {
+def SetStateOp : Gate_Op<"set_state",
```
dime10, line 341: All gates should be pure (due the qubit value semantics).

Although, do the memrefs need a read?

dime10 (changes requested):
Leaving some initial comments on the first part of the PR :)

frontend/catalyst/pipelines.py
paul0403: I agree, migration of buffer deallocation is a change that should be in its own PR.

However, the old `--buffer-deallocation` is completely removed. https://github.com/llvm/llvm-project/pull/126366
Therefore we must address this, or we can't finish the llvm update.

paul0403: The deallocation behavior changed a lot: https://mlir.llvm.org/docs/OwnershipBasedBufferDeallocation/

... which is also why more time should be spent here. But I am unsure, given how much time we have spent on this update.

paul0403: The alternative strategy is to again, patch back the old buffer deallocation pass, and leave this as a TODO for next quarter's update, or just even a regular item. This would also mean we don't need to deal with the memory effects in this PR. 

Although I really want to avoid more patches :melting_face:

mlir/include/Quantum/IR/QuantumOps.td
erick-xanadu: There should be a runtime resource :sweat_smile:

mlir/include/Quantum/IR/QuantumDialect.td
dime10: Yeah it was unused previously, we probably introduced when trying out the memory modelling for registers.

mlir/include/Catalyst/IR/CatalystOps.td
paul0403: Prints will get removed if it's just memread since it has no results and thus no users. 

Between write and alloc, I thought alloc was more appropriate because sometimes it needs to alloc the space for the words it's printing.

dime10: I think write is more appropriate, since print can be modelled as writing to some IO buffer. We could put a write to the runtime resource here for example.

mlir/include/Catalyst/IR/CatalystOps.td
paul0403: Just a note that practically, ops without results (like assert here) and with no memory effects or read-only effects get removed. This is why I added memwrite here. My thinking was "it needs to write the assertion result somewhere". But I don't like this line of reasoning. Assertion being read-only feels natural to me but then they would get removed.

mlir/test/Quantum/AllocationTest.mlir
```diff
@@ -12,25 +12,27 @@
...
     // CHECK: quantum.dealloc [[reg]]
     quantum.dealloc %0 : !quantum.reg
-    return
+    return %user : i64
```
dime10, line 24: Just returning %0 directly should be sufficient to preserve the instructions from DCE 😅

mlir/lib/Quantum/Transforms/emit_catalyst_pyface.cpp
```diff
@@ -212,16 +212,21 @@ struct EmitCatalystPyInterfacePass
         MLIRContext *context = &getContext();
         RewritePatternSet patterns(context);
         patterns.add<EmitCatalystPyInterfaceTransform>(context);
+
         GreedyRewriteConfig config;
         config.strictMode = GreedyRewriteStrictness::ExistingOps;
         config.enableRegionSimplification = mlir::GreedySimplifyRegionLevel::Disabled;
         config.maxIterations = 1;
+        // Update to the following lines the next time we update llvm
```
dime10, line 220:
```suggestion
        // TODO: Update to the following lines the next time we update llvm
```

mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp
```diff
@@ -410,7 +411,9 @@ struct InlineNestedSymbolTablePass : PassWrapper<InlineNestedSymbolTablePass, Op
         GreedyRewriteConfig config;
         config.strictMode = GreedyRewriteStrictness::ExistingOps;
         config.enableRegionSimplification = mlir::GreedySimplifyRegionLevel::Disabled;
-
+        // Update to the following lines the next time we update llvm
```
dime10, line 414:
```suggestion
        // TODO: Update to the following lines the next time we update llvm
```

mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp
```diff
@@ -379,18 +379,19 @@ struct AnnotateWithFullyQualifiedNamePass
     void runOnOperation() override
     {
         MLIRContext *context = &getContext();
-
-        // Do not fold to save in compile time.
         GreedyRewriteConfig config;
         config.strictMode = GreedyRewriteStrictness::ExistingOps;
         config.enableRegionSimplification = mlir::GreedySimplifyRegionLevel::Disabled;
+        // Update to the following lines the next time we update llvm
```
dime10, line 385:
```suggestion
        // TODO: Update to the following lines the next time we update llvm
```

mlir/lib/Catalyst/Transforms/DetectQNodes.cpp
```diff
@@ -926,8 +920,11 @@ struct AddExceptionHandlingPass : impl::AddExceptionHandlingPassBase<AddExceptio
         GreedyRewriteConfig config;
         config.strictMode = GreedyRewriteStrictness::ExistingOps;
         config.enableRegionSimplification = mlir::GreedySimplifyRegionLevel::Disabled;
+        // Update to the following lines the next time we update llvm
```
dime10, line 923:
```suggestion
        // TODO: Update to the following lines the next time we update llvm
```

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
```diff
@@ -82,22 +90,43 @@ func::FuncOp AdjointLowering::discardAndReturnReg(PatternRewriter &rewriter, Loc
         rewriter.setInsertionPointAfter(callee);
         unallocFn =
             rewriter.create<func::FuncOp>(loc, fnName, fnType, visibility, nullptr, nullptr);
-        // clone the body.
+        // Clone the body.
+        // TODO: we do not support submodules, i.e. our gradient functions
+        // will have just one block.
+        assert(callee.getBody().hasOneBlock() &&
+               "Gradients with quantum submodules are not supported");
```
dime10, line 97: subroutines? 😅 
```suggestion
        // TODO: we do not support subroutines, i.e. our gradient functions
        // will have just one block.
        assert(callee.getBody().hasOneBlock() &&
               "Gradients with quantum subroutines are not supported");
```

Although this is odd, if you want to check that there are no subroutines, you should verify that there are no call ops (or call-like ops) in the body. Checking for a single block just asserts that there is no unstructured control flow.

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
```diff
@@ -68,10 +65,21 @@ func::FuncOp AdjointLowering::discardAndReturnReg(PatternRewriter &rewriter, Loc
...
+    // removing the expval op in the body.
+    // TODO: we only support grad on expval op for now
     std::string fnName = callee.getName().str() + ".nodealloc";
     Type qregType = quantum::QuregType::get(rewriter.getContext());
-    FunctionType fnType = rewriter.getFunctionType(callee.getArgumentTypes(), qregType);
+
+    Type f64Type = rewriter.getF64Type();
+    SmallVector<Type> retTypes{qregType};
+    auto expvalOps = callee.getOps<quantum::ExpvalOp>();
+    std::for_each(expvalOps.begin(), expvalOps.end(),
```
dime10, line 79: We should assert that our assumption is true, i.e. that there are no other measurement processes besides expval.

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
```diff
@@ -82,22 +90,43 @@ func::FuncOp AdjointLowering::discardAndReturnReg(PatternRewriter &rewriter, Loc
...
-                rewriter.eraseOp(op);
-            }
-            else if (isa<func::ReturnOp>(op)) {
-                op->setOperands(localDealloc.getOperand());
-            }
-        });
+        SmallVector<Value> returnVals;
+
+        // Fine for now: only one block in body so only one quantum.dealloc
+        auto localDealloc = *unallocFn.getOps<quantum::DeallocOp>().begin();
```
dime10, line 105: Grab the terminator directly, it's faster faster than iterating to find ops of a particular type :)

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
```diff
@@ -82,22 +90,43 @@ func::FuncOp AdjointLowering::discardAndReturnReg(PatternRewriter &rewriter, Loc
...
+        returnVals.push_back(localDealloc.getOperand());
+
+        // We also need to return the expval op so it won't be dead-code-eliminated
+        // TODO: we only support grad on expval op for now
+        for (auto expvalOp : unallocFn.getOps<quantum::ExpvalOp>()) {
```
dime10, line 110: Wait if we're already iterating through the expval ops above (to get the types), can't you do the collection there too?

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
```diff
@@ -82,22 +90,43 @@ func::FuncOp AdjointLowering::discardAndReturnReg(PatternRewriter &rewriter, Loc
...
+            returnVals.push_back(expvalOp);
+        }
+
+        // Erase the device release.
+        // Unfortunately erasing on the fly within the getOps() causes the
+        // "erasing while iterating" segfault
+        // So we need a separate, completely materialized worklist.
```
dime10, line 117: With the `.walk` method from before it was possible though?

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
```diff
@@ -82,22 +90,43 @@ func::FuncOp AdjointLowering::discardAndReturnReg(PatternRewriter &rewriter, Loc
...
+        std::for_each(deviceReleaseOps.begin(), deviceReleaseOps.end(),
+                      [&](const quantum::DeviceReleaseOp &op) {
+                          deviceReleaseOpsEraseWorklist.push_back(op);
+                      });
+        std::for_each(deviceReleaseOpsEraseWorklist.begin(), deviceReleaseOpsEraseWorklist.end(),
+                      [&](const quantum::DeviceReleaseOp &op) { rewriter.eraseOp(op); });
+
+        // Create the return
+        // Again, assume just one block for now
+        (*unallocFn.getOps<func::ReturnOp>().begin())->setOperands(returnVals);
```
dime10, line 129: Just generally it's good to be mindful of how often you're iterating over a possibly large piece of IR, as that can start to add up, especially once a function is called by other functions which can quickly become quadratic (not necessarily the case here, just in general).

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
dime10: Oh right, I remember now. I think you wanted to check that there is only a single block so that we can assume there is only one return statement? Which I think is true anyway, as we only deal with structured control flow until going to llvm, so that makes sense to assert.

dime10 (commented):
Great work, especially the detailed PR description 🥇

mlir/test/Quantum/AllocationTest.mlir
paul0403: https://github.com/PennyLaneAI/catalyst/pull/1752/commits/ab3103496df4ea7c37b46d72d748d19d800e5101

mlir/include/Catalyst/IR/CatalystOps.td
paul0403: Hmm, makes sense, I guess assert does change the "computer state" depending on whether it's true or false. All good then. 

I'm resolving this as there's no change needed to the code.

paul0403:
I split the new buffer dealloc migration into its own PR #1778 and added back the old buffer dealloc.

I am therefore closing all comments regarding the memory effects. We will properly revisit them the next time...

mlir/lib/Catalyst/Transforms/BufferizableOpInterfaceImpl.cpp
```diff
@@ -75,8 +75,6 @@ struct PrintOpInterface
 struct CustomCallOpInterface
     : public bufferization::BufferizableOpInterface::ExternalModel<CustomCallOpInterface,
                                                                    CustomCallOp> {
-    bool bufferizesToAllocation(Operation *op, Value value) const { return true; }
```
paul0403, line 78: I guess we should add these `bufferizesToAllocation` methods back now that we are not migrating this time.

mlir/include/Quantum/IR/QuantumDialect.td
paul0403: Yes, I just realized that these memory effects are only IR side effects used for hints during IR transformations, e.g. ops ordering, ops removal/cse, ..., and maybe shouldn't be taken too literally?

According to [side effect docs](https://mlir.llvm.org/docs/Rationale/SideEffectsAndSpeculation/):

> Many MLIR operations don’t exhibit any behavior other than consuming and producing SSA values. These operations can be reordered with other operations as long as they obey SSA dominance requirements and can be eliminated or even introduced (e.g. for [rematerialization](https://en.wikipedia.org/wiki/Rematerialization)) as needed.
However, a subset of MLIR operations have implicit behavior than isn’t reflected in their SSA data-flow semantics. These operations need special handing, and cannot be reordered, eliminated or introduced without additional analysis.

So it sounds like these side effects are additional hints on top of the usual information available from SSA data flow.

So yes for the gate ops we have it would indeed mean the side effects are not necessary, as we chain gates up (and lock down their order) with explicit SSA semantics already. 

I will remove this line in the description.

paul0403: https://github.com/PennyLaneAI/catalyst/pull/1752/commits/0f54789e513f9cc449642cdc77f4ff0d4c206df8

mlir/Makefile
```diff
@@ -95,11 +95,24 @@ mhlo: TARGET_FILE := $(MK_DIR)/mlir-hlo/mhlo/transforms/CMakeLists.txt
 mhlo: PATCH_FILE := $(MK_DIR)/patches/mhlo-Add-PassesIncGen-in-transforms-CMakeList.patch
 mhlo:
 	@echo "build MLIR-HLO"
+	# Clean mhlo submodule for git patches
+	make clean-mhlo
+
```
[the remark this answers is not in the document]
paul0403: I found the magic incantation that dry-runs for `git apply`: 
https://git-scm.com/docs/git-apply#Documentation/git-apply.txt---check

https://github.com/PennyLaneAI/catalyst/pull/1752/commits/e9fbb2d4e8cac0f68029a0392bdd080c530e4a0c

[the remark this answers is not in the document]
paul0403: (so I removed this clean)

mlir/Makefile
paul0403: Good catch, this is no longer needed!

paul0403: https://github.com/PennyLaneAI/catalyst/pull/1752/commits/b04b0592051f183f1bef9680bf74b4da32317426

.github/workflows/check-catalyst.yaml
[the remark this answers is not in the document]
paul0403: https://github.com/PennyLaneAI/catalyst/pull/1752/commits/b85e6a1649c451af73c3b81a1a02e422b3eb1888

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
paul0403: > I think you wanted to check that there is only a single block so that we can assume there is only one return statement?

Yes

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
paul0403: That was in `callee`, but this is in the cloned `unallocFn`, so they are unrelated onjects in cpp.

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
paul0403: But `DeallocOp` is not a terminator right?

mlir/include/Quantum/IR/QuantumDialect.td
dime10: Agreed, except that they are not hints. Misclassifying effects can lead to miscompilation.

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
dime10: Oh sorry, I probably confused this line with the one that gets the return op 👍

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
dime10: My bad 👍

paul0403: Although, since this is a clone, I do wonder whether it's possible to do something with the mapper, so we save the linear search in the clone.

Let me try it.

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/1752/commits/5b3205769a611eb956b889acfb45c2d14c27ed12

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/1752/commits/4cc38e48ee385d2f2836704737896bc9c902cbf6

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
paul0403: No longer need this extra worklist since I use the mapper from the clone now

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
paul0403: The mapper eliminates the need to iterating over the clone entirely!

I now loop over the original `callee` (the thing being cloned) once at the beginning to the the expvals and deallocs (this iterating needs to happen at least once), and retrival of corresponding ops in the cloned `unallocFn` are all O(1) from the mapper.

paul0403: https://github.com/PennyLaneAI/catalyst/pull/1752/commits/034098e969957399d40cb3fe1d1cf77d291d417a

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
paul0403: done in https://github.com/PennyLaneAI/catalyst/pull/1752/commits/034098e969957399d40cb3fe1d1cf77d291d417a

doc/releases/changelog-dev.md
```diff
@@ -175,6 +175,14 @@
     [(#1671)](https://github.com/PennyLaneAI/catalyst/pull/1671)
     [(#1681)](https://github.com/PennyLaneAI/catalyst/pull/1681)
 
+* (Compiler developers only) The version of LLVM, MHLO and Enzyme used by Catalyst is
```
dime10: I know we call it MHLO colloquially, but since this is a changelog we might want to use the actual name of the project which is `mlir-hlo` (mhlo is just one of the dialects).

requirements.txt
```diff
@@ -5,7 +5,7 @@ pip>=22.3
 # Build dependencies for non-Python components
 # Do not allow NumPy 2.0.0 due to a bug with their C API that blocks the usage of the Stable ABI;
 # this bug was fixed in 2.0.1 (https://github.com/numpy/numpy/pull/26995)
-nanobind
+nanobind==2.4
 numpy!=2.0.0
```
dime10: Ideally if we are pinning we should say why. We also try to be as lenient as possible, is 2.4 required exactly or is there a min/max that's required? 

```suggestion
numpy!=2.0.0
# minimum required by MLIR
nanobind>=2.4
```

(I'm also putting the numpy entry below its comment again.)

mlir/test/Gradient/AdjointTest.mlir
dime10, line 1: What's the point of these test updates?

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
```diff
@@ -47,15 +44,44 @@ void AdjointLowering::rewrite(func::FuncOp op, PatternRewriter &rewriter) const
...
+    return success();
 }
 
 func::FuncOp AdjointLowering::discardAndReturnReg(PatternRewriter &rewriter, Location loc,
                                                   func::FuncOp callee)
 {
+    // TODO: we do not support subroutines, i.e. our gradient functions
+    // will have just one block.
+    assert(callee.getBody().hasOneBlock() &&
+           "Gradients with quantum subroutines are not supported");
```
dime10: one block != no subroutines (i.e. function calls)

one block == no unstructured control flow

mlir/include/Quantum/IR/QuantumOps.td
```diff
@@ -567,7 +564,7 @@ def GlobalPhaseOp : UnitaryGate_Op<"gphase", [DifferentiableGate, AttrSizedOpera
 }
 
 def MultiRZOp : UnitaryGate_Op<"multirz", [DifferentiableGate, NoMemoryEffect,
-                                           AttrSizedOperandSegments, AttrSizedResultSegments]> {
+                            AttrSizedOperandSegments, AttrSizedResultSegments]> {
```
dime10: I think the previous formatting makes more sense no, the alignment is kind of arbitrary here?

mlir/Makefile
```diff
@@ -204,6 +206,7 @@ clean-llvm:
 clean-mhlo:
 	@echo "clean HLO dialect build files"
 	rm -rf $(MHLO_BUILD_DIR)
+	cd mlir-hlo; git clean -fd; git checkout .; \
```
dime10: Is that supposed to be there? 
```suggestion
	cd mlir-hlo; git clean -fd; git checkout .
```

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
```diff
@@ -47,15 +44,44 @@ void AdjointLowering::rewrite(func::FuncOp op, PatternRewriter &rewriter) const
...
+            deallocs.push_back(dyn_cast<quantum::DeallocOp>(op));
+            continue;
+        }
+        else if (isa<quantum::MeasurementProcess>(op)) {
+            if (isa<quantum::ExpvalOp>(op)) {
+                expvalOps.push_back(dyn_cast<quantum::ExpvalOp>(op));
+                continue;
+            }
+            else {
+                callee.emitOpError() << "Adjoint gradient is only supported on expval measurements";
```
dime10, line 78: Returning makes sense right?
```suggestion
                callee.emitOpError() << "Adjoint gradient is only supported on expval measurements";
                return failure();
```

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
```diff
@@ -47,15 +44,44 @@ void AdjointLowering::rewrite(func::FuncOp op, PatternRewriter &rewriter) const
...
+    // TODO: we only support grad on expval op for now
     SmallVector<quantum::DeallocOp> deallocs;
-    for (auto op : callee.getOps<quantum::DeallocOp>()) {
-        deallocs.push_back(op);
-    };
+    SmallVector<quantum::ExpvalOp> expvalOps;
+    SmallVector<quantum::DeviceReleaseOp> deviceReleaseOps;
+    for (Operation &op : callee.getBody().getOps()) {
+        if (isa<quantum::DeallocOp>(op)) {
+            deallocs.push_back(dyn_cast<quantum::DeallocOp>(op));
```
dime10: Minor, but if you already checked there is no reason to dyn cast
```suggestion
            deallocs.push_back(cast<quantum::DeallocOp>(op));
```

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
```diff
@@ -47,15 +44,44 @@ void AdjointLowering::rewrite(func::FuncOp op, PatternRewriter &rewriter) const
...
+    //
+    // We also need to return the expval to avoid dead code elimination downstream from
+    // removing the expval op in the body.
+    // TODO: we only support grad on expval op for now
     SmallVector<quantum::DeallocOp> deallocs;
-    for (auto op : callee.getOps<quantum::DeallocOp>()) {
-        deallocs.push_back(op);
-    };
+    SmallVector<quantum::ExpvalOp> expvalOps;
+    SmallVector<quantum::DeviceReleaseOp> deviceReleaseOps;
```
dime10, line 66: There should only be one device release no?

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
```diff
@@ -67,11 +93,18 @@ func::FuncOp AdjointLowering::discardAndReturnReg(PatternRewriter &rewriter, Loc
         return callee;
     }
 
-    // Since the return value is guaranteed to be discarded, then let's change the return type
-    // to be only the quantum register.
+    size_t numExpvals = expvalOps.size();
+    if (numExpvals != 1) {
+        callee.emitOpError() << "Invalid number of expval ops: " << numExpvals;
+        return callee;
+    }
```
dime10: Why can't there be multiple expvals? I think that is allowed.

mlir/lib/Catalyst/Transforms/TBAAPatterns.cpp
```diff
@@ -12,20 +12,77 @@
...
+bool isFromExtractAlignedPointerAsIndexOp(Operation *op)
+{
+    if (isa<memref::ExtractAlignedPointerAsIndexOp>(op)) {
+        return true;
+    }
+    auto prevOp = op->getPrevNode();
+    if (prevOp) {
+        return isFromExtractAlignedPointerAsIndexOp(prevOp);
+    }
+    return false;
```
dime10: Does this walk the entire list of ops just to check if the index value came from a memref? I don't think it even uses use-def chains, just the ordered list of ops in a block (not 100% sure on that one).

Now that I think about it, what this function does is check whether the block contains _any_ `ExtractAligned...Op` prior to the load/store op, regardless of whether it defined the index value we are currently interested in.

mlir/lib/Catalyst/Transforms/TBAAPatterns.cpp
```diff
@@ -12,20 +12,77 @@
...
+    }
+    else if (isa<memref::StoreOp>(op)) {
+        mem = dyn_cast<memref::StoreOp>(op).getMemref();
+    }
+
+    auto funcArgs = parentOp.getCallableRegion()->getArguments();
+    bool memIsFuncArg = (std::find(funcArgs.begin(), funcArgs.end(), mem) != funcArgs.end());
+    if (!memIsFuncArg) {
+        return false;
+    }
```
dime10: Since we're checking this already, we might as well be specific. We know that only arguments one and two of the function are pointer memrefs.

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
paul0403: Good catch, for some reason I though only `dyn_cast` would compile because I'm going from a more general type (`Operation`) to a more specific one and I need to check it in runtime. But simple `cast` also compiles here.

mlir/Makefile
paul0403: Good catch :sweat_smile:

paul0403: https://github.com/PennyLaneAI/catalyst/pull/1752/commits/8ff300f79d6d97f567d67e05bb35ba0fbdea259c

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/1752/commits/7a7e7632bf191457d5667f078ccbb061c1be8f07

requirements.txt
paul0403: @mehrdad2m has more context on this, is 2.4 required exactly?

mlir/test/Gradient/AdjointTest.mlir
paul0403: I was checking for exactly one expval op.

But I now realize multiple expval is ok :sweat_smile: It will just be a jacobian. I will change them back.

mlir/include/Quantum/IR/QuantumOps.td
paul0403: This is a relic of the memory effect changes. I'll indent them back.

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
paul0403: Ah yes, returning right away here does make sense. I will add this.

mlir/include/Quantum/IR/QuantumOps.td
paul0403: https://github.com/PennyLaneAI/catalyst/pull/1752/commits/90a184f750333233959f5419acf0c4786a39fb0a

doc/releases/changelog-dev.md
paul0403: https://github.com/PennyLaneAI/catalyst/pull/1752/commits/ed103bdfd959d9552f6631201ea2008701480294

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/1752/commits/5d249e2f8a1c387ced03ed164acbb6e61fdbcce1

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/1752/commits/13853d529d759797ed860e1ed1f7d68acd257161

mlir/test/Gradient/AdjointTest.mlir
paul0403: I updated so that (1) number of expvals is not constrained, (2) we check for there's just one device release, and (3) revert these test updates

https://github.com/PennyLaneAI/catalyst/pull/1752/commits/1240c6a0285339fc63d1aa659c996e23af1700fc

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
paul0403: I re-enabled multiple expvals. See the other comment.

.github/workflows/build-wheel-linux-x86_64.yaml
```diff
@@ -210,9 +210,11 @@ jobs:
...
         export PATH=$GITHUB_WORKSPACE/llvm-build/bin:$PATH
 
-        export TARGET_FILE=mlir/mlir-hlo/mhlo/transforms/CMakeLists.txt
-        export PATCH_FILE=mlir/patches/mhlo-Add-PassesIncGen-in-transforms-CMakeList.patch
-        if patch --dry-run -p1 -N $TARGET_FILE $PATCH_FILE > /dev/null 2>&1; then patch -p1 $TARGET_FILE $PATCH_FILE; fi
+        pushd .
+        cd $GITHUB_WORKSPACE/mlir/mlir-hlo
+        git apply $GITHUB_WORKSPACE/mlir/patches/mhlo-remove-shardy.patch
+        git apply $GITHUB_WORKSPACE/mlir/patches/mhlo-add-back-necessary-passes.patch
+        popd
```
erick-xanadu: I've never seen `pushd` used this way to push the current working directory :thinking: 

```suggestion
        pushd $GITHUB_WORKSPACE/mlir/mlir-hlo
        git apply $GITHUB_WORKSPACE/mlir/patches/mhlo-remove-shardy.patch
        git apply $GITHUB_WORKSPACE/mlir/patches/mhlo-add-back-necessary-passes.patch
        popd
```

mlir/lib/Catalyst/Transforms/TBAAPatterns.cpp
erick-xanadu: `getPrevNode` returns the previous operation in the basic block. It is better to use SSA values. Maybe backwards slice may help?

mlir/lib/Catalyst/Transforms/TBAAPatterns.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/1752/commits/24d96ca1d9edf175202bc25155a20443974227f5, https://github.com/PennyLaneAI/catalyst/pull/1752/commits/b6833bf60b6c5a5059c6b8accb3079ccbcaee89d

requirements.txt
dime10: MLIR seems to require this: `nanobind>=2.4, <3.0`

Unless there is a reason to be more specific I would say we use this

dime10: Actually 3 hasn't even been released yet, so I'm guessing they do this as a precaution. Since we don't really follow this practice, I would say just `nanobind>=2.4` is also fine.

Updated the suggestion

mlir/lib/Catalyst/Transforms/TBAAPatterns.cpp
paul0403: I'm updating this to backward slice as we speak

paul0403: https://github.com/PennyLaneAI/catalyst/pull/1752/commits/f7455ddec9458cdab61207d8c932b2939d76cfaf

.github/workflows/build-wheel-linux-x86_64.yaml
paul0403: good point :sweat_smile:

paul0403: https://github.com/PennyLaneAI/catalyst/pull/1752/commits/76613a2a0f55fd160d7bc2ad3529127b24a19558

paul0403: Wait this didn't work...
```
pushd $GITHUB_WORKSPACE/mlir/mlir-hlo
  git apply $GITHUB_WORKSPACE/mlir/patches/mhlo-remove-shardy.patch
  git apply $GITHUB_WORKSPACE/mlir/patches/mhlo-add-back-necessary-passes.patch
  popd
  
  env:
    BUILD_STANDALONE_PLUGIN: false
/__w/catalyst/catalyst/mlir/mlir-hlo /__w/catalyst/catalyst
error: patch failed: bindings/c/CMakeLists.txt:10
error: bindings/c/CMakeLists.txt: patch does not apply
error: patch failed: stablehlo_ext/CMakeLists.txt:[12](https://github.com/PennyLaneAI/catalyst/actions/runs/15422051646/job/43399511338?pr=1752#step:17:12)
error: stablehlo_ext/CMakeLists.txt: patch does not apply
error: patch failed: stablehlo_ext/analysis/CMakeLists.txt:1
error: stablehlo_ext/analysis/CMakeLists.txt: patch does not apply
error: patch failed: stablehlo_ext/transforms/CMakeLists.txt:20
error: stablehlo_ext/transforms/CMakeLists.txt: patch does not apply
error: patch failed: stablehlo_ext/transforms/stablehlo_refine_shapes.cpp:34
error: stablehlo_ext/transforms/stablehlo_refine_shapes.cpp: patch does not apply
error: patch failed: tests/lit.cfg.py:32
error: tests/lit.cfg.py: patch does not apply
error: patch failed: tools/mlir-hlo-opt/mlir-hlo-opt.cc:[20](https://github.com/PennyLaneAI/catalyst/actions/runs/15422051646/job/43399511338?pr=1752#step:17:20)
error: tools/mlir-hlo-opt/mlir-hlo-opt.cc: patch does not apply
```

paul0403: Ok it's because we cache the source and the patch is being applied to the cache :sweat_smile:

requirements.txt
paul0403: https://github.com/PennyLaneAI/catalyst/pull/1752/commits/38cd1291ffcacb932d1658bf51b11d2a20b3e412 updated to >=2.4

paul0403: added a comment explaining https://github.com/PennyLaneAI/catalyst/pull/1752/commits/0084fe745b662f2a3b5c0df565632f58bd91bb4c

requirements.txt
```diff
@@ -5,7 +5,8 @@ pip>=22.3
 # Build dependencies for non-Python components
 # Do not allow NumPy 2.0.0 due to a bug with their C API that blocks the usage of the Stable ABI;
 # this bug was fixed in 2.0.1 (https://github.com/numpy/numpy/pull/26995)
-nanobind
+# llvm requires nanobind 2.4 or higher
+nanobind>=2.4
 numpy!=2.0.0
```
dime10:
```suggestion
numpy!=2.0.0
# llvm requires nanobind 2.4 or higher
nanobind>=2.4
```

.github/workflows/check-catalyst.yaml
dime10:
```suggestion
        python3 -m pip install numpy pybind11 nanobind
```

dime10 (approved):

dime10:
I'm investigating issue #2420, and I'm pretty sure this PR introduced the bug.

Looking at this change specifically, I think it misinterpreted the restriction in place on the adjoint method:
>13. The new pattern applicator (`applyPatternsGreedily`) is more
aggressive in dead code elimination,
and is eliminating dead `Value`s in the adjoint gradient method.
The `nodealloc` function we generate for adjoint gradient lowering used
to only return the qreg, not the expval result. This causes the expval
op to be eliminated since it has no users.
This further causes wrong gradient results, since the entire program,
all ops included (regardless
of dead or not), impacts the gradient through chain rule.
To avoid this, we return the expval result as well.
In doing this, **we implicitly assume that differentiated qnodes can only
return expval.
Although this assumption is true and also restricted by frontend,**
ideally we should not have it hard coded.
We leave this as a TODO for a future feature.

Specifically, we had:
```cpp
        // The callee of the adjoint op must return as a single result the quantum register.
        func::FuncOp callee =
            SymbolTable::lookupNearestSymbolFrom<func::FuncOp>(op, op.getCalleeAttr());
        assert(callee && callee.getNumResults() == 1 && "invalid qfunc symbol in adjoint op");
```
where the generated function only returned the register, but this did not restrict how many QNode MP results could be produced. Now the check reads:
```cpp
        // The callee of the adjoint op must return 2 results: the quantum register and the expval.
        func::FuncOp callee =
            SymbolTable::lookupNearestSymbolFrom<func::FuncOp>(op, op.getCalleeAttr());
        assert(callee && callee.getNumResults() == 2 && "invalid qfunc symbol in adjoint op");
```

In the frontend we only ever had this verification:
```py
    if f.diff_method == "adjoint" and any(prim not in [expval_p] for prim in return_ops):
        raise DifferentiableCompileError(
            "The adjoint method can only be used for QNodes which return qml.expval."
        )
```
It only checks that all results are expvals, but not how many there are.

Looking into whether we can reverse this change / fix the problem that prompted this change another way.

PR #1756, opened 2025-05-21
Set up copybutton options to exclude prompts and output

josh146:


dime10:
While it strips the output, it seems to leave behind empty lines for each line of console output. Are we able to remove those as well?

josh146:
@dime10 I've edited the settings to try and fix that

josh146:
Works now!

dime10 (approved):
Awesome, works flawlessly 🎉 🎉

PR #1758, opened 2025-05-21
Add a builtin MBQC pipeline for MBQC workloads

joeycarter:
**Context:** When we initially added the MBQC dialect, we had to manually specify the full compilation pipeline add drop in the MBQC-to-LLVM dialect-conversion pass, `"convert-mbqc-to-llvm"`, to convert the ops in the `mbqc` dialect to the LLVM dialect. For example:

```python
mbqc_pipeline = [
    (
        "default-pipeline",
        [
            "enforce-runtime-invariants-pipeline",
            "hlo-lowering-pipeline",
            "quantum-compilation-pipeline",
            "bufferization-pipeline",
        ],
    ),
    (
        "mbqc-pipeline",
        [
            "convert-mbqc-to-llvm",
        ],
    ),
    (
        "llvm-dialect-lowering-pipeline",
        [
            "llvm-dialect-lowering-pipeline",
        ],
    ),
]

@qml.qnode(dev)
def workload():
    ...
```

No one is going to want to type that out each time, so it would be handy to have this pipeline already defined somewhere.

**Description of the Change:** Adds a new function, `mbqc_pipeline()` to a new `passes.builtin_pipelines` module. This function first makes a copy of the default pipeline stages (from `catalyst.pipelines.get_stages()`) and inserts the mbqc-to-llvm conversion pass immediately before the quantum-to-llvm conversion pass in the `"MLIRToLLVMDialect"` stage.

This PR also adds a few helper functions, e.g. `catalyst.pipelines.insert_before_pass()` and `catalyst.pipelines.insert_after_pass()` to aid in manipulating existing pass pipelines.

**Benefits:** Less typing to specify the pass pipeline for MBQC workloads (and therefore less error-prone!)

### DIFF
--- a/frontend/catalyst/passes/builtin_pipelines.py
+++ b/frontend/catalyst/passes/builtin_pipelines.py
@@ -1,0 +1,57 @@
+# Copyright 2025 Xanadu Quantum Technologies Inc.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""This module exposes built-in Catalyst MLIR pass pipelines to the frontend."""
+
+from typing import List, Tuple
+
+from catalyst.pipelines import CompileOptions, get_stages, insert_before_pass
+
+PipelineStages = List[Tuple[str, List[str]]]
+
+
+def mbqc_pipeline() -> PipelineStages:
+    """Return the pipeline stages for MBQC workloads.
+
+    The MBQC pipeline is identical to the default Catalyst pipeline, but with the MBQC-to-LLVM
+    dialect-conversion pass inserted immediately before the Quantum-to-LLVM dialect-conversion pass
+    in the ``MLIRToLLVMDialect`` pipeline stage.
+
+    Returns:
+        PipelineStages: The list of pipeline stages.
+    """
+    options = CompileOptions()  # Use all default compile options
+    stages = get_stages(options)  # Get list of default pipeline stages
+
+    # Find the MLIR-to-LLVM dialect-conversion stage, 'MLIRToLLVMDialect'
+    stage_names = [item[0] for item in stages]
+    llvm_dialect_conversion_stage_name = "MLIRToLLVMDialect"
+
+    assert (
+        llvm_dialect_conversion_stage_name in stage_names
+    ), f"Stage 'MLIRToLLVMDialect' not found in default pipeline stages"
+
+    llvm_dialect_conversion_stage_index = stage_names.index(llvm_dialect_conversion_stage_name)
+
+    _, pipeline = stages[llvm_dialect_conversion_stage_index]
+
+    assert len(pipeline) > 0, f"Pipeline for stage 'MLIRToLLVMDialect' is empty"
+
+    # Insert (in-place) the "convert-mbqc-to-llvm" pass immediately before the
+    # "convert-quantum-to-llvm" pass in the MLIRToLLVMDialect pipeline
+    insert_before_pass(
+        pipeline, ref_pass="convert-quantum-to-llvm", new_pass="convert-mbqc-to-llvm"
+    )
+
+    return stages
--- a/frontend/catalyst/pipelines.py
+++ b/frontend/catalyst/pipelines.py
@@ -313,3 +313,57 @@ def get_stages(options):
     stages["BufferizationPass"] = get_bufferization_stage(options)
     stages["MLIRToLLVMDialect"] = get_convert_to_llvm_stage(options)
     return list(stages.items())
+
+
+def insert_after_pass(pipeline: list[str], ref_pass: str, new_pass: str) -> None:
+    """Insert a pass into an existing pass pipeline at the position after the given reference pass.
+
+    Args:
+        pipeline (list[str]): An existing pass pipeline, given as a list of passes.
+        ref_pass (str): The name of the reference pass after which the new pass is inserted.
+        new_pass (str): The name of the pass to insert.
+
+    Raises:
+        ValueError: If `ref_pass` is not found in the pass pipeline.
+
+    Example:
+        >>> pipeline = ["pass1", "pass2"]
+        >>> insert_after_pass(pipeline, "pass1", "new_pass")
+        >>> pipeline
+        ['pass1', 'new_pass', 'pass2']
+    """
+    try:
+        ref_index = pipeline.index(ref_pass)
+    except ValueError as e:
+        raise ValueError(
+            f"Cannot insert pass '{new_pass}' into pipeline; reference pass '{ref_pass}' not found"
+        ) from e
+
+    pipeline.insert(ref_index + 1, new_pass)
+
+
+def insert_before_pass(pipeline: list[str], ref_pass: str, new_pass: str) -> None:
+    """Insert a pass into an existing pass pipeline at the position before the given reference pass.
+
+    Args:
+        pipeline (list[str]): An existing pass pipeline, given as a list of passes.
+        ref_pass (str): The name of the reference pass before which the new pass is inserted.
+        new_pass (str): The name of the pass to insert.
+
+    Raises:
+        ValueError: If `ref_pass` is not found in the pass pipeline.
+
+    Example:
+        >>> pipeline = ["pass1", "pass2"]
+        >>> insert_before_pass(pipeline, "pass1", "new_pass")
+        >>> pipeline
+        ['new_pass', 'pass1', 'pass2']
+    """
+    try:
+        ref_index = pipeline.index(ref_pass)
+    except ValueError as e:
+        raise ValueError(
+            f"Cannot insert pass '{new_pass}' into pipeline; reference pass '{ref_pass}' not found"
+        ) from e
+
+    pipeline.insert(ref_index, new_pass)
--- a/frontend/test/pytest/test_mbqc.py
+++ b/frontend/test/pytest/test_mbqc.py
@@ -25,33 +25,41 @@ import pennylane.ftqc as plft
 import pytest
 
 from catalyst import qjit
+from catalyst.passes.builtin_pipelines import mbqc_pipeline
+from catalyst.pipelines import CompileOptions
 from catalyst.utils.exceptions import CompileError
 
 pytestmark = pytest.mark.usefixtures("disable_capture")
 
-mbqc_pipeline = [
-    (
-        "default-pipeline",
-        [
-            "enforce-runtime-invariants-pipeline",
-            "hlo-lowering-pipeline",
-            "quantum-compilation-pipeline",
-            "bufferization-pipeline",
-        ],
-    ),
-    (
-        "mbqc-pipeline",
-        [
-            "convert-mbqc-to-llvm",
-        ],
-    ),
-    (
-        "llvm-dialect-lowering-pipeline",
-        [
-            "llvm-dialect-lowering-pipeline",
-        ],
-    ),
-]
+MBQC_PIPELINE = mbqc_pipeline()
+
+
+def test_mbqc_pipeline():
+    """Loosely check that the MBQC pipeline is the same as the default Catalyst pipeline, but with
+    the MBQC-to-LLVM dialect-conversion pass inserted before the Quantum-to-LLVM dialect-conversion
+    pass in the ``MLIRToLLVMDialect`` pipeline stage.
+    """
+    options = CompileOptions()
+    default_pipelines = options.get_pipelines()
+
+    # Check that the stage names are the same
+    default_stage_names = [item[0] for item in default_pipelines]
+    mbqc_stage_names = [item[0] for item in MBQC_PIPELINE]
+
+    assert default_stage_names == mbqc_stage_names
+    assert mbqc_stage_names[-1] == "MLIRToLLVMDialect"
+
+    # Check that the conversion pass(es) are in the mbqc pipeline
+    _, mbqc_llvm_conversion_pipeline = MBQC_PIPELINE[-1]
+
+    assert "convert-quantum-to-llvm" in mbqc_llvm_conversion_pipeline
+    assert "convert-mbqc-to-llvm" in mbqc_llvm_conversion_pipeline
+
+    # Check that the mbqc-to-llvm pass comes *before* the quantum-to-llvm pass
+    quantum_to_llvm_index = mbqc_llvm_conversion_pipeline.index("convert-quantum-to-llvm")
+    mbqc_to_llvm_index = mbqc_llvm_conversion_pipeline.index("convert-mbqc-to-llvm")
+
+    assert mbqc_to_llvm_index < quantum_to_llvm_index
 
 
 def test_measure_x():
@@ -68,7 +76,7 @@ def test_measure_x():
 
     qml.capture.enable()
 
-    @qjit(pipelines=mbqc_pipeline)
+    @qjit(pipelines=MBQC_PIPELINE)
     @qml.qnode(dev)
     def workload():
         _ = plft.measure_x(0)
@@ -94,7 +102,7 @@ def test_measure_y():
 
     qml.capture.enable()
 
-    @qjit(pipelines=mbqc_pipeline)
+    @qjit(pipelines=MBQC_PIPELINE)
     @qml.qnode(dev)
     def workload():
         _ = plft.measure_y(0)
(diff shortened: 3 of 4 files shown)


doc/releases/changelog-dev.md
```diff
@@ -219,6 +219,14 @@
 * The unused helper function `genArgMapFunction` in the `--lower-gradients` pass is removed.
   [(#1753)](https://github.com/PennyLaneAI/catalyst/pull/1753)
 
+* Catalyst now includes a new module `catalyst.passes.builtin_pipelines`, which contains built-in
+  compilation-pass pipelines for common workloads that extend the default Catalyst pass pipeline.
+  [(#1758)](https://github.com/PennyLaneAI/catalyst/pull/1758)
+
+  The built-in pass pipelines include:
```
paul0403, line 226: Hey, this is very nice! Definitely something we've been wanting for a while now :parrot: I can see this being used for PPR/PPM too, and in fact I do believe @sengthai needed to add his own pipelines when using them from python!

I will take a closer look tomorrow. Just a first thought: what about adding the oqd pipeline into this infrastructure too?

josh146:
Nice, thanks @joeycarter! This is something @isaacdevlugt I believe would have a stake in, as @paul0403 mentions it connects to how we want to expose the full PPM pipeline to users! So good to get a product review just on the UI/docs/making sure it's consistent with how we planned the built-in PPM pipeline UI.

joeycarter:
> Nice, thanks @joeycarter! This is something @isaacdevlugt I believe would have a stake in, as @paul0403 mentions it connects to how we want to expose the full PPM pipeline to users! So good to get a product review just on the UI/docs/making sure it's consistent with how we planned the built-in PPM pipeline UI.

Thanks @josh146! Yes absolutely, I'll wait for approval from all stakeholders before merging. :+1:

doc/releases/changelog-dev.md
sengthai: Thanks @paul0403! It is a good idea!

frontend/catalyst/passes/builtin_pipelines.py
dime10, line 1: Would it not make sense to store these new pipelines along-side the default pipeline (which is also a built-in pipeline)?

joeycarter: I'm happy to put it anywhere, I'm awaiting input from Product with a decision on how the UI should look (including the import location).

doc/releases/changelog-dev.md
```diff
@@ -215,6 +215,14 @@
 * The unused helper function `genArgMapFunction` in the `--lower-gradients` pass is removed.
   [(#1753)](https://github.com/PennyLaneAI/catalyst/pull/1753)
 
+* Catalyst now includes a new module `catalyst.passes.builtin_pipelines`, which contains built-in
```
isaacdevlugt, line 218:
```suggestion
* Catalyst now includes a new module called `catalyst.passes.builtin_pipelines`, which contains built-in
```

doc/releases/changelog-dev.md
```diff
@@ -215,6 +215,14 @@
...
+  compilation-pass pipelines for common workloads that extend the default Catalyst pass pipeline.
+  [(#1758)](https://github.com/PennyLaneAI/catalyst/pull/1758)
+
+  The built-in pass pipelines include:
+
+  - `mbqc_pipeline()`: for MBQC workloads.
```
isaacdevlugt, line 224: lolol is this the only bullet point, or are there more coming?

frontend/test/pytest/test_mbqc.py
```diff
@@ -25,33 +25,41 @@
 import pytest
 
 from catalyst import qjit
+from catalyst.passes.builtin_pipelines import mbqc_pipeline
+from catalyst.pipelines import CompileOptions
```
paul0403, line 29: Doesn't it make more sense to put everything pipeline-related in a `pipeline` module? e.g. 
```
from catalyst.pipelines import default_pipeline, mbqc_pipeline
```

This also eliminates the need to retrive default pipeline from `CompileOptions` (which feels a bit weird)

frontend/test/pytest/test_pipelines.py
```diff
@@ -0,0 +1,93 @@
...
+        """Test the ``insert_after_pass`` function with valid inputs, inserting into the middle of
+        the pipeline.
+        """
+        new_pass_1 = "new_pass_1"
+        insert_after_pass(pipeline, ref_pass="pass1", new_pass=new_pass_1)
+
+        pipeline_expected = ["pass1", new_pass_1, "pass2", "pass3"]
+        assert pipeline == pipeline_expected
+
+    def test_insert_after_pass_at_end(self, pipeline):
```
paul0403, line 44: This test seems to be duplicating the one above? Is there anything specific that you were afraid of when inserting to the end?

frontend/test/pytest/test_pipelines.py
```diff
@@ -0,0 +1,93 @@
...
+        """Test the ``insert_before_pass`` function with valid inputs, inserting into the middle of
+        the pipeline.
+        """
+        new_pass_1 = "new_pass_1"
+        insert_before_pass(pipeline, ref_pass="pass2", new_pass=new_pass_1)
+
+        pipeline_expected = ["pass1", new_pass_1, "pass2", "pass3"]
+        assert pipeline == pipeline_expected
+
+    def test_insert_before_pass_at_beginning(self, pipeline):
```
paul0403, line 73: Same here for inserting before beginning

frontend/catalyst/pipelines.py
```diff
@@ -315,3 +315,57 @@ def get_stages(options):
...
+        >>> pipeline
+        ['pass1', 'new_pass', 'pass2']
+    """
+    try:
+        ref_index = pipeline.index(ref_pass)
+    except ValueError as e:
+        raise ValueError(
+            f"Cannot insert pass '{new_pass}' into pipeline; reference pass '{ref_pass}' not found"
+        ) from e
+
```
paul0403, line 343: In general I think we'd like to avoid try-except unless there's no alternative. I forgot where I heard this rule but I definitely remember hearing it. 

So, why not just 
```python
if ref_pass in pipeline:
   # do your thing
else:
   # raise error
``` 
? It's also more readable, at least I would think so :sweat_smile:

frontend/catalyst/pipelines.py
```diff
@@ -315,3 +315,57 @@ def get_stages(options):
...
+        >>> insert_before_pass(pipeline, "pass1", "new_pass")
+        >>> pipeline
+        ['new_pass', 'pass1', 'pass2']
+    """
+    try:
+        ref_index = pipeline.index(ref_pass)
+    except ValueError as e:
+        raise ValueError(
+            f"Cannot insert pass '{new_pass}' into pipeline; reference pass '{ref_pass}' not found"
+        ) from e
```
paul0403, line 369: same here

frontend/test/pytest/test_mbqc.py
```diff
@@ -68,7 +76,7 @@ def test_measure_x():
 
     qml.capture.enable()
 
-    @qjit(pipelines=mbqc_pipeline)
+    @qjit(pipelines=MBQC_PIPELINE)
```
paul0403, line 79: Why not just name the variable `mbqc_pipeline`?

frontend/test/pytest/test_mbqc.py
```diff
@@ -25,33 +25,41 @@
...
+MBQC_PIPELINE = mbqc_pipeline()
+
+
+def test_mbqc_pipeline():
+    """Loosely check that the MBQC pipeline is the same as the default Catalyst pipeline, but with
+    the MBQC-to-LLVM dialect-conversion pass inserted before the Quantum-to-LLVM dialect-conversion
+    pass in the ``MLIRToLLVMDialect`` pipeline stage.
+    """
+    options = CompileOptions()
+    default_pipelines = options.get_pipelines()
```
paul0403, line 43: Like mentioned, having to go through `CompileOptions` for the default pipeline seems counterintuitive for users. Especially now we are exposing pipelines to users, why not just expose the default pipeline as well?

I'm thiking something alongside the following:
```
- catalyst/frontend/catalyst (root package location)
   - passes (things regarding pipeline should not go in here at all)
   - pipeline (new directory, aka new package)
       - __init__.py (expose your things here, e.g. mbqc pipeline, default pipeline, ppm pipeline, your insert/remove pass utils, etc)
       - pipeline_utils.py (insert/remove pass API)
       - builtin_pipelines.py (defines the builtin pipelines)
```

frontend/catalyst/passes/builtin_pipelines.py
```diff
@@ -0,0 +1,57 @@
...
+    The MBQC pipeline is identical to the default Catalyst pipeline, but with the MBQC-to-LLVM
+    dialect-conversion pass inserted immediately before the Quantum-to-LLVM dialect-conversion pass
+    in the ``MLIRToLLVMDialect`` pipeline stage.
+
+    Returns:
+        PipelineStages: The list of pipeline stages.
+    """
+    options = CompileOptions()  # Use all default compile options
+    stages = get_stages(options)  # Get list of default pipeline stages
+
```
paul0403, line 36: Same here, can be simplified if we export the default pipeline

paul0403 (commented):
Nice, thanks! I think the functionality is good, just some comments on package organization :100:

frontend/test/pytest/test_pipelines.py
paul0403: Because I would argue this insertion test is just testing python's `list.insert()` is safe to insert to the end.

frontend/catalyst/pipelines.py
joeycarter: This was mainly to rewrite the error message that comes from `list.index(x)` when `x` is not in the list:

```pycon
>>> a = list(range(10))
>>> a.index(11)
  File "<stdin>", line 1, in <module>
ValueError: 11 is not in list
```

Writing it as a try-except block is a small optimization, since both `x in list(...)` and `list.index()` operations require iterating through the list, while the try-except block with one call to `list.index()` requires only one iteration through it.

Here's some evidence :slightly_smiling_face: 

With a try-except block:

```console
$ py -m timeit -s 'a = list(range(1000))' 'el = 500' 'try:' '    i = a.index(el)' 'except ValueError as e:' '    pass'
200000 loops, best of 5: 1.77 usec per loop
```

With an `x in list` check:

```console
$ py -m timeit -s 'a = list(range(1000))' 'el = 500' 'if el in a:' '    i = a.index(el)' 'else:' '    pass'
100000 loops, best of 5: 3.51 usec per loop
```

The try-except block is about twice as fast (as expected :slightly_smiling_face:).

If we were being fast-and-loose with error messages I wouldn't wrap it in a try-except block at all and let it just throw the "not in list" `ValueError`, but this way it gives some more context explaining why the pass couldn't be inserted into the pipeline.

frontend/test/pytest/test_mbqc.py
joeycarter: Because I've written this variable as a module-level constant :slightly_smiling_face:

https://peps.python.org/pep-0008/#constants

frontend/test/pytest/test_pipelines.py
joeycarter: This is really just an effort to cover all corner cases that I can think of.

It's also following the "test the interface, not the implementation" philosophy, meaning the tests are agnostic to the implementation. Suppose we decide to change the data structure that contains the sequence of passes and we can no longer use `list.index()`. These tests should still pass regardless of the implementation. (Granted, we'd have to change the `pipeline` fixture, but the `insert_after_pass` interface should be unaffected).

frontend/test/pytest/test_pipelines.py
joeycarter: Answered here: https://github.com/PennyLaneAI/catalyst/pull/1758/files#r2121962768

frontend/test/pytest/test_mbqc.py
joeycarter: @dime10 suggested something similar.

@isaacdevlugt is this UI that @paul0403 is suggesting alright with you?

frontend/test/pytest/test_mbqc.py
joeycarter: I haven't really made an effort to expose the default pipeline to users; what I've done here is just to test the new mbqc pipeline against the default one. Indeed users could have always accessed the default pipeline in this way. If we do want to expose the default pipeline, we can organize things following the decision made in this thread: https://github.com/PennyLaneAI/catalyst/pull/1758#discussion_r2121966407

frontend/catalyst/pipelines.py
paul0403: Ahh, this makes sense. Thanks! I did not consider the performance nuance. 

> If we were being fast-and-loose with error messages I wouldn't wrap it in a try-except block at all and let it just throw the "not in list" ValueError, but this way it gives some more context explaining why the pass couldn't be inserted into the pipeline.

I agree, we should aim to give meaning messages instead of generic ones from python! (Debugg `KeyError: 1765480912` from jax a while back was certainly an eye opener for me)

frontend/test/pytest/test_mbqc.py
josh146: To me (as a user) it is not really clear what the difference is between a pipeline and a pass (1 pass to one person might be a pipeline of multiple passes to another person).

So if I was looking for an entire PPM/PPR pass, or a full MBQC pass, I would likely still go here: https://docs.pennylane.ai/projects/catalyst/en/stable/code/__init__.html#module-catalyst.passes

josh146: That is my argument for pipelines of passes _being_ passes from a user perspective 😅

josh146: Although I'm not quite convinced, since a pipeline is a dict and a pass is a decorator with options

doc/releases/changelog-dev.md
josh146: @joeycarter would you be able to include an example of how to use this pipeline?

frontend/test/pytest/test_mbqc.py
josh146: Oh wait, I just realized I am confusing the qjit `pipeline` argument with the `circuit_transform_pipeline`, which is equivalent to the [`catalyst.pipeline`](https://docs.pennylane.ai/projects/catalyst/en/stable/code/api/catalyst.pipeline.html) decorator 😱

doc/releases/changelog-dev.md
joeycarter: @josh146 Yes for sure. I'll add an example once we settle on the interface and import location (to avoid having the example out of sync with the final API). I'll leave this conversation unresolved as a reminder.

joeycarter:
> @joeycarter @isaacdevlugt I realized that I find `@qjit(pipelines=)` a bit confusing with `@qjit(circuit_transform_pipeline=)` which co-incides with `catalyst.passes` and provides the same functionality as `catalyst.pipelines` 😱
> 
> I think we need to better separate/denote/merge these concepts to make it clear to end-users, and figure out where things should live in the documentation

I agree that this is confusing! Even I forgot that the `pipelines` option of `qjit` is "considered to be used by advanced users for low-level debugging purposes." (quote from the [`qjit`](https://docs.pennylane.ai/projects/catalyst/en/stable/code/api/catalyst.qjit.html) docs). However, if we start providing full, built-in pipelines, they would currently fit more naturally into the `pipelines` option than the `circuit_transform_pipeline` option based on the expected types for each (although we can change this of course).

One of the troubles here is that the mbqc pipeline, as it is now, does not actually contain any _quantum_ circuit transforms, rather it just adds a pass to the default pipeline to perform the MBQC-to-LLVM dialect conversion, which makes it unsuitable for the `circuit_transform_pipeline` option (the order that passes get inserted also matters here). Normally we would add all the dialect-conversion passes to the default pipeline, but we were trying to keep the mbqc work separate while it's still under development, hence the new pipeline added in this PR. And just to add, if we do eventually add the MBQC-to-LLVM dialect-conversion pass to the default pipeline, I think it still makes sense to define a complete, built-in MBQC/FTQC pipeline, since I expect it to become much more involved as it develops, and specifying all the passes with the `circuit_transform_pipeline` option will become unwieldy.

Maybe for now I would propose keeping the `circuit_transform_pipeline` option as is, and update the documentation for the `pipelines` option to say that it is intended for use with the built-in pipelines that Catalyst provides, and that it is also suitable for advanced users for debugging purposes.

frontend/catalyst/pipelines.py
dime10: > In general I think we'd like to avoid try-except unless there's no alternative. I forgot where I heard this rule but I definitely remember hearing it.

I think this depends. The general advice I remember about exceptions is that they're very fast in the negative case (when no problem arises), but slow in the positive case. So if you expect to frequently hit the exception, a conditional can be better, but if you think you will rarely hit it, an exception can be better.

frontend/test/pytest/test_mbqc.py
dime10: Yeah admittedly passes and pipelines are a bit overloaded. Here we are talking about the entire compilation recipe (maybe we should name it that?), while we otherwise also refer to a group of passes as a pass pipeline, but don't really have a different name for a group of pipelines.

But besides the UI question, just from a code organization point of view I would expect the default pipeline to be in the same place as other pre-defined pipelines. (pipeline = whole compilation recipe)

frontend/catalyst/pipelines.py
mlxd: Exactly as @dime10 said: we try not to use try-except for conditional logic for this exact reason.

PR #1763, opened 2025-05-23
Add more OQD tests

mehrdad2m:
**Context:**
OQD openAPL tests are limited to a single RX gate

- [x] add proper cleanup fixture

Expland the test to cover:
- [x] Target gate set {RX, RY, MS(not added yet)}
- [x] Gates for which there is a decomposition to the target gate set {CNOT, RZ, PhaseShift, Hadamard, S, T, PauliZ}
- [x] More complex circuits (RX_RY, 2-qubit QFT)

Modified OQD toml file to ensure more complicated circuits like the QFT are only decomposed to the {CNOT, RZ, PhaseShift, Hadamard, S, T, PauliZ} which are decomposable to the target gate set.

To avoid a situation where different tests compete over the same openAPL output file, we pass the name of the opanAPL file as kwarg to the OQD device which is used in runtime to make the json file.

**Benefits:**

**Possible Drawbacks:**
Huge number of lines need to be added in order to compare the whole openAPL json files. Alternatively, we only check for more high level infos (e.g. number of Pulses, specific parameters, etc.) instead of adding the huge number of lines for openAPL programs

**Related GitHub Issues:**
[sc-88299]


paul0403:
Didn't they add the transition labels? :sweat_smile: So they added a label to each transition but is not using them as keys?

How much work do you think is needed to use the transition/level labels upstream, in order to reduce the json size?

paul0403:
We discovered that the mysterious CI failure is because of multiple pytests all looking for the `__openapl.json` dumped to the OS

mehrdad2m:
To avoid a situation where different tests compete over the same openAPL outfile file, we pass the name of the opanAPL file as kwarg to the OQD device which is used in runtime to make the json file.

frontend/test/test_oqd/oqd/test_openapl_generation.py
```diff
@@ -24,65 +24,335 @@
...
-    def test_RX_gate(self):
+
+def ordered(obj):
+    """Sort the JSON file."""
+    if isinstance(obj, dict):
+        return sorted((k, ordered(v)) for k, v in obj.items())
+    if isinstance(obj, list):
+        return sorted(ordered(x) for x in obj)
+    else:
+        return obj
```
paul0403: Does `json.load` not guarantee a return type? And are there potential dangers if it returns something other than dict and list?

frontend/test/test_oqd/oqd/test_openapl_generation.py
```diff
@@ -24,65 +24,335 @@
...
-        with open(self.output_f, "r", encoding="utf-8") as f:
-            catalyst_json = json.load(f)
+        stats = profile_openapl(oqd_dev.openapl_file)
+        assert stats["num_ions"] == 1
+        assert stats["num_parallel_protocols"] == 1
+        assert stats["num_beams"] == 2
+        assert stats["num_transitions"] == 4
+        assert stats["num_levels"] == 4
+        expected_f = os.path.join(MODULE_TEST_PATH, "test_single_RX.json")
+        assert verify_json(expected_f, oqd_dev.openapl_file)
```
paul0403: Aren't the asserts covered by the `verify_json` already?

paul0403: Or, I guess my question is why do we need the profiler at all, if we have the `verify_json` checking the entire openapl.

runtime/lib/backend/oqd/OQDDevice.hpp
```diff
@@ -77,8 +78,10 @@ class OQDDevice final : public Catalyst::Runtime::QuantumDevice {
         device_shots = device_kwargs.contains("shots")
                            ? static_cast<size_t>(std::stoll(device_kwargs["shots"]))
                            : 0;
+        openapl_file = device_kwargs.contains("openapl_file") ? device_kwargs["openapl_file"]
```
paul0403: Probably a better name for the kwarg is `openapl_file_name`, since just `open_apl` could be misunderstood as the file object.

paul0403 (commented):
Nice, glad that we have the CI mystery resolved!

runtime/lib/backend/oqd/OQDDevice.hpp
```diff
@@ -31,6 +31,7 @@ class OQDDevice final : public Catalyst::Runtime::QuantumDevice {
 
     size_t device_shots;
     std::string ion_specs;
+    std::string openapl_file;
```
paul0403: Since now this involves more than just changing tests, don't forget to add a changelog entry!

I think this would just be under internal changes since technically OQD device is not released and is only available if someone builds from source.

frontend/test/test_oqd/oqd/test_openapl_generation.py
mehrdad2m: True, we don't need to profile the two tests for which we check the entire json

frontend/test/test_oqd/oqd/test_openapl_generation.py
mehrdad2m: Actually, I think we don't need this anymore. I originally added this function to make sure the order is preserved (I suspected that it might be the source of issues before). I will remove this.

mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/1763/commits/a45d2719dbc632446ddd96247c04007e07a7a260

runtime/lib/backend/oqd/OQDDevice.hpp
mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/1763/commits/9c1e569f92443466131c783f2421d3c75d39fb2a

runtime/lib/backend/oqd/OQDDevice.hpp
mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/1763/commits/e89fe623d4f9ed4f12b5ce9a6760181e92f4cbf4

frontend/test/test_oqd/oqd/test_openapl_generation.py
```diff
@@ -24,65 +24,313 @@
...
+        assert stats["num_levels"] == 4
+
+
+class TestDecomposableGates:
+    """Test OQD device OpenAPL generation for gates decomposable into target gates."""
+
+    def test_CNOT_gate(self, tmp_openapl_file_name):
         """Test OpenAPL generation for a circuit with a single CNOT circuit."""
-        oqd_dev = OQDDevice(backend="default", shots=4, wires=2)
+        print("tmp_openapl_file_name: ", tmp_openapl_file_name)
```
paul0403:
```suggestion
```

paul0403 (approved):

PR #1766, opened 2025-05-26
Add support for `qml.measure` in Catalyst's PlxPR pipeline

dime10:
The [`qml.measure()`](https://docs.pennylane.ai/en/stable/code/api/pennylane.measure.html) operation for mid-circuit measurements can now be used in QJIT-compiled circuits with program capture enabled. For example:

```python
import pennylane as qml
from catalyst import qjit

device = qml.device("lightning.qubit", wires=1)

qml.capture.enable()

@qjit
@qml.qnode(device)
def circuit():
    qml.X(wires=0)
    qml.measure(wires=0, reset=reset)
    return qml.expval(qml.Z(0))
```

```pycon
>>> circuit()
-1.0
```

Note that using `qml.measure()` in this way binds the operation to [`catalyst.measure()`](https://docs.pennylane.ai/projects/catalyst/en/stable/code/api/catalyst.measure.html), which may behave differently than `qml.measure()` in a native PennyLane circuit, as described in the [Functionality differences from PennyLane](https://docs.pennylane.ai/projects/catalyst/en/stable/dev/sharp_bits.html#functionality-differences-from-pennylane) section of the sharp bits and debugging tips guide. In regular QJIT-compiled workloads (_without_ program capture enabled), you must continue to use `catalyst.measure()`.

### DIFF
--- a/frontend/test/pytest/test_capture_integration.py
+++ b/frontend/test/pytest/test_capture_integration.py
@@ -349,6 +349,45 @@ class TestCapture:
 
         assert jnp.allclose(capture_result, circuit(theta))
 
+    @pytest.mark.parametrize("reset", [False, True])
+    @pytest.mark.parametrize("op", [qml.I, qml.X])
+    def test_measure(self, backend, reset, op):
+        """Test the integration for a circuit with a mid-circuit measurement.
+
+        We do not currently have full feature parity between PennyLane's `measure` and Catalyst's.
+        Moreover, there is limited support for the various MCM methods with QJIT and with program
+        capture enabled. Hence, we only test that a simple example with a deterministic outcome
+        returns correct results.
+        """
+        device = qml.device(backend, wires=1)
+
+        # Capture enabled
+
+        qml.capture.enable()
+
+        @qjit
+        @qml.qnode(device)
+        def captured_circuit():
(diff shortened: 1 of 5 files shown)


frontend/test/pytest/test_capture_integration.py
```diff
@@ -349,6 +349,71 @@ def circuit(theta):
...
+
+        qml.capture.disable()
+
+        if reset:
+            expected_result = 1
+        else:
+            if op is qml.I:
+                expected_result = 1
+            elif op is qml.X:
+                expected_result = -1
```
dime10, line 385: What I normally see is that the different inputs are directly paired with their expected values in the parametrize call, but I guess this works too 👍 

```
@pytest.mark.parametrize("op,expected", [(qml.I, 1), (qml.X, -1)])
```

dime10 (commented):
Thanks Joey, approved :)

frontend/test/pytest/test_capture_integration.py
joeycarter: When I initially wrote this I was thinking that the expected result would depend non-trivially on the inputs, but turns out that wasn't the case in the end! Refactored in f974a8e1

PR #1772, opened 2025-05-29
Update tutorial to reflect the removal of separate match and rewrite

mehrdad2m:
**Context:**
LLVM has removed separate `match` and `rewrite` functions. we need to update our [Catalyst Compiler Passes](https://docs.pennylane.ai/projects/catalyst/en/stable/dev/transforms.html#writing-transformations-on-catalyst-s-ir) tutorial  

updated tutorial

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


doc/dev/transforms.rst
```diff
@@ -286,11 +285,11 @@ the second is a list of qubits):
 
     QubitUnitary(*, QubitUnitary(*, *))
 
-Let's implement it in C++:
+Let's add the pattern matching logic to the ``matchAndRewrite`` method:
```
joeycarter, line 288:
```suggestion
Let's add the pattern-matching logic to the ``matchAndRewrite`` method:
```

doc/dev/transforms.rst
```diff
@@ -711,18 +731,20 @@ Alright, our function should now look something like this:
         func.return %dx, %dy, %dz : f64, f64, f64
     }
 
-Finally, we have to amend our rewrite function to invoke the new function we created and delete the
+Finally, we have to amend our matchAndRewrite function to invoke the new function we created and delete the
```
joeycarter, line 734:
```suggestion
Finally, we have to amend our ``matchAndRewrite`` function to invoke the new function we created and delete the
```

doc/dev/transforms.rst
```diff
@@ -711,18 +731,20 @@ Alright, our function should now look something like this:
...
 ``GradOp`` from the IR:
 
 .. code-block:: cpp
 
-    void FiniteDiffLowering::rewrite(GradOp op, PatternRewriter &rewriter)
+    LogicalResult FiniteDiffLowering::matchAndRewrite(GradOp op, PatternRewriter &rewriter)
     {
         ...
             populateFiniteDiffMethod(rewriter, op, gradFn);
         }
```
joeycarter, line 743: Is this right? I'm a little confused reading this, since it opens with `...` and closes with `}` :smile:

joeycarter (commented):
Just a few editorial comments :slightly_smiling_face:

doc/dev/transforms.rst
mehrdad2m: the `}` marks the end of the scope [here](https://github.com/PennyLaneAI/catalyst/blob/1d603a35bb3169c90212256bbace9ad88a5bffe9/doc/dev/transforms.rst?plain=1#L609-L629). I think the beginning of the scope is not shown since there is more code before it.

PR #1776, opened 2025-05-30
Fix badges in README

joeycarter:
The links for the CodeCov and RTD badges in the README weren't correct and were attempting to retrieve statuses that didn't exist. This PR fixes that.


AntonNI8 (approved):
Thanks @joeycarter!

PR #1782, opened 2025-06-03
Tree traversal transform in Catalyst

dime10:
WIP

### DIFF
--- a/frontend/catalyst/passes/if_statement_split.py
+++ b/frontend/catalyst/passes/if_statement_split.py
@@ -89,12 +89,12 @@ class IfOperatorPartitioningPass(RewritePattern):
         if not flat_if:
             return
 
-        print(f"FDX: Should flat: {flat_if}")
+        # print(f"FDX: Should flat: {flat_if}")
 
         # Split IfOps into only true branches
         self.split_nested_if_ops(op, rewriter)
 
-        print_mlir(op, "After splitting IfOps:")
+        # print_mlir(op, "After splitting IfOps:")
 
         # Flatten nested IfOps
         self.flatten_nested_IfOps(op, rewriter)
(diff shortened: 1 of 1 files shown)


dime10:
> You're right! It should also store for right branch after simulation.
>
> right branch:
> restore_state from depth - 1
> simulation at depth 
> store_state for depth

Storing after the simulation doesn't make much sense to me, after all you're trying to save the state before modifying it, so you can go back to it.

For instance, going `right -> M -> right` in your proposal would store and then restore the state right away again in the middle node.

I think the only time you need to restore is if you are at a node that you have visited before, and are now taking a different path. This can only happen if you've gone left at the node and (later) are now going right.

That's why I had this separate flag to tell me if we needed restoring:
```py
while ...:
    match node_status:
        case 0:
            visited_stack[current_depth] = 1
            # store here
            print(f" - storing state")

            print(" - going left")
            current_depth += 1
        case 1:
            visited_stack[current_depth] = 2
            if needs_restore:
                print(f" - restoring state")
                needs_restore = False

            print(" - going right")
            current_depth += 1
        case 2:
            visited_stack[current_depth] = 0
            print(" - going up")
            current_depth -= 1

    if current_depth == tree_depth:
        print(" < hit a leaf, going up >")
        needs_restore = True
```

dime10:
> Originally I was thinking that each simulation starts with an MCM

I think that's a good idea (excluding the first node/segment of course).

> so we need to ensure they all have a copy of the state before next simulation, therefore I store a copy after each simulation (because the first node has no MCM, storing before it makes no sense). 

I suppose storing after each simulation segment makes sense in order to skip the first one. What I was confused by is that restoring would happen before the segment, but I didn't realize you were planning to restore from the _previous node_. This should be correct then.

(side note: I just realized that even storing the state with each segment is not strictly necessary, because technically we only need to store if we _know_ that we will restore.)

> And what you're saying should be that we store a copy of state at before each depth (possibly excluding the first one), and because it's beforehand, we should be able to avoid storing an extra copy of unused data at the last node?

I guess before or after makes no difference if it's handled correctly, my mistake. But the restore only needs to happen once after hitting a leaf, doing it on every segment is unnecessary work right?

> EDIT: And for the left branch if mcm prob is 0, then it wouldn't reach the leaf, so need_restore = False and just go back and run the right branch without restore. And for this case, the store after simulation would have redundant date stored.

That sounds right.

PR #1789, opened 2025-06-05
Test stable/stable without building catalyst

dime10:
Bypasses the need to download caches from GitHub.

### DIFF
--- a/.github/workflows/check-pl-compat.yaml
+++ b/.github/workflows/check-pl-compat.yaml
@@ -42,9 +42,12 @@ jobs:
         git checkout v0.11.0-rc
 
     - name: Install deps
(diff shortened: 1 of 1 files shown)


.github/workflows/check-pl-compat.yaml
```diff
@@ -47,55 +47,71 @@ jobs:
         python-version: ${{ needs.constants.outputs.primary_python_version }}
 
     - name: Install deps
+      if: ${{ inputs.catalyst != 'stable' }}
       run: |
         sudo apt-get update
         sudo apt-get install -y cmake ninja-build ccache libomp-dev libasan6 lld
+
+    - name: Install deps (Python)
```
dime10, line 55:
```suggestion
    - name: Install deps (Python)
      run: |
```

PR #1790, opened 2025-06-06
Fix ubuntu24 compatability

runora95:
**Context:**
The workflow `CPL stable/stable/stable / Build Configuration` is failing after the ubuntu-24.04 upgrade. The default version of python in ubuntu24 is 3.12, the [workflow fails](https://github.com/PennyLaneAI/catalyst/actions/runs/15464703564/job/43534000796) when it tries to grep python 3.10 because it doesn't exist/ 
Add `actions/setup-python@v5` to check pennylane compatability.
**Benefits:**
Python3.10 gets installed and the pipeline uses the correct python version.
**Possible Drawbacks:**
None
**Related GitHub Issues:**

### DIFF
--- a/.github/workflows/CPL_stable-stable-stable.yaml
+++ b/.github/workflows/CPL_stable-stable-stable.yaml
@@ -1,8 +1,8 @@
 name: Check CPL stable/stable/stable
 
 on:
-  # schedule:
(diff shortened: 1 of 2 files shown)


dime10 (approved):
Ah perfect, I was wondering why the workflow was failing in my other PR :)

PR #1795, opened 2025-06-09
Fix a failing test on latest/latest due to a new PL gate metadata

paul0403:
**Context:**
PL added a new metadata to controlled gates https://github.com/PennyLaneAI/pennylane/pull/7612
We thus need to update our tests that check for the number of metadata.

**Benefits:**
latest/latest does not fail


andrijapau (approved):
Thanks! 🥳

paul0403:
(No changelog needed for just updating tests)

PR #1796, opened 2025-06-09
Update `.dep-versions`

mudit2812:
Erick suggested we update `.dep-versions` to use the latest dev release of PennyLane after https://github.com/PennyLaneAI/pennylane/pull/7613 was merged. Since he's currently away, I'm opening the PR.

### DIFF
--- a/.dep-versions
+++ b/.dep-versions
@@ -10,7 +10,7 @@ enzyme=v0.0.180
 
 # For a custom PL version, update the package version here and at
 # 'doc/requirements.txt
-pennylane=0.42.0-dev33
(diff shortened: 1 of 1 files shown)


mlir/Enzyme
mudit2812, line 1: I'm not sure what has happened here

paul0403:
Note that I'm updating this in #1795

mlir/Enzyme
paul0403: Your local environment is on an old llvm/mhlo/enzyme. 

We can update in #1795 to avoid complications for you on this : )

PR #1799, opened 2025-06-10
Add flaky pytest infrastructure and weekend runners 

paul0403:
**Context:**
We wish to reduce flaky tests in CI.
There is a python package `flaky` (which we already have in our `requirements.txt`) that reruns failed tests to hunt for flakiness. 
Let's actually use it.

- Add `flaky` infrastructure. A pytest run that hunts for flaky tests can be launched by `make pytest ENABLE_FLAKY=ON`
- 4 tests that change some global state cannot be run more than once. We force them to be run only once. None of them are of a stochastic nature
- Add a script to launch flaky hunting runs on weekends. 

**Benefits:**
If any new tests are flaky, this weekend runner will catch them!

[sc-87028]


.github/workflows/check-weekly-flaky.yaml
```diff
@@ -0,0 +1,32 @@
+name: Run Catalyst pytests for flaky test checking
+
+on:
+  pull_request: # temporarily set this on PR to check this script
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - ready_for_review
```
paul0403: Remove before merging!

.github/workflows/check-weekly-flaky.yaml
```diff
@@ -0,0 +1,36 @@
...
+        python3 -m pip install -r requirements.txt
+        python3 -m pip install cuda-quantum==0.6.0
+        python3 -m pip install oqc-qcaas-client
+        make frontend
+        pip uninstall pennylane-catalyst
+        pip install -i https://test.pypi.org/simple/ PennyLane-Catalyst
+
+    - name: Run Python Pytest Tests with flaky Checks
+      run: |
+        make pytest ENABLE_FLAKY=ON
```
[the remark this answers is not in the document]
paul0403: done

paul0403:
Since this action installs from testpypi, this PR's CI will fail tests for tests depending on new features still not on testpypi yet. 

This is fine. I only enable the script on PRs to check the script itself.

frontend/test/pytest/device/test_decomposition.py
```diff
@@ -113,6 +112,9 @@ def execute(self, circuits, execution_config):
         return circuits, execution_config
 
 
+NoUnitaryDevice.capabilities.operations.pop("QubitUnitary")
```
paul0403, line 115: @dime10 So for these test-only devices, the issue was because the popping from the capabilities dictionary were happening per-instance, instead of per-class. This breaks because the capabilities are set only once **on the class** from the base PL device class's `__init__subclass__` https://github.com/PennyLaneAI/pennylane/blob/master/pennylane/devices/device_api.py#L148

popping just once during class initialization (as opposed to instance init!) fixes this.

frontend/test/pytest/test_autograph.py
```diff
@@ -73,6 +73,13 @@ def val(self):
         return self.ref
 
 
+@pytest.fixture(autouse=True)
+def reset_Failing():
+    save = Failing.triggered.copy()
+    yield
+    Failing.triggered = save
```
paul0403: This one is more straightforward, this `Failing` class uses a class variable to keep track of labels it already saw, and this track record of labels is not reset over different tests, or different runs of the same test. 

I just add a fixture to reset it everytime.

frontend/test/pytest/test_autograph.py
```diff
@@ -1426,11 +1459,9 @@ def f1():
 
         assert f1() == sum([1, 1, 2, 2])
 
-    @pytest.mark.xfail(reason="this won't run warning-free until we fix the resource warning issue")
```
paul0403, line 1429: I found out that while loops' fallback just do not raise the warning whatsoever, the warning is only on for loops :thinking: :sweat_smile: https://github.com/PennyLaneAI/catalyst/blob/5fc3f6ceb6a9b4f47c6fc76aa1d7ac9d12c6e0ad/frontend/catalyst/autograph/ag_primitives.py#L347C29-L347C55

So I guess this test is checking for that... (which also means it does not need to patch the "no fallback warning" setting)

I don't know why this decision was made, but doesn't it make more sense to warning with while loops too? :sweat_smile:

dime10: I guess the point of the warning in the for loop was to alert users to the fact that the `i` variable is now dynamic, and they can no longer index lists with it for example. I don't know if the same applies to while, since there is no iteration variable (at least none originating from the loop op itself) 🤔

paul0403: I see. This does make sense I think.

In which case, I think we can keep this test and remove the xfail?

dime10: I think the test can be deleted, it would make sense as a for loop test, and we do in fact have it (it's called `test_ignore_warnings`). But without the silencing flag being required, I'm not sure what we would be testing.

frontend/test/pytest/test_autograph.py
```diff
@@ -55,7 +55,33 @@
...
 class Failing:
-    """Test class that emulates failures in user-code"""
+    """
+    Test class that emulates failures in user-code.
+
+    When autograph fails to convert, it will fall back to native python control flow execution.
+    In such cases autograph would raise a warning.
+
+    We want to test that this warning is properly raised, but if we use an actual error to trigger
+    the autograph fallback, the pytest would fail. Therefore, we manually raise an exception.
```
dime10:
```suggestion
    the autograph fallback, the fallback would also fail. Therefore, we raise an exception in a controlled fashion.
```

frontend/test/pytest/test_autograph.py
```diff
@@ -55,7 +55,33 @@
...
+            while acc < 5:
+                acc = Failing(acc, "while").val + 1
+            return acc
+
+    When tracing, the first Failing instance will encounter an unseen label "while".
+    This raises an exception and fails autograph, causing it to fallback.
+    Then in the actual python while loop, future Failing instances will encounter the same "while"
+    label. Since the label is not new, no new exception is raised, and the test finishes without
+    errors.
+    """
```
dime10, line 85: Thanks for adding the explanation here Paul! :) 

I thing I don't quite understand, when I looked at the similar test for the for loop (that tests the fallback warning), it didn't actually need this class. It was just setting up a scenario we know will trigger the fallback but pass otherwise.

```
    @pytest.mark.filterwarnings("error::UserWarning")
    def test_ignore_warnings(self, monkeypatch):
        """Test the AutoGraph config flag properly silences warnings."""
        monkeypatch.setattr("catalyst.autograph_ignore_fallbacks", True)

        @qjit(autograph=True)
        def f():
            acc = 0
            data = [0, 4, 5]
            for i in range(3):
                acc = acc + data[i]

```

Are we sure we even need this class? 😅

dime10 (approved):
Thanks @paul0403 !

frontend/test/pytest/test_autograph.py
paul0403: I think these are actually two separate tests? 

`test_ignore_warnings`: tests the `autograph_ignore_fallback` flag itself
```python
    @pytest.mark.filterwarnings("error::UserWarning")
    def test_ignore_warnings(self, monkeypatch):
        """Test the AutoGraph config flag properly silences warnings."""
        monkeypatch.setattr("catalyst.autograph_ignore_fallbacks", True)
```

`test_whileloop_no_warning`: tests that while loop naturally do not warn when fallback happens, aka no need to set flag, just do a fallback-ed while loop and see no warning is raised

So I actually think they have different purposes and are nice to have.

frontend/test/pytest/test_autograph.py
paul0403: Maybe this was needed for while loops? Because for those it's hard to "force" an error in order to test the fallback. Because for for loops forcing an error is easy, just use the loop variable as an index (so the `for i in ..., data[i]`), but for while it would be a bit more difficult? :thinking:

frontend/test/pytest/test_autograph.py
dime10: I don't think that interpretation aligns with the original design of the test, why would it be setting the flag then? I think it might just have been a careless copy of the for loop test. Also, why would we test for the absence of something if there was no expectation of that thing happening in the first place? It's not like we test that all our functions never raise a warning 🤔

frontend/test/pytest/test_autograph.py
dime10: Ah good point, maybe it is just for while loops then 🤔

frontend/test/pytest/test_autograph.py
paul0403: > I think it might just have been a careless copy of the for loop test.

True, I suspect this is what happened :sweat_smile: This explains the weirdness of this test, i.e. having the setting on when we don't need it.


> Also, why would we test for the absence of something if there was no expectation of that thing happening in the first place? It's not like we test that all our functions never raise a warning 🤔

I have been convinced. I will remove the test.

paul0403: https://github.com/PennyLaneAI/catalyst/pull/1799/commits/d26d0435b554aef9dd5084d1c946f96f49b47f95

PR #1801, opened 2025-06-11
Update from_plxpr to adjust to modified qnode prim

albi3ro:
**Context:**

In pennylane PR: https://github.com/PennyLaneAI/pennylane/pull/7050

I updated `qnode_prim` to get rid of `n_consts` as metadata. Instead, all constants are promoted to arguments. 

Updates use of `qnode_prim` for the new signature.

**Benefits:**

Will work with latest pennylane.

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/frontend/catalyst/from_plxpr.py
+++ b/frontend/catalyst/from_plxpr.py
@@ -178,17 +178,15 @@ class WorkflowInterpreter(PlxprInterpreter):
 # pylint: disable=unused-argument, too-many-arguments
 @WorkflowInterpreter.register_primitive(qnode_prim)
 def handle_qnode(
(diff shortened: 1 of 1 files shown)


albi3ro:
So somehow this change doubled the number of `CNOT`'s present when merge amplitude embedding is applied, and I am so completely confused.

joeycarter:
I'm a little late to the party here, but are there any thoughts on reverting https://github.com/PennyLaneAI/pennylane/pull/7050 until we figure out a solution that works for both PennyLane and Catalyst? latest/latest/latest is failing and this is blocking development work on xDSL passes.

paul0403:
This can be closed now that it's reverted right? @albi3ro @dime10

albi3ro:
Reverted due to this promoting things to tracers.

PR #1803, opened 2025-06-13
Move json and toml cmake `fetchcontent`s to root mlir directory

paul0403:
**Context:**
Multiple dialects are now using the `json` cpp library.  

We just `fetchcontent` once from the mlir root directory, instead of downloading them once per dialect.

**Benefits:**
Faster build.


dime10 (approved):

paul0403:
Let's merge this after #1794.

PR #1807, opened 2025-06-13
Add from_plxpr conversion for new dynamic qubit allocation and deallocation 

paul0403:
**Context:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-93262]


runtime/lib/capi/RuntimeCAPI.cpp
```diff
@@ -389,9 +389,13 @@ void __catalyst__rt__qubit_release(QUBIT *qubit)
...
 static int __catalyst__rt__qubit_release_array__impl(QirArray *qubit_array)
 {
-    getQuantumDevicePtr()->ReleaseAllQubits();
+    //getQuantumDevicePtr()->ReleaseAllQubits();
     std::vector<QubitIdType> *qubit_array_ptr =
         reinterpret_cast<std::vector<QubitIdType> *>(qubit_array);
+    for (QubitIdType qubit : *qubit_array_ptr) {
+        std::cout << "releasing..." << "\n";
+        getQuantumDevicePtr()->ReleaseQubit(qubit);
+    }
```
paul0403: Don't know why, but the individual `ReleaseQubit` does not seem to be working.

paul0403: Maybe it's just because the `QubitManager` never expected multiple allocations. I will check this further.

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -294,15 +292,43 @@ class PLxPRToQuantumJaxprInterpreter(PlxprInterpreter):
...
     def interpret_operation(self, op):
         """Re-bind a pennylane operation as a catalyst instruction."""
 
-        in_qubits = [self.qreg_manager[w] for w in op.wires]
+        in_qubits_and_qregs = [self._get_wire_value_and_qreg(w) for w in op.wires]
+        in_qubits = []
+        in_qregs = []
+        for in_qubit, in_qreg in in_qubits_and_qregs:
+            in_qubits.append(in_qubit)
+            in_qregs.append(in_qreg)
```
```suggestion
       in_qubits, in_qregs = zip(*in_qubits_and_qregs)
```

runtime/lib/capi/RuntimeCAPI.cpp
dime10: In what sense are they not working?

dime10: We might not get around revisiting the AllocateQubits / ReleaseAllQubits asymmetry this time around. Presumably, the device is going to need a need batch release function for qubits that doesn't remove everything? 🤔 
Doing it one-by-one could be quite inefficient.

paul0403: > In what sense are they not working?

After a `ReleaseQubit(index)` call, the `__catalyst_get_num_qubits()` result does not change. This leads to return shapes of measurements to be messed up.

Also this might just be a syndrome, the underlying thing could be more complicated.

dime10: Ah shoot 🤔 Could this be an optimization thing? I believe in lightning we decided not to reduce the size of the statevector for performance reasons (or at least at some point thought about doing this), but the number of qubits should still have gone down in the qubit manager.

paul0403: Another syndrome is that mcm pytests all crash (~~probably segfaults~~ with linux `Killed`). I'm looking into it now.

paul0403:
I'm closing this since we decided on a new interpretation (i.e. no new allocations, just resolve extract indices at runtime)

PR #1810, opened 2025-06-16
Support Static and Dynamic Variables in PLxPR Programs with QJIT

sengthai:
**Context:**
Control over static variables in compilation is a general necessity. At the moment, with `qml.capture.enable()`, additionally specifying a static variable in `qjit` does not work. With this epic, we want to add support for static variables in the plxpr-qjit execution pipeline.

- Updated the `trace_from_pennylane` function signature to include `dynamic_argnums`.
- Modified the function implementation to correctly handle dynamic arguments when calling `from_plxpr`.

 
[[sc-93318]]

### DIFF
--- a/frontend/catalyst/from_plxpr.py
+++ b/frontend/catalyst/from_plxpr.py
@@ -45,6 +45,7 @@ from catalyst.device import (
     extract_backend_info,
     get_device_capabilities,
 )
+from catalyst.tracing.type_signatures import filter_static_args
 from catalyst.jax_extras import jaxpr_pad_consts, make_jaxpr2, transient_jax_config
 from catalyst.jax_primitives import (
     AbstractQbit,
@@ -722,7 +723,9 @@ def handle_measure_in_basis(self, angle, wire, plane, reset, postselect):
 
 
 # pylint: disable=too-many-positional-arguments
-def trace_from_pennylane(fn, static_argnums, abstracted_axes, sig, kwargs, debug_info=None):
+def trace_from_pennylane(
+    fn, static_argnums, dynamic_argnums, abstracted_axes, sig, kwargs, debug_info=None
+):
     """Capture the JAX program representation (JAXPR) of the wrapped function, using
     PL capure module.
 
@@ -746,6 +749,6 @@ def trace_from_pennylane(fn, static_argnums, abstracted_axes, sig, kwargs, debug
         args = sig
 
         plxpr, out_type, out_treedef = make_jaxpr2(fn, **make_jaxpr_kwargs)(*args, **kwargs)
-        jaxpr = from_plxpr(plxpr)(*args, **kwargs)
+        jaxpr = from_plxpr(plxpr)(*dynamic_argnums, **kwargs)
 
     return jaxpr, out_type, out_treedef, sig
--- a/frontend/catalyst/from_plxpr.py
+++ b/frontend/catalyst/from_plxpr.py
@@ -45,7 +45,6 @@ from catalyst.device import (
     extract_backend_info,
     get_device_capabilities,
 )
-from catalyst.tracing.type_signatures import filter_static_args
 from catalyst.jax_extras import jaxpr_pad_consts, make_jaxpr2, transient_jax_config
 from catalyst.jax_primitives import (
     AbstractQbit,
--- a/frontend/catalyst/from_plxpr/from_plxpr.py
+++ b/frontend/catalyst/from_plxpr/from_plxpr.py
@@ -499,10 +499,27 @@ def trace_from_pennylane(
     PL capure module.
 
     Args:
-        args (Iterable): arguments to use for program capture
+        fn(Callable): the user function to be traced
+        static_argnums(int or Seqence[Int]): an index or a sequence of indices that specifies the
+            positions of static arguments.
+        dynamic_argnums(int or Seqence[Int]): an index or a sequence of indices that specifies the
+            positions of dynamic arguments.
+        abstracted_axes (Sequence[Sequence[str]] or Dict[int, str] or Sequence[Dict[int, str]]):
+            An experimental option to specify dynamic tensor shapes.
+            This option affects the compilation of the annotated function.
+            Function arguments with ``abstracted_axes`` specified will be compiled to ranked tensors
+            with dynamic shapes. For more details, please see the Dynamically-shaped Arrays section
+            below.
+        sig(Sequence[Any]): a tuple indicating the argument signature of the function. Static arguments
+            are indicated with their literal values, and dynamic arguments are indicated by abstract
+            values.
+        kwargs(Dict[str, Any]): keyword argumemts to the function.
+        debug_info(jax.api_util.debug_info): a source debug information object required by jaxprs.
 
     Returns:
         ClosedJaxpr: captured JAXPR
+        Tuple[Tuple[ShapedArray, bool]]: the return type of the captured JAXPR.
+            The boolean indicates whether each result is a value returned by the user function.
         PyTreeDef: PyTree metadata of the function output
         Tuple[Any]: the dynamic argument signature
     """
--- a/frontend/catalyst/from_plxpr/from_plxpr.py
+++ b/frontend/catalyst/from_plxpr/from_plxpr.py
@@ -493,7 +493,7 @@ def handle_measure_in_basis(self, angle, wire, plane, reset, postselect):
 
 # pylint: disable=too-many-positional-arguments
 def trace_from_pennylane(
-    fn, static_argnums, dynamic_argnums, abstracted_axes, sig, kwargs, debug_info=None
+    fn, static_argnums, dynamic_args, abstracted_axes, sig, kwargs, debug_info=None
 ):
     """Capture the JAX program representation (JAXPR) of the wrapped function, using
     PL capure module.
@@ -502,8 +502,7 @@ def trace_from_pennylane(
         fn(Callable): the user function to be traced
         static_argnums(int or Seqence[Int]): an index or a sequence of indices that specifies the
             positions of static arguments.
-        dynamic_argnums(int or Seqence[Int]): an index or a sequence of indices that specifies the
-            positions of dynamic arguments.
+        dynamic_args(Seqence[Any]): the abstract values of the dynamic arguments.
         abstracted_axes (Sequence[Sequence[str]] or Dict[int, str] or Sequence[Dict[int, str]]):
             An experimental option to specify dynamic tensor shapes.
             This option affects the compilation of the annotated function.
@@ -534,7 +533,15 @@ def trace_from_pennylane(
 
         args = sig
 
+        if isinstance(fn, qml.QNode) and static_argnums:
+            # `make_jaxpr2` sees the qnode
+            # The static_argnum on the wrapped function takes precedence over the
+            # one in `make_jaxpr`
+            # https://github.com/jax-ml/jax/blob/636691bba40b936b8b64a4792c1d2158296e9dd4/jax/_src/linear_util.py#L231
+            # Therefore we need to coordinate them manually
+            fn.static_argnums = static_argnums
+
         plxpr, out_type, out_treedef = make_jaxpr2(fn, **make_jaxpr_kwargs)(*args, **kwargs)
-        jaxpr = from_plxpr(plxpr)(*dynamic_argnums, **kwargs)
+        jaxpr = from_plxpr(plxpr)(*dynamic_args, **kwargs)
 
     return jaxpr, out_type, out_treedef, sig
--- a/frontend/test/pytest/test_capture_integration.py
+++ b/frontend/test/pytest/test_capture_integration.py
@@ -1532,7 +1532,7 @@ class TestCapture:
         captured_circuit_2_mlir = captured_circuit_2.mlir
         assert "%cst = arith.constant 2.0" in captured_circuit_2_mlir
         assert 'quantum.custom "RY"(%cst)' in captured_circuit_2_mlir
-        assert "%cst = arith.constant 1.5" not in captured_circuit_1_mlir
+        assert "%cst = arith.constant 1.5" not in captured_circuit_2_mlir
 
         assert result_1 == result_2
 
--- a/frontend/test/pytest/test_capture_integration.py
+++ b/frontend/test/pytest/test_capture_integration.py
@@ -1536,4 +1536,20 @@ class TestCapture:
 
         assert result_1 == result_2
 
+        # Test under a non qnode workflow function
+        @qjit(static_argnums=(0,))
+        def workflow(x, y):
+            @qml.qnode(qml.device(backend, wires=1))
+            def c():
+                qml.RX(x, wires=0)
+                qml.RY(y, wires=0)
+                return qml.expval(qml.PauliZ(0))
+
+            return c()
+
+        result_3 = workflow(1.5, 2.0)
+        captured_circuit_3_mlir = workflow.mlir
+        assert "%cst = arith.constant 1.5" in captured_circuit_3_mlir
+        assert 'quantum.custom "RX"(%cst)' in captured_circuit_3_mlir
+
         qml.capture.disable()
--- a/frontend/test/pytest/test_capture_integration.py
+++ b/frontend/test/pytest/test_capture_integration.py
@@ -1500,3 +1500,36 @@ class TestCapture:
             return qml.sample()
 
         assert jnp.allclose(circuit(), capture_result)
+
+    def test_static_variable_qnode(self, backend):
+        """Test the integration for a circuit with a static variable."""
+
+        qml.capture.enable()
+
+        # Capture in qnode level
+        @qjit(static_argnums=(0,))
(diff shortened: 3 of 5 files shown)


paul0403:
This will need to wait for #1801 , by the way.

.github/workflows/check-weekly-flaky.yaml
dime10, line 1: Your commit history might be messed up, this shouldn't be here.

paul0403: This is from #1801 .

I don't know exactly why this is needed though. @sengthai was this just because locally you were working with latest pennylane? :sweat_smile: 

Or, was there something that made you think that this static argnums work cannot be worked on with just `dev48`?

sengthai: Oops, yes, I rebased to [1801](https://github.com/PennyLaneAI/catalyst/pull/1801) branch 😅. It should be only a few lines of change. Will clean up after #1801

frontend/catalyst/from_plxpr.py
```diff
@@ -867,6 +865,6 @@ def trace_from_pennylane(fn, static_argnums, abstracted_axes, sig, kwargs, debug
         args = sig
 
         plxpr, out_type, out_treedef = make_jaxpr2(fn, **make_jaxpr_kwargs)(*args, **kwargs)
-        jaxpr = from_plxpr(plxpr)(*args, **kwargs)
+        jaxpr = from_plxpr(plxpr)(*dynamic_argnums, **kwargs)
```
paul0403: Shouldn't this be called `dynamic_args`? They are actual arguments right? Not just argnums?

paul0403 (commented):
Nice, simple fix :100: 

Remember to add a test : )

.github/workflows/check-weekly-flaky.yaml
paul0403: @sengthai Does this actually need #1801? If not you can just rebase on top of current `main` no?

sengthai: Yes, I only need #1801 for testing.

sengthai: Let's me rebase to main, then wait for #1801 to be merged.

frontend/catalyst/from_plxpr.py
sengthai: I thought they were dynamic argnums as we discussed with Mudit? 🤔

paul0403: No, I think they are the actual arguments, not a list of indices.

frontend/test/pytest/test_capture_integration.py
```diff
@@ -1500,3 +1500,36 @@ def loop_0(i):
...
+        @qjit(static_argnums=(0,))
+        @qml.qnode(qml.device(backend, wires=1))
+        def captured_circuit_1(x, y):
+            qml.RX(x, wires=0)
+            qml.RY(y, wires=0)
+            return qml.expval(qml.PauliZ(0))
+
+        # Ignore static_argnums in the qnode
+        @qjit(static_argnums=1)
+        @qml.qnode(qml.device(backend, wires=1), static_argnums=0)
```
paul0403: This seems confusing... if we have multiple sources of truth, let us explicitly handle them either by asserting or raising an error, instead of us making a silent decision without telling the user what happened. 

I would argue an explicit failure (either assert or raise is fine with me) is good here, as this isn't a "good" piece of user code to begin with.

frontend/test/pytest/test_capture_integration.py
```diff
@@ -1500,3 +1500,36 @@ def loop_0(i):
...
+        def captured_circuit_2(x, y):
+            qml.RX(x, wires=0)
+            qml.RY(y, wires=0)
+            return qml.expval(qml.PauliZ(0))
+
+        result_1 = captured_circuit_1(1.5, 2.0)
+        assert "stablehlo.constant dense<1.500000e+00>" in captured_circuit_1.mlir
+        assert "stablehlo.constant dense<2.000000e+00>" not in captured_circuit_1.mlir
```
paul0403: Let's save the `.mlir` to a string to prevent recompiling here in the tests : )

frontend/test/pytest/test_capture_integration.py
dime10: I believe product had defined the behaviour when configs are defined differently for different scopes (originally for the autograph kwarg). Looking at the epic, this is as intended!

> qjit static_argnums takes precedence over QNode static_argnums
> ```
> @qml.qjit(static_argnums=0, experimental_capture=True)
> @qml.qnode(dev, static_argnums=2) # ignored!
> ```

dime10: As a reviewer, make sure to read the corresponding shortcut epic and compare the implementation against the requirements outlined there (I still forget this myself as well 😏).

Although it's also fine to raise concerns with requirements of the epic at this stage if there are good reasons for doing something differently.

frontend/catalyst/from_plxpr.py
dime10: Don't forget to update the docstring with the new function signature! (which seems to have changed considerably since it was first written, or it has always been wrong?)

frontend/catalyst/from_plxpr.py
```diff
@@ -746,6 +748,6 @@ def trace_from_pennylane(fn, static_argnums, abstracted_axes, sig, kwargs, debug
         args = sig
```
dime10: What happens with this now that it is unused? Is the `sig` still relevant as a function argument, or are we missing something?

frontend/test/pytest/test_capture_integration.py
paul0403: I see, I didn't look at the epic, sorry about that :awkward:

I'm fine with this if this is the epic's description then. Although personally I would like to see a error/warning message of some sort...

sengthai: > Although personally I would like to see a error/warning message of some sort...

I agree with that. I think these two arguments should be one and solved when Catalyst and Pennylane are unified.
The argument should be in one place. 

If we can't unify these arguments in one place, then I am happy to include a warning message as well. 😅

frontend/test/pytest/test_capture_integration.py
sengthai: Good point! Will change in next commit!

frontend/test/pytest/test_capture_integration.py
dime10: I believe we are in fact getting rid of them on the qnode / unifying them next quarter :)

frontend/catalyst/from_plxpr.py
sengthai: Yes, `sig` is still used for the call to `make_jaxpr2`. But I don't know why `args = sig` in the first place... 🤔

dime10: Ah good point, I missed that. But now we have both sig and dynamic args, does that make sense?

dime10: Just had a look, `sig` is the "full signature", after abstractifying dynamic arguments and leaving static arguments as is. Abstractifying means we remove the concrete values for a set of arguments, and only use the type information (the abstract values).

dime10: I believe tracing to jaxpr needs access to both, the abstract dynamic values and the concrete static values. I guess I don't really know what `from_plxpr` requires (and why it needs arguments in the first place?). But before your change, we gave it the same information as for the jaxpr generation I guess.

dime10: Looking inside `from_plxr`, I think what might need to happen is that you need to forward the `static_argnums` (and probably `abstracted_axes` too, so essentially the `make_jaxpr_kwargs`) to the `make_jaxpr` call inside:
```
    return jax.make_jaxpr(partial(WorkflowInterpreter().eval, plxpr.jaxpr, plxpr.consts))
```

frontend/test/pytest/test_capture_integration.py
```diff
@@ -1500,3 +1500,36 @@ def loop_0(i):
...
+
+        result_2 = captured_circuit_2(1.5, 2.0)
+        assert "stablehlo.constant dense<1.500000e+00>" not in captured_circuit_2.mlir
+        assert "stablehlo.constant dense<2.000000e+00>" in captured_circuit_2.mlir
+
+        assert result_1 == result_2
+
+        qml.capture.disable()
```
dime10, line 1555: Don't manually disable capture, if the test fails this statement won't be executed and it will leave the global state enabled causing trouble for other tests. Instead, use the pytest fixture that will guarantee the state is restored.

dime10: Ah never mind, I forgot I had enabled the fixture globally for this module 😅 
```
pytestmark = pytest.mark.usefixtures("disable_capture")
```
We should be good then 👍

frontend/catalyst/from_plxpr.py
paul0403: > it has always been wrong?

Very likely yes :sweat_smile:

paul0403:
I just realized this doesn't actually produce static IR...

```
module @captured_circuit_1 {
  func.func public @jit_captured_circuit_1(%arg0: tensor<f64>) -> tensor<f64> attributes {llvm.emit_c_interface} {
    %cst = stablehlo.constant dense<1.500000e+00> : tensor<f64>
    %0 = catalyst.launch_kernel @module_captured_circuit_1::@captured_circuit_1(%cst, %arg0) : (tensor<f64>, tensor<f64>) -> tensor<f64>
    return %0 : tensor<f64>
  }

    func.func public @captured_circuit_1(%arg0: tensor<f64>, %arg1: tensor<f64>) -> tensor<f64> attributes {diff_method = "parameter-shift", llvm.linkage = #llvm.linkage<internal>, qnode} {
      quantum.device shots(%c0_i64) ["/home/paul.wang/.local/lib/python3.10/site-packages/pennylane_lightning/liblightning_qubit_catalyst.so", "LightningSimulator", "{'mcmc': False, 'num_burnin': 0, 'kernel_name': None}"]
      %extracted = tensor.extract %arg0[] : tensor<f64>
      %out_qubits = quantum.custom "RX"(%extracted) %1 : !quantum.bit
      %extracted_0 = tensor.extract %arg1[] : tensor<f64>
      %out_qubits_1 = quantum.custom "RY"(%extracted_0) %out_qubits : !quantum.bit
      %2 = quantum.namedobs %out_qubits_1[ PauliZ] : !quantum.obs
      %3 = quantum.expval %2 : f64
      %4 = quantum.insert %0[ 0], %out_qubits_1 : !quantum.reg, !quantum.bit
      quantum.device_release
    }
}
```

It's only static in the sense the the qnode is being called with a static number. The underlying kernel is still dynamic.

paul0403:
Note that the above is not `from_plxpr`, but the 
```
```
aka the plxpr does not pick up the static arg already

paul0403:
> https://github.com/jax-ml/jax/blob/main/jax/_src/interpreters/partial_eval.py#L2408 this line in `trace_to_jaxpr_dynamic2` in jax is changing the `in_type`. Before this line `in_type` is just one tracer, after it's two. It somehow promoted the static arg into a tracer again.
> 
> However this is only happening for the plxpr path.

The only difference I can see before reaching this `f.call_wrapped` line is that in the good old catalyst frontend, on the qnode we have `interface="auto"`, but in plxpr we have `interface="jax"`

paul0403:
Ok, I now think all of my troubles are because the qjit static argnums and the qnode static argnums are fighting with each other :sweat_smile: 

To which I strongly suggest we disallow specifying both at the same time!

frontend/catalyst/from_plxpr.py
paul0403: added docstring

paul0403:
> > Ok, I now think all of my troubles are because the qjit static argnums and the qnode static argnums are fighting with each other 😅
> > To which I strongly suggest we disallow specifying both at the same time!
> 
> ~Maybe we can clear it in the qnode when it's specified in qjit?~
> 
> No I guess that's impossible because we never really encounter the qnode on the catalyst side, happy to disallow this case for now :)

So we actually already do this for non-plxpr pipeline https://github.com/PennyLaneAI/catalyst/blob/main/frontend/catalyst/jit.py#L587

But this is not picked up by `from_plxpr` because it goes through its own `make_jaxpr2`, which sees a qnode with the original qnode static argnums. This fails because in `trace_to_jaxpr_dynamic2` the static argnums followed is the one in the wrapped function. I added a comment. 

But anyway, if we are fine with not explicitly disallowing two configs together, then this PR is now ready for merging. I also updated the tests to directly check the gates are using the static angles.

frontend/catalyst/from_plxpr.py
paul0403: No need, all static argnums are dealt with at the 
```
```
step, before reaching `from_plxpr` at all.

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -515,7 +533,15 @@ def trace_from_pennylane(fn, static_argnums, abstracted_axes, sig, kwargs, debug
 
         args = sig
 
+        if isinstance(fn, qml.QNode) and static_argnums:
+            # `make_jaxpr2` sees the qnode
+            # The static_argnum on the wrapped function takes precedence over the
+            # one in `make_jaxpr`
+            # https://github.com/jax-ml/jax/blob/636691bba40b936b8b64a4792c1d2158296e9dd4/jax/_src/linear_util.py#L231
+            # Therefore we need to coordinate them manually
+            fn.static_argnums = static_argnums
```
dime10, line 542: So I guess as a special case we can overwrite static_argnums when qjit is applied directly to a qnode? Not overly useful imo, since the general case will still fail, but I guess we can have this here for now.

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -515,7 +533,15 @@ def trace_from_pennylane(fn, static_argnums, abstracted_axes, sig, kwargs, debug
...
+
         plxpr, out_type, out_treedef = make_jaxpr2(fn, **make_jaxpr_kwargs)(*args, **kwargs)
-        jaxpr = from_plxpr(plxpr)(*args, **kwargs)
+        jaxpr = from_plxpr(plxpr)(*dynamic_args, **kwargs)
```
dime10, line 545: Since there aren't any changes here with respect to my earlier comment, I'm guessing you concluded this is the right approach? (edit: saw your update now)

Thinking about it again, I guess the jaxpr was traced once with the full signature and static argnums, and as a consequence some arguments where embedded as constants into the jaxpr. Therefor, those don't need to be provided as static arguments again?

I'm not sure if the same holds for abstracted axes 🤔 Does omitting that parameter on a second trace erase the fact that some arguments have abstracted axes?

dime10 (approved):
Looks good!

frontend/catalyst/from_plxpr/from_plxpr.py
paul0403: https://github.com/PennyLaneAI/catalyst/pull/1810/commits/fd20f16ced755e8be6ed601ae2468db9b353d020 it actually works if the inner qnode function just captures the workflow function's args directly instead of taking in its own args. 

But yes, unfortunately if the qnode is not capturing outer variables this would be more tricky. It's gonna be very difficult, since the qnode can be nested at an arbitrary level... I guess this check can only be in `make_jaxpr2`

But for now I think this is fine.

frontend/catalyst/from_plxpr/from_plxpr.py
paul0403: Essentially all the static arg business is handled purely in the "make_plxpr" step. From that point onwards we can just pass the dynamic arguments.

paul0403: > Thinking about it again, I guess the jaxpr was traced once with the full signature and static argnums, and as a consequence some arguments where embedded as constants into the jaxpr. Therefor, those don't need to be provided as static arguments again?

Excatly :100:

paul0403: My gut says abstract_axes follows the same pattern, but I'm not 100% sure. Ideally I would go into jax and find the line that determines what abstract axes gets used (I highly suspect it's also the one on the wrapped function), but there's no time right now :sweat_smile:

paul0403: Also, qnode does not have an `abstract_axes` kwarg so this might be a non-issue in the first place? :thinking:

dime10: > Also, qnode does not have an abstract_axes kwarg so this might be a non-issue in the first place? 🤔

It is an issue, because Catalyst supports it so it needs to work through the PLxPR pipeline. Remember that Catalyst can compile completely classical programs too, so this is independent of any QNode business.

But I agree since the PR is not about abstracted_axes we can skip it for now, I just noticed that it might also have been overlooked in the first from plxpr implementation.

PR #1814, opened 2025-06-17
use `dataclass.replace` to update `ExecutionConfig` or `MCMConfig`

andrijapau:
**Context:**

We are freezing the dataclasses `ExecutionConfig` or `MCMConfig` in Pennylane (https://github.com/PennyLaneAI/pennylane/pull/7697) so they are no longer mutable. Turns out there are external Catalyst tests that failed since there are some files in this repo that are mutating the configuration.

Use `dataclass.replace` to update various properties of the dataclass so we don't mutate.

**Benefits:** Not in place mutating the execution configuration.

**Possible Drawbacks:** None identified.

[sc-93592]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -258,6 +258,9 @@
 
 <h3>Internal changes ⚙️</h3>
 
+* Use `dataclass.replace` to update `ExecutionConfig` and `MCMConfig` rather than mutating properties.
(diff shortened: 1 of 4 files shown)


albi3ro (approved):
👍 Let's also merge this before https://github.com/PennyLaneAI/pennylane/pull/7697

Changelog under "Internals"?

PR #1818, opened 2025-06-19
hoist the pulse duration calculation arithmetic out of the parallel protocol

mehrdad2m:
**Context:**
Currently, the `quantum-to-ion` pass lowers the gates to patterns of ParallelProtocol + pulses. The semantic of ParallelProtocol is that all the instructions inside it are executed in parallel. However, we are including some arithmetic that calculates the duration of pulses within the Parallel protocol, therefore creating data dependency between the parallel instructions.
Moreover, the name of the pass `quantum-to-ion` implies a full conversion from quantum dialect to ion dialect which is not the case. We only convert quantum gates and all the other quantum operations remain intact after the pass is applied.

- hoisted the pulse duration calculation arithmetic out of the parallel protocol
- changed the name of the pass to `gates-to-pulses`

**Benefits:**
This prepares the ion dialect for future development.

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/mlir/lib/Ion/Transforms/QuantumToIonPatterns.cpp
+++ b/mlir/lib/Ion/Transforms/QuantumToIonPatterns.cpp
@@ -233,12 +233,13 @@ mlir::LogicalResult oneQubitGateToPulse(CustomOp op, mlir::PatternRewriter &rewr
         auto qubits = op.getInQubits();
         MLIRContext *ctx = op.getContext();
 
+        auto angle = op.getParams().front();
+        auto time = computePulseDuration(rewriter, loc, angle, beam.rabi, beam.detuning);
+
         auto ppOp = rewriter.create<ion::ParallelProtocolOp>(
             loc, qubits, [&](OpBuilder &builder, Location loc, ValueRange qubits) {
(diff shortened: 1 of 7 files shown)


mlir/lib/Ion/Transforms/gates_to_pulses.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright 2022-2023 Xanadu Quantum Technologies Inc.
+// Copyright 2025 Xanadu Quantum Technologies Inc.
```
paul0403:
```suggestion
// Copyright 2024-2025 Xanadu Quantum Technologies Inc.
```

mlir/lib/Ion/Transforms/QuantumToIonPatterns.cpp
```diff
@@ -233,12 +233,13 @@ mlir::LogicalResult oneQubitGateToPulse(CustomOp op, mlir::PatternRewriter &rewr
         auto qubits = op.getInQubits();
```
paul0403, line 233: Let's rename this pattern file as well then

mehrdad2m: Good point

PR #1819, opened 2025-06-20
Export missing names in the new from_plxpr module

paul0403:
**Context:**
We created a new `from_plxpr` module yesterday, but missed a few things to export.

### DIFF
--- a/frontend/catalyst/from_plxpr/__init__.py
+++ b/frontend/catalyst/from_plxpr/__init__.py
@@ -15,4 +15,4 @@
 """Conversion from plxpr to catalyst jaxpr"""
 
 from catalyst.from_plxpr.control_flow import handle_cond, handle_for_loop, handle_while_loop
-from catalyst.from_plxpr.from_plxpr import from_plxpr, trace_from_pennylane
+from catalyst.from_plxpr.from_plxpr import from_plxpr, register_transform, trace_from_pennylane


paul0403:
> > > Tbh I still think `program_capture` would be a better name for this sub-module, one to distinguish it from the individual `from_plxpr.py` module but also to provide a name that's easier to understand even without knowing what plxpr is (that is it describes the purpose rather than the implementation).
> > 
> > 
> > I thought about it, but then I realized changing the names right now would mean external packages (either user or just PL/lightning) will fail lines like
> > ```python
> > from catalyst.from_plxpr import blah
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > so I just left it the same, since I did not want to cause more cross-repo car crashes at the end of the quarter.
> > We can properly rename this in feature freeze.
> 
> Who would be using this module? It's not publicly exported in our API 🤔

So I know at least xdsl tests in the PL repo is importing "internals" from `from_plxpr.py`. It's not public-public, but it still causes some headaches internally for us I think.

dime10:
> > > > Tbh I still think `program_capture` would be a better name for this sub-module, one to distinguish it from the individual `from_plxpr.py` module but also to provide a name that's easier to understand even without knowing what plxpr is (that is it describes the purpose rather than the implementation).
> > > 
> > > 
> > > I thought about it, but then I realized changing the names right now would mean external packages (either user or just PL/lightning) will fail lines like
> > > ```python
> > > from catalyst.from_plxpr import blah
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > so I just left it the same, since I did not want to cause more cross-repo car crashes at the end of the quarter.
> > > We can properly rename this in feature freeze.
> > 
> > 
> > Who would be using this module? It's not publicly exported in our API 🤔
> 
> So I know at least xdsl tests in the PL repo is importing "internals" from `from_plxpr.py`. It's not public-public, but it still causes some headaches internally for us I think.

Got it, thanks! That might not be great, we shouldn't use internal modules in other packages 🤔

PR #1820, opened 2025-06-20
A new JAX primitive for quantum definitions is added to Catalyst

maliasadi:
**Context:**
A new JAX primitive for decomposition-rule definitions is added to Catalyst.

This functionality is not supposed to be used by users, instead it's used in the decomposition system to lower rules from PennyLane --> PLxPR --> MLIR.

**Feature PRs:**
- [X] Added a new JAX primitive to capture and compile the decomposition rule definitions to MLIR. `decomposition_rule` is the decorator integrated with this primitive for development purposes.
- [X] Renamed `QregManager` to `QubitHandler` and extended the class to manage
  converting PLxPR wire indices into Catalyst JAXPR qubits.
  This is especially useful for lowering decomposition rules that take
  in qubits as arguments.

[sc-95656]

### DIFF
--- a/frontend/test/lit/test_decomposition.py
+++ b/frontend/test/lit/test_decomposition.py
@@ -386,9 +386,9 @@ def test_decomposition_rule_shaped_wires():
 
     # CHECK: func.func private @shaped_wires_rule([[QREG:%.+]]: !quantum.reg, [[PARAM_TENSOR:%.+]]: tensor<f64>, [[QUBITS:%.+]]: tensor<3xi64>) -> !quantum.reg
     # CHECK-NEXT: [[IDX_0:%.+]] = stablehlo.slice [[QUBITS]] [0:1] : (tensor<3xi64>) -> tensor<1xi64>
-    # CHECK-NEXT: [[RIDX_0: %.+]] = stablehlo.reshape [[IDX_0]] : (tensor<1xi64>) -> tensor<i64>
-    # CHECK-NEXT: [[EXTRACTED: %.+]] = tensor.extract %1[] : tensor<i64>
-    # CHECK-NEXT: [[QUBIT:%.+]] = quantum.extract [[QREG]][%extracted] : !quantum.reg -> !quantum.bit
+    # CHECK-NEXT: [[RIDX_0:%.+]] = stablehlo.reshape [[IDX_0]] : (tensor<1xi64>) -> tensor<i64>
+    # CHECK-NEXT: [[EXTRACTED:%.+]] = tensor.extract [[RIDX_0]][] : tensor<i64>
+    # CHECK-NEXT: [[QUBIT:%.+]] = quantum.extract [[QREG]][[[EXTRACTED]]] : !quantum.reg -> !quantum.bit
     # CHECK-NEXT: [[EXTRACTED_0:%.+]] = tensor.extract [[PARAM_TENSOR]][] : tensor<f64>
     # CHECK-NEXT: [[OUT_QUBITS:%.+]] = quantum.custom "RX"([[EXTRACTED_0]]) [[QUBIT]] : !quantum.bit
 


test.py
```diff
@@ -10,8 +10,10 @@
 @qdef
 def decomp(x, y, w):
     jax.debug.print("hello add {} and {}", x, y)
-    qml.GlobalPhase(0.1)
-
+    qml.GlobalPhase(1.14)
+    qml.RX(x, wires=1)
+    qml.RY(y, wires=2)
```
dime10: I thought we said this didn't make any sense? 😅

dime10:
Please consider incorporating https://github.com/PennyLaneAI/catalyst/commit/f9647b9efa295e4a4d118998c78e313c7dcd8483 since I would really like to avoid having every decomposed gate loose its qubit connectivity information.

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -43,15 +43,17 @@
...
+from catalyst.from_plxpr.qreg_manager import QbitManager, QregManager
 from catalyst.jax_extras import jaxpr_pad_consts, make_jaxpr2, transient_jax_config
 from catalyst.jax_primitives import (
+    AbstractQbit,
     MeasurementPlane,
     adjoint_p,
     compbasis_p,
     cond_p,
     counts_p,
+    decomposition_rule_p,
```
maliasadi: @dime10 I've changed the name to `decomposition_rule` since the last review

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -505,6 +506,45 @@ def wrapper(qreg, *args):
     return vals_out
 
 
+@PLxPRToQuantumJaxprInterpreter.register_primitive(decomposition_rule_p)
+def handle_decomposition_rule(self, *, pyfun, func_jaxpr, is_qreg, num_params):
+    """
+    Transform a quantum decomposition rule from PLxPR into JAXPR with quantum primitives.
+    """
+
+    if is_qreg:
```
maliasadi, line 512: I had to add this to support
- Wires slicing
- PL's WiresLike data type
- control-flow ops inside rules

I've included lit tests [here](https://github.com/PennyLaneAI/catalyst/pull/1820/files#diff-0f5507ff6a68593e9e594c12dea5ccfaa47299edb0533c6df60a7e8ee637fcd5R231)

frontend/catalyst/passes/__init__.py
```diff
@@ -39,6 +39,7 @@
     disentangle_cnot,
     disentangle_swap,
     get_ppm_specs,
+    graph_decomposition,
```
maliasadi: This is related to the draft pass that we'll work on in a followup PR this week. The decomp pass draft branch is based on this one. 


Update: Removed from this PR to make the reviews easier. Added to [mlir_gdecomp/pl_transform](https://github.com/PennyLaneAI/catalyst/tree/mlir_gdecomp/pl_transform)

frontend/catalyst/from_plxpr/qreg_manager.py
```diff
@@ -188,3 +188,107 @@ def __setitem__(self, index: int, qubit: AbstractQbit):
     def __iter__(self):
         """Iterate over wires map dictionary"""
         return iter(self.wire_map.items())
+
+
+class QbitManager:
```
mlxd: Can we name this something new, more related to its scope and usage?
We have way too many qubit/qbit managers in the library already.

maliasadi: Good idea @mlxd! This class manages `quantum.bit` rather than `quantum.register`. We already have `QRegManager` for registers, so this new class would handle q values. A few suggestions:
- QBitHandler
- QbitIRManager
- QbitValueManager
- QbitValueHandler

dime10: Yes it's just to be analogous to the QregManager. Happy to change it but we should probably update both in that case.

Alternatively, there is the possibility that we merge the two classes and just allow the QregManager to have `None` for the qreg attribute. Admitedly the QubitManager was mostly quickly hacked together to be able to use it in-place of the QregManager (to demonstrate the concept), but it would be good to clean that up @maliasadi 

Alternative names that describe the _function_ of these classes a bit better:
- QubitValueMap
- ValueSemanticMap
- ValueSemanticConverter

mlxd: > we should probably update both in that case.

Good call

> QubitValueMap

I like this one the most, but the other two options sound reasonable too.

maliasadi: > we should probably update both in that case

Yeah, I believe there is actually a product effort focused on finding coherent names and definitions for different qubit managers and maps across the ecosystem. cc: @isaacdevlugt  

> QubitValueMap

I updated the name to this for now :+1:

isaacdevlugt: AFAIK any of the qubit-managing objects we have in the ecosystem aren't user-facing, so I have no strong opinions 👀

dime10: Did you look at merging the two classes @maliasadi?

maliasadi: @dime10 I didn't merge these two classes after renaming because they have quite different goals and APIs. `QubitValueMap` has a much simpler API that doesn't require extracting or inserting registers. I'm happy to merge the classes if you think otherwise, but then we would need to address the following:
- A new name for `QRegManager` that includes the intent of `QubitValueMap`.
- Defining default behaviours for the get, set, insert, and extract methods of `QRegManager` while maintaining the functionality of `QubitValueMap`.

dime10: Well that's exactly it, in my quick prototype I was copying the API on the QRegManager because I used the QubitManager in places where the QRegManager was expected (you can see most of it just raises not-implemented errors). That doesn't feel very clean to me, and since we are merging this into production I thought it make sense to clean this up a bit. You can see `insert_dynamic_qubits` for example does nothing, which also feels hacky.

frontend/catalyst/jax_primitives.py
```diff
@@ -552,6 +574,22 @@ def _func_lowering(ctx, *args, call_jaxpr, fn):
...
 
+#
+# Decomp rule
+#
+@decomprule_p.def_abstract_eval
+def _decomposition_rule_abstract(*, pyfun, func_jaxpr, is_qreg=False, num_params=None):
+    return ()
+
+
+def _decomposition_rule_lowering(ctx, *, pyfun, func_jaxpr):
```
dime10: I think this needs the remaining kwargs too (although we don't use them):
```suggestion
def _decomposition_rule_lowering(ctx, *, pyfun, func_jaxpr, **_):
```

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -43,15 +43,17 @@
 from pennylane.transforms import single_qubit_fusion as pl_single_qubit_fusion
 from pennylane.transforms import unitary_to_rot as pl_unitary_to_rot
 
-from catalyst.device import extract_backend_info, get_device_capabilities
-from catalyst.from_plxpr.qreg_manager import QregManager
+from catalyst.device import extract_backend_info
+from catalyst.from_plxpr.qreg_manager import QregManager, QubitValueMap
```
dime10: I just realized that the QregManager that was here recently is only used for plxpr, thus largely duplicates the existing manager here: [QRegPromise](https://github.com/PennyLaneAI/catalyst/blob/85b1e6ac4cb42cc07bbd93458338dc409f92218c/frontend/catalyst/jax_tracer.py#L340C7-L340C18)

Oh well, not this PR's fault but maybe a reason not to add yet a third one 😅

frontend/catalyst/from_plxpr/from_plxpr.py
dime10: Good point! With control-flow ops in there current form the register is required, so this dual-mode support is not a bad idea :)

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -505,6 +506,45 @@ def wrapper(qreg, *args):
...
+
+        converted_closed_jaxpr_branch = jax.make_jaxpr(wrapper)(
+            self.qreg_manager.get(), *func_jaxpr.in_avals
+        )
+    else:
+
+        def wrapper(*args):
+            manager = QubitValueMap(args[num_params:])
+            converter = copy(self)
+            converter.qreg_manager = manager
```
dime10: This attribute is not super clean, which I was trying to get at in my other comment. I think we should update the converter class and lines using this attribute to make sure it also expects a register-free value mapper.

frontend/catalyst/from_plxpr/qreg_manager.py
```diff
@@ -188,3 +188,106 @@ def __setitem__(self, index: int, qubit: AbstractQbit):
...
+           qml.Hadamard(wires=w)
+           qml.X(wires=0)
+        In the above, because we don't know whether `w` is `0` or not until runtime, we must not
+        cancel the inverse PauliX gates on `0`.
+
+        Thus in the IR def use chain we need to interrupt the first X gate's %out_qubit to be used
+        as the second X gate's qubit operand directly.
+        To do this, for the dynamic wire gate we insert back to the register.
+        """
+        # do not do anything since we don't deal with dynamic wires
```
dime10: This is a bit dirty, since the original interface and call sites of this method would expect it to do something.

frontend/catalyst/from_plxpr/qreg_manager.py
```diff
@@ -188,3 +188,106 @@ def __setitem__(self, index: int, qubit: AbstractQbit):
...
+        # and each gate using the integer wires originally is interpreted with qubit wires instead
+        self.wire_map = dict(zip(init_qubit_refs, init_qubit_refs))
+
+    def get_final_qubits(self) -> list[AbstractQbit]:
+        """Return the final qubit states."""
+        return [self.wire_map[idx] for idx in self.qubit_indices]
+
+    def get(self):
+        """Return the current AbstractQreg value."""
+        raise NotImplementedError()
```
dime10: All of these methods are a bit superfluous of course, I put them there so I would find out if some code tried to call the qubit manager where I didn't expect it, but I don't know if we want to keep these stubs in this form.

frontend/catalyst/jax_primitives.py
```diff
@@ -312,9 +312,11 @@ class MeasurementPlane(Enum):
 quantum_kernel_p.multiple_results = True
 measure_in_basis_p = Primitive("measure_in_basis")
 measure_in_basis_p.multiple_results = True
+decomprule_p = core.Primitive("decomposition_rule")
```
dime10, line 319: I showed this to @mudit2812 btw since he is quite interested in such a function definition primitive without requiring a call from within a qjit/qnode. We can keep the `decomprule_p` for now but just putting this here that we'll probably want to generalize it since it is useful in a few contexts. We might also want to look at the overlap with the `subroutine_p` primitive that Erick implemented.

Anyway, no changes required.

mlir/Makefile
```diff
@@ -53,7 +53,7 @@ help:
 	@echo "  format [version=?] to apply C++ formatter; use with 'version={version}' to run clang-format-{version} instead of clang-format"
 
 .PHONY: all
-all: llvm mhlo enzyme dialects dialect-docs plugin
+all: llvm mhlo  enzyme dialects plugin
```
dime10: I'm guessing this change can be reverted now? (it seems unrelated to the pr)
```suggestion
all: llvm mhlo  enzyme dialects dialect-docs plugin
```

frontend/catalyst/from_plxpr/qreg_manager.py
maliasadi: We could use `get` to return the final qubit states in this class in favour of removing the custom `get_final_qubits`, and remove `set`, `extract`, `insert` `insert_all_dangling_qubits` and `insert_dynamic_qubits` until we find some applications for these methods.

frontend/catalyst/from_plxpr/qreg_manager.py
maliasadi: What do you think about my comment [here](https://github.com/PennyLaneAI/catalyst/pull/1820/files#r2276637674)?

frontend/catalyst/from_plxpr/from_plxpr.py
maliasadi: What do you think of renaming it to `qubit_handler` and use it for both `QregManager` and `QubitValueMap`?

frontend/catalyst/from_plxpr/from_plxpr.py
maliasadi: Ah, I see your point. we can merge these two classes for sure! I can rename `QRegManager` to `QubitHandler( root_qreg_value: AbstractQreg = None, init_qubit_refs: list[AbstractQbit] = [])` and merge `QubitValueMap` with this.

frontend/catalyst/from_plxpr/from_plxpr.py
dime10: That sounds good 👍

frontend/catalyst/from_plxpr/qreg_manager.py
dime10: Any chance we can make the converter not call this method if the reg is not used, would that make it cleaner?

frontend/catalyst/from_plxpr/qreg_manager.py
dime10: Yeah feel free to try some improvements, just keep the code in mind where these functions are used. 

If you don't find anything good we'll just merge what we have

frontend/catalyst/from_plxpr/from_plxpr.py
dime10: Sounds great 🤩

frontend/catalyst/from_plxpr/qreg_manager.py
maliasadi: Done [55e2b6b](https://github.com/PennyLaneAI/catalyst/pull/1820/commits/55e2b6bb1f2eb29877e814e3219fc4fd69d408cd)

dime10:
Thanks @maliasadi! I can't review the new version in detail but from my side consider it approved :) Maybe @paul0403 can have look over the changes to the existing manager and give the final approval 💯

doc/releases/changelog-dev.md
```diff
@@ -23,6 +23,17 @@
...
 
+* Added a new JAX primitive to capture and compile the decomposition rule
+  definitions to MLIR. `decomposition_rule` is the decorator integrated
+  with this primitive for development purposes.
+  [(#1820)](https://github.com/PennyLaneAI/catalyst/pull/1820)
+
+* Renamed `QregManager` to `QubitHandler` and extended the class to manage
+  converting PLxPR wire indices into Catalyst JAXPR qubits.
+  This is especially useful for lowering decomposition rules that take
+  in qubits as arguments.
```
```suggestion
  This is especially useful for lowering subroutines that take
  in qubits as arguments, for example decomposition rules.
```

frontend/test/lit/test_decomposition.py
```diff
@@ -265,3 +268,243 @@ def decompose_to_matrix():
...
+        # CHECK: [[QREG:%.+]] = quantum.alloc
+        # CHECK-NEXT: [[QUBIT:%.+]] = quantum.extract [[QREG]][ 0] : !quantum.reg -> !quantum.bit
+        # CHECK-NEXT: [[QUBIT_1:%.+]] = quantum.custom "Hadamard"() [[QUBIT]] : !quantum.bit
+        # CHECK-NEXT: [[QREG_1:%.+]] = quantum.insert [[QREG]][ 0], [[QUBIT_1]] : !quantum.reg, !quantum.bit
+        # CHECK-NEXT: quantum.compbasis qreg [[QREG_1]] : !quantum.obs
+        identity()
+        all_wires_rx(float, int, int, int)
+        qml.Hadamard(0)
+        return qml.probs()
+
```
paul0403, line 357: Question: in this mlir, what is instructing the later inliner pass @rniczh is writing where to inline the private `identity` and `all_wires_rx` to? Seems like the main circuit ir has lost the two call sites? :thinking: Or am I missing any of the partial instantiation you talked about?

frontend/test/lit/test_decomposition.py
```diff
@@ -265,3 +268,243 @@ def decompose_to_matrix():
...
+        # CHECK: module @circuit_4
+        shaped_wires_rule(float, jax.core.ShapedArray((3,), int))
+        qml.Hadamard(0)
+        return qml.probs()
+
+    # CHECK: func.func private @shaped_wires_rule([[QREG:%.+]]: !quantum.reg, [[PARAM_TENSOR:%.+]]: tensor<f64>, [[QUBITS:%.+]]: tensor<3xi64>) -> !quantum.reg
+    # CHECK-NEXT: [[IDX_0:%.+]] = stablehlo.slice [[QUBITS]] [0:1] : (tensor<3xi64>) -> tensor<1xi64>
+    # CHECK-NEXT: [[RIDX_0: %.+]] = stablehlo.reshape [[IDX_0]] : (tensor<1xi64>) -> tensor<i64>
+    # CHECK-NEXT: [[EXTRACTED: %.+]] = tensor.extract %1[] : tensor<i64>
+    # CHECK-NEXT: [[QUBIT:%.+]] = quantum.extract [[QREG]][%extracted] : !quantum.reg -> !quantum.bit
```
paul0403:
```suggestion
    # CHECK-NEXT: [[EXTRACTED: %.+]] = tensor.extract [[RIDX_0]][] : tensor<i64>
    # CHECK-NEXT: [[QUBIT:%.+]] = quantum.extract [[QREG]][[[EXTRACTED]]] : !quantum.reg -> !quantum.bit
```

frontend/catalyst/from_plxpr/qubit_handler.py
```diff
@@ -20,13 +20,23 @@
...
+
+    If `qubit_or_qreg_ref` is an `AbstractQreg`, this handler manages converting the qubits in qreg.
     The fundamental challenge in from_plxpr is that Plxpr (and frontend PennyLane) uses wire index
     semantics, but Catalyst jaxpr uses qubit value semantics.
 
+    However, if `qubit_or_qreg_ref` is a list/tuple of `AbstractQbit` values, this handler
+    manages converting the qubits in the list/tuple. This is useful when the qubits
+    are passed in as arguments to the function being converted. This feature is mainly
+    useful for lowering decomposition rules that take in qubits as arguments.
+
```
paul0403:
```suggestion
    Args:
        qubit_or_qreg_ref: An `AbstractQreg` value, or a list/tuple of `AbstractQbit` values.

    If `qubit_or_qreg_ref` is an `AbstractQreg`, this handler manages converting the qubits in qreg.     
    However, if `qubit_or_qreg_ref` is a list/tuple of `AbstractQbit` values, this handler
    manages converting the qubits in the list/tuple. This is useful when the qubits
    are passed in as arguments to the function being converted. This feature is mainly
    useful for lowering decomposition rules that take in qubits as arguments.
    
    The fundamental challenge in from_plxpr is that Plxpr (and frontend PennyLane) uses wire index
    semantics, but Catalyst jaxpr uses qubit value semantics.
```

frontend/catalyst/from_plxpr/qubit_handler.py
```diff
@@ -61,31 +71,47 @@ class QregManager:
...
     wire_map: dict[int, AbstractQbit]  # Note: No dynamic wire indices for now in from_plxpr.
 
-    def __init__(self, root_qreg_value: AbstractQreg):
-        self.abstract_qreg_val = root_qreg_value
+    def __init__(self, qubit_or_qreg_ref: AbstractQreg | list[AbstractQbit] | tuple[AbstractQbit]):
+
+        if isinstance(qubit_or_qreg_ref, (list, tuple)):
+            self.abstract_qreg_val = None
+            self.qubit_indices = qubit_or_qreg_ref
+            # oddly enough we want to map the refs to themselves initially, because when we interpret
```
paul0403: Yes I noticed this as well! I don't think it's that odd though.

(Let's satisfy the ling-too-long pylints here btw)

frontend/catalyst/from_plxpr/qubit_handler.py
```diff
@@ -95,16 +121,27 @@ def set(self, qreg: AbstractQreg):
...
             raise CompileError(
                 "Setting new qreg value, but the previous one still has dangling qubits."
             )
+
+        # Devalidate the old qubit values if user wants to set a qreg
+        if self.abstract_qreg_val is None:
+            self.wire_map = {}
+            self.qubit_indices = []
+
         self.abstract_qreg_val = qreg
```
paul0403: I think the logic here can be a bit more clarified by just cleanly separating the modes:
```suggestion
        if self.abstract_qreg_val is not None:
           # qreg mode
           if len(self.wire_map) != 0:
               raise CompileError(
                   "Setting new qreg value, but the previous one still has dangling qubits."
               )
           self.abstract_qreg_val = qreg
        else:
           # qubits mode
           # Devalidate the old qubit values if user wants to set a qreg
            self.wire_map = {}
            self.qubit_indices = []


```

frontend/catalyst/from_plxpr/qubit_handler.py
```diff
@@ -124,6 +167,10 @@ def insert_all_dangling_qubits(self):
         This is necessary, for example, at the end of the qreg lifetime before deallocing,
         or when passing qregs into and out of scopes like control flow.
         """
+        if self.abstract_qreg_val is None:
```
paul0403, line 172: Why not fail this method like the others? Just wondering

frontend/catalyst/from_plxpr/qubit_handler.py
```diff
@@ -141,6 +188,9 @@ def insert_dynamic_qubits(self, wires):
         as the second X gate's qubit operand directly.
         To do this, for the dynamic wire gate we insert back to the register.
         """
+        if self.abstract_qreg_val is None:
+            # do nothing since we don't deal with dynamic qubits
```
paul0403: Same here

frontend/test/pytest/from_plxpr/test_from_plxpr_qubit_handler.py
```diff
@@ -0,0 +1,443 @@
...
+
+        assert isinstance(qubit_handler.get(), list)
+        assert all(isinstance(q, AbstractQbit) for q in qubit_handler.get())
+        assert len(qubit_handler.get()) == 2
+
+        qubit_handler.set("monkey_mock_qubit")
+        assert qubit_handler.get() == "monkey_mock_qubit"
+
+    def test_getitem_setitem(self):
+        """Test that getting and setting items in a QubitHandler raises NotImplementedError"""
```
paul0403: Is this still true? :sweat_smile:

frontend/catalyst/from_plxpr/qubit_handler.py
```diff
@@ -20,13 +20,23 @@
 from catalyst.utils.exceptions import CompileError
 
 
-class QregManager:
+class QubitHandler:
```
paul0403, line 23: Can you add a test where there are gates using the `QubitHandler` in the list of qubits mdoe? I guess it would look something like 

```python
        q0 = AbstractQbit()
        q1 = AbstractQbit()
        qubit_handler = QubitHandler([q0, q1])
        out_qubits_0, out_qubits_1 =  qubit_mock_op_p.bind(q0, q1)

        qubit_handler[q0] = out_qubits_0
        qubit_handler[q1] = out_qubits_1

        # add checks...
```
? Are the keys of the map in qubit mode just the starting qubit values?

paul0403 (commented):
Nice!

frontend/test/lit/test_decomposition.py
maliasadi: This is actually the main difference with subroutines. We don't need call sites when lowering the circuit. What we need is a list of private decomposition rules in the compiler program, which we can then use in the rewrite decomposition MLIR pass (@rniczh's PR) in later stages.

frontend/catalyst/from_plxpr/qubit_handler.py
maliasadi: Hmm.. it doesn't work as it requires qubit extraction and we need a quantum register for that.

frontend/test/pytest/from_plxpr/test_from_plxpr_qubit_handler.py
maliasadi: good catch! Fixed [be0bca8](https://github.com/PennyLaneAI/catalyst/pull/1820/commits/be0bca83fbeca7a6b498f8fbd8d00466ddf9bdd5)

frontend/catalyst/from_plxpr/qubit_handler.py
maliasadi: Fixed [be0bca8](https://github.com/PennyLaneAI/catalyst/pull/1820/commits/be0bca83fbeca7a6b498f8fbd8d00466ddf9bdd5)

frontend/catalyst/from_plxpr/qubit_handler.py
maliasadi: Fixed [be0bca8](https://github.com/PennyLaneAI/catalyst/pull/1820/commits/be0bca83fbeca7a6b498f8fbd8d00466ddf9bdd5)

frontend/catalyst/from_plxpr/qubit_handler.py
maliasadi: Fixed [be0bca8](https://github.com/PennyLaneAI/catalyst/pull/1820/commits/be0bca83fbeca7a6b498f8fbd8d00466ddf9bdd5)

frontend/catalyst/from_plxpr/qubit_handler.py
paul0403: Hmm, I think the mock op does not require registers.... but don't stress too much here, if it's too difficult we don't have to add them.

maliasadi: This is the code I was trying with it:

``` python
        qbits = [AbstractQbit(), AbstractQbit()]
        qubit_handler = QubitHandler(qbits)
        wires = [0, 1]
        in_qubits = [qubit_handler[w] for w in wires]
        out_qubits =  qubit_mock_op_p.bind(*qbits)
```

did I do anything wrong? :thinking:

frontend/test/lit/test_decomposition.py
paul0403: I think your latest commit reverted this update

frontend/catalyst/from_plxpr/qubit_handler.py
paul0403: I think because in the init you just put `qbits = [AbstractQbit(), AbstractQbit()]` as the keys of the map instead of the `wires=[0,1]`

frontend/test/lit/test_decomposition.py
maliasadi: Ah, it didn't work with these changes

paul0403: What do you mean? The test IR has to have the explicit `%1` and `%extracted`? :thinking:

PR #1839, opened 2025-07-01
check for catalyst_cli in $PATH

0mWh:
**Context:**
Catalyst could not find itself in an astral-uv installation.

Search for "catalyst" cli executable on the system path.

**Benefits:**
It works in astral-uv

**Possible Drawbacks:**
Different "catalyst" on the system path?

**Related GitHub Issues:**
This

### DIFF
--- a/frontend/catalyst/utils/runtime_environment.py
+++ b/frontend/catalyst/utils/runtime_environment.py
@@ -19,6 +19,7 @@ import os
 import os.path
 import sys
 import sysconfig
+import shutil
 
 from catalyst._configuration import INSTALLED
 
@@ -86,6 +87,11 @@ def get_cli_path() -> str:  # pragma: nocover
     # Fallback to python location
     path = os.path.join(os.path.dirname(sys.executable), catalyst_cli)
     if os.path.isfile(path):
+        return path
+
+    # Environment path
+    path = shutil.which(catalyst_cli)
+    if path is not None:
         return path
 
     raise RuntimeError(


dime10:
Hi @0mWh, thank you for contributing to Catalyst! We'll have someone from the team review your PR.

dime10:
Thank you again for your PR @0mWh! We've had several discussions around it, and came to the conclusion that we'd like to avoid matching against any executable on `PATH` that just happens to be named `catalyst`.

What we're proposing is adding an environment variable called `CATALYST_EXECUTABLE` that can be customized by the user if desired, and raise an error message pointing out this variable as an alternative if Catalyst can't find it's binary like in your case.

Please let us know if that would cover your use-case. Alternatively, if you'd like to provide some additional details about the issue you're encountering (e.g. what the directory structure of your installation is like, including the path to the catalyst binary) we could also look into that.

0mWh:
Thanks for the questions & sorry for the delay; patching it locally de-prioritized this for me.

---

> we'd like to avoid matching against any executable on `PATH` that just happens to be named `catalyst`.

I'm not sure what % of cases this would affect. Maybe some mechanism could verify if the catalyst binary is actually your pennylane catalyst? Also, at least in my case, the directory in which catalyst_cli resides is at the front of the PATH anyway.

---

> adding an environment variable called `CATALYST_EXECUTABLE` that can be customized by the user if desired   

The problem is that the environment's runtime path can have GUIDs and randomized strings (horrible, I know), so I do not think that this be determined reliably.

---

> Please let us know if that would cover your use-case. Alternatively, if you'd like to provide some additional details about the issue you're encountering (e.g. what the directory structure of your installation is like, including the path to the catalyst binary) we could also look into that.

I patched the install on my system to use it [here, in cell 4](https://github.com/0mWh/pqic-gic-quadrigems/blob/main/notebooks/3_simulated_circuits.ipynb), setup with uv with [this pyproject.toml](https://github.com/0mWh/pqic-gic-quadrigems/blob/main/pyproject.toml)

PR #1845, opened 2025-07-04
Update a test for a new manual decomposition in core

paul0403:
**Context:**
The manual decomposition of `SingleExcitationPlus` was updated in core https://github.com/PennyLaneAI/pennylane/pull/7771

Change the gates we are checking for to follow the new decomposition.

**Benefits:**
Tests pass.
Latest/latest passes.

### DIFF
--- a/frontend/test/pytest/test_preprocess.py
+++ b/frontend/test/pytest/test_preprocess.py
@@ -14,6 +14,7 @@
 """Test for the device preprocessing."""
 import platform
 from dataclasses import replace
+from typing import List
 
 import numpy as np
 import pennylane as qml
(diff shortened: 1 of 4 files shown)


paul0403:
No changelog for just tests.

erick-xanadu (approved):

paul0403:
For the folks who are also updating dep versions, PL version has been updated to latest here. (@JerryChen97 )

frontend/test/pytest/test_preprocess.py
```diff
@@ -122,9 +122,9 @@ def circuit(theta: float):
             return qml.state()
 
         mlir = qjit(circuit, target="mlir").mlir
-        assert "PauliX" in mlir
+        assert "Hadamard" in mlir
```
joeycarter, line 124: I think this is fine for a quick fix, but I'm wondering if we could make the test more robust against future changes in PennyLane by decomposing the tape and comparing the ops in the decomposition with the ops in the MLIR representation, e.g. 

```python
ops = circuit._tape.operations[0].decomposition()

for op in ops:
    assert map_op_name(op) in mlir
```

Where `map_op_name` is some function that maps the PennyLane op name to its quantum-dialect equivalent.

frontend/test/pytest/test_preprocess.py
```diff
@@ -122,9 +122,9 @@ def circuit(theta: float):
             return qml.state()
 
         mlir = qjit(circuit, target="mlir").mlir
```
joeycarter, line 123: While we're here, I noticed we use `qjit` twice. We can probably get rid of the `@qjit` decorator if we use it as a function call here.

frontend/test/pytest/test_preprocess.py
paul0403: So I tried this and this proved to be a bit more difficult. Some ops in pennylane can't just be decomposed "in a vacuum". For example for the blockenocde op I see

```
pennylane.operation.DecompositionUndefinedError
```

joeycarter: I wouldn't say this has to be a universal practice, only where a decomposition is well defined. For example, I see that for `qml.BlockEncode`, all we check is

```
assert "quantum.unitary" in mlir
assert "BlockEncode" not in mlir
```

which I think is fine.

joeycarter: Anyway, just a suggestion. If it's proving too difficult we can tackle it later when we have capacity.

joeycarter: Also this strategy probably won't work for lit tests (which I see also require updating).

paul0403: Yes, this strategy does not work for lit tests since we can't (?) communicate python variables to filecheck variables. 

I actually have this implemented already, it's not a big thing, but then I realized out of the four lit-ish tests in this pytest only one would work from this strategy :sweat_smile: For the three failing ones:
- two are PL gates with undefined decompositions
- one is using a mock gate in this test file, so it also does not have PL decomposition

I will push my prototype for git history to record it in case we want to resurrect it, then revert.

joeycarter: Sounds good, thanks @paul0403

frontend/test/pytest/test_preprocess.py
paul0403: https://github.com/PennyLaneAI/catalyst/pull/1845/commits/2afdc56db6204c7bb197a936d075702a2ddd884d

PR #1846, opened 2025-07-04
Better checking for runtime device kwargs string not having nested dictionaries

paul0403:
**Context:**
#1843 bans nested dictionaries in device kwargs.
I assumed that there was only the global dictionary by asserting the starting and ending char of the string are `{}`, 
but this [breaks some tests in lightning repo](https://github.com/PennyLaneAI/pennylane-lightning/actions/runs/16075489083/job/45369618669?pr=1192#step:6:304).

This assumption is true for devices going through catalyst frontend, but for standalone tests this is not true.

Do an actual nested curly braces check.

**Benefits:**
Tests pass.

**Possible Drawbacks:**
The failing tests in lightning are not actually lightning tests. They are catalyst runtime tests piggy-backing off the lightning device.
This is because null.qubit did not exist back when we moved the lightning device's catalyst interface into the lightning repo, so we moved all the tests. 
We should move the catalyst runtime tests back to the catalyst repo in the future.

### DIFF
--- a/frontend/test/pytest/test_capture_integration.py
+++ b/frontend/test/pytest/test_capture_integration.py
@@ -376,7 +376,7 @@ class TestCapture:
 
         qml.capture.disable()
 
-        assert capture_result == expected
+        assert jnp.allclose(capture_result, expected)
 
     def test_measure_postselect(self, backend):
         """Test the integration for a circuit with a mid-circuit measurement using postselect.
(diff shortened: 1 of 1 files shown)


maliasadi (approved):
:raised_hands:

frontend/test/pytest/test_preprocess.py
```diff
@@ -337,6 +337,7 @@ def test_decomposition_of_forloop_circuit(self, reps, angle):
 
         dev = qml.device("lightning.qubit", wires=2)
 
+        @qjit
```
maliasadi, line 340: :thinking:

paul0403: A test was comparing non qjit with non qjit :sweat_smile: I'm sneaking it in here

PR #1847, opened 2025-07-04
Update a test to use np.allclose instead of `==`

paul0403:
**Context:**
A test is mysteriously failing on ARM machines: https://github.com/PennyLaneAI/catalyst/actions/runs/16077576091/job/45378065361#step:12:178

```
FAILED frontend/test/pytest/test_capture_integration.py::TestCapture::test_transform_single_qubit_fusion_workflow[lightning.qubit] - assert Array(-0.50022394, dtype=float64) == Array(-0.50022394, dtype=float64)
```

The results do look the same. The leading theory is it's a floating point error.

Change to `numpy.allclose`.

### DIFF
--- a/frontend/test/pytest/test_capture_integration.py
+++ b/frontend/test/pytest/test_capture_integration.py
@@ -16,6 +16,7 @@
 
 from functools import partial
 
+import numpy as np
 import jax.numpy as jnp
 import pennylane as qml
 import pytest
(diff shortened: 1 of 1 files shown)


sengthai (commented):
should we use `allclose` with other functions test in that file as well?

paul0403:
> should we use `allclose` with other functions test in that file as well?

I am checking just one test for wheels CI (which is slow). If this actually fixes it I will update the others.

sengthai (approved):

erick-xanadu (approved):

PR #1850, opened 2025-07-07
Adjoint and ctrl transform conversion in `from_plxpr`

albi3ro:
**Context:**

We are pushing to make plxpr as a new valid frontend for catalyst. So new need feature parity.

Add conversion support for adjoint and ctrl transforms.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-93191]


erick-xanadu:
@mehrdad2m I believe this could be merged to the RC.

dime10:
Thanks for working on this, let us know when you want to get a review :)

frontend/catalyst/from_plxpr/control_flow.py
```diff
@@ -251,3 +252,58 @@ def remove_qreg(*args_plus_qreg):
...
+# pylint: disable=unused-argument
+@PLxPRToQuantumJaxprInterpreter.register_primitive(plxpr_adjoint_transform_prim)
+def handle_adjoint_transform(
+    self,
+    *plxpr_invals,
+    jaxpr,
+    lazy,
+    n_consts,
+):
+    """Handle the conversion from plxpr to Catalyst jaxpr for the for loop primitive"""
```
dime10:
```suggestion
    """Handle the conversion from plxpr to Catalyst jaxpr for the adjoint primitive"""
```

frontend/catalyst/from_plxpr/control_flow.py
dime10, line 1: Is adjoint considered control flow? (regarding the chosen module)

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -564,6 +573,21 @@ def handle_measure_in_basis(self, angle, wire, plane, reset, postselect):
...
 
+# pylint: disable=unused-argument
+@PLxPRToQuantumJaxprInterpreter.register_primitive(plxpr_ctrl_transform_prim)
+def handle_ctrl_transform(self, *invals, jaxpr, n_control, control_values, work_wires, n_consts):
+    """Interpret a control transform primitive."""
+    consts = invals[:n_consts]
+    args = invals[n_consts:-n_control]
+    control_wires = invals[-n_control:]
+
+    unroller = copy(self)
```
dime10, line 590: What is the self object / where is its code located?

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -564,6 +573,21 @@ def handle_measure_in_basis(self, angle, wire, plane, reset, postselect):
...
+    unroller.control_wires += tuple(control_wires)
+    unroller.control_values += tuple(control_values)
+    unroller.eval(jaxpr, consts, *args)
```
dime10, line 593: Is this only responsible for resolving the control transform or are there any other changes?

frontend/catalyst/from_plxpr/from_plxpr.py
paul0403: `self` is `PLxPRToQuantumJaxprInterpreter`

paul0403: located in the same file above

frontend/catalyst/from_plxpr/control_flow.py
albi3ro: Placed it there because it follows the same general subroutine structure. Made it easier to compare with the other subroutine-like structures.

frontend/catalyst/from_plxpr/from_plxpr.py
albi3ro: Yep.  The structure for these registrations follows what it would look like as a  method.

frontend/catalyst/from_plxpr/from_plxpr.py
albi3ro: Yep. This executes the subroutine, and just adds the control wires and control values to every operation encountered in the subroutine. When the full execution is traced, the subroutine no longer occurs, just the operations in the subroutine with control wires and control values.

frontend/catalyst/from_plxpr/control_flow.py
dime10: Maybe the module should be renamed then? I would not expect to find adjoint in it

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -320,6 +327,8 @@ def interpret_operation(self, op, is_adjoint=False, control_values=(), control_w
             )
 
         in_qubits = [self.qreg_manager[w] for w in op.wires]
```
erick-xanadu: I think you'll need to merge again. There's a small change that I believe needs to happen here:

```
self.qreg_manager.insert_dynamic_qubits(op.wires + op.control_wires)

```

frontend/catalyst/from_plxpr/control_flow.py
albi3ro: Now that the structure's been worked out, I'll move it back.

paul0403: Yes I think adjoint can be in its own file; a reasonable name for this one would be `meta_ops.py` (ops with jaxprs as params) but that in my opinion is less sensible than `control_flow.py` and `adjoint.py` separately : )

frontend/catalyst/from_plxpr/from_plxpr.py
albi3ro: Updated the branch.

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -322,6 +333,8 @@ def interpret_operation(self, op, is_adjoint=False, control_values=(), control_w
         self.qreg_manager.insert_dynamic_qubits(op.wires)
```
erick-xanadu:
```suggestion
        self.qreg_manager.insert_dynamic_qubits(op.wires + control_wires)
```

erick-xanadu: I'm a little confused by `self.control_wires`. :thinking: not sure if they should be added here.

erick-xanadu: I'll dedicate some time to think about this after lunch, but happy to hear what you think. If you want we can chat about this offline.

albi3ro: potentially do this before and after the execution and trust that control wires don't overlap with operator wires?

erick-xanadu: Yes, if control wires overlap with operator wires that's a runtime error that we can't control. (This is true for any overlapping wires in any operation, not special to control_wires). Give me a little bit of time to think today. Can I get back to you on this after 2:30?

albi3ro: 👍

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -293,13 +299,18 @@ class PLxPRToQuantumJaxprInterpreter(PlxprInterpreter):
...
         self.device = device
         self.shots = shots
         # TODO: we assume the qreg value passed into a scope is the unique qreg in the scope
         # In other words, we assume no new qreg will be allocated in the scope
         self.qreg_manager = qreg_manager
         self.subroutine_cache = cache
+        self.control_wires = control_wires
+        """Any control wires used for a subroutine."""
+        self.control_values = control_values
+        """Any control values for executing a subroutine."""
```
erick-xanadu, line 312: I believe this implies we need to pass in `self.control_wires` and `self.control_values` to all constructors of `PLxPRToQuantumJaxprInterpreter`.

I tried the following example:

```
def test_adjoint_inside_ctrl():

    qml.capture.enable()

    @qml.qjit
    @qml.qnode(qml.device("null.qubit", wires=2), autograph=False)
    def c():
        qml.ctrl(qml.adjoint(qml.X), 0, control_values=(True))(wires=1)
        return qml.probs()

    print(c.mlir)
    qml.capture.disable()

```

and the control_wires and control_values get removed from the IR:

```
      quantum.device shots(%c0_i64) ["/home/ubuntu/Code/catalyst2/frontend/catalyst/utils/../../../runtime/build/lib/librtd_null_qubit.so", "NullQubit", "{'track_resources': False}"]
      %1 = quantum.adjoint(%0) : !quantum.reg {
      ^bb0(%arg0: !quantum.reg):
        %4 = quantum.extract %arg0[ 1] : !quantum.reg -> !quantum.bit
        %out_qubits = quantum.custom "PauliX"() %4 : !quantum.bit
        %5 = quantum.insert %arg0[ 1], %out_qubits : !quantum.reg, !quantum.bit
        quantum.yield %5 : !quantum.reg
      }
      %2 = quantum.compbasis qreg %1 : !quantum.obs
      %3 = quantum.probs %2 : tensor<4xf64>
      quantum.dealloc %1 : !quantum.reg
      quantum.device_release
      return %3 : tensor<4xf64>
    }
```

I would also recommend to remove the use of `copy.copy` just to be able to jump easily to any new instantiation of the class.

albi3ro: Ah... you're right. Maybe we could add an `update` method to indicate "give me a new instance with same properties except these ones".

frontend/catalyst/from_plxpr/from_plxpr.py
erick-xanadu: Thanks for giving me enough time! I was able to reason about this a little bit longer. I believe it would be best to have:

```python
        self.qreg_manager.insert_dynamic_qubits(op.wires + control_wires + self.control_wires)
```

This is just to be more careful and I believe it should also match what Catalyst does.

frontend/catalyst/from_plxpr/from_plxpr.py
erick-xanadu: :thinking: I prefer to use the constructor and pass everything needed, but happy to let others decide on the style here. I understand it is a lot more verbose but it would make readability easier. (For example, I could jump between constructors easily in my editor to find out where we are creating new instances).

albi3ro: I find using `copy` better, since then we *don't* have to go back and find everywhere another instance or child class might be initialized.

For example in the above case, if I use `copy` when executing the adjoint transform, this would have worked with no modifications.

frontend/catalyst/from_plxpr/from_plxpr.py
albi3ro: handled and tested.

frontend/catalyst/from_plxpr/control_flow.py
```diff
@@ -58,13 +57,10 @@ def handle_cond(self, *plxpr_invals, jaxpr_branches, consts_slices, args_slice):
...
             def calling_convention(*args_plus_qreg):
                 *args, qreg = args_plus_qreg
-                device = self.device
-                shots = self.shots
                 # `qreg` is the scope argument for the body jaxpr
                 qreg_manager = QregManager(qreg)
-                converter = PLxPRToQuantumJaxprInterpreter(
-                    device, shots, qreg_manager, self.subroutine_cache
-                )
+                converter = copy(self)
```
paul0403, line 62: This seems like a drastic change, we are no longer launching a brand new interpreter? Can you elaborate a bit?

albi3ro: We are creating a brand new interpreter, just by copying the original configuration. This makes it easier to keep the same top-level configuration in all the child interpreters.  We only have to update the `qreg_manager` that will differ between each call.

Makes is easier to keep the correct configuration and arguments.

We just now *automatically* get the same device, shots, subroutine cache, control wires, and control values, instead of manually needing to provide them.

paul0403: I'm a bit afraid that this is not a deepcopy :thinking: Do you know if this is a deep copy? Or would it matter if this were a shallow copy?

I guess tests pass so that's a good sign, but I'm still a bit worried. But it could just be me being paranoid haha

albi3ro: Right now it's a shallow copy. We could switch to a deep copy, but shouldn't we want to use the same subroutine cache?

All the other properties are immutable, so it wouldn't make a difference.

paul0403: Ah yes, you are right! Thanks for the clarification :rocket:

paul0403 (approved):
:partying_face:

Issue #1838, opened 2025-06-27
Cancel inverses pass gives incorrect results with non-autographed python control flow


paul0403:
```python
import pennylane as qml
import catalyst
from catalyst import qjit

@qjit
def workflow():

    @qml.qnode(qml.device("lightning.qubit", wires=2))
    def circuit_raw():
        qml.Hadamard(wires=0)
        for i in range(2):
            qml.Hadamard(wires=0)
            qml.Hadamard(wires=1)
            qml.Hadamard(wires=0)
            qml.Hadamard(wires=1)
        return qml.probs()  # should be 50/50: the loop has two cancel pairs, and there's one Hadamard outside

    raw = circuit_raw()
    cancelled = catalyst.passes.cancel_inverses(circuit_raw)()

    return raw, cancelled

print(workflow())

```

```
(Array([0.5, 0. , 0.5, 0. ], dtype=float64), Array([1., 0., 0., 0.], dtype=float64))
```

Versions are the previous release:
```
(stable_env) paul.wang@L7450-H298SW3:~/catalyst_new/catalyst$ which-pennies 
amazon-braket-pennylane-plugin    1.31.3
PennyLane                         0.41.1
PennyLane-Catalyst                0.11.0
PennyLane_Lightning               0.41.1
PennyLane_Lightning_Kokkos        0.41.1
jax                               0.4.28
jaxlib                            0.4.28
```

PR #1619, opened 2025-04-07
Add resource tracking to NullQubit

jzaia18:
**Context:**
For benchmarking, we need a method of precise resource tracking. This method has to be compatible with code run directly from PennyLane and should be able to track precise resources after various transformation passes have already occurred.

See also: https://github.com/PennyLaneAI/pennylane-benchmarks/pull/76 and https://github.com/PennyLaneAI/pennylane/pull/7226

**Description of the Change:**
Adds 2 new class attributes to `NullQubit`, as well as an optional argument to its constructor. When this option is toggled on, `NullQubit` will track precise gate counts while "executing" a circuit and save them to a file when all qubits are released.

**Benefits:**
It is now possible to track precise gate counts (with or without gate decomposition or other transformation passes) for measuring circuit complexity.

**Possible Drawbacks:**
This implementation currently write the resources JSON to a file on disk as opposed to neatly returning an object back through to the Python caller in PennyLane. In the future, a more robust solution may be desired.

**Related GitHub Issues:**
[sc-91366]

PR #1859, opened 2025-07-07
v0.12.0 changelog

isaacdevlugt:
**Context:** Logging changes

**Description of the Change:** Organizes changelog

**Benefits:** Changes are logged

**Possible Drawbacks:** 🫥 

**Related GitHub Issues:** [sc-89369](https://app.shortcut.com/xanaduai/story/89369)


doc/releases/changelog-0.12.0.md
```diff
@@ -260,107 +327,114 @@
...
 
-* The device kwarg parsing in runtime has been updated to disallow nested dictionary formats.
+* The device kwarg parsing in the Catalyst runtime has been updated to disallow nested dictionary 
+  formats.
   [(#1843)](https://github.com/PennyLaneAI/catalyst/pull/1843)
   [(#1846)](https://github.com/PennyLaneAI/catalyst/pull/1846)
 
 <h3>Deprecations 👋</h3>
 
+🦗... 🦗...
```
isaacdevlugt: Should we mention Python 3.10 support being deprecated? 🤔

doc/releases/changelog-0.12.0.md
```diff
@@ -2,49 +2,86 @@
...
-* Added integration with PennyLane's experimental python compiler based on xDSL.
-  This allows developers and users to write xDSL transformations that can be used with Catalyst.
-  [(#1715)](https://github.com/PennyLaneAI/catalyst/pull/1715)
-
-* A new compilation pass called :func:`~.passes.ppr_to_ppm` has been added to Catalyst
-  to decompose Pauli product rotations (PPRs), :math:`\exp(-iP_{\{x, y, z\}} \theta)`, into
-  Pauli product measurements (PPMs). Non-Clifford PPR (:math:`\theta = \tfrac{\pi}{8}`) requires
-  the consumption of a magic state, while Clifford PPR (:math:`\theta = \tfrac{\pi}{4}`) will not.
-  The methods are implemented as described in [arXiv:1808.02892](https://arxiv.org/abs/1808.02892v3).
+* A new compilation pass called :func:`~.passes.ppm_compilation` has been added to Catalyst to
```
dime10, line 5: Should we mention something about this being somewhat experimental, in progress, or mainly for internal use? The fact that you can't execute the result and also need to use a special compiler pipeline to run these passes somehow doesn't make it feel like a full-fledged feature.

dime10: At least we might want to consider mentioning that these are non-executable programs and are meant to explore compilation.

doc/releases/changelog-0.12.0.md
mehrdad2m: I think it makes sense to mention that we are deprecating Python 3.10 somewhere and this can be a good place for it. Are we doing the same thing on Pennylane changelog as well? 🤔

doc/releases/changelog-0.12.0.md
```diff
@@ -2,49 +2,86 @@
...
+  %m_14, %out_15:2 = qec.ppm ["Z", "Z"] %6#0, %10 : !quantum.bit, !quantum.bit
+  %from_elements = tensor.from_elements %m_14 : tensor<i1>
+  %m_16, %out_17 = qec.ppm ["Z"] %out_15#1 : !quantum.bit
+  ...
+  ```
 
-  * A new function :func:`~.passes.get_ppm_specs` to get the result statistics after a PPR/PPM compilations is available. The statistics is returned in a Python dictionary.
+* A new function called :func:`~.passes.get_ppm_specs` has been added for acquiring 
+  statistics after PPM compilation.
   [(#1794)](https://github.com/PennyLaneAI/catalyst/pull/1794).
```
AntonNI8:
```suggestion
  statistics after PPM compilation.
  [(#1794)](https://github.com/PennyLaneAI/catalyst/pull/1794)
```

doc/releases/changelog-0.12.0.md
```diff
@@ -114,23 +150,17 @@
   print(results)
```
dime10: Might I suggest a slightly more convenient option:
```suggestion
  snapshots, *results = circuit()
  
  >>> print(snapshots)
  [Array([1.+0.j, 0.+0.j, 0.+0.j, 0.+0.j], dtype=complex128), 
  Array([0.5+0.j, 0.5+0.j, 0.5+0.j, 0.5+0.j], dtype=complex128)]
  >>> print(results)
  Array([0.25, 0.25, 0.25, 0.25], dtype=float64)
```

This will make it clear how the returned values are structured, and printing the snapshots separately from the results should also be easier to read.

doc/releases/changelog-0.12.0.md
isaacdevlugt: Yep!

doc/releases/changelog-0.12.0.md
```diff
@@ -148,30 +178,57 @@
   [0. 1. 0. 0. 0. 0. 0. 0.]
   ```
```
dime10, line 188: I think it's important to mention the caveat here.
````suggestion
  ```

  While this feature adds a lot of convenience, it may also reduce performance on devices where
  reallocating resources can be expensive, such as statevector simulators.
````

File: doc/releases/changelog-0.12.0.md [hunk 6]
```diff
@@ -148,30 +178,57 @@
...
 
-  In this example, the number of wires is not specified at device initialization.
-  When we encounter an X gate on `wires=2`, catalyst automatically expands the size
-  of the qubit register to include the requested wire index.
-  Here, the register will contain (at least) 3 qubits after the X operation.
-  As a result, we can see the QNode returning the probabilities for the state |001>,
-  meaning 3 wires were allocated in total.
+* Two new peephole-optimization compilation passes called :func:`~.passes.disentangle_cnot` and 
+  :func:`~.passes.disentangle_swap` have been added. Each compilation pass replaces `SWAP` or `CNOT`
+  instructions at the MLIR level with other equivalent elementary gates.
```
dime10: Why is the MLIR level relevant here? 

That's what Catalyst is built on so I don't think it needs to be specified every time something happens in mlir.

doc/releases/changelog-0.12.0.md
```diff
@@ -148,30 +178,57 @@
...
-  [(#1817)](https://github.com/PennyLaneAI/catalyst/pull/1817)
+* The :func:`qml.measure <pennylane.measure>` operation for mid-circuit measurements can now be used 
+  in qjit-compiled circuits with program capture enabled.
+  [(#1766)](https://github.com/PennyLaneAI/catalyst/pull/1766)
+
+  Note that using :func:`qml.measure <pennylane.measure>` in this way binds the operation to 
+  :func:`catalyst.measure`, which behaves differently than :func:`qml.measure <pennylane.measure>` 
+  in a native PennyLane circuit, as described in the *Functionality differences from PennyLane* 
+  section of the :doc:`sharp bits and debugging tips <sharp_bits>` guide. In regular qjit-compiled 
+  workflows (without program capture enabled), you must continue to use :func:`catalyst.measure`.
```
dime10: I don't particularly like this framing, as the differing behaviour is dependant on the simulation method chosen and not which measure op the user uses. The sharp bits should probably be updated to include changes we've made since it was written, like the addition of the one-shot method which does behave the same as pennylane.

doc/releases/changelog-0.12.0.md
```diff
@@ -180,34 +237,40 @@
...
+  of Pauli gates (`qml.X`, `qml.Y`, `qml.Z`), the adjoint of `qml.S` gate, and the adjoint of the 
+  `qml.T` gate.
   [(#1738)](https://github.com/PennyLaneAI/catalyst/pull/1738)
 
 * The `keep_intermediate` argument in the `qjit` decorator now accepts a new value that allows for
   saving intermediate files after each pass. The updated possible options for this argument are:
-  * `False` or `0` or `"none"` or `None` : No intermediate files are kept.
-  * `True` or `1` or `"pipeline"`: Intermediate files are saved after each pipeline.
-  * `2` or `"pass"`: Intermediate files are saved after each pass.
+  - `False` or `0` or `"none"` or `None` : No intermediate files are kept.
```
dime10: Maybe it's not necessary to support both "none" and None? Seems a bit redundant and having 4 options to the same thing could be a bit confusing.

If I had to choose I would probably remove the "none" string option 🤔

doc/releases/changelog-0.12.0.md
```diff
@@ -180,34 +237,40 @@
...
-* Two new Python pass decorators, :func:`~.passes.disentangle_cnot` and 
-  :func:`~.passes.disentangle_swap` have been added. They apply their corresponding MLIR passes 
-  `--disentangle-CNOT` and `--disentangle-SWAP` respectively. 
-  [(#1823)](https://github.com/PennyLaneAI/catalyst/pull/1823)
+* Catalyst is compatible with the new :func:`qml.set_shots <pennylane.set_shots>` transform 
+  introduced in PennyLane v0.42.
+  [(#1784)](https://github.com/PennyLaneAI/catalyst/pull/1784)
+
+* `null.qubit` can now support an optional `track_resources` argument, which allows it to record 
+  which gates are executed.
```
dime10: Should this not contain an example? I know many users are interested in this, so it might even be good to push the visibility of this further.

doc/releases/changelog-0.12.0.md
```diff
@@ -180,34 +237,40 @@
...
+  [(#1619)](https://github.com/PennyLaneAI/catalyst/pull/1619)
 
 <h3>Breaking changes 💔</h3>
 
-* (Device Developers Only) The `QuantumDevice` interface in the Catalyst Runtime plugin system
-  has been modified, which requires recompiling plugins for binary compatibility.
+* The `QuantumDevice` interface in the Catalyst Runtime plugin system has been modified, which
```
dime10: Please don't remove the tags I added, they are important to inform which kind of user is affected by a breaking change. This change for example is not breaking for regular users of the package, but it is breaking for device integrators. Same with the changes that would only affect someone integrating at the frontend, rather than regular users.

doc/releases/changelog-0.12.0.md
```diff
@@ -258,112 +322,120 @@
...
-  The mlir-hlo version is updated to [commit 617a9361d186199480c080c9e8c474a5e30c22d1](https://github.com/tensorflow/mlir-hlo/tree/617a9361d186199480c080c9e8c474a5e30c22d1).
-  The Enzyme version is updated to [v0.0.180](https://github.com/EnzymeAD/Enzyme/releases/tag/v0.0.180).
+  The LLVM version has been updated to 
+  [commit a8513158](https://github.com/llvm/llvm-project/tree/179d30f8c3fddd3c85056fd2b8e877a4a8513158).
+  The mlir-hlo version has been updated to 
+  [commit e30c22d1](https://github.com/tensorflow/mlir-hlo/tree/617a9361d186199480c080c9e8c474a5e30c22d1).
+  The Enzyme version has been updated to 
+  [v0.0.180](https://github.com/EnzymeAD/Enzyme/releases/tag/v0.0.180).
 
 * The clang-format and clang-tidy versions used by Catalyst have been updated to v20.
```
dime10: This is not a breaking change, at best this could be classified as an internal change (but we can probably omit it entirely).

doc/releases/changelog-0.12.0.md
```diff
@@ -258,112 +322,120 @@
...
   with the rationale being that such machines are becoming increasingly scarce.
 
-  If support for Mac x86 platforms is still desired, please install
-  Catalyst version 0.11.0, PennyLane version 0.41.0, PennyLane-Lightning
-  version 0.41.0, and Jax version 0.4.28.
+  If support for Mac x86 platforms is still desired, please install Catalyst v0.11.0, PennyLane 
+  v0.41.0, PennyLane-Lightning v0.41.0, and JAX v0.4.28.
 
-* Sphix version has been updated to 8.1. Some other related packages have been updated as well.
+* The Sphinx version has been updated to v8.1.
```
dime10: Same here, not a breaking change, I don't even think this affected any of the distributed code, in which case it shouldn't appear in the changelog at all.

doc/releases/changelog-0.12.0.md
dime10: Might as well since we know we are dropping it next release :)

doc/releases/changelog-0.12.0.md
```diff
@@ -258,112 +322,120 @@
...
+  [commit a8513158](https://github.com/llvm/llvm-project/tree/179d30f8c3fddd3c85056fd2b8e877a4a8513158).
+  The mlir-hlo version has been updated to 
+  [commit e30c22d1](https://github.com/tensorflow/mlir-hlo/tree/617a9361d186199480c080c9e8c474a5e30c22d1).
+  The Enzyme version has been updated to 
+  [v0.0.180](https://github.com/EnzymeAD/Enzyme/releases/tag/v0.0.180).
 
 * The clang-format and clang-tidy versions used by Catalyst have been updated to v20.
   [(#1721)](https://github.com/PennyLaneAI/catalyst/pull/1721)
 
 * Support for Mac x86 has been removed. This includes Macs running on Intel processors.
```
dime10: This seems quite important, we might want to push it further up for more visibility.

doc/releases/changelog-0.12.0.md
```diff
@@ -258,112 +322,120 @@
...
+  If support for Mac x86 platforms is still desired, please install Catalyst v0.11.0, PennyLane 
+  v0.41.0, PennyLane-Lightning v0.41.0, and JAX v0.4.28.
 
-* Sphix version has been updated to 8.1. Some other related packages have been updated as well.
+* The Sphinx version has been updated to v8.1.
   [(#1734)](https://github.com/PennyLaneAI/catalyst/pull/1734)
 
-* The device kwarg parsing in runtime has been updated to disallow nested dictionary formats.
+* The device kwarg parsing in the Catalyst runtime has been updated to disallow nested dictionary 
+  formats.
```
dime10:
```suggestion
* (Device developers only) Device parameters which are forwarded by the Catalyst runtime
 to plugin devices as a string may not contain nested dictionaries. Previously, these would
 be parsed incorrectly, and instead will now raise an error.
```

doc/releases/changelog-0.12.0.md
```diff
@@ -258,112 +322,120 @@
...
 
-* Fixes the conversion of PLxPR to JAXPR with quantum primitives when using control flow.
+* Fixed the conversion of PLxPR to JAXPR with quantum primitives when using control flow.
   [(#1809)](https://github.com/PennyLaneAI/catalyst/pull/1809)
 
-* Fixes canonicalization of insertion and extraction into quantum registers.
+* Fixed a bug that occurred when a qjit'd workflow contained Pythonic control flow and Autograph was 
+  turned off, leading to compilation passes being applied incorrectly and giving incorrect end 
+  results.
   [(#1840)](https://github.com/PennyLaneAI/catalyst/pull/1840)
```
dime10: This doesn't sound right, the message and PR are now mismatched.

doc/releases/changelog-0.12.0.md
```diff
@@ -380,95 +452,96 @@
...
+  interface in MLIR. The new MLIR bufferization interface is required by JAX v0.4.29 or higher.
   [(#1027)](https://github.com/PennyLaneAI/catalyst/pull/1027)
   [(#1686)](https://github.com/PennyLaneAI/catalyst/pull/1686)
   [(#1708)](https://github.com/PennyLaneAI/catalyst/pull/1708)
   [(#1740)](https://github.com/PennyLaneAI/catalyst/pull/1740)
   [(#1751)](https://github.com/PennyLaneAI/catalyst/pull/1751)
   [(#1769)](https://github.com/PennyLaneAI/catalyst/pull/1769)
 
-* Redundant `OptionalAttr` is removed from `adjoint` argument in `QuantumOps.td` TableGen file
+* The package name of the Catalyst distribution has been updated to be consistent with
```
dime10: Hmm I don't think this is internal. It is very public in the sense that official package name has changed. However, it shouldn't affect users in most common interactions.

doc/releases/changelog-0.12.0.md
isaacdevlugt: > At least we might want to consider mentioning that these are non-executable programs and are meant to explore compilation.

This is a good option!

doc/releases/changelog-0.12.0.md
```diff
@@ -2,49 +2,86 @@
...
-  This allows developers and users to write xDSL transformations that can be used with Catalyst.
-  [(#1715)](https://github.com/PennyLaneAI/catalyst/pull/1715)
-
-* A new compilation pass called :func:`~.passes.ppr_to_ppm` has been added to Catalyst
-  to decompose Pauli product rotations (PPRs), :math:`\exp(-iP_{\{x, y, z\}} \theta)`, into
-  Pauli product measurements (PPMs). Non-Clifford PPR (:math:`\theta = \tfrac{\pi}{8}`) requires
-  the consumption of a magic state, while Clifford PPR (:math:`\theta = \tfrac{\pi}{4}`) will not.
-  The methods are implemented as described in [arXiv:1808.02892](https://arxiv.org/abs/1808.02892v3).
+* A new compilation pass called :func:`~.passes.ppm_compilation` has been added to Catalyst to
+  transform Clifford+T gates into Pauli Product Measurements (PPMs) using just one transform.
```
isaacdevlugt:
```suggestion
  transform Clifford+T gates into Pauli Product Measurements (PPMs) using just one transform, allowing for 
  exploring representations of programs in a new paradigm in logical quantum compilation.
```

@dime10 how's that?

doc/releases/changelog-0.12.0.md
isaacdevlugt: Fair point!

```suggestion
  instructions with other equivalent elementary gates.
```

doc/releases/changelog-0.12.0.md
isaacdevlugt:
```suggestion
  - `False` or `0` or `None` : No intermediate files are kept.
```

doc/releases/changelog-0.12.0.md
isaacdevlugt: 🫡

isaacdevlugt:
```suggestion
* (Device Developers Only) The `QuantumDevice` interface in the Catalyst Runtime plugin system has been modified, which
```

doc/releases/changelog-0.12.0.md
```diff
@@ -221,24 +284,25 @@
   Finally, the `PrintState` and the `One`/`Zero` utility functions have been removed, since they
   did not serve a convincing purpose.
 
-* (Frontend Developers Only) Some Catalyst primitives for JAX have been renamed, and the qubit
-  deallocation primitive has been split into deallocation and a separate device release primitive.
+* Some Catalyst primitives for JAX have been renamed, and the qubit deallocation primitive has been 
+  split into deallocation and a separate device release primitive.
```
isaacdevlugt:
```suggestion
* (Frontend Developers Only) Some Catalyst primitives for JAX have been renamed, and the qubit deallocation primitive has been 
  split into deallocation and a separate device release primitive.
```

doc/releases/changelog-0.12.0.md
isaacdevlugt: See change

doc/releases/changelog-0.12.0.md
dime10: That's good! Should we also mention the programs compiled this way are non-executable?

doc/releases/changelog-0.12.0.md
isaacdevlugt:
```suggestion
  Note that using :func:`qml.measure <pennylane.measure>` with Catalyst in this way binds the operation to 
  :func:`catalyst.measure`, which may lead to slight differences in behaviour, as described in the *Functionality differences from PennyLane* 
  section of the :doc:`sharp bits and debugging tips <sharp_bits>` guide.
```

@dime10 how's this?

doc/releases/changelog-0.12.0.md
isaacdevlugt: Wait... kinda confused why this change is logged in Catalyst 🤔 @jzaia18 @AntonNI8 do either of you have context for that?

isaacdevlugt: Also, if null qubit support was added to Catalyst this release, which PR / changelog entry recorded the change? 🤔

dime10: There was a PR in catalyst that added resource tracking, null qubit itself was already supported.
https://github.com/PennyLaneAI/catalyst/pull/1619

doc/releases/changelog-0.12.0.md
isaacdevlugt: That PR fixed this issue: https://github.com/PennyLaneAI/catalyst/issues/1838. Does what I put here not describe that issue being fixed?

doc/releases/changelog-0.12.0.md
isaacdevlugt: Pretty visible yes, but no real impact on users whatsoever right? I found that it was best suited for internal changes because of that

doc/releases/changelog-0.12.0.md
dime10: Hmm the PR is called "Fixes up chained self inverses", and the example in there contains no control flow 🤔

dime10: It's possible the original issue mischaracterized the problem. The current description talks about control flow and autograph, which I don't think are relevant to issue. It is also a bit vague, as it never says what the problem was.

Erick described the change as "Fixes canonicalization of insertion and extraction into quantum registers.". Maybe the following is more accurate and informative:

```
* Fixed a bug in the internal simplification of qubit chains in the compiler, which manifested
  after certain transformations like `cancel_inverses ` and lead to incorrect program results.
```

doc/releases/changelog-0.12.0.md
dime10: It's hard to predict what no "real impact" means. The installed name in a `pip list` command would be different for example, if you were grepping for `PennyLane-Catalyst` without that `-i` flag your command could now be broken (actually even with the `-i` flag it would be broken because we changed `-` to `_`).

Maybe that sounds a bit contrived, but I don't think it's that far out there. Imo this is definitely a user-facing change, unlike internal changes which should be completely invisible.

doc/releases/changelog-0.12.0.md
isaacdevlugt: Thanks! 🙌

doc/releases/changelog-0.12.0.md
isaacdevlugt: Fair! Can't hurt to err on the side of more transparency. I'll move it to Improvements :)

doc/releases/changelog-0.12.0.md
isaacdevlugt: Oh! It might just have been a gap in coverage on the supported devices page https://docs.pennylane.ai/projects/catalyst/en/stable/dev/devices.html

Stable doesn't show NQ, but Latest does

isaacdevlugt: In any case, @dime10 this kwarg is experimental and subject to change. So, not sure how much we want to highlight it. @AntonNI8 do you have more context? Is this something we want to encourage users to use, or is this a temporary stop-gap for the new `specs` features we're adding in v0.43?

doc/releases/changelog-0.12.0.md
dime10: May I suggest the following:
```
  Note that the simulation behaviour of mid-circuit measurements can differ between PennyLane and
  Catalyst, depending on the chosen `mcm_method` QNode parameter. Please see the
  :doc:`sharp bits<sharp_bits#functionality-differences-from-pennylane>` for additional information.
```
(link should also go straight to the section, not 100% sure how to achieve this)

doc/releases/changelog-0.12.0.md
dime10: I know it's "experimental", but it's also something that is incredibly useful and external collaborators have asked for. I'll leave it up you then to decide how to highlight this.

AntonNI8: @isaacdevlugt @dime10 I would recommend including this as an internal change rather than an improvement.
Reason being a) that it is subject to change and b) that it's not the way we would/will generally encourage users to go about resource tracking - we have planned work in Q3 to improve `specs` and resource tracking generally which should make it more accessible and performant.
Since there is already interest in the feature (as you said @dime10), I think it wouldn't hurt to include an example, although I wouldn't say it's strictly necessary.

doc/releases/changelog-0.12.0.md
isaacdevlugt: I think we'd need to add a label along the lines of what is presented here? https://stackoverflow.com/questions/15394347/adding-a-cross-reference-to-a-subheading-or-anchor-in-another-page

isaacdevlugt:
```suggestion
  Note that the simulation behaviour of mid-circuit measurements can differ between PennyLane and
  Catalyst, depending on the chosen `mcm_method`. Please see the 
  *Functionality differences from PennyLane* section in the
  :doc:`sharp bits and debugging tips page <sharp_bits>` for additional information.
```

doc/releases/changelog-0.12.0.md
```diff
@@ -148,66 +185,117 @@
...
-  :func:`~.passes.disentangle_swap` have been added. They apply their corresponding MLIR passes 
-  `--disentangle-CNOT` and `--disentangle-SWAP` respectively. 
-  [(#1823)](https://github.com/PennyLaneAI/catalyst/pull/1823)
+* Catalyst is compatible with the new :func:`qml.set_shots <pennylane.set_shots>` transform 
+  introduced in PennyLane v0.42.
+  [(#1784)](https://github.com/PennyLaneAI/catalyst/pull/1784)
+
+* `null.qubit` can now support an optional `track_resources` keyword argument, which allows it to record 
+  which gates are executed. 
+  [(#1619)](https://github.com/PennyLaneAI/catalyst/pull/1619)
```
isaacdevlugt, line 281: Just adding a comment here so that it's not hidden due to it being outdated (https://github.com/PennyLaneAI/catalyst/pull/1859#discussion_r2200843519)

> @isaacdevlugt @dime10 I would recommend including this as an internal change rather than an improvement.
> Reason being a) that it is subject to change and b) that it's not the way we would/will generally encourage users to go about resource tracking - we have planned work in Q3 to improve specs and resource tracking generally which should make it more accessible and performant.
> Since there is already interest in the feature (as you said @dime10), I think it wouldn't hurt to include an example, although I wouldn't say it's strictly necessary.

isaacdevlugt: For context, we didn't add a code example for this feature in the PL changelog 😅

AntonNI8: @dime10 would you have a moment to write up a minimal code example that highlights this functionality in the way you find most useful? If so, we could add it to these release notes.

doc/releases/changelog-0.12.0.md
```diff
@@ -118,26 +150,26 @@
...
   ```pycon
   >>> print(results)
   ([Array([1.+0.j, 0.+0.j, 0.+0.j, 0.+0.j], dtype=complex128), 
   Array([0.5+0.j, 0.5+0.j, 0.5+0.j, 0.5+0.j], dtype=complex128)], 
   ```
 
-* Catalyst now supports automatic qubit management.
+* Catalyst now supports automatic qubit management, meaning that the number of wires does not need 
+  to be speficied during device initialization.
```
DSGuala:
```suggestion
  to be specified during device initialization.
```

doc/releases/changelog-0.12.0.md
```diff
@@ -265,112 +349,110 @@
...
 * MLIR plugins can now be specified via lists and tuples, not just sets.
   [(#1812)](https://github.com/PennyLaneAI/catalyst/pull/1812)
 
-* Fixes the conversion of PLxPR to JAXPR with quantum primitives when using control flow.
+* Fixed the conversion of PLxPR to JAXPR with quantum primitives when using control flow.
   [(#1809)](https://github.com/PennyLaneAI/catalyst/pull/1809)
 
-* Fixes canonicalization of insertion and extraction into quantum registers.
+* Fixed a bug in the internal simplification of qubit chains in the compiler, which manifested in 
+  certain transformations like `cancel_inverses ` and lead to incorrect results.
```
DSGuala:
```suggestion
  certain transformations like `cancel_inverses ` and led to incorrect results.
```

DSGuala (approved):
Looks good to me. Approved pending a suggestions about typos and the discussion about moving  #1619 to internal changes.

AntonNI8 (approved):
Approved pending final changes regarding resource tracking ✅

dime10 (approved):

PR #1861, opened 2025-07-07
Clean up resources files litter

jzaia18:
**Context:**
The resource tracking feature added in #1619 has a unit test which leaves files in the `runtime/` directory.

Ensures all files are cleaned up after unit test execution. Allows the NullQubit device to track the filename it has written to.

**Benefits:**
Clean workspace, no file litter

**Possible Drawbacks:**
More complex unit tests

**Related GitHub Issues:**
[sc-94685]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -23,6 +23,9 @@
 
 <h3>Improvements 🛠</h3>
 
+* Fix resource tracking unit test polluting the environment with output files
+  [(#1861)](https://github.com/PennyLaneAI/catalyst/pull/1861)
+
 * Adjoint differentiation is used by default when executing on lightning devices, significantly reduces gradient computation time.
   [(#1961)](https://github.com/PennyLaneAI/catalyst/pull/1961)
 
@@ -84,9 +87,6 @@
   [(#1952)](https://github.com/PennyLaneAI/catalyst/pull/1952)
 
 <h3>Bug fixes 🐛</h3>
-
-* Fix resource tracking unit test polluting the environment with output files
-  [(#1861)](https://github.com/PennyLaneAI/catalyst/pull/1861)
 
 * Fix type promotion on conditional branches, where the return values from `cond` should be the promoted one.
   [(#1977)](https://github.com/PennyLaneAI/catalyst/pull/1977)
--- a/runtime/tests/Test_NullQubit.cpp
+++ b/runtime/tests/Test_NullQubit.cpp
@@ -867,9 +867,12 @@ TEST_CASE("Test NullQubit device resource tracking", "[NullQubit]")
     CHECK(sim->ResourcesGetNumGates() == 0);
     CHECK(sim->ResourcesGetNumQubits() == 0);
     CHECK(sim->ResourcesGetFilename() == RESOURCES_FNAME);
+}
 
+TEST_CASE("Test resource tracking filename", "[NullQubit]")
+{
     // Check automatic filename creation
-    dummy = std::make_unique<NullQubit>("{'track_resources':True}");
+    std::unique_ptr<NullQubit> dummy = std::make_unique<NullQubit>("{'track_resources':True}");
     CHECK(dummy->IsTrackingResources() == true);
     dummy->ReleaseAllQubits();
 


runtime/tests/Test_NullQubit.cpp
```diff
@@ -656,6 +658,7 @@ TEST_CASE("Test NullQubit device resource tracking", "[NullQubit]")
 {
     // The name of the file where the resource usage data is stored
     constexpr char RESOURCES_FNAME[] = "__pennylane_resources_data.json";
+    constexpr char RESOURCES_FNAME_PREFIX[] = "__pennylane_resources_data_";
 
     // Open a file for writing the resources JSON
     FILE *resource_file_w = fopen(RESOURCES_FNAME, "wx");
```
mlxd: Can we wrap this, and all of the cleanup code into an RAII class?
That way, the filename is tracked upon creation, so we don't need to query the filesystem, any failure will ensure correct cleanup in the destructor, and the functionality itself can be tested independent of the cleanup routines.

jzaia18: I can certainly wrap this into a RAII class. But the resources fname in line 660 is different than the one that gets written automatically when `releaseAllQubits()` is called. Since the final approach that we took for the resources filename is to append a timestamp to a prefix, I don't think we can avoid querying the filesystem to locate it unless we want to refactor to allow the user to choose the filename that will be written.

mlxd: Is this information not available to the device though?
As in, shouldn't the device be able to have a handle to said RAII file-wrapped instance that has an associated lifetime and access methods to the data we care about? I think I'd prefer if we can go down this route.

jzaia18: I'm a bit confused, could you clarify the question? Are you saying we should pass a custom object that stores file information at device construction time?

runtime/tests/Test_NullQubit.cpp
```diff
@@ -17,6 +17,8 @@
 #include <catch2/catch_test_macros.hpp>
 #include <catch2/matchers/catch_matchers_string.hpp>
 #include <cstdio>
+#include <filesystem>
```
paul0403: Hey, thanks for adding this. One overall wondering I have is why we are doing this in the test (aka user code) instead of the null qubit device itself.

paul0403: Since I assume the creation of these breadcrumb files is in the device, not by the tests.

jzaia18: Doing which particular thing? If its the `#include`, then its because we need to search the filesystem for the created file to clean it up only in the tests, the device itself doesn't need to do anything special other than writing the file

paul0403: No, I mean the cleanup of the dumped files itself. Should the cleanup happen within the object (I assume the null qubit device object) that produced these files?

jzaia18: No, the files are written as output so that the user (or really in our use-case the benchmark suite) can read the resource usage information.

jzaia18: So we need the file to persist even after the device has been destroyed, it should be a manual cleanup by the user whenever they're finished with the info in the file (sort of like logs).

paul0403 (approved):

runtime/tests/Test_NullQubit.cpp
paul0403: I am also in favor of wrapping the (somewhat complicated) cleanup code into its own utility function/class, as opposed to flat code in the test.

Although my opinions on this are not super strong.

mlxd: > Are you saying we should pass a custom object that stores file information at device construction time?

Yes, exactly this.

runtime/lib/backend/null_qubit/NullQubit.hpp
```diff
@@ -474,13 +479,20 @@ struct NullQubit final : public Catalyst::Runtime::QuantumDevice {
      */
     auto ResourcesGetNumQubits() -> std::size_t { return resource_data_["num_qubits"]; }
 
+    /**
+     * @brief Returns the filename where resource tracking information is dumped. Only works if
+     * resource tracking is enabled
+     */
+    auto ResourcesGetFilename() -> std::string { return resources_fname_; }
```
maliasadi:
```suggestion
     ResourcesGetFilename() const -> std::string { return resources_fname_; }
```

runtime/lib/backend/null_qubit/NullQubit.hpp
```diff
@@ -474,13 +479,20 @@ struct NullQubit final : public Catalyst::Runtime::QuantumDevice {
      */
     auto ResourcesGetNumQubits() -> std::size_t { return resource_data_["num_qubits"]; }
```
maliasadi, line 481:
```suggestion
    auto ResourcesGetNumQubits() const -> std::size_t { return resource_data_["num_qubits"]; }
```

runtime/tests/Test_NullQubit.cpp
```diff
@@ -862,7 +862,9 @@ TEST_CASE("Test NullQubit device resource tracking", "[NullQubit]")
 
     // Check that releasing resets
     sim->ReleaseAllQubits();
+    std::remove(RESOURCES_FNAME); // Remove the file automatically created by the device
```
maliasadi, line 865: Couldn't we implement the resource tracking in a RAII class to automate this part?

runtime/lib/backend/null_qubit/NullQubit.hpp
jzaia18: This one we can't do since `resource_data_` is not const, and accessing it is not a constant operation either.

runtime/tests/Test_NullQubit.cpp
jzaia18: Not for much benefit, in the general case we want to file to persist long after the device is deallocated. It's just a special case for unit testing where we want it to be removed so early. I tested this out a bit this morning, but implementing this with a RAII class introduced a lot of extra complexity for not much benefit.

maliasadi: Yeah, that makes sense!

runtime/lib/backend/null_qubit/NullQubit.hpp
maliasadi: Oops. my bad! I oversaw it

doc/releases/changelog-dev.md
```diff
@@ -85,6 +85,9 @@
 
 <h3>Bug fixes 🐛</h3>
 
+* Fix resource tracking unit test polluting the environment with output files
+  [(#1861)](https://github.com/PennyLaneAI/catalyst/pull/1861)
```
paul0403: Usually we don't have changelog for test changes. Since this PR changes nullqubit itself but is still for test purposes, I don't think it's a bug fix. I would propose to have this under improvements maybe?

runtime/tests/Test_NullQubit.cpp
```diff
@@ -862,7 +862,24 @@ TEST_CASE("Test NullQubit device resource tracking", "[NullQubit]")
...
+
+    const std::string dummy_resources_fname = dummy->ResourcesGetFilename();
+
+    std::cout << "Auto-created resources filename: " << dummy_resources_fname << std::endl;
+
+    CHECK(dummy_resources_fname.rfind("__pennylane_resources_data_") == 0);
+    CHECK(dummy_resources_fname.find(".json") != std::string::npos);
+
+    std::remove(
+        dummy_resources_fname.c_str()); // Remove the file automatically created by the device
```
paul0403: Is this section meant to test the specific case when a file name is not provided? In that case I feel like this could be its own separate test.

paul0403 (approved):
:100:

PR #1872, opened 2025-07-09
make input branch not required in the wheel CI

mehrdad2m:
**Context:**
PR #1625 adds input branch to the wheel builder workflows in order to call them on rc branch for the rc nightly build, but it marks the input branch as required and causes a failure for the regular nightly builds on main since it calls the workflows without any input branch

makes the input branch no required and falls back to `github.ref` when no input branch is specified.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


paul0403 (approved):
:laughing:

Issue #1726, opened 2025-05-07
LAPACKE build on macOS is not compatible with CMake 4


dime10:
Building the runtime on macOS with cmake 4 installed produces the following error:
```
Cloning into 'reference-lapack-src'...
HEAD is now at 77a0ceb6c Merge pull request #382 from vladimir-ch/lapacke/dlantr-no-alloc-trans
CMake Error at /catalyst/runtime/build/_lapacke-accelerate/src/lapacke-accelerate-build/_deps/reference-lapack-src/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
```
The [upstream project](https://github.com/Reference-LAPACK/lapack/blob/77a0ceb6c9a3c757e7039027f1f324e2811e6ee0/CMakeLists.txt#L1) in question uses the following line, which is no longer compatible with cmake 4:
```
cmake_minimum_required(VERSION 3.2)
```

In the meantime, we might have to require CMake < 4 in Catalyst to warn users early about the incompatibility.
Eventually, we'll have to find a way to update the dependency to use a more recent CMake version 🤔

lepus2589:
I see you're directly including my project via `ExternalProject_Add()`. My workaround doesn't really help there, because it requires user intervention (modifying `CMakeUserPresets.json`).

The initial project was not intended for direct inclusion. But I'm happy to support alternative approaches or additional changes, that make direct inclusion more convenient. Let me know what would help the most.

lepus2589:
I did some refactoring to the project and integrated the CMake v4 check into the code. Nothing else should have changed for you as a consumer.

Unfortunately, I can't really do anything about `CMAKE_OSX_SYSROOT`. This is explicitly meant to be configured by the end user on their machine per the [CMake docs](https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_SYSROOT.html).

dime10:
Thank you so much for your help on this @lepus2589!

PR #1876, opened 2025-07-09
Remove constraint on CMake Version<4 and add the support for handling CMAKE_OSX_SYSROOT

rniczh:
**Context:**

When compiling catalyst from scratch. Mac user may encounter `CMAKE_OSX_SYSROOT` didn't set properly when compiling the lapacke. (may reproduce by `make runtime`)

lapacke actually run this regex expression `SDKs/MacOSX(.)\.sdk` to get the version number. Run the command `xcrun --sdk macosx --show-sdk-path` may only get the symlink `MacOSX.sdk -> MacOSX${LATEST_VERSION}.sdk` Instead of `MacOSX${LATEST_VERSION}.sdk`, but it's depend on your machine.

Note:

`CMake<4`:
Everything well, just don't set the `SDKROOT` in shell env then wouldn't be facing any issue (If it is set, CMake would use it to init the `CMAKE_OSX_SYSROOT`) and CMake has its own detection logic to figure out the path

`CMake==4`:
If `CMAKE_OSX_SYSROOT` is empty: CMake wouldn't detect the sdk path now, they leave for the user to set it explicitly.

Add fallback logic to automatically detect the macOS SDK path when `CMAKE_OSX_SYSROOT`
is not explicitly set. 

1. Check if `CMAKE_OSX_SYSROOT` is already set and use its realpath if available
2. Fall back to using the realpath of `xcrun --sdk macosx --show-sdk-path` to automatically
    detect the official macOS SDK path when `CMAKE_OSX_SYSROOT` is not set

**Benefits:**

Support CMake>=4, and don't need the user to set the `SDKROOT` in shell env manually 

**Related GitHub Issues:** #1726

### DIFF
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -93,16 +93,8 @@ if(NOT EXISTS ${CMAKE_BINARY_DIR}/lib/liblapacke.3.dylib)
             # get the realpath of the SDK path
             get_filename_component(OSX_SDK_PATH ${XCRUN_SDK_PATH} REALPATH)
         else()
-            # Fallback to the latest SDK available under /Library/Developer/CommandLineTools/SDKs/
-            execute_process(
-                COMMAND ls /Library/Developer/CommandLineTools/SDKs/
-                COMMAND grep "MacOSX[0-9][0-9]*\\.[0-9][0-9]*\\.sdk"
-                COMMAND sort -V
-                COMMAND tail -1
-                OUTPUT_VARIABLE LATEST_SDK_NAME
-                OUTPUT_STRIP_TRAILING_WHITESPACE
-            )
-            set(OSX_SDK_PATH "/Library/Developer/CommandLineTools/SDKs/${LATEST_SDK_NAME}")
+            message(FATAL_ERROR "Failed to get the official SDK path since xcrun failed, try to run "
+                    "`xcrun --sdk macosx --show-sdk-path` for more information")
         endif()
     endif()
 
--- a/runtime/CMakeLists.txt
+++ b/runtime/CMakeLists.txt
@@ -75,31 +75,31 @@ add_subdirectory(tests)
 
 if(APPLE AND (${CMAKE_SYSTEM_PROCESSOR} STREQUAL arm64))
 
-execute_process(
-    COMMAND bash -c "xcrun --sdk macosx --show-sdk-path"
-    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
-    OUTPUT_VARIABLE OSX_SDK_SYMBOL_PATH
-    OUTPUT_STRIP_TRAILING_WHITESPACE
-    RESULT_VARIABLE CMD_RESULT
-)
-
-get_filename_component(SEARCH_DIR ${OSX_SDK_SYMBOL_PATH} DIRECTORY)
-file(GLOB CANDIDATE_FILES "${SEARCH_DIR}/MacOSX*.sdk")
-set(HIGHEST_VERSION "0")
-set(OSX_SDK_LATEST_PATH "${OSX_SDK_SYMBOL_PATH}")
-
-foreach(CURRENT_FILE IN LISTS CANDIDATE_FILES)
-    get_filename_component(CURRENT_FILENAME ${CURRENT_FILE} NAME)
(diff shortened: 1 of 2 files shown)


runtime/CMakeLists.txt
```diff
@@ -74,6 +74,30 @@ add_subdirectory(lib)
...
+    string(REGEX MATCH [=[MacOSX(.*)\.sdk$]=] _REGEX_DUMMY "${CURRENT_FILENAME}")
+    set(CURRENT_VERSION "${CMAKE_MATCH_1}")
+    if(CURRENT_VERSION VERSION_GREATER HIGHEST_VERSION)
+        set(HIGHEST_VERSION ${CURRENT_VERSION})
+        set(OSX_SDK_PATH ${CURRENT_FILE})
+    endif()
+endforeach()
+
 # Don't rerun external project everytime we configure the runtime build.
 if(NOT EXISTS ${CMAKE_BINARY_DIR}/lib/liblapacke.3.dylib)
```
dime10, line 78: Can you move the discovery inside of this block so it doesn't run everytime?

rniczh: Sure, I've moved the correspondence code to the block in the following patch.

rniczh:
@dime10  my CMake version is 3.31.6 and macOS is 15.5

For now, I can reproduce it by removing the `runtime/build` dir, and run the `make runtime` again, the MacOS SDK path would be  `/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk`, instead of `/Library/Developer/CommandLineTools/SDKs/MacOSX15.5.sdk`. And actually, there are different SDK under the path, so the possible drawbacks I wrote is for the user want to choose the different version of SDK but not the latest one they have.

```
CMake Error at /Users/hongsheng.zheng/Workspace/catalyst/runtime/build/_lapacke-accelerate/src/lapacke-accelerate-build/_deps/acceleratelapack-src/CMakeLists.txt:88 (message):
  Please install a more recent XCode for a compatible MacOS SDK.


Preset CMake variables:

  BUILD_INDEX64:BOOL="FALSE"

-- The C compiler identification is AppleClang 17.0.0.17000013
-- The Fortran compiler identification is GNU 15.1.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /opt/homebrew/opt/ccache/libexec/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Checking whether Fortran compiler has -isysroot
-- Checking whether Fortran compiler has -isysroot - yes
-- Checking whether Fortran compiler supports OSX deployment target flag
-- Checking whether Fortran compiler supports OSX deployment target flag - yes
-- Detecting Fortran compiler ABI info
-- Detecting Fortran compiler ABI info - done
-- Check for working Fortran compiler: /opt/homebrew/bin/gfortran - skipped
-- Trying to link to ILP64 interface of LAPACK: 
-- Trying to link to ILP64 interface of LAPACK: FALSE
-- The Darwin kernel version is: 24.5.0
-- The MacOS version is: >=15.5
-- The BLAS/LAPACK version provided by the Accelerate framework should be: 3.12.0
-- The MacOS SDK is: /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk
-- The minimum compatible MacOS SDK version is: 15.5
-- Configuring incomplete, errors occurred!
```

sengthai:
> @sengthai You've encountered this error as well right, how did you resolve it?

Yes, I did. I solved that by updating my XCode SDK and run `make runtime; make runtime` sequentially.  
The first time it fails, it updates the CMake configuration. On the second run, it builds BLAS from source and succeeds.

rniczh:
> > @sengthai You've encountered this error as well right, how did you resolve it?
> 
> Yes, @dime10 I did. I tricked it by updating my Mac with Xcode SDK and running `make runtime; make runtime` sequentially. The first time it fails, it updates the CMake configuration. On the second run, it builds BLAS from source and succeeds.

I cannot reproduce what you reproduced 🤔 Even I build it twice, it still show the same error message, have you using brew to install lapacke before?

dime10:
> > @sengthai You've encountered this error as well right, how did you resolve it?
> 
> Yes, @dime10 I did. I tricked it by updating my Mac with Xcode SDK and running `make runtime; make runtime` sequentially. The first time it fails, it updates the CMake configuration. On the second run, it builds BLAS from source and succeeds.

I think that was a different error (latest macOS updating the underlying reference lapack version for Accelerate). I'm thinking of the issue with CMAKE_OSX_SYSROOT

PR #1880, opened 2025-07-10
upgrade PL and LT versions on rc

mehrdad2m:
**Context:**
We need to be able to test wheels with the latest dev versions of Pennylane and Lightning.

**Benefits:**

**Possible Drawbacks:**
the wheels will still not include the rc branch versions of PL and LT.  

**Related GitHub Issues:**

### DIFF
--- a/.dep-versions
+++ b/.dep-versions
@@ -10,8 +10,8 @@ enzyme=v0.0.180
 
 # For a custom PL version, update the package version here and at
 # 'doc/requirements.txt
-pennylane=0.42.0-dev67
+pennylane=0.42.0-dev68
 
 # For a custom LQ/LK version, update the package version here and at
 # 'doc/requirements.txt'
-lightning=0.42.0-dev16
+lightning=0.42.0-dev45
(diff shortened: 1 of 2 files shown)


paul0403:
> the wheels will still not include the rc branch versions of PL and LT.

Shouldn't we just do this instead? 
Or is this a hack to just test "rcrcrc" manually right now instead of waiting for the rcrcrc action at midnight?

mehrdad2m:
>Shouldn't we just do this instead?
Or is this a hack to just test "rcrcrc" manually right now instead of waiting for the rcrcrc action at midnight?

No this is so when we ship the rc wheels, they include the latest versions of PL and LT. Ideally we should include the rc versions but that still needs a bit of work. Firstly, we don't have any rc wheels uploaded for PL. For LT, we have the rc wheels but I should modify our scripts to correctly parse rc versions. This is the best we can do for now to quickly ship our rc wheels today.
Also rcrcrc does not install from wheels. It manually checks out the rc branches and installs from there.

PR #1881, opened 2025-07-10
Add Python 3.13 support

rniczh:

Add support for python 3.13

### DIFF
--- a/.github/workflows/constants.yaml
+++ b/.github/workflows/constants.yaml
@@ -74,7 +74,7 @@ jobs:
       - name: Python versions
         id: python_versions
         run: |
-          echo 'python_versions=["3.10", "3.11", "3.12"]' >> $GITHUB_OUTPUT
+          echo 'python_versions=["3.10", "3.11", "3.12", "3.13"]' >> $GITHUB_OUTPUT
 
       - name: Primary Python version
         id: primary_python_version
--- a/setup.py
+++ b/setup.py
@@ -161,6 +161,7 @@ classifiers = [
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
     "Programming Language :: Python :: 3 :: Only",
 ]
 


.github/workflows/constants.yaml
```diff
@@ -74,7 +74,7 @@ jobs:
       - name: Python versions
         id: python_versions
         run: |
-          echo 'python_versions=["3.10", "3.11", "3.12"]' >> $GITHUB_OUTPUT
+          echo 'python_versions=["3.10", "3.11", "3.12", "3.13"]' >> $GITHUB_OUTPUT
```
rniczh, line 77: Since `oqc_qcaas_client` doesn't support python3.13, this patch would skip the installation of it.

.github/workflows/build-wheel-macos-arm64.yaml
```diff
@@ -479,6 +479,7 @@ jobs:
         python${{ matrix.python_version }} -m pip install 'amazon-braket-pennylane-plugin>1.27.1'
 
     - name: Install OQC client
+      if: matrix.python_version != '3.13'
```
rniczh, line 495: Since `oqc_qcaas_client` doesn't support python3.13, this patch would skip the installation of it.

.github/workflows/build-wheel-macos-arm64.yaml
```diff
@@ -479,6 +492,7 @@ jobs:
...
```
dime10, line 495: I'm surprised just skipping the installation doesn't fail the tests that require it, I guess those are automatically skipped when oqc is not installed?

.github/workflows/build-wheel-macos-arm64.yaml
```diff
@@ -448,18 +448,31 @@ jobs:
...
+      id: python_build_version
+      env:
+        FORWARD_COMPAT_BASE_VERSION: ${{ needs.constants.outputs.forward_compat_python_version }}
+      run: |
+        SMALLEST_VERSION=$(printf '%s\n' "${{ matrix.python_version }}" "$FORWARD_COMPAT_BASE_VERSION" | sort -V | head -n 1)
+        if [[ "$SMALLEST_VERSION" == "$FORWARD_COMPAT_BASE_VERSION" && "${{ matrix.python_version }}" != "$FORWARD_COMPAT_BASE_VERSION" ]]; then
+          echo "python_build_version=$FORWARD_COMPAT_BASE_VERSION" >> $GITHUB_OUTPUT
+        else
+          echo "python_build_version=${{ matrix.python_version }}" >> $GITHUB_OUTPUT
+        fi
```
dime10: I think something like this might be a bit simpler? (untested)
```suggestion
        if contains(${{ fromJson(needs.constants.outputs.python_versions) }}, ${{ matrix.python_version }}); then
          echo "python_build_version=${{ matrix.python_version }}" >> $GITHUB_OUTPUT
        else
          echo "python_build_version=$FORWARD_COMPAT_BASE_VERSION" >> $GITHUB_OUTPUT
        fi
```

dime10: You might not even need the `${{ }}` within the github functions (contains, fromJson), but I'm not sure (I don't really get the GH language 😅).
```
contains(fromJson(needs.constants.outputs.python_versions), matrix.python_version)
```

.github/workflows/build-wheel-macos-arm64.yaml
rniczh: Yes, exactly. it's marked as skipped in conftest.py  if it fails to import the `qcaas_client` https://github.com/PennyLaneAI/catalyst/blob/d0d8f0ab4162af3304075ef7eef41856a594ac07/frontend/test/test_oqc/conftest.py#L46-L49

.github/workflows/build-wheel-macos-arm64.yaml
rniczh: Thanks for the suggestion! 

After the test on the GH. Seems like we still need the `${{ }}` to get value. For `if:` in GH actions, we could eliminate the need to use it (e.g. [this one](https://github.com/PennyLaneAI/catalyst/pull/1881/files#diff-1d070300e8a4097dfeb320744e6a01cbd8baba474d7424e57aa85a2f3b5dd013R477)). Since we use the shell `if` under the `run:`, so it couldn't works. Because under the `run:`, it just the shell scripts, that means we still need the `${{ }}` to get the value. And seems like the `env:` also need it.

For better readability, I make the `contains(fromJson(needs.constants.outputs.python_versions), matrix.python_version)` as an env variable `IsVersionSupported`, and use it in the `run:`. [Which](https://github.com/PennyLaneAI/catalyst/pull/1881/files#diff-76d47350e5e7d212639f8fcfdaa04cb633cc445eb0671e88231bf0a125366f21R460-R466
) may make it more easier to read.

dime10: Great! (good catch about this not being in the github script but the shell environment 👀)

PR #1882, opened 2025-07-10
Run code formatting on all PRs

mehrdad2m:
**Context:**
Code formatting should be run on all PRs whether they are targeting main or not.


frontend/test/lit/test_from_plxpr.py
```diff
@@ -304,6 +304,7 @@ def circuit(w1: int, w2: int):
 
 test_single_qubit_dynamic_static_interleaved()
 
+
```
paul0403, line 307: :laughing:

PR #1884, opened 2025-07-10
Update the dev requirement for building Runtime

maliasadi:
**Context:**
The runtime requires the `Development.SABIModule` component which is added in cmake 3.26. This short PR updates the dev requirements to ensure that the runtime is always built with a compatible cmake.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,6 +11,9 @@ nanobind>=2.4
 pybind11>=2.12.0
 PyYAML
 
+# Runtime requires the Development.SABIModule component, added in cmake 3.26:
+cmake >= 3.26
+
 # formatting/linting
 black
 clang-format~=20.1


requirements.txt
```diff
@@ -11,6 +11,9 @@ nanobind>=2.4
 pybind11>=2.12.0
 PyYAML
 
+# Runtime requires the Development.SABIModule component, added in cmake 3.26:
+cmake >= 3.26
```
dime10: Thanks Ali! I think we handle the compiler toolchain installation (including CMake) outside of pip. 

You can ensure the cmake minimum is enforced in the cmake files themselves though, via `cmake_minimum_required`, as well as in any scripts that install a compiler toolchain. The docs can also be updated: https://docs.pennylane.ai/projects/catalyst/en/stable/dev/installation.html#requirements

maliasadi: @dime10 the minimum required version of cmake is correctly set to 3.26 [here](https://github.com/PennyLaneAI/catalyst/blob/596d038c7d453a198218636635c7478849bfe8af/runtime/CMakeLists.txt#L1).

This is mainly a dev env issue that I'm trying to solve. When using Conda envs with Python versions < 3.13, the default installed version of cmake is below 3.26. So, every time I try to build Catalyst in a new env, I have to manually upgrade the installed version of CMake.

dime10: Got it! @mudit2812 Actually asked the same question, see here for some more context: https://xanaduhq.slack.com/archives/C030UB33QH3/p1752605858160829?thread_ts=1752604118.743549&cid=C030UB33QH3
If there are no concerns we can start moving more of the compiler toolchain into requirements.txt.

mudit2812: Should ninja also be here?

maliasadi: Yes, we can also install ninja and remove them both from the list of build requirements in the installation guideline

maliasadi:
David and I chatted offline -- there are more places to fix and update the build dependencies in Catalyst

requirements.txt
dime10:
```suggestion
# Compiler toolchain available via PIP
cmake>=3.26,<4
ninja
```

PR #1887, opened 2025-07-10
Replace the VLA use with standard C++ code

rniczh:

Previously, we enable VLA in tests to pass the test suite `runtime/tests/Test_NullQubit.cpp` #1875 But seems it's not the standard C++, and after discussed, the use of vector here wouldn't affect the `__catalyst_*` functions. So just wrap it with a vector and pass the internal data to `MemRefT_double_2d`.


dime10 (approved):
Looks good, thanks!

PR #1888, opened 2025-07-10
Update compilers in GitHub Actions workflow to include Clang

sengthai:
Modified the `constants.yaml` workflow file to include Clang in the list of compilers when the `MULTIPLE` condition is met.

### DIFF
--- a/.github/workflows/constants.yaml
+++ b/.github/workflows/constants.yaml
@@ -88,9 +88,8 @@ jobs:
       - name: Compilers (All)
         id: compilers
         run: |
-          # TODO: add Clang after https://github.com/actions/runner-images/issues/8659 is fixed
           if [[ $MULTIPLE ]]; then
-            echo 'compilers=["gcc"]' >> $GITHUB_OUTPUT
+            echo 'compilers=["gcc", "clang"]' >> $GITHUB_OUTPUT
           else
             echo 'compilers=["gcc"]' >> $GITHUB_OUTPUT
           fi


.github/workflows/constants.yaml
```diff
@@ -84,13 +84,20 @@ jobs:
...
         id: rt_backends
         run: echo 'rt_backends=["lightning", "lightning-kokkos", "openqasm", "oqc", "oqd"]' >> $GITHUB_OUTPUT
+      
+      - name: Print system and compiler info
+        run: |
+          uname -a
+          which clang++
+          which g++
+          clang --version
+          g++ --version
```
sengthai: TODO: remove before merge

mlir/lib/Quantum/Transforms/AdjointPatterns.cpp
```diff
@@ -165,7 +166,7 @@ class AdjointGenerator {
         if (auto parametrizedGate = dyn_cast<quantum::ParametrizedGate>(operation)) {
             OpBuilder::InsertionGuard insertionGuard(builder);
             builder.setInsertionPoint(clone);
-            ValueRange params = parametrizedGate.getAllParams();
+            SmallVector<Value> params = llvm::to_vector(parametrizedGate.getAllParams());
```
mlxd: Just to mention, you can remove the vector altogether and use the range if you opt-out of using the enumerate call in the for-each loop --- this worked for me on libstdcxx15. Tracking the index value with a counter can replace the enumerated `idx` value.

sengthai: Thanks for the tip! I’ll try it with `ValueRange` and just add a counter for indexing manually.

sengthai: It works well in CI! 
[c0b7884](https://github.com/PennyLaneAI/catalyst/pull/1888/commits/c0b788405186ed7205f76deabef1520bae2414e1)

PR #1892, opened 2025-07-11
Bump Pennylane and Lightning versions prior to the release

mehrdad2m:
**Context:**

Bump the PennyLane and Lightning minimum versions in preparation for the Catalyst 0.12.0 release. These changes ensure we can build the wheels after the Lightning release and before the core PennyLane release.


erick-xanadu (approved):
I think it looks correct! You may want to wait for other members to other chime in as well.

dime10 (approved):
🥳

PR #1898, opened 2025-07-14
Bump to the jax commit with the source_info update

paul0403:
**Context:**
Bump to the jax commit with the source_info update


paul0403:
@mehrdad2m format check is now be run on these ones (not targetting main)!

PR #1904, opened 2025-07-15
De-tensorize function boundary pass

josephleekl:
**Context:**
Scalar tensors across function boundaries are prevalent due to JAX, which results in extra instructions required to extract elements from tensor arguments, and construct tensors from scalar elements to pass to the functions.

A new MLIR pass is created to detensorize function boundaries. Extra tensor extract and from_element ops are added then folded.

**Benefits:**
Reduce instructions count

### Time score

|Workflow             |Catalyst (qjit+exec)|  |Catalyst (exec)|  |
| :--- | ---: | :---: | ---: | :---: |
|QPE                  |0.99                |  |0.99           |  |
|QSVT                 |0.99                |  |1.00           |  |
|XAS                  |1.40                |🟢|0.99           |  |
|shor                 |0.98                |  |0.98           |  |
|molecular_hamiltonian|  -                 |  |  -            |  |
|sampling             |1.00                |  |1.02           |  |
|stateprep            |0.98                |  |0.68           |🔴|
|grover               |1.02                |  |0.96           |🔴|
|QAOA_layers_scaling  |1.00                |  |1.01           |  |
|QML                  |0.98                |  |1.00           |  |
|QML_jaxjit           |  -                 |  |  -            |  |
|UCCSD                |0.98                |  |1.02           |  |
|VQE                  |  -                 |  |1.00           |  |

## Detailed Per-Workflow Results

- The assumed noise level for runtime improvements/regressions is 3.0%
- ⚠️ marks workflows with runtime fluctuations greater than 5.0% (std/mean)

### Catalyst (compilation + execution)

|Workflow                              |time [s]|std/mean|  |time score|  |virt mem [MB]|virt mem score|  |phys mem [MB]|phys mem score|  |
| :--- | ---: | ---: | :---: | ---: | :---: | ---: | ---: | :---: | ---: | ---: | :---: |
|QPE[11-12]                            |1.196   |7.1%    |⚠️|0.97      |  |365.729      |  -           |  |116.785      |  -           |  |
|QPE[12-12]                            |2.222   |6.5%    |⚠️|1.00      |  |499.972      |  -           |  |106.713      |  -           |  |
|QSVT[9]                               |26.464  |18.0%   |⚠️|0.99      |  |4,279.710    |  -           |  |2,471.539    |  -           |  |
|XAS[2-1-9]                            |28.783  |0.8%    |  |1.46      |🟢|4,424.084    |  -           |  |2,501.321    |  -           |  |
|XAS[2-2-9]                            |38.110  |0.6%    |  |1.33      |🟢|4,424.159    |  -           |  |2,500.624    |  -           |  |
|shor[15]                              |1.288   |7.7%    |⚠️|0.98      |  |237.171      |  -           |  |138.236      |  -           |  |
|shor[33]                              |3.168   |3.2%    |  |0.99      |  |237.761      |  -           |  |141.140      |  -           |  |
|sampling[24-2]                        |1.563   |5.0%    |  |1.02      |  |1,046.111    |  -           |  |116.871      |  -           |  |
|sampling[25-2]                        |3.411   |3.1%    |  |0.99      |  |1,852.645    |  -           |  |106.881      |  -           |  |
|stateprep[12-MottonenStatePreparation]|7.099   |2.8%    |  |0.98      |  |404.347      |  -           |  |229.323      |  -           |  |
|grover[18]                            |9.678   |0.9%    |  |1.02      |  |459.678      |  -           |  |291.979      |  -           |  |
|QAOA_layers_scaling[19-4]             |111.156 |0.4%    |  |1.00      |  |584.816      |  -           |  |5,387.448    |  -           |  |
|QML[IQPKernelClassifier-12-10]        |121.986 |1.1%    |  |0.98      |  |2,122.109    |  -           |  |912.646      |  -           |  |
|UCCSD[H2O-STO-3G]                     |12.956  |1.4%    |  |0.99      |  |532.305      |  -           |  |286.573      |  -           |  |
|UCCSD[NH3-STO-3G]                     |33.076  |2.3%    |  |0.97      |  |1,025.818    |  -           |  |537.739      |  -           |  |
|VQE[H2O-STO-3G]                       |21.477  |0.8%    |  |0.99      |  |475.394      |  -           |  |271.380      |  -           |  |
|VQE[NH3-STO-3G]                       |72.532  |0.7%    |  |  -       |  |1,224.448    |  -           |  |785.674      |  -           |  |

### Catalyst (execution only)

|Workflow                              |time [s]|std/mean|  |time score|  |virt mem [MB]|virt mem score|  |phys mem [MB]|phys mem score|  |
| :--- | ---: | ---: | :---: | ---: | :---: | ---: | ---: | :---: | ---: | ---: | :---: |
|QPE[11-12]                            |0.896   |1.0%    |  |0.98      |  |365.954      |  -           |  |117.281      |  -           |  |
|QPE[12-12]                            |2.060   |0.7%    |  |0.99      |  |500.201      |  -           |  |116.912      |  -           |  |
|QSVT[9]                               |0.143   |6.2%    |⚠️|1.00      |  |4,279.643    |  -           |  |2,471.780    |  -           |  |
|XAS[2-1-9]                            |9.320   |0.7%    |  |1.00      |  |4,424.198    |  -           |  |2,502.304    |  -           |  |
|XAS[2-2-9]                            |18.713  |0.4%    |  |0.99      |  |4,424.159    |  -           |  |2,496.918    |  -           |  |
|shor[15]                              |0.051   |6.1%    |⚠️|0.96      |🔴|237.228      |  -           |  |138.162      |  -           |  |
|shor[33]                              |1.817   |1.5%    |  |1.00      |  |237.699      |  -           |  |140.599      |  -           |  |
|sampling[24-2]                        |1.368   |1.6%    |  |1.03      |🟢|1,046.060    |  -           |  |103.055      |  -           |  |
|sampling[25-2]                        |3.186   |1.3%    |  |1.00      |  |1,856.906    |  -           |  |123.036      |  -           |  |
|stateprep[12-MottonenStatePreparation]|0.033   |19.5%   |⚠️|0.68      |🔴|404.091      |  -           |  |242.000      |  -           |  |
|grover[18]                            |1.561   |1.6%    |  |0.96      |🔴|468.432      |  -           |  |308.101      |  -           |  |
|QAOA_layers_scaling[19-4]             |106.051 |0.5%    |  |1.01      |  |584.730      |  -           |  |5,745.373    |  -           |  |
|QML[IQPKernelClassifier-12-10]        |97.194  |3.2%    |  |1.00      |  |1,184.518    |  -           |  |623.976      |  -           |  |
|UCCSD[H2O-STO-3G]                     |0.113   |1.9%    |  |1.05      |🟢|530.752      |  -           |  |292.098      |  -           |  |
|UCCSD[NH3-STO-3G]                     |1.153   |4.2%    |  |0.99      |  |1,025.717    |  -           |  |550.011      |  -           |  |
|VQE[H2O-STO-3G]                       |2.067   |1.8%    |  |1.00      |  |475.332      |  -           |  |272.044      |  -           |  |
|VQE[NH3-STO-3G]                       |16.754  |1.0%    |  |1.01      |  |1,224.450    |  -           |  |785.883      |  -           |  |
**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-95476]

### DIFF
--- a/mlir/lib/Catalyst/Transforms/DetensorizeFunctionBoundaryPass.cpp
+++ b/mlir/lib/Catalyst/Transforms/DetensorizeFunctionBoundaryPass.cpp
@@ -64,7 +64,9 @@ struct DetensorizeCallSitePattern : public OpRewritePattern<func::CallOp> {
             return failure();
         }
 
-        // Skip for QNodes (for gradients boundaries)
+        // Skip for QNodes
+        // Some Gradient boundaries only work for Tensor signatures
+        // and not scalar ones, hence we skip them here.
         if (funcOp->hasAttr("qnode")) {
             return failure();
         }
@@ -122,45 +124,6 @@ struct DetensorizeCallSitePattern : public OpRewritePattern<func::CallOp> {
         }
     }
 
-    void replaceCallOp(PatternRewriter &rewriter, func::CallOp &callOp,
-                       func::FuncOp &newFuncOp) const
-    {
-        rewriter.setInsertionPoint(callOp);
-        SmallVector<Value> newOperands;
-        for (Value operand : callOp.getOperands()) {
-            // Insert ExtractOp if the old operand is a scalar tensor to bridge the detensorized
-            // function
-            if (isScalarTensor(operand.getType())) {
-                auto extractOp =
-                    rewriter.create<tensor::ExtractOp>(callOp.getLoc(), operand, ValueRange{});
-                newOperands.push_back(extractOp.getResult());
-            }
-            else {
-                newOperands.push_back(operand);
-            }
-        }
-
-        auto newCallOp = rewriter.create<func::CallOp>(callOp.getLoc(), newFuncOp, newOperands);
-
-        SmallVector<Value> newResults;
-        for (size_t i = 0; i < callOp.getNumResults(); ++i) {
-            Value oldResult = callOp.getResult(i);
-            Value newResult = newCallOp.getResult(i);
-            if (isScalarTensor(oldResult.getType())) {
-                // Insert a FromElementsOp if the old result is a scalar tensor to bridge the
-                // detensorized function
-                auto fromElementsOp = rewriter.create<tensor::FromElementsOp>(
-                    callOp.getLoc(), oldResult.getType(), newResult);
-                newResults.push_back(fromElementsOp.getResult());
-            }
-            else {
-                newResults.push_back(newResult);
-            }
-        }
-
-        rewriter.replaceOp(callOp, newResults);
-    }
-
     void mapFuncOpBodyAndReturnOp(PatternRewriter &rewriter, Block *newEntryBlock,
                                   func::FuncOp &funcOp, IRMapping &mapper) const
     {
@@ -203,6 +166,45 @@ struct DetensorizeCallSitePattern : public OpRewritePattern<func::CallOp> {
             }
         }
         rewriter.create<func::ReturnOp>(oldReturnOp.getLoc(), newReturnOperands);
+    }
+
+    void replaceCallOp(PatternRewriter &rewriter, func::CallOp &callOp,
+                       func::FuncOp &newFuncOp) const
+    {
+        rewriter.setInsertionPoint(callOp);
+        SmallVector<Value> newOperands;
+        for (Value operand : callOp.getOperands()) {
+            // Insert ExtractOp if the old operand is a scalar tensor to bridge the detensorized
+            // function
+            if (isScalarTensor(operand.getType())) {
+                auto extractOp =
+                    rewriter.create<tensor::ExtractOp>(callOp.getLoc(), operand, ValueRange{});
+                newOperands.push_back(extractOp.getResult());
+            }
+            else {
+                newOperands.push_back(operand);
+            }
+        }
+
+        auto newCallOp = rewriter.create<func::CallOp>(callOp.getLoc(), newFuncOp, newOperands);
+
+        SmallVector<Value> newResults;
+        for (size_t i = 0; i < callOp.getNumResults(); ++i) {
+            Value oldResult = callOp.getResult(i);
+            Value newResult = newCallOp.getResult(i);
+            if (isScalarTensor(oldResult.getType())) {
+                // Insert a FromElementsOp if the old result is a scalar tensor to bridge the
+                // detensorized function
+                auto fromElementsOp = rewriter.create<tensor::FromElementsOp>(
+                    callOp.getLoc(), oldResult.getType(), newResult);
+                newResults.push_back(fromElementsOp.getResult());
+            }
+            else {
+                newResults.push_back(newResult);
+            }
+        }
+
+        rewriter.replaceOp(callOp, newResults);
     }
 };
 } // namespace


mlir/lib/Catalyst/Transforms/DetensorizeFunctionBoundaryPass.cpp
```diff
@@ -0,0 +1,309 @@
...
+    return false;
+}
+
+struct DetensorizeFuncPattern : public OpRewritePattern<func::FuncOp> {
+    using OpRewritePattern<func::FuncOp>::OpRewritePattern;
+
+    LogicalResult matchAndRewrite(func::FuncOp funcOp, PatternRewriter &rewriter) const override
+    {
+        // Skip for main function
+        if (funcOp->hasAttr("llvm.emit_c_interface")) {
```
josephleekl: this is to avoid transforming the function called by python - is this a good way to do it?

dime10: Great idea! The entry-point function is also always prefixed with `jit_` but I think the attribute method should be more reliable :)

mlir/lib/Catalyst/Transforms/DetensorizeFunctionBoundaryPass.cpp
```diff
@@ -0,0 +1,309 @@
...
+            return failure();
+        }
+
+        // Collect the argument and result types
+        FunctionType funcType = funcOp.getFunctionType();
+        SmallVector<Type> newArgTypes;
+        SmallVector<Type> newResultTypes;
+
+        for (Type type : funcType.getInputs()) {
+            newArgTypes.push_back(getScalarType(type));
```
dime10: Small suggestion, but something like this could be more readable? 
```suggestion
            newArgTypes.push_back(getScalarOrOriginalType(type));
```

mlir/lib/Catalyst/Transforms/DetensorizeFunctionBoundaryPass.cpp
```diff
@@ -0,0 +1,309 @@
...
+            }
+        }
+
+        rewriter.replaceOp(callOp, newResults);
+        return success();
+    }
+};
+
+// Remove from_elements -> extract
+struct FoldExtractFromElementsPattern : public OpRewritePattern<tensor::ExtractOp> {
```
dime10: Out of curiosity this pattern doesn't exist upstream? Either way though probably no harm in having it here explicitly

dime10 (commented):
Wow this is really clean, excellent work 💯

mlir/lib/Catalyst/Transforms/DetensorizeFunctionBoundaryPass.cpp
```diff
@@ -0,0 +1,370 @@
...
+        if (isScalarTensor(type)) {
+            return true;
+        }
+    }
+    return false;
+}
+
+// Exclude functions that are part of the any gradient or mitigation call chain
+// through attributes e.g. `implementation` or `callee`
+void findAndExcludeGradientMitigationCallChains(ModuleOp module,
```
josephleekl: this is to avoid any calls involving gradients and mitigation (zne), but the way it's implemented of walking through the entire module may be complete overkill 😅 i'll investigate a bit more whether this is necessary/the performance of this

mlir/lib/Catalyst/Transforms/DetensorizeFunctionBoundaryPass.cpp
josephleekl: Indeed both of the foldings (extract -> from_elements and vice versa) already happen with canonicalize! 😅 I'll remove them to avoid duplication

mlir/lib/Catalyst/Transforms/DetensorizeFunctionBoundaryPass.cpp
josephleekl: updated!

mlir/include/Catalyst/Transforms/Passes.td
```diff
@@ -17,11 +17,12 @@
...
 include "mlir/Pass/PassBase.td"
 
-def DetensorizeFunctionBoundaryPass : Pass<"detensorize-function-boundary"> {
+def DetensorizeFunctionBoundaryPass : Pass<"detensorize-function-boundary", "mlir::ModuleOp"> {
     let summary = "Detensorize across function boundary.";
 
     let dependentDialects = [
-        "tensor::TensorDialect"
+        "tensor::TensorDialect",
+        "mlir::func::FuncDialect"
```
dime10: This shouldn't be necessary since func will already be present in the input. The dependent dialects field is used to load additional dialects that may not yet have been loaded because they weren't present in the input, but will or may be created by the pass.

The tensor entry is appropriate though since you are creating tensor operations, but they aren't necessarily present in the input since you are merely matching against func and call ops.

mlir/lib/Catalyst/Transforms/DetensorizeFunctionBoundaryPass.cpp
dime10: Hmm, this does seem potentially overkill. I suppose the issue is that we want to modify FuncOps without knowing whose calling it. I think there is a few options here:
- copy instead of delete the function you are detensorizing, then only "allowed" call-like operations will be modified to point to the new target
- instead of applying the two patterns independently, only match against allowed call ops and then from there transform the function definition (if it hasn't already been transformed)

These should be more efficient than this manual traversal 😅

dime10: I notice your comment says "call chain", is it really necessary to include functions lower down in the call tree from a gradient call for example? The main issues I can think of should be at the interface only.

The same question should also be considered for zne.

dime10 (commented):
Thanks Joseph!

mlir/include/Catalyst/Transforms/Passes.td
josephleekl: Thanks for the explanation! 👓

mlir/lib/Catalyst/Transforms/DetensorizeFunctionBoundaryPass.cpp
josephleekl: Thanks @dime10 ! I've updated now: https://github.com/PennyLaneAI/catalyst/pull/1904/commits/7e61d5a299bfce233c525e003dc79b0685d1b969
- No longer walk the module
- Only update if the FuncOp is actually called by a Func Call
- Additionally we skip if it is used by gradient (through the attribute `diff_method`). The reason for this extra step (which is the original motivation for the overkill call chain) is that the following MLIR fails to compile down, even though `circuit_0` is not *directly* used by the `gradient.grad` call:

```
module @grad.cost {
  func.func public @jit_grad.cost(%arg0: tensor<f64>) -> tensor<f64> attributes {llvm.emit_c_interface} {
    %0 = gradient.grad "auto" @cost(%arg0) {diffArgIndices = dense<0> : tensor<1xi64>} : (tensor<f64>) -> tensor<f64>
    return %0 : tensor<f64>
  }
  func.func private @cost(%arg0: tensor<f64>) -> tensor<f64> attributes {llvm.linkage = #llvm.linkage<internal>} {
    %extracted = tensor.extract %arg0[] : tensor<f64>
    %0 = call @circuit_0(%extracted) : (f64) -> f64
    %from_elements = tensor.from_elements %0 : tensor<f64>
  }
  func.func public @circuit_0(%arg0: f64) -> f64 attributes {diff_method = "parameter-shift", llvm.linkage = #llvm.linkage<internal>, qnode} {
    quantum.device shots(%c0_i64) ["/home/ubuntu/work/detensorize-func-bound/venv-detensorize-func-bound/lib/python3.11/site-packages/pennylane_lightning/liblightning_qubit_catalyst.so", "LightningSimulator", "{'mcmc': False, 'num_burnin': 0, 'kernel_name': None}"]
    %out_qubits = quantum.custom "PauliX"() %1 : !quantum.bit
    %2 = quantum.namedobs %out_qubits[ PauliZ] : !quantum.obs
    %3 = quantum.expval %2 : f64
    %4 = quantum.insert %0[ 0], %out_qubits : !quantum.reg, !quantum.bit
    quantum.device_release
    return %3 : f64
  }
  func.func @setup() {
    quantum.init
    return
  }
  func.func @teardown() {
    quantum.finalize
    return
  }
}
```

dime10: Ah you're totally right, the qnode function does form another boundary for the AD implementation, so that makes sense to exclude those if they are being differentiated!

Nice work :)

mlir/lib/Catalyst/Transforms/DetensorizeFunctionBoundaryPass.cpp
```diff
@@ -108,17 +64,23 @@ struct DetensorizeFuncPattern : public OpRewritePattern<func::FuncOp> {
...
+                if (!isa<func::CallOp>(use.getUser())) {
+                    return failure();
+                }
+            }
         }
 
-        // Skip if function is excluded for gradient/mitigation call chain
-        if (excludedFunctions.count(funcOp.getSymName())) {
+        // Skip if used by gradient
+        if (funcOp->hasAttr("diff_method")) {
```
dime10: Do we know if this attribute is only attached when differentiating, and not when a user set the keyword on the qml.qnode but isn't actually diffing it?

mlir/lib/Catalyst/Transforms/DetensorizeFunctionBoundaryPass.cpp
```diff
@@ -0,0 +1,277 @@
...
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/Tensor/IR/Tensor.h"
+#include "mlir/IR/IRMapping.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+#include "Catalyst/IR/CatalystDialect.h"
+#include "Gradient/IR/GradientOps.h"
+#include "Mitigation/IR/MitigationOps.h"
```
josephleekl: remove

mlir/lib/Catalyst/Transforms/DetensorizeFunctionBoundaryPass.cpp
```diff
@@ -108,17 +64,23 @@ struct DetensorizeFuncPattern : public OpRewritePattern<func::FuncOp> {
...
-            return failure();
+        // Skip if not used directly used func call (e.g. gradient VJP or mitigation ZNE)
+        auto module = funcOp->getParentOfType<ModuleOp>();
+        if (auto uses = mlir::SymbolTable::getSymbolUses(funcOp, module)) {
+            for (const mlir::SymbolTable::SymbolUse &use : *uses) {
+                if (!isa<func::CallOp>(use.getUser())) {
+                    return failure();
+                }
+            }
         }
```
dime10: So I think this should be _relatively_ safe now. I'm a bit cautious though because symbols are more complicated than values, and I don't quite know how this `getSymbolUses` works.
- I don't think a symbol table stores the uses of a symbol (only the definition?), in which case I'm assuming that this still has to walk all the functions in the module to find uses of the symbol you are looking for. Not ideal performance wise if we are doing this many times.
- Additionally, there might be some edge cases with nested modules (which we use for qnodes, for now only in the early stages of the compiler). It looks something like this:
  ```mlir
  module @small_circ {
    func.func public @jit_small_circ() -> tensor<4xcomplex<f64>> attributes {llvm.emit_c_interface} {
      ...
      %0 = catalyst.launch_kernel @module_small_circ::@small_circ(%cst) : (tensor<4x4xcomplex<f64>>) -> tensor<4xcomplex<f64>>
      ...
    }
    module @module_small_circ {
      func.func public @small_circ(%arg0: tensor<4x4xcomplex<f64>>) -> tensor<4xcomplex<f64>> attributes {diff_method = "parameter-shift", llvm.linkage = #llvm.linkage<internal>, qnode} {
        ...
      }
    }
  }
  ```
  When you're matching the `@small_circ` function, I don't know if you'll detect the `launch_kernel` call for instance.
  Admittedly where your pass is running right now you won't encounter this situation yet, but that will likely change in the future and we might as well make it as robust as possible.
- If the same function is both called and differentiated, one could still create a detensorized version for the call, and leaving the original for the differentiation. The downside is of course that now we have an almost identical function twice, which may increase compile time. So this may or may not be worth considering.

dime10: I'm not sure what the best approach is right now. There are the two I proposed in the [previous comment](https://github.com/PennyLaneAI/catalyst/pull/1904#discussion_r2252449714), but copying the function does have its downsides if we don't delete the original once it's no longer used (I don't think mlir does any "dead function elimination" 😅). At least this version is short and easy to understand. I just want to be careful about the edge cases (i.e. future bugs).

mlir/lib/Catalyst/Transforms/DetensorizeFunctionBoundaryPass.cpp
josephleekl: turns out `diff_method` is [automatically added ](https://github.com/PennyLaneAI/catalyst/blob/8a199e255822e14467d0b4b7e3f31e7ef42aa52d/frontend/catalyst/jax_primitives_utils.py#L123)to qnodes. 

I'll have a think about how else to detect the uses along with the comment below, unless we want to skip qnodes completely

mlir/lib/Catalyst/Transforms/DetensorizeFunctionBoundaryPass.cpp
josephleekl: Good points! I tested the nested module case, and you're right that my current implementation fails to detect the use. In this case I'll go with your suggestion:

> instead of applying the two patterns independently, only match against allowed call ops and then from there transform the function definition (if it hasn't already been transformed)

I'm also slightly worried about the extra compilation time from copying/duplicating some functions - I'll try it out and see what difference it makes.

josephleekl: I now learned that `getSymbolUses` does in fact scan the entire module, and is costly in terms of performance. One option is to pre-calculate once with `SymbolUserMap`, but that's still a performance hit. I don't know of any way to check that the caller is a gradient call (needed for this case [here](https://github.com/PennyLaneAI/catalyst/pull/1904#discussion_r2255328233)) - do you have any better idea? For now I'm just checking for the `qnode` attribute, and ignoring those entirely 😅 

I have now updated the implementation to match appropriate CallOps, then copy the corresponding funcOp (append with `.detensorized`) with the new signature.

I also found that the `symbol-dce` [pass](https://mlir.llvm.org/docs/Passes/#-symbol-dce) can eliminate the unused dead function (if they're not public).

mlir/lib/Catalyst/Transforms/DetensorizeFunctionBoundaryPass.cpp
josephleekl: as mentioned in the comment below, I have not found a better way, instead I'm just explicitly ignoring qnodes...

mlir/lib/Catalyst/Transforms/DetensorizeFunctionBoundaryPass.cpp
dime10: > I now learned that getSymbolUses does in fact scan the entire module, and is costly in terms of performance.

Good to know 😅 

> I don't know of any way to check that the caller is a gradient call (needed for this case https://github.com/PennyLaneAI/catalyst/pull/1904#discussion_r2255328233) - do you have any better idea? For now I'm just checking for the qnode attribute, and ignoring those entirely 😅

Unfortunately there might not be a way around traversing the call graph from a diff op to find out if a qnode lies in it🤔 We could solve it from the frontend by only providing the diff_method kwarg if the qnode is actually differentiated, or provide a small analysis that only runs once to check this property on the IR. 

> I have now updated the implementation to match appropriate CallOps, then copy the corresponding funcOp (append with .detensorized) with the new signature.

Great!

> I also found that the symbol-dce [pass](https://mlir.llvm.org/docs/Passes/#-symbol-dce) can eliminate the unused dead function (if they're not public).

Oh awesome, that's good to know! We should use it together with this pass then :) (I'm sure that must traverse the whole IR but at least it only happens once)

dime10: > We could solve it from the frontend by only providing the diff_method kwarg if the qnode is actually differentiated, or provide a small analysis that only runs once to check this property on the IR.

Trying to think whether this is worth it. I think many programs are non-AD programs actually, it would be a shame to have this restriction for all programs. 

But now that I think about it, this is something that should likely be solved in the gradient dialect. It should be able to deal with functions that have scalars in their signatures. I remember at least some of the methods we implemented worked on both tensors and scalars.

But I think for now this might be out of scope. Excellent work 💯

josephleekl: > But now that I think about it, this is something that should likely be solved in the gradient dialect. It should be able to deal with functions that have scalars in their signatures. I remember at least some of the methods we implemented worked on both tensors and scalars.

That would be ideal!

josephleekl: I've tested the performance of this on XAS as well, it seems to give a pretty good performance upgrade:
```
Nothing
---------------------------------------------------------------------------------------- benchmark: 2 tests ---------------------------------------------------------------------------------------
Name (time in s)                         Min                Max               Mean            StdDev             Median               IQR            Outliers     OPS            Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_XAS_with_compilation[2-1-9]     40.5770 (1.0)      42.8336 (1.0)      41.3838 (1.0)      0.9649 (1.21)     40.8129 (1.0)      1.4147 (1.28)          1;0  0.0242 (1.0)           5           1
test_XAS_with_compilation[2-2-9]     49.7151 (1.23)     51.7178 (1.21)     50.4924 (1.22)     0.7988 (1.0)      50.5416 (1.24)     1.1011 (1.0)           1;0  0.0198 (0.82)          5           1
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

just `detensorize-function-boundary`

---------------------------------------------------------------------------------------- benchmark: 2 tests ---------------------------------------------------------------------------------------
Name (time in s)                         Min                Max               Mean            StdDev             Median               IQR            Outliers     OPS            Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_XAS_with_compilation[2-1-9]     28.9980 (1.0)      29.6205 (1.0)      29.3813 (1.0)      0.2535 (1.44)     29.4872 (1.0)      0.3705 (1.16)          1;0  0.0340 (1.0)           5           1
test_XAS_with_compilation[2-2-9]     38.6764 (1.33)     39.0615 (1.32)     38.8957 (1.32)     0.1760 (1.0)      38.9582 (1.32)     0.3196 (1.0)           1;0  0.0257 (0.76)          5           1
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------


`detensorize-function-boundary`' + `symbol-dce`
---------------------------------------------------------------------------------------- benchmark: 2 tests ---------------------------------------------------------------------------------------
Name (time in s)                         Min                Max               Mean            StdDev             Median               IQR            Outliers     OPS            Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_XAS_with_compilation[2-1-9]     27.5732 (1.0)      28.4671 (1.0)      28.2450 (1.0)      0.3794 (1.14)     28.3824 (1.0)      0.3150 (1.20)          1;1  0.0354 (1.0)           5           1
test_XAS_with_compilation[2-2-9]     36.9908 (1.34)     37.7718 (1.33)     37.5823 (1.33)     0.3333 (1.0)      37.7353 (1.33)     0.2635 (1.0)           1;1  0.0266 (0.75)          5           1
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

I'll gather the number for all the tests, then if it all looks good I'll merge this

dime10: Wow that's incredible, we are talking 25% improvement (seconds) just from this? 🫨

PR #1905, opened 2025-07-15
Dependency management update

dime10:
Additional updates & cleanup on #1884

### DIFF
--- a/.github/workflows/check-catalyst.yaml
+++ b/.github/workflows/check-catalyst.yaml
@@ -42,7 +42,7 @@ jobs:
     steps:
       - name: Checkout Catalyst repo
         uses: actions/checkout@v4
-      
+
       - name: Set up Python # Ensure python3.10 is used
         uses: actions/setup-python@v5
         with:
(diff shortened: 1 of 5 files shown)


.github/workflows/check-catalyst.yaml
```diff
@@ -51,7 +51,7 @@ jobs:
       - name: Install dependencies
         run: |
           sudo apt-get update
-          sudo apt-get -y -q install ninja-build make cmake clang libomp-dev
+          sudo apt-get -y -q install ninja-build make cmake clang
```
paul0403, line 54: I'm not super familiar with the context, why are we removing ninja in other places but not here?

dime10: I only removed ninja and cmake installations in places where we are already using `pip install -r requirements.txt` (in lots of places we don't do this and only install what we need because it is slow), you still need those packages to come from somewhere.

paul0403: Makes sense, thanks!

paul0403 (approved):

PR #1910, opened 2025-07-16
Fix qml.prod operator in AutoGraph decorator passthroughs

### LINKED ISSUE #1824: Stateprep in a qml.prod with autograph
This is with PennyLane-Catalyst==0.11.0, PennyLane-qiskit==0.41.0, PennyLane_Lightning==0.41.0

The following code produces a compile error: 

```
@qml.prod
def state_routine():
    qml.StatePrep(np.append(np.ones(7),[0])/7,wires=[0,1,2],normalize=True)
    #qml.adjoint(qml.StatePrep(np.append(np.ones(7),[0])/7,wires=[0,1,2],normalize=True))
    #qml.adjoint(qml.adjoint(qml.StatePrep(np.append(np.ones(7),[0])/7,wires=[0,1,2],normalize=True)))
    #qml.PauliX(wires=0) #I include this just to make sure qml.prod isn't doing something weird
    return qml.state()

@qml.qjit(autograph=True, abstracted_axes=("n",)) 
def sp_test():
    state_routine()
    return qml.state()
out=sp_test()
```
Giving this error:
```
CompileError: StatePrep is not supported on 'lightning.qubit' device
```
This demonstrating that state prep seems to not work on the autograph when in a @qml.prod; can do the qjit without autograph and it works, oddly the adjoint of the stateprep does work (ie the commented bit) as does the adjoint of the adjoint.

The following code does not produce this error:
```
@qml.qjit(autograph=True, abstracted_axes=("n",)) 
def sp_test():
    qml.StatePrep(np.append(np.ones(7),[0])/7,wires=[0,1,2],normalize=True)
    return qml.state()
outMat=sp_test()
```
So, mixed with the observations of the adjoint, this seems to be in how the autograph unravels (or does not unravel) what is in the @qml.prod

If the PauliX line is uncommented in the first bit of code, a different error is thrown, but for that I will open a separate investigation.

dime10:
Relies on https://github.com/PennyLaneAI/pennylane/pull/7889.

The above fixes the issue of AutoGraph transforming internal PennyLane library code when it shouldn't. This PR provides an additional fix to ensure AutoGraph doesn't fail when `qml.prod` operators are passed to other decorators like `qml.adjoint`.

We have a fix in-place that transparently passes through decorators like `qml.adjoint` without transforming the decorator, but still transforming the function being wrapped by it. That code expects the first argument to `qml.adjoint`-like functions to be either a PennyLane operation, or a callable. However, `qml.prod` is a valid target that is an `Operator` but not an `Operation`.
This PR changes the expected type to `Operator` (which is a parent class of `Operation`).

NOTE: This PR does not enable support for AutoGraph to transform code targeted by `qml.prod`, that is a separate issue.

closes #1824
[sc-93971]


doc/requirements.txt
```diff
@@ -31,6 +31,6 @@ lxml_html_clean
 
 # Pre-install PL development wheels
 --extra-index-url https://test.pypi.org/simple/
-pennylane-lightning-kokkos==0.42.0
-pennylane-lightning==0.42.0
-pennylane==0.42.0
+pennylane-lightning-kokkos==0.42
+pennylane-lightning==0.42
```
dime10:
```suggestion
pennylane-lightning-kokkos==0.42.0
pennylane-lightning==0.42.0
```

frontend/test/pytest/test_autograph.py
```diff
@@ -2097,6 +2097,23 @@ def loop(i):
...
+            qml.H(0)
+            qml.X(0)
+
+        @qjit(autograph=True, target="jaxpr")
+        @qml.qnode(qml.device("null.qubit", wires=0))
+        def circuit():
+            qml.adjoint(template())
+            return qml.state()
+
+        assert circuit.jaxpr is not None
```
paul0403, line 2115: Should we check execution results?

frontend/catalyst/autograph/ag_primitives.py
```diff
@@ -549,8 +549,8 @@ def converted_call(fn, args, kwargs, caller_fn_scope=None, options=None):
             if not args:
                 raise ValueError(f"{fn.__name__} requires at least one argument")
 
-            # If first argument is already an operation, pass it through directly
-            if isinstance(args[0], qml.operation.Operation):
+            # If first argument is already an operator, pass it through directly
+            if isinstance(args[0], qml.operation.Operator):
```
paul0403, line 553: So I guess this PR just prevents the crash, but code targeted by `qml.prod` would still fallback to native python interpreted control flow?

paul0403: Also, since support for `qml.prod` needed to be added in PL, we might need to gradually expand here, since there will be other `qml.Operator` that's not supported now?
```suggestion
            if args[0] in (qml.operation.Operation, qml.prod):
```
I'm not sure about the "`isintance` checking for multiple types" syntax, but you get the idea.

dime10: > So I guess this PR just prevents the crash, but code targeted by qml.prod would still fallback to native python interpreted control flow?

Not quite, it's just that code targeted by qml.prod is invisible to autograph (full support for transforming the target of qml.prod is a separate bug).

dime10: > Also, since support for qml.prod needed to be added in PL

That's not quite what happened. I didn't _add support_ in PL for anything, I ensured that wrappers in general weren't misidentified by autograph. While I might have missed some, I tried to apply to fix to all relevant wrappers I could find.

> we might need to gradually expand here, since there will be other qml.Operator that's not supported now?

Given the above, I don't see a good reason to add these one by one. It appears that the original assumption that only `Operations` would be targeted by `qml.adjoint` and similar functions is evidently wrong, and `Operators` are are also allowed. 
@albi3ro Can you confirm if this is common in PL to have `Operators` (that are not also `Operations`) act as gates in a circuit? `qml.prod` is the first one we seem to have come across.

frontend/test/pytest/test_autograph.py
dime10: Do you see a reason to? 

Here I'm just checking that program capture succeeded (because that's the bug I'm fixing, an error during program capture), in an effort to keep the test as unit-y as possible.

paul0403: > Do you see a reason to?

I just feel safer with more tests :laughing:

frontend/catalyst/autograph/ag_primitives.py
paul0403: > I didn't _add support_ in PL for anything, I ensured that wrappers in general weren't misidentified by autograph.

Ah ok, this is what I misunderstood. All good then!

dime10:
Waiting on #1897 to be merged to fix the latest/latest failure introduce in last night's pennylane version.

frontend/test/pytest/test_autograph.py
dime10: Just trying to avoid the old paradigm of making everything a system level test ^^

paul0403: I just realized, maybe this test is better suited as a lit test?

dime10: I'm not checking IR, and the main advantage of using lit is access to filecheck. The main disadvantage of using lit is that it doesn't contribute to coverage. So to me it looks like there are only downsides and no upsides?

paul0403: > I'm not checking IR

I see now, the purpose is just checking that autograph did not crash and a jaxpr was produced. All good then :+1:

PR #1916, opened 2025-07-17
Update llvm, 2025 Q3

paul0403:
**Context:**
Update llvm, mhlo and enzyme, 2025 Q3.
The latest pair of good versions, indicated by mhlo, is https://github.com/tensorflow/mlir-hlo/commit/1dd2e71331014ae0373f6bf900ce6be393357190
```
mhlo=1dd2e71331014ae0373f6bf900ce6be393357190
llvm=f8cb7987c64dcffb72414a40560055cb717dbf74
```

For Enzyme, we go to the latest release https://github.com/EnzymeAD/Enzyme/releases/tag/v0.0.186
```
enzyme=v0.0.186
```
with commit `8c1a596158f6194f10e8ffd56a1660a61c54337e`

Miscellaneous:
1. `GreedyRewriteConfig.stuff = blah` -> `GreedyRewriteConfig.setStuff(blah)` https://github.com/llvm/llvm-project/pull/137122
2. llvm gep op `inbounds` attribute is subsumed under a gep sign wrap enum flag https://github.com/llvm/llvm-project/pull/137272
3. `arith::Constant[Int, Float]Op` builders now have the same argument order as other ops (output type first, then arguments) https://github.com/llvm/llvm-project/pull/144636 (note that Enzyme also noticed this https://github.com/EnzymeAD/Enzyme/pull/2379 :laughing: )
4. The `lookupOrCreateFn` functions now take in a builder instead of instantiating a new one https://github.com/llvm/llvm-project/pull/136421 
5. `getStridedElementPtr` now takes in `rewriter` as the first argument (instead of the last), like all the other utils https://github.com/llvm/llvm-project/pull/138984
6. The following functions now return a `LogicalResult`, and will be caught by warnings as errors as `-Wunused-result`:
   - `func::FuncOp.[insert, erase]Argument(s)` https://github.com/llvm/llvm-project/pull/137130
   - `getBackwardSlice()` https://github.com/llvm/llvm-project/pull/140961

Things related to `transform.apply_registered_pass` op:
1. It now takes in a `dynamic_options` https://github.com/llvm/llvm-project/pull/142683. We don't need to use this as all our pass options are static.
2. The options it takes in are now dictionaries instead of strings https://github.com/llvm/llvm-project/pull/143159

Bufferization:
1. `bufferization.to_memref` op is renamed to `bufferization.to_buffer` https://github.com/llvm/llvm-project/pull/137180
3. `bufferization.to_tensor` op's builder now needs the result type to be explicit https://github.com/llvm/llvm-project/pull/142986. This is also needed by a patched mhlo pass.
4. The `getBuffer()` methods take in a new arg for `BufferizationState` https://github.com/llvm/llvm-project/pull/141019, https://github.com/llvm/llvm-project/pull/141466
5. `UnknownTypeConverterFn` in bufferization options now takes in just a type instead of a full value https://github.com/llvm/llvm-project/pull/144658

**Related GitHub Issues:** 
[sc-95176]
[sc-95664]


paul0403:
Trying CI.

mlir/lib/Gradient/Transforms/GradMethods/PS_FunctionShifting.cpp
```diff
@@ -35,7 +35,7 @@ static Value genSelectiveShift(PatternRewriter &rewriter, Location loc, Value pa
     }
 
     // Make sure all active iteration variables match the selectors.
-    Value shiftCondition = rewriter.create<arith::ConstantIntOp>(loc, true, 1);
+    Value shiftCondition = rewriter.create<arith::ConstantIntOp>(loc, 1, true);
```
paul0403, line 38: TODO: there are 3 `arith::ConstantInt` buidler calls in the mhlo patched passes. We might need to update those too.

.dep-versions
```diff
@@ -2,9 +2,9 @@
 # To update JAX version alongside compatible dependency tags, run the following script:
 # python3 .github/workflows/set_dep_versions.py {JAX_version}
 jax=0.6.2
-mhlo=617a9361d186199480c080c9e8c474a5e30c22d1
-llvm=179d30f8c3fddd3c85056fd2b8e877a4a8513158
-enzyme=v0.0.180
+mhlo=1dd2e71331014ae0373f6bf900ce6be393357190
+llvm=f8cb7987c64dcffb72414a40560055cb717dbf74
+enzyme=v0.0.186
```
paul0403, line 7: changelog

paul0403:
28 pytest failures at the previous commit.

mlir/Makefile
```diff
@@ -97,6 +97,11 @@ mhlo:
 	@if cd mlir-hlo; git apply --check $(MK_DIR)/patches/mhlo-add-back-necessary-passes.patch; then \
 		git apply $(MK_DIR)/patches/mhlo-add-back-necessary-passes.patch; \
 	fi
+
+	# Patch a MHLO bug with std::sort
```
paul0403, line 109: Remember to patch this in github wheels scripts too

mlir/patches/mhlo-rename-sort.patch
```diff
@@ -0,0 +1,33 @@
...
++static void mysort(const Vec<Node*>&, Vec<int32_t>* delta);
+ static void moveToList(GraphCycles::Rep* r, Vec<int32_t>* src,
+                        Vec<int32_t>* dst);
+ static void clearVisitedBits(GraphCycles::Rep* r, const Vec<int32_t>& nodes);
+@@ -199,8 +199,8 @@ static void backwardDfs(GraphCycles::Rep* r, int32_t n, int32_t lowerBound) {
+ // Recomputes rank assignments to make them compatible with the edges (producer
+ // has smaller rank than its consumer)
+ static void reorder(GraphCycles::Rep* r) {
+-  sort(r->nodes, &r->deltab);
+-  sort(r->nodes, &r->deltaf);
```
erick-xanadu, line 10: I think this is equivalent:

```diff
diff --git a/utils/cycle_detector.cc b/utils/cycle_detector.cc
index e3901ae88..890f39654 100644
--- a/utils/cycle_detector.cc
+++ b/utils/cycle_detector.cc
@@ -199,8 +199,8 @@ static void backwardDfs(GraphCycles::Rep* r, int32_t n, int32_t lowerBound) {
 // Recomputes rank assignments to make them compatible with the edges (producer
 // has smaller rank than its consumer)
 static void reorder(GraphCycles::Rep* r) {
-  sort(r->nodes, &r->deltab);
-  sort(r->nodes, &r->deltaf);
+  mlir::sort(r->nodes, &r->deltab);
+  mlir::sort(r->nodes, &r->deltaf);
 
   // Adds contents of delta lists to list (backwards deltas first).
   r->list.clear();
```

frontend/catalyst/jax_extras/lowering.py
```diff
@@ -165,3 +166,54 @@ def custom_lower_jaxpr_to_module(
                 worklist += [*op.body.operations]
 
     return ctx.module, ctx.context
+
+
+with ir.Context() as ctx:
+    with ir.Location.unknown(ctx) as loc:
```
erick-xanadu: You can put these two lines in the same line.

```suggestion
with ir.Context() as ctx, ir.Location.unknown(ctx) as loc:
```

I've encountered issues in the past where types from different contexts are not compared correctly. E.g., i64 from ctx1 was different form i64 from ctx2. This code is not used anywhere because none of our passes have options in them at the moment. I see the tests, but I wonder whether an integration test would be more useful.

Also, I think this method would benefit from pattern matching. Look into the `match` and `case` statements.

mlir/patches/mhlo-rename-sort.patch
paul0403: I think this is the same kind of fix as just renaming? as in explicitly forcing the name?

frontend/catalyst/jax_extras/lowering.py
paul0403: > Also, I think this method would benefit from pattern matching. Look into the `match` and `case` statements.

matching against the type looks weird, you'd have to "call" the types
```python
t = 12.0
match t:
  case int():
    print("int")
  case float():
    print("float")
```

so I went with the if statements :sweat_smile:

erick-xanadu: I think you could (but I am not sure) do the following:

```python
t = 12.0
match type(t):
  case int:
    print("int")
  case float:
    print("float")
```

mlir/patches/mhlo-rename-sort.patch
erick-xanadu: Yes.

frontend/catalyst/jax_extras/lowering.py
paul0403: No, this gives 
```
  File "/home/paul.wang/catalyst_new/catalyst/hi.py", line 3
    case int:
         ^^^
SyntaxError: name capture 'int' makes remaining patterns unreachable
```

dime10: `match`-`case` is actually a super cool feature in Python! 

Inside those "weird" parenthesis you can put parameters which will match against properties you want the matched object to have.
The reason you are putting parenthesis is because `t` is an _instance of_ `int` or `float`, it is not _one of_ `int` or `float`.

It's certainly a lot nicer than a chain of isinstance checks 😅

erick-xanadu: That's true. Thank you! Not to make it more confusing:

```python
import builtins
t = 12.0
match type(t):
  case builtins.int:
    print("int")
  case builtins.float:
    print("float")
```

erick-xanadu: But I agree, the syntax is different from what has been used in the project.

paul0403: > I've encountered issues in the past where types from different contexts are not compared correctly. E.g., i64 from ctx1 was different form i64 from ctx2.

But we are not comparing types here right? 

And if we want to be safe, I can pass the same `Context()` instance into the recursive calls. Or is it done already since the recursive calls are all under the same `with` block?

erick-xanadu: I think jax provides the context it has already instantiated via the `LoweringRuleContext` type which is usually the first parameter in the lowering rules. I would reuse that one.

I believe comparing types is just a symptom. What would it mean when you have a type i64 but then you encounter something else that is i64 but not the same i64? Let me try to find the concrete example...

paul0403: Just a note, recursive calls for a function under the same `with` do not create new `with` context objects, so we are not actually creating new jax `ir.Context()` instances here
```python
class Thing:
   def __init__(self, thing):
      self.thing = thing
   def __enter__(self):
      return self.thing
   def __exit__(self, _, __, ___):
      return

with Thing("apple") as apple:
   def foo(i):
      print(id(apple))
      if i==0:
         return 0
      foo(i-1)

foo(3)
```

```
130706228697584
130706228697584
130706228697584
130706228697584
```

erick-xanadu: Yes, I agree. But I am still worried about using a new Context as opposed to the one provided by JAX. Maybe I am wrong though. Here is the ticket that has a comment regarding precomputing types and using them in the JAX modules: https://github.com/PennyLaneAI/catalyst/pull/610#discussion_r1556672108

erick-xanadu: Essentially, if we remove the `with Context` and place it in the tests, it would allow for the context used by JAX to be used normally (I think?) Which context do you think would be used?

paul0403: Ah I see what you mean. Then I think what we should do is:
- This util function takes in a required jax lowering context (can location still just be a blanket `ir.Location.unknown()`?)
- We pass in the context when actually using it. For the `apply_registered_op`'s lowering we pass the true context for that primitive's lowering, and for the unit test of this util itself we pass a fake `ir.Context()`

erick-xanadu: I think you don't need the context not the Location unknown if it already is under the context and location. So I think you can just remove the contextmanagers and keep the definition, but in the tests you have the context managers.

No need to pass the context directly. I believe the context argument is optional when under the context manager.

erick-xanadu: See for example: frontend/catalyst/utils/types.py and frontend/test/pytest/test_types.py

paul0403: https://github.com/PennyLaneAI/catalyst/pull/1916/commits/3a8fdcea5f97d0ca74627f4ea90f9e0edf1928b0 I think I understand it now, yes I noticed the same, the methods in `ir` do not need contexts and locations to be passed in when there is one instance of it already

erick-xanadu:
> It now takes in a dynamic_options
https://github.com/llvm/llvm-project/pull/142683. We don't need to use this as all our pass options are static.
The options it takes in are now dictionaries instead of strings
https://github.com/llvm/llvm-project/pull/143159

Thank you for pointing this out. I'll try to address these tomorrow in xDSL.

erick-xanadu:
It now takes in a dynamic_options
https://github.com/llvm/llvm-project/pull/142683. We don't need to use this as all our pass options are static.
The options it takes in are now dictionaries instead of strings
https://github.com/llvm/llvm-project/pull/143159It now takes in a dynamic_options
https://github.com/llvm/llvm-project/pull/142683. We don't need to use this as all our pass options are static.
The options it takes in are now dictionaries instead of strings
https://github.com/llvm/llvm-project/pull/143159

frontend/catalyst/jax_extras/lowering.py
```diff
@@ -165,3 +166,49 @@ def custom_lower_jaxpr_to_module(
                 worklist += [*op.body.operations]
 
     return ctx.module, ctx.context
+
+
+def get_mlir_attribute_from_pyval(value):
```
mehrdad2m, line 172: This looks like a very nice utility function. Do you think it could belong to the `utils` folder or is it too specific to this file?

mlir/Makefile
```diff
@@ -204,6 +219,11 @@ clean-dialects:
 clean-llvm:
 	@echo "clean llvm/mlir build files"
 	rm -rf $(LLVM_BUILD_DIR)
+	cd llvm-project; git clean -fd; git checkout .
```
mehrdad2m, line 220: should we add the same logic for the other pathed repos (enzyme, mhlo) for consistency?

frontend/catalyst/jax_extras/lowering.py
erick-xanadu: I have a file which contains similar code named types.py. You could also think of upstreaming something like this to the MLIR repo. (which will then be picked up by JAX since they use the Python bindings) Maybe somewhere around this [file](https://github.com/llvm/llvm-project/blob/main/mlir/python/mlir/extras/types.py)

mlir/Makefile
paul0403: I already have this for mhlo. I will add this for enzyme as well.

frontend/catalyst/jax_extras/lowering.py
paul0403: > This looks like a very nice utility function. Do you think it could belong to the `utils` folder or is it too specific to this file?

I think this is specific to this file, this is a helper just for lowering jax primitive params (which are pure python values) to mlir attributes, so I thought it naturally belonged in the `jax_extras/lowering.py`. This file already _is_ the util file for lowering I think.

paul0403:
@mehrdad2m @dime10 this is now completely ready

frontend/catalyst/jax_extras/lowering.py
dime10: This is a good point, unless it's truly generic, we should avoid separating code from its intended context (which a utils-style organization is prone to). For example, the `gen_mlir.py` module in the utils folder doesn't seem generic at all, it's very specific to our lowering process and should probably belong there (unfortunately the refactor of all tracing and jaxpr lowering functionality into a proper sub-module was never completed). 

I think the same applies for many other modules in the utils folder right now, with the exception of something like `patching.py` which seems truly generic. Imo we should be looking to reduce the amount of code in the utils folder and work towards more cohesive sub-modules in the catalyst package (tracing, lowering, compilation, user api functions, etc).

While creating mlir types from python types does seem useful, the only realistic use case I can see for this right now is when we lower jaxpr to mlir (even xdsl wouldn't use this as it doesn't use the Python bindings). Given this, I would say such a function would ideally live in the bindings themselves (like Erick is suggesting), and barring that in our lowering code.

paul0403: > Given this, I would say such a function would ideally live in the bindings themselves (like Erick is suggesting)

I agree. Given that jax does not offer this util I do think the best place to add it is the `lowering.py` because I also agree that this util won't be used outside jaxpr-to-mlir lowering.

mlir/Makefile
paul0403: https://github.com/PennyLaneAI/catalyst/pull/1916/commits/2038103b64d272efe154ffbd4faa87fa3b767c8b

frontend/catalyst/passes/pass_api.py
```diff
@@ -287,22 +288,17 @@ def __init__(self, name: str, *options: list[str], **valued_options: dict[str, s
     def get_options(self):
         """
         Stringify options according to what mlir-opt expects.
```
mehrdad2m:
```suggestion
        Build a dictionary mapping option names to MLIR attributes.
```

.github/workflows/build-wheel-linux-arm64.yaml
```diff
@@ -120,8 +128,15 @@ jobs:
       if: steps.cache-mhlo-source.outputs.cache-hit != 'true'
       run: |
         cd $GITHUB_WORKSPACE/mlir/mlir-hlo
-        git apply $GITHUB_WORKSPACE/mlir/patches/mhlo-remove-shardy.patch
-        git apply $GITHUB_WORKSPACE/mlir/patches/mhlo-add-back-necessary-passes.patch
+        if git apply --check $GITHUB_WORKSPACE/mlir/patches/mhlo-remove-shardy.patch; then \
```
dime10: This change was intentional. Patching a freshly cloned source should _always_ apply the patch, if the patch application fails we should know about it

mlir/lib/Catalyst/Transforms/GEPInboundsPatterns.cpp
```diff
@@ -29,11 +29,12 @@ struct GEPOpRewritePattern : public mlir::OpRewritePattern<LLVM::GEPOp> {
                                         mlir::PatternRewriter &rewriter) const override
     {
         auto defOp = op.getBase().getDefiningOp();
-        if (op.getInbounds() || (defOp && isa<LLVM::ZeroOp>(defOp))) {
+        if (op.getNoWrapFlags() == LLVM::GEPNoWrapFlags::inbounds ||
```
mehrdad2m, line 32:
```suggestion
        if (op.getNoWrapFlags() & LLVM::GEPNoWrapFlags::inbounds ||
```
When comparing flags, usually bitwise AND is preferred, since there might be other flags that are also set. Also, there is a [isInBounds](https://llvm.org/doxygen/classllvm_1_1GEPNoWrapFlags.html#aaba6873d157dbd7f5f02da723e6ca78f) function that I think does the same thing.

frontend/catalyst/jax_extras/lowering.py
```diff
@@ -165,3 +166,53 @@ def custom_lower_jaxpr_to_module(
...
+
+    attr = None
+    match value:
+        case bool():
+            attr = ir.BoolAttr.get(value)
+
+        case int():
+            if value < 0:
+                attr = ir.IntegerAttr.get(ir.IntegerType.get_signed(64), value)
+            else:
```
mehrdad2m: A little nit-picky but since `value` is in python, it can potentially be outside the range if a i64, so would be nice to check the width of it as well.

mlir/lib/Catalyst/Transforms/GEPInboundsPatterns.cpp
paul0403: I think this `isInBounds` is for the llvm gep instruction, not the mlir gep op

paul0403: As for changing to bitwise AND, I'm getting 
```
/home/paul.wang/catalyst_new/catalyst/mlir/lib/Catalyst/Transforms/GEPInboundsPatterns.cpp:32:13: error: value of type 'mlir::LLVM::GEPNoWrapFlags' is not contextually convertible to 'bool'
        if (op.getNoWrapFlags() & LLVM::GEPNoWrapFlags::inbounds ||
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/paul.wang/catalyst_new/catalyst/mlir/lib/Catalyst/Transforms/GEPInboundsPatterns.cpp:32:66: error: invalid operands to binary expression ('mlir::LLVM::GEPNoWrapFlags' and 'bool')
        if (op.getNoWrapFlags() & LLVM::GEPNoWrapFlags::inbounds ||
```

frontend/catalyst/jax_extras/lowering.py
```diff
@@ -165,3 +166,53 @@ def custom_lower_jaxpr_to_module(
...
+                    raise CompileError(
+                        f"Dictionary keys for MLIR DictionaryAttr must be strings, got: {type(k)}"
+                    )
+                named_attrs[k] = get_mlir_attribute_from_pyval(v)
+            attr = ir.DictAttr.get(named_attrs)
+
+        case None:
+            # MLIR has a UnitAttr for representing a void or "none" value
+            # TODO: is `None` the best flag here?
+            attr = ir.UnitAttr.get()
```
dime10: That's not how I interpret the unit attr. It's a singleton value for attributes where the mere presence of it is meaningful. For instance, an optional bool attribute can have three states (absent, False, True). Often times, you don't need this though and two states are sufficient (absent, present). 
Note that the unit attribute represents the "positive" state here (present), not the "negative" or absent state like Void or None typically do, whereas this branch suggests the opposite.

dime10 (changes requested):
Thanks @paul0403, just two points:

frontend/catalyst/jax_extras/lowering.py
paul0403: https://github.com/PennyLaneAI/catalyst/pull/1916/commits/d82d7ace00ab9f0e890270973071abf48cb9b174

.github/workflows/build-wheel-linux-arm64.yaml
paul0403: I thought the checking was for `make mhlo` twice? So the second time the patch would fail but we don't want the makefile to pause?

paul0403: Oh nevermind, this is wheels script, then this should be in response to issue #1919 ?

frontend/catalyst/jax_extras/lowering.py
paul0403: > Often times, you don't need this though and two states are sufficient (absent, present).

Yes I agree. 

This is why I put this TODO here, it felt a bit weird. I wanted this util to cover unitattrs, but I couldn't figure out the best flag this util could take in as the "unitattr" flag. Given the name `get_mlir_attribute_from_pyval`, it suggests that it takes in the corresponding python value for unitattr, which doesn't really exist. Python doesn't have a "unit" type whose value is not important and only the existence/absense itself is important (or does it?).

Thoughts about removing the unitattr case from the util completely? Since there is no `pyval` for unitattr in the first place?

.github/workflows/build-wheel-linux-arm64.yaml
dime10: Yes that issue was fixed by this PR no? https://github.com/PennyLaneAI/catalyst/pull/1924

frontend/catalyst/jax_extras/lowering.py
dime10: Sounds good to me 👍

paul0403: https://github.com/PennyLaneAI/catalyst/pull/1916/commits/0b17c6c538ff16f5d845cb4e69a7f4f631ac497a

.github/workflows/build-wheel-linux-arm64.yaml
paul0403: I see, I just thought this issue was still open. Reverted in https://github.com/PennyLaneAI/catalyst/pull/1916/commits/dc072e1641bce027c98e266d14b27e0ed0b59c0d

mlir/lib/Catalyst/Transforms/GEPInboundsPatterns.cpp
mehrdad2m: Ah I see, I assumed they are gonna be implemented the same way, but apparently not. Sounds good then

mlir/patches/mhlo-rename-sort.patch
dime10: Would you mind using this patch instead? Seems cleaner :)

mlir/python/CMakeLists.txt
```diff
@@ -3,6 +3,8 @@ include(AddMLIRPython)
 # TODO: Add an upstream cmake param for this vs having a global here.
 add_compile_definitions("MLIR_PYTHON_PACKAGE_PREFIX=mlir_quantum.")
 
+# Ignore nanobind warnings
+add_compile_options(-w)
```
dime10, line 7: Out of curiosity, is this what Lee recommended in the end? I thought he had a somewhat different trick.

Separately though, I would just like to ensure a blind warning disable like this is very targeted. For instance with Stim, I used the following:
```
target_compile_options(libstim PRIVATE -w)
```
Are we sure the line above doesn't affect code we'd like to have warnings on?

mlir/patches/mhlo-rename-sort.patch
paul0403: https://github.com/PennyLaneAI/catalyst/pull/1916/commits/81b0ab66b2449395287c61d05c5750df6eeb69d2

mlir/python/CMakeLists.txt
paul0403: I had to do a blanket `add_compile_options` because there are no targets in this cmake file, there's just calls to mlir's cmake functions here. 

Lee is ok with this, I'll tag you in the slack thread.

dime10 (approved):
🚀

mehrdad2m (approved):

PR #1921, opened 2025-07-18
Migrate to stablehlo

paul0403:
**Context:**
Migrate to stablehlo.

Catalyst depends on the stablehlo dialect because that is what jax lowers classical primitives (additions, jnp array ops, linalg ops, etc) to. The stablehlo dialect was pulled in from the [tensorflow/mlir-hlo](https://github.com/tensorflow/mlir-hlo) repo, but [that repo is slated for deletion](https://groups.google.com/a/openxla.org/g/openxla-discuss/c/Mppuv1Edv1s) .
They suggest users to migrate to [the stablehlo repo](https://github.com/openxla/stablehlo). 

Use the stablehlo version of the mhlo operations, dialects and passes we are currently using.
Remove mhlo build system from regular makefiles and CI scripts.
Remove mhlo patches. 

**Benefits:**
No longer depending on a sunsetting external project.

[sc-96641]


dime10:
> I'm opening CI to check the build system. As of this commit ([ee515f5](https://github.com/PennyLaneAI/catalyst/commit/ee515f5f2af0ac8bb73a519503ac726909021ecc)), everything should succeed up until the python tests (i.e. everything should succeed up until `make test-mlir`).
> 
> I'm still building stablehlo in embedded mode instead of external mode, but the [strange logic in stablehlo's cmakelists](https://github.com/openxla/stablehlo/blob/main/CMakeLists.txt#L69) makes it seem impossible to use their external mode. Embedded mode, aka building the `stablehlo` directory just as another subdirectory in the overall `mlir` directory, is just the easiest mode to get things going for now.
> 
> I will now move on to check the missing passes. We can revisit build mode later.

Thanks @paul0403! What happened to the stdlib error?

.github/workflows/check-catalyst.yaml
```diff
@@ -406,13 +314,20 @@ jobs:
         key: ${{ runner.os }}-ccache-${{ github.run_id }}
         restore-keys: ${{ runner.os }}-ccache-
 
+    # just hard code commit manually, set up this stablehlo dep verisons infra later
```
paul0403: TODO: setup CI infra and stuff

paul0403:
Regular CI is all green at the previous commit https://github.com/PennyLaneAI/catalyst/pull/1921/commits/3f317e8246ba7391423454b7ebad79c733522606, checking wheels now.

mlir/CMakeLists.txt
```diff
@@ -4,7 +4,8 @@ project(Catalyst LANGUAGES CXX C)
 set(CMAKE_BUILD_WITH_INSTALL_NAME_DIR ON)
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
-set(CMAKE_CXX_STANDARD 20 CACHE STRING "C++ standard to conform to")
+# stablehlo is still on cpp17
+set(CMAKE_CXX_STANDARD 17 CACHE STRING "C++ standard to conform to")
```
paul0403: Is there a way to use cpp 20 everywhere in our mlir passes, but only for the `add_subdirectory(stablehlo)` subdirectory use cpp17? @mlxd

.github/workflows/build-wheel-linux-arm64.yaml
```diff
@@ -360,6 +285,13 @@ jobs:
           PYTHON=$(which python${{ matrix.python_version }}) \
           make oqc
 
+    - name: Clone Stablehlo Submodule
```
dime10: Does it not participate in caching?

dime10: Oh, the build is now part of the catalyst build so has to be repeated every time right? That's certainly unfortunate 🤔 

I'd like to confirm there is nothing we can do about this before going ahead

.github/workflows/set_dep_versions.py
```diff
@@ -71,7 +71,7 @@
 # Update each version using sed
 cmds = [
     f"sed -i '' 's/^jax=.*/jax={jax_version}/' {dep_versions_path}",
-    f"sed -i '' 's/^mhlo=.*/mhlo={hlo_commit}/' {dep_versions_path}",
+    f"sed -i '' 's/^stablehlo=.*/stablehlo={hlo_commit}/' {dep_versions_path}",
```
dime10: Did you test this change? I think that's a commit from a different repo (tensorflow), and we now need to determine a corresponding commit in the stablehlo repo.

mlir/CMakeLists.txt
dime10: I don't think it's big deal unless we use c++20 features 🤔

mlir/include/Catalyst/Transforms/Passes.td
```diff
@@ -61,7 +61,7 @@ def ScatterLoweringPass : Pass<"scatter-lowering"> {
     let dependentDialects = [
         "mlir::func::FuncDialect",
         "index::IndexDialect",
-        "mhlo::MhloDialect",
+        "stablehlo::StablehloDialect",
```
dime10: Would be nice to move this pass into the hlo dialect directories as well :)

mlir/CMakeLists.txt
```diff
@@ -29,35 +30,23 @@ endif()
...
-  MhloToStablehlo
-  StablehloToMhlo
-)
-
 list(APPEND CMAKE_MODULE_PATH "${MLIR_CMAKE_DIR}")
 list(APPEND CMAKE_MODULE_PATH "${LLVM_CMAKE_DIR}")
-list(APPEND CMAKE_MODULE_PATH "${MHLO_CMAKE_DIR}")
 list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake/modules")
 
+unset(LLVM_USE_LINKER)
```
dime10: Add a comment for this patch

frontend/catalyst/pipelines.py
```diff
@@ -226,13 +226,13 @@ def get_hlo_lowering_stage(_options: CompileOptions) -> List[str]:
...
-        "func.func(chlo-legalize-to-hlo)",
-        "stablehlo-legalize-to-hlo",
-        "func.func(mhlo-legalize-control-flow)",
-        "func.func(hlo-legalize-to-linalg)",
-        "func.func(mhlo-legalize-to-std)",
-        "func.func(mhlo-legalize-sort)",
-        "convert-to-signless",
+        "func.func(chlo-legalize-to-stablehlo)",
+        "func.func(stablehlo-legalize-control-flow)",
+        "func.func(stablehlo-aggressive-simplification)",
```
dime10, line 231: New pass? :)

mlir/tools/quantum-opt/CMakeLists.txt
```diff
@@ -21,12 +21,11 @@ set(LIBS
     mitigation-transforms
     MLIRIon
     ion-transforms
-    MhloRegisterDialects
     StablehloRegister
+    StablehloCAPI
```
dime10: Huh what is the capi doing here?

dime10 (changes requested):
Nice work 🚀 🚀

.github/workflows/build-wheel-linux-arm64.yaml
paul0403: Yes, I thought about caching the source but that's not really meaningful, instead of download from their repo we'd just download from cache which is kinda pointless :sweat_smile: so I just didn't bother adding it

dime10: If the source code is very large (like with llvm) it actually makes a difference, but I agree here it's not so important. More concerned about the build.

mlir/tools/quantum-opt/CMakeLists.txt
paul0403: `mlirRegisterAllStablehloPasses` is here

https://github.com/openxla/stablehlo/blob/e07debd5e257ec1e118f18c54068977b89f03b2f/stablehlo/integrations/c/StablehloPasses.cpp#L18
https://github.com/openxla/stablehlo/blob/e07debd5e257ec1e118f18c54068977b89f03b2f/stablehlo/integrations/c/CMakeLists.txt#L33

frontend/catalyst/pipelines.py
paul0403: @mehrdad2m discovered this was needed before the `legalize-to-linalg`. Mehrdad do you remember why?

mlir/include/Catalyst/Transforms/Passes.td
paul0403: Good point!

.github/workflows/build-wheel-linux-arm64.yaml
paul0403: No longer relevant now that we moved away from embedded build

mlir/CMakeLists.txt
paul0403: No longer relevant in non-embedded build

.github/workflows/build-wheel-linux-arm64.yaml
```diff
@@ -109,25 +109,18 @@ jobs:
...
-
-    - name: Patch MHLO Source
-      if: steps.cache-mhlo-source.outputs.cache-hit != 'true'
-      run: |
-        cd $GITHUB_WORKSPACE/mlir/mlir-hlo
-        git apply $GITHUB_WORKSPACE/mlir/patches/mhlo-remove-shardy.patch
-        git apply $GITHUB_WORKSPACE/mlir/patches/mhlo-rename-sort.patch
+        repository: openxla/stablehlo
+        ref: ${{ needs.constants.outputs.stablehlo_version }}
+        path: mlir/stablehlo
```
mehrdad2m: should we add a ${{ github.workspace }} to a avoid relative path

.github/workflows/build-wheel-linux-x86_64.yaml
```diff
@@ -128,25 +128,18 @@ jobs:
...
-
-    - name: Patch MHLO Source
-      if: steps.cache-mhlo-source.outputs.cache-hit != 'true'
-      run: |
-        cd $GITHUB_WORKSPACE/mlir/mlir-hlo
-        git apply $GITHUB_WORKSPACE/mlir/patches/mhlo-remove-shardy.patch
-        git apply $GITHUB_WORKSPACE/mlir/patches/mhlo-rename-sort.patch
+        repository: openxla/stablehlo
+        ref: ${{ needs.constants.outputs.stablehlo_version }}
+        path: mlir/stablehlo
```
mehrdad2m: Also here and other places in all workflows

.github/workflows/set_dep_versions.py
paul0403: Good point, I just assumed it would be fine, but I tested it now and it breaks :sweat_smile: 

Do we even care about updating this now? Because I think we are trying to decouple the llvm commit from the jax version?

mlir/CMakeLists.txt
paul0403: We can now go back to cpp20 since stablehlo is no longer embedded!

mlir/include/Catalyst/Transforms/Passes.td
paul0403: https://github.com/PennyLaneAI/catalyst/pull/1921/commits/5dc59ab60fa4cc21beb0c7fcbb3f103e8aa2a184

paul0403: ... and https://github.com/PennyLaneAI/catalyst/pull/1921/commits/c910f32983c213c4a254960a6d271211bd808b73

.github/workflows/build-wheel-linux-arm64.yaml
paul0403: https://github.com/PennyLaneAI/catalyst/pull/1921/commits/2f96459090a99958a0a34551d2987199e93b90ce

I probably don't have time to add this for all the other submodules, I just did it for my stablehlo changes. Maybe someone next week can cleanup the other ones too :)

.github/workflows/set_dep_versions.py
paul0403: > Because I think we are trying to decouple the llvm commit from the jax version?

The current llvm commit we track is not the one that's fetched from 0.6.2 by this python script, it was just the latest good llvm/hlo pair when I made the update.

dime10: > it was just the latest good llvm/hlo pair when I made the update

What is the "latest good" llvm/hlo pair? Are you saying you used the last commit on the llvm main branch that the hlo repo was referencing? (that is it's not related to the llvm version used by jax?)

dime10: > Because I think we are trying to decouple the llvm commit from the jax version?

Good point, are we already using an official release tag of llvm then? If so I think we can remove this script, but we shouldn't have broken code in the rep. If not, then we could repair or remove it depending on our plans.

mlir/include/Catalyst/Transforms/Passes.td
dime10: Nice 👍

mlir/lib/stablehlo/HloCustomCallPatterns.cpp
```diff
@@ -14,22 +14,22 @@
 
 #define DEBUG_TYPE "scatter"
 
+#include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/IR/PatternMatch.h"
+#include "stablehlo/dialect/StablehloOps.h"
+#include "llvm/Support/Debug.h"
```
dime10: L before S 😅  (thanks for fixing the includes though, I wish clang format would do a better job at it)
```suggestion
#include "llvm/Support/Debug.h"
#include "stablehlo/dialect/StablehloOps.h"
```

mlir/lib/stablehlo/scatter_lowering.cpp
```diff
@@ -28,15 +28,16 @@
 #include "mlir/Pass/Pass.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 
-#include "Catalyst/Transforms/Patterns.h"
+#include "stablehlo/Passes.h"
+#include "stablehlo/Patterns.h"
```
dime10: Oh yeah this is awkward to have both internal and external stablehlo headers 😅

paul0403: Opinions on renaming the internal folder to `stablehlo-catalyst`?

dime10: Hmm, a bit unwieldy 🤔 Wait, aren't our dialect folders usually capitalized? Technically it would be `StableHLO` then, but we could also drop the "stable" and just call it `HLO`.
Or, if we want to be more explicit, `HLOExtensions`?
Alternatively, if we do want to distinguish it from the "proper" dialects in our repo I guess `hlo-extensions` also wouldn't be that bad.

.github/workflows/set_dep_versions.py
paul0403: > What is the "latest good" llvm/hlo pair? 

So I go to stablehlo repo and look at their commits https://github.com/openxla/stablehlo/commits/main/

Every once in a while they have a "Integrate with llvm version `<llvm commit tag>`" commit. I just use the latest pair over there. I actually already used this during the llvm update. So...

> (that is it's not related to the llvm version used by jax?)

... yes this is correct.


> are we already using an official release tag of llvm then? If so I think we can remove this script,

Which means although we are not using an official release tag of llvm, we can probably still remove the script? Since the procedure to get the correct pair of versions is not as convoluted as how jax handled it.

dime10: Thanks for the explanation! In that case, I'm happy to remove the script.

mlir/lib/stablehlo/HloCustomCallPatterns.cpp
paul0403: I don't know why but my `clang-format` keeps wanting to put `llvm` after `stablehlo`...

mlir/lib/stablehlo/scatter_lowering.cpp
paul0403: Renamed to `hlo-extensions` in https://github.com/PennyLaneAI/catalyst/pull/1921/commits/8cc592ad0509691011757758ecbb89fe79d991c1

mlir/lib/stablehlo/HloCustomCallPatterns.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/actions/runs/16970453059/job/48105717054?pr=1921

.github/workflows/set_dep_versions.py
paul0403: https://github.com/PennyLaneAI/catalyst/pull/1921/commits/8a3a4846c4bda6f3f2ec83792a98980c8d1d647b

mlir/lib/stablehlo/HloCustomCallPatterns.cpp
dime10: No worries, clang-format is king I suppose 😅

mlir/CMakeLists.txt
```diff
@@ -30,37 +30,61 @@ endif()
...
+    )
+endforeach()
+
+add_library(ExternalStablehloLib INTERFACE)
+
+foreach(STABLEHLO_LIB IN LISTS STABLEHLO_LIBS)
+    target_link_libraries(ExternalStablehloLib INTERFACE ${STABLEHLO_LIB})
+endforeach()
+
+target_include_directories(ExternalStablehloLib INTERFACE
```
dime10: Maybe add SYSTEM here?

mlir/Makefile
```diff
@@ -94,36 +94,26 @@ llvm:
...
-	cmake -G Ninja -S mlir-hlo -B $(MHLO_BUILD_DIR) \
+	cmake -G Ninja -S stablehlo -B $(STABLEHLO_BUILD_DIR) \
+		-DSTABLEHLO_ENABLE_LLD=ON \
 		-DCMAKE_BUILD_TYPE=$(BUILD_TYPE) \
-		-DLLVM_ENABLE_ASSERTIONS=ON \
 		-DMLIR_DIR=$(LLVM_BUILD_DIR)/lib/cmake/mlir \
-		-DPython3_EXECUTABLE=$(PYTHON) \
+		-DLLVM_ENABLE_ASSERTIONS=ON \
+		-DSTABLEHLO_ENABLE_BINDINGS_PYTHON=OFF \
+		-DSTABLEHLO_ENABLE_SPLIT_DWARF=ON \
```
dime10, line 109: What does this do ?

mlir/Makefile
```diff
@@ -94,36 +94,26 @@ llvm:
...
+		-DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
 		-DCMAKE_C_COMPILER=$(C_COMPILER) \
 		-DCMAKE_CXX_COMPILER=$(CXX_COMPILER) \
 		-DCMAKE_C_COMPILER_LAUNCHER=$(COMPILER_LAUNCHER) \
 		-DCMAKE_CXX_COMPILER_LAUNCHER=$(COMPILER_LAUNCHER) \
 		-DCMAKE_EXE_LINKER_FLAGS=$(USE_SANITIZER_FLAGS) \
-		-DLLVM_ENABLE_LLD=$(ENABLE_LLD) \
-		-DLLVM_ENABLE_ZLIB=$(ENABLE_ZLIB) \
-		-DLLVM_ENABLE_ZSTD=$(ENABLE_ZSTD) \
```
dime10, line 122: Does stablehlo not make use of these flags?

mlir/CMakeLists.txt
paul0403: https://github.com/PennyLaneAI/catalyst/pull/1921/commits/fdf96d2c2b06f45dbe466c0b332fa86317bab27f

.github/workflows/build-wheel-linux-arm64.yaml
```diff
@@ -207,31 +200,29 @@ jobs:
         path: ${{ github.workspace }}/llvm-build
         key: ${{ matrix.container_img }}-llvm-${{ needs.constants.outputs.llvm_version }}-${{matrix.python_version}}-wheel-build
 
-    - name: Build MHLO Dialect
-      if: steps.cache-mhlo-build.outputs.cache-hit != 'true'
-      # building with LLD is a strong requirement for mhlo
+    - name: Build Stablehlo Dialect
```
dime10, line 203: Any chance we can use the make target to build it?

mlir/lib/Driver/CMakeLists.txt
```diff
@@ -55,3 +52,5 @@ add_mlir_library(CatalystCompilerDriver
     LINK_LIBS PRIVATE
     ${LIBS}
 )
+
+target_link_libraries(CatalystCompilerDriver PRIVATE ExternalStablehloLib)
```
dime10: Can we not just include this in the list above (no big deal if not)?

mlir/lib/hlo-extensions/CMakeLists.txt
```diff
@@ -25,3 +29,4 @@ target_include_directories(${LIBRARY_NAME} PUBLIC
                            .
                            ${PROJECT_SOURCE_DIR}/include
                            ${CMAKE_BINARY_DIR}/include)
+target_link_libraries(${LIBRARY_NAME} PRIVATE ExternalStablehloLib)
```
dime10: Same here

mlir/tools/catalyst-cli/CMakeLists.txt
```diff
@@ -39,16 +39,14 @@ set(LIBS
...
-    StablehloRegister
     MLIRCatalystTest
-    ${ALL_MHLO_PASSES}
     ${ENZYME_LIB}
     CatalystCompilerDriver
 )
 
 add_mlir_tool(catalyst-cli catalyst-cli.cpp SUPPORT_PLUGINS)
 target_link_libraries(catalyst-cli PRIVATE ${LIBS})
+target_link_libraries(catalyst-cli PRIVATE ExternalStablehloLib)
```
dime10: Same here

mlir/tools/quantum-lsp-server/CMakeLists.txt
```diff
@@ -11,11 +11,10 @@ set(LIBS
     MLIRMBQC
     MLIRMitigation
     MLIRIon
-    MhloRegisterDialects
-    StablehloRegister
 )
 
 add_llvm_executable(quantum-lsp-server quantum-lsp-server.cpp)
 target_link_libraries(quantum-lsp-server PRIVATE ${LIBS})
+target_link_libraries(quantum-lsp-server PRIVATE ExternalStablehloLib)
```
dime10: And here

mlir/tools/quantum-opt/CMakeLists.txt
```diff
@@ -21,16 +21,14 @@ set(LIBS
...
-    StablehloRegister
     MLIRCatalystTest
     MLIRCatalystUtils
     MLIRTestDialect
-    ${ALL_MHLO_PASSES}
 )
 
 add_mlir_tool(quantum-opt quantum-opt.cpp DEPENDS ${LIBS} SUPPORT_PLUGINS)
 target_link_libraries(quantum-opt PRIVATE ${LIBS})
+target_link_libraries(quantum-opt PRIVATE ExternalStablehloLib)
```
dime10: And here

dime10 (approved):
Can't believe it finally happened 😍

mlir/Makefile
paul0403: These are in stablehlo's own build guide, under "if you are actively developing stablehlo" https://github.com/openxla/stablehlo?tab=readme-ov-file#build-instructions

I'm hoping we don't need to add more stablehlo patches in the future :crossed_fingers: but just in case we need to, I want to follow the development build right now instead of making people confused down the line.

mlir/Makefile
paul0403: https://github.com/PennyLaneAI/catalyst/pull/1921/commits/42edcc6599bab7f5dc52fe6abafec7bcde209942 good catch, I put them back. 

(They have the other two but don't have zstd)

.github/workflows/build-wheel-linux-arm64.yaml
paul0403: I am always annoyed that the wheels scripts call cmake and command line "raw" instead of calling the "make blah" like the usual check-catalyst script :sweat_smile: I am all in favor of everyone calling the makefile, but this would have a lot of lines of change, so we can probably do this in another PR later

mlir/Makefile
```diff
@@ -103,6 +103,8 @@ stablehlo:
 		-DCMAKE_BUILD_TYPE=$(BUILD_TYPE) \
 		-DMLIR_DIR=$(LLVM_BUILD_DIR)/lib/cmake/mlir \
 		-DLLVM_ENABLE_ASSERTIONS=ON \
+		-DLLVM_ENABLE_LLD=$(ENABLE_LLD) \
```
dime10, line 106: Is this not already handled by the `STABLEHLO_ENABLE_LLD` flag ? (actually that one needs to use the ENABLE_LLD variable as well rather than being hardcoded)

mlir/lib/Driver/CMakeLists.txt
paul0403: https://github.com/PennyLaneAI/catalyst/pull/1921/commits/5b99ac5d88e9688bd460d53489e40af902828a45

.github/workflows/build-wheel-linux-arm64.yaml
dime10: At least it's an opportunity in to change it for the new code you introduce ;)

mlir/Makefile
paul0403: > Is this not already handled by the `STABLEHLO_ENABLE_LLD` flag ?

They have both
image



> actually that one needs to use the ENABLE_LLD variable as well rather than being hardcoded

True :sweat_smile: I must have forgotten to change this back

paul0403: https://github.com/PennyLaneAI/catalyst/pull/1921/commits/bc1d58ccb0ad0825155763592a2f2c8166a779f6

.github/workflows/build-wheel-linux-arm64.yaml
paul0403: Done (after a series of small mishaps)

paul0403:
> Regarding the current build wheels errors:
> 
> 1. Could it be that `--workflow` is missing from the list of CMAKE_ARGS (`catalyst/runtime/CMakeLists.txt`, `ExternalProject_Add(lapacke-accelerate`? See current Accelerate LAPACKE [README](https://github.com/lepus2589/accelerate-lapacke?tab=readme-ov-file#workflow-with-cmake).
> 2. Would it be worth considering sticking to a given working GIT_TAG (version, UUID...), then doing regular updates for these external projects?

The current failures are caused by github mac runner images being upgraded to maxos 15. We are handling it in #1986

Issue #1586, opened 2025-03-24
Display Catalyst Version in `quantum-opt --version` Output


sengthai:
**Current Behavior:**
Running the command: 
```
quantum-opt --version
```
Currently, return:
```bash
LLVM (http://llvm.org/):
  LLVM version 19.0.0git
  Optimized build with assertions.
```
It only displays the LLVM version.

**Expected Behavior:**
It should also display the Catalyst version alongside the LLVM version. For example:

For example:
```bash
Catalyst version 0.10.0
LLVM (http://llvm.org/):
  LLVM version 19.0.0git
  Optimized build with assertions.
```

**Proposed Solution:**
* Update the version printer in `mlir/tools/quantum-opt/quantum-opt.cpp` to include the Catalyst version.
* Use the `llvm::cl::AddExtraVersionPrinter` function, as documented here: [AddExtraVersionPrinter](https://llvm.org/doxygen/namespacellvm_1_1cl.html#af482fe65d5cc4df7480d60fdf731c14e)

**Notes:**
* The Catalyst version is defined in `frontend/catalyst/_version.py`.
* The version string should be retrieved from there. If that’s not feasible in C++, an alternative such as pulling the version from the Git metadata or generating it at build time would be great.

PR #1922, opened 2025-07-20
Fix issue 1586: Display Catalyst Version in `quantum-opt --version` Output

rturrado:
**Context:**
Catalyst version is maintained at `frontend/catalyst/_version.py`.
We want to have it accessible from the C++ Catalyst libraries and tools, for example from `quantum-opt`.

There are still a few things that C++ cannot do at compile time, like reading and parsing a Python file.
On top of that, there are the issues of:
- How to access a file in a different project (e.g., `frontend` project from `mlir` project).
- How to locate a given source file during runtime. For example, you could run `quantum-opt` from anywhere, and `_version.py` may not be even accessible.
But we can do some things at CMake configure time, as suggested in the description of the issue:
  >  an alternative such as pulling the version from the Git metadata or generating it at build time would be great.

This PR:
1. Updates `mlir/CMakeLists.txt` to generate `include/Catalyst/Utils/frontend_catalyst_version_py.h` from `frontend/catalyst/_version.py`. This file just contains a `static constexpr const char *CATALYST_VERSION` with the value parsed from the Python file.
2. Adds a `catalyst::printVersion(llvm::raw_ostream&)` function to MLIRCatalystUtils library. This functions just prints `CATALYST_VERSION` to an LLVM raw ostream.
3. Uses that function from `tools/quantum-opt/quantum-opt.cpp` to display the Catalyst version when invoked via `quantum-opt --version`.
4. Adds a new unit test binary, `CatalystsUtilsTests`, with only one test that checks that the catalyst version is not `unknown` (the default value when the version couldn't be parsed from `_version.py`).

**Benefits:**
1. We still have a single source of truth in `frontend/catalyst/_version.py`.
2. The version is set in the C++ code during CMake configure time (so accessible during runtime).

**Possible Drawbacks:**
- Current implementation prints the LLVM version before the Catalyst version, because the Catalyst version is an extra .

**Related GitHub Issues:**
#1586

### DIFF
--- a/mlir/CMakeLists.txt
+++ b/mlir/CMakeLists.txt
@@ -121,6 +121,26 @@ FetchContent_Declare(tomlplusplus
 )
 FetchContent_MakeAvailable(tomlplusplus)
 
+################################################################################
+## Generate frontend_catalyst_version_py.h from frontend/catalyst/_version.py ##
+################################################################################
+
+function(make_version_hpp_file input_file output_file)
+    file(READ ${input_file} input_content)
+    if (input_content MATCHES "__version__ = \"([^\"]+)\"")
+        set(version_number "${CMAKE_MATCH_1}")
+    else()
+        message(WARNING "Could not find Catalyst version in ${input_file}")
+        set(version_number "unknown")
+    endif()
+    set(header_content "static constexpr const char *CATALYST_VERSION = \"${version_number}\";")
+    file(WRITE ${output_file} "${header_content}")
+endfunction(make_version_hpp_file)
+
+make_version_hpp_file(
+    "${PROJECT_SOURCE_DIR}/../frontend/catalyst/_version.py"
+    "${PROJECT_BINARY_DIR}/include/Catalyst/Utils/frontend_catalyst_version_py.h"
+)
 
 ######################
 ## Catalyst Sources ##
--- a/mlir/unittests/Utils/PrintVersion.cpp
+++ b/mlir/unittests/Utils/PrintVersion.cpp
@@ -1,0 +1,27 @@
+// Copyright 2024 Xanadu Quantum Technologies Inc.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+
+//     http://www.apache.org/licenses/LICENSE-2.0
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "Catalyst/Utils/PrintVersion.h"
+
+#include "gmock/gmock.h" // HasSubstr
+
+namespace {
+TEST(MLIRCatalystUtils, VersionIsNotUnknown)
+{
+    auto output = std::string{};
+    auto rss = llvm::raw_string_ostream{output};
+    catalyst::printVersion(rss);
+    EXPECT_THAT(output, ::testing::Not(::testing::HasSubstr("unknown")));
+}
+} // namespace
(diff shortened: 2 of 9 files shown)


mlir/CMakeLists.txt
```diff
@@ -121,6 +121,30 @@ FetchContent_Declare(tomlplusplus
...
+    else()
+        file(READ ${input_file} input_content)
+        if (NOT input_content MATCHES "__version__ = \"([^\"]+)\"")
+            message(WARNING "Could not find Catalyst version in ${input_file}")
+        else()
+            set(version_number "${CMAKE_MATCH_1}")
+        endif()
+    endif()
+    set(header_content "static constexpr const char *CATALYST_VERSION = \"${version_number}\";")
+    file(WRITE ${output_file} "${header_content}")
```
dime10: I think this solves a similar problem to the build time macro (`LIBREGISTRY`) we have configured here:
https://github.com/PennyLaneAI/catalyst/blob/f314679ccc577c425b45e8a3804beac93ba317c9/runtime/lib/capi/RuntimeCAPI.cpp#L142
https://github.com/PennyLaneAI/catalyst/blob/f314679ccc577c425b45e8a3804beac93ba317c9/runtime/lib/registry/CMakeLists.txt#L31

I don't know how the two (the macro and your file-based approach) compare though, so I'm happy to go with what you have 👍

dime10 (approved):
Great work, and thank you for picking up some of those issues from the backlog! 🚀🚀 

Happy to approve :)

mlir/unittests/Utils/PrintVersion.cpp
```diff
@@ -0,0 +1,27 @@
+// Copyright 2024 Xanadu Quantum Technologies Inc.
```
sengthai:
```suggestion
// Copyright 2025 Xanadu Quantum Technologies Inc.
```
Small thing. Maybe switch the year to 2025? 😄

rturrado: Good spot!

mlir/CMakeLists.txt
rturrado: Thanks! I'll have a look and see if I can find any similarities.

doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,9 @@
 
 <h3>Improvements 🛠</h3>
 
+*  Displays Catalyst version in `quantum-opt --version` output.
+  [#1922](https://github.com/PennyLaneAI/catalyst/pull/1922)
```
dime10: 👍 
```suggestion
*  Displays Catalyst version in `quantum-opt --version` output.
  [(#1922)](https://github.com/PennyLaneAI/catalyst/pull/1922)
```

mlir/CMakeLists.txt
rturrado: I've had a quick look at the runtime C API, but not enough to properly understand it. My feeling, anyway, is that the C API kind of solution may be quite more convoluted for this version issue.

rturrado:
Many thanks everyone for your comments! Please don't hesitate to be as strict as you consider with the review. In the end it is your codebase, and I'm just trying to learn about it and, if possible, help.

mlir/CMakeLists.txt
dime10: No worries, I was mainly referring to our use of [preprocessor definitions](https://cmake.org/cmake/help/latest/command/add_compile_definitions.html) there, nothing specific about the runtime or C API itself. If you look at the `RuntimeCAPI.cpp` file, the `LIBREGISTRY` name on line 142 isn't actually defined anywhere. Instead, we instruct CMake to provide a replacement value for it to the preprocessor at build time. 

In my opinion, this achieves the same thing you are doing here (since you determine the version number in the cmake script at build time), just without writing to a separate header file. Either way I'm happy with the solution so I'll mark this thread as resolved :)

rturrado: Ah OK, now I see. That would indeed be another way of achieving the same thing. And it may be even simpler and faster as it doesn't involve a file creation.

rturrado:
> Great feel free to merge when you're ready (or let me know in case you don't have access to the merge button) :)

I think I do not have merge privileges, so please go ahead!

PR #1924, opened 2025-07-21
Fix cache and patch issues with CI builds

dime10:


dime10:
Looking good: 
- https://github.com/PennyLaneAI/catalyst/actions/runs/16424925293
- https://github.com/PennyLaneAI/catalyst/actions/runs/16424937855
- https://github.com/PennyLaneAI/catalyst/actions/runs/16424941628

PR #1927, opened 2025-07-21
[MLIR] (documentation) Add notes to the quantum dialect

erick-xanadu:
**Context:** Adding a little bit more documentation to the Quantum dialect

**Description of the Change:**:
* Adding notes where some more discussion could be useful
* Fixes `quantum.state`'s example to take an observable as an operand.

**Benefits:** Small improvements in documentation.

**Possible Drawbacks:** None.

**Related GitHub Issues:**


mlir/include/Quantum/IR/QuantumOps.td
```diff
@@ -369,6 +381,11 @@ def SetStateOp : Gate_Op<"set_state"> {
         This operation is useful for simulators implementing state preparation.
         Instead of decomposing state preparation into multiple operations, this
         operation shortcuts all of that into a single operation.
+
+        .. note::
+
+            This operation should be the first quantum operation of the entry point of
+            a quantum module.
```
dime10: I put this blob in the QuantumDevice interface:
```
     * This instruction initializes a set of qubits to the provided computational basis state.
     * Although the instruction is typically used for state initialization at the beginning of
     * a quantum program, devices are encouraged to support arbitrary re-initialization if
     * capable. Reinitialization on a subset of qubits is equivalent to deallocation -> allocation
     * -> initialization on the same set of qubits.
```

I don't think we should restrict it to the first instruction in a module, at least not at the IR level (this is fine as frontend or device-specific restriction though like we currently have it).

One reason is that this op can be quite useful even at other points in a program, for instance for tree traversal.

mlir/include/Quantum/IR/QuantumOps.td
```diff
@@ -552,7 +574,14 @@ def CustomOp : UnitaryGate_Op<"custom", [DifferentiableGate, NoMemoryEffect,
...
 def GlobalPhaseOp : UnitaryGate_Op<"gphase", [DifferentiableGate, AttrSizedOperandSegments]> {
     let summary = "Global Phase.";
+
     let description = [{
+    	Applies global phase to the current system.
+
+    	.. note::
+
+            With the introduction of dynamic qubit allocation, applying a global
+            phase and then creating a new qubit results in undefined behaviour.
```
dime10: I don't see why this would be undefined behaviour?

dime10:
Thanks Erick!

mlir/include/Quantum/IR/QuantumOps.td
erick-xanadu: My understanding from the discussions we had offline was that it is up to the device how to implement a global phase in the presence of dynamic qubit allocation. But I may be wrong.

Case 1:
Even though the original allocation of qubits is N, a device is always allowed to allocate M > N qubits at the start of the program. A global phase operation would then introduce a global phase for all the M qubits. 

Case 2:
A device is given the allocate operation for N qubits and it allocates N qubits. Global phase is applied on the current system. A user requests a new qubit. This new qubit may be out of phase.

Happy to erase this note after you read this comment. Thanks!

erick-xanadu: We could also call it implementation defined behaviour.

mlir/include/Quantum/IR/QuantumOps.td
erick-xanadu:
```suggestion
```

mlir/include/Quantum/IR/QuantumOps.td
```diff
@@ -402,6 +419,11 @@ def SetBasisStateOp : Gate_Op<"set_basis_state"> {
         operation shortcuts all of that into a single operation.
         This signature matches the one in pennylane-lightning which expects
         only a single integer as opposed to a binary digit.
+
+        .. note::
+
+            This operation should be the first quantum operation of the entry point of
+            a quantum module.
```
erick-xanadu:
```suggestion
```

mlir/include/Quantum/IR/QuantumOps.td
dime10: Hmm, I think the ideas above stem from the misconception that global phase applies to _qubits_ -- it does not.

Let's take the statevector formalism. A global phase $g$ is a complex number of size 1 that is common to all elements of the statevector, that is it can be factored in or out arbitrarily:
```math
g \begin{pmatrix} a \\ b \end{pmatrix} =  \begin{pmatrix} g a \\ g b \end{pmatrix}
```

If we allocate a new qubit, we want to combine the existing statevector with a description of the new qubit, which is done using the tensor product. Typically, the new qubit is in the zero state $$( 1,  0 )$$, but this isn't actually necessary.

Let's compute the new statevector:
```math
g \begin{pmatrix} a \\ b \end{pmatrix} \otimes \begin{pmatrix} c \\ d \end{pmatrix} =  \begin{pmatrix} g a \\ g b \end{pmatrix} \otimes \begin{pmatrix} c \\ d \end{pmatrix} = \begin{pmatrix} g a c \\ g b c \\ g a d \\ g b d \end{pmatrix} = g \begin{pmatrix} a c \\ b c \\ a d \\ b d \end{pmatrix}
```

From this it should be clear that it doesn't matter whether the global phase operation is applied before or after allocating a new qubit, the result is the same. After all, the global phase is really just that, a _global_ factor of the quantum state.

dime10: Note that there is a distinction between the "global phase" as an element of the state ($g$ above), and the "global phase operation". The latter is an operator taking an arbitrary state to the scaled state:

```math
\begin{pmatrix} \bullet \end{pmatrix} \mapsto g \begin{pmatrix} \bullet \end{pmatrix}
```
.

This is the reason I don't particularly like the representation of the global phase operator as one that acts on particular qubits (for example a 2x2 unitary matrix for the 1 qubit version), as that qubit is purely arbitrary and either way you are scaling the whole system.

erick-xanadu:
```suggestion
```

erick-xanadu: Thank you, I have removed the comment.

mlir/include/Quantum/IR/QuantumOps.td
```diff
@@ -552,7 +564,10 @@ def CustomOp : UnitaryGate_Op<"custom", [DifferentiableGate, NoMemoryEffect,
 
 def GlobalPhaseOp : UnitaryGate_Op<"gphase", [DifferentiableGate, AttrSizedOperandSegments]> {
     let summary = "Global Phase.";
+
     let description = [{
+    	Applies global phase to the current system.
+
```
dime10:
```suggestion
```

mlir/include/Quantum/IR/QuantumOps.td
```diff
@@ -1115,6 +1146,11 @@ def ProbsOp : Measurement_Op<"probs", [AttrSizedOperandSegments]> {
             func.return
         }
         ```
+
```
dime10, line 1148:
```suggestion
```

mlir/include/Quantum/IR/QuantumOps.td
```diff
@@ -1115,6 +1146,11 @@ def ProbsOp : Measurement_Op<"probs", [AttrSizedOperandSegments]> {
...
+
+        .. note::
+
+            The optional operand state_in is only used after bufferization.
```
dime10:
```suggestion
            The optional operand ``state_in`` is only used after bufferization.
```

PR #1928, opened 2025-07-22
Naive Qubit Mapping and Routing Algorithm in Runtime

ritu-thombre99:
**Context:**

When running a quantum circuit on a hardware with certain connectivity constraints, two-qubit gates like CNOTs can only be executed using physical qubits that are connected by an edge on the hardware.

Hence, necessary SWAPs need to be inserted in order to route the logical qubits so that they are mapped to an edge on the device, while respecting other compiling constraint i.e. order the gate dependencies from the input quantum circuit, and ensuring compiled quantum circuit is equivalent to the input quantum circuit.

Following image shows a simple example:

image

**Benefits:**

1. This will support qubit mapping and routing in the runtime 
2. Analogue to [qml.transpile](https://docs.pennylane.ai/en/stable/code/api/pennylane.transforms.transpile.html) transform
3. Circuits with control structure will also be supported since they will already be optimized by Catalyst passes and unrolled in the runtime.

**Possible Drawbacks:**
1. Modification of RuntimeCAPI 
2. Currently, not checking if device supports SWAP
3. TODOL Consider automatic qubit management

**Related GitHub Issues:**

### DIFF
--- a/runtime/lib/capi/RuntimeCAPI.cpp
+++ b/runtime/lib/capi/RuntimeCAPI.cpp
@@ -18,6 +18,9 @@
 #include <ctime>
 #include <memory>
 #include <ostream>
+#include <limits>
+#include <set>
+#include <map>
 #include <stdexcept>
 #include <string_view>
 #include <tuple>
@@ -46,6 +49,11 @@ static std::unique_ptr<ExecutionContext> CTX = nullptr;
  * @brief Thread local device pointer with internal linkage.
  */
 thread_local static RTDevice *RTD_PTR = nullptr;
+
+const int MAXIMUM = 1e9;
+std::set<int> physical_qubits;
+std::map<std::pair<int,int>, bool> coupling_map;
+std::map<std::pair<int,int>, int> distance_matrix;
 
 bool getModifiersAdjoint(const Modifiers *modifiers)
 {
@@ -256,7 +264,7 @@ void __catalyst__rt__finalize()
 
 static int __catalyst__rt__device_init__impl(int8_t *rtd_lib, int8_t *rtd_name, int8_t *rtd_kwargs,
                                              int64_t shots, bool auto_qubit_management)
-{
+{   
     // Device library cannot be a nullptr
     RT_FAIL_IF(!rtd_lib, "Invalid device library");
     RT_FAIL_IF(!CTX, "Invalid use of the global driver before initialization");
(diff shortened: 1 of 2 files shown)


runtime/lib/capi/RuntimeCAPI.cpp
```diff
@@ -612,10 +779,13 @@ void __catalyst__qis__CNOT(QUBIT *control, QUBIT *target, const Modifiers *modif
 {
     RT_FAIL_IF(control == target,
                "Invalid input for CNOT gate. Control and target qubit operands must be distinct.");
+    
+    std::pair<int,int> routedQubits = getRoutedQubits(control, target, modifiers);
```
dime10: We should have a flag here that controls whether routing is performed or not.

runtime/lib/capi/RuntimeCAPI.cpp
```diff
@@ -272,6 +345,100 @@ static int __catalyst__rt__device_init__impl(int8_t *rtd_lib, int8_t *rtd_name,
     if (CTX->getDeviceRecorderStatus()) {
         getQuantumDevicePtr()->StartTapeRecording();
     }
+
+    // Extract coupling map from the kwargs passed
```
dime10: This should be split out into its own function.

In fact, we want to take all these functions and put them in a new file (Routing.cpp/hpp) that you can then call into.

runtime/lib/backend/null_qubit/NullQubit.hpp
```diff
@@ -242,6 +242,11 @@ struct NullQubit final : public Catalyst::Runtime::QuantumDevice {
                         const std::vector<QubitIdType> &controlled_wires = {},
                         const std::vector<bool> &controlled_values = {})
     {
+        // Print to see what naive router does on Null qubits
```
dime10, line 251: We should add some runtime tests for the routing functionality

runtime/lib/capi/RuntimeCAPI.cpp
```diff
@@ -612,10 +779,13 @@ void __catalyst__qis__CNOT(QUBIT *control, QUBIT *target, const Modifiers *modif
 {
     RT_FAIL_IF(control == target,
                "Invalid input for CNOT gate. Control and target qubit operands must be distinct.");
+    
+    std::pair<int,int> routedQubits = getRoutedQubits(control, target, modifiers);
     getQuantumDevicePtr()->NamedOperation("CNOT", {},
-                                          {/* control = */ reinterpret_cast<QubitIdType>(control),
-                                           /* target = */ reinterpret_cast<QubitIdType>(target)},
+                                          {/* control = */ routedQubits.first,
+                                           /* target = */ routedQubits.second},
```
dime10: I think we should keep the type that is passed to the function as a `QubitIdType` (even if you don't use it in your algorithm internally).

runtime/lib/capi/RuntimeCAPI.cpp
```diff
@@ -272,6 +345,100 @@ static int __catalyst__rt__device_init__impl(int8_t *rtd_lib, int8_t *rtd_name,
...
+    for (auto i_itr = physicalQubits.begin(); i_itr != physicalQubits.end(); i_itr++)
+    {
+        for (auto j_itr = physicalQubits.begin(); j_itr != physicalQubits.end(); j_itr++ ) 
+        {
+            for (auto k_itr = physicalQubits.begin(); k_itr != physicalQubits.end(); k_itr++ ) 
+            {
+                if (distanceMatrix[std::make_pair(*j_itr,*i_itr)] + distanceMatrix[std::make_pair(*i_itr,*k_itr)] < distanceMatrix[std::make_pair(*j_itr,*k_itr)] )
+                {
+                    distanceMatrix[std::make_pair(*j_itr,*k_itr)] = distanceMatrix[std::make_pair(*j_itr,*i_itr)] + distanceMatrix[std::make_pair(*i_itr,*k_itr)];
+                    predecessorMatrix[std::make_pair(*j_itr,*k_itr)] = predecessorMatrix[std::make_pair(*i_itr,*k_itr)];
```
dime10: The next question is where we want to store this kind of information. If we don't want to touch device backends, but still want it to specific to each device (unlike the global you have right now), I think the natural place for it would be in the `RTDevice` class.

runtime/lib/capi/RuntimeCAPI.cpp
```diff
@@ -272,6 +345,100 @@ static int __catalyst__rt__device_init__impl(int8_t *rtd_lib, int8_t *rtd_name,
     if (CTX->getDeviceRecorderStatus()) {
         getQuantumDevicePtr()->StartTapeRecording();
     }
+
+    // Extract coupling map from the kwargs passed 
+    // If coupling map is provided then it takes in the form {...,'couplingMap' ((a,b),(b,c))}
+    // else {...,'couplingMap' (a,b,c)}
+    size_t start = args[2].find("coupling_map': ") + 15; // Find key and opening parenthesis
```
dime10, line 275: Finally we'll want to device how to pass down the coupling map from the frontend to the runtime.

I think the shortcut via the device_kwargs is okay for now, although normally those are meant to go the Device class constructor. We may want to associate them closer to the runtime then than the device, like we do with shots or qubit management.

runtime/lib/capi/RuntimeCAPI.cpp
ritu-thombre99: Changed types in https://github.com/PennyLaneAI/catalyst/commit/4227d89fe4f3dbf432f8b253531f0137bab1400b

runtime/lib/capi/RuntimeCAPI.cpp
ritu-thombre99: Modularized in Routing.hpp https://github.com/PennyLaneAI/catalyst/commit/4ae4a4670874ae7e570ae8f2c6a55e547db31caa

runtime/lib/capi/RuntimeCAPI.cpp
ritu-thombre99: Added option to execution context https://github.com/PennyLaneAI/catalyst/commit/4ae4a4670874ae7e570ae8f2c6a55e547db31caa#diff-4718a857bb4ed45b3ed299bdb7926e5beca237e53aa645d124f295e0cbbd2357R312

Checks happening before every two-qubit gate https://github.com/PennyLaneAI/catalyst/commit/4ae4a4670874ae7e570ae8f2c6a55e547db31caa#diff-1a3e5dcf898ac446b6355c7f97c3c62d9039d75db6bb79268d07e3e6d7072bd1R655

runtime/lib/capi/RuntimeCAPI.cpp
ritu-thombre99: Defined them as class variables in Routing.hpp https://github.com/PennyLaneAI/catalyst/commit/4ae4a4670874ae7e570ae8f2c6a55e547db31caa which are initialized only if wires passed to PennyLane device are in a format of list of tuples representing coupling map of the hardware

runtime/lib/capi/Routing.hpp
```diff
@@ -0,0 +1,180 @@
+// Copyright 2025 Xanadu Quantum Technologies Inc.
```
mlxd, line 1: Is this file meant to be the header, or the implementation?
If the header, can we add an `#pragma once` at the top? If the implementation, we should probably split the header interface and implementation details separately.

runtime/lib/capi/Routing.hpp
```diff
@@ -0,0 +1,180 @@
...
+
+#include "Exception.hpp"
+#include "ExecutionContext.hpp"
+#include "MemRefUtils.hpp"
+#include "QuantumDevice.hpp"
+#include "Timer.hpp"
+
+#include "RuntimeCAPI.h"
+
+const int MAXIMUM = 1e9;
```
mlxd: Rather than having scope visibility across this header, can this be isolated? Otherwise, this will be exported and visible in the generated translation unit.

Either programmatically passing this as an argument to a function/method where needed with a given default value, or (if it is required we set this magic variable), it can be isolated in an anonymous namespace, which should ensure it isn't visible in the compiled code outside of where it is needed.

runtime/lib/capi/RuntimeCAPI.cpp
```diff
@@ -47,6 +48,11 @@ static std::unique_ptr<ExecutionContext> CTX = nullptr;
  */
 thread_local static RTDevice *RTD_PTR = nullptr;
 
+/**
+ * @brief Global routing pass pointer.
+ */
+static std::unique_ptr<RoutingPass> RUNTIME_ROUTER = nullptr;
```
mlxd: Do we need this to be global? Is there a way we can isolate this so it isn't exposed? Any concerns with thread locality?

mlxd (commented):
Just a few fly-by comments from me.
Feel free to ignore them as needed

runtime/lib/capi/RuntimeCAPI.cpp
dime10: I agree, the router should be able to be scoped to each device instance (RTDevice)

PR #1941, opened 2025-07-25
Add tools for modifying compilation pipelines

joeycarter:
**Context:** As more compilation passes are added to Catalyst, it's becoming increasingly necessary to define custom compilation pass pipelines for specialized workloads.

**Description of the Change:** This PR adds internal developer tools to

* Retrieve the default pipeline (i.e. the default sequence of compilation stages) with `catalyst.pipelines.default_pipeline()`
* Insert passes into an existing pipeline with `catalyst.pipelines.insert_pass_after()` and `catalyst.pipelines.insert_pass_before()`
* Insert named compilation stages into an existing sequence of stages with `catalyst.pipelines.insert_stage_after()` and `catalyst.pipelines.insert_stage_before()`

**Examples:**

Insert a pass into a pipeline:

```pycon
>>> pipeline = ["pass1", "pass2"]
>>> insert_pass_after(pipeline, "new_pass", ref_pass="pass1")
>>> pipeline
['pass1', 'new_pass', 'pass2']
```

Insert a stage into a sequence of stages:

```pycon
>>> stages = [("stage1", ["s1p1", "s1p2"]), ("stage2", ["s2p1", "s2p2"])]
>>> insert_stage_after(stages, ("new_stage", ["p0"]), ref_stage="stage1")
>>> stages
[('stage1', ['s1p1', 's1p2']), ('new_stage', ['p0']), ('stage2', ['s2p1', 's2p2'])]
```

**Benefits:** Easier definition of custom compilation pass pipelines, especially where only one or a few passes need to be inserted at specific positions in the default pipeline.

**Notes:** This PR supersedes #1758.

[sc-96323]

PR #1942, opened 2025-07-25
Add catalyst.ftqc.mbqc_pipeline for experimental MBQC workloads

joeycarter:
**Context:** When we initially added the MBQC dialect, we had to manually specify the full compilation pipeline add drop in the MBQC-to-LLVM dialect-conversion pass, `"convert-mbqc-to-llvm"`, to convert the ops in the `mbqc` dialect to the LLVM dialect. For example:

```python
mbqc_pipeline = [
    (
        "default-pipeline",
        [
            "enforce-runtime-invariants-pipeline",
            "hlo-lowering-pipeline",
            "quantum-compilation-pipeline",
            "bufferization-pipeline",
        ],
    ),
    (
        "mbqc-pipeline",
        [
            "convert-mbqc-to-llvm",
        ],
    ),
    (
        "llvm-dialect-lowering-pipeline",
        [
            "llvm-dialect-lowering-pipeline",
        ],
    ),
]

@qml.qnode(dev)
def workload():
    ...
```

No one is going to want to type that out each time, so it would be handy to have this pipeline available in a common location.

**Description of the Change:** Adds a new function, `mbqc_pipeline()` to a new `catalyst.ftqc` module. This function first makes a copy of the default pipeline stages (from `catalyst.pipelines.default_pipeline()`, recently added in #1941) and inserts the mbqc-to-llvm conversion pass immediately before the quantum-to-llvm conversion pass in the `"MLIRToLLVMDialect"` stage.

**Example:**

```python
import numpy as np
import pennylane as qml
from catalyst.ftqc import mbqc_pipeline

qml.capture.enable()

@qml.qjit(pipelines=mbqc_pipeline())
@qml.qnode(dev, mcm_method="tree-traversal")
def workload():
    m0 = qml.ftqc.measure_arbitrary_basis(wires=0, angle=np.pi / 4, plane="XY")
    qml.cond(m0, qml.X, qml.I)(0)
    return qml.expval(qml.Z(0))

workload()
```

**Benefits:** Less typing to specify the pass pipeline for MBQC workloads (and therefore less error-prone!)

**Possible Drawbacks:** There is planned development work to more broadly support user-definition of custom compilation pipelines in Catalyst. Once this work is complete, it may be beneficial to revert this change and use the new infrastructure.

**Note:** This PR was adapted from and supersedes #1758.


mlxd (approved):
No notes --- looks exactly as discussed.
Thanks @joeycarter

PR #1944, opened 2025-07-29
Move mhlo passes from patch to `Catalyst` dialect

paul0403:
**Context:**
Move mhlo passes from patch to `Catalyst` dialect

**Benefits:**
- Easier to maintain: when llvm updates we would be editing regular cpp files instead of patch files.
- Better organization: fewer patching going on. 
- Much easier review for the stablehlo migration PR.


dime10:
Maybe these should go in a dialect folder called HLO?

mlir/include/Catalyst/Transforms/Passes.td
```diff
@@ -282,4 +282,53 @@ def BufferDeallocation : Pass<"buffer-deallocation", "func::FuncOp"> {
   let constructor = "mlir::bufferization::createBufferDeallocationPass()";
 }
 
+// mhlo legalize sort pass.
+// mhlo dropped the -legalize-sort pass when migrating to stablehlo.
```
dime10: Let's differentiate the project (mlir-hlo) from _one of_ the dialects it contains (mhlo) across these comments and notices.

paul0403:
> Maybe these should go in a dialect folder called HLO?

Done, I split them into `mlir-hlo` folders.

mlir/include/Catalyst/mlir-hlo/mhlo_legalize_to_standard_patterns.td
```diff
@@ -0,0 +1,122 @@
...
+// limitations under the License.
+
+
+// This file is taken from the
+//   tensorflow/mlir-hlo
+// repository, under the Apache 2.0 License, at
+//   https://github.com/tensorflow/mlir-hlo/blob/a5529d99fc4d1132b0c282a053d26c11e6636b3a/mhlo/transforms/legalize_to_standard/legalize_to_standard_patterns.td
+// with the following copyright notice:
+
+    /* Copyright 2019 The OpenXLA Authors.
```
mehrdad2m: I am not sure how the licensing would work when copying external code. Is there a guideline for this out there?

paul0403: See what we did for the upstream buffer dealloc pass https://github.com/PennyLaneAI/catalyst/blob/main/mlir/lib/Catalyst/Transforms/BufferDeallocation.cpp

We also pulled in lapack code from jax like this as well.

mehrdad2m (approved):

mlir/include/Catalyst/mlir-hlo/mhlo_legalize_to_standard_patterns.td
dime10: Yeah this is typically how we do it. It has to include that it's foreign code, the license used, and the copyright if available.

frontend/catalyst/pipelines.py
```diff
@@ -231,7 +231,7 @@ def get_hlo_lowering_stage(_options: CompileOptions) -> List[str]:
         "func.func(mhlo-legalize-control-flow)",
         "func.func(hlo-legalize-to-linalg)",
         "func.func(mhlo-legalize-to-std)",
-        "func.func(hlo-legalize-sort)",
```
dime10, line 234: Is there a reason to rename the original pass name?

mlir/include/Catalyst/Transforms/Passes.h
```diff
@@ -18,6 +18,8 @@
 
 #include "mlir/Pass/Pass.h"
 
+#include "Catalyst/mlir-hlo/Passes.h"
```
dime10: Is this header useful if we don't add any new lines here?

mlir/include/Catalyst/mlir-hlo/CMakeLists.txt
dime10, line 1: Sorry I didn't mean a subfolder in the catalyst dialect, but a dialect folder of it's own. Does that pose any problems?

frontend/catalyst/pipelines.py
paul0403: I wanted all the "patched" passes to start with `mhlo-`.

mlir/include/Catalyst/mlir-hlo/CMakeLists.txt
paul0403: https://github.com/PennyLaneAI/catalyst/pull/1944/commits/22aa47672c3607df3395a0eee8902a2c0254e21e

mlir/include/Catalyst/Transforms/Passes.h
paul0403: You are right, it is not needed. I removed them in https://github.com/PennyLaneAI/catalyst/pull/1944/commits/de202031362e250f8ef0112a3ca304aa486163ac

mlir/include/Catalyst/Transforms/Passes.td
paul0403: Done. All the attribution notices now use the full name of the repo `tensorflow/mlir-hlo`

dime10 (approved):
Neat!

PR #1945, opened 2025-07-29
Error out workflow with capture enabled

JerryChen97:
**Context:**
`for_loop`, `while_loop` and `cond` won't work with capture enabled. So we error out when these happen and provide information to users.

 - New `CompatibilityError`
 - At the entry point of these three workflows, validate the `qml.capture.enabled`

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-95361]


doc/releases/changelog-dev.md
```diff
@@ -4,6 +4,9 @@
 
 <h3>Improvements 🛠</h3>
 
+* Workflows `for_loop`, `while_loop` and `cond` now error out if `qml.capture` is enabled.
+  [(#1945)](https://github.com/PennyLaneAI/catalyst/pull/1945)
```
dime10, line 8: I wonder if this should go into a deprecation section or something. But I guess these have never worked with qml.capture to begin with so this is probably fine as an improvement for now.

frontend/catalyst/api_extensions/control_flow.py
```diff
@@ -239,7 +240,32 @@ def conditional_fn():
         ...     return cond_fn()
         >>> f(1.6)
         Array(2.56, dtype=float64)
+
+    .. note::
+
+        Currently, ``catalyst.cond`` is not supported in program capture mode.
+        If you are using ``qml.qnode`` with PennyLane's ``capture`` feature, you
+        should use ``qml.cond`` instead:
```
dime10: Since we have no plans to support it we can remove "currently".

Another thing is the wording around "qnode". Catalyst control flow can be used inside or outside of qnodes, if this is the case with qml control flow as well we should remove the mention of qnode.

```suggestion
       ``catalyst.cond`` is not supported in program capture mode. If ``qml.capture`` is enabled,
        please use ``qml.cond`` instead.
```

It would also make sense to update the given example because the two are *not* equivalent (qjit vs qnode).

frontend/catalyst/api_extensions/control_flow.py
```diff
@@ -239,7 +240,32 @@ def conditional_fn():
...
+            # This will raise an error with capture mode
+            @qjit
+            def func(x):
+                @catalyst.cond(x > 1.0)
+                def cond_fn():
+                    return x ** 2
+                return cond_fn()
+
+            # Use this instead for capture mode compatibility
+            @qml.qnode(device, interface="jax")
```
dime10: The `interface` keyword is not something that was ever by catalyst, is this required with program capture? Imo using program capture implies jax, so this seems a bit redundant.

frontend/catalyst/api_extensions/control_flow.py
dime10: Similarly for the other functions

frontend/test/pytest/test_conditionals.py
```diff
@@ -451,6 +452,51 @@ def h(x: int):
...
+                def cond_fn():
+                    return 1
+
+            # Verify the error message is specific and helpful
+            error_msg = str(exc_info.value)
+            assert "catalyst.cond is not supported with PennyLane's capture enabled" in error_msg
+
+        finally:
+            # Always disable capture mode after test
+            qml.capture.disable()
```
dime10: Good thinking! We do have a fixture for this though to avoid writing this code in each test :) 
https://github.com/PennyLaneAI/catalyst/blob/d20dfc03559cbe02de11a0fe49cc6f1f2a0c538a/frontend/test/pytest/conftest.py#L43

You can add the following to your test function or class:
```py
@pytest.mark.usefixture("disable_capture")
```

paul0403:
Just for my sanity, the supposed use case is 
```
qml.capture_enable()

...
@qjit
def ...
   qml.for_loop
```

correct? In other words `catalyst.for_loop` is only supposed to go through the non-plxpr pipeline?

JerryChen97:
> Just for my sanity, the supposed use case is
> 
> ```
> qml.capture_enable()
> 
> ...
> @qjit
> def ...
>    qml.for_loop
> ```
> 
> correct? In other words `catalyst.for_loop` is only supposed to go through the non-plxpr pipeline?

That's howI understood the story as well!

doc/releases/changelog-dev.md
JerryChen97: Yeah, it's more like an extra informative UX designed ahead of wildfire

frontend/test/pytest/test_conditionals.py
JerryChen97: https://github.com/PennyLaneAI/catalyst/pull/1945/commits/59e8df36844afb2f473fbab4f0ba6ec745220c8f

frontend/catalyst/api_extensions/control_flow.py
JerryChen97: https://github.com/PennyLaneAI/catalyst/pull/1945/commits/5017f387b96a12648d62cb4190fb6e3701a57d1e

frontend/catalyst/api_extensions/control_flow.py
```diff
@@ -239,7 +240,31 @@ def conditional_fn():
...
+            # This will raise an error with capture mode
+            @qjit
+            def func(x):
+                @catalyst.cond(x > 1.0)
+                def cond_fn():
+                    return x ** 2
+                return cond_fn()
+
+            # Use this instead for capture mode compatibility
+            @qml.qnode(device)
```
dime10: Why not `qjit`? Otherwise the programs are two entirely different things.
```suggestion
            @qjit
```

dime10: Additionally in pennylane, qnodes have to return measurement processes. Not sure if this applies to program capture or not though.

frontend/catalyst/api_extensions/control_flow.py
```diff
@@ -388,7 +413,31 @@ def loop_rx(i, x):
...
+
+    .. note::
+
+       ``catalyst.for_loop`` is not supported in program capture mode.
+       If ``qml.capture`` is enabled, please use ``qml.for_loop`` instead.
+
+        .. code-block:: python
+
+            # This will raise an error with capture mode
+            @qjit
```
dime10, line 426: Minor: We could be even clearer here and show how program capture might be enabled:
```suggestion
            qml.capture.enable()
            # This will raise an error with capture mode
            @qjit
```

frontend/test/pytest/test_loops.py
```diff
@@ -253,6 +254,47 @@ def inner(j):
...
+
+            @while_loop(condition)
+            def loop_fn(x):
+                return x + 1
+
+        # Verify the error message is specific and helpful
+        error_msg = str(exc_info.value)
+        assert "catalyst.while_loop is not supported with PennyLane's capture enabled" in error_msg
+
+    @pytest.mark.xfail(reason="autograph")
```
dime10: What's going on here?

JerryChen97: So it will emit the error we just deisgned and be successfully caught. However, one more error from autograph will show up... I think it might be bad idea to introduce more dependency just for this test so xfailed it here

dime10: Is autograph always on with program capture? (I don't see it enabled in the test)

I wonder what is different about this test than the others. Can you share the error exactly?

I'd would like to avoid just having an xfail test for no reason. What is the dependency you'd like to avoid?

JerryChen97:
```bash
>           test(4)

frontend/test/pytest/test_loops.py:292: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../pennylane/venv/lib/python3.12/site-packages/pennylane/logging/decorators.py:61: in wrapper_entry
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
frontend/catalyst/jit.py:596: in __call__
    requires_promotion = self.jit_compile(args, **kwargs)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../pennylane/venv/lib/python3.12/site-packages/pennylane/logging/decorators.py:61: in wrapper_entry
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
frontend/catalyst/jit.py:669: in jit_compile
    self.jaxpr, self.out_type, self.out_treedef, self.c_sig = self.capture(args, **kwargs)
                                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
frontend/catalyst/debug/instruments.py:145: in wrapper
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
../pennylane/venv/lib/python3.12/site-packages/pennylane/logging/decorators.py:61: in wrapper_entry
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
frontend/catalyst/jit.py:730: in capture
    return trace_from_pennylane(
frontend/catalyst/from_plxpr/from_plxpr.py:718: in trace_from_pennylane
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
frontend/catalyst/jax_extras/tracing.py:499: in make_jaxpr_f
    jaxpr, out_type, consts = trace_to_jaxpr_dynamic2(f)
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^
../pennylane/venv/lib/python3.12/site-packages/jax/_src/profiler.py:354: in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
../pennylane/venv/lib/python3.12/site-packages/jax/_src/interpreters/partial_eval.py:2363: in trace_to_jaxpr_dynamic2
    ans = fun.call_wrapped(*in_tracers)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../pennylane/venv/lib/python3.12/site-packages/jax/_src/linear_util.py:211: in call_wrapped
    return self.f_transformed(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../pennylane/venv/lib/python3.12/site-packages/jax/_src/api_util.py:73: in flatten_fun
    ans = f(*py_args, **py_kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^
../pennylane/venv/lib/python3.12/site-packages/jax/_src/linear_util.py:396: in _get_result_paths_thunk
    ans = _fun(*args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^
../pennylane/venv/lib/python3.12/site-packages/pennylane/workflow/qnode.py:857: in __call__
    return capture_qnode(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../pennylane/venv/lib/python3.12/site-packages/pennylane/workflow/_capture_qnode.py:584: in capture_qnode
    qfunc_jaxpr, out_tree = _extract_qfunc_jaxpr(qnode, abstracted_axes, *args, **kwargs)
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../pennylane/venv/lib/python3.12/site-packages/pennylane/workflow/_capture_qnode.py:461: in _extract_qfunc_jaxpr
    qfunc = qml.capture.run_autograph(qfunc) if qnode._autograph else qfunc
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../pennylane/venv/lib/python3.12/site-packages/pennylane/capture/autograph/transformer.py:197: in run_autograph
    new_fn, module, source_map = TRANSFORMER.transform(fn, user_context)
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <pennylane.capture.autograph.transformer.PennyLaneTransformer object at 0x7fa4dfb6e240>, obj = <function TestWhileLoops.test_while_loop_raises_compatibility_error_with_capture_integration.<locals>.test at 0x7fa4d8ed2340>
user_context = <malt.core.converter.ProgramContext object at 0x7fa4d3fafa70>

    def transform(self, obj, user_context):
        """Launch the transformation process. Typically, this only works on function objects.
        Here we also allow QNodes to be transformed."""
    
        # By default, AutoGraph will only convert function or method objects, not arbitrary classes
        # such as QNode objects. Here we handle them explicitly, but we might need a more general
        # way to handle these in the future.
        # We may also need to check how this interacts with other common function decorators.
        fn = obj
        if isinstance(obj, qml.QNode):
            fn = obj.func
        elif inspect.isfunction(fn) or inspect.ismethod(fn):
            pass
        elif callable(obj):
            # pylint: disable=unnecessary-lambda,unnecessary-lambda-assignment
            fn = lambda *args, **kwargs: obj(*args, **kwargs)
        else:
            raise AutoGraphError(f"Unsupported object for transformation: {type(fn)}")
    
        # Check if the function has already been converted.
        if hasattr(fn, "ag_unconverted"):
            warnings.warn(
                f"AutoGraph will not transform the function {fn} as it has already been transformed.",
                AutoGraphWarning,
            )
            new_fn, module, source_map = (
                fn,
                getattr(fn, "ag_module", None),
                getattr(fn, "ag_source_map", None),
            )
        else:
            try:
                new_fn, module, source_map = self.transform_function(fn, user_context)
            except KeyError as e:
                if "Lambda object" in str(e) and "while_loop" in inspect.getsource(fn):
>                   raise AutoGraphError(
                        "AutoGraph currently does not support lambda functions as a loop condition for `qml.while_loop`."
                        " Please define the condition using a named function rather than a lambda function."
                    ) from e
E                   pennylane.capture.autograph.ag_primitives.AutoGraphError: AutoGraph currently does not support lambda functions as a loop condition for `qml.while_loop`. Please define the condition using a named function rather than a lambda function.

../pennylane/venv/lib/python3.12/site-packages/pennylane/capture/autograph/transformer.py:80: AutoGraphError
```

JerryChen97: > Is autograph always on with program capture? (I don't see it enabled in the test)
> 
> I wonder what is different about this test than the others. Can you share the error exactly?
> 
> I'd would like to avoid just having an xfail test for no reason. What is the dependency you'd like to avoid?

I want to avoid importing and catching the `AutoGraphError` here

frontend/catalyst/utils/exceptions.py
```diff
@@ -24,3 +24,20 @@ class CompileError(Exception):
...
+    """Error raised when Catalyst control flow is used with PennyLane capture enabled.
+
+    This error indicates that Catalyst's control flow functions (for_loop, while_loop, cond)
+    cannot be used in contexts where PennyLane's capture mode is active. Users should use
+    the corresponding PennyLane functions instead for compatibility with capture functionality.
+    """
+
+    def __init__(self, control_flow_type="control flow"):
+        message = (
+            f"catalyst.{control_flow_type} is not supported with PennyLane's capture enabled. "
```
paul0403: Since `catalyst.control flow` is not a thing, I would suggest making this `control_flow_type` argument required, i.e. remove the default value `"control flow"`.

frontend/test/pytest/test_loops.py
JerryChen97: Oh never mind, I just read the error msg and replacing lambda resolved it 😓

frontend/catalyst/utils/exceptions.py
JerryChen97: https://github.com/PennyLaneAI/catalyst/pull/1945/commits/ef365156d211181cffb7e0bdd77731d1bd324f22

frontend/test/pytest/test_loops.py
JerryChen97: https://github.com/PennyLaneAI/catalyst/pull/1945/commits/6db25ece6d732c99e51a5b4f16a7c41d16ba422a

paul0403:
```
pennylane.capture.autograph.ag_primitives.AutoGraphError: AutoGraph currently does not support lambda functions as a loop condition for `qml.while_loop`. Please define the condition using a named function rather than a lambda function.
```
This is news to me :thinking: 

But I guess [this is a restriction in core PL](https://github.com/PennyLaneAI/pennylane/blob/ad4530d4d7dc1dfd59c2959dfff799d8457ba70b/pennylane/capture/autograph/transformer.py#L80), so I think there's no complicated actions needed here if following the message and changing the lambda into a function solves the issue.

JerryChen97: smells like a new story to remove it!

frontend/catalyst/utils/exceptions.py
```diff
@@ -24,3 +24,20 @@ class CompileError(Exception):
 
 class DifferentiableCompileError(CompileError):
     """An error indicating an invalid differentiation configuration."""
+
+
+class CompatibilityError(Exception):
+    """Error raised when Catalyst control flow is used with PennyLane capture enabled.
```
paul0403: If this is only for this situation, maybe name it `PlxprCaptureCFCompatibilityError` or something. `CompatibityError` sounds way too generic :sweat_smile:

JerryChen97: https://github.com/PennyLaneAI/catalyst/pull/1945/commits/66d71616258b16da097b29c814df30d013d685cf

frontend/catalyst/__init__.py
```diff
@@ -79,6 +79,7 @@
 from catalyst.passes.pass_api import pipeline
 from catalyst.utils.exceptions import (
     AutoGraphError,
+    PlxprCaptureCFCompatibilityError,
```
paul0403: Does isort not complain if it's not in alphabetical order? :scream: (Same with the string one down there)

JerryChen97:
lol apparently our own `make format` is fighting with CodeFactor

dime10:
> lol apparently our own `make format` is fighting with CodeFactor

I've never trusted that command 😆

frontend/test/pytest/test_loops.py
dime10: > This is news to me 🤔

I think it's a known problem with AG, or least at some point it was. But yeah different problem!

dime10: Thanks for fixing it @JerryChen97 !

dime10 (approved):

PR #1948, opened 2025-07-29
Enable dynamic one-shot if shots is set

rniczh:
**Context:**

This PR is to make the `one-shot` mid-circuit measurement (MCM) method as default if `shots` is set to `qml.qnode`. Currently, it's not fully supported within the compiler.

- **Issues**

1. Result Shape Handling
    1A. Normal result shape processing
    1B. CountsMP result shape handling
    1C. Shot vector result shape handling
2. Variance measurements on Observables & Unsupported measurement types by design
3. Transform measurements conflict with one-shot
4.  [ #1949 ] 
Dynamic wires handling when `qml.sample()` is used
5. Memory corruption issues
Memory corruption/use-after-free (happened randomly in runtime)
6. `from_plxpr` not work with one-shot 
7. Support return classical values (frontend/test/pytest/test_shotvector.py::TestShotVector::test_shot_vector_with_complex_container_sample)

**1. Result Shape Handling**
The result shape for Non-MCM needs further adjustment within one-shot wrapper, lots of test issues related. Mainly have `CountsMP`, tuple handling, normal result handling, and so on. 

**1A. Normal result shape processing**
    Squeezes unnecessary dimensions from measurement outputs while preserving the expected 2D structure when needed.
    
**1B. CountsMP result shape handling**
    Extracting both keys and counts from results, aggregating the counts across all shots using `sum` reduction, and restructuring the result as a tuple containing the keys and the aggregated counts.

**1C. Shot vector result shape handling**
          Added shot vector detection and reshaping logic in result processing.

**2. Variance measurements on Observables & Unsupported measurement types by design**
     Added [fallback mechanism](https://github.com/PennyLaneAI/catalyst/pull/1948/files#diff-6248f6a10c0364879c7d6f3c40b38db9d59a91a2ad9f06e8cf0fac34e0ccaaeaR144-R163) for mcm when using the one-shot method fails with unsupported measurements. Only attempt fallback when `mcm_method` was auto-determined (not explicitly set by user). For example,  [`test_dynamic_one_shot_unsupported_none_shots`](https://github.com/PennyLaneAI/catalyst/blob/rniczh/enable-dynamic-one-shot/frontend/test/pytest/test_mid_circuit_measurement.py#L495) should not fallback to `single-branch` and expected error should raise.

**3. Transform measurements conflict with one-shot**
     `measurement_from_samples` and `measurement_from_counts` cannot work properly while using `one-shot`. Currently, we raise an error when using both `one-shot` and `measurement_from_{samples/counts}` at the same time. And if the user not specify the `mcm_method`, this PR ensure it would not select the `one-shot` for this case.

**4. Dynamic wires handling when `qml.sample()` is used**
     Raise an error for `one-shot` + `sample([])` or `counts([])` + num of wires is dynamic

**6. `from_plxpr` not work with one-shot**
      Add one-shot XFail test: `frontend/test/pytest/from_plxpr/test_from_plxpr.py::TestCatalystCompareJaxpr::test_sample_with_dynamic_shot`
     
**7. Support return classical values**
     The following behaviours now allowed in one-shot mode. Originally, the output from one-shot only consider the measurements, to make it possible to let classical values happen in return values, this PR first classify the classical values from the `return_values_flat` in [jax_tracer.py](https://github.com/PennyLaneAI/catalyst/pull/1948/files#diff-e3a1c695342cc476e08c91fa99a56eaf17c6a8f78facc356039cba117f8ba7d5R1484) and used it later to insert back to the result with proper position in [qfunc.py](https://github.com/PennyLaneAI/catalyst/pull/1948/files#diff-6248f6a10c0364879c7d6f3c40b38db9d59a91a2ad9f06e8cf0fac34e0ccaaeaR391). But now, it's restrict to the case that only when dynamic one-shot applied.

**Others**
     Disable one-shot for `value_and_grad` temporarily since one issue from JAX (https://github.com/jax-ml/jax/issues/30517) encountered. We could enable it once we update to their next release version.
     
```python
import pennylane as qml

@qjit(autograph=True)
@qml.qnode(qml.device("lightning.qubit", wires=1, shots=10), mcm_method="one-shot")
def f(x):
    qml.Hadamard(0)
    if measure(0):
        x = 42
    else:
        x = 43
    return (x, measure(0), qml.expval(qml.PauliZ(0)))

print(f(42))

# Expected output: (Array([43, 43, 42, 42, 43, 43, 42, 42, 43, 43], dtype=int64), Array([False, False,  True,  True, False, False,  True,  True, False, False], dtype=bool), Array(0.2, dtype=float64))
```
      

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-93649]


frontend/catalyst/jax_tracer.py
```diff
@@ -208,6 +208,44 @@ def _eval_jaxpr(*args, **kwargs):
...
+        qnode: The quantum node to check
+        transform_name: Name of the transform to look for
+        mode: If "only_one", returns True only if transform_name is the ONLY transform
+                 in the program. If "any", returns True if transform_name is present in the program.
+
+    Returns:
+        bool: True if `transform_name` is detected (and is only one if only_one=True),
+              False otherwise
+    """
+    if not (hasattr(qnode, "transform_program") and qnode.transform_program):
```
paul0403, line 226: What's the difference between these two checks? Is the intention to:
1. First check that the qnode has the `.transform_program` attribute; if not, don't need to check anything else
2. If qnode has a transform program, check that is program is not empty; if it is empty, no need to check anything else
?

frontend/catalyst/jax_tracer.py
```diff
@@ -208,6 +208,44 @@ def _eval_jaxpr(*args, **kwargs):
...
+        return False
+
+    transform_funcs = [
+        transform_container.transform
+        for transform_container in qnode.transform_program
+        if hasattr(transform_container, "transform")
```
paul0403, line 232: Will it ever be the case that an entry in `qnode.transform_program` will not have the `transform` attribute? :thinking:

frontend/catalyst/jax_tracer.py
```diff
@@ -208,6 +208,44 @@ def _eval_jaxpr(*args, **kwargs):
...
+        for func in transform_funcs
+        if hasattr(func, "__name__") and transform_name in func.__name__
+    ]
+
+    has_target_transform = len(matching_transforms) > 0
+
+    if mode == "only_one":
+        return has_target_transform and len(transform_funcs) == 1
+
+    return has_target_transform
```
paul0403, line 246: Since this helper only counts, we don't need to copy things into and out of new lists!

```suggestion
   num_matching_transforms = 0
   for func in transform_funcs:
      if transform_name in func.__name__:
          num_matching_transforms += 1
    
    ...
```

Also, why is it `transform_name in func.__name__`? Is it only a substring?

frontend/catalyst/jax_tracer.py
```diff
@@ -208,6 +208,44 @@ def _eval_jaxpr(*args, **kwargs):
 }
 
 
+@debug_logger
+def uses_transform(qnode, transform_name, mode: Literal["only_one", "any"] = "any"):
```
paul0403, line 212: Sidenote: can you put the tracing util helpers into the helper files? `frontend/catalyst/jax_extras/tracing.py`.

frontend/catalyst/jax_tracer.py
```diff
@@ -1126,6 +1164,17 @@ def is_midcircuit_measurement(op):
     return any(map(is_midcircuit_measurement, tape.operations))
 
 
+@debug_logger
+def num_midcircuit_measurement(tape):
+    """Number of mid-circuit measurements."""
+
+    def is_midcircuit_measurement(op):
+        """Only to avoid 100 character per line limit."""
```
paul0403, line 1172:
```suggestion
        """Checks if a PennyLane operation is a midcircuit measurement."""
```

frontend/catalyst/jax_tracer.py
```diff
@@ -1194,7 +1243,19 @@ def apply_transforms(
         # TODO: Ideally we should allow qnode transforms that don't modify the measurements to
         # operate in the permissive tracing mode, but that currently leads to a small number of
         # test failures due to the different result format produced in trace_quantum_function.
-        if has_classical_outputs(flat_results):
+
+        # Special case: dynamic_one_shot with classical outputs
+        # This is allowed because dynamic_one_shot doesn't modify the semantic of measurements,
+        # it just changes how it is executed (via vmap over single shots)
+        is_dynamic_one_shot = len(qnode_program) and all(
```
paul0403, line 1250: Should it be `all` or `any`? :thinking:

frontend/catalyst/jax_tracer.py
```diff
@@ -23,7 +23,7 @@
 from dataclasses import dataclass, replace
 from enum import Enum
 from functools import partial, reduce
-from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, List, Literal, Optional, Tuple, Union
```
paul0403, line 26: I looked at some initial helpers, then I quickly realized this would be a bit difficult to review :melting_face: 

Can you maybe split this PR into several smaller PRs, one for each issue you encountered?

I think this is possible, each PR can add their own tests and specify `mcm_method="one-shot"` in the test. Then the very last PR can be one that changes the default mcm method from single branch statistics to one shot and remove all the specified `mcm_method="one-shot"` in the tests.

paul0403 (changes requested):

frontend/catalyst/jax_tracer.py
dime10: What would the different PRs contain? I think the number of (non-test) source files modified is fairly small here

paul0403: > What would the different PRs contain? I think the number of (non-test) source files modified is fairly small here

I was thinking one per issue listed in the PR description since they seemed quite seperate

rniczh:
Already separated into these PRs:
https://github.com/PennyLaneAI/catalyst/pull/2004
https://github.com/PennyLaneAI/catalyst/pull/2017
https://github.com/PennyLaneAI/catalyst/pull/2019

PR #1950, opened 2025-07-30
Ensure canonical usage of the qjit decorator

dime10:
When no kwargs are supplied, an additional function call with empty
parenthesis is not necessary and discouraged in all public use cases.

Non-functional change, docs and tests only. No changelog


sengthai (approved):
Thanks @dime10 👍

PR #1951, opened 2025-07-30
Add Partition Layers Pass and related functionality

sengthai:
Add PartitionLayersPass and related functionality.

- Introduced `PartitionLayersPass` to group quantum operations into layers based on their qubit interactions and their commutation.
- Updated `LayerOp` to support building with input and output values.
- Added a new utility function in `PauliStringWrapper` to create a wrapper from a QEC operation.
- Modified existing files to register the new pass and include necessary headers.

**-----**
For easier review, I recommend focusing on these files:
- `mlir/lib/QEC/IR/QECDialect.cpp`
- `mlir/include/QEC/Utils/QECLayer.h`
- `mlir/lib/QEC/Utils/QECLayer.cpp`
- `mlir/lib/QEC/Transforms/PartitionLayers.cpp`
- `mlir/test/QEC/PartitionLayers.mlir`

[[sc-96702]]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -33,6 +33,20 @@
   potentially to support parallel execution.
   [(#1951)](https://github.com/PennyLaneAI/catalyst/pull/1951)
   
+* Added a new JAX primitive to capture and compile the decomposition rule
+  definitions to MLIR. `decomposition_rule` is the decorator integrated
+  with this primitive for development purposes.
+  [(#1820)](https://github.com/PennyLaneAI/catalyst/pull/1820)
+
+* Renamed `QregManager` to `QubitHandler` and extended the class to manage
+  converting PLxPR wire indices into Catalyst JAXPR qubits.
+  This is especially useful for lowering subroutines that take
+  in qubits as arguments, for example decomposition rules.
+  [(#1820)](https://github.com/PennyLaneAI/catalyst/pull/1820)
+
+* Fix resource tracking unit test polluting the environment with output files
+  [(#1861)](https://github.com/PennyLaneAI/catalyst/pull/1861)
+
 * Adjoint differentiation is used by default when executing on lightning devices, significantly reduces gradient computation time.
   [(#1961)](https://github.com/PennyLaneAI/catalyst/pull/1961)
 
@@ -63,8 +77,15 @@
   %0 = transform.apply_registered_pass "some-pass" with options = {"an-option" = true, "maxValue" = 1 : i64, "multi-word-option" = 1 : i64}
   ```
 
-* `Commuting Clifford Pauli Product Rotation (PPR) operations, past non-Clifford PPRs, now supports P(π/2) Cliffords in addition to P(π/4)`
-   [(#1966)](https://github.com/PennyLaneAI/catalyst/pull/1966)
+* Added checks to raise an error when the input qubits to the multi-qubit gates in the runtime CAPI are not all distinct. 
+  [(#2006)](https://github.com/PennyLaneAI/catalyst/pull/2006).
+
+* Commuting Clifford Pauli Product Rotation (PPR) operations, past non-Clifford PPRs, now supports P(π/2) Cliffords in addition to P(π/4)
+  [(#1966)](https://github.com/PennyLaneAI/catalyst/pull/1966)
+
+* A new jax primitive `qdealloc_qb_p` is available for single qubit deallocations.
+  [(#2005)](https://github.com/PennyLaneAI/catalyst/pull/2005)
+
 
 <h3>Breaking changes 💔</h3>
 
@@ -116,6 +137,9 @@
 
 * When capture is enabled, `qjit(autograph=True)` will use capture autograph instead of catalyst autograph.
   [(#1960)](https://github.com/PennyLaneAI/catalyst/pull/1960)
+
+* `from_plxpr` can now handle dynamic shots and overridden device shots.
+  [(#1983)](https://github.com/PennyLaneAI/catalyst/pull/1983/)
 
 * QJitDevice helper `extract_backend_info` removed its redundant `capabilities` argument.
   [(#1956)](https://github.com/PennyLaneAI/catalyst/pull/1956)
@@ -212,5 +236,7 @@ David Ittah,
 Christina Lee,
 Joseph Lee,
 Andrija Paurevic,
+Ritu Thombre,
 Roberto Turrado,
-Paul Haochen Wang.
+Paul Haochen Wang,
+Jake Zaia.
--- a/.dep-versions
+++ b/.dep-versions
@@ -10,7 +10,7 @@ enzyme=v0.0.186
 
 # For a custom PL version, update the package version here and at
 # 'doc/requirements.txt'
-pennylane=0.43.0.dev34
+pennylane=0.43.0.dev43
 
 # For a custom LQ/LK version, update the package version here and at
 # 'doc/requirements.txt'
--- a/.github/workflows/determine-workflow-runner.yml
+++ b/.github/workflows/determine-workflow-runner.yml
@@ -13,7 +13,7 @@ on:
         value: ${{ jobs.determine_workflow_runner.outputs.runner_group || inputs.default_runner }}
 
 env:
-  LARGE_RUNNER_GROUP_NAME: pl-4-core-large-runner
+  LARGE_RUNNER_GROUP_NAME: blacksmith-4vcpu-ubuntu-2404
 
 jobs:
   determine_workflow_runner:
@@ -22,7 +22,7 @@ jobs:
         (
           github.event_name == 'pull_request'
           && contains(github.event.pull_request.labels.*.name, 'urgent')
-        ) && 'pl-4-core-large-runner' || 'ubuntu-24.04'
+        ) && 'blacksmith-4vcpu-ubuntu-2404' || 'ubuntu-24.04'
       }}
 
     outputs:
--- a/.github/workflows/trigger_benchmarks.yml
+++ b/.github/workflows/trigger_benchmarks.yml
@@ -1,0 +1,109 @@
+name: Trigger Benchmarks for Pull Request
+on:
+  issue_comment:
+    types:
+      - created
+
+env:
+  BENCHMARK_TRIGGER_COMMAND: /benchmark
+
+jobs:
+  trigger_benchmark_webhook:
+    runs-on: ubuntu-24.04
+
+    # Only run for pull request comments (newly created)
+    if: github.event.issue.pull_request
+
+    steps:
+      - name: Check Comment Body
+        id: comment_body
+        env:
+          COMMENT_BODY: ${{ github.event.comment.body }}
+        run: |
+          if [ "$COMMENT_BODY" = "$BENCHMARK_TRIGGER_COMMAND" ]; then
+            echo "triggers_benchmark=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "triggers_benchmark=false" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Set Repository Info
+        id: repo_info
+        env:
+          REPO_OWNER: ${{ github.repository_owner }}
+        run: |
+          echo "REPO_OWNER=$REPO_OWNER" >> $GITHUB_OUTPUT
+          echo "REPO_NAME=${GITHUB_REPOSITORY#$REPO_OWNER/}" >> $GITHUB_OUTPUT
+
+      - name: Check if commenter has write access
+        id: comment_user
+        if: steps.comment_body.outputs.triggers_benchmark == 'true'
+        uses: actions/github-script@v7
+        env:
+          REPO_OWNER: ${{ steps.repo_info.outputs.REPO_OWNER }}
+          REPO_NAME: ${{ steps.repo_info.outputs.REPO_NAME }}
+        with:
+          retries: 3
+          result-encoding: string
+          script: |
+            const user = context.payload.sender.login;
+            console.log(`Checking permissions for '${user}' on repository ${process.env.REPO_OWNER}/${process.env.REPO_NAME}}`);
+
+            const { data } = await github.rest.repos.getCollaboratorPermissionLevel({
+              owner: process.env.REPO_OWNER,
+              repo: process.env.REPO_NAME,
+              username: user
+            });
+            const userPermissions = data.user.permissions;
+            const permission = userPermissions.admin || userPermissions.maintain || (userPermissions.write || userPermissions.push);
+
+            console.log(`User '${user}' has admin/maintain/write access => ${permission}`);
+
+            return permission;
+      
+      - name: Trigger Benchmark Webhook
+        if: steps.comment_body.outputs.triggers_benchmark == 'true' && steps.comment_user.outputs.result == 'true'
+        uses: peter-evans/repository-dispatch@v3
+        env:
+          # The repo that is triggering the source webhook
+          WEBHOOK_METADATA_SOURCE_REPO: ${{ steps.repo_info.outputs.REPO_OWNER }}/${{ steps.repo_info.outputs.REPO_NAME }}
+          
+          # The URL of the actual actions run that triggered the webhook
+          WEBHOOK_METADATA_ACTIONS_RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+          
+          # The username of the person who commented the trigger command on the pull request
+          WEBHOOK_METADATA_GITHUB_ACTOR: ${{ github.event.comment.user.login }}
+          WEBHOOK_METADATA_COMMENT_LINK: ${{ github.event.comment.html_url }}
+          
+          # Information regarding the pull request
+          WEBHOOK_METADATA_PULL_REQUEST_NUMBER: ${{ github.event.issue.number }}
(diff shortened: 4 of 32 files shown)


paul0403:
Hey, I will take a look, but just wondering whether it's possible to split this into multiple smaller PRs?

mlir/include/QEC/Utils/PauliStringWrapper.h
```diff
@@ -173,5 +181,9 @@ bool isNoSizeLimit(size_t MaxPauliSize);
 // Combine the size check logic in one place
 bool exceedPauliSizeLimit(size_t pauliSize, size_t MaxPauliSize);
 
+// Check if the two Pauli string are the same
+bool equal(const PauliWord lhs, const PauliWord rhs);
+bool equal(const PauliStringWrapper &lhs, const PauliStringWrapper &rhs);
```
rniczh: I would recommend using `operator==` over explicit `euqal()` function. Because it's STL incompatible. Follow the [Google C++ Style Guide]:(https://google.github.io/styleguide/cppguide.html) 
> For a type T whose values can be compared for equality, define a non-member operator== and document when two values of type T are considered equal. If there is a single obvious notion of when a value t1 of type T is less than another such value t2 then you may also define operator<=>, which should be consistent with operator==. Prefer not to overload the other comparison and ordering operators.
Don't go out of your way to avoid defining operator overloads. For example, prefer to define ==, =, and <<, rather than Equals(), CopyFrom(), and PrintTo(). Conversely, don't define operator overloads just because other libraries expect them. For example, if your type doesn't have a natural ordering, but you want to store it in a std::set, use a custom comparator rather than overloading <.

Changes:
```cpp
bool operator==(const PauliWord &lhs, const PauliWord &rhs);
bool operator==(const PauliStringWrapper &lhs, const PauliStringWrapper &rhs);
bool operator!=(const PauliWord &lhs, const PauliWord &rhs);
bool operator!=(const PauliStringWrapper &lhs, const PauliStringWrapper &rhs);
```
And the impl:
```cpp
bool operator==(const PauliWord &lhs, const PauliWord &rhs) {
    return llvm::equal(lhs, rhs);
}

bool operator==(const PauliStringWrapper &lhs, const PauliStringWrapper &rhs)
{
    return (lhs.get_pauli_word() == rhs.get_pauli_word()) &&
           (lhs.isNegative() == rhs.isNegative());
}

bool operator!=(const PauliWord &lhs, const PauliWord &rhs) {
    return !(lhs == rhs);
}

bool operator!=(const PauliStringWrapper &lhs, const PauliStringWrapper &rhs)
{
    return !(lhs == rhs);
}
```

`TLayerReduction.cpp`:
```cpp
auto canMerge = (normalizedOps.first == normalizedOps.second);
```

mlir/lib/QEC/Transforms/TLayerReduction.cpp
```diff
@@ -0,0 +1,277 @@
...
+// Merge `rhsOp` into `mergeOp` in lhsLayer when equal under normalization.
+// To merge this we keep and update the rotation kind of `mergeOp` in lhsLayer,
+// then just remove the `rhsOp` from the rhsLayer.
+void mergePPR(QECOpInterface rhsOp, QECLayer &rhsLayer, QECOpInterface mergeOp, IRRewriter &writer)
+{
+    mergeOp.setRotationKind(mergeOp.getRotationKind() / 2);
+
+    rhsLayer.eraseOp(rhsOp);
+    writer.replaceAllUsesWith(rhsOp->getResults(), rhsOp->getOperands());
+    writer.eraseOp(rhsOp);
```
rniczh: You can just using `replaceOp`
```suggestion
    writer.replaceOp(rhsOp, rhsOp->getOperands());
```

mlir/lib/QEC/Transforms/TLayerReduction.cpp
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    writer.setInsertionPointAfter(lhsOp);
+    writer.insert(newOp);
+    lhsLayer.insertToLayer(newOp);
+
+    rhsLayer.eraseOp(rhsOp);
+    // Rewire users of the erased `rhsOp` to its original operands.
+    writer.replaceAllUsesWith(rhsOp->getResults(), rhsOp->getOperands());
+    writer.eraseOp(rhsOp);
```
rniczh: You can just using `replaceOp`
```suggestion
    writer.replaceOp(rhsOp, rhsOp->getOperands());
```

mlir/include/QEC/Utils/PauliStringWrapper.h
sengthai: Nice suggestion! Thanks @rniczh

sengthai: [df4db4e](https://github.com/PennyLaneAI/catalyst/pull/1951/commits/df4db4e0c5fcb51ed2024ad56bd08a0bbf5bd90a)

mlir/lib/QEC/Utils/QECLayer.cpp
```diff
@@ -0,0 +1,246 @@
...
+    layerEntryQubits.insert(entryQubits.begin(), entryQubits.end());
+}
+
+void QECLayer::eraseOp(QECOpInterface op) { llvm::erase(ops, op); }
+
+void QECLayer::updateResultAndOperand(QECOpInterface op)
+{
+    // Ensure layer operand set contains canonical origins for any input qubits
+    llvm::SmallVector<Value> inQubits(op.getInQubits().begin(), op.getInQubits().end());
+    llvm::SmallVector<Value> outQubits(op.getOutQubits().begin(), op.getOutQubits().end());
```
rniczh: I think you could use ValueRange here to remove redundancy
```suggestion
    ValueRange inQubits = op.getInQubits();
    ValueRange outQubits = op.getOutQubits();
```

mlir/lib/QEC/Utils/QECLayer.cpp
```diff
@@ -0,0 +1,246 @@
...
+        if (results.contains(in)) {
+            results.remove(in);
+        }
+
+        Value origin = in;
+        if (resultToOperand.contains(in)) {
+            origin = resultToOperand[in];
+        }
+        inputOrigins.push_back(origin);
+        operands.insert(origin);
```
rniczh: Do we need an extra copy here?

```suggestion
        if (resultToOperand.contains(in)) {
            in = resultToOperand[in];
        }
        inputOrigins.push_back(in);
        operands.insert(in);
```

mlir/lib/QEC/Utils/QECLayer.cpp
```diff
@@ -0,0 +1,246 @@
...
+        entries.push_back(entry);
+    }
+    return entries;
+}
+
+bool QECLayer::actOnDisjointQubits(QECOpInterface op)
+{
+    // Check for overlap with cached index set
+    return llvm::none_of(getEntryQubitsFrom(op),
+                         [&](const auto &q) { return layerEntryQubits.contains(q); });
```
rniczh, line 120: Actually, llvm provide Set api, you can use `llvm::set_intersects` here.
```suggestion
    return !llvm::set_intersects(getEntryQubitsFrom(op), layerEntryQubits);
```

sengthai: Nice, but `llvm::set_intersects` requires both containers to support `.count(E)`, and `std::vector` doesn't have that. So I think we can't use that for now, unless we update `getEntryQubitsFrom(op)`

rniczh: Good point!

mlir/include/QEC/IR/QECDialect.td
```diff
@@ -593,13 +593,14 @@ def LayerOp : QEC_Op<"layer", [SingleBlockImplicitTerminator<"YieldOp">]> {
     let builders = [
       OpBuilder<(ins
         CArg<"mlir::ValueRange", "std::nullopt">:$initArgs,
-        CArg<"llvm::function_ref<void(mlir::OpBuilder &, mlir::Location, mlir::ValueRange)>",
+        CArg<"mlir::ValueRange", "std::nullopt">:$outArgs,
+        CArg<"llvm::function_ref<void(mlir::OpBuilder &, mlir::Location, mlir::ValueRange, mlir::ValueRange&)>",
```
rniczh: Do we need `ValueRange&` here? or just only need `mlir::ValueRange`?

mlir/include/QEC/IR/QECDialect.td
```diff
@@ -593,13 +593,14 @@ def LayerOp : QEC_Op<"layer", [SingleBlockImplicitTerminator<"YieldOp">]> {
...
             "nullptr">)>
     ];
 
     let extraClassDeclaration = [{
       using BodyBuilderFn =
-          llvm::function_ref<void(mlir::OpBuilder &, mlir::Location, mlir::ValueRange)>;
+          llvm::function_ref<void(mlir::OpBuilder &, mlir::Location, mlir::ValueRange, mlir::ValueRange&)>;
```
rniczh: same as https://github.com/PennyLaneAI/catalyst/pull/1951/files#r2310594838

mlir/lib/QEC/IR/QECDialect.cpp
```diff
@@ -123,6 +125,31 @@ LogicalResult FabricateOp::verify()
...
+    OpBuilder::InsertionGuard guard(builder);
+    Location loc = result.location;
+
+    // Set the operands of the layer op
+    result.addOperands(inValues);
+
+    // Set the result types of the layer op
+    for (Value v : outValues) {
+        result.addTypes(v.getType());
+    }
```
rniczh: Since `ValueRange` has `getTypes()` api, we can simplify it
```suggestion
    result.addTypes(outValues.getTypes());
```

doc/releases/changelog-dev.md
```diff
@@ -2,6 +2,16 @@
 
 <h3>New features since last release</h3>
 
+* A new pass `--t-layer-reduction` has been added to reduce the depth and number of non-Clifford PPR
+  operations by commuting adjacent PPRs and finding possible PPRs that can be merged.
+  For more details, see the Figure 6 in [A Game of Surface Code](https://arXiv:1808.02892v3) paper.
+  [(#1975)](https://github.com/PennyLaneAI/catalyst/pull/1975)
+
+* A new pass `--partition-layers` has been added to group PPR/PPM operations into `qec.layer`
```
paul0403: If this pass is not user facing at the python level, it would be better to put it into improvements or internal changes. I think improvements is better?

mlir/include/QEC/Utils/QECLayer.h
```diff
@@ -0,0 +1,136 @@
...
+
+    // Allow move semantics
+    QECLayerContext(QECLayerContext &&) = default;
+    QECLayerContext &operator=(QECLayerContext &&) = default;
+};
+
+class QECLayer {
+  private:
+    llvm::DenseMap<mlir::Value, mlir::Value> resultToOperand;
+    QECLayerContext *context; // Reference to thread-local context
```
paul0403, line 51: Does this have to be a raw pointer? Can this be converted into a smart pointer?

mlir/lib/QEC/IR/QECDialect.cpp
```diff
@@ -17,7 +17,9 @@
 #include "mlir/IR/OpImplementation.h"
 #include "llvm/ADT/TypeSwitch.h" // needed for enums
 #include <llvm/Support/LogicalResult.h>
+#include <mlir/IR/IRMapping.h>
 #include <mlir/IR/OperationSupport.h>
+#include <mlir/IR/Region.h>
```
paul0403: Can you change these to `".../.../.h" instead of pointy brakets as well? :sweat_smile:

paul0403 (approved):
Awesome job :1st_place_medal: just two very small things from my side

mlir/lib/QEC/IR/QECDialect.cpp
sengthai: Sweet!

mlir/include/QEC/IR/QECDialect.td
sengthai: Ohh, I forgot to reason it here, but you are right. we don't don't need `&` here,

sengthai: I made the change here [5643b75](https://github.com/PennyLaneAI/catalyst/pull/1951/commits/5643b754c34301ef42d335ed35911367e71affa5)

mlir/include/QEC/Utils/QECLayer.h
sengthai: Yes, I think a raw pointer is best for this scenario because converting it to a smart pointer would imply ownership/lifetime management that QECLayer does not have. Of course, we can use `shared_ptr`, but it would force heap allocation and shared ownership that isn’t needed. So simply using a raw pointer is good here.

doc/releases/changelog-dev.md
sengthai: Sounds good!

mlir/lib/QEC/IR/QECDialect.cpp
sengthai: Sure thing! 😅

mlir/include/QEC/Utils/QECLayer.h
paul0403: > ownership/lifetime management that QECLayer does not have

So multiple layers can share the same context? I guess there's one context instance per run of the pass?

sengthai: > So multiple layers can share the same context?  I guess there's one context instance per run of the pass?

Yes Paul, that's right, but QECLayer doesn’t own the context and it just borrows a pass-scoped object. The idea is the same as MLIR implementing context. 
```cpp
class Dialect {
...
MLIRContext *context;
...
}
```

doc/releases/changelog-dev.md
sengthai: I made an update here: [5b3d00c](https://github.com/PennyLaneAI/catalyst/pull/1951/commits/5b3d00cdf9bd5059d7e38e47d380e354062d91d6)

mlir/lib/QEC/IR/QECDialect.cpp
sengthai: I updated it here [5b3d00c](https://github.com/PennyLaneAI/catalyst/pull/1951/commits/5b3d00cdf9bd5059d7e38e47d380e354062d91d6)

mlir/include/QEC/Utils/QECLayer.h
paul0403: I see, thanks!

mlir/test/QEC/TLayerReduction.mlir
```diff
@@ -0,0 +1,226 @@
...
+  // so its second PPR from second layer is moved merged to after the that first PPR from first layer.
+  // And because of these can be merged, those two PPRs are merged into a single pi/4 PPR.
+
+  // CHECK: func.func @test_t_layer_opt_5([[Q0:%.+]]: !quantum.bit, [[Q1:%.+]]: !quantum.bit, [[Q2:%.+]]: !quantum.bit)
+  // layer 1
+  // CHECK: [[q1:%.+]]:3 = qec.ppr ["X", "X", "X"](4) [[Q0]], [[Q1]], [[Q2]]
+  // CHECK: [[q2:%.+]]:3 = qec.ppr ["X", "Z", "Y"](-8) [[q1]]#0, [[q1]]#1, [[q1]]#
+  
+  // layer 2
+  // CHECK: [[q3:%.+]]:3 = qec.ppr ["Z", "Y", "X"](-8) [[q2]]#0, [[q2]]#1, [[q2]]#2
```
dime10, line 217: I don't understand how the layers are enforced / represented in the IR from these checks, would you mind explaining this? (same with the example in the docstring)

sengthai: Actually, this test is from #1975, and it was merged into this PR. So, first we wanted to perform the T depth and count reduction algorithm after constructing `layerOp`, however, it's tricky to configure the operands, so we decided to work on plain PPR/PPM instead in #1975.

paul0403: The TLayerReduction pass simply reorders the pprs based on the layer grouping algorithm. Then the partition layers pass actually constructs the layer ops.

This file is the unit test for just the T layer reduction pass. 

Related, do we have a PPR/PPM pipeline yet? @sengthai

sengthai: We haven't added `t-layer-reduction` into PPR/PPM pipeline (`ppm-compilation`) yet. Right now, it's a standalone pass for optimization.

dime10: Hmm, it's a bit weird to have a pass named "t-layer-reduction" when there aren't any actual layers 😅 How would one know that the number of layers were reduced? Would you have to run the layer once before and once after the layer reduction pass to see this?

paul0403: > Would you have to run the layer once before and once after the layer reduction pass to see this?

Good point, this seems like the case... @sengthai can you confirm?

sengthai: Nice observation! Actually, that test demonstrated a reduction in T count. In the paper, this algorithm does both to reduce the T depth and count. Since the pass name only means for T depth reduction, we should reconsider renaming it.

Issue #1220, opened 2024-10-20
Add merge rotation patterns for `qml.Rot` and `qml.CRot`


paul0403:
### Context

In quantum circuits, two neighbouring rotation gates about the same rotation axis on the same wire can be merged into one, i.e. $R_x(a) R_x(b) = R_x(a+b)$.

In #1162, a merge rotation pass was implemented in Catalyst. The pass included [qml.Rot](https://docs.pennylane.ai/en/stable/code/api/pennylane.Rot.html)/[qml.CRot](https://docs.pennylane.ai/en/stable/code/api/pennylane.CRot.html), but then it was realized that the angles in these two functions don't simply add when being merged, because y and z rotations do not commute. These two functions were removed from merge rotation pass in #1206 .

However, `qml.Rot` can be merged; it's just that the new angles of the merged gate are more complicated: https://docs.pennylane.ai/en/stable/code/api/pennylane.transforms.merge_rotations.html#details-on-rot

### Goal

We should include `qml.Rot` and `qml.CRot` in merge rotation pass, with the full angle formulas, instead of just removing them from the pattern.

Requirements:
- In Catalyst mlir, two neighbouring "Rot" gates, and two neighbouring "CRot" gates, should be merged into a single one, with the new angles appropriately given by the formulas linked to above.
- The pass must not change the functionality (aka returned state/probabilities/...) of the circuit. This should be automatic if the formulas above are followed.
- For CRot gates, since now two wires are taken in, merging should only happen if the two wires on both gates are in the same order. Note that we have a [verifier analysis](https://github.com/PennyLaneAI/catalyst/blob/main/mlir/lib/Quantum/Transforms/VerifyParentGateAnalysis.hpp) that checks this, and [is already used by the existing merge rotation pass](https://github.com/PennyLaneAI/catalyst/blob/main/mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp#L48).

### Technical details

1. Catalyst has its own [mlir `quantum` dialect](https://github.com/PennyLaneAI/catalyst/tree/main/mlir/include/Quantum/IR). The operation definitions are in `QuantumOps.td`. The operation of particular interest to us is the [`CustomOp`](https://github.com/PennyLaneAI/catalyst/blob/main/mlir/include/Quantum/IR/QuantumOps.td#L358), which corresponds to a general named gate, with the name being stored as a string.

[...]

PR #1955, opened 2025-08-01
Issue 1220 add merge rotation patterns for qml rot and qml crot

rturrado:
**Context:**

This task has to do with Catalyst's Merge Rotation pass.
When this pass was originally implemented, arbitary rotations were part of the list of rotations that could be merged.
Only to find later on that, for two arbitrary rotations, the merge was not as easy as for a fixed rotation, e.g., `RX`, where the rotation angles are simply added. At that point, arbitary rotations were removed from the set of mergeable rotations.
For an arbitary rotation, more complex mathematical computations have to be performed in order to obtain the angles of the merged gate. These mathematical formulas are well described [here](https://docs.pennylane.ai/en/stable/code/api/pennylane.transforms.single_qubit_fusion.html#derivation).
In order to add arbitrary rotations back, issue #1220 was created. This issue is also very well documented.
This PR tries to address that issue.

1. Add arbitrary rotations, `Rot` and `CRot` to the list of rotations in MergeRotationsPatterns.cpp.
2. Change the logic of `MergeRotationsRewritePattern::matchAndRewrite` so that it calls `matchAndRewriteFixedRotationOrPhaseShift` or `matchAndRewriteArbitraryRotation`.
3. Implement `matchAndRewriteArbitraryRotation` following the mathematical formulas mentioned before.
4. Add tests for the general cases of `Rot` and `CRot`, and for the special cases of merging arbitrary rotations.

Notice that, in order for the new code to compile, I have needed to include a dependency to the MLIR Math dialect, as well as include guards pointing to the Math header file.

**Benefits:**

Arbitary rotations can now be merged as part of Catalyst's Merge Rotations pass.

**Possible Drawbacks:**

There are still a few TODOs in the code.
One of them, possibly the main one, has to do with the implementation of the problematic scenarios for differentiability, which I haven't addressed yet.

**Related GitHub Issues:**

#1220


doc/releases/changelog-dev.md
```diff
@@ -125,6 +125,9 @@
 * `catalyst.accelerate`, `catalyst.debug.callback`, and `catalyst.pure_callback`, `catalyst.debug.print`, and `catalyst.debug.print_memref` now work when capture is enabled.
   [(#1902)](https://github.com/PennyLaneAI/catalyst/pull/1902)
 
+* Add merge rotation patterns for `qml.Rot` and `qml.CRot`.
+  [(#1955)](https://github.com/PennyLaneAI/catalyst/pull/1955)
```
dime10: I would could classify this as an improvement on the existing pass, and adjust the wording as follows:
```suggestion
* The merge rotation pass in Catalyst (:func:`~.passes.merge_rotations`) now also considers
  `qml.Rot` and `qml.CRot`.
  [(#1955)](https://github.com/PennyLaneAI/catalyst/pull/1955)
```

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
```diff
@@ -18,16 +18,21 @@
 #include "Quantum/Transforms/Patterns.h"
 #include "VerifyParentGateAnalysis.hpp"
 #include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/Math/IR/Math.h"
 #include "llvm/ADT/StringSet.h"
 #include "llvm/Support/Debug.h"
 #include "llvm/Support/Errc.h"
 
+#include <array>
```
dime10: It appears that not all cpp files in the repo follow our include order convention, it would be nice to fix that here :)

- stdlib at the top (like array and cassert)
- external project headers (like mlir/ and llvm/)
- local headers (like Quantum/ and Verify...)

(each section is separated by an empty line, and internally sorted alphabetically)

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
```diff
@@ -42,52 +47,285 @@ SmallVector<mlir::Value> convertOpParamsToValues(CustomOp &op, mlir::PatternRewr
...
     }
+
+    mlir::LogicalResult matchAndRewrite(OpType op, mlir::PatternRewriter &rewriter) const override
+    {
+        LLVM_DEBUG(dbgs() << "Simplifying the following operation:\n" << op << "\n");
+
+        StringRef opGateName = op.getGateName();
+        if (!fixedRotationsAndPhaseShifts.contains(opGateName) &&
+            !arbitraryRotations.contains(opGateName))
+            return failure();
```
dime10: consistent use of brackets
```suggestion
            !arbitraryRotations.contains(opGateName)) {
            return failure();
        }
```

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
```diff
@@ -42,52 +47,285 @@ SmallVector<mlir::Value> convertOpParamsToValues(CustomOp &op, mlir::PatternRewr
...
+
+        return success();
+    }
+
+    // Arbitrary single rotations require more complex maths to be merged
+    mlir::LogicalResult matchAndRewriteArbitraryRotation(OpType op,
+                                                         mlir::PatternRewriter &rewriter) const
+    {
+        ValueRange inQubits = op.getInQubits();
+        auto parentOp = dyn_cast_or_null<ParentOpType>(inQubits[0].getDefiningOp());
```
dime10: This doesn't seem like the right cast to use since we don't check the result (it looks like the original code already made this mistake though 😅).

If we already know/guarantee that the `definingOp` exists and is of `ParentOpType`, then a simple `cast` should suffice. (the cast also asserts this condition)

`dyn_cast` allows the object to be casted to be a different type than the casting target, in which case it would return null. So a `dyn_cast` should always be accompanied by a check.

`*or_null` allows the _input_ to the cast to be null, in which case the cast just propagates the null value. These casts should also always be accompanied by a check (unless they appear in a chain, but eventually you want a check).

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
```diff
@@ -42,52 +47,285 @@ SmallVector<mlir::Value> convertOpParamsToValues(CustomOp &op, mlir::PatternRewr
...
+        ValueRange parentInQubits = parentOp.getInQubits();
+        ValueRange parentInCtrlQubits = parentOp.getInCtrlQubits();
+        ValueRange parentInCtrlValues = parentOp.getInCtrlValues();
+
+        // Extract parameters of the op and its parent,
+        // promoting the parameters to mlir::Values if necessary
+        auto parentParams = convertOpParamsToValues(parentOp, rewriter);
+        auto params = convertOpParamsToValues(op, rewriter);
+
+        // TODO: do we allow Unicode in comments? What about in code?
```
dime10: comments yes, code I would say no :)

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
```diff
@@ -42,52 +47,285 @@ SmallVector<mlir::Value> convertOpParamsToValues(CustomOp &op, mlir::PatternRewr
...
+        // Parent params are ϕ1, θ1, and ω1
+        // Params are ϕ2, θ2, and ω2
+        mlir::Value phi1 = parentParams[0];
```
dime10, line 129: As you can tell, the `mlir::` prefix comes up a lot, so in our compiler we generally use `using namespace mlir;` as well as whatever namespace the current module belongs to (e.g. the quantum dialect), for convenience and visual bloat reduction. Other namespaces are generally used explicitly though, like std or different dialects.

(And headers always use the full namespace for every type.)

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
```diff
@@ -42,52 +47,285 @@ SmallVector<mlir::Value> convertOpParamsToValues(CustomOp &op, mlir::PatternRewr
...
+        mlir::Value phi2 = params[0];
+        mlir::Value theta2 = params[1];
+        mlir::Value omega2 = params[2];
+
+        auto [phi1Opt, theta1Opt, omega1Opt] = getStaticValuesOrNothing(parentParams);
+        auto [phi2Opt, theta2Opt, omega2Opt] = getStaticValuesOrNothing(params);
+
+        auto phiF = mlir::Value{};
+        auto thetaF = mlir::Value{};
+        auto omegaF = mlir::Value{};
```
dime10: Interesting, any reason to use this over
```suggestion
        Value phiF;
        Value thetaF;
        Value omegaF;
```

dime10 (commented):
Thanks a lot @rturrado, this PR is in excellent shape 💯 

Still need to check the math logic, but in the meantime I noticed a couple of minor code things:

(btw no need to merge with main unless your branch has conflicts or you need something from the main branch)

rturrado:
> Thanks a lot @rturrado, this PR is in excellent shape 💯

Thanks for the review @dime10, I'll go over it in the coming days.

> (btw no need to merge with main unless your branch has conflicts or you need something from the main branch)

Cool, thanks so much for this.

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
rturrado: Fair enough!

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
rturrado: Ooops. I'll fix this in a separate commit because I've just noticed that there was a bug related to that variable name.

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
rturrado: Beautiful. I was assuming `make format` was taking care of this.

rturrado: I don't know though why it keeps placing `mlir/...` before `llvm/...`.

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
rturrado: For this file, there seems to be an ambiguity between `llvm::Value` and `mlir::Value`. I suppose that's why it was already being.

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
rturrado: Good spot! Not really. It's just the habit of using `auto blah = type{ initializer }` for declaring variables. That, for cases like here, is overly verbose.

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
dime10: > I don't know though why it keeps placing mlir/... before llvm/....

😵‍💫 

huh, well don't worry about the alphabetical too much then, the logical grouping is the main one 👍

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
dime10: Oh my bad, I haven't experienced that before. I wonder what is causing the conflict, but you don't have worry that then 👍

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
rturrado: Thanks for spotting this.

OK, so for all the merges (fixed rotations, phase shifts, and arbitrary rotations), since Op==ParentOp==CustomOp, `VerifyHeterogeneousParentGateAndNameAnalysis` checks that we only merge gates with the same name. So, as you explained above, I think we can substitute the `dyn_cast_or_null` for an `llvm::cast`.

Likewise, for `MultiRZ`, `VerifyParentGateAnalysis` checks that only `MultiRZ` gates are being merged. So the `dyn_cast_or_null` plus `if (!parentOp)` could just be replaced by an `llvm::cast`.

```
        VerifyParentGateAnalysis<MultiRZOp> vpga(op);
        if (!vpga.getVerifierResult()) {
            return failure();
        }

        auto parentOp = dyn_cast_or_null<MultiRZOp>(inQubits[0].getDefiningOp());
        if (!parentOp)
            return failure();
```

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
```diff
@@ -42,59 +48,290 @@ SmallVector<mlir::Value> convertOpParamsToValues(CustomOp &op, mlir::PatternRewr
...
+        mlir::Value omega2 = params[2];
+
+        auto [phi1Opt, theta1Opt, omega1Opt] = getStaticValuesOrNothing(parentParams);
+        auto [phi2Opt, theta2Opt, omega2Opt] = getStaticValuesOrNothing(params);
+
+        mlir::Value phiF;
+        mlir::Value thetaF;
+        mlir::Value omegaF;
+
+        // TODO: use an epsilon for double comparison here?
```
dime10: I think it's alright, we can leave the todo here but I'd wager most of the time a zero value will be exactly zero, unless it was produced by a previous constant folding or something that happened to cancel out.

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
```diff
@@ -42,59 +48,290 @@ SmallVector<mlir::Value> convertOpParamsToValues(CustomOp &op, mlir::PatternRewr
...
+        // Special cases:
+        //
+        // 1. if (ω1 == 0 && ϕ2 == 0) { ϕF = ϕ1; θF = θ1 + θ2; ωF = ω2; }
+        // 2a. if (θ1 == 0 && θ2 == 0) { ϕF = ϕ1 + ϕ2 + ω1 + ω2; θF = 0; ωF = 0; }
+        // 2b. if (θ1 == 0) { ϕF = ϕ1 + ϕ2 + ω1; θF = θ2; ωF = ω2; }
+        // 2c. if (θ2 == 0) { ϕF = ϕ1; θF = θ1; ωF = ω1 + ω2 + ϕ2; }
+        auto zeroConst = rewriter.create<arith::ConstantOp>(loc, rewriter.getF64FloatAttr(0.0));
+        if (omega1IsZero && phi2IsZero) {
+            phiF = phi1;
+            // TODO: is it worth programming constant folding here?
```
dime10: I think the builtin folders will actually take care of this for you :) 
https://github.com/llvm/llvm-project/blob/5a076e3b4d5ec26d8944bff129c1cae67a44b3ef/mlir/lib/Dialect/Arith/IR/ArithOps.cpp#L1082

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
```diff
@@ -42,59 +48,290 @@ SmallVector<mlir::Value> convertOpParamsToValues(CustomOp &op, mlir::PatternRewr
...
+            auto cF = rewriter.create<math::SqrtOp>(
+                loc, rewriter.create<arith::AddFOp>(
+                         loc, firstAddend,
+                         rewriter.create<arith::AddFOp>(loc, secondAddend, thirdAddend)));
+
+            // TODO: can we check these problematic scenarios for differentiability by code?
+            // Problematic scenarios for differentiability:
+            //
+            // 1. if (is_close_to(cF, 0)) { /* sqrt not differentiable at 0 */ return failure(); }
+            // 2. if (is_close_to(cF, 1)) { /* acos not differentiable at 1 */ return failure(); }
```
dime10: Hmm great spot!

In general, not all programs will be differentiated though, so if we do add verification we should check it based on context. Having said that, I think it's okay to leave this as a todo for the purpose of this pr.

dime10 (approved):
Happy to merge 💯 (just looks like one of the tests failed to match)

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
rturrado: Do you find it OK to leave the TODO comment in the code? Should a new issue be open in order not to forget to address this in a future?

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
rturrado: Do you find it OK to leave the TODO comment in the code? Should a new issue be open in order not to forget to address this in a future?

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
dime10: Yes the todo here is fine :)

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
dime10: Posted here: #1967

Issue #1171, opened 2024-10-02
Use adjoint method as default grad method with lightning


dime10:
For compatibility reasons, the diff method in catalyst is set to `parameter-shift` when the user selects `diff_method="best"`. The performance will generally suffer significantly on simulators as a result of this. Implement a device-aware choice to select `adjoint` if the device supports it.

PR #1961, opened 2025-08-06
Set default diff_method to adjoint for lightning devices

josephleekl:
**Context:**
Currently parameter-shift is used by default for differentiation, when the `diff_method` is not provided. This is the most universal method, but not the most performant, especially on lightning devices which support adjoint differentiation (for expval)

For lightning devices, set `diff_method` to `adjoint` when it is set to `best`, which is the default argument.

Note: currently only a single expval is supported - this needs to be fixed. See [discussion](https://github.com/PennyLaneAI/catalyst/pull/1961#discussion_r2263890832)

**Benefits:**
Significantly faster differentiation performance on lightning devices.

**Possible Drawbacks:**

**Related GitHub Issues:**
https://github.com/PennyLaneAI/catalyst/issues/1171

[sc-74760]

### DIFF
--- a/frontend/catalyst/_version.py
+++ b/frontend/catalyst/_version.py
@@ -16,4 +16,4 @@
 Version number (major.minor.patch[-label])
 """
 
-__version__ = "0.13.0-dev29"
+__version__ = "0.13.0-dev30"


frontend/test/pytest/test_gradient.py
```diff
@@ -1273,31 +1273,7 @@ def loop(i):
     assert np.allclose(result, expected)
 
 
-def test_pytrees_return_qnode(backend):
```
josephleekl, line 1276: this test is a duplicate of [this](https://github.com/PennyLaneAI/catalyst/blob/308ea068124ed8cc59f8bbe3eb6356084843ee6d/frontend/test/pytest/test_gradient.py#L1378)

frontend/catalyst/jax_primitives_utils.py
```diff
@@ -114,15 +114,21 @@ def lower_callable_to_funcop(ctx, callable_, call_jaxpr):
...
+                    elif name == "expval":
+                        if found_expval:
+                            return False
+                        found_expval = True
                 return True
 
             device_name = getattr(getattr(callable_, "device", None), "name", None)
 
-            if device_name and "lightning" in device_name and only_expval():
+            if device_name and "lightning" in device_name and only_single_expval():
```
mlxd, line 131: Doesn't lightning support multiple expval calls with adjoint?
The PennyLane + Lightning should, no?

dime10: Yeah I'm surprised by this as well

josephleekl: https://github.com/PennyLaneAI/catalyst/blob/8a199e255822e14467d0b4b7e3f31e7ef42aa52d/mlir/lib/Gradient/Transforms/ConversionPatterns.cpp#L225

this seems to be where it's forbidden/errors out, but I'm not sure why we can't do this? @maliasadi ?

dime10: Oh weird, it looks like restriction was added only recently, it used to just require the register as a result, without mention of any expvals: 
https://github.com/PennyLaneAI/catalyst/commit/c79ab47739610edf7f1a0ff1fa0b7888b54bb898#diff-38cc90494ab765c4ef70616586894c0e7e5533cb44ae4a1767bd8657be2c9a0f

dime10: > The new pattern applicator (applyPatternsGreedily) is more aggressive in dead code elimination,
and is eliminating dead Values in the adjoint gradient method.
The nodealloc function we generate for adjoint gradient lowering used to only return the qreg, not the expval result. This causes the expval op to be eliminated since it has no users.
This further causes wrong gradient results, since the entire program, all ops included (regardless
of dead or not), impacts the gradient through chain rule.
To avoid this, we return the expval result as well.
In doing this, we implicitly assume that differentiated qnodes can only return expval.
Although this assumption is true and also restricted by frontend, ideally we should not have it hard coded.
We leave this as a TODO for a future feature.

I believe it's possible that this was a mistake. Yes we can only differentiate expvals, but I don't know if it's true that we can only differentiate _one_ expval. 

Easy way to test this I suppose by installing the prior catalyst release.

maliasadi: The C++ `adjointJacobian` doesn't support multiple MPs as it doesn't need to. Lightning without QJIT supports multiple `expval` using our Python logic for qfunc serialization as such the C++ implementation doesn't need to deal with multiple MPs as it's already taken care of in Python.
 
I believe PL's `grad` method also doesn't support generic PyTree retvals. Autograd has diff support for an autograd array of retvals, similar to `qml.jacobian`! The only way around supporting such workflows is by wrapping multiple expvals in a numpy array, or by using VJP that uses `qml.simplify` under the hood.

In Catalyst, I wasn't aware of the recent changes. We had support for multiple expval backed by the device gradient method in the past, that I'm pretty sure! 

Update: This support is still baked into the cache manager of the Runtime and the `QuantumDevice` impl for Lightning devices. So, this could be a new limitation caused by changes in the frontend and/or compiler.

dime10: Okay I did a little bit of testing:

- on `0.12` and dev: "pameter-shift" works but "adjoint" hits the assertion
- on `0.11`: the example **segfaults** for me with either method
- on `0.10`: the example **works** with both methods

So I'm not sure what happened in `0.11`, but I would say that this was actually a supported feature and at some point we introduced a regression.

I'm happy to merge this PR without it and @paul0403 can take a look at this regression when he comes back next week.

josephleekl:
> parametrize all the existing tests to run with both configurations

@dime10 do you mean `use_both_frontend`, or both `diff_method`?

dime10:
Both diff_methods :)

frontend/test/pytest/test_gradient.py
```diff
@@ -349,53 +349,70 @@ def f(tree):
...
+            match="The adjoint method can only be used for QNodes which return qml.expval",
+        ):
+            qjit(value_and_grad(workflow_variant))(1.1)
+    else:
+        result = qjit(value_and_grad(workflow_variant))(1.1)
+        expected = (workflow_variant(1.1), qjit(grad(workflow_variant))(1.1))
+        assert np.allclose(result, expected)
 
 
+@pytest.mark.parametrize("diff_method", ["parameter-shift", "adjoint", "best"])
```
dime10: Do you think it is worth it to test `"best"`? Afaik it will run one of the other two anyway, so it just duplicates tests (longer testing).
If we want to test that the "best" dispatch works as expected, a couple of unit tests that test the different scenarios without executing anything seem like a better approach.

dime10 (approved):
Thanks a lot @josephleekl

frontend/test/pytest/test_gradient.py
josephleekl: I agree it's redundant, I just put it there to be safe. In fact `"best"` is already tested with my added tests, so there's no need for these here. Happy to take them out.

josephleekl: updated https://github.com/PennyLaneAI/catalyst/pull/1961/commits/0abd9f55dd473960004747c6214539d8e28ac43d

PR #1965, opened 2025-08-07
 Add a `graph_state_prep` operation to the MBQC dialect

joeycarter:
**Context:** We require a more compact way to express graph states in the IR for MBQC workloads, and one that is more akin to eventual hardware-native instructions for allocating resources for a new graph state.

**Description of the Change:** Adds a new operation to the MBQC dialect called `mbqc.graph_state_prep`, which logically is equivalent to the PennyLane operation [`qml.ftqc.GraphStatePrep`](https://docs.pennylane.ai/en/stable/code/api/pennylane.ftqc.GraphStatePrep.html), except that it only expresses the desired form of the graph state in the IR without decomposing it to its equivalent set of quantum gate operations (such a decomposition can be added in the future as a compilation pass). For more details on this new operation, please refer to the documentation included in the op definition.

Some unit tests for other MBQC-dialect ops have also been renamed for better code organization.

**Example**

```mlir
%adj_matrix = arith.constant dense<[1, 0, 1, 0, 0, 1]> : tensor<6xi1>
%graph_reg = mbqc.graph_state_prep (%adj_matrix : tensor<6xi1>) [init "Hadamard", entangle "CZ"] : !quantum.reg
```

[sc-97319]

### DIFF
--- a/mlir/include/MBQC/IR/MBQCOps.td
+++ b/mlir/include/MBQC/IR/MBQCOps.td
@@ -92,9 +92,53 @@ def MeasureInBasisOp : MBQC_Op<"measure_in_basis"> {
 }
 
 def GraphStatePrepOp : MBQC_Op<"graph_state_prep"> {
-    let summary = "TODO";
+    let summary = "Allocate resources for a new graph state.";
     let description = [{
-        TODO
+        This operation allocates the resources for a new quantum *graph state* for use in MBQC
+        workloads. A graph state is a highly entangled group of qubits arranged according to an
+        arbitrary graph structure, where each qubit is represented by a vertex of the graph and
+        where there is an edge between every interacting pair of neighbouring qubits.
+
+        A graph state requires three parameters to be fully expressed:
+
+        1. A graph representing the qubit connectivity.
+        2. The qubit initial state. All qubits in the graph state are assumed to be initialized
+           to the same state. This parameter is expressed as the name of the one-qubit gate that
+           would have to be applied to the :math:`|0\rangle` state to realize the desired state.
+        3. The type of entangling interaction along each edge. All interactions between
+           neighbouring qubits are assumed to be the same. This parameter is expressed as the name
+           of the two-qubit gate that would have to be applied to the pairs of neighbouring qubits
+           to realize the desired entangled state.
+
+        A *densely packed adjacency matrix* is used to represent the graph structure. This
+        representations is a more compact form of the familiar adjacency matrix graph
+        representation, where only the lower-triangular part of the matrix is stored (since the
+        graph is undirected) and where the diagonal elements are excluded (since there are no qubit
+        self-interactions). The matrix values are then packed into a flat array. For example, the
+        following graph structure,
+
+        .. code-block::
+
+            0 -- 1 -- 2 -- 3
+
+        has the adjacency matrix representation (with row and column indices shown for reference):
+
+        .. code-block::
+
+                 0  1  2  3
+
+            0    0  1  0  0
+            1    1  0  1  0
+            2    0  1  0  1
+            3    0  0  1  0
+
+        The equivalent densely packed adjacency matrix has the lower-triangular part of the above
+        matrix packed into a flat array using row-major order as follows (with extra space added
+        between sequences of values to indicate the rows of the matrix):
+
+        .. code-block::
+
+            [1,   0, 1,   0, 0, 1]
     }];
 
     let arguments = (ins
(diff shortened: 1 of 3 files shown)


mlir/include/MBQC/IR/MBQCOps.td
```diff
@@ -91,4 +91,71 @@ def MeasureInBasisOp : MBQC_Op<"measure_in_basis"> {
...
+            2    0  1  0  1
+            3    0  0  1  0
+
+        The equivalent densely packed adjacency matrix has the lower-triangular part of the above
+        matrix packed into a flat array using row-major order as follows (with extra space added
+        between sequences of values to indicate the rows of the matrix):
+
+        .. code-block::
+
+            [1,   0, 1,   0, 0, 1]
```
mlxd, line 141: So, just for my understanding of the structure, can you confirm this matches with the intention:

```
                  0  1  2  3

            0    0  1  0  0
            1   [1] 0  1  0
            2   [0  1] 0  1
            3   [0  0  1] 0
```
The `[ x ]` are assumed as the values we want, and the number of entries we expect per row grows by 1 for each consecutive row? In that way, we can always index into arbitrary rows and columns using the associated row size and an offset index for the row value? So the number of elements to step-over are `f(r) = (r+1)(r)/2`, with `r` as the row (aka node) index, and the the offset into the row?

mlir/test/MBQC/DialectTest.mlir
```diff
@@ -82,9 +82,17 @@ func.func @testPostSelectInvalid(%q1 : !quantum.bit) {
...
     // expected-error@below {{op attribute 'postselect' failed to satisfy constraint}}
     %res, %new_q = mbqc.measure_in_basis [XY, %angle] %q1 postselect 2 : i1, !quantum.bit
     func.return
 }
+
+// -----
+
+func.func @test_graph_state_prep() {
+    %adj_matrix = arith.constant dense<[1, 0, 1, 0, 0, 1]> : tensor<6xi1>
+    %graph_reg = mbqc.graph_state_prep (%adj_matrix : tensor<6xi1>) [init "Hadamard", entangle "CZ"] : !quantum.reg
```
mlxd, line 96: As discussed elsewhere, happy to revisit this when we want this op to be a primitive without the CZ/H combo.

mlxd (commented):
Nice work @joeycarter 
Just a quick Q on the matrix structure for my understanding, but no blockers I can see.

dime10 (commented):
Thanks for the great docstring! no comments from my side

mlir/include/MBQC/IR/MBQCOps.td
joeycarter: Bingo, that's exactly it.

It might also be useful to note that the number of qubits, N, in the graph state is related to the number of elements, m, in the densely packed adjacency matrix via

$m = N(N - 1) / 2$

or inversely

$N = (1 + \sqrt{1 + 8m}) / 2$

joeycarter: It's also reasonably easy to map between index in the packed array and the row and column indices, e.g. for a packed array of length 6:

```cpp
int j = 1;
int k = 0;
for (int i = 0; i<6; i++) {
    std::cout << "Element i = " << i
              << "; edge [" << j << ", " << k << "]"
              << std::endl;
    k++;
    if (k == j) {
        k = 0;
        j++;
    }
}
```

Output:

```
Element i = 0; edge [1, 0]
Element i = 1; edge [2, 0]
Element i = 2; edge [2, 1]
Element i = 3; edge [3, 0]
Element i = 4; edge [3, 1]
Element i = 5; edge [3, 2]
```

mlir/test/MBQC/DialectTest.mlir
joeycarter: Sounds good :+1: 

Just to clarify, the strings after `init` and `entangle` can be any gate name, we're not limited to H and CZ. As it is now, it's up to the creator of the op to choose which gate names to insert here. Also, just to highlight that the intent here is less to say that these gates _should_ be applied to the graph-state qubits, but rather these are the gates that _would_ be applied to realize the desired graph state. For example, the way to express that the desired initial qubit state is $|+\rangle$ is to say that a Hadamard gate would be applied to a qubit in $|0\rangle$ to realize this state.

mlxd: Ah, gotcha. thanks Joey

mlir/include/MBQC/IR/MBQCOps.td
mlxd: Awesome, thanks Joey. Maybe adding some of the above to the docstring will help with the indexing/sizing info when needing to make any updates, etc.

joeycarter: Done: 6e5bb315

I've left out the code snippet that maps the indices since I plan to write a function that does exactly this when it comes time to write transforms for this op.

Issue #1713, opened 2025-05-05
Support commuting `P(pi/2)` past non-Clifford PPR


sengthai:
Currently, `commute_ppr` (https://github.com/PennyLaneAI/catalyst/pull/1563) considers only commutes `P(pi/4)`. However, `P(pi/2)` can also be commuted. The catch is that it applies a different commutation rule from `P(pi/4)P'`, so if the rule for `P(pi/2)` states that `PP' = -P'P`, then: `P(pi/2) P'(x) = -P'(x) P(pi/2)`.

This support should be extended to `commute_ppr`.

dime10:
Hi @rturrado, no problem feel free to take it :) Let me know if you have any questions!

PR #1966, opened 2025-08-07
Issue 1713: Support commuting `P(pi/2)` past non-Clifford PPR

rturrado:
**Context:**
At the moment, only pi/4 Cliffords can be commuted past non-Clifford. With this PR, pi/2 Cliffords should also be able to be commuted.

The main change is implemented in `PauliStringWrapper::computeCommutationRulesWith`. Here, we keep returning `PP'i` for a pi/4 Clifford, but we now return `-P'` for a pi/2 Clifford.
The other main change needed is to update the definition of `isClifford` in QECDialect.td. Now, a Clifford can also have a rotation of pi/2.

**Benefits:**
pi/2 Cliffords can now be commuted past non-Clifford.

**Possible Drawbacks:**
I have only added 2 anti-conmute tests. More tests may be needed.

**Related GitHub Issues:**
#1713

### DIFF
--- a/mlir/test/QEC/MergePPRIntoPPM.mlir
+++ b/mlir/test/QEC/MergePPRIntoPPM.mlir
@@ -104,7 +104,7 @@ func.func public @merge_ppr_ppm_test_7(%q1: !quantum.bit) -> tensor<i1> {
 func.func public @merge_ppr_ppm_test_8(%q1: !quantum.bit) -> (tensor<i1>, !quantum.bit) {
 
     // CHECK: qec.ppm ["Y"](-1) %
-    // CHECK: qec.ppr ["X"]
+    // CHECK: qec.ppr ["X"](2)
     %0 = qec.ppr ["X"](2) %q1: !quantum.bit
     %m, %out_qubits = qec.ppm ["Y"] %0 : !quantum.bit
     %from_elements = tensor.from_elements %m : tensor<i1>
@@ -114,7 +114,7 @@ func.func public @merge_ppr_ppm_test_8(%q1: !quantum.bit) -> (tensor<i1>, !quant
 func.func public @merge_ppr_ppm_test_9(%q1: !quantum.bit) -> (tensor<i1>, !quantum.bit) {
 
     // CHECK: qec.ppm ["Z"] %
-    // CHECK: qec.ppr ["X"]
+    // CHECK: qec.ppr ["X"](-2)
     %0 = qec.ppr ["X"](-2) %q1: !quantum.bit
     %m, %out_qubits = qec.ppm ["Z"](-1) %0 : !quantum.bit
     %from_elements = tensor.from_elements %m : tensor<i1>


doc/releases/changelog-dev.md
```diff
@@ -28,6 +28,9 @@
   %0 = transform.apply_registered_pass "some-pass" with options = {"an-option" = true, "maxValue" = 1 : i64, "multi-word-option" = 1 : i64}
   ```
 
+*  `Commuting Clifford Pauli Product Rotation (PPR) operations, past non-Clifford PPRs, now supports P(π/2) Cliffords, and not only P(π/4)`
```
```suggestion
*  `Commuting Clifford Pauli Product Rotation (PPR) operations, past non-Clifford PPRs, now supports P(π/2) Cliffords in addition to P(π/4)`
```

mlir/lib/QEC/Utils/PauliStringWrapper.cpp
```diff
@@ -79,12 +78,19 @@ bool PauliStringWrapper::commutes(const PauliStringWrapper &other) const
 PauliStringWrapper
 PauliStringWrapper::computeCommutationRulesWith(const PauliStringWrapper &rhs) const
 {
-    // P * P' * i
-    stim::FlexPauliString result =
-        (*this->pauliString) * (*rhs.pauliString) * stim::FlexPauliString::from_text("i");
-
+    stim::FlexPauliString result = *rhs.pauliString;
+    assert(llvm::isa<PPRotationOp>(this->op) && "Clifford Operation is not PPRotationOp");
+    if (auto this_op = llvm::cast<PPRotationOp>(this->op); this_op.hasPiOverTwoRotation()) {
```
sengthai: Nice change on this function! 
Just a minor point. Since we already assert the type with `isa` just above, I would recommend hoisting `llvm::cast` out of the init-if for readability. 
```suggestion
    auto this_op = llvm::cast<PPRotationOp>(this->op); 
    if (this_op.hasPiOverTwoRotation()) {
```

mlir/lib/QEC/Utils/PauliStringWrapper.cpp
```diff
@@ -79,12 +78,19 @@ bool PauliStringWrapper::commutes(const PauliStringWrapper &other) const
...
+        // -P'
+        result.value.sign = !result.value.sign;
+    }
+    else {
+        assert(this_op.hasPiOverFourRotation() &&
+               "Rotation of Clifford gate is neither pi/2 nor pi/4");
+        // P * P' * i
+        result = (*this->pauliString) * result * stim::FlexPauliString::from_text("i");
+    }
```
sengthai: I would recommend making the two Clifford cases explicit. This would keep the flow simple and encode the invariant.

```suggestion
    else if (this_op.hasPiOverFourRotation()) {
        // P * P' * i
        result = (*this->pauliString) * result * stim::FlexPauliString::from_text("i");
    }
    else {
       llvm_unreachable("Clifford rotation should be π/2 or π/4");
    }
```

sengthai (commented):
Nice work! Thanks @rturrado for this PR! It's not only support commuting pi/2 past non-Clifford PPR, but this leads to work with merging into PPM as well. 🚀

doc/releases/changelog-dev.md
rturrado: Nice, thanks!

mlir/lib/QEC/Utils/PauliStringWrapper.cpp
rturrado: Beautiful, thanks!

mlir/lib/QEC/Utils/PauliStringWrapper.cpp
rturrado: Much nicer, thanks! There seems to be a lot of useful LLVM stuff. The other day I came against this `llvm::enumerate`, now `llvm_unreachable`.

rturrado:
> Nice work! Thanks @rturrado for this PR! It's not only supporting commuting pi/2 past non-Clifford PPR, but this leads to supporting merging PPR(pi/2) into PPM as well. 🚀

Thanks @sengthai! That's a good point. When  working on this I noticed `computeCommutationRulesWith` was also being called from  MergePPRIntoPPM.cpp. But then I forgot about it. I will add a few tests using PPR(pi/2) with PPM.

rturrado:
> I will add a few tests using PPR(pi/2) with PPM.

Following on this. I think I've uncovered a bug in the MergePPRIntoPPM.mlir, and thus in the MergePPRIntoPPM code. You can test this on `main`:

- For `merge_ppr_ppm_test_2` and `merge_ppr_ppm_test_3`, change the `// CHECK-NOT:   qec.ppr["X"](4)` into `// CHECK-NOT:   qec.ppr ["X"]`.
The removal of `(4)` is not necessary, but makes the test stricter, as that check wouldn't allow, for example, a `qec.ppr ["X"](-4)`.
But the bug is in the missing space between `qec.ppr` and `["X"]`.
- Run `quantum-opt --merge-ppr-ppm --split-input-file -verify-diagnostics %s | FileCheck %s` with `%s` pointing to the MergePPRIntoPPM.mlir file.

You will see that the output contains a `qec.ppr ["X"]` for both tests (which effectively passes the `// CHECK-NOT:   qec.ppr["X"](4)`:

```
  func.func public @merge_ppr_ppm_test_2(%arg0: !quantum.bit) -> (tensor<i1>, !quantum.bit) {
    %mres, %out_qubits = qec.ppm ["Z"](-1) %arg0 : !quantum.bit
    %from_elements = tensor.from_elements %mres : tensor<i1>
    %0 = qec.ppr ["X"](4) %out_qubits : !quantum.bit
    return %from_elements, %0 : tensor<i1>, !quantum.bit
  }
  func.func public @merge_ppr_ppm_test_3(%arg0: !quantum.bit) -> (tensor<i1>, !quantum.bit) {
    %mres, %out_qubits = qec.ppm ["Y"] %arg0 : !quantum.bit
    %from_elements = tensor.from_elements %mres : tensor<i1>
    %0 = qec.ppr ["X"](-4) %out_qubits : !quantum.bit
    return %from_elements, %0 : tensor<i1>, !quantum.bit
  }
```

sengthai (approved):
@rturrado Looks great to me! 💯🚀

mlir/test/QEC/MergePPRIntoPPM.mlir
```diff
@@ -92,6 +92,39 @@ func.func public @merge_ppr_ppm_test_6(%q1: !quantum.bit, %q2: !quantum.bit) ->
...
+    %m, %out_qubits = qec.ppm ["X"] %0 : !quantum.bit
+    %from_elements = tensor.from_elements %m : tensor<i1>
+    return %from_elements : tensor<i1>
+}
+
+func.func public @merge_ppr_ppm_test_8(%q1: !quantum.bit) -> (tensor<i1>, !quantum.bit) {
+
+    // CHECK-NOT: qec.ppr ["X"]
+    // CHECK: qec.ppm ["Z"] %
+    // CHECK-NOT: qec.ppr ["X"]
```
sengthai: So, for example like this we would check if PPR after X, since the function returns its qubit.
```suggestion
    // CHECK: qec.ppm ["Y"](-1) %
    // CHECK: qec.ppr ["X"]
```

mlir/test/QEC/MergePPRIntoPPM.mlir
```diff
@@ -92,6 +92,39 @@ func.func public @merge_ppr_ppm_test_6(%q1: !quantum.bit, %q2: !quantum.bit) ->
...
+    %m, %out_qubits = qec.ppm ["Y"] %0 : !quantum.bit
+    %from_elements = tensor.from_elements %m : tensor<i1>
+    return %from_elements, %out_qubits : tensor<i1>, !quantum.bit
+}
+
+func.func public @merge_ppr_ppm_test_9(%q1: !quantum.bit) -> (tensor<i1>, !quantum.bit) {
+
+    // CHECK-NOT: qec.ppr ["X"]
+    // CHECK: qec.ppm ["Y"] %
+    // CHECK-NOT: qec.ppr ["X"]
```
sengthai: Since commuting pi/2 to the other won't change their Pauli string, but only switch its sign.
```suggestion
    // CHECK: qec.ppm ["Z"] %
    // CHECK: qec.ppr ["X"]
```

mlir/test/QEC/MergePPRIntoPPM.mlir
```diff
@@ -28,9 +28,9 @@ func.func public @merge_ppr_ppm_test_1(%q1: !quantum.bit) -> tensor<i1> {
 
 func.func public @merge_ppr_ppm_test_2(%q1: !quantum.bit) -> (tensor<i1>, !quantum.bit) {
 
-    // CHECK-NOT:   qec.ppr["X"](4)
+    // CHECK-NOT:   qec.ppr ["X"]
```
sengthai:
```suggestion
```

mlir/test/QEC/MergePPRIntoPPM.mlir
```diff
@@ -28,9 +28,9 @@ func.func public @merge_ppr_ppm_test_1(%q1: !quantum.bit) -> tensor<i1> {
...
     // CHECK:       qec.ppm ["Z"](-1) %
-    // CHECK-NOT:   qec.ppr["X"](4)
+    // CHECK-NOT:   qec.ppr ["X"]
```
sengthai:
```suggestion
    // CHECK:       qec.ppr ["X"](4)
```

mlir/test/QEC/MergePPRIntoPPM.mlir
```diff
@@ -39,9 +39,9 @@ func.func public @merge_ppr_ppm_test_2(%q1: !quantum.bit) -> (tensor<i1>, !quant
 
 func.func public @merge_ppr_ppm_test_3(%q1: !quantum.bit) -> (tensor<i1>, !quantum.bit) {
 
-    // CHECK-NOT:   qec.ppr["X"](4)
+    // CHECK-NOT:   qec.ppr ["X"]
```
sengthai:
```suggestion
```

mlir/test/QEC/MergePPRIntoPPM.mlir
```diff
@@ -39,9 +39,9 @@ func.func public @merge_ppr_ppm_test_2(%q1: !quantum.bit) -> (tensor<i1>, !quant
...
     // CHECK:       qec.ppm ["Y"] %
-    // CHECK-NOT:   qec.ppr["X"](4)
+    // CHECK-NOT:   qec.ppr ["X"]
```
sengthai:
```suggestion
   // CHECK:       qec.ppr ["X"](-4)
```

mlir/test/QEC/MergePPRIntoPPM.mlir
rturrado: Many thanks for the explanation!

mlir/test/QEC/MergePPRIntoPPM.mlir
rturrado: Roger that!

rturrado:
> Only one question about those code changes. Don't you want to keep the `CHECK-NOT: qec.ppr` before the `CHECK: qec.ppm` for tests 2, 3, 8, and 9? To ensure that PPRs were actually moved past the PPM?

Ah, OK, I understand why that `CHECK-NOT` doesn't make much sense. The test uses one PPR. After the merge pass is run you just check that that PPR sits now after the PPM.

PR #1973, opened 2025-08-11
Improved `cond` support in `from_plxpr`

albi3ro:
**Context:**

We want to switch to using program capture as the frontend for catalyst, so we need better coverage of features.

Allows `cond` to be translated when it is outside the qnode.

Allows `cond` to have elif branches.

Allows an empty false branch when the function has arguments.

Makes sure the predicates are always booleans.

A couple other smaller fixes.

**Benefits:**

Closer to using plxpr as the capture frontend.

**Possible Drawbacks:**

Still some open questions and tests marked xfail for now. 

**Related GitHub Issues:**
[sc-95436]


frontend/catalyst/from_plxpr/control_flow.py
```diff
@@ -31,6 +31,58 @@
...
+
+
+def _to_bool_if_not(arg):
+    if getattr(arg, "dtype", None) == jax.numpy.bool:
+        return arg
+    return jax.numpy.bool(arg)
+
+
+@WorkflowInterpreter.register_primitive(plxpr_cond_prim)
+def _(self, *plxpr_invals, jaxpr_branches, consts_slices, args_slice):
```
paul0403: Should this function be deleted?

frontend/catalyst/from_plxpr/control_flow.py
```diff
@@ -51,31 +103,17 @@ def handle_cond(self, *plxpr_invals, jaxpr_branches, consts_slices, args_slice):
 
         if plxpr_branch is None:
             # Emit a new Catalyst jaxpr branch that simply returns a qreg
-            converted_jaxpr_branch = jax.make_jaxpr(lambda x: x)(*args_plus_qreg).jaxpr
+            converted_jaxpr_branch = jax.make_jaxpr(lambda *x: x[-1])(*args_plus_qreg).jaxpr
```
paul0403: What is this change about?

frontend/catalyst/from_plxpr/control_flow.py
albi3ro: ? The main point of this PR is adding this function?

frontend/catalyst/from_plxpr/control_flow.py
albi3ro:
```
qml.cond(True, RX)(0.5, 0)
```
We want the converted false branch to only return the qreg, not the phase and wire.

albi3ro: Note that this PR that just went into master will make this branch unnecessary.  Will version bump the PR tomorrow.

https://github.com/PennyLaneAI/pennylane/pull/8080

frontend/catalyst/from_plxpr/control_flow.py
paul0403: Ah sorry, I saw the `_` name and my brain thought it's unused. :sweat_smile:  

Let's call it `workflow_cond` like the others!

frontend/test/pytest/test_conditionals.py
```diff
@@ -425,36 +494,52 @@ def f(x: int):
...
+                TypeError,
+                match="Conditional 'else if' function can have arguments only if it is a PennyLane gate.",  # pylint:disable=line-too-long
+            ):
+                qjit(h)
 
     @pytest.mark.usefixtures("disable_capture")
     def test_cond_raises_compatibility_error_with_capture(self):
         """Test that cond raises PlxprCaptureCFCompatibilityError when capture mode is enabled."""
+        if not qml.capture.enable():
+            pytest.skip("capture only test")
```
paul0403: I think it's not just that this test is capture-only. Now that the `cond` in this test file is `qml.cond` (instead of `catalyst.cond`), and this test file uses both modes, to maintain the original purpose this test should be run with capture (since it was run with capture originally), but the `cond` in this test should also be `catalyst.cond`?

Same with the one below.

frontend/test/pytest/test_conditionals.py
```diff
@@ -663,7 +761,8 @@ def else_fn():
                 return 0
 
             cond_fn()
-            assert isinstance(cond_fn.operation, api_extensions.control_flow.Cond)
+            if not qml.capture.enabled():
+                assert isinstance(cond_fn.operation, api_extensions.control_flow.Cond)
```
paul0403, line 881: Do we need a corresponding assert in capture mode here?

frontend/test/pytest/test_conditionals.py
```diff
@@ -263,18 +305,30 @@ def cond_fn():
...
     def test_branch_multi_return_type_unification_qnode_1(self, backend):
         """Test that an exception is not raised when the return types of all branches do not match
         but could be unified."""
 
+        if qml.capture.enabled():
+            pytest.xfail("capture does not allow returning mcm's or classical values")
+
         @qjit
-        @qml.qnode(qml.device(backend, wires=1))
+        @qml.qnode(qml.device(backend, wires=1), autograph=False)
```
paul0403, line 334: What are the autograph changes about? Does qnode have autograph turned on by default in capture?

frontend/test/pytest/test_conditionals.py
```diff
@@ -915,11 +1014,15 @@ def workflow(x):
...
-            workflow(3)
+        if qml.capture.enabled():
+            with pytest.raises(TypeError, match="is not a valid JAX type"):
+                workflow(3)
+        else:
+            with pytest.raises(
+                TypeError,
+                match="Conditional predicates are required to be of bool, integer or float type",
+            ):
+                workflow(3)
```
paul0403, line 1141: Nothing to do here, but probably it would be a good idea to unify the error messages inside and outside capture.

frontend/test/pytest/test_conditionals.py
albi3ro: Yep. We will be removing `autograph` from the qnode soon, but for now it is there and on by default.  And it was causing various issues.

frontend/catalyst/from_plxpr/control_flow.py
albi3ro:
```suggestion
def workflow_cond(self, *plxpr_invals, jaxpr_branches, consts_slices, args_slice):
```

frontend/test/pytest/test_conditionals.py
albi3ro: I'm not sure exactly what it is testing, other than it seems a bit implementation specific.

frontend/test/pytest/test_conditionals.py
albi3ro: I think  as long as the error messages continue to be informative enough about the problem, we should be ok. But yes, there is some room for improving error messages with program capture.

frontend/test/pytest/test_conditionals.py
paul0403: I just took a look, I remember what this is about now. It's a convenience interface to get the internal `catalyst.api_extension.Cond` object (basically the "impl" object for catalyst.cond). 

Since this is strictly catalyst internals we are all good here :ok_hand:

frontend/test/pytest/test_conditionals.py
albi3ro: Turns out I was using `enable()` instead of `enabled()` 🤦‍♀️ . Now properly running this test and using `catalyst_cond`.

frontend/test/pytest/test_conditionals.py
```diff
@@ -11,19 +11,31 @@
...
-
+# pylint: disable=too-many-lines
 from textwrap import dedent
 
 import jax
 import jax.numpy as jnp
 import numpy as np
 import pennylane as qml
 import pytest
+from pennylane import cond
```
paul0403, line 22: Just want to make sure I understand the updated testing strategy here. 

The `cond` in this file will all be `qml.cond` now. This file also runs the tests both with and without capture. With the capture path `qml.cond` will go to catalyst's from_plxpr through the new code added in this PR. What happens to the no-capture path? Has `qml.cond` and `catalyst.cond`'s dispatching been completed there?

paul0403 (approved):
NIce! 

I think the test failures look like PL no longer squeezing out `1`-dimensions? (from the version bump in this PR)

frontend/test/pytest/test_conditionals.py
albi3ro: `qml.cond` dispatches to `catalsyt.cond` if capture is not enabled.  So the only difference is that `qml.cond` can always be used, but `catalyst.cond` can only be used with non-capture qjit.

albi3ro:
> NIce!
> 
> I think the test failures look like PL no longer squeezing out `1`-dimensions? (from the version bump in this PR)

Yep.  Trying to figure out the version bump at the moment.  So this PR will just have to wait for that to happen.

PR #1977, opened 2025-08-12
Fix type promotion on branch

### LINKED ISSUE #1958: Type demotion with `cond`
For the code:

```
@qml.qjit(autograph=True)
@qml.qnode(qml.device('lightning.qubit', wires=1))
def circuit(cond1, cond2):
    if cond1:
        res = False
    elif cond2:
        res = 0.5
    else:
        res = False

    return res

circuit(False, True)
```
```
Array(True, dtype=bool)
```

I would expect either

1) an error about mismatching branch return types
2) a promotion of all outputs to float

instead of a demotion of the float to a boolean.

rniczh:
**Context:**

```python
@qml.qjit(autograph=True)
@qml.qnode(qml.device('lightning.qubit', wires=1))
def circuit(cond1, cond2):
    if cond1:
        res = False
    elif cond2:
        res = 0.5
    else:
        res = False

    return res

circuit(False, True)
```

Should return value with type float64 instead of bool 

Choose the type that match the promoted type.

For an example:

```
branch_avals:    [[ShapedArray(float64[], weak_type=True)], [ShapedArray(bool[])]]
promoted_dtypes: [dtype('float64')]
```

Expected to return matched type `((ShapedArray(float64[], weak_type=True), ...),)`

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

Fixes #1958 

[sc-97050]

### DIFF
--- a/frontend/catalyst/api_extensions/control_flow.py
+++ b/frontend/catalyst/api_extensions/control_flow.py
@@ -218,7 +218,8 @@ def cond(pred: DynamicJaxprTracer):
         ...         return x ** 2
         ...     return cond_fn()
         TypeError: Conditional requires a consistent return structure across all branches! Got
-        PyTreeDef(None) and PyTreeDef(*).
+        PyTreeDef(None) and PyTreeDef(*). Please specify an else branch if PyTreeDef(None) was
+        specified.
 
         >>> @qjit
         ... def f(x: float):


frontend/test/pytest/test_conditionals.py
```diff
@@ -311,6 +311,29 @@ def cond_else():
...
+            @cond_fn.else_if(cond2)
+            def cond_fn_2():
+                return 0.5
+
+            @cond_fn.otherwise
+            def cond_fn_3():
+                return False
+
+            r = cond_fn()
+            assert r.dtype is jnp.dtype("float64")
```
joeycarter: It might be more future-proof to check if JAX 64-bit precision is enabled before asserting the type is `float64`, e.g.

```suggestion
            assert r.dtype is jnp.dtype("float64") if jax.config.values["jax_enable_x64"] else r.dtype is jnp.dtype("float32") 
```

joeycarter (changes requested):
Thanks @rniczh! I had one minor comment on the test you included. Also, can you add a changelog entry under "Bug fixes"?

frontend/catalyst/api_extensions/control_flow.py
```diff
@@ -774,18 +775,21 @@ def _call_with_quantum_ctx(self):
...
-        # all_jaxprs = [s.out_initial_jaxpr() for s in out_sigs]
-        # all_noimplouts = [s.num_implicit_outputs() for s in out_sigs]
-        # _, out_type, _, all_consts = unify_convert_result_types(
-        #     all_jaxprs, all_consts, all_noimplouts
-        # )
-        # Unfortunately, we can not do this beacuse some tracers (specifically, the results of
-        # ``qml.measure``) might not have their source Jaxpr equation yet. Thus, we delay the
-        # unification until the quantum tracing is done. The consequence of that: we have to guess
-        # the output type now and if we fail to do so, we might face MLIR type error down the
-        # pipeline.
```
dime10, line 787: Did you look into this comment? Do you think it's no longer a concern, or are we just ignoring it?

rniczh: We still cannot do `unify_convert_result_types`, since it depend on jaxpr equation, but we don't have the jaxpr equation for the hybrid op at this point.

But actually, we don't need to construct the new jaxpr when selecting the output type of the `cond`. So just only use the `promoted_types = _promote_jaxpr_types([[v.aval for v in j.outvars] for j in jaxprs])` in `unify_convert_result_types` is far enough.

frontend/catalyst/api_extensions/control_flow.py
```diff
@@ -774,18 +775,21 @@ def _call_with_quantum_ctx(self):
...
+
+        if promoted_dtypes:
+            out_type = next(
+                (
+                    cand_out_type
+                    for cand_out_type in out_types
+                    if cand_out_type[0][0].dtype == promoted_dtypes[0]
+                ),
+                out_types[-1],
+            )
```
rniczh: Just realize, only match the first data type is not enough

For an example:
e.g. `return {0: True, 1: 0.5}` will have the type: `((ShapedArray(bool[]), True), (ShapedArray(float64[], weak_type=True), True))`

So we might need to reconstruct the `out_type` here

rniczh: Resolved

frontend/catalyst/api_extensions/control_flow.py
dime10: Good point, I wonder what issue the original author was talking about then since we can just look at the output types of each branch and promote those 🤔 Anyway, thanks for looking into it!

frontend/catalyst/api_extensions/control_flow.py
```diff
@@ -774,18 +775,38 @@ def _call_with_quantum_ctx(self):
...
-        # the output type now and if we fail to do so, we might face MLIR type error down the
-        # pipeline.
+
+        # Select the output type of the one with the promoted dtype among all branches
         out_type = out_types[-1]
+        branch_avals = [[aval for aval, _ in branch_out_type] for branch_out_type in out_types]
+        promoted_dtypes = _promote_jaxpr_types(branch_avals)
+
+        if promoted_dtypes:
+            columns = list(zip(*out_types))
```
dime10: I'm having a little trouble reading the code below. Is the procedure we're doing not 
- unpack the avals from the `(aval, bool)` sequence (i.e. the out_types)
- promote the avals to a common type
- repack the promoted avals into the `(aval, bool)` sequence

?

Since there are multiple type sequences at the start and only one at the end, it does beg the question though what we do with the "explicit" flag (the bool). Do we assert that each branch has the same explicit/implicit signature? 
That would certainly make the repacking step above very straightforward with something like the following:
```py
out_type = [(aval.update(dtype=dtype), expl) for dtype, (aval, expl) in zip(promoted_dtypes, out_type)]
```

rniczh: Thanks! You are right, It much cleaner to just `update` the dtype!

frontend/catalyst/api_extensions/control_flow.py
```diff
@@ -782,31 +782,10 @@ def _call_with_quantum_ctx(self):
         promoted_dtypes = _promote_jaxpr_types(branch_avals)
 
         if promoted_dtypes:
```
dime10: Is there ever a case where `_promote_jaxpr_types` returns nothing?

frontend/catalyst/api_extensions/control_flow.py
dime10: Great :)

dime10: And we know that the explicit flag sequence is the same for all branches?

frontend/catalyst/api_extensions/control_flow.py
rniczh: Yes, for an example, `_promote_jaxpr_types` would return `[]` in the following case

```python
@qjit
@qml.qnode(qml.device(backend, wires=1))
def circuit(x: float):

    def ansatz_true():
        qml.RX(x, wires=0)
        qml.Hadamard(wires=0)

    qml.cond(x > 1.4, ansatz_true)()

    return qml.expval(qml.Z(0))
```

dime10: Got it. The for loop in the list comprehension would just exist out immediately in that case, so we can save the extra if statement in the code :)

frontend/catalyst/api_extensions/control_flow.py
rniczh: I haven't see any explicit promise that all branches should have the same explicit flag sequence in the `_call_with_quantum_ctx`. But if not, the assumption of `nimplouts` would fail right?

dime10: Maybe it would be worth it to try and break this then to see what happens. That is create an example that has different implicit outputs across the branches.

I'm actually not sure what would happen because it seems like later the code does collect one `nimplout` for each branch, and uses it for something 🤔

rniczh: Seems like it would first run into an assertion on checking different things but related.

```
signatures = [[ShapedArray(int64[], weak_type=True), f64[InDBIdx(val=0)]], [ShapedArray(float64[3])]]

    def _assert_cond_result_types(signatures: List[List[AbstractValue]]):
        """Ensure a consistent type signature across branch results."""
        num_results = len(signatures[0])
>       assert all(len(sig) == num_results for sig in signatures), "mismatch: number or results"
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E       AssertionError: mismatch: number or results
```

The out_types from the following example
```
DEBUG out_types[0] ((ShapedArray(int64[], weak_type=True), False), (f64[InDBIdx(val=0)], True))
DEBUG out_types[1] ((ShapedArray(float64[3]), True),)
```

Reproduce:
```python
@qjit
@qml.qnode(qml.device("lightning.qubit", wires=4))
def f(flag, sz):
    a = jnp.ones([sz], dtype=float)
    b = jnp.zeros([3], dtype=float)

    @cond(flag)
    def case():
        return a

    @case.otherwise
    def case():
        return b

    c = case()
    return c
```

dime10: Nice, thanks for checking. In that case we might want to change the assert to an exception since it can be encountered by the user from the frontend.

rniczh: But just wonder, even the frontend user encounter this, they may not easily to understand what happen and how to solve it.

Is it worth to do another check before that and raise an error more clearly describe the root cause?

dime10: Ah yes definitely, can you give it a clearer message for the user? :) Or if you think there is a better (earlier) place for this check that's good too, but I think here is when we first analyze the code the user provided.

doc/releases/changelog-dev.md
```diff
@@ -56,6 +56,9 @@
 
 <h3>Bug fixes 🐛</h3>
 
+* Fix type promotion on branch
```
paul0403: I feel like the changelog entry should be a bit more descriptive :sweat_smile:

frontend/test/pytest/test_conditionals.py
```diff
@@ -311,6 +313,57 @@ def cond_else():
...
+
+            @cond_fn.otherwise
+            def cond_fn_3():
+                return {0: True, 1: False}
+
+            r = cond_fn()
+            expected_dtype = jnp.dtype(
+                "float64" if jax.config.values["jax_enable_x64"] else "float32"
+            )
+            assert all(r[i].dtype is expected_dtype for i in range(len(r)))
```
paul0403:
```suggestion
            assert all(v.dtype is expected_dtype for _, v in r.items())
```

paul0403:
I just realized, does this mean this line in documentation is no longer true? https://docs.pennylane.ai/projects/catalyst/en/stable/code/api/catalyst.cond.html

image

frontend/catalyst/api_extensions/control_flow.py
rniczh: ~I added a explict flag check for this~ Yeap, I think this place is enough, just add clearer error message

paul0403:
> @paul0403
> 
> > I just realized, does this mean this line in documentation is no longer true? https://docs.pennylane.ai/projects/catalyst/en/stable/code/api/catalyst.cond.html
> 
> Some tests are actually try to test this by returning different dtype. I also have the same question on the issue #1958 And run into test failures when trying to raise an error for this.

Ah, so I guess it is ok for user code to have different types in different branches, but this PR you added will make the entire `cond`'s output type to be the biggest type across all branches?

frontend/test/pytest/test_conditionals.py
```diff
@@ -311,6 +313,57 @@ def cond_else():
...
+            def cond_fn_3():
+                return False
+
+            r = cond_fn()
+            assert r.dtype is jnp.dtype(
+                "float64" if jax.config.values["jax_enable_x64"] else "float32"
+            )
+            return r
+
+        assert 0.5 == circuit(False, True)
```
paul0403, line 339: Can you add another test that returns false on the second branch but 0.5 on the third, and also call it with `False, True`? Just to make sure that even the bigger type (float) is not on the actual executed branch, we still promote to the bigger type, which would be 0.0 in this test.

dime10:
@paul0403 We already had type promotion in place, but it didn't work correctly in all cases. This PR fixes that.

I guess strictly speaking the statement is wrong, and it should say "if provided, return types need to be identical or at least promotable across both branches''.

frontend/test/pytest/test_conditionals.py
rniczh: Added

rniczh:
> @paul0403 We already had type promotion in place, but it didn't work correctly in all cases. This PR fixes that.
> 
> I guess strictly speaking the statement is wrong, and it should say "if provided, return types need to be identical or at least promotable across both branches''.

Thanks. I've made corrections and added information to the doc.

frontend/catalyst/api_extensions/control_flow.py
```diff
@@ -224,10 +217,8 @@ def conditional_fn():
         ...     def cond_fn():
         ...         return x ** 2
         ...     return cond_fn()
-        TypeError: Conditional requires consistent return types across all branches, got:
-        - Branch at index 0: [ShapedArray(float64[], weak_type=True)]
-        - Branch at index 1: []
-        Please specify an else branch if none was specified.
+        TypeError: Conditional requires a consistent return structure across all branches! Got
+        PyTreeDef(None) and PyTreeDef(None).
```
dime10: This looks weird, the previous message was clearer I think and also `None` and `None` for both branches doesn't sound right

frontend/catalyst/api_extensions/control_flow.py
```diff
@@ -185,31 +187,22 @@ def conditional_fn():
...
 
-        The return values of all branches of :func:`~.cond` must be the same type.
-        Returning different types, or ommitting a return value in one branch (e.g.,
+        The return values of all branches of :func:`~.cond` must be the same shape.
+        Returning different shapes, or ommitting a return value in one branch (e.g.,
         returning ``None``) but not in others will result in an error.
 
+        However, the return values of all branches of :func:`~.cond` can be different data types.
+        In this case, the return types need to be identical or at least promotable across both
+        branches.
```
dime10:
```suggestion
        In this case, the return types will automatically be promoted to the next common larger type.
```

dime10 (approved):
Thanks!

frontend/catalyst/api_extensions/control_flow.py
rniczh: Typo, the rhs should be `PyTreeDef(*)`. But seems like this error is captured by the different check instead of the original one.

rniczh: I've added a sentence to this error message to clarify the need for an `else` branch when a `PyTreeDef(None)` is present.

PR #1952, opened 2025-07-31
Replacing `device(..., shots=...)` with `set_shots`

JerryChen97:
**Context:**
PennyLane is [deprecating](https://github.com/PennyLaneAI/pennylane/pull/7979) the interface of setting shots on device level. Hopefully this PR can help Catalyst to avoid `PennyLaneDeprecationWarning`.
We will start from tests and then carefully check if within source code there's any places to modify

**Description of the Change:**
Source code:
 - `get_device_capabilities` accepts an extra kwarg `shots=` in order to be able to receive `tape.shots`.
 - The `capabilities` of `QJITDevice` has been heavily adjusted, but only to extend to accept `shots` information from `tape(s)`. As the typical deprecation cycle intended all the time, the previous `QJITDevice.shots` and `QJITDevice.capabilities` passes keep as the same as before if a user want to use them.
 - `validate_measurements` arg `shots` has been changed from positional to keyword. Internally, this help prefers the shots of its input `tape`
 - `QFunc` does not re-initialize the copied devices with shots any more.

Tests:
 - A lot. But just to transit from `device(..., shots=...)` to `set_shots`.
 - Tests related to `capture` remain the same but with warnings caught.

**Benefits:**

**Possible Drawbacks:**
Unlike a typical deprecation, where we completely remove the deprecated usage, this one will be co-existing with the new UI.

**Related GitHub Issues:**
[sc-95431]

PR #1980, opened 2025-08-13
Add `SingleExcitation` and `DoubleExcitation` to RuntimeCAPI.

josephleekl:
### Before submitting

**Context:**
Lightning Devices support SingleExcitation and DoubleExcitation, and with the corresponding qis we will be able to target it directly in the runtime, without further decomposing.

Add SingleExcitation and DoubleExcitation to RuntimeCAPI

**Benefits:**
Reduce need to decompose SingleExcitation and accelerate XAS benchmark with capture enabled + qjit.

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/frontend/catalyst/device/qjit_device.py
+++ b/frontend/catalyst/device/qjit_device.py
@@ -72,6 +72,7 @@ RUNTIME_OPERATIONS = [
     "IsingYY",
     "IsingZZ",
     "SingleExcitation",
+    "DoubleExcitation",
     "ISWAP",
     "MultiRZ",
     "PauliX",
--- a/frontend/catalyst/device/qjit_device.py
+++ b/frontend/catalyst/device/qjit_device.py
@@ -71,6 +71,7 @@ RUNTIME_OPERATIONS = [
     "IsingXY",
     "IsingYY",
     "IsingZZ",
+    "SingleExcitation",
     "ISWAP",
     "MultiRZ",
     "PauliX",
--- a/frontend/test/pytest/test_operations.py
+++ b/frontend/test/pytest/test_operations.py
@@ -262,7 +262,7 @@ def test_singleexcitation():
 
     assert np.allclose(circuit(0.1), circuit.original_function(0.1))
     assert "SingleExcitation" in str(circuit.jaxpr)
-    assert "SingleExcitation" in circuit.mlir
+    assert 'quantum.custom "SingleExcitation"' in circuit.mlir
 
 
 def test_doubleexcitation():
@@ -278,7 +278,7 @@ def test_doubleexcitation():
 
     assert np.allclose(circuit(0.1), circuit.original_function(0.1))
     assert "DoubleExcitation" in str(circuit.jaxpr)
-    assert "DoubleExcitation" in circuit.mlir
+    assert "quantum.custom "DoubleExcitation"" in circuit.mlir
 
 
 if __name__ == "__main__":
--- a/frontend/test/pytest/test_operations.py
+++ b/frontend/test/pytest/test_operations.py
@@ -127,6 +127,9 @@ def test_param(backend):
         qml.SingleExcitation(x, wires=[0, 1])
         qml.SingleExcitation(y, wires=[1, 2])
 
+        qml.DoubleExcitation(x, wires=[0, 1, 2, 3])
+        qml.DoubleExcitation(y, wires=[2, 3, 0, 1])
+
         qml.CRX(x, wires=[0, 1])
         qml.CRY(x, wires=[0, 1])
         qml.CRZ(x, wires=[0, 1])
--- a/frontend/test/pytest/test_operations.py
+++ b/frontend/test/pytest/test_operations.py
@@ -278,7 +278,7 @@ def test_doubleexcitation():
 
     assert np.allclose(circuit(0.1), circuit.original_function(0.1))
     assert "DoubleExcitation" in str(circuit.jaxpr)
-    assert "quantum.custom "DoubleExcitation"" in circuit.mlir
+    assert 'quantum.custom "DoubleExcitation"' in circuit.mlir
 
 
 if __name__ == "__main__":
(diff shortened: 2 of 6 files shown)


runtime/lib/capi/RuntimeCAPI.cpp
```diff
@@ -678,6 +678,15 @@ void __catalyst__qis__IsingZZ(double theta, QUBIT *control, QUBIT *target,
                                           /* modifiers */ MODIFIERS_ARGS(modifiers));
 }
 
+void __catalyst__qis__SingleExcitation(double phi, QUBIT *wire0, QUBIT *wire1,
```
maliasadi, line 681: We need to update `RUNTIME_OPERATIONS` in `qjit_device.py` and add integration tests to `tests/pytest/test_operations.py`

runtime/include/RuntimeCAPI.h
```diff
@@ -70,6 +70,7 @@ void __catalyst__qis__IsingXX(double, QUBIT *, QUBIT *, const Modifiers *);
 void __catalyst__qis__IsingYY(double, QUBIT *, QUBIT *, const Modifiers *);
 void __catalyst__qis__IsingXY(double, QUBIT *, QUBIT *, const Modifiers *);
 void __catalyst__qis__IsingZZ(double, QUBIT *, QUBIT *, const Modifiers *);
+void __catalyst__qis__SingleExcitation(double, QUBIT *, QUBIT *, const Modifiers *);
```
mlxd, line 73: Just a quick fly-by: is the goal to also add `DoubleExcitation` after this PR?

josephleekl: we're adding this specifically for a bottleneck in the XAS benchmark - but we can add all of them in one go? I don't have too much opinion about this @maliasadi ?

mlxd: Most workloads I've seen that use SE also use (a lot more of) DE, so just curious if this was the case. If not, all good.

josephleekl: in this case I've added them too.

runtime/lib/capi/RuntimeCAPI.cpp
josephleekl: tests added!

frontend/catalyst/device/qjit_device.py
```diff
@@ -71,6 +71,8 @@
     "IsingXY",
     "IsingYY",
     "IsingZZ",
+    "SingleExcitation",
+    "DoubleExcitation",
```
paul0403, line 75: Oh, is this where you tell the frontend to no longer decompose these? :star_struck:

frontend/test/pytest/test_operations.py
```diff
@@ -124,6 +124,12 @@ def circuit(x: float, y: float):
         qml.IsingZZ(x, wires=[0, 1])
         qml.IsingZZ(y, wires=[1, 2])
 
+        qml.SingleExcitation(x, wires=[0, 1])
+        qml.SingleExcitation(y, wires=[1, 2])
+
+        qml.DoubleExcitation(x, wires=[0, 1, 2, 3])
+        qml.DoubleExcitation(y, wires=[2, 3, 0, 1])
```
paul0403, line 131: If the above is true, it would be nice to add a lit test to check that the compiled mlir has `quantum.custom "SingleExcitation"` !

maliasadi: It's a good idea! We may need to add lit tests for other named gates as well

frontend/test/pytest/test_operations.py
```diff
@@ -243,5 +249,37 @@ def circuit():
...
+
+    @qjit
+    @qml.qnode(dev)
+    def circuit(phi):
+        qml.SingleExcitation(phi, wires=[0, 1])
+        return qml.expval(qml.PauliZ(0))
+
+    assert np.allclose(circuit(0.1), circuit.original_function(0.1))
+    assert "SingleExcitation" in str(circuit.jaxpr)
+    assert 'quantum.custom "SingleExcitation"' in circuit.mlir
```
paul0403: So, if the purpose of a test is just checking the compiled mlir, there is no need for it to be a pytest. We can do this with a lit test and save some execution time (since lit tests only need to compile, as opposed to pytest's both compiling and running).

The way to write a lit test is really simple: `FileCheck` is a tool that takes in some text and check it against some expected patterns. For example:

```python
# main.py
print("apple")
print("banana")
```

```
# expected.txt
# A plain text file in whatever format that contains the expected text pattern with `CHECK:`
CHECK: apple
CHECK: banana
CHECK-NOT: orange
```

```
python3 main.py | FIleCheck expected.txt
```
(If you want to play with it, after building catalyst from source `FileCheck` is at `catalyst/mlir/llvm-project/build/bin/FileCheck`)

Now, for the python lit tests, we have them in `frontend/test/lit`. We have it setup so that all you need to do is embed the `CHECK` lines inside the python file itself:
```python
# CHECK: apple
print("apple")
```

I think the best file to add this file to is `test_decomposition.py`, to check that the excitation gates are not decomposed on lightning!

frontend/test/pytest/test_operations.py
paul0403: Agreed, but maybe not in this PR?

paul0403: Or if @josephleekl wants to do it here I'm also not against it : )

josephleekl: I think not in this PR 😅

frontend/test/pytest/test_operations.py
josephleekl: Thanks @paul0403 ! I've now removed these tests and added the Lit test in `test_decomposition.py` https://github.com/PennyLaneAI/catalyst/pull/1980/commits/b4336fb4103c6d3a389904e20a609fe493e3fc9b

paul0403 (approved):
:partying_face:

paul0403:
(There's a frontend test that needs to wait for PL version update in #1952 )

paul0403:
> > ~(There's a frontend test that needs to wait for PL version update in #1952 )~ or maybe not, let me rerun. I don't think lit tests can be flaky but let me rerun...
> 
> Seems to be passing now @paul0403 🥳

Ok so there's a flaky lit test... I'll look at it later : ) :melting_face:

maliasadi (approved):
:raised_hands:

PR #1982, opened 2025-08-14
Remove xfail on a seeded non-flaky test

paul0403:
**Context:**
A test marked xfail with the reason being flaky is not actually flaky. We probably just forgot to remove the xfail?

Remove the xfail mark on the test.

**Benefits:**
One less xfail.

### DIFF
--- a/frontend/test/pytest/test_mid_circuit_measurement.py
+++ b/frontend/test/pytest/test_mid_circuit_measurement.py
@@ -548,9 +548,6 @@ class TestDynamicOneShotIntegration:
         assert result.shape == (shots,)
         assert jnp.allclose(result, expected)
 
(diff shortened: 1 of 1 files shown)


paul0403:

frontend/test/pytest/test_mid_circuit_measurement.py
```diff
@@ -548,9 +548,6 @@ def circuit(x):
         assert result.shape == (shots,)
         assert jnp.allclose(result, expected)
 
-    @pytest.mark.xfail(
```
mlxd, line 551: Thank you.

JerryChen97 (approved):
Thanks!!!

PR #1983, opened 2025-08-14
[Capture] Update `from_plxpr` handling of qnode primitive to adjust to new signature

albi3ro:
Depends on pennylane https://github.com/PennyLaneAI/pennylane/pull/7652 and  https://github.com/PennyLaneAI/pennylane/pull/7944 both.

So we won't be able to proceed with this PR till both of those are merged.  Unfortunately, we will also need to have a very quick turnaround between #7652 and this one, since that change breaks the XDSL tests and will require this PR.  But at least we can wait till #7944 is in before start more extensive testing and integration.

**Context:**

We are adjusting the program capture qnode primitive to support shot vectors and dynamic shots.  This adjusts `from_plxpr` accordingly.

Adjusts the conversion of `qnode_prim` to accomodate the shots being passed positionally.

**Benefits:**

Adds support for dynamic shots.

**Possible Drawbacks:**

The qnode primitive's signature is getting complicated.

Shot vector support not yet enabled.

**Related GitHub Issues:**

[sc-97220]


doc/releases/changelog-dev.md
```diff
@@ -75,6 +75,8 @@
 * When capture is enabled, `qjit(autograph=True)` will use capture autograph instead of catalyst autograph.
   [(#1960)](https://github.com/PennyLaneAI/catalyst/pull/1960)
 
+* `from_plxpr` can now handle dynamic shots and overridden device shots.
```
albi3ro, line 113:
```suggestion
* `from_plxpr` can now handle dynamic shots and overridden device shots.
  [(#1983)](https://github.com/PennyLaneAI/catalyst/pull/1983/)
```

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -715,6 +709,7 @@ def trace_from_pennylane(
             fn.static_argnums = static_argnums
 
         plxpr, out_type, out_treedef = make_jaxpr2(fn, **make_jaxpr_kwargs)(*args, **kwargs)
+        print(plxpr)
```
albi3ro:
```suggestion
```

.github/workflows/check-catalyst.yaml
```diff
@@ -475,6 +475,10 @@ jobs:
         python3 -m pip install oqc-qcaas-client
         make frontend
 
+    - name: Install PennyLane branch
+      run: |
+        pip install --no-deps --force git+https://github.com/PennyLaneAI/pennylane@capture-new-shots
+
```
paul0403: Just so I'm clear, #7652 (the `capture-new-shots` branch) will break xdsl because xdsl needs to go through plxpr, which in turn means CI on #7652 itself will not pass until it has a catalyst that has the qnode prim signature updated.

So the catalyst PR that updates the qnode primitive (this one) need to use #7652 manually. 

This is fine, but I would like to have this PR and #7652 to be merged in the same day if possible, so the nightly wheels are unaffected and still built with main branches!

Also, remember to update the wheel scripts accordingly, otherwise the wheel CI on main will fail. These are `build-wheel-[linux-arm64 | linux-x86 | macos-arm64].yaml`.

frontend/test/pytest/from_plxpr/test_from_plxpr.py
```diff
@@ -173,6 +173,25 @@ def circuit():
...
+
+        qml.capture.enable()
+        try:
+            jaxpr = jax.make_jaxpr(c)()
+
+            with pytest.raises(NotImplementedError, match="not yet supported"):
+                from_plxpr(jaxpr)()
+        finally:
+            qml.capture.disable()
+
```
paul0403: Is there a fixture to disable capture automatically for these kind of supposed-to-fail tests? 

I don't remember all the plxpr test fixtures added :sweat_smile: But it's fine if there isn't one.

frontend/test/pytest/from_plxpr/test_from_plxpr.py
```diff
@@ -559,7 +578,7 @@ def circuit():
             return qml.sample(wires=0)
 
         def f():
-            return circuit(shots=100)
+            return qml.set_shots(circuit, shots=100)()
```
paul0403, line 560: Does this mean the 
```python
        # TODO: try set_shots after capture work is completed
        with pytest.warns(
```
can be deleted? :star_struck:

frontend/test/pytest/test_measurement_dynamic_shapes.py
```diff
@@ -26,6 +26,7 @@
 import catalyst
 
 
+@pytest.mark.usefixtures("use_both_frontend")
```
paul0403, line 29: Can we run both frontends for the counts test in this file as well? Or does that use shotvectors and is still not working entirely now?

frontend/test/pytest/test_measurements_shots_results.py
```diff
@@ -44,8 +44,10 @@ def circuit():
             qml.CNOT(wires=[0, 1])
             return qml.expval(qml.Identity(wires=0)), qml.expval(qml.Identity(wires=1))
 
-        expected = circuit()
         result = qjit(circuit, seed=37)()
+        qml.capture.disable()
+        expected = circuit()
```
paul0403: What are these changes about? Aren't these tests already using the `use_both_frontend` fixture? :thinking:

albi3ro: So this was computing the `expected` using capture execution pipeline, which is no longer maintained. So the capture execution pipeline does not have any integration with the updated shots syntax. Do to that, I need to disable capture to get the expected results.

frontend/test/pytest/test_measurement_dynamic_shapes.py
albi3ro: Counts is not yet supported in program capture. We also do not yet support dynamic numbers of device wires.

.github/workflows/check-catalyst.yaml
albi3ro: So yes, once this PR is approved we will find a quiet time to:

* override merge in pennylane
* push new PL to test pypi
* Update this PR to use the new version
* Merge this
* Upload new catalyst wheels to test pypi

paul0403: Ah I see, so this PR will eventually just use a PL on testpypi. In that case no need to update the wheels scripts to manually pip install the `capture-new-shots` branch :+1:

frontend/test/pytest/test_measurement_dynamic_shapes.py
paul0403: > Counts is not yet supported in program capture.

Ah yes completely forgot about this :sweat_smile: all good then

frontend/test/pytest/test_measurements_shots_results.py
paul0403: > So this was computing the `expected` using capture execution pipeline, which is no longer maintained.

Does this mean these tests should not be using both frontends? And should only be using catalyst's legacy frontend?

albi3ro: The non-qjit execution should not be using capture.

We do need to figure out how to selectively turn on capture just for qjit-execution and then turn it off for normal execution, but that's a problem for later me.

frontend/test/pytest/from_plxpr/test_from_plxpr.py
albi3ro: I guess I was just mimicing the old tests in the same class, but we should probably just use a fixture for all these tests.

frontend/test/pytest/from_plxpr/test_from_plxpr.py
albi3ro: Thanks for catching that. Updating.

frontend/test/pytest/test_measurements_shots_results.py
paul0403: Ahh, so this change has nothing to do with qjit, it's that for these cases PL's execution itself (which is the reference for these tests) cannot be in capture mode now. I see. 

All good then :+1:

paul0403: Can you add a comment saying that PL cannot run `expected` with capture?

frontend/test/pytest/from_plxpr/test_from_plxpr.py
albi3ro: Added a `use_capture` fixture for use here instead.

frontend/test/pytest/conftest.py
```diff
@@ -49,6 +49,15 @@ def disable_capture():
             qml.capture.disable()
 
 
+@pytest.mark.fixture(scope="function")
```
albi3ro:
```suggestion
@pytest.fixture(scope="function")
```

frontend/test/pytest/conftest.py
```diff
@@ -49,6 +49,15 @@ def disable_capture():
...
+def use_capture():
```
albi3ro, line 53:
```suggestion
def use_capture():
    """Enable capture before and disable capture after the test."""
```

PR #1986, opened 2025-08-18
Pin mac wheels CI runner to MacOS 14

paul0403:
**Context:**
Github recently updated MacOS in CI images to 15.5. This macos works with xcode 16.4.
https://github.com/actions/runner-images/issues/12541
https://github.com/actions/runner-images/issues/12520

New xcode registers the sdk file a bit differently, which broke the macos/xcode version parsing in lapacke.

Pin macos version in mac wheels CI to `macos-14`

**Benefits:**
Mac wheels no longer fail.

**Possible Drawbacks:**
xcode 16 seems painfully slow. Forcing single threaded build and disabling macos spotlight/OS indexing fixes this.

### DIFF
--- a/.github/workflows/build-wheel-macos-arm64.yaml
+++ b/.github/workflows/build-wheel-macos-arm64.yaml
@@ -387,7 +387,9 @@ jobs:
         sudo xcode-select --switch /Library/Developer/CommandLineTools
         cmake -S mlir -B $GITHUB_WORKSPACE/quantum-build -G Ninja \
               -DCMAKE_BUILD_TYPE=Release \
+              -DCMAKE_OSX_SYSROOT=$(xcrun --show-sdk-path) \
               -DLLVM_ENABLE_ASSERTIONS=ON \
+              -DLLVM_USE_SANITIZER=OFF \
               -DQUANTUM_ENABLE_BINDINGS_PYTHON=ON \
               -DPython3_EXECUTABLE=$(which python${{ matrix.python_version }}) \
               -DPython3_NumPy_INCLUDE_DIRS=$(python${{ matrix.python_version }} -c "import numpy as np; print(np.get_include())") \
--- a/.github/workflows/build-wheel-macos-arm64.yaml
+++ b/.github/workflows/build-wheel-macos-arm64.yaml
@@ -367,8 +367,8 @@ jobs:
     # Build Quantum and Gradient Dialects
     - name: Build MLIR Dialects
       run: |
-        # TODO: turn off spotlight when upgrading to macos 15 and xcode 16 in github mac runners
-        # sudo mdutil -i off /Users
+        # xcode 16 builds very slow if we don't turn off spotlight
+        sudo mdutil -i off /Users
         cmake -S mlir -B $GITHUB_WORKSPACE/quantum-build -G Ninja \
               -DCMAKE_BUILD_TYPE=Release \
               -DLLVM_ENABLE_ASSERTIONS=ON \
--- a/.github/workflows/build-wheel-macos-arm64.yaml
+++ b/.github/workflows/build-wheel-macos-arm64.yaml
@@ -400,7 +400,10 @@ jobs:
               -DLLVM_ENABLE_LLD=OFF \
               -DLLVM_DIR="$GITHUB_WORKSPACE/llvm-build/lib/cmake/llvm"
 
-        cmake --build $GITHUB_WORKSPACE/quantum-build --target check-dialects catalyst-cli
+        # xcode >= 16 turns on verifier after building every translation unit
+        # This will verify all llvm builds and take a huge amount of time
+        # To turn it off, we need to set ENABLE_MODULE_VERIFIER
+        cmake --build $GITHUB_WORKSPACE/quantum-build --target check-dialects catalyst-cli -- -sdk macosx ENABLE_MODULE_VERIFIER=NO
 
     - name: Build Plugin wheel
       # Run only on Thursday at the given time
--- a/.github/workflows/build-wheel-macos-arm64.yaml
+++ b/.github/workflows/build-wheel-macos-arm64.yaml
@@ -384,6 +384,7 @@ jobs:
     # Build Quantum and Gradient Dialects
     - name: Build MLIR Dialects
       run: |
+        sudo mdutil -i off /Users
         sudo xcode-select --switch /Library/Developer/CommandLineTools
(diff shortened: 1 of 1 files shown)


.github/workflows/build-wheel-macos-arm64.yaml
```diff
@@ -325,12 +325,33 @@ jobs:
         key: ${{ runner.os }}-${{ runner.arch }}-enzyme-${{ needs.constants.outputs.llvm_version }}-${{ needs.constants.outputs.enzyme_version }}-wheel-build
         fail-on-cache-miss: True
 
+    - name: Setup Xcode
+      uses: maxim-lobanov/setup-xcode@v1
```
dime10: Just want to mention that we should be careful with third-party actions. If it isn't strictly necessary it's usually better to avoid them

paul0403: We were also wondering what this is, but apparently this is a very popular `setup-xcode` action step https://grep.app/search?q=setup-xcode

We are trying to find alternatives.

rniczh: I also have a same concerns about this, and found a previous discussion about this thing. This guy actually is belong to GitHub Inc. btw.
https://github.com/orgs/community/discussions/160490

mlxd: When in doubt, please ask in the CI channel for an eval.

.github/workflows/build-wheel-macos-arm64.yaml
```diff
@@ -37,7 +37,7 @@ on:
         type: string
 
 env:
-  MACOSX_DEPLOYMENT_TARGET: 14.0
+  MACOSX_DEPLOYMENT_TARGET: 15.0
```
mlxd: Just a quick fly-by note, we should make sure we don't update the target support beyond what we want for user systems when shipping wheels. We aim to still support MacOS 14 (and ideally 13).

rniczh:
For CI, the build doesn't stuck during linking stage when using only a single thread (`-- -j1` with `cmake --build`). Disabling indexing (`mdutil -i off ...`) further accelerates the process.

Py3.12 Build Wheels:
1 thread: 16m 8s
1 thread + disable indexing: 8m 23s

.github/workflows/build-wheel-macos-arm64.yaml
paul0403: So this change is because of the following:

- CI mac runner images have been upgraded to MacOS 15 upstream
- This means the wheels will be built on 15
- If the deployment target stays on 14, all programs using the built-on-15 wheels will emit a warning saying (paraphrasing) "Program being run on 14 but the package is built on 15"
- This breaks some pytests that check for an empty output stream.

I'm not sure how to proceed, as the macos image upgrade upstream means we have to build wheels on 15? Or can we manually downgrade the images back to 14 somehow?

mlxd: https://github.com/actions/runner-images/blob/main/images/macos/macos-13-arm64-Readme.md Can we not defer to this?

paul0403: Ahh, I see, so we can just fallback to the `macos-14` label. 

https://github.com/actions/runner-images/issues/12520

> Switch back to macOS-14 by specifying the macos-14 label. We will continue to support macOS-14 for some time.


What do you think? We can stick with 14 (so status quo) until the `macos-14` label goes away (which will be around for "some time")

mlxd: Currently chatting with @rashidnhm about using the 13 version.

.github/workflows/build-wheel-macos-arm64.yaml
```diff
@@ -325,12 +325,22 @@ jobs:
...
         fail-on-cache-miss: True
 
+    - name: Setup Xcode version
+      #uses: maxim-lobanov/setup-xcode@v1
+      #with:
+        # On GH images, MacOS is current 15.5
+        # This macos works with xcode 16.4
+        # xcode-version: '16.4.0'
+      run:
+        sudo xcode-select -s /Applications/Xcode_16.4.app
```
rniczh: I think this line could just replaced by L343

.github/workflows/build-wheel-macos-arm64.yaml
```diff
@@ -325,12 +325,22 @@ jobs:
...
+
     # Build Catalyst-Runtime
     - name: Build Catalyst-Runtime
       id: runtime-build
       run: |
         # On GH images, gfortran is only available as a specific version.
         export FC=gfortran-14
+        sudo xcode-select --switch /Library/Developer/CommandLineTools
```
rniczh: removed after L335 updated to `sudo xcode-select --switch /Library/Developer/CommandLineTools`

.github/workflows/build-wheel-macos-arm64.yaml
```diff
@@ -363,6 +373,8 @@ jobs:
     # Build Quantum and Gradient Dialects
     - name: Build MLIR Dialects
       run: |
+        sudo mdutil -i off /Users
+        sudo xcode-select --switch /Library/Developer/CommandLineTools
```
rniczh: removed after L335 updated to `sudo xcode-select --switch /Library/Developer/CommandLineTools`

.github/workflows/build-wheel-macos-arm64.yaml
paul0403: I thought so, but my concern is how does `/Library/Developer/CommandLineTools` know to use 16.4?

rniczh: By default, it points to the developer tools associated with the version of macOS, `xcode-select --show-sdk-path` could print what the path it currently use. But unfortunately, you can't change it to another version when switching to `/Library/Developer/CommandLineTools`. It does not guarantee that it will use the specific version of Xcode (e.g., Xcode 16.4).

paul0403: I changed to command line tools in setup.

.github/workflows/build-wheel-macos-arm64.yaml
paul0403: No longer using maxim-lobanov now.

.github/workflows/build-wheel-macos-arm64.yaml
paul0403: @mlxd I will be pinning to `macos-14` now to unblock things. Let me know when the discussion on using 13 has progressed.

PR #1987, opened 2025-08-19
Add end-to-end PCPhase support

maliasadi:
**Context:**
LQ and LGPU provide native support for PCPhase. To enable this in Catalyst, we had to define a new custom MLIR gate op in the Quantum dialect and extend both the frontend and runtime to lower PCPhase into a new QIR instruction. This allows execution on devices with native support and unlocks simulation of QSVT workloads w/ ~30 qubits on a single GPU.

**Benefits:**
- End-to-end PCPhase support
- Simulate QSVT workflows with ~30 qubits backed by LGPU

**Possible Drawbacks:**
- We need a better representation in Catalyst to support generic PL gates with hyperparameters.

**Related GitHub Issues:**
[sc-98149]


benchmark/batchrun.py
```diff
@@ -14,15 +14,7 @@
 """Measurement cycle management entry point"""
 import sys
 
-from catalyst_benchmark.toplevel import (
```
maliasadi, line 17: `make format` updated these headers

runtime/tests/Test_NullQubit.cpp
```diff
@@ -353,7 +353,7 @@ TEST_CASE("Test NullQubit state vector after AllocateQubits", "[NullQubit]")
 
     CHECK(sim->AllocateQubits(0).size() == 0);
 
-    auto &&q = sim->AllocateQubits(2);
+    sim->AllocateQubits(2);
```
maliasadi, line 356: It's an unused variable. `make runtime-test` fails without this change.

mlir/include/Quantum/IR/QuantumOps.td
```diff
@@ -637,10 +637,60 @@ def MultiRZOp : UnitaryGate_Op<"multirz", [DifferentiableGate, NoMemoryEffect,
             return getODSOperands(getParamOperandIdx());
         }
     }];
-    
+
+    let hasCanonicalizeMethod = 1;
+}
+
+def PCPhaseOp : UnitaryGate_Op<"pcphase", [DifferentiableGate, NoMemoryEffect,
```
paul0403, line 644: Hey, I think this looks good and is relatively straightforward, just a question: why did you decide to add it as its own new op, instead of subsuming it under custom ops?

frontend/catalyst/device/qjit_device.py
```diff
@@ -76,6 +76,7 @@
     "PauliX",
     "PauliY",
     "PauliZ",
+    "PCPhase",
```
paul0403, line 81: Remember to add a changelog!

runtime/lib/capi/RuntimeCAPI.cpp
```diff
@@ -504,6 +504,29 @@ void __catalyst__qis__SetState(MemRefT_CplxT_double_1d *data, uint64_t numQubits
...
+void __catalyst__qis__PCPhase(double theta, double dim, const Modifiers *modifiers,
+                              int64_t numQubits, ...)
+{
+    RT_ASSERT(numQubits >= 0);
+
+    // Compute the abs value to support adjoint
+    // Note this will be revisited once we have a better solution
+    // for capturing generic PL gates. Currently, this is the optimal
+    // workaround to reduce type casting throughout the pipeline.
+    dim = dim < 0 ? -dim : dim;
```
paul0403: Ah, is this why PCPhase needs to be its own op? Because its runtime needs its specific preprocessing before going into the `NamedOperation`?

(Also, why can the dimension be a negative float? :sweat_smile: )

mlir/include/Quantum/IR/QuantumOps.td
maliasadi: Mainly because of differences in the signature of `PCPhase`; this op works with variadic number of wires and has the `dim` hyperparameter. I followed the same decision made for `GlobalPhase` and `MultiRZ`, and implement it as a new op instead of trying to update the legacy custom op.

runtime/lib/capi/RuntimeCAPI.cpp
maliasadi: Yeah, `NamedOperation` in Catalyst Runtime and `ApplyOperation` in Lightning C++ APIs don't accept hyperparameters  as the rsult we need to serialize them as regular params before calling the gate kernels. Here for PCPhase, negative dim is not accepted by [PL](https://docs.pennylane.ai/en/stable/code/api/pennylane.PCPhase.html) and Lightning. When applying adjiont to the op, the list of params will be multiplied by negative one, this little patch here is needed to fix this.

I wanted to ensure that we have `dim` in the assembly of PCPhase while reducing the number of type casts to one place so that we can fix this later in future when we have a better support for Operator's hyperparamters in Catalyst.

doc/releases/changelog-dev.md
```diff
@@ -76,6 +76,9 @@
 
 <h3>Internal changes ⚙️</h3>
 
+* Add `PCPhase` native compiler and runtime operator support.
+  [(#1987)](https://github.com/PennyLaneAI/catalyst/pull/1987)
+
 * Add `SingleExcitation` and `DoubleExcitation` to native `RUNTIME_OPERATIONS` and `RuntimeCAPI`.
   [(#1980)](https://github.com/PennyLaneAI/catalyst/pull/1980)
```
dime10: Aren't these user-facing changes? They can now use these gates without decomposition in their programs. We should also mention that this significantly speeds up simulation runtime in the changelog (the "so what").

mlir/include/Quantum/IR/QuantumOps.td
dime10: I think @rniczh recently unified/simplified the lowering of variadic qubit ops?

mlir/include/Quantum/IR/QuantumOps.td
```diff
@@ -637,10 +637,60 @@ def MultiRZOp : UnitaryGate_Op<"multirz", [DifferentiableGate, NoMemoryEffect,
             return getODSOperands(getParamOperandIdx());
         }
     }];
-    
+
+    let hasCanonicalizeMethod = 1;
```
paul0403, line 641: What is this canonicalization method on multirz about?

mlir/include/Quantum/IR/QuantumOps.td
paul0403: > I think @rniczh recently unified/simplified the lowering of variadic qubit ops?

Only in the custom op lowering patterns, i.e. only in the sense of `quantum.custom "gate_with_variadic_inqubits"` can be lowered to `__catalyst__qis__gate_with_variadic_inqubits(...)` through the `CustomOpPattern` https://github.com/PennyLaneAI/catalyst/blob/main/mlir/lib/Quantum/Transforms/ConversionPatterns.cpp#L443. The underlying `__catalyst__qis__pcphase` still needs to exist. 

Hmm, actually now as I am typing this, I realized maybe we can go through custom op for pcphase as well? i.e.

1. During binding `qinst_p`, pcphase still has the special treatement for `dim`:
```python
jax_tracer.py
            params = (
                op.parameters + [op.hyperparameters["dimension"][0]]
                if isinstance(op, qml.PCPhase)
                else op.parameters
            )
            qinst_p.bind(...params...)
``` 

2. During lowering jaxpr to mlir, just treat `dim` as another param in `float_params` on custom op
```python
jax_primitives.py
        float_param = float_params[0]
        dim_param = float_params[1]
        return CustomOp(
            out_qubits=[qubit.type for qubit in qubits],
            out_ctrl_qubits=[qubit.type for qubit in ctrl_qubits],
            theta=float_param,
            dim=dim_param,
            in_qubits=qubits,
            in_ctrl_qubits=ctrl_qubits,
            in_ctrl_values=ctrl_values_i1,
            adjoint=adjoint,
        ).results
```

3. During custom op lowering to runtime, add pcphase as a variadic op https://github.com/PennyLaneAI/catalyst/blob/main/mlir/lib/Quantum/Transforms/ConversionPatterns.cpp#L443
```
Quantum/ConversionPatterns.cpp
      const auto hasVariadicInputQbits = (op.getGateName() == "Identity") || (op.getGateName() == "PCPhase"));
```

and this would lower to your new `__catalyst__qis__PCPhase` runtime stub.

@maliasadi so I think going through custom op is still possible. But I don't really have a preference either way, so I'm fine with whichever method is easier for you. @dime10 David do you have a preference here?

dime10: It's already a separate op, so I don't mind tbh (using custom is also fine though). Who knows maybe at some point we will have a separate operation for each known gate 😅 

However, make sure all new operations implement all applicable quantum interfaces!

doc/releases/changelog-dev.md
maliasadi: yeah, I was thinking the same. I'm benchmarking QSVT w/ this PR for the presentation. I'll update the changelog adding more details about the "so what" :smile:

mlir/include/Quantum/IR/QuantumOps.td
maliasadi: > Who knows maybe at some point we will have a separate operation for each known gate

I’m kinda leaning toward this idea, honestly

maliasadi: @paul0403 @dime10 Let me know what's the preferred approach so that I can finalize and merge this one today or tomorrow.

dime10: Up to you, as long as this is respected:
> However, make sure all new operations implement all applicable quantum interfaces!

dime10: I you don't care I would just say use the new op 👍

mlir/lib/Quantum/IR/QuantumOps.cpp
```diff
@@ -84,6 +84,21 @@ LogicalResult MultiRZOp::canonicalize(MultiRZOp op, mlir::PatternRewriter &rewri
     return failure();
 }
 
+LogicalResult PCPhaseOp::canonicalize(PCPhaseOp op, mlir::PatternRewriter &rewriter)
+{
+    if (op.getAdjoint()) {
+        auto paramNeg = rewriter.create<mlir::arith::NegFOp>(op.getLoc(), op.getTheta());
+        auto paramDim = rewriter.create<mlir::arith::NegFOp>(op.getLoc(), op.getDim());
```
paul0403: Maybe we can simply not negate the dimension here if adjoint is present? This would avoid the need to absolute value the dim in the runtime stub right?

mlir/lib/Quantum/Transforms/merge_rotation.cpp
```diff
@@ -53,6 +53,8 @@ struct MergeRotationsPass : impl::MergeRotationsPassBase<MergeRotationsPass> {
                                                                  &getContext());
         catalyst::quantum::MultiRZOp::getCanonicalizationPatterns(patternsCanonicalization,
                                                                   &getContext());
+        catalyst::quantum::PCPhaseOp::getCanonicalizationPatterns(patternsCanonicalization,
```
paul0403: Do we care about PCPhase in merge-rotation pass? I feel like that would require some extra work (i.e. check the merged pcphase dimensions are the same, etc.). I propose we leave PCPhase's merge rotation as a separate issue (which would also be a good first issue for assessment/onboarding/...)

mlir/test/Quantum/ConversionTest.mlir
```diff
@@ -331,6 +331,35 @@ func.func @multirz(%q0 : !quantum.bit, %p : f64) -> (!quantum.bit, !quantum.bit,
...
+
+    // CHECK: [[d:%.+]] = llvm.mlir.zero : !llvm.ptr
+    // CHECK: [[c1:%.+]] = llvm.mlir.constant(1 : i64)
+    // CHECK: llvm.call @__catalyst__qis__PCPhase(%arg1, %arg2, [[d]], [[c1]], %arg0)
+    %q1 = quantum.pcphase(%p, %d) %q0 : !quantum.bit
+
+    // CHECK: [[d:%.+]] = llvm.mlir.zero : !llvm.ptr
+    // CHECK: [[c2:%.+]] = llvm.mlir.constant(2 : i64)
+    // CHECK: llvm.call @__catalyst__qis__PCPhase(%arg1, %arg2, [[d]], [[c2]], %arg0, %arg0)
+    %q2:2 = quantum.pcphase(%p, %d) %q1, %q1 : !quantum.bit, !quantum.bit
```
paul0403, line 347: Why did the second pcphase take in `%arg0` again? Shouldn't it take in `%q1`, the out qubits of the first pcphase? :scream: :thinking:

paul0403: Ah this is runtime IR now, so they both apply to the same qubit that has been getelemntptr-ed. My bad!

runtime/lib/backend/common/Utils.hpp
```diff
@@ -220,6 +220,7 @@ enum class SimulatorGate : uint8_t {
     Toffoli,
     // n-qubit
     MultiRZ,
+    PCPhase,
```
paul0403: What are the two lists of gates in `Utils.hpp` for? I did not see them for Joseph's single/double excitation gates :thinking:

runtime/lib/capi/RuntimeCAPI.cpp
```diff
@@ -504,6 +504,29 @@ void __catalyst__qis__SetState(MemRefT_CplxT_double_1d *data, uint64_t numQubits
     getQuantumDevicePtr()->SetState(data_view, wires);
 }
 
+void __catalyst__qis__PCPhase(double theta, double dim, const Modifiers *modifiers,
+                              int64_t numQubits, ...)
+{
+    RT_ASSERT(numQubits >= 0);
+
```
paul0403: Do we want to assert that `dim` here is a integer? e.g. floor(dim) == dim or something?

runtime/tests/Test_NullQubit.cpp
```diff
@@ -353,7 +353,7 @@ TEST_CASE("Test NullQubit state vector after AllocateQubits", "[NullQubit]")
...
 
     CHECK(sim->GetNumQubits() == 2);
```
paul0403, line 358: Maybe we can add a runtime test with null qubit :wink:

mlir/include/Quantum/IR/QuantumOps.td
maliasadi: To negate the param when adjoint is provided.

mlir/lib/Quantum/IR/QuantumOps.cpp
maliasadi: hahaha... good catch! :smile:

mlir/lib/Quantum/Transforms/merge_rotation.cpp
maliasadi: PCPhase is a non-global version of GlobalPhase w/ controls. We should be able to write a merge rotation transform for it. But I agree this takes a bit of time. I removed the draft impl but forgot this one. Removed it [ad9b147](https://github.com/PennyLaneAI/catalyst/pull/1987/commits/ad9b14740d5f2e019f26488b837750d562e9f334)

runtime/lib/backend/common/Utils.hpp
maliasadi: Akh! old habit :smile: we used to use this in all the device classes. This is still being used in `OpenQasmDevice`. I never liked these lists btw! I implemented this fast and static `gate_lookup` in the first version of the runtime but never revisited it since :melting_face:

runtime/tests/Test_NullQubit.cpp
maliasadi: Although we don't tests all these gates.. it's a good idea! Added [ad9b147](https://github.com/PennyLaneAI/catalyst/pull/1987/commits/ad9b14740d5f2e019f26488b837750d562e9f334)

runtime/lib/capi/RuntimeCAPI.cpp
maliasadi: Yeah, this can be fixed now after updating the adjoint in canonicalization [ad9b147](https://github.com/PennyLaneAI/catalyst/pull/1987/commits/ad9b14740d5f2e019f26488b837750d562e9f334)

doc/releases/changelog-dev.md
maliasadi: Updated [10a0616](https://github.com/PennyLaneAI/catalyst/pull/1987/commits/10a06168edab2f1425e3f952823cc31c74806327)

maliasadi: benchmark script is still running. I'll share the results later

mlir/include/Quantum/IR/QuantumOps.td
paul0403: Ah, so multirz always had a canonicalization pattern but its `let hasCanonicalizeMethod = 1;` was never set?

maliasadi: hmm... it's more like a GH rendering issue :smile: I've added this for PCPhase

paul0403: :exploding_head: I see it now

PR #1996, opened 2025-08-21
The runtime `qubit_release_array` CAPI should release only the qubits in the array, not all qubits on the device

paul0403:
**Context:**
The runtime `qubit_release_array` CAPI should release only the qubits in the array, not all qubits on the device.
This is especially important for dynamic qubit allocation work. When lowering `qml.deallocate`, the current `quantum.dealloc` op will be lowered to the `qubit_release_array` to release qubits in the register, but because the dynamically allocated qregisters are only temporary in the program, we cannot just release all qubits on the device.

We were ok with this because there was always just one global qreg. But now with dynamic allocation the asymmetry between batched allocation and (the currently non-existing) batched release needs to be properly handled.

- Add `ReleaseQubits(std::vector<QubitIdType>)` interface in `QuantumDevice`, and remove `ReleaseAllQubits()`;
- `qubit_release_array` CAPI uses `ReleaseQubits(std::vector<QubitIdType>)` instead of `ReleaseAllQubits()`.
- Add guards for current devices that don't support dynamic allocation:oqc, oqd, openqasm to check that there is only one `ReleaseQubits()` call in all their programs.

**Benefits:**
- Symmetry between qubit allocation and deallocation in device api!
- Better support implementation of dynamic qubit allocation.

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -318,6 +318,9 @@
 * The `NoMemoryEffect` trait has been removed from the `quantum.alloc` operation.
   [(#2044)](https://github.com/PennyLaneAI/catalyst/pull/2044)
 
+* Enhance `ppm_specs` function to prevent duplicate pass addition
+  [(#2049)](https://github.com/PennyLaneAI/catalyst/pull/2049)
+
 <h3>Documentation 📝</h3>
 
 * The Catalyst Command Line Interface documentation incorrectly stated that the `catalyst`
--- a/frontend/catalyst/passes/builtin_passes.py
+++ b/frontend/catalyst/passes/builtin_passes.py
@@ -1060,7 +1060,9 @@ def ppm_specs(fn):
 
         # add ppm-spec pass at the end to existing pipeline
         _, pass_list = new_options.pipelines[0]  # first pipeline runs the user passes
-        pass_list.append("ppm-specs")
+        # check if ppm-specs is already in the pass list
+        if "ppm-specs" not in pass_list:  # pragma: nocover
+            pass_list.append("ppm-specs")
 
         new_options = _options_to_cli_flags(new_options)
         raw_result = _quantum_opt(*new_options, [], stdin=str(fn.mlir_module))
@@ -1072,7 +1074,7 @@ def ppm_specs(fn):
         except Exception as e:  # pragma: nocover
             raise CompileError(
                 "Invalid json format encountered in ppm_specs. "
-                f" but got {raw_result[: raw_result.index('module')]}"
+                f"Expected valid JSON but got {raw_result[: raw_result.index('module')]}"
             ) from e
 
     else:


dime10:
I don't think releasing them one-by-one should be forced on devices, rather we should make this change in conjunction with the device api to allow a batched release. It's okay though if a specific device implements it as a one-by-one release.

paul0403:
> I don't think releasing them one-by-one should be forced on devices, rather we should make this change in conjunction with changes in the device api to allow a batched release. It's okay though if a specific device implements it as a one-by-one release.
> 
> Things to consider:
> 
> * performance
> * existing documented semantics in the device api
> * what to do with the `release_all` function, probably obsolete now?
> * this is most likely a breaking change for device plugins

I agree this should be thought about more. Specifically I think performance on lightning.qubit will take a big hit if it resizes the state vector one at a time instead of just blanketly deleting the state vector.

runtime/include/QuantumDevice.hpp
```diff
@@ -100,6 +100,28 @@ struct QuantumDevice {
      */
     virtual void ReleaseAllQubits() = 0;
 
+    /**
+     * @brief (Optional) Release an array of qubits.
```
dime10: This cannot be optional, the program has to be able to deallocate the initial allocation. What is optional is supporting multiple calls to this function, or supporting a partial deallocation (where not all active ids are provided).

runtime/include/QuantumDevice.hpp
```diff
@@ -100,6 +100,28 @@ struct QuantumDevice {
...
+     *
+     * Release the qubits in the provided array. This method is only needed for devices with
+     * dynamic qubit allocation (on devices with static allocation there would only be one matching
+     * `AllocateQubits` - `ReleaseAllQubits` pair).
+     *
+     * Note that the interface does not require the qubit to be in a particular state. The behaviour
+     * for releasing an entangled qubit is left up to the device, and could include:
+     *  - raising a runtime error (assuming the device can detect impure states)
+     *  - continuing execution with an entangled but inaccessible qubit
+     *  - resetting the qubit with or without measuring its state
```
dime10: Typically I don't duplicate information like this, but point to relevant operation that already discusses it

runtime/lib/capi/RuntimeCAPI.cpp
```diff
@@ -390,9 +390,9 @@ void __catalyst__rt__qubit_release(QUBIT *qubit)
 
 static int __catalyst__rt__qubit_release_array__impl(QirArray *qubit_array)
 {
-    getQuantumDevicePtr()->ReleaseAllQubits();
     std::vector<QubitIdType> *qubit_array_ptr =
         reinterpret_cast<std::vector<QubitIdType> *>(qubit_array);
+    getQuantumDevicePtr()->ReleaseQubits(*qubit_array_ptr);
     delete qubit_array_ptr;
```
dime10, line 396: This is fairly minor, but here since we immediately delete the array after the call, we don't _have_ to use `const` for the parameter in the signature, although doing so would allow us more flexibility in the future 🤔 
```
void ReleaseQubits(const std::vector<QubitIdType> &qubits)
```

runtime/include/QuantumDevice.hpp
paul0403: This is just that I haven't removed `ReleaseAllQubits` yet, will update : )

(Same for the comment below)

runtime/lib/capi/RuntimeCAPI.cpp
paul0403: done

runtime/lib/backend/null_qubit/NullQubit.hpp
```diff
@@ -140,13 +140,14 @@ struct NullQubit final : public Catalyst::Runtime::QuantumDevice {
...
-    void ReleaseAllQubits()
+    void ReleaseQubits(std::vector<QubitIdType> &qubits)
     {
-        num_qubits_ = 0;
-        this->qubit_manager.ReleaseAll();
-        if (this->track_resources_) {
+        for (auto q : qubits) {
+            this->ReleaseQubit(q);
+        }
+        if ((num_qubits_ == 0) && (this->track_resources_)) {
```
paul0403, line 150: Thoughts on moving resource tracking output filewrite to destructor?

doc/dev/custom_devices.rst
```diff
@@ -35,7 +35,7 @@ have been stubbed out. For a complete description of the interface check out the
                 device_nqubits = num_qubits;
                 return ids
             }
-            void ReleaseAllQubits() override {}
+            void ReleaseQubits() override {}
```
dime10:
```suggestion
            void ReleaseQubits(std::vector<QubitIdType> &qubits) override {}
```

frontend/catalyst/third_party/oqc/src/OQCDevice.hpp
```diff
@@ -53,6 +54,7 @@ class OQCDevice final : public Catalyst::Runtime::QuantumDevice {
   public:
     explicit OQCDevice(const std::string &kwargs = "{device_type : oqc, backend : default}")
     {
+        only_one_pair_of_allocation = true;
```
dime10: Should this not be set on allocation?

dime10: Or rather, shouldn't the allocation function raise an error rather than the dealloc function when called more than once?  A little confused by the use of this flag

runtime/include/QuantumDevice.hpp
```diff
@@ -88,17 +88,22 @@ struct QuantumDevice {
...
-     * many allocation calls have been run (although on devices with static allocation there would
-     * only be one matching `AllocateQubits` - `ReleaseAllQubits` pair).
-     * After executing this call, a device should be in a state ready to allocate new qubits and
-     * execute another quantum program. Configurable execution parameters, whether explicit in
-     * the interface (like shots) or internal via the constructor kwargs, should be maintained
-     * in the same state as when entering this method.
+     * @brief Release an array of qubits.
+     *
+     * Devices without dynamic allocation support must ensure that there will only be
+     * exactly one matching `AllocateQubits` - `ReleaseQubits` pair in their accepted programs.
```
dime10: Devices can't enforce anything on programs, they can only refuse to work:
```suggestion
     * For devices without dynamic allocation support it is expected that this function
     * only succeed if the ID array contains the same values as those produced by the
     * initial `AllocateQubits` call, otherwise the device is encouraged to raise an error.
```

runtime/lib/backend/null_qubit/NullQubit.hpp
dime10: This was intentional I think to handle multiple executions on the same device, otherwise it would accumulate across multiple qnode runs.

dime10 (dismissed):

runtime/lib/capi/RuntimeCAPI.cpp
dime10: It was more a question than request 😅

runtime/lib/backend/null_qubit/NullQubit.hpp
paul0403: > it would accumulate across multiple qnode runs.

Because we recycle from the device pool in runtime right? Then it makes sense :+1: 

Then I think adding this check for `num_qubits==0` is needed. I think null qubit should (1) support dynamic allocation, and (2) report resource tracking results only on the final dealloc, aka when the program finishes. To achieve this we must have this check.

dime10: I don't see anything wrong with the check

runtime/lib/capi/RuntimeCAPI.cpp
paul0403: Asked @mlxd about this, if the semantics of the API function itself is read-only, regardless of how it is used, then the argument should have `const`.

dime10: I see, the advantage of the API not being read-only is that it allows the implementation to modify the array without copying it first, but I'm also happy with const if that is preferred.

doc/dev/custom_devices.rst
paul0403: Good catch :+1:

frontend/catalyst/third_party/oqc/src/OQCDevice.hpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/1996/commits/7cc099f006eca8d6df96aea415840abafac3c5eb I updated the guards to better follow the updated documentation:

     * For devices without dynamic allocation support it is expected that this function
     * only succeed if the ID array contains the same values as those produced by the
     * initial `AllocateQubits` call, otherwise the device is encouraged to raise an error.

paul0403:
> Do you want to add one more test for this?
> 
> For an example:
> 
> ```c++
> TEST_CASE("Test ReleaseQubits", "[NullQubit]")
> {
>     std::unique_ptr<NullQubit> sim = std::make_unique<NullQubit>();
>     auto qubits = sim->AllocateQubits(3);
>     CHECK(sim->GetNumQubits() == 3);
>     std::vector<QubitIdType> qubits_to_release = {qubits[0], qubits[2]};
>     sim->ReleaseQubits(qubits_to_release);
>     CHECK(sim->GetNumQubits() == 1);
> }
> ```

Nice catch! I definitely planned to add this test but then realized this test was blocked by the lightning PR https://github.com/PennyLaneAI/pennylane-lightning/pull/1254#discussion_r2342286640

For context, lightning internally maintains a statevector. Upon partial deallocation, lightning doesn't actually resize the statevector for performance reasons. This means the lightning devices' "get num qubits" methods do not decrease in size when partial release is hit.

So I must have added this test to my mental to-do list... and forgot about it haha

But you are also right that this can be just a test on null qubit. I will add it. Thanks!

PR #2000, opened 2025-08-25
Support array-backed registers

joeycarter:
**Context:** Currently, the `quantum.insert` op in Catalyst's quantum dialect is effectively a no-op once it has been lowered to the LLVM dialect (and therefore also in the runtime). For example, in the following program,

```mlir
%out_qubits = quantum.custom "<gate>"() %1 : !quantum.bit
%2 = quantum.insert %0[ 0], %out_qubits : !quantum.reg, !quantum.bit
```

the `insert` op does not actually "insert" the qubit represented by the SSA value `%out_qubits` into the register (represented by value `%0`) at position 0 at runtime. In other words, it does not manipulate the array of qubit IDs represented by the `quantum.reg` type and update the qubit ID at position 0 in the array. Instead, it simply replaces the uses of the register returned by the `quantum.insert` op with the register given as its input operand, as can be seen in the source code here:

https://github.com/PennyLaneAI/catalyst/blob/38064f849a17653ff176ce0a9d0478759a1d0222/mlir/lib/Quantum/Transforms/ConversionPatterns.cpp#L415-L421

It would be useful to have the ability to insert individually allocated qubits, or qubits from an auxiliary register, into a specific position in a given register (assuming the qubits at those locations have already been individually deallocated). Indeed, this is an essential component of nearly all MBQC workloads. We say that compilers with this ability support *array-backed registers*.

**Description of the Change:** This PR contains two major additions:

1. The addition of a new runtime stub
    ```cpp
    void __catalyst__rt__array_update_element_1d(QirArray *ptr, int64_t idx, QUBIT *qubit)
    ```
    which updates the value in the `QirArray` at position `idx` to be the value of `qubit`.
2. The addition of the `InsertOpArrayBackedPattern` conversion pattern that lowers `quantum.insert` ops to a call to this runtime stub during conversion to the LLVM dialect.

The original `InsertOpPattern` conversion pattern has been renamed to `InsertOpDefaultPattern`. **By default, Catalyst workloads will continue to use the original conversion pattern.** A `use-array-backed-registers` option has been added to the `convert-quantum-to-llvm` pass to opt in to use the `InsertOpArrayBackedPattern` conversion pattern instead.

The MBQC pipeline, available via `catalyst.ftqc.mbqc_pipeline()`, has also been updated to enable array-backed registers.

For example, this option can be enabled from the command line with,

```console
$ quantum-opt --pass-pipeline="builtin.module(convert-quantum-to-llvm{use-array-backed-registers=true})"
```

**Benefits:** Array-backed registers are essential to represent MBQC workloads. This PR will enable ongoing and future FTQC development.

**Possible Drawbacks:** There is a risk in the runtime stub that we overwrite the ID of an active qubit contained in the register (one that has not been deallocated). Ideally we should be able to query whether a qubit is active given its ID, but we do not have a general, device-agnostic way to do this yet. For NullQubit, for example, we can query its `qubit_manager` object to determine if a given qubit is active, but not all devices implement a qubit manager.

[sc-97937]


runtime/include/RuntimeCAPI.h
```diff
@@ -35,6 +35,7 @@ void __catalyst__rt__print_string(char *);
 void __catalyst__rt__assert_bool(bool, char *);
 int64_t __catalyst__rt__array_get_size_1d(QirArray *);
 int8_t *__catalyst__rt__array_get_element_ptr_1d(QirArray *, int64_t);
+void __catalyst__rt__insert_element_into_array_1d(QirArray *, int64_t, QUBIT *);
```
multiphaseCFD: Do we need to add this to a pass in the MLIR layer?

paul0403:
Discussed with Joey offline, this PR will only deal with the IR lowerings of the insert op. 

The handling of qubit reuse and/or keeping track of active and inactive qubits will be done by me in the qubit allocation work.

runtime/include/RuntimeCAPI.h
joeycarter: Yes, see the TODO item in the PR description :slightly_smiling_face:

runtime/lib/capi/RuntimeCAPI.cpp
```diff
@@ -1159,6 +1161,48 @@ int8_t *__catalyst__rt__array_get_element_ptr_1d(QirArray *ptr, int64_t idx)
...
+    const QubitIdType current_qubit_id = data[idx];
+
+    // If the ID of the qubit to insert is equal to the ID currently at the requested position in
+    // the register, there is nothing to do, and we return the unmodified array.
+    if (current_qubit_id == qubit_id) {
+        return;
+    }
+
+    // ~~ TODO ~~ //
+    // DANGER!! Ideally we should have a way to check that we are not overwriting an active qubit,
```
dime10: Isn't that why we are implementing this? To write a qubit into an array position which currently holds a different ID (which may then be inserted into another position, or just handled / dealloc'ed separately I suppose).

mlir/lib/Quantum/Transforms/ConversionPatterns.cpp
```diff
@@ -415,8 +415,30 @@ struct InsertOpPattern : public OpConversionPattern<InsertOp> {
     LogicalResult matchAndRewrite(InsertOp op, InsertOpAdaptor adaptor,
                                   ConversionPatternRewriter &rewriter) const override
     {
-        // Unravel use-def chain of quantum register values, converting back to reference semantics.
-        rewriter.replaceOp(op, adaptor.getInQreg());
```
dime10: I might be nice to gate this new behaviour behind some flag? Not a must, but since you mention several pitfalls that  could allow us to test it out safely

joeycarter: I added an option to the conversion pass: 63e08981

(I updated the PR description for context)

runtime/lib/capi/RuntimeCAPI.cpp
joeycarter: The danger is overwriting a qubit ID in the array representing the register. If we had the means to check at runtime if the qubit ID that would be overwritten is still active (i.e. it hadn't been deallocated yet), that at least this would trigger an assertion.

It depends on the design philosophy, to an extent. It's a similar problem to overwriting a pointer that points to dynamically allocated memory, resulting in a leak. That's a legal thing to do in many languages, so I could see the argument being made that it should be legal here as well. That is, it's up to the user to ensure that they are not inserting a qubit into the register and overwriting another, active qubit.

frontend/catalyst/ftqc/pipelines.py
```diff
@@ -62,12 +65,20 @@ def workload():
 
     _, pipeline = stages[llvm_dialect_conversion_stage_index]
 
+    print(pipeline)
```
multiphaseCFD: Could we remove it?

doc/releases/changelog-dev.md
```diff
@@ -223,6 +223,17 @@
...
 
+* Catalyst now supports *array-backed registers*, meaning that `quantum.insert` operations can be
+  configured to allow for the insertion of a qubit into an arbitrary position within a register.
+  [(#2000)](https://github.com/PennyLaneAI/catalyst/pull/2000)
+
+  This feature is disabled by default. To enable it, configure the pass pipeline to set the
+  `use-array-backed-registers` option of the `convert-quantum-to-llvm` pass to `true`. For example,
+
+  ```console
+  $ quantum-opt --pass-pipeline="builtin.module(convert-quantum-to-llvm{use-array-backed-registers=true})"
```
multiphaseCFD: Could we use `catalyst` to configure as well?

frontend/catalyst/ftqc/pipelines.py
joeycarter: Oops! :smile: 

Fixed: e50c0882

doc/releases/changelog-dev.md
joeycarter: Yes, you could also do

```console
catalyst --tool=opt --pass-pipeline="builtin.module(convert-quantum-to-llvm{use-array-backed-registers=true})" <input file>
```

I'll change this, since we don't ship `quantum-opt` in our wheels.

joeycarter: 9a8e1e49

frontend/catalyst/ftqc/pipelines.py
```diff
@@ -22,9 +22,12 @@
 def mbqc_pipeline() -> PipelineStages:
     """Return the pipeline stages for MBQC workloads.
 
-    The MBQC pipeline is identical to the default Catalyst pipeline, but with the MBQC-to-LLVM
-    dialect-conversion pass inserted immediately before the Quantum-to-LLVM dialect-conversion pass
-    in the ``MLIRToLLVMDialect`` pipeline stage.
+    The MBQC pipeline is identical to the default Catalyst pipeline, but where
```
paul0403:
```suggestion
    The MBQC pipeline is identical to the default Catalyst pipeline, but with
```

frontend/catalyst/ftqc/pipelines.py
```diff
@@ -63,11 +66,18 @@ def workload():
     _, pipeline = stages[llvm_dialect_conversion_stage_index]
 
     assert len(pipeline) > 0, "Pipeline for stage 'MLIRToLLVMDialect' is empty"
+    assert (
+        "convert-quantum-to-llvm" in pipeline
+    ), "Pipeline for stage 'MLIRToLLVMDialect' missing required pass 'convert-quantum-to-llvm'"
```
paul0403: I guess this new assert means the first assert can be removed?

mlir/lib/Quantum/Transforms/quantum_to_llvm.cpp
```diff
@@ -31,6 +31,7 @@ using namespace catalyst::quantum;
 namespace catalyst {
 namespace quantum {
 
+#define GEN_PASS_DECL_QUANTUMCONVERSIONPASS
```
paul0403, line 34: old-man-yells-at-mlir-boilerplate moment :laughing: 

I think `GEN_PASS_DECL_...` is only needed for passes with options (not sure), that's why this was never caught before.

runtime/lib/capi/RuntimeCAPI.cpp
paul0403: > Isn't that why we are implementing this? To write a qubit into an array position which currently holds a different ID

Yes, this is why we are implementing this. 

To put in another way, the current "insert runtime capi" (there isn't one) is just inserting back the same ID, i.e. if the ID array is `[100, 200, 300]`, the insert ops are "inserting `200` back into the array where `200` is", since runtime IR is back into reference semantics. 

Unlike the quantum dialect IR, gates in runtime IR are again just operations acting on a specified ID, instead of producing new qubit values. In other words the representation of a qubit is unchanged before and after an operation on it. e.g. `__catalyst__rt__Hadamard(qubit_id=200)` leaves the qubit ID still as 200. So, while in the quantum dialect IR the new qubit value needs to be inserted back into the register (it's a new `Value` now), runtime does not have this need.

And ftqc team wants to insert other IDs into where the ID 200 is currently, essentially.

--------------

I think this PR does not need to worry about the activity for 3 reasons:
1. Like Joey said, overwriting un-dealloced pointers/references and leading to a leak is a legal thing to do in many languages. It is up to the user to make sure their workflows do not cause leaks
2. The new insert stub is behind a flag already, so there already is protection in place for the "regular" program flow
3. Most importantly, if we were to implement the activity check right now, it will be significantly more work in this PR, as we don't have anything in runtime for this purpose:
- There is the `QubitManager.hpp` that kind of maintains the activities but the class is only used by lightning and null qubit. Not all devices are guaranteed to use it.
- Even for devices that use it, adding the activity query interface in the base `QuantumDevice` involves asking downstream device implementors to rebuild, which is something I think we should avoid unless there truly is an absolute need.
- But mostly, this keeping track of activity should really happen within the `RTDevice` layer, not the actual underlying device (I think Joey was only trying to do it in the actual device layer because the only thing resembling activity checks was in `QubitManager.hpp`). In my dynamic allocation work I actually need to implement it, because over there people want a clean pool (known to be fresh) and dirty pool (vice versa) of qubits, for things like qubit reuse. I would need to keep track of what qubits are deallocated into which pool, and as a happy by product I would be keeping track of which qubits are, well, dealloced. The insert guard can just use that functionality when it is in place.

runtime/lib/capi/RuntimeCAPI.cpp
```diff
@@ -1159,6 +1161,40 @@ int8_t *__catalyst__rt__array_get_element_ptr_1d(QirArray *ptr, int64_t idx)
...
+{
+    RT_ASSERT(getQuantumDevicePtr() != nullptr);
+    RT_ASSERT(CTX->getMemoryManager() != nullptr);
+    std::vector<QubitIdType> *qubit_vector_ptr = reinterpret_cast<std::vector<QubitIdType> *>(ptr);
+
+    RT_ASSERT(idx >= 0);
+
+    if (static_cast<size_t>(idx) >= qubit_vector_ptr->size()) {
+        std::string error_msg =
+            std::format("The qubit register does not contain the requested wire: {}", idx);
```
paul0403: Small comment: I always like more unification when possible, so let's use the same error message string handling as the one above in the extract capi? :thinking: 

```
    std::string error_msg = "The qubit register does not contain the requested wire: ";
    error_msg += std::to_string(idx);
```

This also means we don't need to pull in the `<format>` header. Though, if you want to update the extract capi above to also use std::format I'm also good with that!

runtime/tests/Test_NullQubit.cpp
```diff
@@ -16,6 +16,8 @@
 #include <catch2/catch_approx.hpp>
 #include <catch2/catch_test_macros.hpp>
 #include <catch2/matchers/catch_matchers_string.hpp>
+#include <cstdint>
+#include <cstdio>
```
paul0403, line 20: What are these new headers for?

runtime/tests/Test_NullQubit.cpp
```diff
@@ -31,6 +33,42 @@ using namespace Catch::Matchers;
 using namespace Catalyst::Runtime;
 using namespace Catalyst::Runtime::Devices;
 
+/**
+ * @brief A test fixture that initializes the Catalyst Runtime and sets up a null.qubit runtime
+ *        device with shots=0 and auto_qubit_management=false. The teardown operation releases the
+ *        device and terminates the Catalyst Runtime.
+ */
+struct NullQubitRuntimeFixture {
```
paul0403, line 41: Awesome!

I can't tell you how many times I have wanted this, having gone through multiple changes of the runtime device init capi's signature :laughing: 

We should find some time (not now) to port all eligible tests to use this fixture :+1:

runtime/tests/Test_NullQubit.cpp
```diff
@@ -414,6 +452,99 @@ TEST_CASE("test AllocateQubits generates a proper std::vector<QubitIdType>", "[N
...
+    QirArray *reg = __catalyst__rt__qubit_allocate_array(3);
+
+    auto reg_vec_before = *reinterpret_cast<std::vector<QubitIdType> *>(reg);
+
+    CHECK(reg_vec_before[0] == 0);
+    CHECK(reg_vec_before[1] == 1);
+    CHECK(reg_vec_before[2] == 2);
+
+    // Release qubit 1; note that array `reg` is still [0, 1, 2]
+    __catalyst__rt__qubit_release(reinterpret_cast<QUBIT *>(1));
```
paul0403:
```suggestion
    // Release qubit 1; note that array `reg` is still [0, 1, 2]
    __catalyst__rt__qubit_release(reinterpret_cast<QUBIT *>(reg_vec_before[1]));
```

let's get the ID from the register array properly instead of hard coding the ID : ) I know the ID generation scheme for null qubit will probably never change, but I always like to reduce tech debt even in small places whenever possible. Especially in tests, since random unrelated tests failing when I'm developing something else always gives me anxiety :laughing:

runtime/tests/Test_NullQubit.cpp
```diff
@@ -414,6 +452,99 @@ TEST_CASE("test AllocateQubits generates a proper std::vector<QubitIdType>", "[N
...
+    CHECK(reg1_vec[0] == 0);
+
+    auto reg2_vec = *reinterpret_cast<std::vector<QubitIdType> *>(reg2);
+
+    CHECK(reg2_vec[0] == 1);
+    CHECK(reg2_vec[1] == 2);
+
+    // Release qubit 0 (from reg1) and qubit 1 (from reg2)
+    __catalyst__rt__qubit_release(reinterpret_cast<QUBIT *>(0));
+    __catalyst__rt__qubit_release(reinterpret_cast<QUBIT *>(1));
```
paul0403: same here

runtime/tests/Test_NullQubit.cpp
```diff
@@ -414,6 +452,99 @@ TEST_CASE("test AllocateQubits generates a proper std::vector<QubitIdType>", "[N
...
+    QirArray *reg = __catalyst__rt__qubit_allocate_array(3);
+
+    auto reg_vec_before = *reinterpret_cast<std::vector<QubitIdType> *>(reg);
+
+    // Allocate an individual qubit; internally it has ID 3
+    QUBIT *q = __catalyst__rt__qubit_allocate();
+
+    // Insert the individual qubit (ID: 3) into `reg` at position 3
+    // This should raise an exception position 3 is beyond the size of the register
+    CHECK_THROWS(__catalyst__rt__array_update_element_1d(reg, 3, reinterpret_cast<QUBIT *>(q)));
```
paul0403: You can check the error message string as well with 
```C
    REQUIRE_THROWS_WITH(
        ...bad code...,
        ContainsSubstring("Error message"));
```

frontend/catalyst/ftqc/pipelines.py
joeycarter: Good catch. "where" matches the second bullet point, although yes, "with" matches the first. I'll clean this up.

frontend/catalyst/ftqc/pipelines.py
joeycarter: Hm, strictly speaking, yes. If the second assertion triggers, then the first one will always also trigger. Although, I like having both because they offer different error messages. Maybe overkill for production code though. I'll clean this up.

mlir/lib/Quantum/Transforms/quantum_to_llvm.cpp
joeycarter: This took me way to long to figure out, but *yes!* We have to define `GEN_PASS_DECL_QUANTUMCONVERSIONPASS` because this is what's in the `Passes.h.inc` header file generated by the TableGen:

```c
#ifdef GEN_PASS_DECL_QUANTUMCONVERSIONPASS
struct QuantumConversionPassOptions {
  bool useArrayBackedRegisters = false;
};
#undef GEN_PASS_DECL_QUANTUMCONVERSIONPASS
#endif // GEN_PASS_DECL_QUANTUMCONVERSIONPASS
``` 

We never caught this before because it wasn't necessary until now; the quantum-conversion pass didn't have any pass options before this PR.

runtime/lib/capi/RuntimeCAPI.cpp
joeycarter: Fair enough, I'll change it to string concatenation. We can open a "good first issue" for updating our code base to use all the modern C++20 features :slightly_smiling_face:

runtime/tests/Test_NullQubit.cpp
joeycarter: I have a VS Code extension ([clangd](https://clangd.llvm.org/)) that automatically adds headers for types/functions/classes/etc. that you use.

`cstdint` is for fixed width integer types, like `int8_t`, and `cstdio` is for file I/O, like `fopen`. We don't need these includes here, strictly speaking, because they're included indirectly from other headers. But having them here follows the principles of [include-what-you-use](https://include-what-you-use.org/).

runtime/tests/Test_NullQubit.cpp
joeycarter: Agreed!

runtime/tests/Test_NullQubit.cpp
joeycarter: Good call.

runtime/tests/Test_NullQubit.cpp
joeycarter: Nice! I'll update it

joeycarter: d8b39cb5

runtime/lib/capi/RuntimeCAPI.cpp
joeycarter: d8b39cb5

runtime/tests/Test_NullQubit.cpp
joeycarter: 6f213f8d

runtime/tests/Test_NullQubit.cpp
joeycarter: 6f213f8d

frontend/catalyst/ftqc/pipelines.py
joeycarter: b901d211

frontend/catalyst/ftqc/pipelines.py
joeycarter: b901d211

paul0403 (approved):
:partying_face:

PR #2004, opened 2025-08-26
Add support for classical return values in one-shot mode

rniczh:
**Context:**

Originally, the output from one-shot only consider the measurements, to make it possible to let classical values happen in return values, this PR first classify the classical values from the `return_values_flat` in `jax_tracer.py` and used it later to insert back to the result with proper position in `qfunc.py`.

Now, we can support the following behaviours in one-shot. `circuit_1()` and `circuit_2()` would expect to produce `[True|False]+` and `[42|43]+` with respect to the mcm result of each shot. 

```python
import pennylane as qml
@qjit
@qml.set_shots(10)
@qml.qnode(qml.device("lightning.qubit", wires=1), mcm_method="one-shot")
def circuit_1():
    qml.Hadamard(wires=0)
    return measure(0)
print(circuit_1())
```

```python
import pennylane as qml
@qjit(autograph=True)
@qml.set_shots(10)
@qml.qnode(qml.device("lightning.qubit", wires=1), mcm_method="one-shot")
def circuit_2():
    qml.Hadamard(wires=0)
    if measure(0):
        return 42
    else:
        return 43

print(circuit_2())
```

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


frontend/catalyst/jax_tracer.py
```diff
@@ -38,6 +38,7 @@
     CountsMP,
     ExpectationMP,
     MeasurementProcess,
+    MidMeasureMP,
```
paul0403:
```suggestion
```
since codefactor says it's unused

frontend/catalyst/jax_tracer.py
```diff
@@ -1137,6 +1139,16 @@ def is_midcircuit_measurement(op):
     return any(map(is_midcircuit_measurement, tape.operations))
 
 
+@debug_logger
+def num_midcircuit_measurement(tape):
+    """Number of mid-circuit measurements."""
+
+    def is_midcircuit_measurement(op):
+        return isinstance(op, catalyst.api_extensions.MidCircuitMeasure)
```
paul0403: def inside a def? :thinking: I guess making this a lambda hits the line-too-long?

frontend/catalyst/jax_tracer.py
```diff
@@ -1441,7 +1476,14 @@ def is_leaf(obj):
                 # changed the output. See `split_non_commuting`
                 if tracing_mode == TracingMode.TRANSFORM:
                     # TODO: In the future support arbitrary output from the user function.
-                    output = tape.measurements
+                    if uses_transform(qnode, "dynamic_one_shot_partial", mode="only_one"):
```
paul0403: Why do we need only-one mode here?

frontend/catalyst/qfunc.py
```diff
@@ -145,20 +240,26 @@ def __call__(self, *args, **kwargs):
...
+                trace_quantum_function(
+                    self.func,
+                    qjit_device,
+                    args,
+                    kwargs,
+                    self,
+                    static_argnums,
+                    debug_info,
+                )
             )
```
paul0403: Is there an extra layer of brakets here? :sweat_smile:

frontend/test/pytest/test_mid_circuit_measurement.py
```diff
@@ -779,6 +779,26 @@ def cost():
...
+                "first": qml.sample(),
+                "second": [100, qml.sample()],
+                "third": (qml.sample(), qml.sample()),
+            }
+
+        assert list(circuit().keys()) == ["first", "second", "third"]
+        assert jnp.array(circuit()["first"]).shape == (12, 1)
+        assert jnp.allclose(circuit()["second"][0], jnp.full(12, 100))
+        assert jnp.array(circuit()["second"][1]).shape == (12, 1)
+        assert jnp.array(circuit()["third"]).shape == (2, 12, 1)
```
paul0403:
```suggestion
        result = circuit()
        
        assert list(result.keys()) == ["first", "second", "third"]
        assert jnp.array(result["first"]).shape == (12, 1)
        assert jnp.allclose(result["second"][0], jnp.full(12, 100))
        assert jnp.array(result["second"][1]).shape == (12, 1)
        assert jnp.array(result["third"]).shape == (2, 12, 1)
```
let's save some testing time whenever we can : )

paul0403: Also is it weird that `result["second"][0]` is now an array of 12 `100`s instead of just one `100`? :thinking: What performed the broadcasting?

frontend/catalyst/jax_extras/tracing.py
```diff
@@ -22,7 +22,7 @@
 from copy import copy
 from dataclasses import dataclass
 from functools import partial
-from typing import Any, Callable, Dict, Generator, List, Optional, Set, Tuple, TypeVar
+from typing import Any, Callable, Dict, Generator, List, Literal, Optional, Set, Tuple, TypeVar
```
paul0403: Don't forget the changelog!

frontend/test/pytest/test_mid_circuit_measurement.py
rniczh: It's not promise that classical values always be the same, taking the following examples, `circuit_1()` and `circuit_2()` would expect to produce `[True|False]+` and `[42|43]+`. But of course, when changing to single-branch, it should only return a single value here.

```python
import pennylane as qml
@qjit
@qml.set_shots(10)
@qml.qnode(qml.device("lightning.qubit", wires=1), mcm_method="one-shot")
def circuit_1():
    qml.Hadamard(wires=0)
    return measure(0)
print(circuit_1())
```

```python
@qjit(autograph=True)
@qml.set_shots(10)
@qml.qnode(qml.device("lightning.qubit", wires=1), mcm_method="one-shot")
def circuit_2():
    qml.Hadamard(wires=0)
    if measure(0):
        return 42
    else:
        return 43

print(circuit_2())
```

frontend/catalyst/qfunc.py
rniczh: I think it's because this line exceeds 100 char limits, python formatter does it automatically. 🤔

rniczh: But also, here is an issue that codefactor report that there are too many return values from trace_quantum_function ...👀 I will try to simplify it. So I guess after I simplify it, I can remove this bracket as well.

frontend/catalyst/jax_tracer.py
rniczh: There might have been an issue before. I replaced both of those functions with lambdas, and it seems to be working now.

frontend/catalyst/qfunc.py
rniczh: I refactored the code, changed the structure of the `trace_quantum_function`

frontend/catalyst/jax_tracer.py
rniczh: This issue doesn't belong here, so I've removed this `mode` for now

frontend/catalyst/jax_tracer.py
rniczh: Done

frontend/test/pytest/test_mid_circuit_measurement.py
paul0403: Yes the classical values might be different so they need one per shot, but this shouldn't affect dictionary keys (which are never different) since dictionary keys are handled by pytrees. It should still be a pytree with dictionary key 10

paul0403: I feel like this is a product question, let's ask first.

paul0403: Discussed with product (@isaacdevlugt ), for cases where classical values are returned we fallback mcms to single branch statistics for now and raise a warning saying the user's `mcm_method="one-shot"` has been changed to single branch

dime10: What is the concern? I think this PR is a great addition and solves the problem of handling classical values in a shot-based setting. Falling back to single-branch entirely defeats the purpose of using one-shot and isn't very useful imo.

paul0403: Some classical values are not guaranteed to be the same per shot. The question is whether to broadcast the classical value across all shots. [slack thread](https://xanaduhq.slack.com/archives/C06VCNYC3LH/p1756827854653189)

isaacdevlugt: @dime10 happy to litigate this again if we've missed anything!

dime10: I think it's okay, the proposed plan is equivalent to what @rniczh is implementing here, except that we'll support arbitrary MPs for classical return values, not just samples (this PR).

The question is what we are doing in the meantime, if we don't merge this PR then in the meantime we are disallowing classical return values (regression)? I don't think it makes sense to fallback to single-branch-statistics, since that behaviour will be very confusing 🤔 
The reason I like this PR until the proposed MPs for classical values are implemented, is because it's a very natural extension of the existing pattern in catalyst (where `return x` produces a single classical value with single-branch-statistics, and produces one value for each shot with dynamic-one-shot), and is a stepping stone to adding support for arbitrary MPs (being that sample will already be implemented).

paul0403: > the existing pattern in catalyst (where `return x` ...  produces one value for each shot with dynamic-one-shot)

Wait, this is the current behavior? :exploding_head:

paul0403: Ah wait, you are saying 

> produces a single classical value with single-branch-statistics

is the existing behavior, and 

> produces one value for each shot with dynamic-one-shot

is the natural extension :sweat_smile: I was tripped by the sentence structure

If we are ok with this behavior then this PR can go in!

paul0403: @isaacdevlugt is the above proposed behavior ok? Essentially, 

```python
dev = qml.device("lightning.qubit", wires=1, shots=5)

@qml.qjit(autograph=True)
@qml.qnode(dev, mcm_method="one-shot")
def circuit():
   qml.Hadamard(wires=0)
   m = measure(wires=0)
   if m:
      x = 10
   else:
      x = 20
   return x
```

```
[[10], [10], [20], [10], [20]]  # pseudo result, true result will be properly wrapped in np arrays with proper shapes
```

so one classical return value for each shot in one shot mode.

doc/releases/changelog-dev.md
```diff
@@ -28,6 +28,27 @@
 
 <h3>Improvements 🛠</h3>
 
+* Catalyst now supporting classical return values with one-shot mode.
```
```suggestion
* Catalyst now supports classical return values with one-shot mode.
```

doc/releases/changelog-dev.md
```diff
@@ -28,6 +28,27 @@
...
+  For example, the code below will generate 10 values, with an equal probability of 42 and 43
+  appearing. You can also return `measure(0)` as well.
+  [(#2001)](https://github.com/PennyLaneAI/catalyst/pull/2001)
+
+  ```python
+  import pennylane as qml
```
paul0403, line 38:
```suggestion
```

In our documentation and blog posts we always assume `import pennylane as qml` is run

doc/releases/changelog-dev.md
```diff
@@ -28,6 +28,27 @@
...
+  @qml.set_shots(10)
+  @qml.qnode(qml.device("lightning.qubit", wires=1), mcm_method="one-shot")
+  def circuit_2():
+      qml.Hadamard(wires=0)
+      if measure(0):
+          return 42
+      else:
+          return 43
+
+  print(circuit_2())
```
```suggestion
  def circuit():
      qml.Hadamard(wires=0)
      if measure(0):
          return 42
      else:
          return 43

  print(circuit())
```

paul0403: Also, make sure to actually show the output when there is codeblock in documentation!

frontend/catalyst/jax_tracer.py
```diff
@@ -1340,10 +1394,232 @@ def _get_total_shots(qnode):
...
+            tapes=tapes,
+            post_processing=post_processing,
+            tracing_mode=tracing_mode,
+            return_values_flat=return_values_flat,
+            return_values_tree=return_values_tree,
+            is_leaf=is_leaf,
+        )
+
+
+def _trace_quantum_phase(
```
paul0403:
```suggestion
def _trace_quantum_step(
```

Let's rename them to `trace_quantum/classical_step` because "quantum phase" sounds like the phase gate : )

frontend/test/pytest/test_mid_circuit_measurement.py
isaacdevlugt: I have to admit I'm a little confused now 😅

dime10: > [[10], [10], [20], [10], [20]]

Why the second list dimension?

paul0403: > Why the second list dimension?

There might be multiple classical return values, so I think each shot needs its own container.


> I have to admit I'm a little confused now 😅

What would you @isaacdevlugt expect the above example to return in one shot mcm? I think what David is saying 

>  (where return x produces a single classical value with single-branch-statistics, and produces one value for each shot with dynamic-one-shot), and is a stepping stone to adding support for arbitrary MPs (being that sample will already be implemented)

is that `return x` in one shot mode can just be interpreted as `return sample(x)` directly, since catalyst already supports returning classical values from qnodes.

isaacdevlugt: @paul0403 I would expect the output to be `[[10], [10], [20], [10], [20]]` yes! I think this might be a divergence from the conversation internally (though I like the ability to return `x` directly instead of wrapping it in an MP 😅). Is `qml.sample(x)` also supported in this PR? This should give feature parity with how PL can eventually support this as well

rniczh: @isaacdevlugt No, `qml.sample(x)` is not supported in this PR, it will route to pennylane and ran into an error

paul0403: > Is `qml.sample(x)` also supported in this PR?

No, catalyst sample operation must take an observable. In other words it must be quantum. This PR doesn't change any catalyst internals, it is just broadcasting classical return values across all shots.

dime10: > There might be multiple classical return values, so I think each shot needs its own container.

That's not how we handle output shapes. That is determined by the actual Pytree format the user specified. 
For
```
return x
```
the output should be
```
Array([10, 20, ...])
```
for
```
return x, qml.expval
```
the output should be
```
(Array([10, 20, ...]), Array(0.313))
```
for
```
return {"help": x}, qml.expval
```
the output should be
```
({"help": Array([10, 20, ...])}, Array(0.313))
```
etc.

dime10: > No, catalyst sample operation must take an observable. In other words it must be quantum. This PR doesn't change any catalyst internals, it is just broadcasting classical return values across all shots.

Catalyst sample doesn't work with observables, but like Hongsheng said this needs to implemented in upstream PL first to allow classical values in MPs.

isaacdevlugt: Gotcha! Thanks guys. 

>That's not how we handle output shapes. That is determined by the actual Pytree format the user specified.

I think David is right here. This (i.e., `Array([10, 10, 20, 10, 20])`) is also consistent with PL I believe

paul0403: > That is determined by the actual Pytree format the user specified.

Ah I see, so when the user returns multiple classical values we still follow the pytree they asked for in each shot :+1:

rniczh: Just quick check here, does anyone still has any questions or concerns about this right now?

paul0403: All good from my end.

doc/releases/changelog-dev.md
```diff
@@ -28,6 +28,33 @@
 
 <h3>Improvements 🛠</h3>
 
+* Catalyst now supports classical return values with one-shot mode.
```
```suggestion
* Catalyst now supports returning classical and MCM values with the dynamic one-shot MCM method.
```

doc/releases/changelog-dev.md
```diff
@@ -28,6 +28,33 @@
...
+          return 43
+
+  print(circuit())
+  ```
+
+  which produces either 42 or 43 with equal probability (approximately 50% each)
+
+  ```pycon
+  [43 42 42 42 43 42 43 42 43 42]
+  ```
```
````suggestion
  
  For example, the code below will generate 10 values, with an equal probability of 42 and 43
  appearing. 

  ```python
  import pennylane as qml
  
  @qjit(autograph=True)
  @qml.set_shots(10)
  @qml.qnode(qml.device("lightning.qubit", wires=1), mcm_method="one-shot")
  def circuit():
      qml.Hadamard(wires=0)
      m = measure(0)
      if m:
          return 42, m
      else:
          return 43, m
  ```

  ```pycon
  >>>  print(circuit())
  (Array([42, 43, 42, 42, 43, 42, 42, 43, 42, 42], dtype=int64),
   Array([ True, False,  True,  True, False,  True,  True, False,  True,
           True], dtype=bool))

  ```
````

Comment (isaacdevlugt) [reply doc/releases/changelog-dev.md]: Shows off both features (classical vals and MCM vals in return statement)!

**Review by isaacdevlugt — CHANGES_REQUESTED** [review]
Nice! Just left suggestions on the changelog.

File: doc/releases/changelog-dev.md (hunk 12 above)
Comment (rniczh) [reply doc/releases/changelog-dev.md]: Thanks!

File: doc/releases/changelog-dev.md [hunk 13]
```diff
@@ -28,6 +28,37 @@
...
+      else:
+          return 43, m
+  ```
+
+  ```pycon
+  >>>  print(circuit())
+  (Array([42, 43, 42, 42, 43, 42, 42, 43, 42, 42], dtype=int64),
+   Array([ True, False,  True,  True, False,  True,  True, False,  True,
+           True], dtype=bool))
+
```
```suggestion
```

PR #2005, opened 2025-08-26
Add a jax primitive for single qubit dealloc

paul0403:
**Context:**
There is an mlir op for the single qubit deallocation (which lowers to `__catalyst__rt__qubit_release`), but there is no frontend jax primitive for it.

Add a jax primitive for single qubit dealloc.

**Benefits:**
Things requiring single qubit deallocation like ftqc, decomposition, and dynamic allocation itself, can start doing so in plxpr conversion.

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -56,11 +56,11 @@
   %0 = transform.apply_registered_pass "some-pass" with options = {"an-option" = true, "maxValue" = 1 : i64, "multi-word-option" = 1 : i64}
   ```
 
-*  Commuting Clifford Pauli Product Rotation (PPR) operations, past non-Clifford PPRs, now supports P(π/2) Cliffords in addition to P(π/4)
-   [(#1966)](https://github.com/PennyLaneAI/catalyst/pull/1966)
+* Commuting Clifford Pauli Product Rotation (PPR) operations, past non-Clifford PPRs, now supports P(π/2) Cliffords in addition to P(π/4)
+  [(#1966)](https://github.com/PennyLaneAI/catalyst/pull/1966)
 
-*  A new jax primitive `qdealloc_qb_p` is available for single qubit deallocations.
-   [(#2005)](https://github.com/PennyLaneAI/catalyst/pull/2005)
+* A new jax primitive `qdealloc_qb_p` is available for single qubit deallocations.
+  [(#2005)](https://github.com/PennyLaneAI/catalyst/pull/2005)
 
 
 <h3>Breaking changes 💔</h3>
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -53,8 +53,12 @@
   %0 = transform.apply_registered_pass "some-pass" with options = {"an-option" = true, "maxValue" = 1 : i64, "multi-word-option" = 1 : i64}
   ```
 
-*  `Commuting Clifford Pauli Product Rotation (PPR) operations, past non-Clifford PPRs, now supports P(π/2) Cliffords in addition to P(π/4)`
+*  Commuting Clifford Pauli Product Rotation (PPR) operations, past non-Clifford PPRs, now supports P(π/2) Cliffords in addition to P(π/4)
    [(#1966)](https://github.com/PennyLaneAI/catalyst/pull/1966)
+
+*  A new jax primitive `qdealloc_qb_p` is available for single qubit deallocations.
+   [(#2005)](https://github.com/PennyLaneAI/catalyst/pull/2005)
+
 
 <h3>Breaking changes 💔</h3>
 
--- a/frontend/catalyst/jax_primitives.py
+++ b/frontend/catalyst/jax_primitives.py
@@ -76,6 +76,7 @@ from mlir_quantum.dialects.quantum import (
     CountsOp,
     CustomOp,
     DeallocOp,
+    DeallocQubitOp,
     DeviceInitOp,
     DeviceReleaseOp,
     ExpvalOp,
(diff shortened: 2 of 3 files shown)


paul0403:
@mehrdad2m seems like lit test codecov is not working?

mehrdad2m:
> @mehrdad2m seems like lit test codecov is not working?

🙀

doc/releases/changelog-dev.md
```diff
@@ -56,9 +56,13 @@
   %0 = transform.apply_registered_pass "some-pass" with options = {"an-option" = true, "maxValue" = 1 : i64, "multi-word-option" = 1 : i64}
   ```
 
-*  `Commuting Clifford Pauli Product Rotation (PPR) operations, past non-Clifford PPRs, now supports P(π/2) Cliffords in addition to P(π/4)`
+*  Commuting Clifford Pauli Product Rotation (PPR) operations, past non-Clifford PPRs, now supports P(π/2) Cliffords in addition to P(π/4)
    [(#1966)](https://github.com/PennyLaneAI/catalyst/pull/1966)
 
+*  A new jax primitive `qdealloc_qb_p` is available for single qubit deallocations.
+   [(#2005)](https://github.com/PennyLaneAI/catalyst/pull/2005)
```
joeycarter: Looks like an extra space snuck its way in (and in the bullet point above as well)

```suggestion
* A new jax primitive `qdealloc_qb_p` is available for single qubit deallocations.
  [(#2005)](https://github.com/PennyLaneAI/catalyst/pull/2005)
```

frontend/test/lit/test_dynamic_qubit_allocation.py
```diff
@@ -0,0 +1,44 @@
...
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Unit tests for the dynamic qubit allocation.
+Note that this feature is only available under the plxpr pipeline.
```
joeycarter: Do you think it might be misleading to state this in the test suite given that we do not have to enable capture for this test?

frontend/catalyst/jax_primitives.py
```diff
@@ -259,6 +260,8 @@ class MeasurementPlane(Enum):
 qalloc_p = Primitive("qalloc")
 qdealloc_p = Primitive("qdealloc")
 qdealloc_p.multiple_results = True
+qdealloc_qb_p = Primitive("qdealloc_qb")
+qdealloc_qb_p.multiple_results = True
```
joeycarter, line 264: Can you remind me why we need `multiple_results = True` here given that the abstract_eval function defined below returns `()`?

joeycarter (commented):
Looks good, just a few questions.

frontend/catalyst/jax_primitives.py
paul0403: So jax has a bunch of misnomers (or straight up incomprehensible names... I still don't know what "partial eval" means lol). `multiple_results` on a primitive needs to be set as long as the return type is a list/tuple, regardless of how many things that list/tuple actually has. And this includes having zero things.

doc/releases/changelog-dev.md
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2005/commits/d0169e8e3ca908d1b85ba9a4fdc2be3eb731c411

frontend/test/lit/test_dynamic_qubit_allocation.py
paul0403: Good point, I couldn't find any lit test files that this unit test would fit cleanly into, so I put it into the dynamic allocation lit test file that I will be creating. 

The dynamic allocation tests themselves will be plxpr-only, but you are right the dealloc qb unit test does not need plxpr. What would be your suggestion for wording/organizing this?

joeycarter: I think just `Unit tests for dynamic qubit allocation` will suffice for now. The documentation stating that dynamic qubit allocation is only available with PLxPR should probably be put in a more visible location. Do we have user-facing docs for dynamic qubit allocation yet? If so, it should go there.

frontend/catalyst/jax_primitives.py
joeycarter: Got it, thanks :slightly_smiling_face:

frontend/test/lit/test_dynamic_qubit_allocation.py
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2005/commits/0b0afc80e316fc025a9902599b689767410fb1ad

The documentation of dynamic qubit allocation will be in core PL so we don't need to worry about it here.

paul0403:
> Regarding the coverage, I tested running coverage locally and it worked fine which means it doesn't show those lines as missing. not sure why the patch coverage isn't catching the lit coverage. Let's hope this was a codecov glitch.

I don't think it's a glitch, since this is reproducible in CI. If it works locally but not on CI chances are the check catalyst script did not call the coverage step properly?

joeycarter:
> I know ftqc will not be inserting single qubit deallocs from frontend, but from inside the mlir layer with a pass. (@joeycarter can confirm this?)

Yes, we will need the ability to deallocate individual qubits in a register at the IR layer (but not necessarily from the frontend).

paul0403:
(I am happy to revert this if there is no use case, I made this before 

> it turned out to be very awkward for the "qml.allocate = quantum.alloc" implementation to work well with the capacity interpretation, and it got strong pushback from core.

this happened and thought this would be necessary.)

PR #2006, opened 2025-08-26
Checks if all control and targets in multi-qubit operations at runtime are different

ritu-thombre99:
**Context:**
RuntimeCAPI.cpp checks if control and target qubit operand for CNOT gate `__catalyst__qis__CNOT` are different.
But the same checks are missing in other multi-qubit gates like CRY, CRZ, CRot, Toffoli, CSWAP etc.

```python
import pennylane as qml
import numpy as np
dev = qml.device("lightning.qubit")
@qml.qjit(autograph=True)
@qml.qnode(dev)
def circuit(x):
    qml.Hadamard(0)
    qml.CRY(0.5, [x,1])
    return qml.state()
print(circuit(1))
```
prints the following (printed NamedOperation from NullQubit)

```pycon
Name: Hadamard
Wires : 0,
Name: CRY
Wires : 1,1,
[1.+0.j 0.+0.j 0.+0.j 0.+0.j]
```
The code still executes and produces state when executed with lightning qubit:

```python
import pennylane as qml
import numpy as np
dev = qml.device("lightning.qubit")=
@qml.qjit(autograph=True)
@qml.qnode(dev)
def circuit(x):
    qml.Hadamard(0)
    qml.CRY(0.5, [x,1])
    return qml.state()
print(circuit(1))
```

which prints 
```pycon
[0.70710678+0.j 0.        +0.j 0.70710678+0.j 0.        +0.j]
``` 

i.e. $\frac{1}{2}(|0\rangle + |1\rangle)|0\rangle$. The state produced is as if only first Hadamard gate was applied. This could have been hard to debug with printing all the operations using null qubit

Added checks to make sure all the operands in all the multi-qubit gates at runtime are distinct.

**Related GitHub Issues:**

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -56,6 +56,9 @@
   %0 = transform.apply_registered_pass "some-pass" with options = {"an-option" = true, "maxValue" = 1 : i64, "multi-word-option" = 1 : i64}
   ```
 
+* Added checks to make sure all the operands in all the multi-qubit gates in `RuntimeCAPI.cpp` are distinct. 
+  [(#2006)](https://github.com/PennyLaneAI/catalyst/pull/2006).
+
 *  `Commuting Clifford Pauli Product Rotation (PPR) operations, past non-Clifford PPRs, now supports P(π/2) Cliffords in addition to P(π/4)`
    [(#1966)](https://github.com/PennyLaneAI/catalyst/pull/1966)
 
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -56,7 +56,7 @@
   %0 = transform.apply_registered_pass "some-pass" with options = {"an-option" = true, "maxValue" = 1 : i64, "multi-word-option" = 1 : i64}
   ```
 
-* Added checks to make sure all the operands in all the multi-qubit gates in `RuntimeCAPI.cpp` are distinct. 
+* Added checks to raise an error when the input qubits to the multi-qubit gates in the runtime CAPI are not all distinct. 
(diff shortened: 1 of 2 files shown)


runtime/lib/capi/RuntimeCAPI.cpp
```diff
@@ -776,6 +814,8 @@ void __catalyst__qis__CSWAP(QUBIT *control, QUBIT *aswap, QUBIT *bswap, const Mo
 
 void __catalyst__qis__Toffoli(QUBIT *wire0, QUBIT *wire1, QUBIT *wire2, const Modifiers *modifiers)
 {
+    RT_FAIL_IF((wire0 == wire1 || wire1 == wire2 || wire0 == wire2),
+               "Invalid input for Toffoli gate. All three qubit operands must be distinct..");
```
paul0403:
```suggestion
               "Invalid input for Toffoli gate. All three qubit operands must be distinct.");
```

paul0403 (commented):
Don't forget the changelog!

doc/releases/changelog-dev.md
```diff
@@ -56,6 +56,9 @@
   %0 = transform.apply_registered_pass "some-pass" with options = {"an-option" = true, "maxValue" = 1 : i64, "multi-word-option" = 1 : i64}
   ```
 
+* Added checks to make sure all the operands in all the multi-qubit gates in `RuntimeCAPI.cpp` are distinct.
```
```suggestion
* Added checks to raise an error when the input qubits to the multi-qubit gates in the runtime CAPI are not all distinct. 
```

1. It's just the qubit arguments, as opposed to all the inputs (which includes other things)
2. "Operands" are not a thing outside of mlir now, so these are just inputs to C functions : )

PR #2009, opened 2025-08-28
chore: remove autograph kwarg from qnode

andrijapau:
**Context:**

We are removing the `autograph` kwarg at the QNode level in favour of `disable_autograph` / `run_autograph` + `qjit` (see PL PR https://github.com/PennyLaneAI/pennylane/pull/8104).

~~⚠️ Tests are failing but that is expected and should be fixed once the PL PR merges as autograph will be off by default.~~

[sc-92903]

### DIFF
--- a/.dep-versions
+++ b/.dep-versions
@@ -10,7 +10,7 @@ enzyme=v0.0.186
 
 # For a custom PL version, update the package version here and at
 # 'doc/requirements.txt'
-pennylane=0.43.0.dev39
+pennylane=0.43.0.dev42
 
 # For a custom LQ/LK version, update the package version here and at
 # 'doc/requirements.txt'
--- a/.dep-versions
+++ b/.dep-versions
@@ -10,7 +10,7 @@ enzyme=v0.0.186
 
 # For a custom PL version, update the package version here and at
 # 'doc/requirements.txt'
-pennylane=0.43.0.dev42
+pennylane=0.43.0.dev43
 
 # For a custom LQ/LK version, update the package version here and at
 # 'doc/requirements.txt'
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -33,4 +33,4 @@ lxml_html_clean
 --extra-index-url https://test.pypi.org/simple/
 pennylane-lightning-kokkos==0.43.0.dev15
 pennylane-lightning==0.43.0.dev15
-pennylane==0.43.0.dev42
+pennylane==0.43.0.dev43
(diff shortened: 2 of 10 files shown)


andrijapau:
> Remember to update dep versions once changes in PL are available on test pypi 👍

https://github.com/PennyLaneAI/catalyst/pull/2009/commits/14e153c972d5066f1d2e04d974325fb638b75369. Might push wheels today or just wait until tomorrow - will see.

.dep-versions
```diff
@@ -10,7 +10,7 @@ enzyme=v0.0.186
 
 # For a custom PL version, update the package version here and at
 # 'doc/requirements.txt'
-pennylane=0.43.0.dev39
+pennylane=0.43.0.dev42
```
andrijapau:
```suggestion
pennylane=0.43.0.dev43
```

doc/requirements.txt
```diff
@@ -33,4 +33,4 @@ lxml_html_clean
 --extra-index-url https://test.pypi.org/simple/
 pennylane-lightning-kokkos==0.43.0.dev15
 pennylane-lightning==0.43.0.dev15
-pennylane==0.43.0.dev39
+pennylane==0.43.0.dev42
```
andrijapau:
```suggestion
pennylane==0.43.0.dev43
```

frontend/test/pytest/test_autograph.py
```diff
@@ -992,6 +992,7 @@ def f(n: int):
     def test_for_in_dynamic_range_indexing_array(self):
         """Test for loop over a Python range with dynamic bounds that is used to index an array."""
 
+        @qjit(autograph=True)
         @qml.qnode(qml.device("lightning.qubit", wires=1))
```
paul0403, line 996: I am a bit afraid that this test used to be outside qjit but is now inside qjit.

andrijapau: was just about to tag you lol

paul0403: Looking at it more, I think it actually makes more sense to have this test under qjit :sweat_smile: the old test seems to just be testing lightning itself

andrijapau: The test file is `test_autograph` and given that `qjit` is now the entry to autograph wouldn't this change make sense?

paul0403: ...so adding qjit is actually something we want here I would say! All good then

andrijapau: Cool

PR #2013, opened 2025-08-29
Migrate some more lit tests to `qml.set_shots`

paul0403:
**Context:**
Some python lit tests have not been migrated to `qml.set_shots` and are still emitting deprecation warnings.

Use `qml.set_shots(...)` instead of `qml.device(shots=...)` in more python lit tests.

**Benefits:**
No warnings.


paul0403:

PR #2014, opened 2025-08-30
Add depth analysis on PPM Specs

sengthai:
**Context:**
Previously, `--ppm-specs` analyzed the PPR/PPM's instructions count and weight. Here, we implement PPR/PPM's depth analysis.

- Update `mlir/lib/QEC/Transforms/CountPPMSpecs.cpp`

### DIFF
--- a/mlir/lib/QEC/Transforms/CountPPMSpecs.cpp
+++ b/mlir/lib/QEC/Transforms/CountPPMSpecs.cpp
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "QEC/IR/QECOpInterfaces.h"
 #define DEBUG_TYPE "ppm-specs"
 
 #include <algorithm>
@@ -25,6 +26,8 @@
 
 #include "Catalyst/Utils/SCFUtils.h"
 #include "QEC/IR/QECDialect.h"
+#include "QEC/Utils/QECLayer.h"
+#include "QEC/Utils/QECOpUtils.h"
 #include "Quantum/IR/QuantumOps.h"
 
 using namespace llvm;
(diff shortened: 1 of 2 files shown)


sengthai:
We should also add more tests in Python.

mlir/lib/QEC/Transforms/CountPPMSpecs.cpp
```diff
@@ -117,37 +120,111 @@ struct CountPPMSpecsPass : public impl::CountPPMSpecsPassBase<CountPPMSpecsPass>
...
+            // Count PPRs
             else if (isa<qec::PPRotationOp>(op)) {
                 if (failed(countPPR(cast<qec::PPRotationOp>(op), &PPMSpecs, &stringAllocator))) {
                     return WalkResult::interrupt();
                 }
                 return WalkResult::advance();
             }
+
+            // Skip other ops
             else {
```
rniczh, line 216: Is there any reason to use `skip` here? `skip` will skip the child operations, normally,  it's used after erase something containing region, so you don't want them to keep traversing the region (subtrees). So if you tend to iterate all operation, `return` is enough?

mlir/lib/QEC/Transforms/CountPPMSpecs.cpp
```diff
@@ -117,37 +120,111 @@ struct CountPPMSpecsPass : public impl::CountPPMSpecsPassBase<CountPPMSpecsPass>
...
+        if (currentLayer.empty())
+            return true;
+
+        auto lastOp = currentLayer.getOps().back();
+        return equalTypes(lastOp, op) && commuteToLayer(op, currentLayer);
+    }
+
+    void countDepths(std::vector<QECLayer> &layers,
+                     llvm::DenseMap<StringRef, llvm::DenseMap<StringRef, int>> *PPMSpecs,
+                     llvm::BumpPtrAllocator *stringAllocator)
```
mlxd: Any reason these are raw pointers instead of references?

sengthai: @mlxd No strong reason here. Thanks for pointing out the right spot! Per LLVM style, use references when a parameter must be non-null and there's no ownership transfer. I will refactor this.

mlir/lib/QEC/Transforms/CountPPMSpecs.cpp
sengthai: Thanks @rniczh! Make sense! Using just `return` is should not work in this case. I think `return WalkResult::advance()` should be good for this case.

mlir/lib/QEC/Transforms/CountPPMSpecs.cpp
```diff
@@ -117,37 +120,111 @@ struct CountPPMSpecsPass : public impl::CountPPMSpecsPassBase<CountPPMSpecsPass>
...
+        // Add the last layer if it is not empty.
+        if (!currentLayer.empty()) {
+            layers.emplace_back(std::move(currentLayer));
+        }
+
+        countDepths(layers, &PPMSpecs, &stringAllocator);
+
         if (wr.wasInterrupted()) {
             return failure();
         }
```
rniczh: Since the `walk` process has been interrupted, the logic from lines L221 to L226 appears to be redundant? Would it better to move L228 to L220?

mlir/lib/QEC/Transforms/CountPPMSpecs.cpp
sengthai: I made a change here [82bb1c6](https://github.com/PennyLaneAI/catalyst/pull/2014/commits/82bb1c6bd979b44ba722ee40c9c858632b9adabf)

mlir/lib/QEC/Transforms/CountPPMSpecs.cpp
sengthai: Make sense!

sengthai: [d6c6df2](https://github.com/PennyLaneAI/catalyst/pull/2014/commits/d6c6df2241dc818b8385140669f11c56a72f8e4a)

mlir/lib/QEC/Transforms/CountPPMSpecs.cpp
```diff
@@ -109,49 +112,123 @@ struct CountPPMSpecsPass : public impl::CountPPMSpecsPassBase<CountPPMSpecsPass>
...
+        // Add the last layer if it is not empty.
+        if (!currentLayer.empty()) {
+            layers.emplace_back(std::move(currentLayer));
+        }
+
+        countDepths(layers, PPMSpecs, stringAllocator);
+
         json PPMSpecsJson = PPMSpecs;
         llvm::outs() << PPMSpecsJson.dump(4)
                      << "\n"; // dump(4) makes an indent with 4 spaces when printing JSON
```
rniczh, line 234: Just a quick check, are you intent to always print json if running with depth analysis?

rniczh: I think `llvm::outs()` is not preferable if you considering in production-level, it's okay for local experiments. If you just want to have some debug logs and print on screen easily, how about using `LLVM_DEBUG(llvm::dbgs() << ... << "\n");`

sengthai: > Just a quick check, are you intent to always print json if running with depth analysis?

Yes, we want to print out json when running this.

> I think llvm::outs() is not preferable if you considering in production-level, it's okay for local experiments. If you just want to have some debug logs and print on screen easily, how about using LLVM_DEBUG(llvm::dbgs() << ... << "\n");

I have also tried that before, but it didn't display the result on the screen. 🤔
`LLVM_DEBUG(llvm::dbgs() << PPMSpecsJson.dump(4) << "\n");`

sengthai: It would be better to make this an analysis pass. Currently, this pass is like the transformation pass. I think it is fine for now.

rniczh: Discussed with @sengthai offline. Using `llvm::outs` here is okay for now, but will be changed to AnalysisPass in the future.

PR #2017, opened 2025-09-02
Fix result handling for one-shot

rniczh:
**Context:**

The result of non-mcm cases are not process properly. This PR is going to fix the result of non-mcm cases.

1. Fix normal result shape handling
2. Add support for CountsMP result shape handling
3. Add support for Shot vector result shape handling
4. Add support for snapshot result handling

**Test**

1. `test_mitigation.py` 
Lit test of `test_mitigation.py` needs to be updated. The reason for this change is in the handling of the result, where the `is_valid` parameter provided to gather_non_mcm has been changed from:
```python
is_valid = jnp.array([True] * len(out[i]))
```
to:
```python
is_valid = jnp.full((result.shape[0],), True)
```
Originally, this required passing `%c = stablehlo.constant dense<true> : tensor<{shots}xi1>` to the one-shot wrapper. This modification eliminates the need to pass that extra parameter. Instead, the logic is handled within the body by using broadcast:
```mlir
%c = stablehlo.constant dense<true> : tensor<i1>
%1 = stablehlo.broadcast_in_dim %c, dims = [] : (tensor<i1>) -> tensor<{shots}xi1>
```

2. `test_mid_circuit_measurement.py`

`value_and_grad` now will be supported (fix the result handling), but jax has an issue with the current version we use as it results in a random memory error. https://github.com/tensorflow/tensorflow/pull/97681 will be available in the next jax release. we will re-enable this test when the jax release is available and tested on our end.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-93649]


frontend/test/pytest/test_mid_circuit_measurement.py
```diff
@@ -809,37 +809,41 @@ def grad_g(x):
...
-        assert np.allclose(result1, result2)
+    # value_and_grad now will be supported, but jax has an issue with the current version we use
+    # as it results in a random memory error. https://github.com/tensorflow/tensorflow/pull/97681
+    # It will be fixed in the next jax release. we will re-enable this test when the jax
+    # release is available and tested on our end.
+    # @pytest.mark.xfail(
+    #     reason="value_and_grad with dynamic one-shot is not yet supported.",
+    # )
+    # def test_mcm_method_with_value_and_grad(self):
+    #     """Test that the dynamic_one_shot works with value_and_grad."""
```
paul0403: Let's just update the xfail message instead of commenting out the entire test!

frontend/test/pytest/test_measurements_results.py
```diff
@@ -151,6 +151,43 @@ def counts_2qbit(x: float):
...
+            qml.RX(x, wires=0)
+            return qml.counts()
+
+        expected = [np.array([0, 1]), np.array([1000, 0])]
+        observed = counts_1qbit(0.0)
+        assert np.array_equal(observed, expected)
+
+        expected = [np.array([0, 1]), np.array([0, 1000])]
+        observed = counts_1qbit(np.pi)
+        assert np.array_equal(observed, expected)
```
paul0403:
```suggestion
        expected = [np.array([0, 1]), np.array([1000, 0])]
        observed = counts_1qbit(0.0)
        assert np.array_equal(observed, expected)
```

No need to duplicate tests like this : )

frontend/test/pytest/test_measurements_results.py
```diff
@@ -151,6 +151,43 @@ def counts_2qbit(x: float):
...
+            qml.RY(x, wires=1)
+            return qml.counts()
+
+        expected = [np.array([0, 1, 2, 3]), np.array([1000, 0, 0, 0])]
+        observed = counts_2qbit(0.0)
+        assert np.array_equal(observed, expected)
+
+        expected = [np.array([0, 1, 2, 3]), np.array([0, 0, 0, 1000])]
+        observed = counts_2qbit(np.pi)
+        assert np.array_equal(observed, expected)
```
paul0403:
```suggestion
        expected = [np.array([0, 1, 2, 3]), np.array([1000, 0, 0, 0])]
        observed = counts_2qbit(0.0)
        assert np.array_equal(observed, expected)
```

frontend/test/pytest/test_pytree_args.py
```diff
@@ -344,6 +344,115 @@ def workflow1(param):
         assert isinstance(result, dict)
         assert result["w"] == 1
 
+    def test_return_value_dict_one_shot(self, backend, tol_stochastic):
+        """Test dictionaries."""
+
+        @qml.qnode(qml.device(backend, wires=2))
```
paul0403: Why did this particular test not need `mcm_method="one-shot"`? :thinking:

paul0403: Actually, aren't these tests duplicating the `test_return_value_dict` tests in this file already?

frontend/test/pytest/test_measurements_results.py
paul0403: I just realized these tests are also duplicating already existing tests. Did you forget to delete them from developing? Or do they have a separate purpose?

frontend/test/pytest/test_jit_behaviour.py
```diff
@@ -42,7 +42,7 @@ def f_aot_builder(backend, wires=1, shots=1000):
 
     @qjit
     @qml.set_shots(shots)
-    @qml.qnode(qml.device(backend, wires=wires))
```
paul0403: Why was this change necessary? Does this no longer work with single branch statistics (current default)?

frontend/test/pytest/test_shotvector.py
```diff
@@ -69,14 +69,42 @@ def circuit_dict():
         assert jnp.array(circuit_dict()["first"]).shape == (4, 3, 1)
         assert jnp.array(circuit_dict()["second"]).shape == (4, 3, 1)
 
+    @pytest.mark.parametrize("shots", [((3, 4),), (3,) * 4, (3, 3, 3, 3), [3, 3, 3, 3]])
+    def test_multiple_sample_measurement_one_shot(self, shots):
```
paul0403: Here too. 

If you just want to test the one-shot implementation you added on the existing tests (which are still using single branch statistics as default), you can do with pytest parametrization on the mcm method : )

```python
@pytest.mark.parametrize("mcm_method", ["single-branch-statistics", "one-shot"])
def test(mcm_method):
   ...
   @qml.qnode(..., mcm_method=mcm_method)

```

frontend/catalyst/qfunc.py
```diff
@@ -177,6 +177,120 @@ def _eval_quantum(*args, **kwargs):
...
+                                otherwise, returns the original output.
+        measurement_results: The corresponding measurement results.
+    """
+    # if no snapshot are present, return None, out
+    if not any(isinstance(op, qml.Snapshot) for op in tape.operations):
+        return None, out
+
+    # Snapshots present: out[0] = snapshots, out[1] = measurements
+    assert len(out) == 2
+    snapshot_results, measurement_results = out
```
paul0403: I guess the unpacking syntax on the second line makes the first assert redundant, but I'm also not against having it. I'm good with whichever you prefer :+1:

frontend/catalyst/qfunc.py
```diff
@@ -177,6 +177,120 @@ def _eval_quantum(*args, **kwargs):
...
+
+    # Take first shot for each snapshot
+    processed_snapshots = [
+        snapshot[0] if hasattr(snapshot, "shape") and len(snapshot.shape) > 1 else snapshot
+        for snapshot in snapshot_results
+    ]
```
paul0403, line 214: Will this logic fail for all options other than one-shot?

frontend/catalyst/qfunc.py
```diff
@@ -177,6 +177,120 @@ def _eval_quantum(*args, **kwargs):
...
+        aggregated_counts = jnp.sum(counts_array, axis=0)
+        counts_result = (keys, aggregated_counts)
+    else:
+        aggregated_counts = jnp.sum(counts, axis=0)
+        counts_result = (keys[0], aggregated_counts)
+
+    return counts_result, idx
+
+
+def _process_regular_measurement(m, out, idx, shot_vector):
```
paul0403:
```suggestion
def _process_non_counts_measurement(m, out, idx, shot_vector):
```

paul0403: In fact, can the two functions above be merged into one, just called `_process_terminal_measurement`, and have an if block for counts?

paul0403 (commented):
Thanks! :100: No major changes needed from my side, just some small points

frontend/catalyst/jax_tracer.py
```diff
@@ -1205,7 +1205,12 @@ def apply_transforms(
         # TODO: Ideally we should allow qnode transforms that don't modify the measurements to
         # operate in the permissive tracing mode, but that currently leads to a small number of
         # test failures due to the different result format produced in trace_quantum_function.
-        if has_classical_outputs(flat_results):
+        only_with_dynamic_one_shot = all(
```
paul0403, line 1208: Remember to add a changelog!

I think this would be improvements? So it would be something like 
```
Improvements Section

- The default mid-circuit measurement method in catalyst has been changed from `"single-branch-statistics"` to `"one-shot"`. 
   # and add the list of PRs
```

frontend/test/pytest/test_mid_circuit_measurement.py
rniczh: The reason of commenting out the test is it cannot trigger the xfail anymore, it ran into a memory error.

frontend/test/pytest/test_shotvector.py
rniczh: Good point! I added `pytest.mark.parametrize` for `mcm_method` to all the others that do the same.

frontend/catalyst/qfunc.py
rniczh: Yes, it's only works when one-shot, and more specifically, it only for non-mcm case with one-shot

frontend/catalyst/qfunc.py
rniczh: I removed the assert

frontend/catalyst/qfunc.py
rniczh: Done

frontend/test/pytest/test_jit_behaviour.py
rniczh: fixed, same as https://github.com/PennyLaneAI/catalyst/pull/2017#discussion_r2320643515

frontend/test/pytest/test_pytree_args.py
rniczh: fixed, same as https://github.com/PennyLaneAI/catalyst/pull/2017#discussion_r2320643515

frontend/test/pytest/test_measurements_results.py
rniczh: fixed, same as https://github.com/PennyLaneAI/catalyst/pull/2017#discussion_r2320643515

frontend/test/pytest/test_measurements_results.py
rniczh: fixed, same as https://github.com/PennyLaneAI/catalyst/pull/2017#discussion_r2320643515

frontend/catalyst/jax_tracer.py
rniczh: It's not yet changed from `single-branch` to `one-shot`, we stilling missing the fallback mechanism here, which is in another PR https://github.com/PennyLaneAI/catalyst/pull/2019

I updated the changelog to describe what this PR did:
> Add support for one-shot result handling in non-mcm cases, allowing the following items to continue functioning: counts measurement result, shot vector result, snapshot result. And also optimize the arguments passing to one-shot wrapper in MLIR as well.

rniczh: Oh! I know what we mean, that make sense. I updated it, also include the PR #2019 as well.

frontend/test/pytest/test_mid_circuit_measurement.py
paul0403: I see, thanks for explaining. In that case we can use pytest.skip I think.

doc/releases/changelog-dev.md
```diff
@@ -28,11 +28,15 @@
 
 <h3>Improvements 🛠</h3>
 
+* The default mid-circuit measurement method in catalyst has been changed from `"single-branch-statistics"` to `"one-shot"`.
+  [[#2017]](https://github.com/PennyLaneAI/catalyst/pull/2017)
+  [[#2019]](https://github.com/PennyLaneAI/catalyst/pull/2019)
```
paul0403: Let's add 2019 in 2019 itself : )

paul0403:
```suggestion
```

frontend/catalyst/qfunc.py
paul0403: Then, is there any sort of guard to check that these helpers are only reached when `mcm_methos="one-shot"`? I don't think I saw any :thinking:

frontend/test/pytest/test_jit_behaviour.py
```diff
@@ -37,25 +37,25 @@
 # pylint: disable=too-many-lines
 
 
-def f_aot_builder(backend, wires=1, shots=1000):
+def f_aot_builder(backend, mcm_method, wires=1, shots=1000):
```
paul0403: Why do the tests in this `test_jit_behavior.py` need to be changed at all? I don't think they have anything to do with mcm. How were they failing under your changes?

rniczh: It's only for codefactor coverage, but if it's doesn't matter, I could remove the mcm_method here

paul0403: How so? What lines were only reached by this test file? :open_mouth:

rniczh: Umm...I think cofactor is good now 🤔 I removed the `mcm_method` specified here

frontend/test/pytest/test_mid_circuit_measurement.py
rniczh: Done

frontend/catalyst/qfunc.py
rniczh: Added

frontend/catalyst/qfunc.py
```diff
@@ -177,6 +177,111 @@ def _eval_quantum(*args, **kwargs):
...
+    Args:
+        tape: The tape to get the snapshot results from.
+        out: The output of the tape.
+    Returns:
+        processed_snapshots: The extracted snapshot results if available;
+                                otherwise, returns the original output.
+        measurement_results: The corresponding measurement results.
+    """
+    # if no snapshot are present, return None, out
+    assert mcm_method == "one-shot"
```
paul0403, line 202: Final nitpick: let's add this protection to all the other one-shot specific helpers here you added. Then we are good to go!

frontend/test/pytest/test_mid_circuit_measurement.py
```diff
@@ -809,7 +809,11 @@ def grad_g(x):
 
         assert np.allclose(grad_f(1.0), grad_g(1.0))
 
-    @pytest.mark.xfail(
+    # value_and_grad now will be supported, but jax has an issue with the current version we use
+    # as it results in a random memory error. https://github.com/tensorflow/tensorflow/pull/97681
+    # It will be fixed in the next jax release. we will re-enable this test when the jax
+    # release is available and tested on our end.
+    @pytest.mark.skip(
         reason="value_and_grad with dynamic one-shot is not yet supported.",
```
paul0403:
```suggestion
    # value_and_grad now will be supported, but jax has an issue with the current version we use
    # as it results in a random memory error. https://github.com/tensorflow/tensorflow/pull/97681
    # It will be fixed in the next jax release. we will re-enable this test when the jax
    # release is available and tested on our end.
    @pytest.mark.skip(
        reason="https://github.com/tensorflow/tensorflow/pull/97681",
```

frontend/catalyst/qfunc.py
rniczh: Sure, I updated the helper functions that are one-shot specified

Issue #1949, opened 2025-07-30
Dynamic-one-shot is incompatible with automatic qubit management for some measurement processes


rniczh:
## Context

When using `@qjit` with finite shots, `qml.sample()` measurements fail with a shape canonicalization error involving `OutDBIdx` references, while `qml.expval()` measurements work correctly. This occurs in the `dynamic_one_shot` transformation that's automatically applied when using finite shots and in the case of `wires` is not set to `qml.qnode`.

### ❌ Failing Case (qml.sample())
```python
import pennylane as qml
from catalyst import qjit

backend = "lightning.qubit"

@qjit
@qml.qnode(qml.device(backend, shots=10), mcm_method='one-shot')
def circuit():
    qml.RX(0.0, wires=3)
    return qml.sample()

circuit()
```

### ✅ Working Case (qml.expval())
```python
@qjit
@qml.qnode(qml.device(backend, shots=10), mcm_method='one-shot')
def circuit():
    qml.RX(0.0, wires=3)
    return qml.expval(qml.PauliZ(0))  # This works

circuit()
```

### Why expval() works
- No `OutDBIdx` references are created, the shape for expval(...) case is `ShapedType` which can be handled by `jnp.zeros()`

## Full Stack Trace

```python
Traceback (most recent call last):
  File "/path/to/test.py", line 7, in <module>
    @qjit
     ^^^^
  File "/path/to/venv/catalyst/lib/python3.13/site-packages/pennylane/logging/decorators.py", line 61, in wrapper_entry
    return func(*args, **kwargs)
  File "/path/to/catalyst/frontend/catalyst/jit.py", line 502, in qjit
    return QJIT(fn, CompileOptions(**kwargs))
  File "/path/to/venv/catalyst/lib/python3.13/site-packages/pennylane/logging/decorators.py", line 65, in wrapper_exit
    output = func(*args, **kwargs)
  File "/path/to/catalyst/frontend/catalyst/jit.py", line 565, in __init__
    self.aot_compile()
    ~~~~~~~~~~~~~~~~^^
  File "/path/to/venv/catalyst/lib/python3.13/site-packages/pennylane/logging/decorators.py", line 61, in wrapper_entry
    return func(*args, **kwargs)
  File "/path/to/catalyst/frontend/catalyst/jit.py", line 618, in aot_compile
    self.jaxpr, self.out_type, self.out_treedef, self.c_sig = self.capture(
                                                              ~~~~~~~~~~~~^
[...]

PR #2019, opened 2025-09-03
Add constraints for one-shot and mechanism to fallback to the single-branch

rniczh:
**Context:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/frontend/catalyst/qfunc.py
+++ b/frontend/catalyst/qfunc.py
@@ -37,7 +37,9 @@ from pennylane.transforms.dynamic_one_shot import (
 import catalyst
 from catalyst.api_extensions import MidCircuitMeasure
 from catalyst.device import QJITDevice
+from catalyst.device.qjit_device import is_dynamic_wires
 from catalyst.jax_extras import deduce_avals, get_implicit_and_explicit_flat_args, unzip2
+from catalyst.jax_extras.tracing import uses_transform
 from catalyst.jax_primitives import quantum_kernel_p
 from catalyst.jax_tracer import Function, trace_quantum_function
 from catalyst.logging import debug_logger
@@ -59,7 +61,9 @@ def _resolve_mcm_config(mcm_config, shots):
     )
     if mcm_config.mcm_method is None:
         updated_values["mcm_method"] = (
-            "one-shot" if mcm_config.postselect_mode == "hw-like" else "single-branch-statistics"
+            "one-shot"
+            if mcm_config.postselect_mode == "hw-like" or (isinstance(shots, int) and shots > 0)
+            else "single-branch-statistics"
         )
     if mcm_config.mcm_method == "deferred":
         raise ValueError("mcm_method='deferred' is not supported with Catalyst.")
@@ -94,6 +98,73 @@ def _get_total_shots(qnode):
     return shots
 
 
+def configure_mcm_and_try_one_shot(qnode, args, kwargs):
+    """Configure mid-circuit measurement settings and handle one-shot execution."""
+    dynamic_one_shot_called = getattr(qnode, "_dynamic_one_shot_called", False)
+    if not dynamic_one_shot_called:
+        mcm_config = copy(
+            qml.devices.MCMConfig(
+                postselect_mode=qnode.execute_kwargs["postselect_mode"],
+                mcm_method=qnode.execute_kwargs["mcm_method"],
+            )
+        )
+        total_shots = _get_total_shots(qnode)
+        user_specified_mcm_method = mcm_config.mcm_method
+        mcm_config = _resolve_mcm_config(mcm_config, total_shots)
+
+        # Check if measurements_from_{samples/counts} is being used
+        uses_measurements_from_samples = uses_transform(qnode, "measurements_from_samples")
+        uses_measurements_from_counts = uses_transform(qnode, "measurements_from_counts")
+
+        # For cases that user are not tend to executed with one-shot, and facing measurement
+        # transform, fallback to single-branch-statistics
+        if (
+            (uses_measurements_from_samples or uses_measurements_from_counts)
+            and user_specified_mcm_method is None
+            and mcm_config.mcm_method == "one-shot"
+        ):
+            mcm_config = replace(mcm_config, mcm_method="single-branch-statistics")
+
+        if mcm_config.mcm_method == "one-shot":
+            # If measurements_from_samples/counts while one-shot is used, raise an error
+            if uses_measurements_from_samples:
+                raise CompileError("measurements_from_samples is not supported with one-shot")
+            if uses_measurements_from_counts:
+                raise CompileError("measurements_from_counts is not supported with one-shot")
+
+            mcm_config = replace(
+                mcm_config, postselect_mode=mcm_config.postselect_mode or "hw-like"
+            )
+
+            try:
+                return Function(dynamic_one_shot(qnode, mcm_config=mcm_config))(*args, **kwargs)
+            except (TypeError, ValueError, CompileError, NotImplementedError) as e:
+
+                # If user specified mcm_method, we can't fallback to single-branch-statistics,
+                # reraise the original error
+                if user_specified_mcm_method is not None:
+                    raise
+
+                # Fallback only if mcm was auto-determined
+                error_msg = str(e)
+                unsupported_measurement_error = any(
+                    pattern in error_msg
+                    for pattern in [
+                        "Native mid-circuit measurement mode does not support",
+                        "qml.var(obs) cannot be returned when `mcm_method='one-shot'`",
+                        "empty wires is not supported with dynamic wires in one-shot mode",
+                    ]
+                )
+
+                # Fallback if error is related to unsupported measurements
+                if unsupported_measurement_error:
+                    logger.debug("Fallback to single-branch-statistics: %s", e)
+                    mcm_config = replace(mcm_config, mcm_method="single-branch-statistics")
+                else:
+                    raise
+    return None
+
+
 class QFunc:
     """A device specific quantum function.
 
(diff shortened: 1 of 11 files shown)


frontend/catalyst/qfunc.py
```diff
@@ -94,6 +98,103 @@ def _get_total_shots(qnode):
...
+        return False
+
+    # Check if device has get_c_interface method returning "oqd"
+    if hasattr(qnode.device, "get_c_interface"):
+        try:
+            c_interface_name, _ = qnode.device.get_c_interface()
+            if c_interface_name == "oqd":
+                return False
+        except (AttributeError, TypeError, ValueError):
+            pass
```
rniczh: Seems like checking the device class name is far enough, I will remove the checking of `get_c_interface`

frontend/test/pytest/test_measurement_transforms.py
```diff
@@ -939,3 +940,39 @@ def circuit(a: float):
...
+    def test_measurements_from_samples_raises_error_with_one_shot(self):
+        """Test raise an error when measurements_from_samples is used with one-shot."""
+        device = qml.device("lightning.qubit", wires=2, shots=1000)
+
+        with pytest.raises(
+            CompileError, match="measurements_from_samples is not supported with one-shot"
+        ):
+
+            @qjit
+            @partial(measurements_from_samples, device_wires=device.wires)
```
dime10: Feel free to ignore, but since the tests are identical you could parametrize pytest over the transform function

frontend/test/lit/test_measurements.py
```diff
@@ -71,11 +71,13 @@ def sample2(x: float, y: float):
...
 
 
+# CHECK-LABEL: public @jit_sample3(
+# CHECK: func.func private @one_shot_wrapper({{%.+}}) -> tensor<1000x2xi64>
 # CHECK-LABEL: public @sample3(
 @qjit(target="mlir")
 @qml.set_shots(1000)
 @qml.qnode(qml.device("lightning.qubit", wires=2))
-# CHECK: [[shots:%.+]] = arith.constant 1000 : i64
+# CHECK: [[shots:%.+]] = arith.constant 1 : i64
```
dime10: > And this PR will now default to the one-shot whenever shots > 0

Does this PR switch to one-shot regardless of whether mcms are present in the program?

rniczh: Yes, for testing the coverage. I will make it only works for the case that mcms are present.

frontend/catalyst/qfunc.py
```diff
@@ -94,6 +98,103 @@ def _get_total_shots(qnode):
...
+        if (
+            (
+                uses_measurements_from_samples
+                or uses_measurements_from_counts
+                or not _is_one_shot_compatible_device(qnode)
+            )
+            and user_specified_mcm_method is None
+            and mcm_config.mcm_method == "one-shot"
+        ):
+            mcm_config = replace(mcm_config, mcm_method="single-branch-statistics")
```
paul0403: Nice :+1:

frontend/catalyst/qfunc.py
```diff
@@ -286,6 +376,54 @@ def _process_terminal_measurements(mcm_method, cpy_tape, out, snapshots, shot_ve
...
+            raise TypeError(
+                f"Native mid-circuit measurement mode does not support {type(m).__name__} "
+                "measurements."
+            )
+
+        # Check variance with observable
+        if isinstance(m, VarianceMP) and m.obs:
+            raise TypeError(
+                "qml.var(obs) cannot be returned when `mcm_method='one-shot'` because "
+                "the Catalyst compiler does not handle qml.sample(obs)."
```
paul0403:
```suggestion
                "the Catalyst compiler does not support qml.sample(obs)."
```

frontend/catalyst/qfunc.py
```diff
@@ -286,6 +376,54 @@ def _process_terminal_measurements(mcm_method, cpy_tape, out, snapshots, shot_ve
...
+            )
+
+        # Check if the measurement is supported with shot-vector
+        if has_shot_vector and not isinstance(m, SampleMP):
+            raise NotImplementedError(
+                f"Measurement {type(m).__name__} is not supported a shot-vector. "
```
paul0403:
```suggestion
                f"Measurement {type(m).__name__} does not support shot-vectors. "
```

frontend/catalyst/qfunc.py
```diff
@@ -286,6 +376,54 @@ def _process_terminal_measurements(mcm_method, cpy_tape, out, snapshots, shot_ve
...
+                "Use qml.sample() instead."
+            )
+
+        # Check dynamic wires with empty wires
+        if not has_wires and isinstance(m, (SampleMP, CountsMP)) and (m.wires.tolist() == []):
+            raise NotImplementedError(
+                f"Measurement {type(m).__name__} with empty wires is not supported with "
+                "dynamic wires in one-shot mode. Please specify constant wires on device."
```
paul0403:
```suggestion
                "dynamic wires in one-shot mode. Please specify a constant number of wires on the device."
```

paul0403 (commented):
Straightforward PR, thanks! :+1:

frontend/catalyst/qfunc.py
paul0403: Yes, we never exported a proper `"oqd.qubit"` since it was not end-to-end yet so we decided not to ship it.

Just checking the class name is enough, yes.

rniczh: Done 👍

frontend/test/lit/test_measurements.py
rniczh: @dime10 Now, the one-shot only working when mcms are present, these tests are reverted 👍

frontend/test/pytest/test_measurement_transforms.py
rniczh: @dime10  I merged these two tests and parametrized the pytest here.
```python3
@pytest.mark.parametrize(
        "transform_measurement", (measurements_from_samples, measurements_from_counts)
    )
@pytest.mark.parametrize("mcm_method", ("one-shot", "single-branch-statistics"))
def test_measurements_transform(
```

doc/releases/changelog-dev.md
```diff
@@ -30,14 +30,14 @@
 
 * Catalyst now supports returning classical and MCM values with the dynamic one-shot MCM method.
   [(#2001)](https://github.com/PennyLaneAI/catalyst/pull/2001)
```
```suggestion
  [(#2004)](https://github.com/PennyLaneAI/catalyst/pull/2004)
```

?

frontend/catalyst/qfunc.py
```diff
@@ -72,7 +74,9 @@ def _resolve_mcm_config(mcm_config, shots):
     )
     if mcm_config.mcm_method is None:
         updated_values["mcm_method"] = (
-            "one-shot" if mcm_config.postselect_mode == "hw-like" else "single-branch-statistics"
+            "one-shot"
+            if mcm_config.postselect_mode == "hw-like" or (isinstance(shots, int) and shots > 0)
+            else "single-branch-statistics"
```
paul0403: We can probably just have `"one-shot"` here since the original purpose here was to fall back to single branch, but after this PR we already do it in the `configure_mcm_and_try_one_shot`.

Also, this check disables dynamic shots. Which I guess runs into the "can't insert dynamic arrays into dynamic arrays" jax issue you mentioned? But in any case, I think just specifying one shot here if user doesn't specify any mcm method is what we want to do here.

doc/releases/changelog-dev.md
rniczh: 😮

frontend/catalyst/qfunc.py
rniczh: Are you mean enable the one-shot even shots is not set? Because if there is no shots, we can't use the dynamic one shot

paul0403: No, I mean when the shot value is dynamic:

```python
@qjit
def workflow(shots):
   @qml.set_shots(shots)
   @qml.qnode(...)
   def circuit():
      ...
```

paul0403: In these cases `shots` will be a tracer instead of an int.

rniczh: I just moved the logic to `configure_mcm_and_try_one_shot`, let me know if this address your concern here

frontend/catalyst/qfunc.py
```diff
@@ -37,9 +37,11 @@
 import catalyst
 from catalyst.api_extensions import MidCircuitMeasure
 from catalyst.device import QJITDevice
+from catalyst.device.qjit_device import is_dynamic_wires
 from catalyst.jax_extras import deduce_avals, get_implicit_and_explicit_flat_args, unzip2
+from catalyst.jax_extras.tracing import uses_transform
 from catalyst.jax_primitives import quantum_kernel_p
-from catalyst.jax_tracer import Function, trace_quantum_function
+from catalyst.jax_tracer import DynamicJaxprTracer, Function, trace_quantum_function
```
paul0403:
```suggestion
from catalyst.jax_tracer import Function, trace_quantum_function
```

frontend/catalyst/qfunc.py
```diff
@@ -107,6 +107,98 @@ def _get_total_shots(qnode):
     return shots
 
 
+def _is_one_shot_compatible_device(qnode):
+    device_name = qnode.device.name
+    exclude_devices = {"softwareq.qpp", "nvidia.custatevec", "nvidia.cutensornet"}
```
dime10, line 112: I don't think it's a good idea to just have a hardcoded list of devices in here 🤔

frontend/catalyst/qfunc.py
```diff
@@ -107,6 +107,98 @@ def _get_total_shots(qnode):
...
+        uses_measurements_from_counts = uses_transform(qnode, "measurements_from_counts")
+        has_finite_shots = isinstance(total_shots, int) and total_shots > 0
+
+        # For cases that user are not tend to executed with one-shot, and facing
+        # 1. non-one-shot compatible device,
+        # 2. non-finite shots,
+        # 3. measurement transform,
+        # fallback to single-branch-statistics
+        one_shot_compatible = _is_one_shot_compatible_device(qnode)
+        one_shot_compatible &= has_finite_shots
```
dime10, line 148: Don't we already raise an error for the non-finite shot case further up, why do we go to a fallback here?

frontend/catalyst/qfunc.py
```diff
@@ -107,6 +107,98 @@ def _get_total_shots(qnode):
...
+
+        if should_fallback:
+            mcm_config = replace(mcm_config, mcm_method="single-branch-statistics")
+
+        if mcm_config.mcm_method == "one-shot":
+            # If measurements_from_samples/counts while one-shot is used, raise an error
+            if uses_measurements_from_samples:
+                raise CompileError("measurements_from_samples is not supported with one-shot")
+            if uses_measurements_from_counts:
+                raise CompileError("measurements_from_counts is not supported with one-shot")
```
dime10, line 166: I think the logic is a bit convoluted between the errors and the fallbacks, and there does seem to be overlap on the conditions. What I would propose is simplify this function such unsupported cases always raise an error and let the user know if that case would be supported with single-branch.

dime10 (commented):
Just some comments after coming over here from issue #1949 (not directly related but something I was thinking about regarding the restrictions)

PR #2023, opened 2025-09-04
chore: update version of PL and lightning to latest

andrijapau:
Need to update required version of lightning after this [PR](https://github.com/PennyLaneAI/pennylane/pull/8190/) in PennyLane is merged. This is because https://github.com/PennyLaneAI/pennylane-lightning/pull/1246 is the fix.

[sc-98660]


.dep-versions
```diff
@@ -14,4 +14,4 @@ pennylane=0.43.0.dev45
 
 # For a custom LQ/LK version, update the package version here and at
 # 'doc/requirements.txt'
-lightning=0.43.0.dev24
+lightning=0.43.0.dev25
```
paul0403, line 17: remember to also bump this version after PL fix goes in :partying_face:

andrijapau: Yup, I'll use this PR as the vehicle for that change!

andrijapau: Does it need a changelog?

paul0403: No

.github/workflows/check-catalyst.yaml
andrijapau, line 1: Revert this before merging

andrijapau: Done!

PR #1798, opened 2025-06-09
Fix pass decorator order

paul0403:
**Context:**
When chaining the catalyst pass decorators, the order in the transform sequence is reversed than the order suggested by the python decorator application.  
```python
@qjit
@catalyst.passes.cancel_inverses
@catalyst.passes.merge_rotations
@qml.qnode(dev)
def circuit():
    ...

print(circuit.mlir)

###################
....
    module attributes {transform.with_named_sequence} {
      transform.named_sequence @__transform_main(%arg0: !transform.op<"builtin.module">) {
        %0 = transform.apply_registered_pass "remove-chained-self-inverse" to %arg0 : (!transform.op<"builtin.module">) -> !transform.op<"builtin.module">
        %1 = transform.apply_registered_pass "merge-rotations" to %0 : (!transform.op<"builtin.module">) -> !transform.op<"builtin.module">
        transform.yield 
      }
    }
....
```

**Description of the Change:**
Register the passes in the reverse order (with respect to its current buggy order)

**Benefits:**
Bug fix!

**Related GitHub Issues:** closes #1797 

[sc-92945]

Issue #1797, opened 2025-06-09
Pass decorators are applied in reverse order


paul0403:
When chaining the catalyst pass decorators, the order seems to be reversed

```python
dev = qml.device("lightning.qubit", wires=1)
@qjit
@catalyst.passes.cancel_inverses
@catalyst.passes.merge_rotations
@qml.qnode(dev)
def circuit():
    qml.RX(1.1, wires=0)
    qml.Hadamard(wires=0)
    qml.Hadamard(wires=0)
    qml.RX(2.2, wires=0)
    return qml.probs()

circuit()
print(circuit.mlir)

###################
....
    module attributes {transform.with_named_sequence} {
      transform.named_sequence @__transform_main(%arg0: !transform.op<"builtin.module">) {
        %0 = transform.apply_registered_pass "remove-chained-self-inverse" to %arg0 : (!transform.op<"builtin.module">) -> !transform.op<"builtin.module">
        %1 = transform.apply_registered_pass "merge-rotations" to %0 : (!transform.op<"builtin.module">) -> !transform.op<"builtin.module">
        transform.yield 
      }
    }
....
```
With python decorator's order, the merge rotation pass should come before the cancel inverse pass. However in the transform sequence the cancel inverse pass comes first.

PR #2027, opened 2025-09-04
Fix pass decorator ordering in from_plxpr

### LINKED ISSUE #2025: [BUG] Transform order is reversed with program capture enabled
When program capture is used, the top-most transform decorator gets applied first, when the expected behaviour is for the bottom-most decorator to be applied first. An example is shown below:

```python
qml.capture.enable()

dev = qml.device("lightning.qubit", wires=5)

@qml.qjit
@qml.transforms.merge_rotations
@qml.qnode(dev)
def circ():
    qml.RX(1.5, 0)
    qml.X(0)
    qml.X(0)
    qml.RX(1.5, 0)
    return qml.expval(qml.Z(0))
```
```pycon
>>> print(circ.mlir)
module @circ {
  func.func public @jit_circ() -> tensor<f64> attributes {llvm.emit_c_interface} {
    %0 = catalyst.launch_kernel @module_circ::@circ() : () -> tensor<f64>
    return %0 : tensor<f64>
  }
  module @module_circ {
    module attributes {transform.with_named_sequence} {
      transform.named_sequence @__transform_main(%arg0: !transform.op<"builtin.module">) {
        %0 = transform.apply_registered_pass "remove-chained-self-inverse" to %arg0 : (!transform.op<"builtin.module">) -> !transform.op<"builtin.module">
        %1 = transform.apply_registered_pass "merge-rotations" to %0 : (!transform.op<"builtin.module">) -> !transform.op<"builtin.module">
        transform.yield 
      }
    }
    func.func public @circ() -> tensor<f64> attributes {diff_method = "adjoint", llvm.linkage = #llvm.linkage<internal>, qnode} {
      %cst = arith.constant 1.500000e+00 : f64
      quantum.device shots(%c0_i64) ["/Users/mudit.pandey/.pyenv/versions/pennylane-xdsl/lib/python3.12/site-packages/pennylane_lightning/liblightning_qubit_catalyst.dylib", "LightningSimulator", "{'mcmc': False, 'num_burnin': 0, 'kernel_name': None}"]
      %0 = quantum.alloc( 5) : !quantum.reg
      %out_qubits = quantum.custom "RX"(%cst) %1 : !quantum.bit
      %out_qubits_0 = quantum.custom "PauliX"() %out_qubits : !quantum.bit
      %out_qubits_1 = quantum.custom "PauliX"() %out_qubits_0 : !quantum.bit
      %out_qubits_2 = quantum.custom "RX"(%cst) %out_qubits_1 : !quantum.bit
      %2 = quantum.namedobs %out_qubits_2[ PauliZ] : !quantum.obs
      %3 = quantum.expval %2 : f64
      %4 = quantum.insert %0[ 0], %out_qubits_2 : !quantum.reg, !quantum.bit
      quantum.device_release
    }
  }
  func.func @setup() {
    quantum.init
    return
  }
  func.func @teardown() {
    quantum.finalize
    return
  }
}
```

We can see in the printed MLIR that the cancel-inverses pass is to be applied first, but the expected first pass should be merge-rotations. This bug is specific to program capture. The order of passes is correct without program capture.

paul0403:
**Context:**
It was discovered a while ago that the pass decorator ordering was in reverse, and a fix was added (#1797 , #1798 )
However that fix was not applied to the plxpr path.

Implement pass decorator order fix to plxpr path.

**Benefits:**
Correctness of pass order.

**Related GitHub Issues:** closes #2025 

[sc-98689]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -135,7 +135,8 @@
   qubits on the `"null.qubit"` device has been fixed.
   [(#1926)](https://github.com/PennyLaneAI/catalyst/pull/1926)
 
-* Fix ordering of pass decorators when passes are applied with program capture enabled.
+* Stacked Python decorators for built-in Catalyst passes are now applied in the correct order when
+  program capture is enabled.
   [(#2027)](https://github.com/PennyLaneAI/catalyst/pull/2027)
 
 <h3>Internal changes ⚙️</h3>


frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -284,7 +284,7 @@ def wrapper(*args):
             return self.eval(final_jaxpr.jaxpr, final_jaxpr.consts, *non_const_args)
         else:
             # Apply the corresponding Catalyst pass counterpart
-            self._pass_pipeline.append(Pass(catalyst_pass_name))
+            self._pass_pipeline = [Pass(catalyst_pass_name)] + self._pass_pipeline
```
mlxd:
```suggestion
            self._pass_pipeline.insert(0, Pass(catalyst_pass_name))
```
?

paul0403: Good point :sweat_smile:

doc/releases/changelog-dev.md
```diff
@@ -135,6 +135,9 @@
   qubits on the `"null.qubit"` device has been fixed.
   [(#1926)](https://github.com/PennyLaneAI/catalyst/pull/1926)
 
+* Fix ordering of pass decorators when passes are applied with program capture enabled.
```
joeycarter: Just to match the wording in the v0.12 changelog for the related PR:

```suggestion
* Stacked Python decorators for built-in Catalyst passes are now applied in the correct order when
  program capture is enabled.
```

frontend/test/lit/test_from_plxpr.py
```diff
@@ -338,3 +338,27 @@ def circuit(w1: int, w2: int, w3: int):
...
 
 test_multi_qubit_gates_on_different_dynamic_wires()
+
+
+def test_pass_application():
+    """Application of pass decorators."""
+
+    dev = qml.device("null.qubit", wires=1)
+
+    qml.capture.enable()
```
joeycarter, line 348: Can we use the `use_capture` test fixture here? (If so, we can also remove the `qml.capture.disable()` at the end).

```suggestion
@pytest.mark.usefixtures("use_capture")
def test_pass_application():
    """Application of pass decorators."""

    dev = qml.device("null.qubit", wires=1)
```

joeycarter (commented):
Thanks @paul0403 for the quick fix! :rocket: 

I only had two small suggestions; happy to approve once they're resolved.

frontend/test/lit/test_from_plxpr.py
paul0403: This is lit test and doesn't see the fixtures

joeycarter: Ah, my mistake. Never mind then.

PR #2028, opened 2025-09-04
Specify the correct python version for llvm-lit in CI

mehrdad2m:
**Context:**
`llvm-lit` is a python with a shebang line and calling it without specifying the python version could fail the CI if it lacks the python version required by the shebang. 

specify the python version for `llvm-lit`

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


paul0403:
Hey, just wondering is this a test PR or is it actually supposed to be merged?

paul0403 (approved):
Nice! So we use the python package lit instead of the built llvm lit, neat :100: 

(No changelog needed for CI)

paul0403:
> > Nice! So we use the python package lit instead of the built llvm lit, neat 💯
> > (No changelog needed for CI)
> 
> I believe we are still using the llvm-lit, but invoking the python interpreter ourselves vs using the shebang in the llvm-lit file.

Ah yes, makes sense! Thanks for clarifying!

PR #2030, opened 2025-09-04
Fix path to the catalyst CLI tool installation directory in the docs

joeycarter:
The Catalyst Command Line Interface documentation incorrectly stated that the `catalyst` executable is available in the `catalyst/bin/` directory relative to the environment's installation directory when installed via pip. The documentation has been updated to point to the correct location, which is the `bin/` directory relative to the environment's installation directory.


paul0403 (approved):
Thanks!

PR #2031, opened 2025-09-04
Renaming function and keys in ppm specs

lazypanda10117:
**Context:** The function `get_ppm_specs` and its results' properties naming can be made clearer and more consistent with existing convention. 

**Description of the Change:** Renaming the function `get_ppm_specs` to `ppm_specs` to be consistent with the naming of other functions. Also, changed the results's properties from `num_logical_qubits` to `logical_qubits`, `num_pi8_gates` to `pi8_ppr`, `num_pi4_gates` to `pi4_ppr`, and `num_pi2_gates` to `pi2_ppr`.

**Benefits:** 

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-98501]


frontend/catalyst/passes/builtin_passes.py
```diff
@@ -1037,7 +1037,7 @@ def loop(i):
         {
             'circuit_0': {
                         'max_weight_pi2': 2,
-                        'num_logical_qubits': 2,
+                        'logical_qubits': 2,
                         'num_of_ppm': 44,
                         'num_pi2_gates': 16
```
mehrdad2m:
```suggestion
                        'pi2_ppr': 16
```

mlir/test/QEC/PPMSpecsTest.mlir
```diff
@@ -444,4 +444,4 @@ func.func public @while_error(%arg0: !quantum.bit, %b: i1) {
     }
 
     return
-}
+}
```
mehrdad2m: We usually have one empty line at the end of all files. There is a setting in vscode you can set to do this for you automatically
```suggestion
}

```

paul0403: So when github diff looks like the final line is text, there's actually already another empty final line. Github doesn't render empty lines in PR diff. 

When there's truly no final empty line github shows a stop sign symbol looking thingy
image

paul0403: Ah wait it's here, never mind 😅 I didn't click into the file diff

Issue #2021, opened 2025-09-03
Compilation error for OQC device


devx5f:
**Description:**
Using the OQC device is returning a compile error

**Code Snippet:**
```python
dev = qml.device("oqc.cloud", backend="lucy", shots=2012, wires=1)

@qjit
@qml.qnode(dev)
def circuit(x: float):
    qml.RX(x, wires=0)
    return qml.counts(wires=[0])
``` 

**Error:**
```
CompileError: Device at /opt/anaconda3/envs/test_bed/lib/python3.12/site-packages/catalyst/utils/../lib/librtd_oqc.so cannot be found!
```

<details>
<summary> Stack Trace </summary>

```
---------------------------------------------------------------------------
CompileError                              Traceback (most recent call last)
Cell In[5], line 3
      1 dev = qml.device("oqc.cloud", backend="lucy", shots=2012, wires=1)
----> 3 @qjit
      4 @qml.qnode(dev)
      5 def circuit(x: float):
      6     qml.RX(x, wires=0)
      7     return qml.counts(wires=[0, 1])

File /opt/anaconda3/envs/test_bed/lib/python3.12/site-packages/pennylane/logging/decorators.py:61, in log_string_debug_func.<locals>.wrapper_entry(*args, **kwargs)
     54     s_caller = "::L".join(
     55         [str(i) for i in inspect.getouterframes(inspect.currentframe(), 2)[1][1:3]]
     56     )
     57     lgr.debug(
     58         f"Calling {f_string} from {s_caller}",
     59         **_debug_log_kwargs,
     60     )
---> 61 return func(*args, **kwargs)

File /opt/anaconda3/envs/test_bed/lib/python3.12/site-packages/catalyst/jit.py:502, in qjit(fn, autograph, autograph_include, async_qnodes, target, keep_intermediate, verbose, logfile, pipelines, static_argnums, static_argnames, abstracted_axes, disable_assertions, seed, circuit_transform_pipeline, pass_plugins, dialect_plugins)
    499 if fn is None:
    500     return functools.partial(qjit, **kwargs)
--> 502 return QJIT(fn, CompileOptions(**kwargs))

File /opt/anaconda3/envs/test_bed/lib/python3.12/site-packages/pennylane/logging/decorators.py:65, in log_string_debug_func.<locals>.wrapper_exit(*args, **kwargs)
     63 @wraps(func)
     64 def wrapper_exit(*args, **kwargs):
[...]

PR #2032, opened 2025-09-04
Fix system extension used in OQC device

rniczh:
**Context:**

use `librtd_oqc.dylib` for Darwin instead of `*.so`

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
#2021 2021

### DIFF
--- a/frontend/catalyst/third_party/oqc/oqc_device.py
+++ b/frontend/catalyst/third_party/oqc/oqc_device.py
@@ -15,6 +15,7 @@
 """This module contains the OQC device."""
 
 import os
+import platform
 from typing import Optional
 
 from pennylane.devices import Device, ExecutionConfig
@@ -44,8 +45,9 @@ class OQCDevice(Device):
         the location to the shared object with the C/C++ device implementation.
         """
 
-        # TODO: Replace with the oqc shared library
-        return "oqc", get_lib_path("oqc_runtime", "OQC_LIB_DIR") + "/librtd_oqc.so"
+        system_extension = ".dylib" if platform.system() == "Darwin" else ".so"
+        lib_path = get_lib_path("oqc_runtime", "OQC_LIB_DIR") + "/librtd_oqc" + system_extension
+        return "oqc", lib_path
 
     def __init__(self, wires, backend, shots=1024, **kwargs):
         self._backend = backend


frontend/catalyst/third_party/oqc/oqc_device.py
```diff
@@ -15,6 +15,7 @@
 """This module contains the OQC device."""
 
 import os
+import platform
```
paul0403, line 18: Thanks! Remember to add a changelog entry under bug fix :+1:

frontend/catalyst/third_party/oqc/src/OQCRunner.hpp
```diff
@@ -35,10 +35,12 @@ struct OQCRunner {
         DynamicLibraryLoader libLoader(OQC_PY);
 
         using countsImpl_t =
-            std::vector<size_t> (*)(const char *, const char *, size_t, const char *);
+            void (*)(const char *, const char *, size_t, size_t, const char *, void *);
```
paul0403: Let's add a test in 
https://github.com/PennyLaneAI/catalyst/blob/main/frontend/catalyst/third_party/oqc/src/tests/Test_OQCDevice.cpp
that actually uses counts!

doc/releases/changelog-dev.md
```diff
@@ -139,6 +139,9 @@
   program capture is enabled.
   [(#2027)](https://github.com/PennyLaneAI/catalyst/pull/2027)
 
+* Fix usage of OQC in macOS, include fixing system extension, and fixing wrong type signature of counts
```
```suggestion
* Fix usage of OQC device, including:
   - fix object file system extension on macOS
   - fix wrong type signature of `Counts` API function
```

frontend/catalyst/third_party/oqc/src/tests/Test_OQCDevice.cpp
```diff
@@ -69,3 +69,18 @@ TEST_CASE("Test the bell pair circuit", "[openqasm]")
...
+    auto wires = device->AllocateQubits(2);
+
+    device->NamedOperation("Hadamard", {}, {wires[0]}, false);
+
+    std::vector<double> eigvals(4);
+    std::vector<int64_t> counts(4);
+    DataView<double, 1> eigvals_view(eigvals);
+    DataView<int64_t, 1> counts_view(counts);
+
+    REQUIRE_NOTHROW(device->PartialCounts(eigvals_view, counts_view, {wires[0], wires[1]}));
```
paul0403: Nice! 

Just checked, the [`OQCDevice::PartialCounts`](https://github.com/PennyLaneAI/catalyst/blob/main/frontend/catalyst/third_party/oqc/src/OQCDevice.cpp#L65) (thing called here , aka the capi call from llvm ir) will call into [`OQCRunner::Counts`](https://github.com/PennyLaneAI/catalyst/blob/main/frontend/catalyst/third_party/oqc/src/OQCRunner.hpp#L31), which dlsyms the lowercase `counts` that goes into [the oqc python binding module](https://github.com/PennyLaneAI/catalyst/blob/main/frontend/catalyst/third_party/oqc/src/oqc_python_module.cpp#L49) and sends a request to the cloud oqc device 👍 💯 🏆

paul0403 (approved):

frontend/catalyst/third_party/oqc/src/tests/Test_OQCDevice.cpp
```diff
@@ -69,3 +69,18 @@ TEST_CASE("Test the bell pair circuit", "[openqasm]")
 
     CHECK(device->Circuit() == toqasmempty);
 }
+
+TEST_CASE("Test counts", "[openqasm][counts]")
```
paul0403, line 73: Are all the extra changes you are making just trying to make the test pass?

We don't need to strictly no throw, we can also just test that we get the expected "no oqc email" error message. The purpose is to just check the counts function signatures right?

rniczh: Not entirely. Some are necessary, but what's clear now is that `RT_FAIL_IF(!msg.empty(), msg.c_str());` is the main cause of the final test failure.

paul0403: How so? What's the message?

rniczh: Just segfault

rniczh: I'll give it a try to raise an error for "no oqc email" and show the message and still pass the test. Is it make sense to you?

paul0403: Yes I think this is the best way to proceed here. Probably segfault was on some empty string business...

rniczh: Now, it will show the messasge:
```
/home/runner/work/catalyst/catalyst/frontend/catalyst/third_party/oqc/src//build/tests/runner_tests_oqc
[OQC INFO] OQC credentials not configured. No email, password, or url found in environment variables.
```

paul0403: Yes, checking against this message is all we want here, since we just want to make sure our `counts` signatures are good. We don't really care about oqc's underlying execution here.

paul0403: Let me take a final look

frontend/catalyst/third_party/oqc/src/oqc_python_module.cpp
```diff
@@ -22,30 +22,31 @@
...
     oqc_task = QPUTask(circuit, oqc_config)
     res = client.execute_tasks(oqc_task)
     counts = res[0].result["cbits"]
-
 except Exception as e:
     print(f"circuit: {circuit}")
     msg = str(e)
 )";
 
+extern "C" {
```
paul0403, line 52: Why extern C here? :thinking:

frontend/catalyst/third_party/oqc/src/oqc_python_module.cpp
```diff
@@ -59,15 +60,23 @@ except Exception as e:
...
     // Evaluate in scope of main module
     nb::object scope = nb::module_::import_("__main__").attr("__dict__");
     nb::exec(nb::str(program.c_str()), scope, locals);
 
     auto msg = nb::cast<std::string>(locals["msg"]);
+    if (msg == "OQC credentials not found in environment variables") {
+        std::cout << "[OQC INFO] OQC credentials not configured. No email, password, or url found "
+                     "in environment variables.\n";
+        return;
+    }
```
paul0403: Why can't we use `RT_FAIL_IF`?

paul0403: You can check error messages in cpp tests with 
```
    REQUIRE_THROWS_WITH(
        ...bad code...,
        ContainsSubstring("Error message"));
```

frontend/catalyst/third_party/oqc/src/oqc_python_module.cpp
rniczh: otherwise, it will still report the symbol undefined

paul0403: Ah, so this is to force the name to be exactly `counts`? i.e. prevent cpp name mangling?

(which I think is what's supposed to happen... we use extern C for the standard runtime capi too. Maybe this has been a bug since always...)

frontend/catalyst/third_party/oqc/src/oqc_python_module.cpp
```diff
@@ -22,30 +22,31 @@
...
-
-RES_FORMAT = QuantumResultsFormat().binary_count()
 
 try:
+    import os
     email = os.environ.get("OQC_EMAIL")
     password = os.environ.get("OQC_PASSWORD")
     url = os.environ.get("OQC_URL")
+    if not all([email, password, url]):
+        raise ValueError("OQC credentials not found in environment variables")
```
paul0403: For error messages, the rule of thumb is suggest user actions whenever possible! 
```suggestion
        raise ValueError("OQC credentials not found in environment variables. Please set the environment variables `OQC_EMAIL`, `OQC_PASSWORD` and `OQC_URL`.")
```

frontend/catalyst/third_party/oqc/src/oqc_python_module.cpp
rniczh: Yes, exactly

frontend/catalyst/third_party/oqc/src/oqc_python_module.cpp
rniczh: Ohh..because extern "C" is declared, we cannot throw an error now...

frontend/catalyst/third_party/oqc/src/tests/CMakeLists.txt
```diff
@@ -32,16 +32,16 @@ target_include_directories(runner_tests_oqc PRIVATE
...
-# Locate PyBind11
+# Locate nanobind
 execute_process(
-    COMMAND "${Python_EXECUTABLE}" -c "import pybind11; print(pybind11.get_cmake_dir())"
-    OUTPUT_VARIABLE pybind11_DIR OUTPUT_STRIP_TRAILING_WHITESPACE
+    COMMAND "${Python_EXECUTABLE}" -c "import nanobind; print(nanobind.cmake_dir())"
+    OUTPUT_VARIABLE nanobind_DIR OUTPUT_STRIP_TRAILING_WHITESPACE
 )
-find_package(pybind11 CONFIG REQUIRED)
+find_package(nanobind CONFIG REQUIRED)
```
mlxd: Do we need to use nanobind here? Wouldn't pybind11 be a better fit, given the need for the embedding.

Just for info, trying to avoid using the CPython APIs throughout is preferred. If we can keep using pybind11 here, that is preferable.

paul0403: Done, it is pybind now.

frontend/catalyst/third_party/oqc/src/CMakeLists.txt
```diff
@@ -27,17 +23,18 @@ find_package(Python REQUIRED
     OPTIONAL_COMPONENTS Development.Embed Development.SABIModule
 )
 
-# Locate nanobind
+# Locate pybind11
 execute_process(
-    COMMAND "${Python_EXECUTABLE}" -c "import nanobind; print(nanobind.cmake_dir())"
-    OUTPUT_VARIABLE nanobind_DIR OUTPUT_STRIP_TRAILING_WHITESPACE
+    COMMAND "${Python_EXECUTABLE}" -c "import pybind11; print(pybind11.get_cmake_dir())"
```
dime10, line 28: This will break in our stable_abi wheels no?

PR #2033, opened 2025-09-04
Extend `null.qubit` resource tracking feature to gather and report significantly more data

jzaia18:
**Context:**
PennyLane's `qml.specs()` does not support Catalyst `qjit`'d code. However, `null.qubit` resource tracking is an avenue to provide this support.

Significantly improves `null.qubit` resource tracking to provide all necessary information for feature parity with `qml.specs`.

Improves documentation within existing `NullQubit` class.

**Benefits:**
Allows a UI to exist that provides device level circuit specs from PennyLane. Significantly cleans up and robustifies existing resource tracking code.

**Possible Drawbacks:**

**Related GitHub Issues:**
[[sc-98157](https://app.shortcut.com/xanaduai/story/98157)]
Builds upon resource tracking feature introduced in #1619 
Changes to `qml.specs()` to enable an entrypoint into resource tracking are in https://github.com/PennyLaneAI/pennylane/pull/8202


runtime/lib/backend/null_qubit/NullQubit.hpp
```diff
@@ -29,31 +29,53 @@
...
 #include "QubitManager.hpp"
+#include "ResourceTracker.hpp"
 #include "Types.h"
 #include "Utils.hpp"
 
 namespace Catalyst::Runtime::Devices {
 
 /**
- * @brief struct API for a null backend quantum device.
+ * @brief A null quantum device implementation for testing and resource tracking purposes
```
paul0403:
```suggestion
 * @brief A null backend quantum device.
```

We should not restrict the purpose of null qubit in documentation

runtime/lib/backend/null_qubit/NullQubit.hpp
```diff
@@ -29,31 +29,53 @@
...
- * This device API contains several NULL methods:
- * - a set of methods to manage qubit allocations and deallocations, device shot
- *   noise, and quantum tape recording as well as reference values for the result
- *   data-type; these are used to implement Quantum Runtime (QR) instructions.
+ * This device API provides a complete null backend implementation that:
+ * - Manages qubit allocations and deallocations without actual quantum state
+ * - Supports device shot configuration
+ * - Implements all Quantum Runtime (QR) and Quantum Instruction Set (QIS) methods as no-ops
+ * - Optionally tracks resource usage including gate counts, wire usage, and circuit depth
+ * - Returns deterministic "ground state" results for all measurements and observations
```
paul0403:
```suggestion
 * - Returns mock results for all measurements and observations
```

runtime/lib/backend/null_qubit/NullQubit.hpp
```diff
@@ -29,31 +29,53 @@
...
  *
- * - a set of methods for quantum operations, observables, measurements, and gradient
- *   of the device; these are used to implement Quantum Instruction Set (QIS) instructions.
+ * The null device is particularly useful for:
+ * - Testing quantum program compilation and execution without quantum simulation overhead
+ * - Resource estimation and circuit analysis
```
paul0403:
```suggestion
 * - Resource estimation
```

there's no circuit analysis in runtime currently

runtime/lib/backend/null_qubit/NullQubit.hpp
```diff
@@ -29,31 +29,53 @@
...
  */
 struct NullQubit final : public Catalyst::Runtime::QuantumDevice {
     std::unordered_map<std::string, std::string> device_kwargs;
 
-    NullQubit(const std::string &kwargs = "{}", const std::string &resources_fname = "")
+    /**
+     * @brief Constructs a NullQubit device with optional configuration parameters
+     *
+     * Parses device configuration from JSON-like string format and sets up
+     * resource tracking options if specified. Supported parameters:
```
paul0403:
```suggestion
     * resource tracking options if specified. By default, resource tracking is turned off. Supported parameters:
```

runtime/lib/backend/null_qubit/NullQubit.hpp
```diff
@@ -29,31 +29,53 @@
...
+     * - "track_resources": Enable/disable resource tracking ("True"/"False")
+     * - "resources_fname": Static filename for resource output [requires resource tracking]
+     * - "compute_depth": Enable/disable circuit depth computation ("True"/"False") [requires
+     * resource tracking]
+     *
+     * @param kwargs JSON-like string containing device configuration parameters
```
paul0403: A small thing: we actually have some bugs in parsing this kwarg string and now we don't support nested JSON strings :sweat_smile: https://github.com/PennyLaneAI/catalyst/pull/1843
```suggestion
     * @param kwargs non-nested JSON-like string containing device configuration parameters
```

runtime/lib/backend/null_qubit/NullQubit.hpp
```diff
@@ -29,31 +29,53 @@
...
+     */
+    NullQubit(const std::string &kwargs = "{}")
     {
         this->device_kwargs = Catalyst::Runtime::parse_kwargs(kwargs);
         if (device_kwargs.find("track_resources") != device_kwargs.end()) {
-            track_resources_ = device_kwargs["track_resources"] == "True";
-            this->resources_fname_ = resources_fname;
+            this->track_resources_ = device_kwargs["track_resources"] == "True";
+        }
+        if (device_kwargs.find("resources_fname") != device_kwargs.end()) {
```
paul0403: Since `device_kwargs` is already a std map, can we just use `std::unordered_map::contains()` for these checks? A bit more readale imo.

Catalyst is now completely on cpp20, so this should be possible.

runtime/lib/backend/null_qubit/NullQubit.hpp
```diff
@@ -64,59 +86,31 @@ struct NullQubit final : public Catalyst::Runtime::QuantumDevice {
...
-        fwrite(resources.str().c_str(), 1, resources.str().size(), resources_file);
-
-        // Restore 2 special variables
-        resource_data_["num_qubits"] = num_qubits;
-        resource_data_["num_gates"] = num_gates;
-    }
-
-    /**
-     * @brief Allocate a "null" qubit.
+     * @brief Allocates a new "null" qubit and returns its identifier
```
paul0403:
```suggestion
     * @brief Allocates a new "null" qubit and returns its qubit ID
```

Recently we improved the quantum device interface's documentation quite drastically, so it would be better to use the same words as the ones over there, to minimize potential user confusion! 

We call these "qubit IDs".

runtime/lib/backend/null_qubit/NullQubit.hpp
```diff
@@ -64,59 +86,31 @@ struct NullQubit final : public Catalyst::Runtime::QuantumDevice {
...
      *
-     * @return `QubitIdType`
+     * @return QubitIdType The identifier of the newly allocated qubit
```
paul0403:
```suggestion
     * @return QubitIdType The qubit ID of the newly allocated qubit
```

runtime/lib/backend/null_qubit/ResourceTracker.hpp
```diff
@@ -0,0 +1,367 @@
...
+ * This class provides comprehensive tracking of quantum circuit resources during execution,
+ * including counting gate types and sizes, tracking maximum wire usage, and optionally
+ * computing circuit depth. It can export the collected data to JSON format for analysis.
+ */
+struct ResourceTracker final {
+  private:
+    std::unordered_map<std::string, std::size_t> gate_types_;
+    std::unordered_map<std::size_t, std::size_t> gate_sizes_;
+    std::unordered_map<QubitIdType, std::size_t> wire_depths;
+    QubitIdType max_num_wires_;
```
paul0403: Shouldn't the number of wires just be a regular integer type? 

`QubitIdType` is reserved for integers that should be interpreted (concretely, `reinterpret_cast`-ed) as qubit IDs.

paul0403: Also, in the current code, this is just `size_t`:

```C
std::unordered_map<std::string, std::size_t> resource_data_;
```

runtime/lib/backend/null_qubit/NullQubit.hpp
```diff
@@ -64,59 +86,31 @@ struct NullQubit final : public Catalyst::Runtime::QuantumDevice {
...
+            this->resource_tracker_.SetMaxWires(num_qubits_);
         }
         return this->qubit_manager.Allocate(num_qubits_);
     }
 
     /**
-     * @brief Allocate a vector of "null" qubits.
-     *
-     * @param num_qubits The number of qubits to allocate.
+     * @brief Allocates a vector of "null" qubits and returns their identifiers
```
paul0403:
```suggestion
     * @brief Allocates a vector of "null" qubits and returns their qubit IDs
```

paul0403: ...same with all other places where this is a thing

runtime/lib/backend/null_qubit/NullQubit.hpp
```diff
@@ -140,162 +138,150 @@ struct NullQubit final : public Catalyst::Runtime::QuantumDevice {
...
-                std::string err_msg =
-                    "Error opening file '" + this->resources_fname_ + "'."; // LCOV_EXCL_LINE
-                RT_FAIL(err_msg.c_str());                                   // LCOV_EXCL_LINE
-            }
-            else {
-                PrintResourceUsage(resources_file);
-                fclose(resources_file);
-            }
-            this->resource_data_.clear();
+            this->resource_tracker_.WriteOut();
```
paul0403, line 153: Does it make more sense for the file write to happen in the destructor instead of this release method? This `ReleaseAllQubits` is only at the end of a program now by accident, and not guaranteed, since the device implementer can do whatever they want internally for dynamic qubit allocation/deallocation handling?

paul0403: Or I guess the better question is, is the resource tracking per device instance (which is what implementing it on the null qubit class will lead to), or per qnode (which would be per user program)?

What happens when a qjitted function contains two qnodes on null.qubit?

runtime/lib/backend/null_qubit/ResourceTracker.hpp
```diff
@@ -0,0 +1,367 @@
...
+        else {
+            PrintResourceUsage(resources_file);
+            fclose(resources_file);
+        }
+
+        Reset();
+    }
+};
+
+} // namespace Catalyst::Runtime
```
paul0403: We leave a blank line at the end of all our files.

runtime/lib/backend/null_qubit/NullQubit.hpp
```diff
@@ -140,162 +138,150 @@ struct NullQubit final : public Catalyst::Runtime::QuantumDevice {
...
-     * the same instance state will predictable and reproducibly generate the same
-     * program results. It is also expected that the provided PRNG state is evolved
-     * sufficiently so that two device executions sharing the same instance do not produce
-     * identical results.
-     * The provided PRNG instance is not thread-locked, and devices wishing to share it
-     * across threads will need to provide their own thread-safety.
+     * The null device doesn't perform any quantum simulation, so PRNG configuration
+     * is not required. This method exists to satisfy the QuantumDevice interface.
+     * In real devices, this would configure the pseudo-random number generator used
+     * for sampling and measurements to enable reproducible results.
```
paul0403:
```suggestion
```

We don't need to talk about other devices in null qubit : ) The documentation is already on the base class (which is also rendered on the documentation websites)

runtime/lib/backend/null_qubit/NullQubit.hpp
```diff
@@ -140,162 +138,150 @@ struct NullQubit final : public Catalyst::Runtime::QuantumDevice {
...
-     * the device implementation.
+     * The null device doesn't maintain quantum state, so tape recording is not applicable.
+     * This method exists to satisfy the QuantumDevice interface for compatibility with
+     * caching and gradient computation systems.
      */
     void StopTapeRecording() {}
 
     /**
-     * @brief Doesn't prepare subsystems using the given ket vector in the computational basis.
+     * @brief No-op implementation for state preparation using a statevector
```
paul0403, line 207: Thanks for cleaning up these docstrings! Definitely better than `Doesn't do something` :laughing:

runtime/lib/backend/null_qubit/NullQubit.hpp
```diff
@@ -140,162 +138,150 @@ struct NullQubit final : public Catalyst::Runtime::QuantumDevice {
...
+     *
+     * @param basis_state The computational basis state (ignored)
+     * @param wires The qubits to prepare (ignored)
      */
     void SetBasisState(DataView<int8_t, 1> &, std::vector<QubitIdType> &) {}
 
     /**
-     * @brief Doesn't apply a single gate to the state vector of a device with its name if this is
-     * supported.
+     * @brief Records a named quantum operation for resource tracking (no-op for execution)
```
paul0403:
```suggestion
     * @brief No-op implementation for a named quantum operation
```

The "if resource tracking is enabled" is already covered by the paragraph below I'd say.

runtime/lib/backend/null_qubit/NullQubit.hpp
```diff
@@ -140,162 +138,150 @@ struct NullQubit final : public Catalyst::Runtime::QuantumDevice {
...
+     *
+     * If resource tracking is enabled, records the operation details including name,
+     * parameters, and wire usage. The actual quantum operation is not performed since
+     * this is a null device.
```
paul0403:
```suggestion
     * parameters, and wire usage.
```

runtime/lib/backend/null_qubit/NullQubit.hpp
```diff
@@ -140,162 +138,150 @@ struct NullQubit final : public Catalyst::Runtime::QuantumDevice {
...
+     * If resource tracking is enabled, records the operation as a general unitary matrix
+     * operation. The actual quantum operation is not performed since this is a null device.
      *
+     * @param matrix The unitary matrix defining the operation (ignored)
+     * @param wires The target qubits for the operation
+     * @param inverse Whether this is an adjoint (inverse) operation
+     * @param controlled_wires Control qubits for controlled operations
+     * @param controlled_values Control values for multi-controlled operations
      */
     void MatrixOperation(const std::vector<std::complex<double>> &,
```
paul0403:
```suggestion
    void MatrixOperation(const std::vector<std::complex<double>> &matrix,
```

runtime/lib/backend/null_qubit/NullQubit.hpp
paul0403: Same changes for other gate ops like matrix op.

runtime/lib/backend/null_qubit/NullQubit.hpp
paul0403: Same changes for other gate ops like matrix op.

runtime/tests/Test_NullQubit.cpp
```diff
@@ -930,91 +922,56 @@ TEST_CASE("Test NullQubit device resource tracking", "[NullQubit]")
...
-
-    std::cout << "Auto-created resources filename: " << dummy_resources_fname << std::endl;
-
-    CHECK(dummy_resources_fname.rfind("__pennylane_resources_data_") == 0);
-    CHECK(dummy_resources_fname.find(".json") != std::string::npos);
-
-    std::remove(
-        dummy_resources_fname.c_str()); // Remove the file automatically created by the device
-}
+}
```
paul0403: EOF line

runtime/lib/backend/null_qubit/ResourceTracker.hpp
```diff
@@ -0,0 +1,367 @@
...
+        // Sanity check that wire numbers make sense
+        size_t total_wires = wires.size() + controlled_wires.size();
+        std::vector<QubitIdType> combined_wires = {};
+        combined_wires.insert(combined_wires.end(), wires.begin(), wires.end());
+        combined_wires.insert(combined_wires.end(), controlled_wires.begin(),
+                              controlled_wires.end());
+        for (const auto &i : combined_wires) {
+            if (i >= max_num_wires_) {
+                RT_FAIL(("Wire index " + std::to_string(i) + " exceeds allocated wires").c_str());
+            }
```
paul0403: We have a `RT_FAIL_IF` macro

runtime/lib/backend/null_qubit/ResourceTracker.hpp
```diff
@@ -0,0 +1,367 @@
...
+     */
+    auto GetNumGates(const std::string &gate_name = "") -> std::size_t
+    {
+        if (gate_name != "") {
+            return gate_types_[gate_name];
+        }
+
+        size_t num_gates = 0;
+        for (const auto &[gate_type, count] : gate_types_) {
+            // TODO: Probably a nice functional way to do this
```
paul0403: No need, sure we can use things like std::accumulate, but then it needs a lambda and it is probably just looping behind the scenes. It's neither more readable nor more efficient.

runtime/lib/backend/null_qubit/NullQubit.hpp
```diff
@@ -396,26 +407,28 @@ struct NullQubit final : public Catalyst::Runtime::QuantumDevice {
...
-     * number of counts for each sample.
+     * @brief Generates measurement count statistics for ground state (all shots in state 0)
+     *
+     * Since the null device simulates the ground state |0⟩^⊗n, all measurements yield 0.
+     * The eigenvalues array is filled with integers 0, 1, ..., 2^num_qubits - 1,
+     * while the counts array has all shots in the first position (state 0) and zeros elsewhere.
      *
-     * Always fills eigenvalues with integers ranging from 0, 1, ..., 2**num_qubits
-     * Always sets the first element of counts to `shots` and fills the rest with 0.
+     * @param eigvals Array to fill with possible measurement outcomes (0 to 2^n - 1)
```
paul0403:
```suggestion
     * @param eigvals Array to fill with possible computational basis measurement outcomes (0 to 2^n - 1, inclusive)
```

runtime/tests/Test_NullQubit.cpp
```diff
@@ -1,4 +1,3 @@
-
```
paul0403, line 1: :laughing:

runtime/tests/Test_NullQubit.cpp
```diff
@@ -897,30 +896,23 @@ TEST_CASE("Test NullQubit device shots methods", "[NullQubit]")
...
-        FAIL("Failed to open resource usage file for writing."); // LCOV_EXCL_LINE
-    }
-
     std::unique_ptr<NullQubit> dummy = std::make_unique<NullQubit>();
     CHECK(dummy->IsTrackingResources() == false);
 
     std::unique_ptr<NullQubit> sim =
-        std::make_unique<NullQubit>("{'track_resources':True}", RESOURCES_FNAME);
+        std::make_unique<NullQubit>("{'track_resources':True, 'resources_fname':'" +
+                                    std::string(RESOURCES_FNAME) + ", 'compute_depth':True'}");
```
paul0403: Did the quotation marks get messed up here?
```suggestion
        std::make_unique<NullQubit>("{'track_resources':True, 'resources_fname':'" +
                                    std::string(RESOURCES_FNAME) + "', 'compute_depth':True}");
```

runtime/tests/Test_NullQubit.cpp
```diff
@@ -930,91 +922,56 @@ TEST_CASE("Test NullQubit device resource tracking", "[NullQubit]")
...
-    // Check that releasing resets
-    sim->ReleaseAllQubits();
     std::remove(RESOURCES_FNAME); // Remove the file automatically created by the device
-
-    CHECK(sim->ResourcesGetNumGates() == 0);
-    CHECK(sim->ResourcesGetNumQubits() == 0);
-    CHECK(sim->ResourcesGetFilename() == RESOURCES_FNAME);
-}
-
-TEST_CASE("Test resource tracking filename", "[NullQubit]")
```
paul0403, line 1014: Why did you remove this test?

runtime/tests/Test_ResourceTracker.cpp
```diff
@@ -0,0 +1,357 @@
...
+    tracker->PrintResourceUsage(resource_file_w);
+    fclose(resource_file_w);
+
+    // Open the file of resource data
+    std::ifstream resource_file_r(RESOURCES_FNAME);
+    CHECK(resource_file_r.is_open()); // fail-fast if file failed to create
+
+    std::string full_json;
+    std::string line;
+    // Check all fields have the correct value
```
paul0403, line 252: I'm actually not against checking the raw string in the collected resource file. The programmatic checking isn't really more readable imo. 

But I don't really have a preference, so if you want to leave it as is I'm also fine with that :+1:

paul0403: (In fact, catalyst uses lit tests very heavily, which is essentially just checking outputs as raw strings)

runtime/tests/Test_ResourceTracker.cpp
```diff
@@ -0,0 +1,357 @@
...
+    // Ensure all expected fields were present
+    CHECK(full_json.find("num_wires") != std::string::npos);
+    CHECK(full_json.find("num_gates") != std::string::npos);
+    CHECK(full_json.find("depth") != std::string::npos);
+    CHECK(full_json.find("gate_types") != std::string::npos);
+    CHECK(full_json.find("gate_sizes") != std::string::npos);
+    for (const auto &name : RESOURCE_NAMES) {
+        CHECK(full_json.find(name) != std::string::npos);
+    }
+}
```
paul0403: EOF line

runtime/lib/backend/null_qubit/NullQubit.hpp
jzaia18: Is it okay to give this parameter a name if it goes unused? That's not a linting issue?

runtime/tests/Test_NullQubit.cpp
jzaia18: They totally did haha, good catch

runtime/tests/Test_NullQubit.cpp
jzaia18: It's been moved to `Test_ResourceTracker.cpp`. That's where the actual semantics of the tracking is tested. In `Test_NullQubit.cpp` I only test that NQ correctly passes info to the tracker.

runtime/tests/Test_ResourceTracker.cpp
jzaia18: I like this slightly more since it allows us to be more flexible in the future if e.g. the order of params in the file changes, a new field is added, we have to support optional fields, etc.

runtime/tests/Test_ResourceTracker.cpp
jzaia18: VS Code moment :sob: my EMACS config would never betray me like this

jzaia18: VS Code preferences updated :saluting_face:

runtime/lib/backend/null_qubit/NullQubit.hpp
paul0403: Does `[[maybe unused]]` not work here?

jzaia18: It should. Is that preferable to not giving the parameter a name at all? I suppose with the name its more clear what the parameter represents

paul0403: I think the asymmetry looks weird, that some args of this function are named but others are not.

jzaia18: Valid, I'll add it

runtime/lib/backend/null_qubit/NullQubit.hpp
jzaia18: Back when we first implemented this feature we decided that doing it on release was the better way https://github.com/PennyLaneAI/catalyst/pull/1619#discussion_r2037862591 . Are there extra drawbacks/benefits that we hadn't considered at that time? CC @dime10

paul0403: Oh I remember now, there's an internal thing where catalyst runtime recycles device instances across qnodes when possible. Then it makes sense to have it on the release op :+1:

paul0403 (approved):
:100:

doc/releases/changelog-dev.md
```diff
@@ -28,6 +28,9 @@
 
 <h3>Improvements 🛠</h3>
 
+* Significantly improved resource tracking with `null.qubit`, allowing integration of `qml.specs()` with `qjit`'d code.
```
paul0403: Can you add a frontend test to show behavior with `qml.specs()`?

jzaia18: Currently writing up these integration tests, but I'm realizing that these can't pass CIs since they depend on a version of PennyLane that hasn't been merged yet. Could we implement these as a followup PR?

paul0403: A follow-up PR sounds good :+1:

paul0403: But then in the mean time, I think the changelog should be reworded to not mention `qml.specs()`, since we don't have tests for it yet!

paul0403: (We can mention qml.specs in the changelog in the actual PR that adds it)

runtime/lib/backend/null_qubit/ResourceTracker.hpp
```diff
@@ -0,0 +1,365 @@
...
+ *
+ * This class provides comprehensive tracking of quantum circuit resources during execution,
+ * including counting gate types and sizes, tracking maximum wire usage, and optionally
+ * computing circuit depth. It can export the collected data to JSON format for analysis.
+ */
+struct ResourceTracker final {
+  private:
+    std::unordered_map<std::string, std::size_t> gate_types_;
+    std::unordered_map<std::size_t, std::size_t> gate_sizes_;
+    std::unordered_map<QubitIdType, std::size_t> wire_depths;
```
josephleekl: why is this member not `wire_depths_`? 😅

runtime/lib/backend/null_qubit/ResourceTracker.hpp
```diff
@@ -0,0 +1,365 @@
...
+    std::size_t max_num_wires_;
+    bool static_fname_;
```
josephleekl: maybe call this `static_filename_`?

josephleekl: similar for `resources_filename_`?

runtime/lib/backend/null_qubit/ResourceTracker.hpp
```diff
@@ -0,0 +1,365 @@
...
+    void Operation(const std::string &name, const std::vector<QubitIdType> &wires,
+                   const std::vector<QubitIdType> &controlled_wires)
+    {
+        // Sanity check that wire numbers make sense
+        std::size_t total_wires = wires.size() + controlled_wires.size();
+        std::vector<QubitIdType> combined_wires = {};
+        combined_wires.insert(combined_wires.end(), wires.begin(), wires.end());
+        combined_wires.insert(combined_wires.end(), controlled_wires.begin(),
+                              controlled_wires.end());
+        for (const auto &i : combined_wires) {
```
josephleekl, line 62: I'm guessing PL already checks this, but if not, might want to also check that `wires` and `controlled_wires` have no overlap? (If we do add this, lightning has some checks that do this, just copy from there)

runtime/lib/backend/null_qubit/ResourceTracker.hpp
```diff
@@ -0,0 +1,365 @@
...
+        gate_types_[name]++;
+        gate_sizes_[total_wires]++;
+        if (compute_depth_) {
+            std::size_t max_depth = 0;
+            for (const auto &i : combined_wires) {
+                max_depth = std::max(max_depth, wire_depths[i]);
+            }
+            // All wires used in this operation must now have their depth set based on this max
+            max_depth++;
+            for (const auto &i : wires) {
```
josephleekl:
```suggestion
            for (const auto &i : combined_wires) {
```

runtime/lib/backend/null_qubit/ResourceTracker.hpp
```diff
@@ -0,0 +1,365 @@
...
+    }
+
+  public:
+    /**
+     * @brief Default constructor that initializes the ResourceTracker with default settings
+     *
+     * Initializes the tracker with depth computation disabled and no static filename set.
+     * Calls Reset() to ensure all tracking data structures are properly initialized.
+     */
+    ResourceTracker()
```
josephleekl:
```suggestion
    ResourceTracker() : compute_depth (false), static_filename_(false)
```

or you can just set the member variable default as false?

runtime/lib/backend/null_qubit/ResourceTracker.hpp
jzaia18: I would expect this to be checked earlier. But this doesn't actually break anything for resource tracking, just some of these metrics become nonsense with an illegal set of wires. E.g. `CNOT(0,0)` is still a 2-wire operation.

runtime/lib/backend/null_qubit/ResourceTracker.hpp
```diff
@@ -0,0 +1,365 @@
...
+     * @brief Internal method to record an operation being applied to the device
+     *
+     * Updates the gate type and size counts, and if depth tracking is enabled,
+     * updates the depth of the wires involved in the operation.
+     *
+     * @param name The name of the operation being applied
+     * @param wires The wires the operation is being applied to
+     * @param controlled_wires The control wires the operation is being applied to
+     */
+    void Operation(const std::string &name, const std::vector<QubitIdType> &wires,
```
josephleekl: Can we call this e.g. `RecordOperation`?

runtime/lib/backend/null_qubit/ResourceTracker.hpp
```diff
@@ -0,0 +1,365 @@
...
+     * @param controlled_wires The control wires for controlled operations (empty for
+     * non-controlled)
+     */
+    void MatrixOperation(bool inverse, const std::vector<QubitIdType> &wires,
+                         const std::vector<QubitIdType> &controlled_wires = {})
+    {
+        std::string op_name = "QubitUnitary";
+
+        if (!controlled_wires.empty()) {
+            op_name = "Controlled" + op_name;
```
josephleekl, line 269: and here we don't distinguish how many controlled wires there are, like in NamedOperation?

runtime/lib/backend/null_qubit/ResourceTracker.hpp
```diff
@@ -0,0 +1,365 @@
...
+     * @brief Prints resource usage statistics in JSON format to the specified file
+     *
+     * Outputs comprehensive resource tracking data including number of wires,
+     * total gate count, breakdown by gate types, breakdown by gate sizes,
+     * and circuit depth (if enabled) in JSON format.
+     *
+     * @param resources_file File pointer where JSON data will be written
+     * @throws Runtime error if file writing fails
+     */
+    void PrintResourceUsage(FILE *resources_file)
```
josephleekl:
```suggestion
    void PrintResourceUsageToFile(FILE *resources_file)
```

Maybe?

runtime/lib/backend/null_qubit/NullQubit.hpp
```diff
@@ -140,162 +139,147 @@ struct NullQubit final : public Catalyst::Runtime::QuantumDevice {
...
+     * parameters, and wire usage.
+     *
+     * @param name The name of the quantum operation (e.g., "PauliX", "CNOT", "RZ")
+     * @param params Parameters for parametric gates (ignored)
+     * @param wires The target qubits for the operation
+     * @param inverse Whether this is an adjoint (inverse) operation
+     * @param controlled_wires Control qubits for controlled operations
+     * @param controlled_values Control values for multi-controlled operations
      */
     void NamedOperation(const std::string &name, const std::vector<double> &params,
```
josephleekl:
```suggestion
    void NamedOperation(const std::string &name, [[maybe_unused]] const std::vector<double> &params,
```

does the compiler -Werror not complain without this? :O

runtime/lib/backend/null_qubit/ResourceTracker.hpp
paul0403: PL already checks this

runtime/lib/backend/null_qubit/NullQubit.hpp
```diff
@@ -140,162 +139,147 @@ struct NullQubit final : public Catalyst::Runtime::QuantumDevice {
...
+     * This method always returns 0.0 to satisfy the interface requirements.
+     *
+     * @param obs_id The type of observable (Identity, PauliX, PauliY, PauliZ, Hadamard, or
+     * Hermitian)
+     * @param matrix The matrix representation for Hermitian observables (ignored)
+     * @param wires The qubits the observable acts on (ignored)
+     * @return ObsIdType Always returns 0.0
      */
     auto Observable(ObsId, const std::vector<std::complex<double>> &,
                     const std::vector<QubitIdType> &) -> ObsIdType
```
josephleekl, line 286: for the return below, isn't `ObsIdType` alias for `intptr_t`? why are we returning 0.0? :O

runtime/lib/backend/null_qubit/ResourceTracker.hpp
paul0403: https://github.com/PennyLaneAI/catalyst/blob/main/runtime/lib/capi/RuntimeCAPI.cpp#L633 we also recently added these checks at runtime for cases where wires are dynamic and the frontend can't catch them #2006

runtime/lib/backend/null_qubit/NullQubit.hpp
jzaia18: Weirdly it does not

runtime/lib/backend/null_qubit/NullQubit.hpp
paul0403: I think since it's nullqubit we just didn't really care about the validity of whatever's returned :laughing: 

But yes if we can clean these up as we encounter them I'm all for it :raised_hands:

runtime/tests/Test_NullQubit.cpp
```diff
@@ -897,30 +896,23 @@ TEST_CASE("Test NullQubit device shots methods", "[NullQubit]")
...
 {
     // The name of the file where the resource usage data is stored
-    constexpr char RESOURCES_FNAME[] = "__pennylane_resources_data.json";
-
-    // Open a file for writing the resources JSON
-    FILE *resource_file_w = fopen(RESOURCES_FNAME, "wx");
-    if (resource_file_w == nullptr) {                            // LCOV_EXCL_LINE
-        FAIL("Failed to open resource usage file for writing."); // LCOV_EXCL_LINE
-    }
+    constexpr char RESOURCES_FILENAME[] = "__pennylane_resources_data.json";
```
josephleekl: can we have RESOURCES_FILENAME as std::string? I think ifstream can also take std:string as argument

runtime/lib/backend/null_qubit/ResourceTracker.hpp
jzaia18: Correct, that's just how things are for `ControlledQubitUnitary`

runtime/lib/backend/null_qubit/NullQubit.hpp
jzaia18: I also thought this was weird (that it didn't give a warning even)! It's from before my PR but I can change it to 0

runtime/tests/Test_ResourceTracker.cpp
```diff
@@ -0,0 +1,357 @@
...
+                                             "Adjoint(T)",
+                                             "C(S)",
+                                             "2C(S)",
+                                             "CNOT",
+                                             "Adjoint(ControlledQubitUnitary)",
+                                             "ControlledQubitUnitary",
+                                             "Adjoint(QubitUnitary)",
+                                             "QubitUnitary"};
+
+static void ApplyGates(std::unique_ptr<ResourceTracker> &tracker)
```
josephleekl: do we need a pointer to ResourcerTracker? can't we just use a reference?

runtime/lib/backend/null_qubit/NullQubit.hpp
```diff
@@ -29,31 +29,54 @@
...
+ * - Supports device shot configuration
+ * - Implements all Quantum Runtime (QR) and Quantum Instruction Set (QIS) methods as no-ops
+ * - Optionally tracks resource usage including gate counts, wire usage, and circuit depth
+ * - Returns mock results for all measurements and observations
  *
- * - a set of methods for quantum operations, observables, measurements, and gradient
- *   of the device; these are used to implement Quantum Instruction Set (QIS) instructions.
+ * The null device is particularly useful for:
+ * - Testing quantum program compilation and execution without quantum simulation overhead
+ * - Resource estimation
```
AntonNI8:
```suggestion
 * - Resource tracking & resource estimation
```

doc/releases/changelog-dev.md
```diff
@@ -28,6 +28,9 @@
 
 <h3>Improvements 🛠</h3>
 
+* Significantly improved resource tracking with `null.qubit`.
+  [[#2033]](https://github.com/PennyLaneAI/catalyst/pull/2033)
```
AntonNI8, line 34: Can we add a list of new things that we now track?

AntonNI8 (approved):
⚡

PR #2035, opened 2025-09-05
Remove `format` header from Runtime CAPI

joeycarter:
We accidentally left in a stray

```cpp
#include <format>
```

in the runtime CAPI in #2000, but it is not being used. This PR removes it.


paul0403:
Actually, since we just bumped to gcc 13 and we can now use the format header, why not actually make both error messages use it? :wink:

PR #2037, opened 2025-09-08
Fix incorrect target link

rniczh:
**Context:**

`target_link_libraries(oqc_python_module PRIVATE Python::Python)` causes

```
CMake Error at CMakeLists.txt:37 (target_link_libraries):
  Target "oqc_python_module" links to:

    Python::Python

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```

Remove `Python::Python`, use `pybind11::embed` instead

run `py::scoped_interpreter guard{};` to initialize python env

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
#2021

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -28,6 +28,9 @@
 
 <h3>Improvements 🛠</h3>
 
+* Improve the pass `--ppm-specs` to count the depth of PPRs and PPMs in the circuit.
+  [(#2014)](https://github.com/PennyLaneAI/catalyst/pull/2014)
+
 * The default mid-circuit measurement method in catalyst has been changed from `"single-branch-statistics"` to `"one-shot"`.
   [[#2017]](https://github.com/PennyLaneAI/catalyst/pull/2017)
 
@@ -250,6 +253,10 @@
   ```console
   $ catalyst --tool=opt --pass-pipeline="builtin.module(convert-quantum-to-llvm{use-array-backed-registers=true})" <input file>
   ```
+
+* Fix auxiliary qubit deallocation in `decompose-non-clifford-ppr` pass 
+  in the `clifford-corrected` method.
+  [(#2039)](https://github.com/PennyLaneAI/catalyst/pull/2039)
 
 <h3>Documentation 📝</h3>
 
--- a/mlir/lib/QEC/Transforms/CountPPMSpecs.cpp
+++ b/mlir/lib/QEC/Transforms/CountPPMSpecs.cpp
@@ -25,6 +25,9 @@
 
 #include "Catalyst/Utils/SCFUtils.h"
 #include "QEC/IR/QECDialect.h"
+#include "QEC/IR/QECOpInterfaces.h"
+#include "QEC/Utils/QECLayer.h"
+#include "QEC/Utils/QECOpUtils.h"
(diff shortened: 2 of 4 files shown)


frontend/catalyst/third_party/oqc/src/CMakeLists.txt
```diff
@@ -18,9 +18,35 @@ endif()
 # Locate Python
 # The optional component is only used for the C++ test suite (to spin up its own interpreter),
 # and requires libpython.so to be available on the system.
```
dime10: It looks like your making Embed a required component. On some systems, in particular the manylinux images, the embed component is not available by default. So we don't require it to be available for the building and distribution of wheels, but only for the C++ testing environment, which is what this comment is saying. 

Do you still think making Embed always a required component is necessary?

rniczh: No, just discussed with @mlxd and @paul0403. We will try to use `pybind11::embed` instead. For the `Python::Embed`, it may not work for the user, and also definitely not for CI now.

dime10: Good, I think that's how we handle embedded python interpreters in the runtime tests as well 👍

.github/workflows/check-catalyst.yaml
```diff
@@ -51,7 +51,7 @@ jobs:
       - name: Install dependencies
         run: |
           sudo apt-get update
-          sudo apt-get -y -q install ninja-build make cmake clang
+          sudo apt-get -y -q install ninja-build make cmake clang python3-dev libpython3-dev
```
dime10: I'm surprised if this is needed (since we've already been running [this setup]( https://github.com/PennyLaneAI/catalyst/blob/faee8105b95d191ad6da4c7d856a04f3995dab4b/runtime/tests/CMakeLists.txt#L60-L63)), but there is also no harm in it 👍

.github/workflows/check-catalyst.yaml
```diff
@@ -666,10 +666,15 @@ jobs:
         with:
           python-version: ${{ needs.constants.outputs.primary_python_version }}
 
+      - name: Verify Python development setup
+        run: |
+          python3 -c "import sysconfig; print('Python executable:', sysconfig.get_config_var('BINDIR')); print('Python include dir:', sysconfig.get_path('include')); print('Python lib dir:', sysconfig.get_config_var('LIBDIR'))"
+          echo "Python3_ROOT_DIR=$(python3 -c 'import sys; print(sys.prefix)')" >> $GITHUB_ENV
```
dime10: Is there something wrong with our configuration elsewhere? We shouldn't need such bespoke configurtion in scripts if the make and cmake configs are setup correctly 🤔

.github/workflows/check-catalyst.yaml
rniczh: No we don't need this, it's redundant, I'm going to clean it. It's only exist for debugging

.github/workflows/check-catalyst.yaml
rniczh: No, I'm going to clean it, it's only exist for debugging

PR #2038, opened 2025-09-08
Add functionality to obtain some specs/info on quantum operations

zsb-xqc:
**Context:**
The current `--ppm-specs` pass provides informative specs on a circuit related to ppr/ppm components, such as ppr operation count.  This work extends that to include non-ppr/ppm operations, such as standard gate count, making the specs more generic.  This activity was chosen as a first issue for @zsb-xqc, so it serves as a learning exercise.
[sc-98690]

This modification includes adding a new pass in the quantum dialect `--quantum-specs-info`, which is implemented in:
`QuantumSpecsInfo.cpp`
The test file is:
`QuantumSpecsInfoTest.mlir`

**Benefits:**
Provides general specs for all quantum operations.  As a pass, it can be added at various locations in the pipeline, to see how the compilation has adjusted the gate count.

**Possible Drawbacks:**
None, but there may be better ways to implement spec gathering.

**Related GitHub Issues:**
None

### DIFF
--- a/frontend/catalyst/compiler.py
+++ b/frontend/catalyst/compiler.py
@@ -449,7 +449,7 @@ class Compiler:
 
             if self.options.debug_compiler != True:
                 result = subprocess.run(cmd, check=True, capture_output=True, text=True)
-            else: # self.options.debug_compiler == True:
+            else:  # self.options.debug_compiler == True:
                 with subprocess.Popen(
                     cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True
                 ) as p:
@@ -480,7 +480,7 @@ class Compiler:
                         print(result.stdout.strip(), file=self.options.logfile)
                     if result.stderr:
                         print(result.stderr.strip(), file=self.options.logfile)
-                else: # self.options.debug_compiler == True:
+                else:  # self.options.debug_compiler == True:
                     if res_stdout:
                         print(res_stdout.strip(), file=self.options.logfile)
                     if res_stderr:


mlir/lib/QEC/Transforms/CountPPMSpecs.cpp
```diff
@@ -229,9 +233,18 @@ struct CountPPMSpecsPass : public impl::CountPPMSpecsPassBase<CountPPMSpecsPass>
 
         countDepths(layers, PPMSpecs, stringAllocator);
 
+        std::error_code EC;
+        llvm::raw_fd_ostream fileOutputStream("test.txt", EC, llvm::sys::fs::OF_Append);
+        if (EC) {
+            llvm::errs() << "Error opening file: " << EC.message() << "\n";
+            return failure(); // Handle error
+        }
```
sengthai, line 241: It's nice to write into a file for future use. 👍  It would be great if we had an option for the user to disable or enable this feature.

sengthai (commented):
It's great to start exploring Catalyst through this PR. This workflow looks good for building Catalyst's pass. 👍 
Normally, for full process, we'd like to exposed the pass for end users in Python as well. We can config in `frontend/catalyst/passes/builtin_passes.py, so then the user could used this pass as decorator (e.g `@quantum-specs-info`)

mlir/lib/QEC/Transforms/CountPPMSpecs.cpp
zsb-xqc: Thanks for reviewing @sengthai!,  Yes I will add to front end also and give option to enable/disable writing to file.

Issue #1773, opened 2025-05-29
Decompose pi/8 PPR into PPM with "stale magic state"


sengthai:
**Context:**
#1664 implemented two methods (`inject-magic-state` and `auto-corrected`) for decomposing PPR into PPM. 

There is a decomposition method in  [Active volume: An architecture for efficient fault-tolerant quantum computers with limited non-local connections](https://arxiv.org/pdf/2211.15465), Figure 13(a). The pi/8 PPR uses a T magic state and turns it into a _stale T state_. _Stale magic state_ is introduced where the previous measurement didn't destroy that qubit's state.

**Objective:**
Here, we want to add this new method, `inject-stale-magic-state`.

**Implementation:**
- Implement a new method in `DecomposeNonCliffordPPR.cpp`

sengthai:
Here is Figure 13(a). 
![Image](https://github.com/user-attachments/assets/20a1af62-1808-495a-9cf9-16f763c62a95)

PR #2043, opened 2025-09-11
Efficient PPR to PPM Decomposition based on the Active Volume paper

lazypanda10117:
**Context:** 
Non-clifford PPRs can be decomposed more efficiently using only 2 PPMs with a magic state (see https://arxiv.org/abs/2211.15465), compared to the methods shown in the GOSC paper.

Screenshot 2025-09-10 at 22 47 24

We implemented the new decomposition rule here, under the name `pauli-corrected`. It replaces the existing `auto-corrected` decomposition method to be the new default decomposition method. 

Along the way, we also fixed a bug in the existing `auto-corrected` PPR to PPM decomposition method.

**Benefits:** This decomposition rule is more efficient that the existing ones.

**Possible Drawbacks:**

**Related GitHub Issues:** #1773 

[sc-99149]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -101,7 +101,6 @@
 * A new decomposition rule for non-Clifford PPRs into two PPMs based on the Active Volume paper.
   [(#2043)](https://github.com/PennyLaneAI/catalyst/pull/2043)
 
-
 <h3>Breaking changes 💔</h3>
 
 * The `shots` property has been removed from `OQDDevice`. The number of shots for a qnode execution is now set directly on the qnode via `qml.set_shots`,


frontend/test/pytest/test_peephole_optimizations.py
```diff
@@ -501,11 +541,21 @@ def g():
...
+                qml.CNOT(wires=[idx, idx + 1])
+                qml.T(idx)
+                qml.T(idx + 1)
+
+        return f(), g(), h()
 
     assert 'transform.apply_registered_pass "ppm-compilation"' in test_clifford_to_ppm_workflow.mlir
     optimized_ir = test_clifford_to_ppm_workflow.mlir_opt
     assert 'transform.apply_registered_pass "ppm-compilation"' not in optimized_ir
+    assert "qec.select.ppm" in optimized_ir
```
rniczh, line 558: Align the usage with other assertion here?
```suggestion
    assert 'qec.select.ppm' in optimized_ir
```

mlir/lib/QEC/Transforms/DecomposeNonCliffordPPR.cpp
```diff
@@ -12,8 +12,11 @@
...
 // limitations under the License.
 
+#include <llvm/ADT/SmallVector.h>
+#include <llvm/ADT/StringRef.h>
 #define DEBUG_TYPE "decompose-non-clifford-ppr"
-
+#include "llvm/Support/Debug.h"
+#include "llvm/Support/Errc.h"
 #include <mlir/Dialect/Arith/IR/Arith.h> // for arith::XOrIOp and arith::ConstantOp
 #include <mlir/IR/Builders.h>
```
rniczh: 1. `#define DEBUG_TYPE` should always at the begin , it ensures all headers see the macro and prevents macro pollution issues.
2. use `"..."` instead of `<...>` for third-party headers.

Standard Include Order:
- Standard library headers (`<...>`)
- Third-party library headers LLVM (`"llvm/..."`) -> Normally, MLIR will rely on LLVM infra, so include them before MLIR
- Third-party library headers MLIR (`"mlir/..."`) 
- Third-party library headers MLIR (`"{others}/..."`)  
- Project internal headers (`"Quantum/..."`)

mlir/lib/QEC/Transforms/DecomposeNonCliffordPPR.cpp
```diff
@@ -40,6 +43,62 @@ LogicalInitKind getMagicState(QECOpInterface op)
...
+/// |m⟩──| Z |────────| X / Y |
+///      └───┘        └───────┘
+/// All the operations in second diagram are PPM except for the last PPR P(π/2)
+/// For P(-π/8) we need to use complex conjugate|m̅⟩ as the magic state.
+///
+/// Details:
+/// - If P⊗Z measurement yields -1 then apply Y measurement, otherwise apply X measurement
+///   * The measurement results are stored as i1 values, -1 is true and 1 is false
+/// - If the X or Y measurement yields -1, apply P(π/2) on the input qubits
+void decompose_pauli_corrected_pi_over_eight(PPRotationOp op, PatternRewriter &rewriter)
```
rniczh: Following the [llvm coding standards](https://llvm.org/docs/CodingStandards.html#name-types-functions-variables-and-enumerators-properly), you should use camelCase.
```suggestion
void decomposePauliCorrectedPiOverEight(PPRotationOp op, PatternRewriter &rewriter)
```
Can you also help fixing other naming style in this file?

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -825,7 +828,7 @@ def ppr_to_ppm(qnode=None, *, decompose_method="auto-corrected", avoid_y_measure
         @to_ppr
         @commute_ppr
         @merge_ppr_ppm
-        @ppr_to_ppm
+        @ppr_to_ppm(default_method="auto-corrected")
```
sengthai: Do you mean `decompose_method` here? 🤔

mlir/lib/QEC/Transforms/DecomposeNonCliffordPPR.cpp
lazypanda10117: Thanks for pointing this out.

frontend/catalyst/passes/builtin_passes.py
lazypanda10117: yes, i have changed that now.

mlir/lib/QEC/Transforms/DecomposeNonCliffordPPR.cpp
lazypanda10117: Done!

doc/releases/changelog-dev.md
```diff
@@ -150,6 +154,9 @@
    - fix wrong type signature of `Counts` API function
   [(#2032)](https://github.com/PennyLaneAI/catalyst/pull/2032)
 
+* Fixed the Clifford PPR decomposition rule where using the Y measurement should be take the inverse.
+  [(#2043)](https://github.com/PennyLaneAI/catalyst/pull/2043)
```
sengthai: 👍

mlir/lib/QEC/Transforms/DecomposeNonCliffordPPR.cpp
```diff
@@ -40,6 +41,64 @@ LogicalInitKind getMagicState(QECOpInterface op)
...
+/// ─────|   |────────└───╦───┘
+///      |   ╠═════╗      ║
+///      |   |     ║  ┌───╩───┐
+/// |m⟩──| Z |─────╚══╣ X / Y |
+///      └───┘        └───────┘
+/// All the operations in second diagram are PPM except for the last PPR P(π/2)
+/// For P(-π/8) we need to use flip the ordering in which the X and Y measurements are applied.
+///
+/// Details:
+/// - If P⊗Z measurement yields -1 then apply Y measurement, otherwise apply X measurement
```
sengthai, line 64: I would recommend making this condition consistent with `decomposeAutoCorrectedPiOverEight`, where if the P⊗Z measurement yields -1, then apply the first measurement.

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -781,25 +781,28 @@ def circuit():
     return PassPipelineWrapper(qnode, merge_ppr_ppm_pass)
 
 
-def ppr_to_ppm(qnode=None, *, decompose_method="auto-corrected", avoid_y_measure=False):
+def ppr_to_ppm(qnode=None, *, decompose_method="pauli-corrected", avoid_y_measure=False):
```
sengthai, line 784: Since `avoid_y_measure` is not working in `pauli-corrected yet, I think we should notice in its docstring as well.

mlir/lib/QEC/Transforms/DecomposeNonCliffordPPR.cpp
```diff
@@ -40,6 +41,64 @@ LogicalInitKind getMagicState(QECOpInterface op)
...
+///   * The measurement results are stored as i1 values, -1 is true and 1 is false
```
dime10: `i1` can't have -1, only 0 and 1 right? 🤔

frontend/catalyst/passes/builtin_passes.py
lazypanda10117: I will leave a comment in the doc-string. I am actually working on the implementation of the y_measure decomposition for `pauli-corrected` in another ticket right now, so it should be included in this release as well.

mlir/lib/QEC/Transforms/DecomposeNonCliffordPPR.cpp
lazypanda10117: Done.

mlir/lib/QEC/Transforms/DecomposeNonCliffordPPR.cpp
lazypanda10117: Yes, I think it is a little unclear. I meant `-1` equals to `true` and `1` equals to `false`, so in the `i1 = true = 1 `corresponds to measuring `-1` and `i1 = false = 0` corresponds to measuring `1`. 

I've updated the docstring for this function and also for the other decomposition rules.

doc/releases/changelog-dev.md
```diff
@@ -98,6 +98,10 @@
 * Renaming `get_ppm_specs` to `ppm_specs` and the corresponding results' properties.
   [(#2031)](https://github.com/PennyLaneAI/catalyst/pull/2031)
 
+* A new decomposition rule for non-Clifford PPRs into two PPMs based on the Active Volume paper.
+  [(#2043)](https://github.com/PennyLaneAI/catalyst/pull/2043)
+
+
```
```suggestion
```

frontend/test/pytest/test_peephole_optimizations.py
```diff
@@ -336,14 +336,14 @@ def f():
     assert ppm_specs_output["f_0"]["logical_qubits"] == 2
 
 
-def test_ppr_to_ppm():
+def test_ppr_to_ppm_auto_corrected():
```
rniczh, line 339: I think you can use `pytest.mark.parametrize` to distinguish the `auto_corrected` and `pauli_corrected`. And check the result by using `if decompose_method == "..."`. Do you think it would be more maintainable?

lazypanda10117: I think I prefer the current implementation instead of mixing the assertion logic inside the test case. I would agree if the different methods yield the same outcome.

PR #360, opened 2023-11-09
[Frontend & MLIR] Last quantum register is dealloc: Quantum AllocOp and DeallOc use Value semantics

rmoyard:
**Context:**

AllocOp has both memory semantics and value semantics. We want to be able to use the last quantum register of quantum functions.

**Description of the Change:**

- Dealloc only the last quantum register
- AllocOp and DealloOp have Value semantics only

**Benefits:**

- Quantum passes can use the last register.

**Possible Drawbacks:**

- Cse eliminates two similar quantum allocs (because pure)
- It would not work with multiple allocation in the same function

PR #2044, opened 2025-09-11
Remove `NoMemoryEffect` from quantum alloc op

paul0403:
**Context:**
We mark `quantum.alloc` operation with the `NoMemoryEffect` trait, because we want passes like ~~CSE and~~ DCE to automatically remove them. #360 

However, with the upcoming dynamic allocation work, we must no longer ask CSE to merge two similar alloc ops. Because obviously in the following IR both allocations should stay:
```python
   # user code in python
   wires0 = qml.allocate(3)
   wires1 = qml.allocate(3)
```
which becomes 
```mlir
     // lowered mlir
     %0 = quantum.alloc(3) : !quantum.reg
     %1 = quantum.alloc(3) : !quantum.reg
```

The multiple allocation IR is also a common pattern in FTQC.

In addition, just semantically, quantum allocation op is definitely not without memory effects.

- Remove `NoMemoryEffect` trait on quantum alloc op
- Update the alloc op cse lit test to check that cse now does **not** merge two similar alloc ops
- We still want to remove unused allocations. We now do this properly in alloc op's canonicalization instead of hinting DCE to do it.
- Update gradient lowering to trigger the new alloc op dead-removal canonicalization. Gradient lowering was the place that needed this in this first place because of the callee cloning.

**Benefits:**
Correct semantics.
Unblocks dynamic wire allocation and FTQC.

[sc-98713]

### DIFF
--- a/mlir/lib/Gradient/Transforms/gradient_lowering.cpp
+++ b/mlir/lib/Gradient/Transforms/gradient_lowering.cpp
@@ -66,18 +66,16 @@ struct GradientLoweringPass : impl::GradientLoweringPassBase<GradientLoweringPas
         }
 
         // Guarantee that functions intended to be free of quantum ops are indeed so after folding.
-        for (Region &region : getOperation()->getRegions()) {
-            for (Operation &op : region.getOps()) {
-                if (isa<func::FuncOp>(op) && op.hasAttr("QuantumFree")) {
-                    if (failed(quantum::verifyQuantumFree(cast<func::FuncOp>(op)))) {
-                        op.emitOpError() << "cloned during the gradient pass is not free of "
-                                            "quantum ops:\n"
-                                         << op;
-                        return signalPassFailure();
-                    }
+        getOperation()->walk([&](func::FuncOp f) {
+            if (f->hasAttr("QuantumFree")) {
+                if (failed(quantum::verifyQuantumFree(f))) {
+                    f->emitOpError() << "cloned during the gradient pass is not free of "
+                                        "quantum ops:\n"
+                                     << *f;
+                    return signalPassFailure();
                 }
             }
-        }
+        });
     }
 };
 


mlir/include/Quantum/Utils/RemoveQuantum.h
```diff
@@ -20,6 +20,7 @@
 namespace catalyst {
 namespace quantum {
 
+void removeUnunsedAllocs(mlir::Operation *op);
```
rniczh: remove**Unused**Allocs?

paul0403: Yes :sweat_smile:

mlir/lib/Gradient/Transforms/gradient_lowering.cpp
```diff
@@ -63,16 +64,20 @@ struct GradientLoweringPass : impl::GradientLoweringPassBase<GradientLoweringPas
...
             for (Operation &op : region.getOps()) {
-                if (isa<func::FuncOp>(op) && op.hasAttr("QuantumFree"))
+                if (isa<func::FuncOp>(op) && op.hasAttr("QuantumFree")) {
                     if (failed(quantum::verifyQuantumFree(cast<func::FuncOp>(op)))) {
                         op.emitOpError() << "cloned during the gradient pass is not free of "
                                             "quantum ops:\n"
                                          << op;
                         return signalPassFailure();
                     }
+                }
```
rniczh: I know it's not what you did but this part just has a little confusing me, why not just using `getOperation()->walk([&](func::FuncOp funcOp)`

paul0403: https://github.com/PennyLaneAI/catalyst/pull/2044/commits/6792d28f3afc26494b5cab14374ce5b22bed79a3

PR #2047, opened 2025-09-12
Implement Y-basis PPM decomposition in the Active Volume paper

lazypanda10117:
**Context:** #2043 introduced conditional measurements in the Y-basis when decomposing PPRs into PPMs. There is a way to circumvent having Y-basis measurements by using an injected Y state (See figure 13(b) [here](https://arxiv.org/pdf/2211.15465)).

**Description of the Change:** This introduces an option for users to decompose Y-basis measurements into X-basis and Z- basis measurements.

**Benefits:** It allows users to choose whether they want to allow Y basis measurement or decompose it into X and Z bases measurements with an injected Y state.

[[sc-99154]]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -29,15 +29,15 @@
 <h3>Improvements 🛠</h3>
 
 * Catalyst now supports returning classical and MCM values with the dynamic one-shot MCM method.
-  [(#2001)](https://github.com/PennyLaneAI/catalyst/pull/2001)
-  
+  [(#2004)](https://github.com/PennyLaneAI/catalyst/pull/2004)
+
   For example, the code below will generate 10 values, with an equal probability of 42 and 43
-  appearing. 
+  appearing.
 
   ```python
   import pennylane as qml
   from catalyst import qjit, measure
-  
+
   @qjit(autograph=True)
   @qml.set_shots(10)
   @qml.qnode(qml.device("lightning.qubit", wires=1), mcm_method="one-shot")
@@ -57,12 +57,12 @@
            True], dtype=bool))
   ```
 
-
 * Improve the pass `--ppm-specs` to count the depth of PPRs and PPMs in the circuit.
   [(#2014)](https://github.com/PennyLaneAI/catalyst/pull/2014)
 
 * The default mid-circuit measurement method in catalyst has been changed from `"single-branch-statistics"` to `"one-shot"`.
   [[#2017]](https://github.com/PennyLaneAI/catalyst/pull/2017)
+  [[#2019]](https://github.com/PennyLaneAI/catalyst/pull/2019)
 
 * A new pass `--partition-layers` has been added to group PPR/PPM operations into `qec.layer`
   operations based on qubit interactive and commutativity, enabling circuit analysis and
--- a/frontend/catalyst/jax_tracer.py
+++ b/frontend/catalyst/jax_tracer.py
@@ -1035,7 +1035,7 @@ def trace_quantum_measurements(
                 output, qml.measurements.SampleMP
             ):
                 raise NotImplementedError(
-                    f"Measurement {type(output).__name__} is not supported a shot-vector. "
+                    f"Measurement {type(output).__name__} does not support shot-vectors. "
                     "Use qml.sample() instead."
                 )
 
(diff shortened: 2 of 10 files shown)


mlir/lib/QEC/Transforms/DecomposeNonCliffordPPR.cpp
```diff
@@ -14,9 +14,12 @@
 
 #define DEBUG_TYPE "decompose-non-clifford-ppr"
 #include "mlir/Dialect/Arith/IR/Arith.h" // for arith::XOrIOp and arith::ConstantOp
+#include "mlir/Dialect/SCF/IR/SCF.h"
 #include "mlir/IR/Builders.h"
+#include "mlir/IR/BuiltinTypes.h"
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/ADT/StringRef.h"
+#include <llvm/Support/Debug.h>
```
sengthai: Just minor. I wonder if we are using functions from `Debug.h`?

doc/releases/changelog-dev.md
```diff
@@ -131,6 +131,9 @@
 * A new decomposition rule for non-Clifford PPRs into two PPMs based on the Active Volume paper.
   [(#2043)](https://github.com/PennyLaneAI/catalyst/pull/2043)
 
+* Added support for avoid Y basis measurement for `pauli-corrected` PPR decomposition.
```
```suggestion
* Added support to avoid Y-basis measurements in `pauli-corrected` PPR decomposition.
```

mlir/lib/QEC/Transforms/DecomposeNonCliffordPPR.cpp
```diff
@@ -76,28 +84,75 @@ void decomposePauliCorrectedPiOverEight(PPRotationOp op, PatternRewriter &rewrit
...
-    rewriter.create<DeallocQubitOp>(loc, ppmPZ.getOutQubits().back());
-
-    rewriter.replaceOp(op, pprPI2.getOutQubits());
+    if (avoidPauliYMeasure) {
+        auto constFalse =
+            rewriter.create<arith::ConstantOp>(loc, rewriter.getIntegerAttr(ppmPZRes.getType(), 0));
+        auto constTrue =
+            rewriter.create<arith::ConstantOp>(loc, rewriter.getIntegerAttr(ppmPZRes.getType(), 1));
+        auto ifCondition = rewriter.create<arith::XOrIOp>(
+            loc, ppmPZRes, rotationKind > 0 ? constTrue.getResult() : constFalse.getResult());
```
sengthai: Nice to use the XOR op for this case. Will this XOR op evaluate in compilation? (I think some optimization will do that) 🤔
In the PPM pipeline, we assume that we know the `rotationKind` at compile time, so I would recommend checking the condition with that directly without using XOR.

sengthai (changes requested):
Nice work! The PR title feels a bit broad. Could you specify that it's about Y measurement decomposition?

mlir/lib/QEC/Transforms/DecomposeNonCliffordPPR.cpp
lazypanda10117: Ahh my bad, i forgot to remove this after debugging.

mlir/lib/QEC/Transforms/DecomposeNonCliffordPPR.cpp
lazypanda10117: The XOR will be compile-time evaluated. When the rotation kind is negative, the `if else` block will be swapped in the underlying IR. I did it this way since it is cleaner without having to do further branching.

sengthai: >The XOR will be compile-time evaluated. 

Nice! 

>I did it this way since it is cleaner without having to do further branching.

Ohh i mean not creating a new branch for checking this. However, we are using check on rotation directly when compiling, like this:
```c++
if (rotationkind > 0){
    . . . 
} else {
    . . .
}
```

lazypanda10117: Oh yea, sure that would work too and probably easier to reason about.

sengthai (approved):
Looks great! 💯 🚀 
Thanks @lazypanda10117!

PR #2049, opened 2025-09-16
Enhance `ppm_specs` function to prevent duplicate pass addition

sengthai:
Enhance ppm_specs function to prevent duplicate pass addition

Added a check to ensure "ppm-specs" is not already in the pass list before appending it. This prevents unnecessary duplication in the pipeline configuration.


lazypanda10117 (approved):
Looks good!

PR #2051, opened 2025-09-16
print module scope when using keep_intermediate

mehrdad2m:
**Context:**
Currently, when using `keep_intermediate="pass"` the functions passes only print the scope they are applied to. This can make investigating the IRs harder. 

- Added an option to the compiler driver that controls whether to print module scope.
- Always print the module scope with using `keep_intermediate="pass"`. 

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


mlir/lib/Driver/CompilerDriver.cpp
```diff
@@ -505,7 +505,20 @@ LogicalResult preparePassManager(PassManager &pm, const CompilerOptions &options
...
+                else if (auto parentMod = op->getParentOfType<mlir::ModuleOp>()) {
+                    s << parentMod;
+                }
+                else {
+                    s << *op;
+                }
+            }
+            else {
+                s << *op;
+            }
```
paul0403, line 516: This block seems a bit complicated, it's just trying to print the closest surrounding module right?

Is the problem that `getParentOfType<blah>(op)` will nullptr if `op` is already an `blah`? In which case can we just do
```C
if (dumpModuleScope){
   ModuleOp mod = isa<ModuleOp>(op) ? op :  op->getParentOfType<ModuleOp>(op);
   s << mod;
} else {
   s << *op;
}
```

paul0403 (approved):

PR #2054, opened 2025-09-18
Issue 1714 add variable names to the ir

rturrado:
**Context:**
We can investigate the option of passing down source code information like variable names from the frontend down to the generated ir, using MLIR's location mechanism. The additional info will make generated code more readable.

1) User -> Frontend: `@qjit(use_nameloc=True)` to use name location information in the generated MLIR code.
2) Frontend -> Catalyst CLI: whenever `use_nameloc` is enabled:
    - Pass `stdin=self.mlir_module.operation.get_asm(enable_debug_info=True)` instead of `stdin=str(self.mlir_module)`.
    - Pass `--use-nameloc-as-prefix` in `options`.

3) Catalyst CLI -> frontend: whenever `--use-nameloc-as-prefix` is enabled:
    - Enable `useNameLocAsPrefix` printing flag.
    - Print the MLIR module to stdout with printing flags, e.g., `mlirModule->print(oss, opPrintingFlags)` instead of `oss << *mlirModule`.

**Benefits:**
- `@qjit(use_nameloc=True)` is an eay way for the user to specify that they want to use name location information.
- The same applies to Catalyst CLI's `--use-nameloc-as-prefix` option.

**Possible Drawbacks:**
- `get_arg_names` function may not be robust enough.
- Name location information works nicely for function parameter names (e.g., `%x` instead of `%arg0`), but maybe not so nicely for results of equations (e.g., `%jit28ok292Fjit28jit_f292Fmul` instead of `%0`). An option would be to add name location information only for function parameters.

**Related GitHub Issues:**
#1714


mlxd:
Thanks @rturrado for the submission.
We are currently finalizing our plans for Q4, so may take a little longer to get around to reviewing this. 
Thanks in advance for your patience.

doc/catalyst-cli/catalyst-cli.rst
```diff
@@ -98,6 +98,11 @@ intermediate files are saved.
 Keep intermediate files after each pipeline in the compilation. By default, no intermediate files
 are saved. Using ``--keep-intermediate`` is equivalent to using ``--save-ir-after-each=pipeline``.
 
+``--use-nameloc-as-prefix[=<true|false>]``
+""""""""""""""""""""""""""""""""""""""""""
+
+Print SSA IDs using their name location, if provided, as prefix. By default, name location information is not used.
```
dime10, line 104: Maybe we could say a bit more about what a "name location" is :)

frontend/test/lit/test_option_use_nameloc.py
```diff
@@ -0,0 +1,38 @@
...
+    """
+    # CHECK: %x: tensor<f64>, %y: tensor<f64>
+    return x * y
+
+
+assert str(f.mlir_module.body.operations[0].arguments[0].location) == 'loc("x")'
+assert str(f.mlir_module.body.operations[0].arguments[1].location) == 'loc("y")'
+
+print_mlir(f, 0.3, 0.4)
+print_mlir_opt(f, 0.3, 0.4)
```
dime10: Does the second line do anything here? Typically we the print the IR in the lit tests so that the CHECK statements can be matched against it, but if they already matched once, they aren't going to be matched against anything further down in the stream.

mlir/test/Catalyst/PrintUseNameLocAsPrefix.mlir
```diff
@@ -0,0 +1,31 @@
+// RUN: quantum-opt %s -mlir-use-nameloc-as-prefix -split-input-file | FileCheck %s
```
dime10: This is not our pass is it? If not, we're just testing upstream code right 🤔

frontend/catalyst/jit.py
```diff
@@ -562,20 +565,28 @@ def __init__(self, fn, compile_options):
...
+        """Obtain the MLIR representation after canonicalization"""
         # Canonicalize the MLIR since there can be a lot of redundancy coming from JAX.
         if not self.mlir_module:
             return None
 
-        return canonicalize(stdin=str(self.mlir_module))
+        if self.compile_options.use_nameloc:
+            stdin = self.mlir_module.operation.get_asm(enable_debug_info=True)
+        else:
+            stdin = str(self.mlir_module)
```
dime10: Is this different from using this method unconditionally?
```suggestion
        stdin = self.mlir_module.operation.get_asm(enable_debug_info=self.compile_options.use_nameloc)
```

frontend/catalyst/jit.py
```diff
@@ -766,6 +775,28 @@ def closure(qnode, *args, **kwargs):
 
         return jaxpr, out_type, treedef, dynamic_sig
 
+    def get_arg_names(self):
```
dime10: I'm hesitant to expose another method in the QJIT class as we want to keep it relatively stable and avoid class bloat. Could this be implemented as a free-standing function in the `tracing` submodule (or other jax-related modules if more appropriate, but I think `type_signatures.py` looks pretty relevant)?

frontend/test/lit/test_option_use_nameloc.py
rturrado: I added the `print_mlir_opt` line only to exercise the `mlir_opt` property and thus remove a code coverage warning.

mlir/test/Catalyst/PrintUseNameLocAsPrefix.mlir
rturrado: That's very right! I'll have a look if the test still makes sense using `--use-nameloc-as-prefix`, or I can remove it altogether.

frontend/catalyst/jit.py
rturrado: I think you may be right here, yes. `self.mlir_module` may be a synonym for `self.mlir_module.operation.get_asm()`.

frontend/test/lit/test_option_use_nameloc.py
dime10: Got it, in that case I'd rather have the coverage be reported as missing, since there is no meaningful test verifying the functionality, or duplicate this test so that the mlir_opt property is meaningfully tested if that is easily done (but no big deal if not).

dime10:
Kind of random but I just noticed this PR which is interesting and somewhat relevant: https://github.com/llvm/llvm-project/commit/a40f47c9725f3aed6752f050f3010f3701d0cff7

dime10:
> > Nice work, another excellent PR! 🥇
> 
> Thanks so much for the feedback and the review, @dime10 !
> 
> I have a [problem](https://github.com/PennyLaneAI/catalyst/discussions/2082) though when testing changes on this PR. I am getting a build error since a recent update.

Thanks for the updates! I left a comment in the discussion thread, hopefully it can help you. Otherwise, the PR looks great 💯

mlir/test/Catalyst/PrintUseNameLocAsPrefix.mlir
dime10: Oops, `use-nameloc-as-prefix` doesn't exist (sorry I didn't mean to imply we had our own version of the pass). 

I would say this lit test is superfluous then, since even though we _use_ this pass, there is no code of our own that we are testing here.

rturrado: You're right. I added this test at the very beginning to check myself  how `-mlir-use-nameloc-as-prefix` worked. But it was more a tool for me than anything else.

dime10 (approved):
🚀

frontend/test/pytest/test_get_arg_names.py
dime10, line 1: Looks like these tests need updating to the free function.

dime10:
I want ahead and added a changelog entry [here](https://github.com/PennyLaneAI/catalyst/commit/480e0a984429af0bbcd07a6bd8e47a002ff1694c) for expediency. Thank you for the awesome contribution 🎉

rturrado:
Thanks a lot for your help and your comments, @dime10 and @paul0403!

PR #2055, opened 2025-09-19
Add system tests for qml.specs using Catalyst resource tracking

jzaia18:
**Context:**
PRs https://github.com/PennyLaneAI/catalyst/pull/2033 and https://github.com/PennyLaneAI/pennylane/pull/8202 integrate `qml.specs` with Catalyst `qjit`'d code via the resource tracking feature.

Adds end-to-end tests of `qml.specs` on `qjit`'d code.

**Benefits:**
Code coverage.

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-99166]

### DIFF
--- a/frontend/test/pytest/test_device_api.py
+++ b/frontend/test/pytest/test_device_api.py
@@ -139,6 +139,18 @@ def test_track_resources():
     dev = NullQubit(wires=2, track_resources=True)
     assert "track_resources" in QJITDevice.extract_backend_info(dev).kwargs
     assert QJITDevice.extract_backend_info(dev).kwargs["track_resources"] is True
+    assert "resources_filename" not in QJITDevice.extract_backend_info(dev).kwargs
+    assert "compute_depth" not in QJITDevice.extract_backend_info(dev).kwargs
+
+    dev = NullQubit(
+        wires=2, track_resources=True, resources_filename="my_resources.txt", compute_depth=True
+    )
+    assert "track_resources" in QJITDevice.extract_backend_info(dev).kwargs
+    assert QJITDevice.extract_backend_info(dev).kwargs["track_resources"] is True
+    assert "resources_filename" in QJITDevice.extract_backend_info(dev).kwargs
+    assert QJITDevice.extract_backend_info(dev).kwargs["resources_filename"] == "my_resources.txt"
+    assert "compute_depth" in QJITDevice.extract_backend_info(dev).kwargs
+    assert QJITDevice.extract_backend_info(dev).kwargs["compute_depth"] is True
 
 
 if __name__ == "__main__":


frontend/test/pytest/test_specs.py
```diff
@@ -0,0 +1,109 @@
...
+
+    @qml.qnode(dev)
+    def circuit():
+        qml.Hadamard(wires=0)
+        return qml.expval(qml.PauliZ(0))
+
+    pl_specs = qml.specs(circuit, level=level)()
+    cat_specs = qml.specs(qjit(circuit), level=level)()
+
+    assert pl_specs["resources"].num_wires == cat_specs["resources"].num_wires
```
paul0403: Is it possible to factor out the comparisons into its own function? `check_pl_and_cat_has_same_specs()` or something (you might come up with a better name).

frontend/test/pytest/test_specs.py
```diff
@@ -0,0 +1,109 @@
...
+    for gate, count in pl_specs["resources"].gate_sizes.items():
+        assert gate in cat_specs["resources"].gate_sizes
+        assert count == cat_specs["resources"].gate_sizes[gate]
+
+
+@pytest.mark.parametrize("level", ["device"])
+def test_complex(level):
+    """Test a complex case of qml.specs() against PennyLane"""
+
+    dev = qml.device("lightning.qubit", wires=4)
```
paul0403, line 64: A bit confused, why is it lightning qubit now? Wasn't your catalyst resource tracking feature implemented on null qubit? Or are these new tests not related to the null qubit implementation you added?

jzaia18: This isn't testing resource tracking, it's tracking `qml.specs` which has to be compatible with any device. Under the hood, specs replaces the device with `null.qubit` (with some of the original device's info monkey-patched in) and uses resource tracking.

paul0403: If we are replacing the device, aka explicitly going against a user directive, we should raise a warning in my opinion, something like "resource tracking with qjit is only supported with null qubit, but got lightning qubit. Replacing device with null qubit instead".

jzaia18: The purpose of the feature is to get the resources for the original device. But we handle that as a part of how we do the replacement to `null.qubit`, taking the capabilities from the original device TOML and overriding the `preprocess` function. CC @AntonNI8

paul0403: > explicitly going against a user directive, we should raise a warning

Thanks for the context, but I still think this should be true.

paul0403: > But we handle that as a part of how we do the replacement to null.qubit

Especially if it's because of our implementation details

AntonNI8: My understanding is that when the device is replaced, it is only done in the context of `specs`. When the user proceeds to compile and execute (as normal) they will not find that their device has suddenly been replaced. If it were, I would agree that we should raise a warning.

paul0403: Where is the replacement done? Because if the IR has null qubit then all executions will be on null qubit

jzaia18: The executions have to be done on `null.qubit`, that's the only place that we have resource tracking. Plus, we don't want to actually execute (on the real device) since that would be extremely expensive (or even impossible considering we want to work with thousands to millions of qubits). The code for how everything happens is here: https://github.com/PennyLaneAI/pennylane/blob/34de4fae7366f6a8a6a96bfa562ed4c373250caa/pennylane/resource/specs.py#L102

paul0403: So under `qml.specs` regular execution (aka doing maths on lightning) is not actually happening?

AntonNI8: Correct.

paul0403: Thanks for confirming, in that case maybe it's fine then... I still feel quite uncomfortable about just changing a user parameter but if a decision has already been made here I don't have objections :+1:

paul0403: Or maybe, the idea was "we want to track resource usage on lightning.qubit, but because of implementation details we will implement the underlying feature on null qubit, but to the user it is as if it's still run on lightning.qubit"? Because `qml.specs` does not actually return circuit results. 

So user is still under the impression/facade of "this resource tracking is run on lightning"

jzaia18: From the docs:

> This transform converts a QNode into a callable that provides resource information about the circuit after applying the specified amount of transforms/expansions first.

I think that communicates that the code doesn't actually get executed. There's a separate feature, `qml.Tracker` which executes AND tracks info, `qml.specs()` is purely for returning resource information.

Side note: `qml.Tracker` is actually also Catalyst incompatible, but that's an issue for another day

AntonNI8: > So user is still under the impression/facade of "this resource tracking is run on lightning"

@paul0403 yes, this is why `null.qubit` is configured to imitate the user-specified device. This way, the `specs` returned should match what would occur for execution on the user-specified device.

PR #2056, opened 2025-09-19
Fixing qml demo one shot issue

lazypanda10117:
**Context:** The recent one-shot implementaiton in Catalyst is causing a build [failure](https://github.com/PennyLaneAI/qml/actions/runs/17844479263/job/50741097686) in the QML demo repository, giving the error:

```
      ...

      File "/home/runner/work/qml/qml/.venv-build/lib/python3.11/site-packages/catalyst/qfunc.py", line 292, in __call__
        fn_result = configure_mcm_and_try_one_shot(self, args, kwargs)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/runner/work/qml/qml/.venv-build/lib/python3.11/site-packages/catalyst/qfunc.py", line 173, in configure_mcm_and_try_one_shot
        return Function(dynamic_one_shot(qnode, mcm_config=mcm_config))(*args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      ...

      File "/home/runner/work/qml/qml/.venv-build/lib/python3.11/site-packages/catalyst/qfunc.py", line 334, in __call__
        flattened_fun, _, _, out_tree_promise = deduce_avals(
                                                ^^^^^^^^^^^^^
      File "/home/runner/work/qml/qml/.venv-build/lib/python3.11/site-packages/catalyst/jax_extras/tracing.py", line 401, in deduce_avals
        verify_static_argnums_type(static_argnums)
      File "/home/runner/work/qml/qml/.venv-build/lib/python3.11/site-packages/catalyst/tracing/type_signatures.py", line 101, in verify_static_argnums_type
        raise TypeError(
    TypeError: The `static_argnums` argument to `qjit` must be an int or convertable to atuple of ints, but got value (Traced<~int64[]>with<DynamicJaxprTrace>,)
```

As a side note, the failure comes from the checks in`frontend/catalyst/tracing/type_signatures.py`.
```python
def verify_static_argnums_type(static_argnums):
    """Verify that static_argnums have correct type.

    Args:
        static_argnums (Iterable[int]): indices to verify

    Returns:
        None
    """
    is_tuple = isinstance(static_argnums, tuple)
    is_valid = is_tuple and all(isinstance(arg, int) for arg in static_argnums)
    if not is_valid:
        raise TypeError(
            "The `static_argnums` argument to `qjit` must be an int or convertable to a"
            f"tuple of ints, but got value {static_argnums}"
        )
    return None
```
When we ignore the verification, the aforementioned issue will be resolved as well. However, it doesn't make sense to remove this check as a solution, as it may lead to more issues down the line. 

**Description of the Change:** We solve this by passing `static_argnums` correctly to internally transformed kernel functions,
which includes the one-shot mid circuit measurement transform here.

**Benefits:** Resolve the build failure in the QML repo.

### DIFF
--- a/frontend/catalyst/qfunc.py
+++ b/frontend/catalyst/qfunc.py
@@ -51,6 +51,7 @@ from catalyst.utils.exceptions import CompileError
 logger = logging.getLogger(__name__)
 logger.addHandler(logging.NullHandler())
 
+
 @dataclass
 class OutputContext:
     """Context containing parameters needed for finalizing quantum function output."""
--- a/frontend/catalyst/qfunc.py
+++ b/frontend/catalyst/qfunc.py
@@ -50,7 +50,7 @@ from catalyst.utils.exceptions import CompileError
 
 logger = logging.getLogger(__name__)
 logger.addHandler(logging.NullHandler())
-
+logging.basicConfig(filename='mylog.log', level=logging.DEBUG)
 
 @dataclass
 class OutputContext:
@@ -179,6 +179,7 @@ def configure_mcm_and_try_one_shot(qnode, args, kwargs):
 
                 # Fallback only if mcm was auto-determined
                 error_msg = str(e)
+                # TODO: handle the underlying reasons for the unsupported errors
                 unsupported_measurement_error = any(
                     pattern in error_msg
                     for pattern in [
@@ -192,7 +193,7 @@ def configure_mcm_and_try_one_shot(qnode, args, kwargs):
 
                 # Fallback if error is related to unsupported measurements
                 if unsupported_measurement_error:
-                    logger.debug("Fallback to single-branch-statistics: %s", e)
+                    logger.warning("Fallback to single-branch-statistics: %s", e)
                     mcm_config = replace(mcm_config, mcm_method="single-branch-statistics")
                 else:
                     raise
--- a/frontend/catalyst/qfunc.py
+++ b/frontend/catalyst/qfunc.py
@@ -172,7 +172,6 @@ def configure_mcm_and_try_one_shot(qnode, args, kwargs):
             try:
                 return Function(dynamic_one_shot(qnode, mcm_config=mcm_config))(*args, **kwargs)
             except (TypeError, ValueError, CompileError, NotImplementedError) as e:
-
                 # If user specified mcm_method, we can't fallback to single-branch-statistics,
                 # reraise the original error
                 if user_specified_mcm_method is not None:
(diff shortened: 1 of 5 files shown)


sengthai (approved):
👍

frontend/catalyst/qfunc.py
```diff
@@ -172,27 +172,28 @@ def configure_mcm_and_try_one_shot(qnode, args, kwargs):
...
                 error_msg = str(e)
+                # TODO: handle the underlying reasons for the unsupported errors
                 unsupported_measurement_error = any(
                     pattern in error_msg
                     for pattern in [
                         "Native mid-circuit measurement mode does not support",
                         "qml.var(obs) cannot be returned when `mcm_method='one-shot'`",
                         "empty wires is not supported with dynamic wires in one-shot mode",
                         "No need to run one-shot mode",
+                        "The `static_argnums` argument to `qjit` must be an int",
```
paul0403: Sorry, why are we falling back?

paul0403: Doesn't this mean there is an issue inside our own one-shot transform code? Likely due to jax's omnistaging (jax ops will return tracers regardless of whether it received tracers or static literals) some jax reshaping op in our one shot transform code is producing a tracer. That would be my guess.

lazypanda10117: Yes, i agree this implies some issue deeper down in the one-shot transform code. We would need to track this and resolve this later. This is more of a quick fix to resolve the qml demo build issue. @rniczh should have better context as to how we can actually resolve the underlying issue.

paul0403: found the issue and fixed it

paul0403:
> @paul0403 Thanks for clarify
> 
> The following solution works for me, do you want to give it a try?
> 
> ```
> --- a/frontend/catalyst/jax_tracer.py
> +++ b/frontend/catalyst/jax_tracer.py
> @@ -225,8 +225,11 @@ class Function:
>  
>      @debug_logger
>      def __call__(self, *args, **kwargs):
> +        static_argnums = kwargs.pop("static_argnums", ())
> +
>          jaxpr, _, out_tree = make_jaxpr2(
>              self.fn,
> +            static_argnums=static_argnums,
>              debug_info=kwargs.pop("debug_info", jdb("Function", self.fn, args, kwargs)),
>          )(*args, **kwargs)
> ```

Yes I found the same issue :laughing: so this bug was there for god knows how long

It is only exposed now because I think the one shot wrapper is the only thing that uses this `Function` class (maybe also gradients? but not sure there)

rniczh (approved):
Looks good to me 🎉 Thanks @lazypanda10117 and @paul0403 for jumping right into this issue and getting it solved.

PR #2057, opened 2025-09-19
Add PPR/PPM To MBQC lowering pass

sengthai:
**Context:**
Introduces a bottom-of-stack MBQC execution pathway by shimming PPR/PPM into native MBQC primitives, enabling end-to-end experimentation on a mocked backend.

- Adds a new pass `--ppr-to-mbqc` that lowers `qec.ppr`/`qec.ppm` into native MBQC-style operations.
- Lowering strategy:
   * Per-qubit Pauli conjugation to map the Pauli word to Z-basis (H/RotXZX as needed).
   * Right-to-left CNOT ladder to accumulate parity onto a designated qubit; reverse ladder to uncompute.
   * Kernel emission: RZ for rotations (angle derived from rotation kind) or measurement for PPM.
   * Reverse conjugation to restore basis post-kernel.
   * Conversion target legalizes quantum and arith dialects; PPR/PPM become illegal and are replaced.

**Benefits:**
Enables a functional MBQC pipeline on virtual hardware using existing PPR/PPM infrastructure.

**Usage**
```python
import pennylane as qml
from catalyst.passes import to_ppr, ppr_to_mbqc

pipeline = [("pipe", ["enforce-runtime-invariants-pipeline"])]

@qjit(pipelines=pipeline, keep_intermediate=True, target="mlir")
@ppr_to_mbqc
@to_ppr
def circuit():
    qml.H(0)
    qml.CNOT([0, 1])
    return measure(1)

print(circuit.mlir_opt)

```

[[sc-98879]]


frontend/catalyst/passes/builtin_passes.py
```diff
@@ -1159,3 +1159,89 @@ def circuit():
...
+      - **PPR**: apply an ``RZ`` with an angle derived from the rotation kind.
+      - **PPM**: perform a measurement and return an ``i1`` result.
+    - Uncomputing by reversing the CNOT ladder and the conjugations.
+    - Conjugating the qubits back to the original basis.
+
+    .. note::
+
+        This pass expects PPR/PPM operations to be present. In practice, use it
+        after :func:`~.passes.to_ppr` and/or :func:`~.passes.commute_ppr` and/or
+        :func:`~.passes.merge_ppr_ppm`.
```
lazypanda10117, line 1207: Is this compatible with the `ppr_to_ppm` pass as well?

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -1159,3 +1159,89 @@ def circuit():
...
+    measurements. Conceptually, each Pauli product is handled by:
+
+    - Mapping its Pauli string to the Z basis via per-qubit conjugations
+      (e.g., ``H`` for ``X``; specialized ``RotXZX`` sequences for ``Y``).
+    - Accumulating parity onto the first qubit with a right-to-left CNOT ladder.
+    - Emitting the kernel operation:
+      - **PPR**: apply an ``RZ`` with an angle derived from the rotation kind.
+      - **PPM**: perform a measurement and return an ``i1`` result.
+    - Uncomputing by reversing the CNOT ladder and the conjugations.
+    - Conjugating the qubits back to the original basis.
```
lazypanda10117, line 1201: If it's convenient, I think it would be helpful if you can also link the source (e.g, arxiv) of this decomposition here.

lazypanda10117 (approved):
Looks good to me! I think it would be nice to also add the circuit diagrams for the decomposition to the PR description.

doc/releases/changelog-dev.md
```diff
@@ -26,6 +26,11 @@
       return qml.probs()
   ```
 
+* Added ``--ppr-to-mbqc`` to lower ``qec.ppr``/``qec.ppm`` into an MBQC-style quantum circuit.
```
paul0403: I think all mbqc entries go into the internal changes section now because that project is not user facing yet.

paul0403: Also just a note a changelog format: the link goes after the first summary sentence.

```
 * Added a new change.
   [(#...)](httprs://...)

   ...details of the change...
```

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -1159,3 +1159,89 @@ def circuit():
...
+
+
+def ppr_to_mbqc(qnode):
+    R"""
+    Specify that the MLIR compiler pass for lowering Pauli Product Rotations (PPR)
+    and Pauli Product Measurements (PPM) to a measurement-based quantum computing
+    (MBQC) style circuit will be applied.
+
+    This pass replaces QEC operations (``qec.ppr`` and ``qec.ppm``) with a
+    gate-based sequence in the Quantum dialect using standard gates and
```
paul0403: Shouldn't it be called `@ppr_to_quantum` then? Where does mbqc come in?

mlir/lib/QEC/Transforms/ppr_to_mbqc.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+        RewritePatternSet patterns(ctx);
+        populatePPRToMBQCPatterns(patterns);
+
+        if (failed(applyPartialConversion(getOperation(), target, std::move(patterns)))) {
+            return signalPassFailure();
+        }
+
+        // Run CSE to deduplicate constants and trivial common constants
+        PassManager pm(&getContext());
```
rniczh: you can use `ctx` right? No need to `getContext()` again
```suggestion
        PassManager pm(ctx);
```

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
```diff
@@ -0,0 +1,238 @@
...
+    return rewriter.create<quantum::MeasureOp>(qubit.getLoc(),
+                                               /*mres=*/rewriter.getI1Type(),
+                                               /*out_qubits=*/qubit.getType(),
+                                               /*in_qubits=*/qubit,
+                                               /*postselect=*/nullptr);
+}
+
+// Applies per-qubit conjugations that map the provided Pauli string to the Z
+// basis. If `isReverse` is true, applies the inverse mapping.
+void constructPauliConjugationLayer(SmallVector<Value> &qubits, SmallVector<StringRef> pauliString,
```
rniczh:
```suggestion
void constructPauliConjugationLayer(SmallVector<Value> &qubits, ArrayRef<StringRef> pauliString,
```

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
```diff
@@ -0,0 +1,238 @@
...
+//
+// Returns success after replacing `op` with the produced SSA values.
+//
+// LLVM_ATTRIBUTE_USED is used to suppress the warning about the function being unused.
+LLVM_ATTRIBUTE_USED LogicalResult convertToMBQC(QECOpInterface op,
+                                                ConversionPatternRewriter &rewriter)
+{
+    Value measResult;
+    auto pauliString = extractPauliString(op);
+    SmallVector<Value> qubits = op.getInQubits();
```
rniczh, line 191: I think it's not promise to has at least one qubit in QEC Dialect right? from `QECDialect.td`, it's allow to have zero elements.
```
Variadic<QubitType>:$in_qubits,  // The qubits to apply the rotation to
```
You can add the limitation on the dialect or add an validation here to ensure no 0 qubit happen when transforming.
```
    if (qubits.empty()) {
        return emitError(op.getLoc(), "QEC operation must have at least one input qubit");
    }
    
    if (pauliString.empty()) {
        return emitError(op.getLoc(), "QEC operation must have a Pauli product");
    }
    
    if (qubits.size() != pauliString.size()) {
        return emitError(op.getLoc(), "Number of qubits must match Pauli product length");
    }
```

frontend/catalyst/passes/builtin_passes.py
sengthai: Good question! It is not compatible since the `ppr_to_ppm` involves a multiplex PPR op, and we don't want it to be more complicated for now.

frontend/catalyst/passes/builtin_passes.py
sengthai: Good point!

doc/releases/changelog-dev.md
sengthai: Nice one! Thanks, Paul!

frontend/catalyst/passes/builtin_passes.py
sengthai: The reason is that the decomposition gates are targeting the subset of universal gates that are specifically supported in the MBQC workload. These universal gates are kind of the basis gates for MBQC. 

You remind me to update this docstring to clarify that we target the MBQC gate set here.

mlir/lib/QEC/Transforms/ppr_to_mbqc.cpp
sengthai: Ohh nice point! Thanks!

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
sengthai: Sweet!

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
sengthai: You are right!

doc/releases/changelog-dev.md
sengthai: >I think all mbqc entries go into the internal changes section now because that project is not user facing yet.

We also expose this pass to the user as well. So, it's fine to put in new feature section.

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
sengthai: I think it is a good idea to verify them when initializing it. Example for PPR we have this:
```C++
// mlir/lib/QEC/IR/QECDialect.cpp
LogicalResult PPRotationOp::verify()
{

    size_t numPauliProduct = getPauliProduct().size();

    if (numPauliProduct == 0) {
        return emitOpError("Pauli string must be non-empty");
    }

    if (numPauliProduct != getInQubits().size()) {
        return emitOpError("Number of qubits must match number of pauli operators");
    }
    return mlir::success();
}
```

However, I'm not sure if there is no op verifier for `CustomOp`. Now I'm creating a new verifier for CustomOp as well.

doc/releases/changelog-dev.md
sengthai: [254d8bf](https://github.com/PennyLaneAI/catalyst/pull/2057/commits/254d8bf70262f2e668b99798365dcd06599acd6a)

frontend/catalyst/passes/builtin_passes.py
sengthai: [254d8bf](https://github.com/PennyLaneAI/catalyst/pull/2057/commits/254d8bf70262f2e668b99798365dcd06599acd6a)

frontend/catalyst/passes/builtin_passes.py
sengthai: >add the circuit diagrams for the decomposition to the PR description.

Adding the circuit diagrams would take a time for this large circuit. 😅 
So, I will add examples for that decomposition instead.

[254d8bf](https://github.com/PennyLaneAI/catalyst/pull/2057/commits/254d8bf70262f2e668b99798365dcd06599acd6a)

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
rniczh: I'm not sure, what kinds of verify needed for CustomOp?

sengthai: This verifier: https://github.com/PennyLaneAI/catalyst/blob/0c496da057957db78556567fb6166dfb48f90c01/mlir/lib/Quantum/IR/QuantumOps.cpp#L181

rniczh: I mean what you want to create for CustomOp?

rniczh: Oh, sorry, it's your added XD

rniczh: Yup, you can add it as well 👍

sengthai: I built the CustomOp for representing the gates that are supported in MBQC. And before we built that `::verify` will be performed.

sengthai: hahah okay

doc/releases/changelog-dev.md
paul0403: Sorry, so it's still in the new features section? I think it should be either:

- ppr to quantum, and new feature section, or
- ppr to mbqc, and internal changes section

This is because mbqc is not a user facing feature.

paul0403: If this pass is user facing, then this pass is not just for mbqc, and should be named something else.

paul0403: I know the mbqc _team_ will be the main user of this pass, but that's just the user. This pass itself is just lowering ppr dialect into the quantum dialect right?

sengthai: >Sorry, so it's still in the new features section? 

I mentioned above.

> I know the mbqc team will be the main user of this pass, but that's just the user. This pass itself is just lowering ppr dialect into the quantum dialect right?

It's hard to tell since this gate rule is based on paper as well.

paul0403: @mlxd @isaacdevlugt @dime10 was a decision on this made? To me, the scope of this pass (which converts ppr dialect ops to quantum dialect ops for mock execution) is a bit fuzzy, and we should properly segment the user impacts. 

This pass does the following:
- convert PPR/PPM dialect ops into quantum dialect ops
- however, a particular op in its output `quantum.custom "RotXZX"` does not have a runtime lowering yet, and hence is not executable end-to-end.  

So the question is, is this pass a new user facing feature? And is this pass for any user, or is this pass just for our in-house mbqc team?

I remember @josh146 saying "new feature" section in changelog should be used for completed new features that the user can try out. Putting "mbqc" support in this section implies that users can already try out the mbqc pipeline end-to-end, which is not true. This is also why all the mbqc sections are put under "internal changes" for now.

So my position is, if this is a mbqc only pass, then it should go into internal changes, not new features.

Then, is this a mbqc only pass? The existence of the rot xzx gate seems to imply so, but if we also decide to add runtime stub for that gate (or decomposition of that gate onto other "usual" gates on devices that are not null qubit, since e.g. lightning will not have a method for this gate I suppose), then execution of PPR dialect can actually now happen end-to-end, and this will indeed be a new feature. But then in this case this pass is not mbqc specific: it is just converting ppm dialect to quantum dialect now, and should not be named mbqc.

So to summarize, my opinion is
- either, we don't add the backend support for rotxzx, and keep the pass name as mbqc, and put it under internal changes
- or, we add the backend support for rotxzx, change the pass name to just `-ppr-to-quantum`, and put it under new feature.

Sorry for causing all this trouble, but I think properly defining feature boundaries is important. I don't know if any decision was already made here; if there already was a decision then feel free to ignore this and go with the decision.

sengthai: >however, a particular op in its output quantum.custom "RotXZX" does not have a runtime lowering yet, and hence is not executable end-to-end.

@paul0403 You got the point that we haven't supported RotXZX yet. So, It appears that it needs another pass to lower into MBQC formalism for executions. For this reason, I agree to put this into "Internal Change". However, if we do want and can support executable RotXZX, it will still be good in the new feature section.

josh146: I don't have a strong opinion here, and happy to defer to others.

> I remember @josh146 saying "new feature" section in changelog should be used for completed new features that the user can try out.

this is correct. Something to note is that the new features section does not distinguish between external and internal users -- it applies to new features that can be used by _any_ user, whether external _or_ internal.

As long as the release note accurately states what can/can't be done with this feature, that is fine.

paul0403: Thanks for the responses, sounds like we decided to keep mbqc in the pass name and put it under internal changes :+1: 

We can put PPR execution into new features when end-to-end execution is completely. That work may require another pass to lower ppm to quantum dialect (one that does not lower to rotxzx) and can leave the mbqc pass separate. But that is for another day.

paul0403: > As long as the release note accurately states what can/can't be done with this feature, that is fine.

@sengthai remember to state in changelog that this pass still does not support end-to-end execution of ppm dialect.

mlir/lib/QEC/Transforms/CMakeLists.txt
```diff
@@ -17,6 +17,8 @@ file(GLOB SRC
     ppm_compilation.cpp
     PartitionLayers.cpp
     TLayerReduction.cpp
+    ppr_to_mbqc.cpp
+    PPRToMBQC.cpp
```
mlxd, line 21: Feel free to ignore, but just wondering why we have mixed snake_case and CamelCase in this CMake glob? Do we expect to use both?

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
```diff
@@ -0,0 +1,247 @@
...
+            // Y Pauli, and it would be tedious to write a function to generate them.
+            // ─────RotXZX(11.33,0.00,2.80)────RZ(PHI/2)────RotXZX(1.23,0.00,9.76)────
+            auto params = !isReverse ? std::array<double, 3>({11.33, 0.00, 2.80})
+                                     : std::array<double, 3>({1.23, 0.00, 9.76});
+            qubits[idx] = buildSingleQubitGate(qubits[idx], "RotXZX", params, rewriter)
+                              .getOutQubits()
+                              .front();
+            break;
+        }
+        case 'I':
```
mlxd, line 102: Can we add the https://en.cppreference.com/w/cpp/language/attributes/fallthrough.html statement here for clarity?

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
```diff
@@ -0,0 +1,247 @@
...
+//  - PPR XX(π/4) [0, 1]: H(0) • H(1) • CNOT(1, 0) • RZ(π/2)   • CNOT(1, 0) • H(0) • H(1)
+//  - PPR ZZ(π/8) [0, 1]:               CNOT(1, 0) • RZ(π/4)   • CNOT(1, 0)
+//  - PPR YZ(π/8) [0, 1]: RotXZX(0)   • CNOT(1, 0) • RZ(π/4)   • CNOT(1, 0) • RotXZX(0)†
+//  - PPM XX      [0, 1]: H(0) • H(1) • CNOT(1, 0) • Measure(0)• CNOT(1, 0) • H(0) • H(1)
+//  - PPM ZZ      [0, 1]:               CNOT(1, 0) • Measure(0)• CNOT(1, 0)
+//  - PPM YZ      [0, 1]: RotXZX(0)   • CNOT(1, 0) • Measure(0)• CNOT(1, 0) • RotXZX(0)†
+//
+// LLVM_ATTRIBUTE_USED is used to suppress the warning about the function being unused.
+//
+LLVM_ATTRIBUTE_USED LogicalResult convertToMBQC(QECOpInterface op,
```
mlxd: Would https://en.cppreference.com/w/cpp/language/attributes/maybe_unused.html be a better fit here, or is the `LLVM_ATTRIBUTE_USED` a better fit? I'd prefer the C++ standard option if it is semantically correct, but fine by me to use the LLVM option if that is more correct.

mlxd (commented):
Nice work on this @sengthai 
Just a few quick q's, but happy to defer to others for the final review & approval.

mlir/lib/QEC/Transforms/CMakeLists.txt
paul0403: I have noticed that we use snake case for the files that are the actual passes, and camel case for the files that are the rewrite patterns...

I don't know if it's intentional or not

mlir/lib/QEC/Transforms/ppr_to_mbqc.cpp
```diff
@@ -0,0 +1,74 @@
...
+    void runOnOperation() final
+    {
+        auto ctx = &getContext();
+        ConversionTarget target(*ctx);
+
+        target.addIllegalOp<qec::PPRotationOp>();
+        target.addIllegalOp<qec::PPMeasurementOp>();
+
+        // The Dialects that we want to convert to
+        target.addLegalDialect<catalyst::quantum::QuantumDialect>();
```
multiphaseCFD, line 51: Wondering if the LegalDialect of the target is empty until we add some dialects to it?

mlir/lib/QEC/Transforms/CMakeLists.txt
mlxd: May be good for us to agree on the approach and document it somewhere.

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
```diff
@@ -0,0 +1,247 @@
...
+using namespace catalyst;
+using namespace catalyst::quantum;
+using namespace catalyst::qec;
+
+namespace {
+
+CustomOp buildCNOTGate(Value control, Value target, ConversionPatternRewriter &rewriter)
+{
+    return rewriter.create<quantum::CustomOp>(
+        control.getLoc(),
```
multiphaseCFD, line 32: Could we add some comments to this arg `control.getLoc()`? Since it's not an arg required by the `quantum::CustomOp` operation.

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
```diff
@@ -0,0 +1,247 @@
...
+        /*out_ctrl_qubits=*/mlir::TypeRange({}),
+        /*params=*/mlir::ValueRange(),
+        /*in_qubits=*/mlir::ValueRange({control, target}),
+        /*gate_name=*/"CNOT",
+        /*adjoint=*/false,
+        /*in_ctrl_qubits=*/mlir::ValueRange({}),
+        /*in_ctrl_values=*/mlir::ValueRange());
+}
+
+CustomOp buildSingleQubitGate(Value qubit, StringRef gateName, ArrayRef<double> params,
```
multiphaseCFD, line 43: Is `ArrayRef` here a SSA Value type or STL container derived type?

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
```diff
@@ -0,0 +1,247 @@
...
+                                              /*in_ctrl_values=*/mlir::ValueRange());
+}
+
+MeasureOp buildMeasurementOp(Value qubit, ConversionPatternRewriter &rewriter)
+{
+    return rewriter.create<quantum::MeasureOp>(qubit.getLoc(),
+                                               /*mres=*/rewriter.getI1Type(),
+                                               /*out_qubits=*/qubit.getType(),
+                                               /*in_qubits=*/qubit,
+                                               /*postselect=*/nullptr);
```
multiphaseCFD, line 71: Is it best practice to set `postselect`  as `nullptr`?

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
```diff
@@ -0,0 +1,247 @@
...
+            break;
+        }
+        }
+    }
+}
+
+// Applies a right-to-left CNOT ladder to accumulate parity onto `qubits[0]`.
+void constructCNOTLadder(SmallVector<Value> &qubits, ConversionPatternRewriter &rewriter)
+{
+    size_t numQubits = qubits.size();
```
multiphaseCFD, line 115: This method entangles two qubits with adjacent idx in qubits with CNOT for the qubits[-1] to qubits[0], right? Is there any difference if we build this ladder from [0] to [-1]?  I'm also wondering why we need to build this ladder?

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
```diff
@@ -0,0 +1,247 @@
...
+        qubits[i] = outQubits[0];
+        qubits[i - 1] = outQubits[1];
+    }
+}
+
+// Reverses the CNOT ladder to uncompute parity accumulation.
+void constructReverseCNOTLadder(SmallVector<Value> &qubits, ConversionPatternRewriter &rewriter)
+{
+    size_t numQubits = qubits.size();
+    for (size_t i = 0; i + 1 < numQubits; i++) {
```
multiphaseCFD: Okay, the reverse method should match the original build method. 
What about
`i < qubits.size() - 1`

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
```diff
@@ -0,0 +1,247 @@
...
+    }
+}
+
+// Emits the kernel corresponding to `op`:
+//  - `PPMeasurementOp`: measures `qubits[0]` and assigns the i1 to `measResult`.
+//  - `PPRotationOp`: applies an RZ on `qubits[0]` with angle from `getRotationKind()`.
+void constructKernelOperation(SmallVector<Value> &qubits, Value &measResult, QECOpInterface op,
+                              ConversionPatternRewriter &rewriter)
+{
+    if (isa<PPMeasurementOp>(op)) {
```
multiphaseCFD, line 140: i assume `isa<>()` is a MLIR equivalent to python `isinstance`?

mlir/lib/QEC/Transforms/CMakeLists.txt
sengthai: > snake case for the files that are the actual passes, and camel case for the files that are the rewrite patterns

@paul0403 is right! I saw that pattern files since I joined. I thought it was the norm. 

> May be good for us to agree on the approach and document it somewhere.

I agree with @mlxd that we should approach it in the developer's document.

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
```diff
@@ -0,0 +1,247 @@
...
+        auto measOp = buildMeasurementOp(qubits[0], rewriter);
+        measResult = measOp.getMres();
+        qubits[0] = measOp.getOutQubit();
+    }
+    else {
+        int16_t signedRk = static_cast<int16_t>(op.getRotationKind());
```
multiphaseCFD, line 146: any reason why we use `int16_t` here? why not cast is to double directly for later usage?

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
sengthai: I will do! It would be great!

multiphaseCFD (commented):
Thanks @sengthai for the nice work. Just a few comments from my end for now.

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
paul0403: This is one of the quirks of mlir capi :sweat_smile: all ops need a "location". I guess xdsl does not need it? But yeah it's a standard thing so no worries here I'd say.

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
paul0403: `ArrayRef` can be used on arbitrary types, not just mlir capi types.

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/blob/main/mlir/include/Quantum/IR/QuantumOps.td#L958 it's an optional attribute, so nullptr says it's not there on this instance of the measure op.

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
multiphaseCFD: Thanks @paul0403 . Is the corresponding arg passed by value or ref?

paul0403: `ArrayRef` is a lightweight object that is non owning, so no copying anywhere. It's just const reference.

It actually is just the starting pointer and length of array https://llvm.org/doxygen/classllvm_1_1ArrayRef.html#details

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
paul0403: yes

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
sengthai: Yes, using `LLVM_ATTRIBUTE_USED` seems overkill here, since we only want to suppress the warning. So, using `[[maybe_unused]]` should be better.

sengthai: I will use `[[maybe_unused]]` instead.

rniczh: @mlxd I think `[[maybe_unused]]` has different semantic from `LLVM_ATTRIBUTE_USED`. `LLVM_ATTRIBUTE_USED` is used when this function is called by external, or through indirect call(ptr), and you want to guarantee it exists in the final binary, and avoid DCE. However, in this case, the reason of why compiler show the warning just because the caller is template function, it's not yet be specialized when checking the unused function/variable. In your understanding, which is more better fit?

sengthai: @rniczh agree! Since we don't have any TableGen or JIT that looks up for this function, `[[maybe_used]]` should be fine.

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
sengthai: > Is there any difference if we build this ladder from [0] to [-1]?

Good question! The difference is that build the ladder with qubits[-1] to qubits[0], we don't need to handle the result of that op by replacing it results value every CNOT. 


> I'm also wondering why we need to build this ladder?

The ladder is part of the decomposition of a PPR into a standard circuit. It’s essentially the structure that makes the decomposition possible. In principle, the necessity could be proven mathematically, but I don’t have that proof yet. 😅 We can reference to this paper: https://arxiv.org/pdf/2509.20052 


Screenshot 2025-09-26 at 4 35 44 PM

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
sengthai: Good cache! It would work!

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
mlxd: Happy to go with whatever you prefer here.

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
sengthai: getRotationKind() is defined to return an `uint16_t`. So if we convert directly to `double`, we lose the sign.

mlir/lib/QEC/Transforms/ppr_to_mbqc.cpp
sengthai: Yes you are right! And adding legal dialect is important if we want to build the op from that specific dialect.

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
sengthai: [d7a8e97](https://github.com/PennyLaneAI/catalyst/pull/2057/commits/d7a8e977a02e090b57ea505b01bca49f8cc8d8c5)

doc/releases/changelog-dev.md
sengthai: 👨‍💻
[d7a8e97](https://github.com/PennyLaneAI/catalyst/pull/2057/commits/d7a8e977a02e090b57ea505b01bca49f8cc8d8c5)

PR #2058, opened 2025-09-19
Fix conversion target rule in `to_ppr` and `ppm_compilation` pass

sengthai:
**Context:**
Updates the quantum gate conversion logic to add support for the identity gate (`I`) in the Clifford+T to PPR decomposition passes, and refactors the conversion targets. The main changes include extending gate handling to cover `I`, updating conversion targets to list both illegal and legal operations explicitly, and improving error messages.

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -136,7 +136,7 @@
 * Added support to avoid Y-basis measurements in `pauli-corrected` PPR decomposition.
   [(#2047)](https://github.com/PennyLaneAI/catalyst/pull/2047)
 
-* Update conversion targets and refactor `to_ppr` and `ppm_compilation` passes to handle gate `I`.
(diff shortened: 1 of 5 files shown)


paul0403 (commented):
Two things I'm wondering:

1. If the idea is just to replace identity gates with zero-angle rotations, why not just ignore the identity gate in the first place?
2. As I was looking through the code, it seems like the `applySingleQubitConversion` function can take in a const reference for the custom gate conversion structs, since it is being used read-only? i.e. 
```C
void applySingleQubitConversion(CustomOp op, 
   const SmallVector<GateConversion> &gateConversions,
                                ConversionPatternRewriter &rewriter)
```

rniczh:
@sengthai 
> Nice feedback! Yes, it is read-only. So using const is the best!
How about `ArrayRef` ?

PR #2068, opened 2025-09-30
Add support of MCM for decompose pass

rniczh:
**Context:**

Added traversal logic of mcm. For an example, finding the `qreg` and index of `qb3` in the following case, it needs to traverse the def-use chain through the ops.

```
qb0 = extract(...)
...
meas_result, qb2 = qml.measure(qb1, 0)
...
qb4 = op(qb3, ...) 
```

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-100311]

### DIFF
--- a/mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
+++ b/mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
@@ -346,6 +346,9 @@ class OpSignatureAnalyzer {
                 }
                 qubit = customOp.getQubitOperands()[0];
             }
+            else if (auto measureOp = dyn_cast_or_null<quantum::MeasureOp>(qubit.getDefiningOp())) {
+                qubit = measureOp.getInQubit();
+            }
         }
 
         return nullptr;
@@ -376,6 +379,9 @@ class OpSignatureAnalyzer {
                         continue;
                     }
                 }
+            }
+            else if (auto measureOp = dyn_cast_or_null<quantum::MeasureOp>(qubit.getDefiningOp())) {
+                qubit = measureOp.getInQubit();
             }
 
             break;


mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
```diff
@@ -346,6 +346,9 @@ class OpSignatureAnalyzer {
                 }
                 qubit = customOp.getQubitOperands()[0];
             }
+            else if (auto measureOp = dyn_cast_or_null<quantum::MeasureOp>(qubit.getDefiningOp())) {
```
dime10: We should use interfaces whenever we can, which might save you from having to add all these instructions one by one :) Unfortunately, measure doesn't seem to implement any of our interfaces, but the CustomOp above may be better served with the [QuantumOperation](https://github.com/PennyLaneAI/catalyst/blob/07a23c992207ab1f7c19a11128d703c2457f7804/mlir/include/Quantum/IR/QuantumInterfaces.td#L56) or [QuantumGate](https://github.com/PennyLaneAI/catalyst/blob/07a23c992207ab1f7c19a11128d703c2457f7804/mlir/include/Quantum/IR/QuantumInterfaces.td#L92) interface.

Although I'm a little confused why we just get the 0th qubit operand, are we guaranteed to only deal with single-qubit operations here?

dime10: Also, why not get the index and register together in one walk, rather than in separate functions? 🤔

rniczh: > Also, why not get the index and register together in one walk, rather than in separate functions?

Yes, I agreed, I updated this in another PR  https://github.com/PennyLaneAI/catalyst/pull/2074, but it might take more time to handle that one, so I can move that change to this PR

rniczh: > Although I'm a little confused why we just get the 0th qubit operand, are we guaranteed to only deal with single-qubit operations here?

It's because this function here is used to get the source qreg, at the point, only one qreg, the global one exists. So traverse different qubit operands would still reach the same qreg. But now we have dynamic qreg! So this method will need to be changed and follow the solution from https://github.com/PennyLaneAI/catalyst/pull/2074

dime10: Makes sense 👍

rniczh: @dime10 Thanks for the suggestions, I changed to use the `QuantumGate` to traverse the gates instead of `CustomOp` and also use only one walk to collect the qreg and qubit index in the latest change.https://github.com/PennyLaneAI/catalyst/pull/2068/commits/d01ef10b237b8e689de87ec478831c16bd5c4064

doc/releases/changelog-0.13.0.md
maliasadi, line 1: Could you add this PR to the changelog?

frontend/test/pytest/from_plxpr/test_capture_integration.py
```diff
@@ -1313,6 +1313,38 @@ def circuit(x: float, y: float, z: float):
...
+        # Capture enabled
+
+        qml.capture.enable()
+        qml.decomposition.enable_graph()
+
+        @qjit(target="mlir")
+        @partial(qml.transforms.decompose, gate_set=[qml.RX, qml.RY, qml.RZ])
+        @qml.qnode(qml.device(backend, wires=2))
+        def captured_circuit(x: float, y: float, z: float):
+            qml.measure(0)
```
maliasadi: could we update the test to post-process the result of the measure in `qml.cond` and conditionally add a gate? The graph itself doesn't fully support `cond`s so decomposing gates inside `qml.cond` are not yet in the scope.

mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
```diff
@@ -453,4 +434,4 @@ void populateDecomposeLoweringPatterns(RewritePatternSet &patterns,
 }
 
 } // namespace quantum
-} // namespace catalyst
+} // namespace catalyst
```
maliasadi:
```suggestion
} // namespace catalyst

```

doc/releases/changelog-0.13.0.md
rniczh: Sure, I added it 👍

frontend/test/pytest/from_plxpr/test_capture_integration.py
rniczh: No problem, I added to the https://github.com/PennyLaneAI/catalyst/pull/2068/commits/65cdd45bd97d33b6c4acb378724790923b1fe001

doc/releases/changelog-0.13.0.md
```diff
@@ -13,13 +13,14 @@
   [(#2091)](https://github.com/PennyLaneAI/catalyst/pull/2091)
   [(#2029)](https://github.com/PennyLaneAI/catalyst/pull/2029)
   [(#2001)](https://github.com/PennyLaneAI/catalyst/pull/2001)
+  [(#2068)](https://github.com/PennyLaneAI/catalyst/pull/2068)
```
maliasadi, line 14:
```suggestion
  [(#2091)](https://github.com/PennyLaneAI/catalyst/pull/2091)
  [(#2068)](https://github.com/PennyLaneAI/catalyst/pull/2068)
  [(#2029)](https://github.com/PennyLaneAI/catalyst/pull/2029)
  [(#2001)](https://github.com/PennyLaneAI/catalyst/pull/2001)
```

doc/releases/changelog-0.13.0.md
```diff
@@ -11,6 +11,7 @@
   and a ``UserWarning`` is raised.
   [(#2099)](https://github.com/PennyLaneAI/catalyst/pull/2099)
   [(#2091)](https://github.com/PennyLaneAI/catalyst/pull/2091)
+  [(#2068)](https://github.com/PennyLaneAI/catalyst/pull/2068)
```
sengthai: Should we sort these PRs by #number?

mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
```diff
@@ -117,13 +114,23 @@ class OpSignatureAnalyzer {
...
         signature.outCtrlQubitIndices = signature.inCtrlWireIndices;
     }
 
     operator bool() const { return isValid; }
 
+    Value getUpdatedQreg(PatternRewriter &rewriter, Location loc)
+    {
+        // FIXME: This will cause an issue when the decomposition function has cross-qreg
+        // inputs and outputs. Now, we just assume has only one qreg input, the global one exists.
+        return signature.inWireIndices[0].getReg();
```
sengthai: How about we give an error message when multiple `qreg` appear?

mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
```diff
@@ -54,17 +56,21 @@ namespace quantum {
 ///       Value idx = dynamicIdx.getValue();   // Get the Value
 ///     }
 ///   }
-struct QubitIndex {
```
sengthai, line 57: I wonder what drove to switch from `struct` to `class`?  🤔

rniczh: That's because I don't want to expose these two members in the QubitIndex. So I'm thinking, probably, change to use `class` would be much more make sense here.
```cpp
    std::variant<std::monostate, Value, IntegerAttr> index;
    Value qreg;
```

mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
rniczh: Sure,  I added it in the https://github.com/PennyLaneAI/catalyst/pull/2068/commits/1ad5c8b94ac9115a64d645f4e504da6f22151702

doc/releases/changelog-0.13.0.md
rniczh: Sure, https://github.com/PennyLaneAI/catalyst/pull/2068/commits/1ad5c8b94ac9115a64d645f4e504da6f22151702

mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
sengthai: Ahh I see. does `struct` allow private properties as well right? 🤔

rniczh: Yes, but should avoiding to do this
Here is an section in llvm coding standards talked about this thing, if you need `private`, should using `class` instead
https://llvm.org/docs/CodingStandards.html#use-of-class-and-struct-keywords

sengthai: Good to know!

mlxd: Also in our C++ guidance docs :)

PR #2069, opened 2025-09-30
Minor changelog changes

isaacdevlugt:
[sc-100377](https://app.shortcut.com/xanaduai/story/100377/initial-changelog-ordering-for-pl)

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -87,64 +87,62 @@
 
 * A new quantum compilation pass that reduces the depth and count of non-Clifford Pauli product
   rotations (PPRs) in circuits is now available. This compilation pass works by commuting
-  non-Clifford PPRs (often referred to as ``T`` gates) in adjacent
+  non-Clifford PPRs (often just referred to as ``T`` gates) in adjacent
   layers and merging compatible ones. More details can be found in Figure 6 of
   [A Game of Surface Codes](https://arXiv:1808.02892v3).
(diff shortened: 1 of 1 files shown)


doc/releases/changelog-dev.md
```diff
@@ -87,65 +87,63 @@
...
+      "t_layer_reduction": {}
+  }
 
-  After performing the ``catalyst.passes.to_ppr`` and ``catalyst.passes.merge_ppr_ppm``
-  passes, the circuit contains a depth of four of non-Clifford PPRs. Subsequently applying the
-  ``t_layer_reduction`` pass will move PPRs around via commutation, resulting in a circuit with a
-  smaller PPR depth of three.
+  for pipeline in [reduce_T, no_reduce_T]:
+
+      @qjit(pipelines=pips, target="mlir", circuit_transform_pipeline=pipeline)
```
paul0403, line 85: I've been extremely concerned with having two separate "pipeline" kwargs for a long time, I would think this is very confusing.... but I guess at the moment there's no better way to do it if we don't want to stack decorators...

Or maybe we can change their names? But I can't think of any good names of the top of my head.

isaacdevlugt: @paul0403 yes! My understanding is that the `pipelines` arg is a bit more developer/power-user centric, while `circuit_transform_pipeline` is for most users. 

We have work planned in Q4 to make specifying compilation pipelines uniform in PL and Catalyst that should address the confusion 🚀

PR #2072, opened 2025-09-30
Remove `qml.capture.disable()` from changelog code examples

paul0403:
**Context:**
The common usage of capture is to enable it for a session at the top and then use it throughout, so unless the user explicitly needs both modes we should not recommend the disable function.

Remove `qml.capture.disable()` from changelog code examples


paul0403:
closing in favor of #2073

PR #2075, opened 2025-10-01
Better error messages for dynamic qubit allocation

### LINKED ISSUE #2071: Stress test for Dynamic Qubit Allocation feature [v0.13]
### Issue 1: Implicit wires in terminal MP raises an error

```py
qml.capture.enable()

@qjit
@qml.qnode(qml.device("lightning.qubit", wires=2))  # 2 initial qubits
def circuit():
    qml.X(0)                        # |10>

    with qml.allocate(1) as q:      # |10> and |0>, 1 dynamically allocated qubit
        qml.X(q[0])                 # |10> and |1>
        qml.CNOT(wires=[q[0], 1])   # |11> and |1>

    return qml.probs()

>>> circuit()
RuntimeError: [LightningSimulator.cpp:322][Function:Probs] Error in Catalyst Runtime: Invalid size for the pre-allocated probabilities
```

Possibly a restriction in the current feature implementation, but it does not appear to be documented in release notes or documentation.

The second issue is the error message which a) only happens at runtime, and b) does not explicitly point to the problem.

---

paul0403:
**Context:**
There are some error cases where we should issue a better, more actionable error message regarding the dynamic allocation feature.

Add specific error messages for the following error cases:
- Using dynamic allocation without program capture
- Implicit wires in terminal MP
- Dynamically allocated wires in terminal MP
- Use-after-free
- Subroutines

**Benefits:**
Better error message.

**Related GitHub Issues:**
closes #2071

[sc-100475]

### DIFF
--- a/.github/workflows/check-catalyst.yaml
+++ b/.github/workflows/check-catalyst.yaml
@@ -576,7 +576,8 @@ jobs:
     - name: Install PennyLane-Lightning-Kokkos release candidate
       if: github.event.pull_request.base.ref == 'v0.13.0-rc'
       run: |
-        pip install --no-deps --pre -U --index-url https://test.pypi.org/simple/ pennylane-lightning-kokkos'<=0.43.0'
+        # Need to install the latest versions of both lightning.qubit and lightning.kokkos
+        pip install --no-deps --pre -U --index-url https://test.pypi.org/simple/ pennylane-lightning'<=0.43.0' pennylane-lightning-kokkos'<=0.43.0'
 
     - name: Get Cached LLVM Build
       id: cache-llvm-build


frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -594,6 +628,17 @@ def handle_subroutine(self, *args, **kwargs):
     """
     Transform the subroutine from PLxPR into JAXPR with quantum primitives.
     """
+
+    if any(is_dynamically_allocated_wire(arg) for arg in args):
+        raise NotImplementedError(
+            textwrap.dedent(
+                """
```
rniczh, line 648:
```suggestion
            """
```

paul0403: This is actually what black (the formatter) wants :sweat_smile: I don't know why either

rniczh: Got it XD

frontend/test/pytest/test_dynamic_qubit_allocation.py
```diff
@@ -299,6 +302,96 @@ def circuit(N):
...
+            with qml.allocate(1) as _:
+                pass
+            return qml.probs(wires=[0])
+
+
+def test_use_after_free(backend):
+    """
+    Test error message when used after free.
+    """
+    qml.capture.enable()
```
dime10: Always use the test fixture to guarantee capture state is restored to default even if the test fails.

frontend/catalyst/jax_tracer.py
```diff
@@ -1454,6 +1454,10 @@ def _trace_classical_phase(
         # Therefore we need to compute the tree with measurements as leaves and it comes
         # with an extra computational cost
 
+        if any(isinstance(wire, qml.wires.DynamicWire) for wire in quantum_tape.wires):
```
dime10, line 1457: That's interesting, how do the wires make their way onto the tape in your test example which doesn't use the wires anywhere?

frontend/test/lit/test_dynamic_qubit_allocation.py
```diff
@@ -89,7 +89,7 @@ def test_basic_dynalloc():
         qml.X(qs1[1])
         qml.CNOT(wires=[qs1[2], 1])
 
-    return qml.probs()
+    return qml.probs(wires=[0])
```
dime10, line 92: Did this test not raise an error without the argument?

paul0403: The underlying error was a runtime error, and this was a lit test (so it had target=mlir)

frontend/test/pytest/test_dynamic_qubit_allocation.py
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2075/commits/7f961e2049608aa76bf95cb511daa3dd010bc366 I forgot it existed :sweat_smile: thanks for the reminder!

frontend/catalyst/jax_tracer.py
paul0403: PL stores them all beforehand, and then resolves them at runtime.

But now that you mention it, maybe it's better to check the actual qml.allocate() op when we encounter it instead of relying on the tape. Let me see if this is possible.

dime10: Oh, didn't know allocate generated an instruction on the tape :O 

I think it's alright then, the solution you have makes sense 👍

PR #2078, opened 2025-10-01
[Capture] handle grad and jacobian

albi3ro:
**Context:**

Handles the `grad` primitive from pennylane and adds a lowering for it.

This requires PennyLane PR: https://github.com/PennyLaneAI/pennylane/pull/8357

Once that merges in, we can retarget this branch against the newer pennylane.

Instead of translating the pennylane grad prim to the catalyst one, we just register a lowering for the grad primitive. We also have to register a translation rule where we are translating the target jaxpr.

**Benefits:**

Can handle grad and jacobian calls.

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-100560]

### DIFF
--- a/frontend/catalyst/from_plxpr/from_plxpr.py
+++ b/frontend/catalyst/from_plxpr/from_plxpr.py
@@ -164,7 +164,9 @@ def handle_grad(self, *args, jaxpr, n_consts, **kwargs):
     new_jaxpr = jax.make_jaxpr(f)(*args[n_consts:])
 
     new_args = (*new_jaxpr.consts, *args[n_consts:])
-    return pl_jac_prim.bind(*new_args, jaxpr=new_jaxpr.jaxpr, n_consts=len(new_jaxpr.consts), **kwargs)
+    return pl_jac_prim.bind(
+        *new_args, jaxpr=new_jaxpr.jaxpr, n_consts=len(new_jaxpr.consts), **kwargs
+    )
 
 
 # pylint: disable=unused-argument, too-many-arguments


frontend/catalyst/jax_primitives_utils.py
```diff
@@ -75,7 +105,7 @@ def lower_callable(ctx, callable_, call_jaxpr, pipeline=None, context=None, publ
     add more than one FuncOps. This depends on the contents of call_jaxpr.
 
     Args:
-      ctx: LoweringRuleContext
+      ctx: LoweringRulemetadata
```
albi3ro:
```suggestion
      ctx: LoweringRuleContext
```

frontend/catalyst/jax_primitives.py
```diff
@@ -692,6 +690,33 @@ def _grad_lowering(ctx, *args, jaxpr, fn, grad_params):
...
 
 
+# pylint: disable=too-many-arguments
+def _capture_grad_lowering(ctx, *args, argnums, jaxpr, n_consts, method, h, fn, scalar_out):
+    mlir_ctx = ctx.module_context.context
+    if h:
+        f64 = ir.F64Type.get(mlir_ctx)
+        finiteDiffParam = ir.FloatAttr.get(f64, h)
+    else:
+        finiteDiffParam = None
```
albi3ro:
```suggestion
    f64 = ir.F64Type.get(mlir_ctx)
    finiteDiffParam = ir.FloatAttr.get(f64, h)
```

.github/workflows/check-catalyst.yaml
```diff
@@ -475,6 +475,10 @@ jobs:
         python3 -m pip install oqc-qcaas-client
         make frontend
 
+    - name: Install PennyLane branch
+      run: |
+        pip install --no-deps --force git+https://github.com/PennyLaneAI/pennylane@add-fn-to-grad-prim
```
paul0403: Are there any mutual dependencies between the two? If not, is it ok to wait for the PL one to get merged first, then just update the PL dep's dev version?

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -399,7 +409,7 @@ def trace_from_pennylane(
             fn.static_argnums = static_argnums
 
         plxpr, out_type, out_treedef = make_jaxpr2(fn, **make_jaxpr_kwargs)(*args, **kwargs)
-        jaxpr = from_plxpr(plxpr)(*dynamic_args, **kwargs)
+        jaxpr = from_plxpr(plxpr)(*plxpr.in_avals)
```
paul0403, line 425: Can you briefly explain why this change was introduced?

frontend/catalyst/jax_primitives.py
```diff
@@ -692,6 +690,30 @@ def _grad_lowering(ctx, *args, jaxpr, fn, grad_params):
     ).results
 
 
+# pylint: disable=too-many-arguments
+def _capture_grad_lowering(ctx, *args, argnums, jaxpr, n_consts, method, h, fn, scalar_out):
```
paul0403, line 694: Does this mean we will have a mix of PL primitives and catalyst primitives in catalyst jaxpr now (if the user is using grad)?

I'll defer to @dime10 as to whether this is ok.

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -27,6 +27,7 @@
 from jax.extend.linear_util import wrap_init
 from pennylane.capture import PlxprInterpreter, qnode_prim
 from pennylane.capture.expand_transforms import ExpandTransformsInterpreter
+from pennylane.capture.primitives import grad_prim
```
paul0403: Let's call it `pl_grad_p` like the other PL-sourced things in here : )

frontend/test/pytest/test_gradient.py
```diff
@@ -1026,11 +1091,12 @@ def f(x):
 
     with pytest.raises(DifferentiableCompileError, match="higher order derivatives"):
 
+        # not sure how to get this working with qml.grad TODO
         @qjit
         def workflow(x: float):
             g = qml.qnode(qml.device(backend, wires=1), diff_method=method)(f)
-            h = grad(g, method="auto")
-            i = grad(h, method="auto")
+            h = catalyst.grad(g, method="auto")
```
paul0403, line 1098: Why are some tests updated to `qml.grad/jacobian` but some others stayed as `catalyst.grad/jacobian`? Is there any other work needed to support grad/jacobian? Or are the `catalyst.grad/jacobian` tests run without capture? If that is the case, what's stopping us from running them under capture as well, like the other tests you updated? :thinking:

.github/workflows/check-catalyst.yaml
albi3ro: That's the plan. But just want to make sure we don't need any more changes there before it goes itn.

frontend/catalyst/from_plxpr/from_plxpr.py
albi3ro: Pytree inputs.  I might actually turn this change into an independent PR.

frontend/catalyst/jax_primitives.py
albi3ro: I am thinking we should probably error out with `catalyst.grad` and `catalyst.jacobian` if program capture is enabled, similar to what we do for `for_loop`, `while_loop` and `cond`.

frontend/test/pytest/test_gradient.py
albi3ro: Some are because they are `catalyst.grad` specific behavior. For this one, I am at a loss.

frontend/catalyst/jax_primitives.py
paul0403: Yes we can ban the API call, but I was thinking more about internally catalyst jaxpr will now have both PL primitives and catalyst primtives.

paul0403: For things like `qml.cond`, the binded primitive to catalyst jaxpr is still the catalyst cond primitive.

albi3ro: For now. I had a story I never got around to that would investigate unifying them. 

What exactly makes something a PL primitive or catalyst primitive? Where it's defined? Whether or not there's a lowering?

paul0403: Discussed, using the plxpr primitive directly is ok :+1:

frontend/catalyst/from_plxpr/from_plxpr.py
albi3ro: SPlit out into #2165

frontend/catalyst/from_plxpr/from_plxpr.py
albi3ro: renamed.

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -157,6 +157,15 @@ def __init__(self):
         super().__init__()
 
 
+@WorkflowInterpreter.register_primitive(pl_jac_prim)
+def handle_grad(self, *args, jaxpr, n_consts, **kwargs):
+    """Translate a grad equation."""
+    f = partial(copy(self).eval, jaxpr, args[:n_consts])
+    new_jaxpr = jax.make_jaxpr(f)(*args[n_consts:]).jaxpr
```
dime10: Do we have to retrace the function if we're essentially just cloning the primitive?

frontend/catalyst/jax_primitives_utils.py
```diff
@@ -86,33 +118,33 @@ def lower_callable(ctx, callable_, call_jaxpr, pipeline=None, context=None, publ
...
     """
-    if context is None:
-        context = tuple()
-    key = (callable_, *context, *pipeline)
-    if func_op := get_cached(ctx, key):
-        return func_op
+    if metadata is None:
+        metadata = tuple()
+    key = (callable_, *metadata, *pipeline)
+    if callable_ is not None:
```
dime10, line 146: Do we expect callable to be None sometimes?

frontend/catalyst/jax_primitives_utils.py
```diff
@@ -86,33 +118,33 @@ def lower_callable(ctx, callable_, call_jaxpr, pipeline=None, context=None, publ
     if pipeline is None:
         pipeline = tuple()
```
dime10, line 120: Is this needed after you changed the default to `()`?

frontend/catalyst/jax_primitives_utils.py
```diff
@@ -45,28 +70,35 @@ def get_call_equation(jaxpr):
...
 
-def lower_jaxpr(ctx, jaxpr, context=None):
+def lower_jaxpr(ctx, jaxpr, metadata=None, fn=None):
     """Lowers a call primitive jaxpr, may be either func_p or quantum_kernel_p
 
     Args:
         ctx: LoweringRuleContext
         jaxpr: JAXPR to be lowered
-        context: additional context to distinguish different FuncOps
+        metadata: additional metadata to distinguish different FuncOps
```
dime10, line 79: `fn` is missing

frontend/catalyst/jax_primitives_utils.py
```diff
@@ -45,28 +70,35 @@ def get_call_equation(jaxpr):
...
-    return lower_callable(ctx, callable_, call_jaxpr, pipeline=pipeline, context=context)
+
+    if fn is None or isinstance(fn, qml.QNode):
+        equation = get_call_equation(jaxpr)
+        call_jaxpr = equation.params["call_jaxpr"]
+        pipeline = equation.params.get("pipeline")
+        callable_ = equation.params.get("fn")
+        if callable_ is None:
+            callable_ = equation.params.get("qnode", None)
+    else:
```
dime10, line 93: What's the purpose of the `fn` parameter and this branch?

frontend/catalyst/from_plxpr/from_plxpr.py
albi3ro: We need to convert the target from plxpr to catalyst-jaxpr.

dime10: Good point!

frontend/catalyst/jax_primitives_utils.py
albi3ro: Might be getting passed from somewhere else, so thought to just play it safe.

frontend/catalyst/jax_primitives_utils.py
albi3ro: So previously, the `fn` was always embedded into the target jaxpr itself. For the catalyst grad, the jaxpr would contain one equation that was a `func` primitive. Then that `func` primitive would have both jaxpr and a `fn`.  Instead of having that additional nesting, I'm just storing the `fn` itself on the equation.

So that is something I could actually use some feedback on. Is there actually any difference between the equation storing both the jaxpr and the function, versus the equation storing a jaxpr that contains both the jaxpr and the function? To me, the second option just seems overly complicated.

For the catalyst grad, we had:
```
{'jaxpr': { lambda ; a:f64[]. let
    b:f64[] = func[
      call_jaxpr={ lambda ; c:f64[]. let
          d:f64[] = quantum_kernel[
            call_jaxpr={ lambda ; e:f64[]. let
                ...
            }
            pipeline=()
            qnode=...
          ] c
          l:f64[] = integer_pow[y=2] d
        in (l,) }
      fn=<function f at 0x12b4afa60>
    ] a
  in (b,) },
 'fn': <catalyst.jax_tracer.Function at 0x12b205460>,
 'grad_params': GradParams(method='auto', scalar_out=True, h=None, argnums=[0], scalar_argnums=True, expanded_argnums=[0], with_value=False)}
```

Even though it was storing `<catalyst.jax_tracer.Function at 0x12b205460>` on the primitive, we were using the `<function f at 0x12b4afa60>` from the `func` equation, not the `catalyst.jax_tracer.Function`.

I'm just adding logic to use the same level `fn`.

frontend/catalyst/jax_primitives_utils.py
albi3ro: I can't remember anymore. I still don't entirely understand this file and the assumptions it relies on. I might have just been trying to play it safe because I'm not entirely sure how things get used.

If it is indeed `None`, we could retrieve the wrong thing from the cache, so better safe than sorry.

dime10: Yeah no problem 👌

frontend/catalyst/jax_primitives_utils.py
dime10: Hmm, good question. It's hard to say I think that code has gone through a few iterations (like the cache on Function that is now gone). If you found a way to simplify it then I think that's great!

frontend/catalyst/jax_primitives_utils.py
albi3ro:
```suggestion
        metadata: additional metadata to distinguish different FuncOps
        fn (Callable | None): the function the jaxpr corresponds to. Used for naming and caching.
```

Issue #1911, opened 2025-07-16
`qml.prod` is not supported in AutoGraph passthrough


dime10:
The following code fails, even with #1910. The reason is that prod is not explicitly listed in the [transform call hotfix](https://github.com/PennyLaneAI/catalyst/blob/c54b13229525eae6336012ac96598623c1337451/frontend/catalyst/autograph/ag_primitives.py#L535-L536) nor the [transform object hotfix](https://github.com/PennyLaneAI/catalyst/blob/c54b13229525eae6336012ac96598623c1337451/frontend/catalyst/autograph/ag_primitives.py#L572-L577).

```py
@qml.prod
def template(b):
    if b:
        qml.H(0)
        qml.X(0)

@qjit(autograph=True)
@qml.qnode(qml.device("null.qubit", wires=0))
def circuit(b: bool):
    qml.adjoint(template(b))
    return qml.state()
```
```
  File "test.py", line 12, in template
    if b:

jax.errors.TracerBoolConversionError: Attempted boolean conversion of traced array with shape bool[].
This concrete value was not available in Python because it depends on the value of the argument b.
```

dime10:
Tagging @PennyLaneAI/core for visibility regarding autograph work in PennyLane / for the unification.

PR #2083, opened 2025-10-02
Fix qml.prod with autograph

rniczh:
**Context:**

Fix qml.prod with autograph

```python
@qml.prod
def template(b):
    if b:
        qml.H(0)
        qml.X(0)

@qjit(autograph=True)
@qml.qnode(qml.device("null.qubit", wires=1))
def circuit(b: bool):
    template(b)
    return qml.state()

print(circuit(True))
```

Error message:
```
jax.errors.TracerBoolConversionError: Attempted boolean conversion of traced array with shape bool[].
The error occurred while tracing the function circuit at ... for qjit_capture. This concrete value was not available in Python because it depends on the value of the argument b.
See https://docs.jax.dev/en/latest/errors.html#jax.errors.TracerBoolConversionError
```

HOTFIX: Handle calls to functions that were decorated with qml.prod, qml.adjoint, etc. These decorators return wrapper functions that call the original function without  autograph conversion. We detect these wrappers and unwrap them to convert the original function with autograph.

TODO: Once PL has has dedicated way to propagate autograph through decorators, we should remove this way of handling `qml.prod`.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-95653]
https://github.com/PennyLaneAI/catalyst/issues/1911

### DIFF
--- a/.github/workflows/check-catalyst.yaml
+++ b/.github/workflows/check-catalyst.yaml
@@ -478,12 +478,12 @@ jobs:
     - name: Install PennyLane release candidate
       if: github.event.pull_request.base.ref == 'v0.13.0-rc'
       run: |
-        pip install --no-deps --force git+https://github.com/PennyLaneAI/pennylane@v0.43.0-rc0
+        pip install --no-deps --pre -U --index-url https://test.pypi.org/simple/ pennylane'<=0.43.0'
 
     - name: Install PennyLane-Lightning release candidate
       if: github.event.pull_request.base.ref == 'v0.13.0-rc'
       run: |
-        pip install --no-deps --pre -U --extra-index-url https://test.pypi.org/simple/ pennylane-lightning'<=0.43.0'
+        pip install --no-deps --pre -U --index-url https://test.pypi.org/simple/ pennylane-lightning'<=0.43.0'
 
     - name: Get Cached LLVM Build
       id: cache-llvm-build
@@ -571,12 +571,12 @@ jobs:
     - name: Install PennyLane release candidate
       if: github.event.pull_request.base.ref == 'v0.13.0-rc'
       run: |
-        pip install --no-deps --force git+https://github.com/PennyLaneAI/pennylane@v0.43.0-rc0
+        pip install --no-deps --pre -U --index-url https://test.pypi.org/simple/ pennylane'<=0.43.0'
 
     - name: Install PennyLane-Lightning-Kokkos release candidate
       if: github.event.pull_request.base.ref == 'v0.13.0-rc'
       run: |
-        pip install --no-deps --pre -U --extra-index-url https://test.pypi.org/simple/ pennylane-lightning-kokkos'<=0.43.0'
+        pip install --no-deps --pre -U --index-url https://test.pypi.org/simple/ pennylane-lightning-kokkos'<=0.43.0'
 
     - name: Get Cached LLVM Build
       id: cache-llvm-build
@@ -643,7 +643,7 @@ jobs:
     - name: Install PennyLane release candidate
       if: github.event.pull_request.base.ref == 'v0.13.0-rc'
       run: |
-        pip install --no-deps --force git+https://github.com/PennyLaneAI/pennylane@v0.43.0-rc0
+        pip install --no-deps --pre -U --index-url https://test.pypi.org/simple/ pennylane'<=0.43.0'
 
     - name: Get Cached LLVM Build
       id: cache-llvm-build
--- a/frontend/catalyst/_version.py
+++ b/frontend/catalyst/_version.py
@@ -16,4 +16,4 @@
 Version number (major.minor.patch[-label])
 """
 
-__version__ = "0.13.0-rc1"
+__version__ = "0.13.0-rc2"


frontend/catalyst/autograph/ag_primitives.py
```diff
@@ -532,10 +532,11 @@ def converted_call(fn, args, kwargs, caller_fn_scope=None, options=None):
         (ag_config, "CONVERSION_RULES", module_allowlist),
         (ag_py_builtins, "BUILTIN_FUNCTIONS_MAP", py_builtins_map),
     ):
-        # HOTFIX: pass through calls of known Catalyst wrapper functions
-        if fn in (
+        # List of known wrapper functions that should be handled specially
+        _known_wrapper_functions = (
             catalyst.adjoint,
             qml.adjoint,
+            qml.prod,
```
dime10, line 540: Just wondering, are any of the changes besides this line relevant, or just refactoring?

rniczh: Yes, it's because I use it to find the decorator and apply it again to the converted call of the `fn.__wrapped`.
Actually, there is another way to achieve the same behaviour is through `fn.__globals__` (as shown below). It doesn't need to create the `_known_wrapper_functions`, and also doesn't need to find which decorator need to be applied as well. With that being said, it didn't seem robust enough for me, so I didn't use this solution 🤔. But if you find this solution is better than the previous one, I'm happy to update it 👍
```
decorator_name = fn.__module__.split(".")[-1]
decorator = fn.__globals__.get(decorator_name)
```

https://github.com/PennyLaneAI/catalyst/blob/67505c528e3976d22b6f7caa99ba2cfc3a71b2c6/frontend/catalyst/autograph/ag_primitives.py#L611-L613

dime10: > Yes, it's because I use it to find the decorator and apply it again to the converted call of the fn.__wrapped

Right, I am wondering about the robustness, but maybe going via __wrapped__ is actually the best way here? Not sure.

Normally, we grab the decorated function here since it's the first argument to the decorator.
The exception to this is if the decorator was already applied, outside the qjit scope, in which case it's too late to hit this branch. For that we use the `CatalystCallable`, or handle the `QNode` specially. I guess the difference with `qml.prod` and the other qml decorators here is that ctrl and adjoint are dispatched to their catalyst equivalent, so can be handled via CatalystCallable, whereas `qml.prod` is not dispatched 🤔

rniczh: >  I guess the difference with qml.prod and the other qml decorators here is that ctrl and adjoint are dispatched to their catalyst equivalent, so can be handled via CatalystCallable, whereas qml.prod is not dispatched 

Yes, exactly, and `prod` aslo doesn't has the catalyst equivalent impl as well.

frontend/catalyst/autograph/ag_primitives.py
```diff
@@ -598,6 +602,26 @@ def qnode_call_wrapper():
...
             return new_qnode(**new_kwargs)
 
+        # HOTFIX: Handle calls to functions that were decorated with qml.prod, qml.adjoint, etc.
+        # These decorators return wrapper functions that call the original function without
+        # autograph conversion. We detect these wrappers and unwrap them to convert the
+        # original function with autograph.
+        if hasattr(fn, "__wrapped__"):
+            # Find the decorator function
+            if decorator := next(
+                (f for f in _known_wrapper_functions if f.__module__ == fn.__module__), None
```
dime10: What is the intention here? That the wrapped function is from the same module as the wrapper? 🤔

rniczh: Yes, because we can't apply the fn on converted_inner: `fn(converted_inner)`. So find the decorator and apply on the `converted_inner`

https://github.com/PennyLaneAI/catalyst/blob/1e8e1e78c6d0c91c35e12d5a0bf3bbf3f30ea65c/frontend/catalyst/autograph/ag_primitives.py#L623

rniczh: The fn here is `template` and the module is: str`"pennylane.ops.op_math.prod"`
```python
@qml.prod
def template(b):
    if b:
        qml.H(0)
        qml.X(0)

@qjit(autograph=True)
@qml.qnode(qml.device("null.qubit", wires=1))
def circuit(b: bool):
    template(b)
    return qml.state()

print(circuit(True))
```

rniczh: So taking an example:

```python
@qml.adjoint
@qml.prod
def template(b: bool):
    qml.H(0)
    qml.X(0)
```

The function will be applied twice with two different `__module__`, and they are both `template` as fn
```
pennylane.ops.op_math.adjoint
pennylane.ops.op_math.prod
```

dime10: Ok I think I understand, you're assuming that qml decorator functions return new functions defined in the same module, and you are only matching against the list on known decorators above. Realistically, that will only be `qml.prod` from that list at the moment, since catalyst functions have their system and adjoint/ctrl get dispatched to catalyst equivalents.

While I do think the module approach is a bit brittle, since it is fairly restrictive I think it is sufficient for a hotfix. Just to confirm, I went to the PL source code to check the module of the api function and returned function [are the same](https://github.com/PennyLaneAI/pennylane/blob/3ead0f4f9bdbb2a459c591c812c5414c1d3a4013/pennylane/ops/op_math/prod.py#L115-L116) (I think the only reason that's the case is because of the special `wraps` method that excludes the `__module__` as a attribute to update).

Let's add a note like `# TODO: remove once PL has dedicated way to propagate autograph through decorators`.

frontend/catalyst/autograph/ag_primitives.py
```diff
@@ -598,6 +602,26 @@ def qnode_call_wrapper():
...
+                original_fn = fn.__wrapped__
+
+                # Convert the original function with autograph
+                def converted_inner(*inner_args, **inner_kwargs):
+                    return converted_call(
+                        original_fn, inner_args, inner_kwargs, caller_fn_scope, options
+                    )
+
+                # Apply the decorator to the converted function and call it
+                return decorator(converted_inner)(*args, **(kwargs if kwargs is not None else {}))
```
dime10: This still makes me a little uneasy actually, for two reasons:
- finding the decorator to apply relies merely on the module attribute, if there were two decorators in the same module it would break
- how we propagate additional decorator arguments, for instance with prod we have:
```
def prod(*ops, id=None, lazy=True):
```
  The `id` and `lazy` arguments are not available here since the original API call would have happened in a different scope, and here he call the "top-level" api function again. Admittedly, we don't probably don't use those specific arguments much in Catalyst, but still 🤔

rniczh: > finding the decorator to apply relies merely on the module attribute, if there were two decorators in the same module it would break

Yes, that did causing the problem 🤔 The module matching only works correctly when the `fn` of `@wraps(fn)` is itself. So it can't allow to have multiple different `fn` of `@wraps(fn)` within a function of course

> how we propagate additional decorator arguments, for instance with prod we have

I hadn't thought about the keyword argument part. I'm trying to capture the closure vars in the latest change https://github.com/PennyLaneAI/catalyst/pull/2083/commits/09bd293f20cdac9fdd9ce90b9138ee93f40f371f

The idea is:
Take closure vars: `ClosureVars(nonlocals={'fn': <function template at 0x....>, 'id': None, 'lazy': False}, ...)`, extract the nonlocals part, remove the function component `fn`, and use the rest as kwargs to re-apply it.

And I changed the implementation of prod a little bit from my side, mainly just to test that this usage is possible.
```python
if len(ops) == 0:
        def decorator(fn):
            return prod(fn, id=id, lazy=lazy)
        return decorator
```
This way, when using as follows, it can extract the nonlocals from updated closure vars: `ClosureVars(nonlocals={'fn': <function template at 0x...>, 'id': None, 'lazy': False}.` And apply with `decorator(converted_inner, **decorator_kwargs)` 
```python
@qml.prod(lazy=False)
...
```

frontend/catalyst/autograph/ag_primitives.py
rniczh: I added the TODO in https://github.com/PennyLaneAI/catalyst/pull/2083/commits/26d3b2a0c2c515150a73d2221963e5c18541eaf3

frontend/catalyst/autograph/ag_primitives.py
rniczh: @dime10 Now the restriction that only allows to handle the `qml.prod` this way is added @https://github.com/PennyLaneAI/catalyst/pull/2083/commits/3cfd412d9aa5f7ab97294e30b657b8bfcd1a9990
So now we don't search the decorator, just use `qml.prod` directly

frontend/catalyst/autograph/ag_primitives.py
```diff
@@ -598,6 +603,35 @@ def qnode_call_wrapper():
...
+
+            # Extract decorator arguments from the closure
+            # There is an assumption that the closure variables are
+            # the same as the decorator arguments
+            closure_vars = inspect.getclosurevars(fn)
+            decorator_kwargs = {
+                k: v
+                for k, v in closure_vars.nonlocals.items()
+                if v is not original_fn  # Exclude the wrapped function itself
+            }
```
dime10, line 621:
```suggestion
            decorator_kwargs = {"id": closure_vars.nonlocals["id"], "lazy": closure_vars.nonlocals["lazy"]}
```

PR #2085, opened 2025-10-03
Rename `t_layer_reduction` to `reduce_t_depth` pass

sengthai:
Rename the`t_layer_reduction` transform  to `reduce_t_depth` for more user-friendly

[[sc-99454]]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -92,19 +92,20 @@
   [A Game of Surface Codes](https://arXiv:1808.02892v3).
   [(#1975)](https://github.com/PennyLaneAI/catalyst/pull/1975)
   [(#2048)](https://github.com/PennyLaneAI/catalyst/pull/2048)
+  [(#2084)](https://github.com/PennyLaneAI/catalyst/pull/2084)
 
   Consider the following circuit.
 
   ```python
   import pennylane as qml
   from catalyst import qjit, measure
-  from catalyst.passes import to_ppr, commute_ppr, t_layer_reduction, merge_ppr_ppm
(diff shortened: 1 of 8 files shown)


dime10:
> I think the `reduce` here still weird, the behaviour of this pass doesn't involve any reduction. Even the `canonicalize_t` is much more better then `reduce_t`. It helps other pass to reduce the t depth, but not in this pass, right?
> 
> Edited:
> 
> > resulting in a circuit with a smaller PPR depth.
> 
> These description of this pass still not sounds right? From the user, they probably cannot get any reason of why this pass reduces the PPR depth? Because it's only happen when the optimization pass run after this pass.

Sengthai might correct me, but I believe this _is_ the optimization pass, and it does reduce the T depth. However, because the layer structure is not explicitly represented yet, it is not obvious that that depth was reduced. If you apply the layer generation pass before and after this optimization it becomes obvious that the depth is lower.
Imo it would be better to run the optimization once the layers are generated, but Sengthai mentioned it was easier this way. Maybe a signal that we need to work on the representation a bit to make it more amenable.

sengthai:
> I believe this is the optimization pass, and it does reduce the T depth. However, because the layer structure is not explicitly represented yet, it is not obvious that that depth was reduced. If you apply the layer generation pass before and after this optimization it becomes obvious that the depth is lower.

@dime10 💯 exactly is! We may need 

> Sounds like we can run the layer generation just after this pass? If so, would it fine to run the layer generation within this pass?

@rniczh Forming the layer after this pass is useful to show the depth reduction, however, it may distract other passes to perform on that layer structure. 

I think we can state from that pass's docstring about using `ppm_spec` to know the impact of that pass.

PR #2089, opened 2025-10-03
Various small OQC fixes

paul0403:
**Context:**
Various small OQC fixes.

- The error message when account info is missing now includes the env variables to set
- OQC cpp device itself should not be the one starting the python interpreter (this will cause python's "cannot start multiple interpreters" error from user python script). Instead, whatever cpp tests that need this should start their python interpreter with pybind.
- update to new OQC API 
- Properly link up the backend to use, instead of always falling back to lucy
- Migrate to new shots scheme (shots goes with qnode, not the device)

**Benefits:**
OQC device works

**Related GitHub Issues:** #2021
[sc-98652]

### DIFF
--- a/frontend/catalyst/qfunc.py
+++ b/frontend/catalyst/qfunc.py
@@ -456,7 +456,9 @@ def _process_terminal_measurements(mcm_method, cpy_tape, out, snapshots, shot_ve
     return (snapshots, tuple(new_out)) if snapshots else tuple(new_out)
 
 
-def _validate_one_shot_measurements(mcm_config, tape: qml.tape.QuantumTape, qnode) -> None:
+def _validate_one_shot_measurements(
+    mcm_config, tape: qml.tape.QuantumTape, user_specified_mcm_method, shot_vector, wires
+) -> None:
     """Validate measurements for one-shot mode.
 
     Args:
@@ -469,13 +471,11 @@ def _validate_one_shot_measurements(mcm_config, tape: qml.tape.QuantumTape, qnod
         NotImplementedError: If measurement configuration is not supported
     """
     mcm_method = mcm_config.mcm_method
-    user_specified_mcm_method = qnode.execute_kwargs["mcm_method"]
     assert mcm_method == "one-shot"
 
     # Check if using shot vector with non-SampleMP measurements
-    shot_vector = qnode._shots.shot_vector if qnode._shots else []
     has_shot_vector = len(shot_vector) > 1 or any(copies > 1 for _, copies in shot_vector)
-    has_wires = qnode.device.wires is not None and not is_dynamic_wires(qnode.device.wires)
+    has_wires = wires is not None and not is_dynamic_wires(wires)
 
     # Raise an error if there are no mid-circuit measurements, it will fallback to
     # single-branch-statistics
@@ -558,7 +558,6 @@ def dynamic_one_shot(qnode, **kwargs):
     """
 
     cpy_tape = None
-    aux_tapes = None
     mcm_config = kwargs.pop("mcm_config", None)
 
     def transform_to_single_shot(qnode):
@@ -567,15 +566,20 @@ def dynamic_one_shot(qnode, **kwargs):
                 "dynamic_one_shot is only supported with finite shots."
             )
 
+        user_specified_mcm_method = qnode.execute_kwargs["mcm_method"]
+        shot_vector = qnode._shots.shot_vector if qnode._shots else []
+        wires = qnode.device.wires
+
         @qml.transform
         def dynamic_one_shot_partial(
             tape: qml.tape.QuantumTape,
         ) -> tuple[Sequence[qml.tape.QuantumTape], Callable]:
             nonlocal cpy_tape
             cpy_tape = tape
-            nonlocal aux_tapes
 
-            _validate_one_shot_measurements(mcm_config, tape, qnode)
+            _validate_one_shot_measurements(
+                mcm_config, tape, user_specified_mcm_method, shot_vector, wires
+            )
 
             if tape.batch_size is not None:
                 raise ValueError("mcm_method='one-shot' is not compatible with broadcasting")
--- a/frontend/catalyst/jax_tracer.py
+++ b/frontend/catalyst/jax_tracer.py
@@ -206,15 +206,6 @@ class Function:
         AssertionError: Invalid function type.
     """
 
-    CACHE = {}
-
-    def __new__(cls, fn):
-        if cached_instance := cls.CACHE.get(fn):
-            return cached_instance
-        new_instance = super().__new__(cls)
-        cls.CACHE[fn] = new_instance
-        return new_instance
-
     @debug_logger_init
     def __init__(self, fn):
         self.fn = fn
--- a/mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
+++ b/mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
@@ -87,7 +87,11 @@ class OpSignatureAnalyzer {
                                 .inCtrlQubits = op.getCtrlQubitOperands(),
                                 .inCtrlValues = op.getCtrlValueOperands(),
                                 .outQubits = op.getNonCtrlQubitResults(),
-                                .outCtrlQubits = op.getCtrlQubitResults()})
+                                .outCtrlQubits = op.getCtrlQubitResults(),
+                                .inWireIndices = {},
+                                .inCtrlWireIndices = {},
+                                .outQubitIndices = {},
+                                .outCtrlQubitIndices = {}})
     {
         if (!enableQregMode)
             return;


frontend/catalyst/third_party/oqc/src/oqc_python_module.cpp
```diff
@@ -33,17 +32,22 @@ std::string program = R"(
...
+    optimisations = qcaas_client._externalized_compiler_config.Tket()
+    optimisations.tket_optimizations = qcaas_client._externalized_compiler_config.TketOptimizations.DefaultMappingPass
     RES_FORMAT = QuantumResultsFormat().binary_count()
     client = OQCClient(url=url, email=email, password=password)
     client.authenticate()
     oqc_config = CompilerConfig(repeats=shots, results_format=RES_FORMAT, optimizations=optimisations)
-    oqc_task = QPUTask(circuit, oqc_config)
-    res = client.execute_tasks(oqc_task)
+    oqc_task = QPUTask(program=circuit, config=oqc_config, qpu_id="qpu:uk:2:d865b5a184")
+    res = client.execute_tasks(oqc_task, qpu_id="qpu:uk:2:d865b5a184")
```
dime10: Let's figure out how to not hard code this and we are good to go I would say. Thanks for all the fixes!

paul0403: https://github.com/PennyLaneAI/catalyst/pull/2089/commits/3514fe5686e5a0b32c414fe31f8ba68e9cf20b46

frontend/catalyst/qfunc.py
```diff
@@ -101,6 +102,15 @@ def _get_total_shots(qnode):
...
     if shots_value is None:
+        if qnode.device.name == "OQCDevice":
+            raise CompileError(
+                textwrap.dedent(
+                    """
+                OQC device does not support analytical simulation.
+                Please supply the number of shots on the qnode.
+                """
+                )
+            )
```
dime10: This is not great, is there not a better spot to put this device specific logic than in a generic function?

paul0403: I agree, I'm trying to find somewhere better (and struggling actually). 

Because the OQC device file does not actually see the qnode (which is where shots is now :melting_face: )

paul0403: But let me try finding somewhere better.

dime10: One option could be at runtime I suppose (within the device), but it would be better earlier 🤔

paul0403: No, the other random error we saw (something about can't run sample/counts) is still in the frontend, so this new message needs to be efore that...

paul0403: ~~I guess the `extract_backend_info` on `QJITDevice` seems a bit better... the "preprocessing" for things like braket device also happens there.~~ Never mind `QJITDevice` does not see the qnode

frontend/catalyst/third_party/oqc/src/OQCDevice.hpp
```diff
@@ -43,6 +43,9 @@ class OQCDevice final : public Catalyst::Runtime::QuantumDevice {
     std::set<QubitIdType> initial_allocated_QubitIds;
     std::unordered_map<std::string, std::string> device_kwargs;
 
+    std::string qpu_id;
+    static const std::unordered_map<std::string, std::string> qpu_map;
```
rniczh: I'm wondering could `qpu_map` be defined by `inline static const`? Then L96 to L98 can be removed.
```cpp
inline static const std::unordered_map<std::string, std::string> qpu_map = {
        {"lucy", "qpu:uk:2:d865b5a184"}, 
        {"toshiko", "qpu:jp:3:673b1ad43c"}
    };
```

frontend/catalyst/third_party/oqc/src/oqc_python_module.cpp
```diff
@@ -33,38 +32,41 @@ std::string program = R"(
...
+        raise ValueError(
+            """
+            OQC credentials not found in environment variables.
+            Please set the environment variables `OQC_EMAIL`, `OQC_PASSWORD` and `OQC_URL`.
+            """
+        )
+    import qcaas_client
+    from qcaas_client.client import OQCClient, QPUTask, CompilerConfig, QuantumResultsFormat
+    optimisations = qcaas_client._externalized_compiler_config.Tket()
+    optimisations.tket_optimizations = qcaas_client._externalized_compiler_config.TketOptimizations.DefaultMappingPass
```
rniczh: Is there any reason we cannot use `compiler_config` directly?
From https://docs.oqc.app/code_examples.html :

> from qcaas_client.compiler_config import CompilerConfig, QuantumResultsFormat, Tket, TketOptimizations
> ...
> optimisations = Tket()
optimisations.tket_optimizations = TketOptimizations.Two

frontend/catalyst/qfunc.py
paul0403: I think this works https://github.com/PennyLaneAI/catalyst/pull/2089/commits/422174dcd7b97a3c129bbf32c8fceb6f11b81ec5

I will need to tweak some other things, but I think this idea works.

frontend/catalyst/device/qjit_device.py
```diff
@@ -363,7 +364,7 @@ def preprocess(
 
         if execution_config is None:
             execution_config = qml.devices.ExecutionConfig()
-        _, config = self.original_device.preprocess(execution_config)
+        _, config = self.original_device.preprocess(execution_config, shots)
```
paul0403: Actually, how about just 
```
   if self.original_device is OQCDevice and shots is None:
      ...error...
```

right here? This way we don't need to change this general signature, and qjit_device file does do device-specific preprocessing already (like braket)

frontend/catalyst/qfunc.py
dime10: good idea. the preprocess has a standard signature though, right?

dime10: Is this something we can do in the verification with respect to the toml capabilities? 🤔

frontend/catalyst/device/qjit_device.py
dime10: It does but it shouldn't 🤔

dime10: This issue might point to a problem in our device api. The device either needs to be able to verify this parameter itself or provide us the necessary information declaratively (i.e. toml) so that we can do it for any device.

paul0403: I have a new idea, one sec...

paul0403: @JerryChen97 did add the pathway for the qjit device to see the program-instance's shots from the qnode (the `shots` arg on preprocess). But around there is also where we get the capabilities from the tomls. Seems like we can check this around here. Let me try.

paul0403: https://github.com/PennyLaneAI/catalyst/pull/2089/commits/8a90d4acb1d29741824c5bfec91674458710d9f2 I tried various things, but eventually I couldn't get it any better than this. There are so many hacks around this set shots business.

For starters, there seems to be 3 levels of "capabilities": the completely static toml-level capabilities, the capabilities recomputed with the info of whether the user provided shots, and then finally the qjit device capabilities. 

And then, there seems to be 2 sources of truth for shots: on the `original_device`, there always will be a `shots` attribute that essentially `bool`s to `False`; and then the shots from user's qnode `set_shots` is to be taken in in the device's `preprocess` during `trace_quantum_function` (in `jax_tracer.py`).

This means whether the user has specified shots or not is not actually known at the initialization of a `QJITDevice`. I guess the point was so that the initilization always just needs to care about the static toml capabilities? But then, the initilization also has the following lines
```python
class QJITDevice:
   def __init__(...):
      device_capabilities = get_device_capabilities(original_device, self.original_device.shots)
      self.capabilities = get_qjit_device_capabilities(device_capabilities)

...

def get_device_capabilities(device, shots=None) -> DeviceCapabilities:
    ...
    shots_present = bool(shots)
    device_capabilities = _load_device_capabilities(device).filter(finite_shots=shots_present)
```

(Note that `_load_device_capabilities` is the true toml capabilities)

But, due to 

> on the `original_device`, there always will be a `shots` attribute that essentially `bool`s to `False`

the `get_device_capabilities` always filters as if shots is not present, regardless of whether the user has supplied shots through `set_shots`. This means after initialization, **MP with finite shots are always filtered out**. I do not know whether this is intended or a bug.

But then, this new message we're adding requires we know whether the user has provided shots (so at `preprocess`), and whether the toml capabilities says the device needs shots. But at `preprocess`, due to the above issue, it is as if all devices do not require shots! So at the point of the check (beginning of `preprocess`, where user's `set_shots` is first available during the device-ish files), I need to reload the static toml capabilities. 

-----------------------

I cannot think of any better ways to do this. I tried refactor this but it was just too much. The refactoring idea I had was that, loading of capabilities should only happen either just in init, or just in preprocess. Both had some challenges.
- If we load capabilities just in preprocess, a bunch of tests break because they maunally create a dummy device, edit the capabilities (which requires capabilities to be loaded at init!), then run preprocess to test the transform program.
- If we load capabilities just in init, this does seem nicer because now the qjit device instance's capabilities are completely static from the toml, but then again we run into the above issue.

paul0403: ~I also don't know why almost all kokkos tests are breaking in CI (I do not see this locally...)~ latest lightning kokkos rc was broken

dime10: > the get_device_capabilities always filters as if shots is not present, regardless of whether the user has supplied shots through set_shots. This means after initialization, MP with finite shots are always filtered out. I do not know whether this is intended or a bug.

That sounds like a bug introduced by the migration to set_shots 🤔

frontend/catalyst/third_party/oqc/src/oqc_python_module.cpp
paul0403: Good points! I added your suggestions in https://github.com/PennyLaneAI/catalyst/pull/2089/commits/d0124f0540146e3e5bb531375c15d31c1471664d

frontend/catalyst/device/qjit_device.py
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2089/commits/069a641edcf716ee6dc193216dbe45fb09821d8e I refactored around the capabilities a bit. This bug is no longer there as I no longer use `original_device.shots` (the thing that PL is now leaving as an empty shots object just as a deprecation cycle thing and always `bool`s to false).

Essentially, the idea is that during `QJITDevice`'s init, we only load the static toml device specs. Then in preprocess I update the capabilities with shot info, then compute the transform program with the updated capabilities. I needed to change a few tests for the device internals because I moved the shot processing from init to preprocess now, but I think it's all good.

paul0403: Wait I forgot to cleanup... this commit looks very ugly :sweat_smile: sorry

frontend/catalyst/device/qjit_device.py
```diff
@@ -366,11 +369,29 @@ def preprocess(
...
 
         program = TransformProgram()
-        if shots is None:
-            capabilities = self.capabilities
+
+        # During preprocessing, we now have info on whether the user is requesting execution
+        # with shots.
+        # Note that this new set of capabilities are only temporarily needed for the computation
+        # of the preprocessing transform program.
+        if not shots:
```
dime10: I think the previous condition (`is None`) makes more sense no? Technically 0 shots is finite shots (with empty results)

frontend/catalyst/device/qjit_device.py
```diff
@@ -366,11 +369,29 @@ def preprocess(
...
+                self.original_device, shots=False, capabilities=self.capabilities
+            )
+            capabilities = get_qjit_device_capabilities(device_caps)
         else:
-            # recompute device capabilities if shots were provided through set_shots
-            device_caps = get_device_capabilities(self.original_device, shots)
+            device_caps = get_device_capabilities(
+                self.original_device, shots=True, capabilities=self.capabilities
+            )
             capabilities = get_qjit_device_capabilities(device_caps)
```
dime10: One thing that's interesting now is that we reload the qjit capabilities no matter what 🤔 Is it even useful to do it in the device initially then?

frontend/catalyst/device/qjit_device.py
```diff
@@ -531,13 +550,22 @@ def _load_device_capabilities(device) -> DeviceCapabilities:
     return capabilities
 
 
-def get_device_capabilities(device, shots=None) -> DeviceCapabilities:
-    """Get or load the original DeviceCapabilities from device"""
+def get_device_capabilities(device, shots=None, capabilities=None) -> DeviceCapabilities:
```
dime10: I think if the sole purpose is to filter mp capabilities based on shots, we might as well rename this to something more descirptive, like `filter_shot_capabilities` or similar.

dime10: Actually I don't quite like that the function is dual purpose with both `device` and `capabilities`, the only two uses are in this file anyway and we always provide the capabilities. Maybe the function in it's original form is no longer needed?
Lots of test use it though, so it might require some updates...
Alternatively, the filtering could be split out into it's own function, and `get_device_capabilities` calls load + filtering. Some kind of cleanup can be done here I think :)

paul0403: Yes, the only need for it to take in old capabilities is because of the uses in this file...

> Lots of test use it though, so it might require some updates...

But this was why I couldn't change the name.

But indeed we can do the filtering in another function. Let me see what I can do.

frontend/catalyst/device/qjit_device.py
paul0403: This was changed because it was still needing to take in the PL empty shots object, but I now realize I'm not even doing that anymore. I'll change it back.

paul0403: Actually no, it's still needed. It's called in jax_tracer.py with either the empty shots object on the qnode or a concrete user shots. So this was a bug if comparing against None. 

In other words, the kwargs `shots=None` in this function never actually uses the default `None`. 

https://github.com/PennyLaneAI/catalyst/blob/main/frontend/catalyst/jax_tracer.py#L1450
https://github.com/PennyLaneAI/catalyst/blob/main/frontend/catalyst/jax_tracer.py#L1485

dime10: Can we check against the proper "null" value then, rather than a boolean? `shots=0` would also trigger this branch otherwise

frontend/catalyst/device/qjit_device.py
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2089/commits/fb10c8d1ea807331d367487d87843b53c39c1b0f

I made a new function, `filter_device_capabilities_with_shots`, to just do the filtering. The old `get_device_capabilities` is now unused in the code base. It was just used in preprocess for this filtering, but now I call the filter function instead.

Now, the issues are mainly with how the tests are written. My changes work with everything that isn't the manual tests using these internals.

There's two kinds of problems in the problematic tests with the `get_device_capabilities` function:
- They create a custom set of device capabilities by copying the "full capabilities" of lightning and manually delete some. They call `get_device_capabilities` on lightning device.
- Then they mock out `get_device_capabilities` by letting it return the new, reduced set of capabilities. This was needed because `preprocess` used to call this `get_device_capabilities`.

What did this demand of the code?

Let's look at the old design (before my latest "desparado" commit):
0. `get_device_capabilities` has a "read from raw device toml mode" and a "take in existing capability and update mode"
1. `QJITDevice`'s init reads raw toml, not through `get_device_capabilities`, but just through the raw loading function
2. preprocess filters raw toml by calling `get_device_capabilities` in "capability mode": it was doing the filtering.
3. Tests accomplish point 1 by calling `get_device_capabilities` in "read raw toml mode", since they need the raw lightning capabilities
4. Tests accomplish point 2 because `get_device_capabilities` was in preprocess, so now preprocess (which computes the transform program, which is the thing being tested) correctly runs with the mock capabilities. 

My new design:
0.  `get_device_capabilities` now always reads from raw toml directly. It no longer has the "update old capacity mode". This ensures point 1 of what the tests want (load raw lightning capabilities).
1. The new filter function is just the "take in existing capability and update mode".
2. preprocess calls the new filter function, since that's all it actually needs. 
3. Because `preprocess` now calls the filter function instead, I had to change the test mock to mock out the new filter function instead of the old `get_device_capabilities`, to accomplish point 2 of the tests.

This is the minimally intrusive design I can think of. The codebase itself is now very clean: init loads from raw toml, preprocess filters based on shots, and that's it. It's just that the unused `get_device_capabilities` needs to still be maintained because of the weirdness of the tests. 

What should the tests actually do? Well, maybe they should have an actual toml for the mock custom test devices, or 
maybe the `get_device_capabilities` can just be downstreamed as a testing util function, instead of living in the actual code base? But anyway, there is obviously no time to clean all the tests now. The offending tests are just the ones in `test_measurement_transforms.py` that check the capabilities with these internals directly. The "regular" tests are all fine. This test suite maintainance also sounds like a good first issue for one of the newer members of the team actually!

frontend/catalyst/device/qjit_device.py
paul0403: shots cannot be 0

https://github.com/PennyLaneAI/pennylane/blob/master/pennylane/measurements/shots.py#L150

But I agree checking against the empty shot object is better, since that's what we are actually calling it with. Let me see how to do this.

frontend/catalyst/device/qjit_device.py
dime10: Great, thank you Paul! Agreed the new design is pretty clean :)

frontend/catalyst/device/qjit_device.py
dime10: Do you know the answer to this one?

frontend/catalyst/device/qjit_device.py
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2089/commits/861404cb04a18fe69cfcbd08d3ca200902f223f1

Apprantly shots can be `None` (default kwarg value of preprocess function), the shot object (second check), or just an int (which will not be zero since PL already checks it, so I add the isinstance check).

I guess technically it can also be a tracer, but if the user has provided a tracer for shots then for the purposes of this check the shots are not absent. Whether user calls it with zero shots is now up to them I guess.

dime10: looks good 👍

frontend/catalyst/device/qjit_device.py
paul0403: The second `get_qjit_device_capabilities` (in `preprocess`) does feel redundant... but if I remove it two tests fail. Let me look into it.

paul0403: https://github.com/PennyLaneAI/catalyst/pull/2089/commits/7e4b9edd17fde6ba9312566c8779b4b0460cad04 never mind, it was another mock

dime10: Oh wait, don't all those transforms below the lines in your commit need the accurate capabilities including the qjit modifications?

dime10: Ah never mind, the input is already the qjit-adjusted capabilities, okay I think this looks good 👍

paul0403: Yes, the init already adds the qjit level capabilities to the raw toml capabilities.

PR #2091, opened 2025-10-03
[Decomposition] Update default fallback message following changes in PL

maliasadi:
**Context:**
Update the default fallback message and behaviour so that the system always falls back to the legacy implementation when PL’s decomposition graph fails to solve. This ensures consistency with PL `UserWarning`s and allows us to continue compiling and executing the circuit end-to-end.

- Capture warnings and raise a `UserWarning` in `from_plxpr` when graph fails.
- Consistency with the behaviour of PL when the graph-decomposition fails to find rules

**Benefits:**
- Clearer user warnings/errors (consistent with PL warnings/errors) using `qml.transforms.decompose`

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-100399]

### DIFF
--- a/frontend/catalyst/from_plxpr/from_plxpr.py
+++ b/frontend/catalyst/from_plxpr/from_plxpr.py
@@ -1025,7 +1025,7 @@ def _collect_and_compile_graph_solutions(inner_jaxpr, consts, tkwargs, ncargs):
         warnings.showwarning(w.message, w.category, w.filename, w.lineno)
         # TODO: use a custom warning class for this in PennyLane to remove this
         # string matching and make it more robust.
-        if "The graph-based decomposition system is unable" in str(w.message):
+        if "The graph-based decomposition system is unable" in str(w.message): # pragma: no cover
             graph_succeeded = False
             warnings.warn(
                 "Falling back to the legacy decomposition system.",
--- a/frontend/catalyst/from_plxpr/from_plxpr.py
+++ b/frontend/catalyst/from_plxpr/from_plxpr.py
@@ -17,6 +17,7 @@ This submodule defines a utility for converting plxpr into Catalyst jaxpr.
 # pylint: disable=protected-access
 # pylint: disable=too-many-lines
 
+import warnings
 from copy import copy
 from functools import partial
 from typing import Callable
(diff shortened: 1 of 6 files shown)


frontend/test/pytest/from_plxpr/test_from_plxpr.py
```diff
@@ -1000,6 +1000,27 @@ def circuit(x):
...
+        @partial(qml.transforms.decompose, gate_set={qml.RX, qml.RZ})
+        @qml.qnode(qml.device("lightning.qubit", wires=2))
+        def circuit(x):
+            qml.Hadamard(x)
+            return qml.state()
+
+        with pytest.warns(
+            UserWarning,
+            match="The graph-based decomposition system is unable to find a decomposition for",
+        ):
```
paul0403: I was confused for a bit since I though hadamard can decompose to these two, then I remembered globalphase :sweat: 

Can we make it clearer by changing this set to e.g. empty? So it's obvious that this decomposition solution would fail.

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -986,15 +1008,31 @@ def _collect_and_compile_graph_solutions(inner_jaxpr, consts, tkwargs, ncargs):
...
+
+    with warnings.catch_warnings(record=True) as captured_warnings:
+        warnings.simplefilter("always", UserWarning)
+        final_jaxpr = jax.make_jaxpr(gds_wrapper)(*ncargs)
+
+    for w in captured_warnings:
+        warnings.showwarning(w.message, w.category, w.filename, w.lineno)
+        # TODO: use a custom warning class for this in PennyLane to remove this
+        # string matching and make it more robust.
+        if "The graph-based decomposition system is unable" in str(w.message):
```
paul0403: What's the difference between the two checks, here and in decompose.py? :thinking:

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -219,14 +220,26 @@ def handle_qnode(
...
             inner_jaxpr=closed_jaxpr.jaxpr,
             consts=closed_jaxpr.consts,
             tkwargs=self.decompose_tkwargs,
             ncargs=non_const_args,
         )
 
+        # Fallback to the legacy decomposition if the graph-based decomposition failed
+        if not graph_succeeded:
+            # Remove the decompose-lowering pass from the pipeline
+            self._pass_pipeline = [p for p in self._pass_pipeline if p.name != "decompose-lowering"]
```
paul0403, line 235: Does the decompose-lowering pass not just do nothing when its targets (the decomp rule functions) are not present in the IR? Can we avoid manually editing the pipeline here?

maliasadi: Yup, but this simply avoids extra decompose-lowering pass checks at MLIR.

maliasadi: I also wanted to ensure that we don't hardcode the target gateset when we decompose in PLxPR

frontend/catalyst/from_plxpr/from_plxpr.py
maliasadi: The first check (decompose.py) avoids capturing and compiling partially solved decomposition rules (this is de facto something that will be fixed in PL in the graph module, as in the current implementation of the graph, the solver doesn't raise those partially solved operations and returns rules which will not be useful in the fallback). This wasn't a problem in PL but needs to be fixed for the Catalyst integration to remove this patch. The second check tells the compiler to trigger decomposition in PLxPR and avoids handling `decompse-lowering` to the program.

frontend/test/pytest/from_plxpr/test_from_plxpr.py
maliasadi: I updated this test to the following:

``` python
        @partial(qml.transforms.decompose, gate_set={qml.GlobalPhase})
        def circuit(x):
            qml.Hadamard(x)
            return qml.state()
```

the graph still fails while the legacy system will be able to decompose it! 
I don't check the lit test to check the correctness of the fallback decomposition.

Let me know if it clears it out

PR #2095, opened 2025-10-06
Modify RC CI to install RC branches from PL and Lightning

paul0403:
**Context:**
PRs targeting `v0.13.0-rc` should be tested with the release candidate branches of PL and Lightning as well.
We manually overwrite the CI to install these directly after `make frontend` (which gets the version from `.dep_versions`).

Note that when merging rc branch back to main after the release, this PR must be reverted for `.dep_versions` to take back control of the versions.

### DIFF
--- a/.github/workflows/check-catalyst.yaml
+++ b/.github/workflows/check-catalyst.yaml
@@ -478,13 +478,13 @@ jobs:
     - name: Install PennyLane release candidate
       if: github.event.pull_request.base.ref == 'v0.13.0-rc'
       run: |
-        pip install --no-deps --force git+https://github.com/PennyLaneAI/pennylane@0.43.0-rc0
+        pip install --no-deps --force git+https://github.com/PennyLaneAI/pennylane@v0.43.0-rc0
     - name: Checkout PennyLane-Lightning release branch
       if: github.event.pull_request.base.ref == 'v0.13.0-rc'
       uses: actions/checkout@v4
       with:
         repository: PennyLaneAI/pennylane-lightning
-        ref: 0.43.0_rc
+        ref: v0.43.0_rc
         path: lightning_build
         fetch-depth: 0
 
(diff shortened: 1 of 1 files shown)


.github/workflows/check-catalyst.yaml
```diff
@@ -620,6 +662,11 @@ jobs:
         python3 -m pip install -r requirements.txt
         make frontend
 
+    - name: Install PennyLane release candidate
+      if: github.event.pull_request.base.ref == 'v0.13.0-rc'
+      run: |
+        pip install --no-deps --force git+https://github.com/PennyLaneAI/pennylane@v0.43.0-rc0
+
```
mehrdad2m, line 647: What about installing `Lightning` and `Lightning-kokkos` here.

paul0403: This is for the openqasm tests. So no need.

Similarly I did not install kokkos for the lightning.qubit frontend tests.

mehrdad2m (approved):

.github/workflows/check-catalyst.yaml
```diff
@@ -478,22 +478,12 @@ jobs:
...
-        pip install --no-deps --force git+https://github.com/PennyLaneAI/pennylane@v0.43.0-rc0
-    - name: Checkout PennyLane-Lightning release branch
-      if: github.event.pull_request.base.ref == 'v0.13.0-rc'
-      uses: actions/checkout@v4
-      with:
-        repository: PennyLaneAI/pennylane-lightning
-        ref: v0.43.0_rc
-        path: lightning_build
-        fetch-depth: 0
+        pip install --no-deps --pre -U --extra-index-url https://test.pypi.org/simple/ pennylane
```
mehrdad2m: hmm, this could still be tricky. How can we be sure that latest testPyPi version that is installed is the rc wheel and not the latest dev wheel? I don't think the `--pre` would exclude the dev wheels does it? 🤔

paul0403: I just added a latest commit to pin lightning to check for versions with the rc name.

PL has not uploaded its first rc testpypi yet, so I'll update that tomorrow.

paul0403: Never mind this command doesn't work... pip parsed it into
```
pip install --no-deps --pre -U --extra-index-url https://test.pypi.org/simple/ pennylane-lightning
```
and then pipe (`>`) the output of that command to a file called `=0.43.0rc0`...

paul0403: https://github.com/PennyLaneAI/catalyst/pull/2095/commits/1396161d0a75289c8d235778deeec4528b852d32

paul0403: Tomorrow I need to update this for the PL rc testpypi package name as well.

mehrdad2m (approved):

PR #2100, opened 2025-10-07
Selective unitary decomposition fallback to fix IR blow-up

### LINKED ISSUE #2077: PC-relative offset overflow in PLT entry for `__catalyst__rt__device_init'
Using PL==0.43.0-dev65 and Catalyst==0.13.0-dev68.

I ran into this problem while catalys-ing a larger workflow, so here only represent a minimal example which demonstrates the bug; this does not arise for fewer qubits. When running the following circuit I get warning messages followed by an error:
```python
nQub=14

@qml.qjit(autograph=True, abstracted_axes=("n",)) 
@qml.qnode(qml.device("lightning.qubit", wires=nQub)) 
def circuit(): 
    wires=[i for i in range(nQub)] #wires 0 thru the number needed
    qml.ctrl(qml.RX,control=[1,2,3,nQub-1])(1,wires=[0])
    qml.ctrl(qml.Adder,control=[1,2,3])(1,x_wires=wires[4:])
    return qml.state() 

circuit()
```

I receive many warning messages such as:
`[/lib/x86_64-linux-gnu/crti.o](http://localhost:8888/lib/x86_64-linux-gnu/crti.o): in function _init':
(.init+0xb): relocation truncated to fit: R_X86_64_REX_GOTPCRELX against undefined symbol __gmon_start__'`

and at the end of these compilation messages an error message such as 
`[/tmp/circuitr4_k_uhd/circuit.so](http://localhost:8888/tmp/circuitr4_k_uhd/circuit.so): PC-relative offset overflow in PLT entry for __catalyst__rt__device_init'`

The repeated nature of the 'relocation truncated to fit' messages and the final one appear similar to [this](https://github.com/apache/mxnet/issues/15971) or [this](https://stackoverflow.com/questions/76456034/remove-small-code-style-relocations-from-pre-built-libraries).

This seems to block a successful compilation, even as removing the qjit decorator allows the circuit to run succesfully with the lightning simulator.

Additional note: for this size of a circuit, I get a warning soon after beginning to compile the circuit (but before getting the error message), saying something such as:
`[my_environment_location]/lib/python3.12/site-packages/jax/_src/interpreters/mlir.py:1135: UserWarning: A large amount of constants were captured during lowering (2.15 GB total). If this is intentional, disable this warning by setting JAX_CAPTURED_CONSTANTS_WARN_BYTES=-1.` 
I have seen this get up to 30GB with larger (depthwise) circuits.

This may warrant an additional issue: I also note that when trying to compile circuits of this size, my /tmp/ directory quickly fills up with remnants of the almost compiled circuits, ie mlir files of tens of GB. These can be easily purged from the terminal, but it is slightly concerning that I am storing mid-compile data from compilations that did not complete.

dime10:
Do not fallback to QubitUnitary unless operator provides no decomposition. Previously, a QubitUnitary would be constructed for any Controlled operator not natively supported by the device. While this can have some advantages, e.g. better performance on simulators for gates up to a certain size (which was never the intention of the fallback), it can also cause issues like overly large constant matrices embedded into the IR, and can also be unexpected to users since they might expect a decomposition to elementary gates.

The proposed update should be more robust in that it still uses the QubitUnitary fallback for gates that don't implement a decomposition, but otherwise makes use of the provided one.

closes #2077

### DIFF
--- a/frontend/catalyst/device/decomposition.py
+++ b/frontend/catalyst/device/decomposition.py
@@ -16,6 +16,7 @@
 This module contains the decomposition functions to pre-process tapes for
 compilation & execution on devices.
 """
+
 import copy
 import logging
 from functools import partial
@@ -79,11 +80,9 @@ def catalyst_decomposer(op, capabilities: DeviceCapabilities):
     if op.name in getattr(capabilities, "to_matrix_ops", {}):
         return _decompose_to_matrix(op)
 
-    if op.has_matrix and isinstance(op, qml.ops.Controlled):
-
+    if not op.has_decomposition and op.has_matrix and "QubitUnitary" in capabilities.operations:
         # If the device supports unitary matrices, apply the fallback.
-        if "QubitUnitary" in capabilities.operations:
-            return _decompose_to_matrix(op)
+        return _decompose_to_matrix(op)
 
     return op.decomposition()
 
--- a/doc/releases/changelog-0.13.0.md
+++ b/doc/releases/changelog-0.13.0.md
@@ -175,6 +175,14 @@
 
 <h3>Improvements 🛠</h3>
 
+* `QubitUnitary` is no longer favoured in the decomposition of controlled operators when the
+  operator is not natively supported by the device, and the device supports `QubitUnitary`. Instead,
+  conversion to `QubitUnitary` only happens if the operator does not define another decomposition.
+  The previous behaviour was the cause of performance issues when dealing with large controlled
+  operators, as their matrix representation could be embedded as dense constant data into the
+  program. The performance difference can span multiple orders of magnitude depending on regime.
+  [(#2100)](https://github.com/PennyLaneAI/catalyst/pull/2100)
+
 * Catalyst conditional operators, such as :func:`~.cond` or :func:`pennylane.cond`, now allow the
   target / branch functions to use arguments in their call signature. Previously, one had to supply
   all values via closure, but this is now done automatically under the hood.
--- a/frontend/test/pytest/test_preprocess.py
+++ b/frontend/test/pytest/test_preprocess.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Test for the device preprocessing."""
+
 import platform
 from dataclasses import replace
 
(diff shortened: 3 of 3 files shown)


frontend/catalyst/device/decomposition.py
```diff
@@ -79,11 +80,9 @@ def catalyst_decomposer(op, capabilities: DeviceCapabilities):
     if op.name in getattr(capabilities, "to_matrix_ops", {}):
         return _decompose_to_matrix(op)
 
-    if op.has_matrix and isinstance(op, qml.ops.Controlled):
-
+    if not op.has_decomposition and op.has_matrix and "QubitUnitary" in capabilities.operations:
         # If the device supports unitary matrices, apply the fallback.
```
mehrdad2m:
```suggestion
        # If the operator has no decomposition and the device supports unitary matrices, 
        # apply the fallback.
```

frontend/catalyst/device/decomposition.py
```diff
@@ -79,11 +80,9 @@ def catalyst_decomposer(op, capabilities: DeviceCapabilities):
     if op.name in getattr(capabilities, "to_matrix_ops", {}):
         return _decompose_to_matrix(op)
 
-    if op.has_matrix and isinstance(op, qml.ops.Controlled):
-
+    if not op.has_decomposition and op.has_matrix and "QubitUnitary" in capabilities.operations:
```
mehrdad2m: The new condition seems to accept any operation regardless of whether it is controlled or not. is that intentional? If so, does it make sense to add a test for a non-controlled circuit that would be triggered by this?

dime10: Yes I suppose this change was intentional, because if there is no decomposition available, and at this point we've already requested the op to be decomposed, the alternative would be to fail. 

An example of an op that only defines a matrix and not a decomp is BlockEncode. I think we already have tests for this, but it turns out there is [another mechanism](https://github.com/PennyLaneAI/catalyst/blob/a0f600172fb9109ab83f7db2a7d5bcd8d577de73/frontend/catalyst/device/qjit_device.py#L542-L548) that is used to turn a specific set of gates into QubitUnitaries when requested by the device. That mechanism doesn't deal with controlled versions of those gates though, so `Controlled(BlockEncode)` needs the mechanism we are looking at here.

Maybe another test should be added, let me check 🤔

dime10: Thanks Mehrdad, I went back and verified all the tests that would trigger this new branch. Turns out some tests were weak and didn't specify the exact error that was expected.

For differentiation certain ops might need to be decomposed, and if those ops don't have a decomposition they would be turned into unitaries here. That changed the error message from:
```
catalyst.utils.exceptions.CompileError: Operator RX(Traced<~float64[]>with<DynamicJaxprTrace>, wires=[0]) not supported with catalyst on this device and does not provide a decomposition.
```
into:
```
catalyst.utils.exceptions.CompileError: catalyst failed with error code 1: Failed to run pipeline: MLIRToLLVMDialect
Compilation failed:
grad.circuit:56:13: error: failed to legalize operation 'memref.load' that was explicitly marked illegal
      %38 = stablehlo.concatenate %32, %37, dim = 0 : (tensor<1xcomplex<f64>>, tensor<1xcomplex<f64>>) -> tensor<2xcomplex<f64>>
```

The latter is obviously horrible, but I think points to our verification not being strong enough to reject the program 🤔 

For now I'm reverting to only consider controlled operators though.

frontend/catalyst/device/decomposition.py
dime10: I think that is evident from the condition just above (don't particularly like comments that just repeat clear code) 😅 but we might as well update or remove this one

PR #2101, opened 2025-10-08
Actions use rc PennyLane from testpypi instead of git

paul0403:
**Context:**
Change two actions to use rc PennyLane from testpypi instead of git:
- check-catalyst (CI on PRs in catalyst repo targeting the rc branch)
- check-pl-compat (the rcrcrc check CI)

### DIFF
--- a/.github/workflows/check-catalyst.yaml
+++ b/.github/workflows/check-catalyst.yaml
@@ -478,7 +478,7 @@ jobs:
     - name: Install PennyLane release candidate
       if: github.event.pull_request.base.ref == 'v0.13.0-rc'
       run: |
-        pip install --no-deps --force git+https://github.com/PennyLaneAI/pennylane@v0.43.0-rc0
+        pip install --no-deps --pre -U --extra-index-url https://test.pypi.org/simple/ pennylane'<=0.43.0'
 
     - name: Install PennyLane-Lightning release candidate
       if: github.event.pull_request.base.ref == 'v0.13.0-rc'
@@ -571,7 +571,7 @@ jobs:
     - name: Install PennyLane release candidate
       if: github.event.pull_request.base.ref == 'v0.13.0-rc'
       run: |
-        pip install --no-deps --force git+https://github.com/PennyLaneAI/pennylane@v0.43.0-rc0
+        pip install --no-deps --pre -U --extra-index-url https://test.pypi.org/simple/ pennylane'<=0.43.0'
 
     - name: Install PennyLane-Lightning-Kokkos release candidate
       if: github.event.pull_request.base.ref == 'v0.13.0-rc'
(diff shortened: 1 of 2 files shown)


.github/workflows/check-catalyst.yaml
```diff
@@ -643,7 +643,7 @@ jobs:
     - name: Install PennyLane release candidate
       if: github.event.pull_request.base.ref == 'v0.13.0-rc'
       run: |
-        pip install --no-deps --force git+https://github.com/PennyLaneAI/pennylane@v0.43.0-rc0
+        pip install --no-deps --pre -U --extra-index-url https://test.pypi.org/simple/ pennylane'<=0.43.0'
```
andrijapau: Wait this works? lol

andrijapau (approved):

.github/workflows/check-catalyst.yaml
andrijapau: This will always install rc?

andrijapau: I'm gonna steal this

paul0403: Yes! https://peps.python.org/pep-0440/#pre-releases

> The pre-release segment consists of an alphabetical identifier for the pre-release phase, along with a non-negative integer value. Pre-releases for a given release are ordered first by phase (alpha, beta, release candidate) and then by the numerical component within that phase.

So this will look for the latest version that is the pre-release of `0.43.0`, which is the latest rc version!

Alex-Preciado: yeah, I think this, combined with `--pre`, allows installing release candidates if they are available in TestPyPI

paul0403: https://peps.python.org/pep-0440/#summary-of-permitted-suffixes-and-relative-ordering

> Within a numeric release (1.0, 2.7.3), the following suffixes are permitted and MUST be ordered as shown:
> ```
> .devN, aN, bN, rcN, <no suffix>, .postN
> ```

.github/workflows/check-catalyst.yaml
```diff
@@ -478,7 +478,7 @@ jobs:
...
```
Alex-Preciado: Looking good!! One question from my end ... do we need to use `--extra-index-url` here, or should we use `--index-url`?

paul0403: Yes I think, from pip man page 

>        -i,--index-url <url>
>              Base URL of Python Package Index (default https://pypi.python.org/simple/).
>
>       --extra-index-url <url>
>              Extra URLs of package indexes to use in addition to --index-url.

Alex-Preciado: Yeah, that's what I was about to say. I think we want `--index-url` in this case since our release candidates are only stored in TestPyPI. `--extra-index-url` is equivalent to "look at this other index in addition to the default PyPI".

paul0403: Ah yes, good point!

Alex-Preciado (approved):

PR #2102, opened 2025-10-08
rcrcrc action use rc PennyLane from testpypi instead of git

paul0403:
**Context:**
rcrcrc action use rc PennyLane from testpypi instead of git

Note that this needs to be updated on main, not rc branch, since the rcrcrc action is triggered from main.


Alex-Preciado (approved):
Looks good!

PR #2105, opened 2025-10-08
Fixing catalyst measure error message when mcms are used in MP

lazypanda10117:
### Before submitting

**Context:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


frontend/catalyst/jax_tracer.py
```diff
@@ -1048,9 +1063,15 @@ def trace_quantum_measurements(
                 d_wires = len(device.wires)
 
             m_wires = output.wires if output.wires else None
-            obs_tracers, nqubits = trace_observables(output.obs, qrp, m_wires)
+            obs_tracers, nqubits, contains_compbasis = trace_observables(output.obs, qrp, m_wires)
             nqubits = d_wires if nqubits is None else nqubits
 
+            if contains_compbasis and mcm_config.mcm_method == "single-branch-statistics":
```
dime10: I'm not sure I get this flag, why does compbasis imply a mid-circuit measurement?

frontend/catalyst/jax_tracer.py
```diff
@@ -1048,9 +1063,15 @@ def trace_quantum_measurements(
...
+                raise NotImplementedError(
+                    "single-branch-statistics does not support measurement process "
+                    "on mid circuit measurements."
+                )
+
             using_compbasis = obs_tracers.primitive == compbasis_p
 
             if isinstance(output, qml.measurements.SampleMP):
```
dime10, line 1069: I think you can just check the presence of the `mv` field in the MP (`output`), like we do for SampleMP, but disallow it in the other MPs. Probably no complex logic needed.

dime10: There is also the `obs` field, but that is only used if the MP received a pennylane operator, so should always be empty when a measure result is used.
https://github.com/PennyLaneAI/pennylane/blob/2343bf2ccfa2009a49dd8ce602635b380e278ce7/pennylane/measurements/measurements.py#L170-L172

frontend/catalyst/jax_tracer.py
lazypanda10117: I am no longer using this method. Instead, I am now checking for the presence of the `output.mv` field as you've suggested.

frontend/catalyst/jax_tracer.py
lazypanda10117: Yep, i am doing this now instead. Thanks for the suggestion!

frontend/test/pytest/test_mid_circuit_measurement.py
```diff
@@ -420,6 +420,58 @@ def circuit(x):
         with pytest.raises(TypeError, match="postselect must be '0' or '1'"):
             _ = circuit(1.8)
 
+    def test_single_branch_statistics_not_implemented_error(self, backend):
```
dime10: minor suggestion: this is a classic example where pytest parametrize can come in handy over the meausurement classes :)

lazypanda10117: oh right, lemme fix that before merging : )

PR #2112, opened 2025-10-10
Fix a clang-tidy warning on main

paul0403:
**Context:**
A clang-tidy warning on main branch is blocking benchmark runs. 

We do have `-Wall` turned on by default, but we might have slightly different configs than the benchmark suite? 
But anyway, this is the needed fix.

A class is instantiated with some fields missing initial values. 
We provide the empty init values.

**Benefits:**
Benchmark suite runs.

### DIFF
--- a/mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
+++ b/mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
@@ -87,7 +87,11 @@ class OpSignatureAnalyzer {
                                 .inCtrlQubits = op.getCtrlQubitOperands(),
                                 .inCtrlValues = op.getCtrlValueOperands(),
                                 .outQubits = op.getNonCtrlQubitResults(),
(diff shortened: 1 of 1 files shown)


mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
```diff
@@ -87,7 +87,11 @@ class OpSignatureAnalyzer {
...
                                 .outQubits = op.getNonCtrlQubitResults(),
-                                .outCtrlQubits = op.getCtrlQubitResults()})
+                                .outCtrlQubits = op.getCtrlQubitResults(),
+                                .inWireIndices = {},
+                                .inCtrlWireIndices = {},
+                                .outQubitIndices = {},
+                                .outCtrlQubitIndices = {}})
     {
         if (!enableQregMode)
             return;
```
mlxd, line 97: Just an aside comment --- needs braces.
Ignore for now, but let's be sure this standard is maintained,

PR #2117, opened 2025-10-13
Runtime router

ritu-thombre99:
**Context:**

Original PR: https://github.com/PennyLaneAI/catalyst/pull/1928#issue-3251031857

When running a quantum circuit on a hardware with certain connectivity constraints, two-qubit gates like CNOTs can only be executed using physical qubits that are connected by an edge on the hardware.

Hence, necessary SWAPs need to be inserted in order to route the logical qubits so that they are mapped to an edge on the device, while respecting other compiling constraint i.e. order the gate dependencies from the input quantum circuit, and ensuring compiled quantum circuit is equivalent to the input quantum circuit.

Following image shows a simple example:

image

**Benefits:**

1. This will support qubit mapping and routing in the runtime 
2. Analogue to [qml.transpile](https://docs.pennylane.ai/en/stable/code/api/pennylane.transforms.transpile.html) transform
3. Circuits with control structure will also be supported since they will already be optimized by Catalyst passes and unrolled in the runtime.

**Possible Drawbacks:**
1. Modification of RuntimeCAPI 
2. Currently, not checking if device supports SWAP
3. There are multi-qubit gates like Toffoli, MultiRZ, QubitUnitary in RuntimeCAPI.cpp. Routing is not applied to these qubits yet.

**Related GitHub Issues:**

Example usage:

Currently, seeing what SWAP gates are added can be viewed by using `null.qubit` done by printing `NamedOperation` in `runtime/lib/backend/null_qubit/NullQubit.hpp`

```python
import pennylane as qml
import numpy as np
dev = qml.device("null.qubit", wires = [(0,1),(1,2),(2,3)])
@qml.qjit(autograph=True)
@qml.qnode(dev)
def circuit(c: bool, x: int, y: int):
    if c:
        qml.Hadamard(x)
    qml.Hadamard(y)
    qml.CRZ(0.4,[x,y])
    qml.CNOT([x,2])
    qml.CRX(0.6,[2,y])
    qml.CRY(0.2,[x,y])
    qml.IsingXX(0.24,[0,y])
    qml.CNOT([2,4])
    qml.CNOT([0,5])
    return qml.state()
print(circuit(True,0,3))
```

Output:

```pycon
Name: Hadamard
Wires : 0,
Name: Hadamard
Wires : 3,
Name: SWAP
Wires : 0,1,
Name: SWAP
Wires : 1,2,
Name: CRZ
Wires : 2,3,
Name: CNOT
Wires : 2,1,
Name: SWAP
Wires : 1,2,
Name: CRX
Wires : 2,3,
Name: SWAP
Wires : 1,2,
Name: CRY
Wires : 2,3,
Name: IsingXX
Wires : 2,3,
Name: SWAP
Wires : 1,2,
Name: SWAP
Wires : 2,3,
Name: CNOT
Wires : 3,4,
Name: SWAP
Wires : 1,2,
Name: SWAP
Wires : 2,3,
Name: SWAP
Wires : 3,4,
Name: CNOT
Wires : 4,5,
Name: SWAP
Wires : 1,3,
Name: SWAP
Wires : 0,1,
Name: SWAP
Wires : 4,0,
[1.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j
 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j
 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j
 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j
 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j
 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j 0.+0.j
 0.+0.j 0.+0.j 0.+0.j 0.+0.j]
```


frontend/catalyst/device/qjit_device.py
```diff
@@ -319,14 +320,40 @@ def __init__(self, original_device):
...
+            for wire_label in original_device.wires.labels
+        ):
+            wires_from_cmap = set()
+            for wire_label in original_device.wires.labels:
+                wires_from_cmap.add(wire_label[0])
+                wires_from_cmap.add(wire_label[1])
+            wires_from_cmap = qml.wires.Wires(list(wires_from_cmap))
+            # check_device_wires(wires_from_cmap) not called
+            # since automatic qubit management
+            super().__init__(wires=wires_from_cmap, shots=original_device.shots)
```
paul0403, line 334: core PL has deprecated setting shots on devices
```suggestion
            super().__init__(wires=wires_from_cmap)
```

frontend/catalyst/device/qjit_device.py
```diff
@@ -319,14 +320,40 @@ def __init__(self, original_device):
...
+        #       when customizable multi-pathway decomposition is implemented. (Epic 74474)
+        if hasattr(original_device, "_to_matrix_ops"):
+            _to_matrix_ops = getattr(original_device, "_to_matrix_ops")
+            setattr(device_capabilities, "to_matrix_ops", _to_matrix_ops)
+            if _to_matrix_ops and not device_capabilities.supports_operation("QubitUnitary"):
+                raise CompileError(
+                    "The device that specifies to_matrix_ops must support QubitUnitary."
+                )
+        if original_device.wires is not None:
+            setattr(device_capabilities, "coupling_map", original_device.wires.labels)
```
paul0403, line 352: A couple thoughts here:

1. An immediate issue with this logic here is that it would interfere with the regular way of wires, i.e. `qml.device("lightning.qubit", wires=3)`, because `3` is also not `None`. You could do the same check as above, i.e. check that each entry in the `Wires` object is a tuple, but...
2. ... I'm not so sure about the UI. So I guess your UI is to take in the map through the wires ketword on the device? This only happens to work because in PennyLane, the device wires only need to be a list of hashable objects, which are then treated as labels. However in Catalyst we simplify this a bit, by letting the device wires kwarg to just mean "the number of wires this circuit uses". This UI breaks that assumption, and could cause difficulties in other places (for example, there's work going on about formalizing device capacity and qnode algorithm wires, which will likely change the device wire UI).

If the entire purpose is to just have a UI to take in the wire map from the user, I would suggest taking in it through its own dedicated decorator. The decorator would be on the device, and its entire purpose is just set the coupling map:

```python

dev = catalyst.set_coupling_map(dev, [(0,1),(1,2),(2,3)])

@qjit
@qml.qnode(dev)
def circuit():
   ...
```

This way, your functionality is properly modulated. The `set_coupling_map` function could live in a separate file, e.g. `routing.py`, and that module would take care of, e.g. verifiying the map makes sense on the device, etc.

Implementation wise, this also prevents the need to piggy back the map on the capabilities (which I see (a) you need to pass into other functions now, and (b) has its own complicated song and dance already). Your function could just add the map to the `device_kwargs` of the PL device it takes in. This is because in `extract_backend_info`, at the end all PL device kwargs will be moved onto the qjit device:
```python
qjit_device.py

def extract_backend_info(device: qml.devices.QubitDevice):
   device_kwargs = {}
   ...
    for k, v in getattr(device, "device_kwargs", {}).items():
        if k not in device_kwargs:  # pragma: no branch
            device_kwargs[k] = v
    return BackendInfo(...., device_kwargs)
```

frontend/catalyst/device/qjit_device.py
```diff
@@ -319,14 +320,40 @@ def __init__(self, original_device):
...
+            isinstance(wire_label, tuple) and (len(wire_label) >= 2)
+            for wire_label in original_device.wires.labels
+        ):
+            wires_from_cmap = set()
+            for wire_label in original_device.wires.labels:
+                wires_from_cmap.add(wire_label[0])
+                wires_from_cmap.add(wire_label[1])
+            wires_from_cmap = qml.wires.Wires(list(wires_from_cmap))
+            # check_device_wires(wires_from_cmap) not called
+            # since automatic qubit management
```
paul0403, line 333: I don't think your functionality is restricted to automatic management. The number of wires used by the qnode is still known, it's just that there's some connectivity between them. 

The action by your functionality is just inserting SWAP gates right? So it doesn't require allocation of new qubits. The IR would look like the IR with a known number of wires.

frontend/catalyst/device/qjit_device.py
paul0403: Of course, a restriction of doing it through kwargs in the first place is that the coupling map must be static. However as a first draft we don't need to be concerned over this. The verification in `routing.py` can just check this as well.

runtime/lib/capi/RuntimeCAPI.cpp
```diff
@@ -267,8 +269,23 @@ static int __catalyst__rt__device_init__impl(int8_t *rtd_lib, int8_t *rtd_name,
...
+    // else {...,'couplingMap' (a,b,c)}
+    size_t start = args[2].find("coupling_map': ") + 15; // Find key and opening parenthesis
+    size_t end = args[2].find("}", start);               // Find closing parenthesis
+    std::string coupling_map_str = std::string(args[2].substr(start, end - start));
+
+    if (coupling_map_str.find("((") != std::string::npos) {
+        RT_FAIL_IF(
+            !initRTDevicePtr(args[0], args[1], args[2], auto_qubit_management, coupling_map_str),
+            "Failed initialization of the backend device");
+    }
```
paul0403, line 284: Can we do this parsing in [the runtime device's `parse_kwargs` util](https://github.com/PennyLaneAI/catalyst/blob/5eec7c770fc5fd1d5c5a9c03124db3967de8769d/runtime/lib/backend/common/Utils.hpp#L107), instead of here in the stub? This way the parsing functionality is all properly modulated in one place. 

This way you also don't need to change the signature of a bunch of things to take the coupling map in (here in device init capi and in the `RTDevice` class in `ExecutionContext.hpp`).

runtime/lib/capi/Routing.hpp
```diff
@@ -0,0 +1,185 @@
...
+#include <memory>
+#include <ostream>
+#include <set>
+#include <stdexcept>
+#include <string_view>
+#include <tuple>
+
+namespace Catalyst::Runtime {
+
+class RoutingPass final {
```
paul0403, line 30: This isn't a pass since this is runtime : )

runtime/lib/capi/ExecutionContext.hpp
```diff
@@ -170,6 +171,8 @@ class RTDevice {
 
     std::unique_ptr<SharedLibraryManager> rtd_dylib{nullptr};
     std::unique_ptr<QuantumDevice> rtd_qdevice{nullptr};
+    // device specific routing pass pointer.
```
paul0403, line 174: This comment isn't really necessary, there's already the variable name : )

runtime/lib/capi/RuntimeCAPI.cpp
paul0403: Actually, I just realized the `parse_kwargs` is called by the specific devices, not at the `RTDevice` level, so maybe it would be a bit tricky :thinking: 

Maybe it would then make sense to make the parsing into its own separate util function in `Routing.hpp`? Like `parse_coupling_map_from_kwargs` or something, and just call it here?

runtime/lib/capi/RuntimeCAPI.cpp
```diff
@@ -77,9 +77,11 @@ std::vector<bool> getModifiersControlledValues(const Modifiers *modifiers)
  * to the new initialized device pointer.
  */
 [[nodiscard]] bool initRTDevicePtr(std::string_view rtd_lib, std::string_view rtd_name,
-                                   std::string_view rtd_kwargs, bool auto_qubit_management)
+                                   std::string_view rtd_kwargs, bool auto_qubit_management,
+                                   std::string_view coupling_map_str = {})
```
paul0403, line 81: Can we use std::optional? This way in `ExecutionContext.hpp` you can just check against std::nullopt instead of the more fragile 
```C
if (coupling_map_str.find("((") != std::string::npos)
```

runtime/lib/capi/ExecutionContext.hpp
```diff
@@ -320,13 +341,14 @@ class ExecutionContext final {
     }
 
     [[nodiscard]] auto getOrCreateDevice(std::string_view rtd_lib, std::string_view rtd_name,
-                                         std::string_view rtd_kwargs, bool auto_qubit_management)
+                                         std::string_view rtd_kwargs, bool auto_qubit_management,
+                                         std::string_view coupling_map_str = {})
```
paul0403, line 345: Likewise std::optional

runtime/lib/capi/ExecutionContext.hpp
```diff
@@ -224,6 +227,20 @@ class RTDevice {
...
+                      std::string_view _rtd_kwargs, bool _auto_qubit_management,
+                      std::string_view coupling_map_str)
+        : rtd_lib(_rtd_lib), rtd_name(_rtd_name), rtd_kwargs(_rtd_kwargs),
+          auto_qubit_management(_auto_qubit_management)
+    {
+        // Extract coupling map from the kwargs passed
+        // If coupling map is provided then it takes in the form {...,'couplingMap' ((a,b),(b,c))}
+        // else {...,'couplingMap' (a,b,c)}
+        if (coupling_map_str.find("((") != std::string::npos)
+            RUNTIME_ROUTER = std::make_unique<RoutingPass>(coupling_map_str);
```
paul0403, line 240: Just a reminder that we always use braces, even if it's one-line : ) 

```C
// Don't do this 
if (blah)
   thing;

// Do this
if (blah) {
   thing;
}
```

@mlxd I'm spreading the good word

runtime/lib/capi/RuntimeCAPI.cpp
```diff
@@ -554,176 +574,468 @@ void __catalyst__qis__Identity(const Modifiers *modifiers, int64_t numQubits, ..
...
 void __catalyst__qis__PauliX(QUBIT *qubit, const Modifiers *modifiers)
 {
-    getQuantumDevicePtr()->NamedOperation("PauliX", {}, {reinterpret_cast<QubitIdType>(qubit)},
-                                          MODIFIERS_ARGS(modifiers));
+    if (RTD_PTR != nullptr && RTD_PTR->getRuntimeRouter() != nullptr) {
+        QubitIdType mapped_wire =
+            RTD_PTR->getRuntimeRouter()->getMappedWire(reinterpret_cast<QubitIdType>(qubit));
+        getQuantumDevicePtr()->NamedOperation("PauliX", {}, {mapped_wire},
+                                              MODIFIERS_ARGS(modifiers));
+    }
```
paul0403, line 592: So with this functionality, I feel like we will lock down the quantum dialect as a high-level "expression of intent"? 

i.e. when quantum dialect says "perform CNOT on wires 0 and 2", it no longer maps to what's actually happening on the device. The runtime will change this to "perform SWAP on wires 0 and 1, then CNOT on wires 1 and 2", and send it to the device's `NamedOperation`. 

I am not super concerned by this, but just want to point out this subtle lock down of semantics. 

But I guess this isn't too much of an issue, since the entire PL is quite ambiguous already on whether it's high-level or low-level, even at the python layer. 

cc @dime10 @mlxd

runtime/lib/capi/RuntimeCAPI.cpp
```diff
@@ -554,176 +574,468 @@ void __catalyst__qis__Identity(const Modifiers *modifiers, int64_t numQubits, ..
...
 void __catalyst__qis__PauliY(QUBIT *qubit, const Modifiers *modifiers)
 {
-    getQuantumDevicePtr()->NamedOperation("PauliY", {}, {reinterpret_cast<QubitIdType>(qubit)},
-                                          MODIFIERS_ARGS(modifiers));
+    if (RTD_PTR != nullptr && RTD_PTR->getRuntimeRouter() != nullptr) {
+        QubitIdType mapped_wire =
+            RTD_PTR->getRuntimeRouter()->getMappedWire(reinterpret_cast<QubitIdType>(qubit));
+        getQuantumDevicePtr()->NamedOperation("PauliY", {}, {mapped_wire},
+                                              MODIFIERS_ARGS(modifiers));
+    }
```
paul0403, line 606: One thing though, there is a lot of duplicate code here for all the gates. Can you outline them into its own function/macro and put it in `Routing.hpp`? Again, just for better modularity.

Seems like one-qubit and two-qubit gates have different code, but I'll leave you to handle the details.

runtime/lib/capi/RuntimeCAPI.cpp
```diff
@@ -1062,6 +1613,14 @@ void __catalyst__qis__Probs(MemRefT_double_1d *result, int64_t numQubits, ...)
...
 
     if (wires.empty()) {
+        if (RTD_PTR != nullptr && RTD_PTR->getRuntimeRouter() != nullptr) {
+            std::vector<std::tuple<QubitIdType, QubitIdType>> finalSwaps =
+                RTD_PTR->getRuntimeRouter()->getFinalPermuteSwaps();
+            for (auto i = 0; i < finalSwaps.size(); i++) {
+                RTD_PTR->getQuantumDevicePtr()->NamedOperation(
+                    "SWAP", {}, {std::get<0>(finalSwaps[i]), std::get<1>(finalSwaps[i])});
+            }
+        }
```
paul0403, line 1623: I'm a bit worried about swapping back only at the terminal measurement.

I haven't looked into the logic of the router itself, but I assume the `getFinalPermuteSwaps` can properly deal with multiple swaps occurring in the circuit? In other words, there is not need to immediately swap back in each gate, and we're ok with only swapping back at the end when a user requests terminal readouts?

What about mid circuit measurements?

runtime/lib/capi/RuntimeCAPI.cpp
```diff
@@ -1093,6 +1652,14 @@ void __catalyst__qis__Sample(MemRefT_double_2d *result, int64_t numQubits, ...)
...
```
paul0403, line 1662: Similarly, can you outline the terminal measurement's routing code into its own function as well?

paul0403 (commented):
Fantastic work :100: :rocket: This functionality will be very valuable for targeting hardware!

I haven't looked into the logic of the router, just left some initial thoughts and possible improvements on the overall code structure. But I think it looks really great!

PR #2121, opened 2025-10-14
Update to Pylint 4.0.0

paul0403:
**Context:**
Update to Pylint 4.0.0.
This needs to be coordinated cross the ecosystem, to determine the target pylint version we're going to.
For now, I'm saving the work I have done so far on trying to update catalyst to pylint 4.0.0. I am handing this off to @rniczh 

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/frontend/test/pytest/test_gradient.py
+++ b/frontend/test/pytest/test_gradient.py
@@ -1800,7 +1800,7 @@ def test_forloop_vmap_worflow_derivation(backend):
             result_i = circuit(transposed_data[i], weights)
             return result_array.at[i].set(result_i)
 
-        return body(transposed_result) # pylint: disable=no-value-for-parameter
+        return body(transposed_result)  # pylint: disable=no-value-for-parameter
 
     cat_res = qjit(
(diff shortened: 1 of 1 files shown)


paul0403:
> In the past we have kept up with latest, if we now pin across the ecosystem that's fine too

I think the versions are decided at the ecosystem level now.

.github/workflows/check-formatting.yaml
```diff
@@ -14,7 +14,7 @@ jobs:
 
     steps:
       - name: Install dependencies
-        run: pip install black isort pylint==3.3.9
+        run: pip install black isort pylint
```
paul0403: @rniczh can you pin here as well?

rniczh: done https://github.com/PennyLaneAI/catalyst/pull/2121/commits/11af6ac27bc73f886a83de02ff1df1910016f822

rniczh (approved):

PR #2122, opened 2025-10-14
Update LLVM version, 2025 Q4

rniczh:
**Context:**

Update llvm, stablehlo and enzyme, 2025 Q4
The latest pair of good versions, indicated by stablehlo, is https://github.com/openxla/stablehlo/commit/0a4440a5c8de45c4f9649bf3eb4913bf3f97da0d
```
stablehlo=0a4440a5c8de45c4f9649bf3eb4913bf3f97da0d
llvm=113f01aa82d055410f22a9d03b3468fa68600589
```

For Enzyme, we go to the latest release https://github.com/EnzymeAD/Enzyme/releases/tag/v0.0.203
```
enzyme=v0.0.203
```
with commit `476c8e3193a8577ba24ff845ae2294109225f83a`

- `llvm-project/mlir/include/mlir/InitAllPasses.h` header no longer includes individual pass headers. 
To incorporate specific passes, you now need to include either `mlir/Conversion/Passes.h` or the header for each dialect pass. This change stems from the upstream cleanup of `mlir/InitAllDialects.h`, which previously handled these includes. Following the decision from the owner (as seen in this pull request comment: https://github.com/ftynse/water/pull/16/files#r2259526343), we will specifically include each required header, faster compilation (no unnecessary headers). For further context, refer to this PR: https://github.com/llvm/llvm-project/pull/151150
- To bufferize custom tensors into custom buffers. The upstream PR https://github.com/llvm/llvm-project/pull/144867 changed the the return type of `bufferization::detail::defaultGetBufferType`  to be `BufferLikeType` instead of `BaseMemRefType`. We aligned the behaviour with the upstream PR, return the BufferLikeType from our `getBufferType()` implementation as well.
- `elemwise_unary` and `elemwise_binary` of `linalg` have been deprecated from the upstream, replaced with `elementwise`. https://github.com/llvm/llvm-project/pull/147082. Follow this PR, using `linalg.add` directly.
 register_all_dialects/passes/extension` should now be configured as a static library in CMake, refer to this PR: https://github.com/llvm/llvm-project/pull/151150
- An opt-in feature (`{op}::create` method) be added to support the same behaviour of `rewriter.create` method. But it provides a meaningful message when using this new opt-in feature compared to `rewrite.create`. Since the new generated `create` function will call the `build` function, it requires us to define the instance of the builder (`GraidentOps.td`). Refer to the PR: https://github.com/llvm/llvm-project/pull/147168.
**Take `linalg.abs` as an example:**
    ```cpp
    void AbsOp::build(::mlir::OpBuilder &odsBuilder, ::mlir::OperationState &odsState, ValueRange inputs, ValueRange outputs, ArrayRef<NamedAttribute> attributes) {
           buildStructuredOp(odsBuilder, odsState, std::nullopt, inputs, outputs,
             attributes, AbsOp::getRegionBuilder());
        
    }
    
    AbsOp AbsOp::create(::mlir::OpBuilder &builder, ::mlir::Location location, ValueRange inputs, ValueRange outputs, ArrayRef<NamedAttribute> attributes) {
     ::mlir::OperationState __state__(location, getOperationName());
     build(builder, __state__, inputs, outputs, attributes);
     auto __res__ = ::llvm::dyn_cast<AbsOp>(builder.create(__state__));
     assert(__res__ && "builder didn't return the right type");
     return __res__;
    }
    
    AbsOp AbsOp::create(::mlir::ImplicitLocOpBuilder &builder, ValueRange inputs, ValueRange outputs, ArrayRef<NamedAttribute> attributes) {
     return create(builder, builder.getLoc(), inputs, outputs, attributes);
    }
    ```
- Changed `EnzymeStatic-21` to `EnzymeStatic-22`
- Mock `_ods_cext.globals.register_traceback_file_exclusion` due to API conflicts between Catalyst's MLIR version and the MLIR version used by JAX. The current JAX version we used has not yet updated to the latest MLIR, causing compatibility issues. This workaround will be removed once JAX updates to a compatible MLIR version. https://github.com/llvm/llvm-project/pull/151246

TODO: 
- [x] Update .dep_version
- [ ] Change all `rewrite.create` to `{op}::create` 
 Even it’s compatible to the current `rewrite.create` method, we don’t need to change all `rewrite.create` to `{op}::create` right away, but this new feature provides more friendly development experience. Just for an example: https://github.com/llvm/llvm-project/pull/147311

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-100911]
[sc-100912]

### DIFF
--- a/mlir/lib/Quantum/Transforms/decompose_lowering.cpp
+++ b/mlir/lib/Quantum/Transforms/decompose_lowering.cpp
@@ -109,9 +109,7 @@ struct DecomposeLoweringPass : impl::DecomposeLoweringPassBase<DecomposeLowering
                     std::string newTargetOpStr =
                         targetOp.str() + "_" + std::to_string(DecompUtils::getNumWires(func));
 
-                    auto newtargetOp =
-                        targetOp + "_" + std::to_string(DecompUtils::getNumWires(func));
-                    decompositionRegistry[newtargetOp.str()] = func;
+                    decompositionRegistry[newTargetOpStr] = func;
                 }
                 else {
                     decompositionRegistry[targetOp] = func;


frontend/catalyst/jax_extras/patches.py
```diff
@@ -43,6 +43,29 @@
 )
 
 
+def mock_attributes(obj, attrs: dict[str, any]):
```
paul0403, line 46: Is this needed once we go to jax 0.7.2?

mlir/include/Catalyst/IR/CatalystOps.td
```diff
@@ -138,8 +138,17 @@ def CallbackOp : Catalyst_Op<"callback",
 
   let builders = [OpBuilder<(ins
     "mlir::StringRef":$name, "mlir::FunctionType":$type,
-    CArg<"mlir::ArrayRef<mlir::NamedAttribute>", "{}">:$attrs)
-  >];
+    CArg<"mlir::ArrayRef<mlir::NamedAttribute>", "{}">:$attrs), [{
```
paul0403, line 141: So we don't have to update all ops right away, since as far as I understand this is just QoL. However, for the ops that you have updated here to include the new builders, can we change their `rewriter.create<op>` to `op.create` in the codebase? This is because ideally we don't want unused/untested code, even in tablegen definitions.

frontend/catalyst/jax_extras/patches.py
```diff
@@ -43,6 +43,29 @@
...
+    """Mock the attribute of an object by returning a wrapper.
+
+    Args:
+        obj: The object to mock the attributes of.
+        attrs: A dictionary of attributes to mock.
+            Example: {"attribute_name": attribute_value}
+    """
+
+    class MockAttributeWrapper:
```
paul0403, line 55: Why not use [the patcher infra we already have](https://github.com/PennyLaneAI/catalyst/blob/aa22429cacd374baa0c224e8b8e2deff4b479ccc/frontend/catalyst/utils/patching.py#L20)? That infra also properly restores the third party object to the original one after exiting catalyst.

All you want is to patch the `jaxlib.blahblah._ods_cext.globals` object to have the attribute `register_traceback_file_exclusion` with value `lambda x: None` right? Can't we just do 
```
_ods_cext_globals2 = jaxlib.blahblah._ods_cext.globals
_ods_cext_globals2.register_traceback_file_exclusion = lambda x: None
```

and then where you need it (probably in the lowering rules), do
```
with Patcher((jax.blahblah._ods_cext, "globals", _ods_cext_globals2))
```
?

paul0403: Ah, I just went to check it, the `_ods_cext` object is a python module, so it might not be copyable... I guess this is why you used a separate pool of attributes...

paul0403: But you can still do this right? 
```python
import jaxlib
from jaxlib.mlir._mlir_libs import _mlir as _ods_cext

import catalyst
from catalyst.utils.patching import Patcher

class MyGlobals:
	def __init__(self, original_globals, attr_patches):
		self.original_globals = original_globals
		self.attr_patches = attr_patches

	def __getattr__(self, attr):
		if attr in self.attr_patches:
			return self.attr_patches[attr]
		return original_globals.attr

my_globals = MyGlobals(_ods_cext.globals, {"thing": "stuff"})
with Patcher((_ods_cext, "globals", my_globals)):
	print(_ods_cext.globals.thing)

print(_ods_cext.globals.thing)

```

```
stuff
  File "/home/paul.wang/catalyst_new/catalyst/temp.py", line 21, in <module>
    print(_ods_cext.globals.thing)
          ^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: '_Globals' object has no attribute 'thing'

```

paul0403: i.e. still use the separate pool of attributes, but just go through the patcher infra for proper restoring when exiting catalyst.

frontend/catalyst/jax_extras/patches.py
rniczh: Probably no, but I will give it a try in jax 0.7.2, if it's going well then I will remove those things from 0.7.2 update

mlir/include/Catalyst/IR/CatalystOps.td
rniczh: The CallbackOp create method is already being used, just not in C++ code directly. The python bindings in `frontend/catalyst/jax_primitives.py` create CallbackOp via:

```python
callbackOp = CallbackOp(f"callback_{name}_{callback_id}", *attrs)
```

This constructor calls the generated `CallbackOp::create()` method, which in turn calls the `build()` implementation. The generated code at `mlir/build/include/Catalyst/IR/CatalystOps.cpp.inc`, So the `build` is required.

```cpp
CallbackOp CallbackOp::create(::mlir::OpBuilder &builder, ::mlir::Location location, mlir::StringRef name, mlir::FunctionType type, mlir::ArrayRef<mlir::NamedAttribute> attrs) {
  ::mlir::OperationState __state__(location, getOperationName());
  build(builder, __state__, name, type, attrs);
  auto __res__ = ::llvm::dyn_cast<CallbackOp>(builder.create(__state__));
  assert(__res__ && "builder didn't return the right type");
  return __res__;
}
```

frontend/catalyst/jax_extras/patches.py
rniczh: Fixed https://github.com/PennyLaneAI/catalyst/pull/2122/commits/3d07aac0858b3cf390efafffd01b986af7baa9be

frontend/catalyst/__init__.py
```diff
@@ -68,21 +69,41 @@
...
+    (
+        _ods_cext,
+        "globals",
+        mock_attributes(
+            # pylint: disable=c-extension-no-member
+            _ods_cext.globals,
+            {"register_traceback_file_exclusion": lambda x: None},
+        ),
+    ),
+):
```
paul0403: I'm surprised that patching in init.py worked, since it looks like after the user's `import catalyst` line the patch will be undone...

But since CI is passing I guess everything's fine?...

Can you try patching this only around the lowering fucntions in `jit.py`? (I think it's called `generate_ir` or something)

rniczh: The reason it's placed here is due to the import in `__init.py__`, not due to lowering, but yup, we can narrow the scope to only happen in `jax_primitives.py` https://github.com/PennyLaneAI/catalyst/pull/2122/commits/11aee123f6758e23e9de49de23477a29c4c82fd2
```python
    import catalyst
frontend/catalyst/__init__.py:91: in <module>
    from catalyst import debug, logging, passes
frontend/catalyst/debug/__init__.py:17: in <module>
    from catalyst.debug.callback import callback
frontend/catalyst/debug/callback.py:16: in <module>
    from catalyst.api_extensions.callbacks import base_callback
frontend/catalyst/api_extensions/__init__.py:19: in <module>
    from catalyst.api_extensions.callbacks import accelerate, pure_callback
frontend/catalyst/api_extensions/callbacks.py:35: in <module>
    from catalyst.jax_primitives import python_callback_p
frontend/catalyst/jax_primitives.py:55: in <module>
    from mlir_quantum.dialects.catalyst import (
mlir/build/python_packages/quantum/mlir_quantum/dialects/catalyst.py:17: in <module>
    from ._catalyst_ops_gen import *  # noqa: F401
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
mlir/build/python_packages/quantum/mlir_quantum/dialects/_catalyst_ops_gen.py:12: in <module>
    _ods_cext.globals.register_traceback_file_exclusion(__file__)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E   AttributeError: 'jaxlib.mlir._mlir_libs._mlir._Globals' object has no attribute 'register_traceback_file_exclusion'
```

frontend/catalyst/__init__.py
```diff
@@ -68,6 +68,26 @@
 sys.modules["mlir_quantum.ir"] = __import__("jaxlib.mlir.ir").mlir.ir
 sys.modules["mlir_quantum._mlir_libs"] = __import__("jaxlib.mlir._mlir_libs").mlir._mlir_libs
 
+# # Mock _ods_cext.globals.register_traceback_file_exclusion due to API conflicts between
```
paul0403: Remember to remove those : )

frontend/catalyst/__init__.py
paul0403: Beautiful, thanks!

Was all of this just for a single import line (that isn't even meaningfully used I guess, since you patched it with an empty lambda) in a dialect's autogenerated python bindings? :sob:

frontend/catalyst/__init__.py
rniczh: Thanks for the reminder! https://github.com/PennyLaneAI/catalyst/pull/2122/commits/73d87139387098d545f212a6af189c794dd87295

frontend/catalyst/jax_primitives.py
```diff
@@ -52,54 +53,83 @@
...
-    SampleOp,
-    SetBasisStateOp,
-    SetStateOp,
-    StateOp,
-    TensorOp,
-    VarianceOp,
-)
-from mlir_quantum.dialects.quantum import YieldOp as QYieldOp
+
+# Mock _ods_cext.globals.register_traceback_file_exclusion due to API conflicts between
```
dime10, line 58: Makes it easier to find these temporary code lines :)
```suggestion
# TODO: remove after jax v0.7.2 upgrade
# Mock _ods_cext.globals.register_traceback_file_exclusion due to API conflicts between
```

frontend/catalyst/__init__.py
rniczh: Yes, for now. Interesting is that thing is from this PR https://github.com/llvm/llvm-project/pull/151246 and come from jax team, in latest jax version they allow to track the loc info of mlir from python. So in latest jax version, enable `--use-nameloc-as-prefix` can have the feature similar to what this PR did https://github.com/PennyLaneAI/catalyst/pull/2054 from our side 😅

dime10 (approved):

PR #2125, opened 2025-10-15
Merge v0.13 rc back to main

paul0403:
**Context:**
Merge v0.13.0-rc back into main.

Merge after core PL is released.


.dep-versions
```diff
@@ -10,8 +10,8 @@ enzyme=v0.0.186
 
 # For a custom PL version, update the package version here and at
 # 'doc/requirements.txt'
-pennylane=0.43.0.dev71
+pennylane=0.43.0
```
paul0403, line 13: Do we usually go back to the stable version or the stable-dev0?

dime10: I don't think there is much of a difference 🤔 You would start with basic release until we need a higher version for some reason

dime10 (approved):

.dep-versions
maliasadi: IIRC It's usually dev instead of stable

dime10: dev0 might be behind stable in some important ways, I would not recommend it

paul0403: True, dev0 is actually start of feature freeze so it's very behind.

I'll just stick with stable then. Or, maybe go to the latest dev on testpypi?

PR #2128, opened 2025-10-16
Fix weird consts issue in `from_plxpr`

albi3ro:
**Context:**

When updating tests for #2078 , I was getting weird failures on this test: https://github.com/PennyLaneAI/catalyst/blob/5eec7c770fc5fd1d5c5a9c03124db3967de8769d/frontend/test/pytest/test_gradient.py#L757

Basically what was happening was that the conversion of the target jaxpr was reordering the consts.   So we need to use the consts for the *new jaxpr*, not the *input jaxpr*. 

In all relevant translation steps, we pull the consts from the new jaxpr, not the original inputs.

**Benefits:**

Get correct results without errors.

**Possible Drawbacks:**

I didn't test all the other control flow primitives that might have the same issue, because I don't know how to trigger this is other cases.

**Related GitHub Issues:**

[sc-101609]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -12,6 +12,7 @@
 
 * Fixes the translation of plxpr control flow for edge cases where the `consts` were being
   reordered.
+  [(#2128)](https://github.com/PennyLaneAI/catalyst/pull/2128)
 
 <h3>Internal changes ⚙️</h3>
 


doc/releases/changelog-dev.md
```diff
@@ -10,6 +10,9 @@
 
 <h3>Bug fixes 🐛</h3>
 
+* Fixes the translation of plxpr control flow for edge cases where the `consts` were being
+  reordered.
```
albi3ro, line 14:
```suggestion
  reordered.
  [(#2128)](https://github.com/PennyLaneAI/catalyst/pull/2128)
```

PR #2131, opened 2025-10-20
Bump jax to 0.7.0

rniczh:
**Context:**

- The API of `lower_jaxpr_to_fun` has changed, replacing `public` with `main_function`. Now, custom function names (e.g., `jit_…`) were not permitted, with `main` being the default. To address this, `main_function` is now set to False, while `sym_visibility` is set to True for MLIR, retaining the `jit_func` function name. Additionally, `name_stack` has been removed. https://github.com/jax-ml/jax/pull/29783
- Params are now required to be hashable. (PL) https://github.com/jax-ml/jax/pull/29498
- `make_var` and `get_var` has been removed. Should set the var (`val`) when initializing the `DynamicJaxprTracer`. https://github.com/jax-ml/jax/pull/29937. There is a typo in jax https://github.com/jax-ml/jax/pull/32718, we will patch the installation of jax in this PR.
- Replace `flatten_lowering_ir_args` with `flatten_ir_values` https://github.com/jax-ml/jax/pull/29706
- jax introduced the `TracingEqn` and replace `frame.eqns` (originally using `JaxprEqn`) with `frame.tracing_eqns`. https://github.com/jax-ml/jax/pull/30135
- `pjit_p` has been removed, use `jit_p` instead
- Fix jax source code by patching it when after jax is installed (run `bin/patch_jax_installation.py`)
- Fix jax source code by patching it with (`frontend/catalyst/jax_extras/patches.py`)

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -223,7 +223,7 @@ wheel:
 	# Copy mlir bindings & compiler driver to frontend/mlir_quantum
 	mkdir -p $(MK_DIR)/frontend/mlir_quantum/dialects
 	cp -R $(COPY_FLAGS) $(DIALECTS_BUILD_DIR)/python_packages/quantum/mlir_quantum/runtime $(MK_DIR)/frontend/mlir_quantum/runtime
-	for file in gradient quantum _ods_common catalyst mbqc mitigation _transform; do \
+	for file in gradient quantum _ods_common catalyst mbqc mitigation qec _transform; do \
 		cp $(COPY_FLAGS) $(DIALECTS_BUILD_DIR)/python_packages/quantum/mlir_quantum/dialects/*$${file}* $(MK_DIR)/frontend/mlir_quantum/dialects ; \
 	done
 	mkdir -p $(MK_DIR)/frontend/bin
--- a/frontend/catalyst/jax_tracer.py
+++ b/frontend/catalyst/jax_tracer.py
@@ -258,7 +258,7 @@ KNOWN_NAMED_OBS = (qml.Identity, qml.PauliX, qml.PauliY, qml.PauliZ, qml.Hadamar
 # Take care when adding primitives to this set in order to avoid introducing a quadratic number of
 # edges to the jaxpr equation graph in ``sort_eqns()``. Each equation with a primitive in this set
 # is constrained to occur before all subsequent equations in the quantum operations trace.
-FORCED_ORDER_PRIMITIVES = {device_init_p, gphase_p}
+FORCED_ORDER_PRIMITIVES = {device_init_p, gphase_p, set_basis_state_p, set_state_p}
 
 PAULI_NAMED_MAP = {
     "I": "Identity",
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -206,6 +206,10 @@
   one-shot qnode when `one-shot` mcm method is used.
   [(#2198)](https://github.com/PennyLaneAI/catalyst/pull/2198)
 
+* Fixed a bug where `qml.StatePrep` and `qml.BasisState` might be pushed after other
+  gates, overwriting their effects.
+  [(#2239)](https://github.com/PennyLaneAI/catalyst/pull/2239)
+
 <h3>Internal changes ⚙️</h3>
 
 * Resource tracking now writes out at device destruction time instead of qubit deallocation
--- a/frontend/test/lit/test_skip_initial_state_prep.py
+++ b/frontend/test/lit/test_skip_initial_state_prep.py
@@ -73,3 +73,49 @@ def state_prep_example_double():
 #       CHECK:   quantum.set_state
 #   CHECK-NOT:   quantum.set_state
 print(state_prep_example_double.mlir)
+
+
+@qjit(target="mlir")
+@qml.qnode(qml.device("lightning.qubit", wires=2))
+def state_prep_trotter():
+    """
+    Test state prep used with Trotter Product.
+    https://github.com/PennyLaneAI/catalyst/issues/2235
+    """
+    H = qml.Hamiltonian([1, 1], [qml.PauliZ(0), qml.PauliZ(1)])
+    qml.StatePrep([0, 0, 0, 1], wires=[0, 1])
+    qml.Hadamard(wires=2)
+    qml.ctrl(
+        qml.TrotterProduct(H, 1, n=3, order=2),
+        control=2,
+    )
+    return qml.probs(wires=[2])
+
+
+# CHECK-LABEL: func.func public @state_prep_trotter
+# CHECK: quantum.set_state
+# CHECK: quantum.custom "Hadamard"
+print(state_prep_trotter.mlir)
+
+
+@qjit(target="mlir")
+@qml.qnode(qml.device("lightning.qubit", wires=2))
+def basis_state_trotter():
+    """
+    Test basis state used with Trotter Product.
+    https://github.com/PennyLaneAI/catalyst/issues/2235
+    """
+    H = qml.Hamiltonian([1, 1], [qml.PauliZ(0), qml.PauliZ(1)])
+    qml.BasisState([1, 1], wires=[0, 1])
+    qml.Hadamard(wires=2)
+    qml.ctrl(
+        qml.TrotterProduct(H, 1, n=3, order=2),
+        control=2,
+    )
+    return qml.probs(wires=[2])
+
+
+# CHECK-LABEL: func.func public @basis_state_trotter
+# CHECK: quantum.set_basis_state
+# CHECK: quantum.custom "Hadamard"
+print(basis_state_trotter.mlir)
--- a/frontend/test/pytest/test_skip_initial_state_prep.py
+++ b/frontend/test/pytest/test_skip_initial_state_prep.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Test skipping iniitial state prep"""
+"""Test skipping initial state prep"""
 
 import jax.numpy as jnp
 import pennylane as qml
--- a/frontend/test/pytest/test_state_prep.py
+++ b/frontend/test/pytest/test_state_prep.py
@@ -120,5 +120,48 @@ class Test2QubitStatePrep:
         assert np.allclose(result_obs, result_exp)
 
 
+class TestTrotterProduct:
+    """
+    Test state prep and basis state when used with TrotterProduct.
+    https://github.com/PennyLaneAI/catalyst/issues/2235
+    """
+
+    def test_state_prep_trotter(self, backend):
+        """
+        Test state prep.
+        """
+
+        @qml.qnode(qml.device(backend, wires=3))
+        def circ():
+            H = qml.Hamiltonian([1, 1], [qml.PauliZ(0), qml.PauliZ(1)])
+            qml.StatePrep([0, 0, 0, 1], wires=[0, 1])
+            qml.Hadamard(wires=2)
+            qml.ctrl(
+                qml.TrotterProduct(H, 1, n=3, order=2),
+                control=2,
+            )
+            return qml.probs(wires=[2])
+
+        assert np.allclose(circ(), qjit(circ)())
+
+    def test_basis_state_trotter(self, backend):
+        """
+        Test basis state.
+        """
+
+        @qml.qnode(qml.device(backend, wires=3))
+        def circ():
+            H = qml.Hamiltonian([1, 1], [qml.PauliZ(0), qml.PauliZ(1)])
+            qml.BasisState(np.array([1, 1]), wires=[0, 1])
+            qml.Hadamard(wires=2)
+            qml.ctrl(
+                qml.TrotterProduct(H, 1, n=3, order=2),
+                control=2,
+            )
+            return qml.probs(wires=[2])
+
+        assert np.allclose(circ(), qjit(circ)())
+
+
 if __name__ == "__main__":
     pytest.main(["-x", __file__])


frontend/catalyst/__init__.py
```diff
@@ -68,9 +71,20 @@
...
 sys.modules["mlir_quantum._mlir_libs"] = __import__("jaxlib.mlir._mlir_libs").mlir._mlir_libs
 
+from jaxlib.mlir._mlir_libs import _mlir as _ods_cext
 from catalyst.jax_extras.patches import patch_primitives
+from catalyst.jax_extras.patches import mock_attribute
 
 patch_primitives()
+_ods_cext.globals = mock_attribute(
+    _ods_cext.globals, "register_traceback_file_exclusion", lambda x: None
+)
```
JerryChen97: Should PennyLane also do this mock? @rniczh

rniczh: I'm not super sure, have you encountered any issue without this?

.dep-versions
```diff
@@ -1,7 +1,7 @@
 # Always update the version check in catalyst.__init__ when changing the JAX version.
 # To update JAX version alongside compatible dependency tags, run the following script:
 # python3 .github/workflows/set_dep_versions.py {JAX_version}
-jax=0.6.2
+jax=8bc0ae6da4f6d334d03be7b76be841d72a56d330
```
paul0403: Is this supposed to be `jax=0.7.0`?

bin/patch_jax_installation.py
dime10, line 1: I don't think we can move forward with this, physically writing to another installed package is not something we should do.

dime10 (changes requested):

bin/patch_jax_installation.py
```diff
@@ -0,0 +1,59 @@
...
+def find_package_root(package_name):
+    """Find a package path in a package."""
+    # resolve package name to package
+    package = importlib.import_module(package_name)
+    return os.path.dirname(package.__file__)
+
+
+jax_root = find_package_root("jax")
+
+patch_setup = {
```
paul0403: So we discovered that these patterns only exist in one function, `pjit.py/pjit_staging_rule`, which is only ~100 lines.

Can you try moving this into `jax_extras/patches.py` as a regular patch, instead of modifying the source code?

.dep-versions
paul0403: image
ok I think so, this commit tag is just 0.7.0

frontend/catalyst/from_plxpr/control_flow.py
```diff
@@ -56,15 +60,15 @@ def _to_bool_if_not(arg):
 @WorkflowInterpreter.register_primitive(plxpr_cond_prim)
 def workflow_cond(self, *plxpr_invals, jaxpr_branches, consts_slices, args_slice):
     """Handle the conversion from plxpr to Catalyst jaxpr for the cond primitive"""
-    args = plxpr_invals[args_slice]
+    args = plxpr_invals[_tuple_to_slice(args_slice)]
```
paul0403, line 114: Is this because @JerryChen97 is making the slice params of things like for loop primitives into tuples, for the new hashability requirement for params in jax?

frontend/catalyst/from_plxpr/control_flow.py
```diff
@@ -56,15 +60,15 @@ def _to_bool_if_not(arg):
...
     converted_jaxpr_branches = []
     all_consts = []
 
     # Convert each branch from plxpr to jaxpr
     for const_slice, plxpr_branch in zip(consts_slices, jaxpr_branches):
 
         # Store all branches consts in a flat list
-        branch_consts = plxpr_invals[const_slice]
+        branch_consts = plxpr_invals[_tuple_to_slice(const_slice)]
```
paul0403, line 122: Also the names are a bit misleading now. Does plxpr still pass a slice to us? Or do they pass a tuple?

Since this is params from plxpr equations, I guess the `const_slice` is actually a tuple now? Since plxpr equations are storing tuples as params now I assume?

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -55,6 +55,61 @@
...
+    This function converts them back to slice objects for use with indexing.
+
+    Args:
+        t: Either a slice object (returned as-is) or a tuple (start, stop, step)
+
+    Returns:
+        slice: A slice object
+    """
+    if isinstance(t, slice):
+        return t
```
```suggestion
```

This branch is unnecessary.

mlir/Enzyme
paul0403, line 1: Why are the mlir submodules changed in this PR? This PR is targeting your llvm update base PR right? :thinking:

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -55,6 +55,61 @@
...
+    if isinstance(t, tuple) and len(t) == 3:
+        return slice(*t)
+    return t
+
+
+def _tuple_to_dict(t):
```
paul0403: This function feels super confusing :sweat_smile: Can you try this simplified version from Gemini?

```
def _is_dict_like_tuple(t):
    """Checks if a tuple t is structured like a list of (key, value) pairs."""
    return isinstance(t, tuple) and all(isinstance(item, tuple) and len(item) == 2 for item in t)

def _tuple_to_dict(t):
    """
    Recursively converts JAX-hashable tuple representations back to dicts, 
    and list-like tuples back to lists.
    
    Args:
        t: The item to convert. Can be a dict, a tuple, or a scalar.
        
    Returns:
        The converted dict, list, or the original scalar value.
    """
    
    # ------------------
    # Base Case (No recursion needed for scalars)
    # ------------------
    if not isinstance(t, (dict, tuple, list)):
        return t

    # ------------------
    # Recursive Steps
    # ------------------
    
    # 1. Handle actual dict objects (recursively process values)
    if isinstance(t, dict):
        return {k: _tuple_to_dict(v) for k, v in t.items()}

    # 2. Handle list objects (recursively process elements)
    if isinstance(t, list):
        return [_tuple_to_dict(item) for item in t]
    
    # 3. Handle tuple objects (The core JAX conversion logic)
    if isinstance(t, tuple):
        
        # A. Dict-like tuple: Convert to dict, then recurse on values
        if _is_dict_like_tuple(t):
            # This handles the main (key, value) pair structure
            return {key: _tuple_to_dict(value) for key, value in t}
        
        # B. List-like tuple: Convert to list, then recurse on elements
        else:
            # This handles tuples used to represent lists
            return [_tuple_to_dict(item) for item in t]

    return t # Should only be hit if logic above is incomplete, but included for safety
```

frontend/test/pytest/from_plxpr/test_from_plxpr_qubit_handler.py
```diff
@@ -239,13 +239,19 @@ def test_auto_extract(self):
         assert qubit_handler[0] is new_qubit
         with take_current_trace() as trace:
             # Check that an extract primitive is added
-            assert trace.frame.eqns[-1].primitive is qextract_p
+            last_eqn = trace.frame.tracing_eqns[-1]
+            if callable(last_eqn):
+                last_eqn = last_eqn()
```
paul0403: Can we use the `_get_eqn_from_tracing_eqn` you added instead? This will also test that new util.

.dep-versions
dime10: definitely use the version number a hash :)

.dep-versions
```diff
@@ -10,7 +10,7 @@ enzyme=v0.0.203
 
 # For a custom PL version, update the package version here and at
 # 'doc/requirements.txt'
-pennylane=0.43.0
+pennylane=bump-jax-to-0.7.0
```
paul0403: What we usually do to install temporary branches from PL or lightning is to just modify the makefile and CI scripts (check-catalyst.ymal, and the wheels) to `pip install git+https://github.com/PennyLaneAI/pennylane@bump-jax-to-0.7.0` directly in the "make frontend" steps. That way the make frontend itself will install whatever version from .dep-versions, but you manually overwrite it with the PL branch you want afterwards.

This way you don't need to change the version here in .dep-versions, and you don't need the little `is_git_reference` song and dance in setup.py :wink:

paul0403: It's also ok if you don't do the wheels now, since this is just a temporary dependency on a PL's PR branch after all.

frontend/catalyst/from_plxpr/control_flow.py
JerryChen97: Yes

frontend/catalyst/from_plxpr/control_flow.py
JerryChen97: A tuple now; plxpr can't accept native slice (for py 3.11) due to the same hashability issue

frontend/catalyst/__init__.py
paul0403: Is this outdated? I didn't see this code during the review.

bin/patch_jax_installation.py
rniczh: removed https://github.com/PennyLaneAI/catalyst/pull/2131/commits/5d1e65d63dca19939d9a578e2acdc5a0b5f8bcc7

bin/patch_jax_installation.py
rniczh: Removed https://github.com/PennyLaneAI/catalyst/pull/2131/commits/5d1e65d63dca19939d9a578e2acdc5a0b5f8bcc7

frontend/test/pytest/from_plxpr/test_from_plxpr_qubit_handler.py
rniczh: Fixed https://github.com/PennyLaneAI/catalyst/pull/2131/commits/345911f362c6b4b0e7f3460fdb860b73f7712e47

.dep-versions
rniczh: I converted it back it to the version number https://github.com/PennyLaneAI/catalyst/pull/2131/commits/55ac16121edefb9fd42810269e499e373222ea58

frontend/catalyst/__init__.py
rniczh: @paul0403 Yes, this has already been moved to the parent PR (LLVM update) https://github.com/PennyLaneAI/catalyst/pull/2122

JerryChen97:
These two new fails should be from the newest update today on PL side regarding `is_hermitian` -> `is_verified_hermitian`

frontend/catalyst/from_plxpr/control_flow.py
paul0403: In that case we should update the variable names.

Makefile
```diff
@@ -117,6 +117,7 @@ frontend:
 	# versions of a package with the same version tag (e.g. 0.38-dev0).
 	$(PYTHON) -m pip uninstall -y pennylane
 	$(PYTHON) -m pip install -e . --extra-index-url https://test.pypi.org/simple $(PIP_VERBOSE_FLAG)
+	$(PYTHON) -m pip install git+https://github.com/PennyLaneAI/pennylane@bump-jax-to-0.7.0
```
paul0403:
```suggestion
    # TODO: remove after https://github.com/PennyLaneAI/pennylane/pull/8525 is merged.
	$(PYTHON) -m pip install git+https://github.com/PennyLaneAI/pennylane@bump-jax-to-0.7.0
```

Let's add a reminder.

paul0403: (Same with the commented out version in `.dep-versions`.)

frontend/catalyst/jax_extras/patches.py
```diff
@@ -231,3 +234,233 @@ def _gather_shape_rule_dynamic(
...
+            else:
+                out_tracers = trace.default_process_primitive(
+                    jit_p, args, params, source_info=source_info
+                )
+            return out_tracers
+
+        jax._src.pjit.pjit_staging_rule = (
+            patched_pjit_staging_rule  # pylint: disable=protected-access
+        )
+        pe.custom_staging_rules[jit_p] = patched_pjit_staging_rule
```
paul0403: The issue with patching this way is that it permanently changes jax source code everytime a user `import catalyst`. So if the user is using catalyst and then later jax in the same python program, jax code will not work with itself.

Our `Patcher` infra uses a context manager to restore jax code back as soon as we don't need the patch anymore: https://github.com/PennyLaneAI/catalyst/blob/main/frontend/catalyst/utils/patching.py#L40. 

Let's identify where your patches are needed, and do the same thing. For example, https://github.com/PennyLaneAI/catalyst/blob/main/frontend/catalyst/jit.py#L730

Very likely, there's some place in `jit.py`'s `capture` method (not plxpr capture; the literal `QJIT.capture` function) where we can apply the patches. I'm pretty sure that the patches here are only needed for the `capture` function (which is essentially just the tracing phase).

Makefile
rniczh: `.dep-versions`: https://github.com/PennyLaneAI/catalyst/pull/2131/commits/a4838979be4081bdf0cc17b0ace455df7936e6d9

frontend/catalyst/from_plxpr/from_plxpr.py
JerryChen97: https://github.com/PennyLaneAI/catalyst/pull/2131/commits/297c7d618a683720e25ecfbbd3548d0df23af74a

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -55,6 +55,74 @@
...
+        t: The item to convert. Can be a dict, a tuple, or a scalar.
+
+    Returns:
+        The converted dict, list, or the original scalar value.
+    """
+
+    # ------------------
+    # Base Case (No recursion needed for scalars)
+    # ------------------
+    if not isinstance(t, (dict, tuple, list)):
```
paul0403: Let's remove all the gemini comments :sweat_smile:

frontend/catalyst/jax_extras/patches.py
rniczh: Thanks! that's really helpful! I added those cases to patcher https://github.com/PennyLaneAI/catalyst/pull/2131/commits/d6d1c6a76b4b979bdc2b6aea510fb1881df70441 https://github.com/PennyLaneAI/catalyst/pull/2131/commits/709cb873dffbb4296f0b442baf9ce0a57f744eed

frontend/catalyst/from_plxpr/from_plxpr.py
rniczh: Done https://github.com/PennyLaneAI/catalyst/pull/2131/commits/aca66f0bbeb7e04c880f7d08412b587476fb5fc3

frontend/catalyst/jax_extras/patches.py
```diff
@@ -231,3 +237,222 @@ def _gather_shape_rule_dynamic(
...
+    original_drop_unused_vars = pe._drop_unused_vars
+
+    # ClosedJaxpr requires constvals to be a list
+    def patched_drop_unused_vars(
+        constvars, constvals, eqns=None, outvars=None
+    ):  # pylint: disable=unused-argument
+        constvars, constvals = original_drop_unused_vars(constvars, constvals, eqns, outvars)
+        return constvars, list(constvals)
+
+    return patched_drop_unused_vars
```
paul0403: Just a nitpick: why the nested defs?

Actually, why a getter in the first place? :sweat_smile: Can't we just patch like 
```
with Patcher(pe, "_drop_unused_vars", patched_drop_unused_vars)
```

directly, instead of 
```
with Patcher(pe, "_drop_unused_vars", get_patched_drop_unused_vars())
```
? :thinking:

frontend/catalyst/jit.py
```diff
@@ -726,13 +736,31 @@ def capture(self, args, **kwargs):
...
+        patched_pjit_staging_rule = get_patched_pjit_staging_rule()
+
         if qml.capture.enabled():
             with Patcher(
                 (
                     jax._src.interpreters.partial_eval,  # pylint: disable=protected-access
                     "get_aval",
                     get_aval2,
                 ),
+                (pe, "_drop_unused_vars", patched_drop_unused_vars),
```
paul0403: Since these patches are common across the plxpr and non plxpr path, we can just do them once at the start when entering this `capture` function.

frontend/catalyst/jax_extras/patches.py
rniczh: Sure, make sense https://github.com/PennyLaneAI/catalyst/pull/2131/commits/ebff8a4551a6e44c04bdde6945c92e45e10473cd

frontend/catalyst/jit.py
rniczh: Done https://github.com/PennyLaneAI/catalyst/pull/2131/commits/ebff8a4551a6e44c04bdde6945c92e45e10473cd

frontend/catalyst/from_plxpr/control_flow.py
rniczh: Sure, https://github.com/PennyLaneAI/catalyst/pull/2131/commits/020537f938ef9f3bb4e13b648fc3a952de455849
For more readable, I added some comment on those args as well

paul0403 (approved):
Fantastic job :pray: :partying_face:

.dep-versions
```diff
@@ -10,7 +10,9 @@ enzyme=v0.0.203
 
 # For a custom PL version, update the package version here and at
 # 'doc/requirements.txt'
-pennylane=0.44.0.dev17
+# TODO: uncomment and update to latest version of pennylane
+#       after https://github.com/PennyLaneAI/pennylane/pull/8525 is merged.
```
dime10, line 14: Are we waiting on PL to merge first?

frontend/catalyst/__init__.py
```diff
@@ -77,17 +76,13 @@
...
-#             _ods_cext.globals,
-#             {"register_traceback_file_exclusion": lambda x: None},
-#         ),
-#     ),
-# ):
+# Disable JAX's Shardy partitioner for JAX 0.7+ compatibility
+# Shardy adds 'sdy' dialect attributes that Catalyst doesn't support yet
+import jax
+
+jax.config.update("jax_use_shardy_partitioner", False)
```
dime10: Can this config be updated during tracing (or lowering) only? Someone added a specific function for this so we could do it temporarily (search `transient_jax_config`)

frontend/catalyst/jit.py
```diff
@@ -26,6 +26,10 @@
 import jax
 import jax.numpy as jnp
 import pennylane as qml
+from jax._src.interpreters import partial_eval as pe
+from jax._src.interpreters.partial_eval import DynamicJaxprTrace
+from jax._src.lax import lax
+from jax._src.pjit import jit_p
```
dime10: I will say this is a bit of a bad sign, since we should avoid jax internal imports as much as we can, and if we really have to they should be encapsulated in the tracing or jax_extras module 🤔 

I guess they're needed here since these are the functions we're patching, but I wonder if it's worth it to move the patching into the tracing functions instead (like trace_from_pennylane/trace_to_jaxpr)? Not a must, but it is getting unwieldy in here (and the matches are not related to the core functionality of this module.

dime10 (approved):
Thanks! Just a couple of extra comments:

.dep-versions
JerryChen97: I think the plan is Catalyst go first? Current PL CI is dependent on Catalyst actually (for the external tests) @rniczh

paul0403: Catalyst will merge first, then PL, then we will track the new PL from testpypi in a regular dev version.

frontend/catalyst/jit.py
rniczh: Sure, that make sense. https://github.com/PennyLaneAI/catalyst/pull/2131/commits/939ec48731b38a38e5afe8def97c95e22b58d619

frontend/catalyst/__init__.py
rniczh: Done https://github.com/PennyLaneAI/catalyst/pull/2131/commits/939ec48731b38a38e5afe8def97c95e22b58d619

paul0403:
> > @paul0403 Just realized the change of the name of slice to tuple caused test failures, that should also need the pennylane to change all the occurrence of `..._slice` to `..._tuple` in `pennylane/capture/base_interpreter.py`, `pennylane/control_flow/for_loop.py`, ... etc. as well. @paul0403 @JerryChen97 How do you guys think about this
> 
> I'll see how much work it involves on PennyLane side!

I see, don't worry about it if it involves API changes to the primitives then. 

If this indeed the case, @JerryChen97 make sure to run it through the core team.

frontend/catalyst/utils/patching.py
```diff
@@ -17,6 +17,37 @@
...
+        key: The key to access in the dictionary
+    """
+
+    def __init__(self, dictionary, key):
+        self.dictionary = dictionary
+        self.key = key
+
+    def __getattr__(self, name):
+        if name in ("dictionary", "key"):
+            return object.__getattribute__(self, name)
```
JerryChen97, line 37: Seems this branch was never used. Should we `# pragma: no cover` it since it looks useful? Or just delete this branch

frontend/catalyst/utils/patching.py
```diff
@@ -17,6 +17,37 @@
...
+        if name == "value":
+            return self.dictionary[self.key]
+        raise AttributeError(f"'{type(self).__name__}' object has no attribute '{name}'")
+
+    def __setattr__(self, name, value):
+        if name in ("dictionary", "key"):
+            object.__setattr__(self, name, value)
+        elif name == "value":
+            self.dictionary[self.key] = value
+        else:
```
JerryChen97:
```suggestion
        else:  # pragma: no cover
```

frontend/catalyst/jax_tracer.py
```diff
@@ -136,6 +136,17 @@ def mark_gradient_tracing(method: str):
...
 
+def _get_eqn_from_tracing_eqn(eqn_or_callable):
+    """Helper function to extract the actual equation from JAX 0.7's TracingEqn wrapper."""
+    if callable(eqn_or_callable):
+        actual_eqn = eqn_or_callable()
+        if actual_eqn is None:
+            raise RuntimeError("TracingEqn weakref was garbage collected")
+        return actual_eqn
+    else:
+        return eqn_or_callable
```
JerryChen97, line 157: Codecov says this default branch was never reached. So `eqn_or_callable` is actually all the time `callable`?

frontend/catalyst/jax_tracer.py
```diff
@@ -136,6 +136,17 @@ def mark_gradient_tracing(method: str):
         TRACING_GRADIENTS.pop()
 
 
+def _get_eqn_from_tracing_eqn(eqn_or_callable):
+    """Helper function to extract the actual equation from JAX 0.7's TracingEqn wrapper."""
+    if callable(eqn_or_callable):
+        actual_eqn = eqn_or_callable()
+        if actual_eqn is None:
```
JerryChen97:
```suggestion
        if actual_eqn is None:  # pragma: no cover
```

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -57,6 +59,63 @@
...
+
+    Args:
+        t: Either a slice object (returned as-is) or a tuple (start, stop, step)
+
+    Returns:
+        slice: A slice object
+    """
+    if isinstance(t, tuple) and len(t) == 3:
+        return slice(*t)
+    return t
```
JerryChen97:
```suggestion
    assert isinstance(t, tuple) and len(t) == 3, "Please only use _tuple_to_slice on a tuple of length 3!"
    return slice(*t)
```

rniczh: Done https://github.com/PennyLaneAI/catalyst/pull/2131/commits/68b85e1fad83fc84bb841930e0f0ee3b9e6ac7da

frontend/catalyst/jax_tracer.py
rniczh: In our test, yes, but jax doesn't made any guarantees.

frontend/catalyst/utils/patching.py
rniczh: We should keep it

frontend/catalyst/jax_tracer.py
rniczh: I just added no cover for this

JerryChen97:
Fantastic, now even Codecov passes! @rniczh

JerryChen97 (changes requested):
Hold on temporarily until the PennyLane PR is also ready-to-merge

PR #2135, opened 2025-10-21
Tidy strange code in jit.py

albi3ro:
**Context:**

Trying to understand how transforms get passed around in the catalyst frontend. This code was confusing, and I think it is dead code. So let's try getting rid of the dead logic.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/frontend/catalyst/jit.py
+++ b/frontend/catalyst/jit.py
@@ -743,9 +743,7 @@ class QJIT(CatalystCallable):
             params["_out_tree_expected"] = []
             params["_classical_return_indices"] = []
             params["_num_mcm_expected"] = []
-            default_pass_pipeline = self.compile_options.circuit_transform_pipeline
-            pass_pipeline = params.get("pass_pipeline", default_pass_pipeline)
-            params["pass_pipeline"] = pass_pipeline
+            params["pass_pipeline"] = self.compile_options.circuit_transform_pipeline
             params["debug_info"] = dbg
 
             return QFunc.__call__(


frontend/catalyst/jit.py
```diff
@@ -743,9 +743,7 @@ def closure(qnode, *args, **kwargs):
             params["_out_tree_expected"] = []
             params["_classical_return_indices"] = []
             params["_num_mcm_expected"] = []
-            default_pass_pipeline = self.compile_options.circuit_transform_pipeline
-            pass_pipeline = params.get("pass_pipeline", default_pass_pipeline)
-            params["pass_pipeline"] = pass_pipeline
+            params["pass_pipeline"] = self.compile_options.circuit_transform_pipeline
```
dime10, line 746: Is this worth trying out then? 
```suggestion
            default_pass_pipeline = self.compile_options.circuit_transform_pipeline
            pass_pipeline = kwargs.pop("pass_pipeline", default_pass_pipeline)
            params["pass_pipeline"] = pass_pipeline
```

albi3ro: So how would we hit that?

Also, seems like that might be a new feature we would have to document and test.

dime10: Didn't you say `PassPipelineWrapper` is generating the `pass_pipeline` kwarg? There is also evidence that we expect a `pass_pipeline` kwarg in qnodes [here](https://github.com/PennyLaneAI/catalyst/blob/bb0b7dfab437fb0f5a8db89e3100dae0953d8641/frontend/catalyst/qfunc.py#L286-L287) and [here](https://github.com/PennyLaneAI/catalyst/blob/bb0b7dfab437fb0f5a8db89e3100dae0953d8641/frontend/catalyst/autograph/ag_primitives.py#L586).

I'm not sure I see a new feature, just obviously buggy code that very likely doesn't match the intent of the author who wrote it. Maybe you're thinking of a new feature because this kwarg could be provided by users as well? But to me it looks like the kwarg is used an internal "secret" parameter rather than a user facing one.

PR #2138, opened 2025-10-21
Add split non commuting pass

rniczh:
**Context:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/frontend/catalyst/passes/split_non_commuting.py
+++ b/frontend/catalyst/passes/split_non_commuting.py
@@ -1,9 +1,5 @@
-from dataclasses import dataclass, field
-from itertools import chain
 from typing import Type, TypeVar
 
-import logging
-import numpy as np
 import pennylane as qml
 
 from catalyst.passes.xdsl_plugin import getXDSLPluginAbsolutePath
@@ -19,16 +15,6 @@ from xdsl.pattern_rewriter import PatternRewriter, RewritePattern
 from xdsl.rewriter import InsertPoint
 
 T = TypeVar("T")
-
-logger = logging.getLogger(__name__)
-logger.disabled = True
-
-if not logger.handlers:
-    handler = logging.StreamHandler()
-    handler.setLevel(logging.DEBUG)
-    formatter = logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
-    handler.setFormatter(formatter)
-    logger.addHandler(handler)
 
 
 def get_parent_of_type(op: Operation, kind: Type[T]) -> T | None:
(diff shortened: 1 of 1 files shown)


frontend/catalyst/passes/split_non_commuting.py
```diff
@@ -0,0 +1,605 @@
...
+
+logger = logging.getLogger(__name__)
+logger.disabled = True
+
+if not logger.handlers:
+    handler = logging.StreamHandler()
+    handler.setLevel(logging.DEBUG)
+    formatter = logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
+    handler.setFormatter(formatter)
+    logger.addHandler(handler)
```
mlxd: You should be able to use the PennyLane logging functionality that provides default config already.

I'd change the above to use something similar to https://github.com/PennyLaneAI/catalyst/blob/9376631b57b62d0153950142b8087baa807b6b9d/frontend/catalyst/device/qjit_device.py#L51 and https://github.com/PennyLaneAI/catalyst/blob/9376631b57b62d0153950142b8087baa807b6b9d/frontend/catalyst/device/qjit_device.py#L57 with usage matching https://github.com/PennyLaneAI/catalyst/blob/9376631b57b62d0153950142b8087baa807b6b9d/frontend/catalyst/device/qjit_device.py#L145

frontend/catalyst/passes/split_non_commuting.py
```diff
@@ -0,0 +1,605 @@
...
+                if "group" in op.attributes:
+                    group_attr = op.attributes["group"]
+                    group_id = group_attr.value.data
+
+                    # If it's not the target group, mark for removal
+                    if group_id != target_group:
+                        expval_ops_to_remove.append(op)
+
+        if not expval_ops_to_remove:
+            print(f"No expvals to remove for group {target_group}")
```
mlxd: Do we want to print here? Same q with all other print statements

frontend/catalyst/passes/split_non_commuting.py
```diff
@@ -0,0 +1,605 @@
...
+        return dup_func
+
+    def removeGroup(self, dup_func: func.FuncOp, target_group: int):
+        """Remove expvals from other groups and update return statement.
+
+        For the dup function of group i, this removes all expval operations that
+        belong to other groups, traces their results to the return statement,
+        and removes those return values.
+        """
+        print(f"removeGroup: keeping only group {target_group}")
```
mlxd: Do we want to print here?

frontend/catalyst/passes/split_non_commuting.py
rniczh: Thanks for the suggestion! But I just removed all the print and debug log from this xdsl pass

frontend/catalyst/passes/split_non_commuting.py
rniczh: No, those are debug purpose, I already removed them

frontend/catalyst/passes/split_non_commuting.py
rniczh: No, those are debug purpose, I already removed them

PR #2149, opened 2025-10-24
Use `pass_name` from pennylane transform

albi3ro:
**Context:**

Trying to improve our integration between pennylane transforms and catalyst passes.

See #2206  for where we are able to update all built-in catalyst passes to be transforms instead. That PR helps validate we can make this change with no breaking behaviour.

Going forward, we can setup a plan to get rid of the catalyst-specific pass handling infrastructure.

Also note that this PR targets the PennyLane branch, so we can confirm we won't need anymore changes there to get catalyst working correctly before approving and merging.

Uses the `transform.pass_name` property as the higher priority source of the `pass_name`.

Updates the `pass_pipeline` from the `quantum_kernel_p` primitive to be a tuple of `TransformContainer` objects instead of `Pass`, but leaves in the logic to handle `tuple[Pass,...]` for safer backwards compatibility.

Also makes using transforms with pass names by splitting a `QNode`'s transform program into "things at the start of the program that are tape transforms" and "things at the end of the program that are MLIR passes".  This will allow us to get rid of `PassPipelineWrapper` use for both systems and soften the change from old frontend to new frontend.

**Benefits:**

Unified handling of transforms across pennylane and catalyst. 

**Possible Drawbacks:**
 

**Related GitHub Issues:**

Depends on https://github.com/PennyLaneAI/pennylane/pull/8539 [sc-103775]

### DIFF
--- a/frontend/test/lit/test_decomposition.py
+++ b/frontend/test/lit/test_decomposition.py
@@ -46,6 +46,7 @@ def skip_if_pauli_rot_issue(test_func):
             error_msg = str(e)
             if (
                 "Unsupported type annotation None for parameter pauli_word" in error_msg
+                or "Unsupported type annotation <class 'str'> for parameter pauli_word" in error_msg
                 or "index is out of bounds for axis" in error_msg
             ):
                 print(f"# SKIPPED {test_func.__name__}: PauliRot type annotation issue")


frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -279,6 +276,47 @@ def register_transform(pl_transform, pass_name, decomposition):
     transforms_to_passes[pl_transform] = (pass_name, decomposition)
 
 
+def _handle_decompose_transform(self, inner_jaxpr, consts, non_const_args, tkwargs):
```
albi3ro, line 349: Just extracting this to a helper function to make the function a bit tidier.

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -304,44 +342,11 @@ def handle_transform(
...
-        #     return gds_interpreter.eval(inner_jaxpr, consts, *args)
-
-        # final_jaxpr = jax.make_jaxpr(gds_wrapper)(*args)
-        # return self.eval(final_jaxpr.jaxpr, consts, *non_const_args)
-        return next_eval.eval(inner_jaxpr, consts, *non_const_args)
-
-    catalyst_pass_name = transforms_to_passes.get(transform, (None,))[0]
+    catalyst_pass_name = transform.pass_name
+    if catalyst_pass_name is None:
+        catalyst_pass_name = transforms_to_passes.get(transform, (None,))[0]
```
paul0403, line 420: With this change do we still need the `transforms_to_passes` dictionary at all?

frontend/catalyst/jax_primitives_utils.py
```diff
@@ -366,11 +376,16 @@ def transform_named_sequence_lowering(jax_ctx: mlir.LoweringRuleContext, pipelin
         with ir.InsertionPoint(bb_named_sequence):
             target = bb_named_sequence.arguments[0]
             for _pass in pipeline:
-                options = _pass.get_options()
+                if isinstance(_pass, qml.transforms.core.TransformContainer):
```
paul0403, line 377: Opinions on separating based on if capture is turned on or not?

```suggestion
                if qml.capture.enabled():
                   assert isinstance(_pass, qml.transforms.core.TransformContainer):
```

frontend/catalyst/jax_primitives_utils.py
```diff
@@ -320,6 +320,16 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         self.ctx.module_context = self.old_module_context
 
 
+def _lowered_options(args, kwargs):
+    lowered_options = {}
+    for arg in args:
+        lowered_options[str(arg)] = get_mlir_attribute_from_pyval(True)
```
paul0403, line 324: It warms my heart to see this util being used :heart:

frontend/test/lit/test_peephole_optimizations.py
```diff
@@ -87,6 +87,42 @@ def test_pipeline_lowering_workflow(x):
...
+        qml.RX(x, wires=[0])
+        qml.Hadamard(wires=[1])
+        qml.Hadamard(wires=[1])
+        return qml.expval(qml.PauliY(wires=0))
+
+    # CHECK: pipeline=(<cancel_inverses((), {})>, <merge_rotations((), {})>)
+    print_jaxpr(test_pipeline_lowering_workflow, 1.2)
+
+    # CHECK: transform.named_sequence @__transform_main
+    # CHECK-NEXT: {{%.+}} = transform.apply_registered_pass "remove-chained-self-inverse" to {{%.+}}
```
paul0403:
```suggestion
    # CHECK-NEXT: {{%.+}} = transform.apply_registered_pass "cancel-inverses" to {{%.+}}
```
we recently renamed this

frontend/test/pytest/from_plxpr/test_capture_integration.py
```diff
@@ -1048,7 +1048,7 @@ def circuit(x: float):
         assert jnp.allclose(circuit(0.1), capture_result)
 
     @pytest.mark.usefixtures("use_capture")
-    def test_pass_with_options(self, backend):
+    def test_pass_with_options_patch(self, backend):
```
paul0403, line 1051: We can finally remove `register_transform` now? :star_struck:

frontend/test/pytest/test_peephole_optimizations.py
```diff
@@ -204,6 +211,26 @@ def test_chained_apply_passes_workflow(x: float):
...
+    @qml.qnode(qml.device("lightning.qubit", wires=2))
+    def test_chained_apply_passes_workflow(x: float):
+        qml.Hadamard(wires=[1])
+        qml.RX(x, wires=[0])
+        qml.RX(-x, wires=[0])
+        qml.Hadamard(wires=[1])
+        return qml.expval(qml.PauliY(wires=0))
+
+    assert "cancel-inverses" in test_chained_apply_passes_workflow.mlir
+    assert "merge-rotations" in test_chained_apply_passes_workflow.mlir
```
paul0403: I feel like this test duplicates the lit test you also added

frontend/test/pytest/test_transform_pass_name.py
```diff
@@ -0,0 +1,80 @@
+# Copyright 2023 Xanadu Quantum Technologies Inc.
```
paul0403:
```suggestion
# Copyright 2025 Xanadu Quantum Technologies Inc.
```

frontend/test/pytest/test_transform_pass_name.py
```diff
@@ -0,0 +1,80 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Testing use of transforms with pass name integrating with qnodes."""
+from functools import partial
+
+import pennylane as qml
+import pytest
+
+
+def test_pass_with_options(backend):
```
paul0403, line 21: This one also duplicates the test you added to `test_capture_integration.py`. 

It probably makes the most sense to just move the old one there that uses `register_transform()` here into this new file?

frontend/test/pytest/test_transform_pass_name.py
```diff
@@ -0,0 +1,80 @@
...
+    @tape_only_cancel_inverses
+    @qml.qnode(qml.device(backend, wires=1))
+    def c():
+        qml.X(0)
+        qml.X(0)
+        return qml.state()
+
+    # check inverses canceled
+    assert 'quantum.custom "PauliX"()' not in c.mlir
+    assert 'transform.apply_registered_pass "my-pass"' in c.mlir
```
paul0403:
```suggestion
    # check inverses canceled
    c_mlir = c.mlir
    assert 'quantum.custom "PauliX"()' not in c_mlir
    assert 'transform.apply_registered_pass "my-pass"' in c_mlir
```
`QJIT.mlir` actually triggers recompilation. We can save some testing time this way.

frontend/catalyst/from_plxpr/from_plxpr.py
albi3ro: Given some things in pennylane are currently using it, I decided to leave it be for the moment for backwards compatibility. Once these go in, we get rid of those patches, then we get rid of `transforms_to_passes`.

frontend/catalyst/jax_primitives_utils.py
albi3ro: I'm actually also creating the `TransformContainer`'s in `qfunc.py`.  This allows us to just split the qnode's transform program in two and pass the mlir passes part of it right to `pass_pipeline`.  I'm not actually sure when `Pass` objects would still be getting passed on, but I'm just leaving both cases here to be safe.

frontend/catalyst/jax_primitives_utils.py
albi3ro: Mostly just copy-pasteing from the pass code here.

frontend/test/lit/test_peephole_optimizations.py
albi3ro: Yeah... that caused me a few headaches.

frontend/test/pytest/from_plxpr/test_capture_integration.py
albi3ro: Yep.  But I wanted to get this in, then update the places that use `register_transform`, then come back and get rid of it. Maybe being a bit overly cautious, but keeps things going smoothly for others.

frontend/test/pytest/test_transform_pass_name.py
albi3ro: That one is specific to program capture.  This one is for non-program capture old frontend.

paul0403: Makes sense :+1:

frontend/test/pytest/test_transform_pass_name.py
albi3ro: I feel like that is rather misleading that a property triggers an action. But 🤷

paul0403: Yeah, I was also surprised when I first learned this

frontend/test/pytest/test_transform_pass_name.py
```diff
@@ -0,0 +1,81 @@
...
+
+    @qml.transform
+    def tape_transform(tape):
+        return (tape,), lambda x: x[0]
+
+    @qml.qjit
+    @tape_transform
+    @my_pass
+    @qml.qnode(qml.device(backend, wires=1))
+    def f(x):
```
albi3ro:
```suggestion
    def f(x): # pylint: disable=unused-argument
```

frontend/test/pytest/test_peephole_optimizations.py
albi3ro: So actually decides  what should be a pytest and what should be a lit test?

albi3ro: This is basically a copy paste of the test above, but with the catalyst namespace version replaced with the pl namespace version.

paul0403: If the only thing being checked is the compiled IR (aka not checking end-to-end execution results), then it should be a lit test.

So you're right, there's many pytests that should've been lit tests... part of the reason is for some reason we can't get codecov to see lit tests :melting_face: 

But in this particular case I think your other pytests already cover codecov

frontend/catalyst/jax_primitives_utils.py
```diff
@@ -320,6 +320,16 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+    for option, value in kwargs.items():
+        mlir_option = str(option).replace("_", "-")
+        lowered_options[mlir_option] = get_mlir_attribute_from_pyval(value)
+    return lowered_options
```
mudit2812, line 328: I would like this functionality to be extended in the future to also allow a way to lower python objects that do not have a straightforward way of lowering to MLIR. This is necessary for some of the xDSL passes to work. It would be very cumbersome for the default pass decorators to use arguments that are MLIR compatible for complex passes.

paul0403 (approved):

frontend/catalyst/jax_primitives_utils.py
paul0403: But this has to be per-pass right? We don't know what python objects a xdsl pass might take in as options in general.

mudit2812: Yep, passes would need to define their own strategies for serializing/deserializing, but the compilation pipeline will have to be updated to integrate with these strategies.

paul0403: But I don't know what concrete actions we can do now without knowing what the specific passes need :thinking:

albi3ro: Put something in a backlog TODO for now?

paul0403: I think that's good

PR #2153, opened 2025-10-28
Document linker warning on macOS

dime10:
```
ld: warning: building for macOS-15.0, but linking with dylib '@rpath/libcustom_calls.so' which was built for newer version 15.7
```

### DIFF
--- a/doc/dev/installation.rst
+++ b/doc/dev/installation.rst
@@ -308,7 +308,7 @@ The following target runs all available test suites with the default execution d
 
   make test
 
-You can also test each module separately by using running the ``test-frontend``,
+You can also test each module separately by running the ``test-frontend``,
 ``test-dialects``, and ``test-runtime`` targets instead.
 Alternately, the ``test-frontend`` target can be broken up into two subsets,
 ``make pytest`` and ``make lit``, where the pytest suite
@@ -445,14 +445,14 @@ Known Issues
 
       In some mac setups, it's possible that certain Catalyst libraries are compiled with a
       different OSX deployment version than what Catalyst uses when compiling a user program.
-      You may see the following *warning* in that case when using Catalyst:
+      In this case, you may see a *warning* like the following when using Catalyst:
 
       .. code-block:: console
 
         ld: warning: building for macOS-15.0, but linking with dylib '@rpath/libcustom_calls.so' which was built for newer version 15.7
 
-      While this warning does not prevent using Catalyst, if you would like to eliminate it set the
-      following variable (adjust value as need to the warning you see):
+      While this warning does not prevent using Catalyst, if you would like to eliminate it, set the
+      following environment variable (adjust the value as needed given the warning you see):
 
       .. code-block:: console
 


doc/dev/installation.rst
```diff
@@ -436,6 +439,30 @@ Known Issues
       If not, PyTest might try to use the default Python binary: ``/usr/bin/python3``.
       (See user's report `here <https://github.com/PennyLaneAI/catalyst/issues/377>`_)
 
+      .. raw:: html
+
+        <hr>
+
+      In some mac setups, it's possible that certain Catalyst libraries are compiled with a
+      different OSX deployment version than what Catalyst uses when compiling a user program.
+      You may see the following *warning* in that case when using Catalyst:
```
joeycarter: [Optional] Rewording to make it more general in case users are using different macOS versions:

```suggestion
      In this case, you may see a *warning* like the following when using Catalyst:
```

doc/dev/installation.rst
```diff
@@ -436,6 +439,30 @@ Known Issues
...
+
+      .. code-block:: console
+
+        ld: warning: building for macOS-15.0, but linking with dylib '@rpath/libcustom_calls.so' which was built for newer version 15.7
+
+      While this warning does not prevent using Catalyst, if you would like to eliminate it set the
+      following variable (adjust value as need to the warning you see):
```
joeycarter:
```suggestion
      While this warning does not prevent using Catalyst, if you would like to eliminate it, set the
      following environment variable (adjust the value as needed given the warning you see):
```

doc/dev/installation.rst
```diff
@@ -309,8 +309,11 @@ The following target runs all available test suites with the default execution d
   make test
 
 You can also test each module separately by using running the ``test-frontend``,
```
joeycarter: Just noticed this typo:

```suggestion
You can also test each module separately by running the ``test-frontend``,
```

joeycarter (approved):
Looks good, thanks! Just a few minor editorial suggestions and fixes.

doc/dev/installation.rst
```diff
@@ -309,8 +309,11 @@ The following target runs all available test suites with the default execution d
...
-``test-dialects``, and ``test-runtime`` targets instead. Jupyter Notebook demos are also testable
-via ``test-demos``.
+``test-dialects``, and ``test-runtime`` targets instead.
+Alternately, ``make pytest`` (a subset of ``test-frontend``) and ``make lit`` (a combination of
+subsets from ``test-frontend`` and ``test-dialects``) are also available, where the pytest suite
```
paul0403: I'm pretty sure `make lit` just runs the frontend lit tests, not the dialect ones.

paul0403 (approved):

doc/dev/installation.rst
dime10:
```suggestion
Alternately, the ``test-frontend`` target can be broken up into two subsets,
``make pytest`` and ``make lit``, where the pytest suite
```

PR #2155, opened 2025-10-28
Fix two tests that are still using the deprecated `is_hermitian` property

JerryChen97:
**Context:**
A recent PennyLane [commit](https://github.com/PennyLaneAI/pennylane/pull/8494) deprecated `is_hermitian` and replaced it with a new `is_verified_hermitian`. Therefore some tests of Catalyst need adjustment as well 

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


JerryChen97:
Need to update dep version tmrw after the LLL fails show this bug up

PR #2160, opened 2025-10-30
Enable decomposition of MultiRZ at MLIR

maliasadi:
**Context:**
This PR refactors the implementation of `SignatureAnalyzer` in the decompose-lowering pass extending the analyser and the pass to support special operator definitions (e.g., MultiRZ) along with CustomOps as new patterns.

- Add a `BaseSignatureAnalyzer` class that servers a base analyzer in decompose-lowering
- Add MultiRZ decomposition patterns to the lowering pass
- Fix the issue with trying to capture and compile unutilized abstracted Adjoint and Controlled decomposition rules from the graph. (We don't need these rules from the graph as we handle these symbolic ops using Catalyst AdjointOp and ControlledOp). Avoiding this saves compilation time and failures for templated rules. 
- Include MLIR and Python tests.

**Related Issues:**
[sc-102152]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -17,6 +17,9 @@
   :func:`~.qjit` with program capture enabled.
   [(#2154)](https://github.com/PennyLaneAI/catalyst/pull/2154)
 
+* Pytree inputs can now be used when program capture is enabled.
+  [(#2165)](https://github.com/PennyLaneAI/catalyst/pull/2165)
+
 <h3>Breaking changes 💔</h3>
 
 <h3>Deprecations 👋</h3>
--- a/frontend/catalyst/from_plxpr/from_plxpr.py
+++ b/frontend/catalyst/from_plxpr/from_plxpr.py
@@ -399,7 +399,7 @@ def trace_from_pennylane(
             fn.static_argnums = static_argnums
 
         plxpr, out_type, out_treedef = make_jaxpr2(fn, **make_jaxpr_kwargs)(*args, **kwargs)
-        jaxpr = from_plxpr(plxpr)(*dynamic_args, **kwargs)
+        jaxpr = from_plxpr(plxpr)(*plxpr.in_avals)
 
     return jaxpr, out_type, out_treedef, sig
 
--- a/frontend/test/pytest/test_pytree_args.py
+++ b/frontend/test/pytest/test_pytree_args.py
@@ -22,13 +22,15 @@ import numpy as np
 import pennylane as qml
 import pytest
 from jax._src.tree_util import tree_flatten
+from pennylane import adjoint, cond, for_loop, grad, qjit
 
-from catalyst import adjoint, cond, for_loop, grad, measure, qjit
+from catalyst import measure
 
 
 class TestPyTreesReturnValues:
     """Test QJIT workflows with different return value data-types."""
 
+    @pytest.mark.usefixtures("use_both_frontend")
     def test_return_value_float(self, backend):
         """Test constant."""
 
@@ -41,9 +43,12 @@ class TestPyTreesReturnValues:
         jitted_fn = qjit(circuit1)
 
         params = [0.4, 0.8]
-        expected = 0.64170937
+        expected = jnp.cos(params[0]) * jnp.cos(params[1])
         result = jitted_fn(params)
         assert jnp.allclose(result, expected)
+
+    def test_return_value_mcm(self, backend):
+        """Test that a qnode can return a scalar mcm."""
 
         @qml.qnode(qml.device(backend, wires=2))
         def circuit2():
@@ -166,6 +171,7 @@ class TestPyTreesReturnValues:
         assert result[0] == 4.0
         assert result[1] == 6.0
 
+    @pytest.mark.usefixtures("use_both_frontend")
     def test_return_value_hybrid(self, backend):
         """Test tuples."""
 
@@ -238,6 +244,9 @@ class TestPyTreesReturnValues:
     @pytest.mark.parametrize("mcm_method", ["single-branch-statistics", "one-shot"])
     def test_return_value_dict(self, backend, tol_stochastic, mcm_method):
         """Test dictionaries."""
+
+        if mcm_method == "one-shot" and qml.capture.enabled():
+            pytest.xfail()
 
         @qml.qnode(qml.device(backend, wires=2))
         def circuit1(params):
@@ -438,6 +447,7 @@ class TestPyTreesFuncArgs:
         }
         result = jitted_fn(params)
 
+    @pytest.mark.usefixtures("use_both_frontend")
     def test_args_workflow(self, backend):
         """Test arguments with workflows."""
 
@@ -527,6 +537,7 @@ class TestPyTreesFuncArgs:
         assert np.allclose(result_flatten, result_flatten_expected)
         assert tree == tree_expected
 
+    @pytest.mark.usefixtures("use_both_frontend")
     @pytest.mark.parametrize("inp", [(np.array([0.2, 0.5])), (jnp.array([0.2, 0.5]))])
     def test_args_control_flow(self, backend, inp):
         """Test arguments with control-flows operations."""
@@ -574,6 +585,7 @@ class TestPyTreesFuncArgs:
         result = circuit({"wire": 1})
         assert jnp.allclose(result, True)
 
+    @pytest.mark.usefixtures("use_both_frontend")
     def test_dev_wires_have_pytree(self, backend):
         """Device wires are pytree-compatible."""
 
@@ -592,6 +604,7 @@ class TestPyTreesFuncArgs:
         test_function()
 
 
+@pytest.mark.usefixtures("use_both_frontend")
 class TestAuxiliaryData:
     """Test PyTrees with Auxiliary data."""
 


mlir/test/Quantum/DecomposeLoweringTest.mlir
```diff
@@ -554,3 +554,58 @@ module @mcm_example {
...
+    // CHECK: [[CST_RZ:%.+]] = arith.constant 5.000000e-01 : f64
+    // CHECK: [[CST_PI2:%.+]] = arith.constant 1.5707963267948966 : f64
+    // CHECK: [[CST_PI:%.+]] = arith.constant 3.1415926535897931 : f64
+    // CHECK: [[REG:%.+]] = quantum.alloc( 2) : !quantum.reg
+
+    // CHECK: [[QUBIT1:%.+]] = quantum.custom "RZ"([[CST_RZ]]) {{%.+}} : !quantum.bit
+    // CHECK-NOT: quantum.multirz
+    %cst = stablehlo.constant dense<5.000000e-01> : tensor<f64>
+    %extracted_2 = tensor.extract %cst[] : tensor<f64>
+    %out_qubits = quantum.multirz(%extracted_2) %1 : !quantum.bit
```
maliasadi, line 574: The new quantum-opt tool should replace `multirz`! I don't know what's going on CIs :thinking:

maliasadi: Thanks @rniczh! I had to recompile my LLVM/MLIR without ccache to be able to reproduce the issue.

mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
```diff
@@ -14,411 +14,117 @@
...
+            rewriter.replaceOp(op, callOp->getResults());
         }
 
-        return QubitIndex();
+        return success();
     }
 };
 
-struct DecomposeLoweringRewritePattern : public OpRewritePattern<CustomOp> {
+struct DLMultiRZOpPattern : public OpRewritePattern<MultiRZOp> {
```
rniczh, line 94: I think these two patterns are almost do the same thing. Do you think it's worth to unify them with template? And specialize the differences
```cpp
  auto numQubits = op.getInQubits().size();
  auto MRZNameWithQubits = gateName + "_" + std::to_string(numQubits);

  auto it = decompositionRegistry.find(MRZNameWithQubits.str());
  if (it == decompositionRegistry.end()) {
      return failure();
  }
```

It could be look like this
```cpp
template<typename T> 
struct QuantumOpPattern  : public OpRewritePattern<T>
...
```  

And add to the patterns with:
```cpp
patterns.add<QuantmOpPattern<CustomOp>>(patterns.getContext(), decompositionRegistry, targetGateSet);
patterns.add<QuantumOpPattern<MultiRZOp>>(patterns.getContext(), decompositionRegistry, targetGateSet);
```

rniczh (commented):
Thanks! Just one comment from my side

mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
maliasadi: The only reason I slightly wasn't in favour of using templates here is that using a base class with inheritance provides greater flexibility to support future, specialized gate operators that require a custom assembly format. For instance, to support patterns like PCPhase or QuantumUnitary in the future, we would need to adapt the definition of `signature`. While templates seem a reasonable solution for the current scope of work, I can update the code as you suggested. Lmk what the common practice is for this kind of extensibility.

rniczh: No worries, I think it's good for now 💯 , but we might need to check the thing again once we have to involve more and more operations to decompose

frontend/test/pytest/from_plxpr/test_from_plxpr.py
```diff
@@ -1051,6 +1051,45 @@ def circuit():
         expected = np.array([1, 0, 0, 1]) / np.sqrt(2)
         assert qml.math.allclose(circuit(), expected)
 
+        expected_resources = {"CZ": 1, "Hadamard": 3}
```
maliasadi, line 1054: I have a follow-up PR extending our E2E decomp tests with `qml.specs`.

mlir/lib/Quantum/Transforms/DecomposeLoweringImpl.hpp
```diff
@@ -0,0 +1,424 @@
...
+  public:
+    QubitIndex() : index(std::monostate()), qreg(nullptr) {}
+    QubitIndex(Value val, Value qreg) : index(val), qreg(qreg) {}
+    QubitIndex(IntegerAttr attr, Value qreg) : index(attr), qreg(qreg) {}
+
+    bool isValue() const { return std::holds_alternative<Value>(index); }
+    bool isAttr() const { return std::holds_alternative<IntegerAttr>(index); }
+    operator bool() const { return isValue() || isAttr(); }
+    Value getReg() const { return qreg; }
+    Value getValue() const { return isValue() ? std::get<Value>(index) : nullptr; }
```
multiphaseCFD, line 70: Is `Value` a pointer type or value type or reference?   🤔

mlir/lib/Quantum/Transforms/DecomposeLoweringImpl.hpp
```diff
@@ -0,0 +1,424 @@
...
+        llvm::SmallVector<QubitIndex> outCtrlQubitIndices;
+    } signature;
+
+    BaseSignatureAnalyzer(mlir::Operation *op, mlir::ValueRange params, mlir::ValueRange inQubits,
+                          mlir::ValueRange inCtrlQubits, mlir::ValueRange inCtrlValues,
+                          mlir::ValueRange outQubits, mlir::ValueRange outCtrlQubits,
+                          bool enableQregMode)
+        : paramsStorage(params.begin(), params.end()),
+          signature(Signature{.params = mlir::ValueRange(paramsStorage),
+                              .inQubits = inQubits,
```
multiphaseCFD, line 106: more clear/clean to use designated initializers.

mlir/lib/Quantum/Transforms/DecomposeLoweringImpl.hpp
```diff
@@ -0,0 +1,424 @@
...
+                return;
+            }
+            signature.inWireIndices.emplace_back(index);
+        }
+
+        // input ctrl wire indices
+        for (mlir::Value ctrlQubit : signature.inCtrlQubits) {
+            const QubitIndex index = getExtractIndex(ctrlQubit);
+            if (!index) {
+                op->emitError("Cannot get index for ctrl qubit");
```
multiphaseCFD: Just wondering what are the scenarios this line would be hit

mlir/lib/Quantum/Transforms/DecomposeLoweringImpl.hpp
```diff
@@ -0,0 +1,424 @@
...
+    operator bool() const { return isValue() || isAttr(); }
+    Value getReg() const { return qreg; }
+    Value getValue() const { return isValue() ? std::get<Value>(index) : nullptr; }
+    IntegerAttr getAttr() const { return isAttr() ? std::get<IntegerAttr>(index) : nullptr; }
+};
+
+// The goal of this class is to analyze the signature of a custom operation to get the enough
+// information to prepare the call operands and results for replacing the op to calling the
+// decomposition function.
+class BaseSignatureAnalyzer {
```
multiphaseCFD, line 77: Nice to create a Base class for the analyzer. There could be a few exceptions to be taken care of in the future.

mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
```diff
@@ -14,411 +14,117 @@
...
   private:
     const llvm::StringMap<func::FuncOp> &decompositionRegistry;
```
multiphaseCFD, line 96: I see decomposition funcs are registered

mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
```diff
@@ -14,411 +14,117 @@
...
+
+        auto it = decompositionRegistry.find(MRZNameWithQubits.str());
         if (it == decompositionRegistry.end()) {
             return failure();
         }
-        func::FuncOp decompFunc = it->second;
 
-        // Here is the assumption that the decomposition function must have at least one input and
-        // one result
+        func::FuncOp decompFunc = it->second;
```
multiphaseCFD, line 125: Just wondering: how the `MultiRZ` gate is decomposed? Is it a direct decomposition from the a MultRZ op matrix or apply the decomposition rule defined in the `qml.MultiRZ` class first and then decompose those results?

multiphaseCFD (approved):
Nice work @maliasadi .  Happy to approve! Also left a few questions for discussion.

mlir/lib/Quantum/Transforms/DecomposeLoweringImpl.hpp
maliasadi: It's a value type not a ptr or reference. It behaves like a lightweight wrapper around the actual value data.

mlir/lib/Quantum/Transforms/DecomposeLoweringImpl.hpp
maliasadi: Thanks! Fixed [0a5ca61](https://github.com/PennyLaneAI/catalyst/pull/2160/commits/0a5ca61cf2c40950462989683874b93d33e440f6)

mlir/lib/Quantum/Transforms/DecomposeLoweringImpl.hpp
maliasadi: For cases that we use qubit values instead of qubit registers. For those cases, we have a different logic which doesn't require `initializeQregMode`. This is a check to ensure that we don't init qregs for qubit values. It's associated with `is_qreg=False` in the `decomposition_rule` decorator.

mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
maliasadi: It uses a compiled version of the structured rule defined [here](https://github.com/PennyLaneAI/pennylane/blob/d6a87834c1bae7dd71b87330e28a8fb2c7eabe12/pennylane/ops/qubit/parametric_ops_multi_qubit.py#L228) to rewrite `quantum.multirz` ops with this rule at MLIR.

PR #2161, opened 2025-10-30
AutoGraph in the graph decomposition framework  

maliasadi:
**Context:**
 The new graph-based decomposition framework has Autograph feature parity with PennyLane
  when capture enabled. When compiling with `qml.qjit(autograph=True)`, the decomposition rules
  returned by the graph-based framework are now correctly compiled using Autograph.
  This ensures compatibility and deeper optimization for dynamically generated rules.

- Enable and test AutoGraph in the new graph decomposition framework

**Benefits:**

**Possible Drawbacks:**
- compiling with `qml.qjit(autograph=True)`, will compile the rules with autograph. This is current set globally and not per decomposition rule. Discussed offline -- we need to implement a toggle that works per subroutines or decomposition_rules in PL and as a follow-up update the Catalyst behaviour. re:  sc-104028

**Related GitHub Issues:**
[sc-100295]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -150,6 +150,10 @@
   for the qec dialect.
   [(#2189)](https://github.com/PennyLaneAI/catalyst/pull/2189)
 
+  * Added a canonicalization pattern for `qec.ppr` to remove any PPRs consisting only
+  of identities.
+  [(#2192)](https://github.com/PennyLaneAI/catalyst/pull/2192)
+
 <h3>Documentation 📝</h3>
 
 * A typo in the code example for :func:`~.passes.ppr_to_ppm` has been corrected.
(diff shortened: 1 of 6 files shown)


frontend/test/lit/test_decomposition.py
```diff
@@ -1199,3 +1200,34 @@ def circuit_26(x: float, y: float, z: float):
...
+    qml.decomposition.enable_graph()
+
+    @qml.qjit(autograph=True)
+    @partial(
+        qml.transforms.decompose, gate_set={"Rot"}
+    )
+    @qml.qnode(qml.device("lightning.qubit", wires= 1))
+    def circuit_27(n: int, x: float):
+        # CHECK: %2:2 = scf.for %arg2 = %c0 to %1 step %c1 iter_args(%arg3 = %arg1, %arg4 = %0) -> (tensor<f64>, !quantum.reg)
+        @qml.for_loop(0, n, 1)
```
paul0403: It works out of the box directly? :star_struck:

paul0403 (approved):

frontend/test/lit/test_decomposition.py
maliasadi: Oh yeah. re: #2160

PR #2162, opened 2025-10-30
Add canonicalize to extract

rniczh:
**Context:**

Fix canonicalization of eliminating redundant `quantum.insert` and `quantum.extract` pairs. When extracting a qubit immediately after inserting it at the same index, the operations can be cancelled out while properly updating remaining uses of the register. For an example:
```mlir
// Before canonicalization
%1 = quantum.insert %0[%idx], %qubit1 : !quantum.reg, !quantum.bit
%2 = quantum.extract %1[%idx] : !quantum.reg -> !quantum.bit
...
%3 = quantum.insert %1[%i0], %qubit2 : !quantum.reg, !quantum.bit
%4 = quantum.extract %1[%i1] : !quantum.reg -> !quantum.bit
// ... use %1
// ... use %4

// After canonicalization
// %2 directly uses %qubit1
// %3 and %4 updated to use %0 instead of %1
%3 = quantum.insert %0[%i0], %qubit2 : !quantum.reg, !quantum.bit
%4 = quantum.extract %0[%i1] : !quantum.reg -> !quantum.bit
// ... use %qubit1
// ... use %4
```

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-102833]

### DIFF
--- a/.dep-versions
+++ b/.dep-versions
@@ -10,7 +10,7 @@ enzyme=v0.0.186
 
 # For a custom PL version, update the package version here and at
 # 'doc/requirements.txt'
-pennylane=0.44.0.dev17
+pennylane=0.44.0.dev20
 
 # For a custom LQ/LK version, update the package version here and at
 # 'doc/requirements.txt'
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -20,6 +20,9 @@
 * Pytree inputs can now be used when program capture is enabled.
   [(#2165)](https://github.com/PennyLaneAI/catalyst/pull/2165)
 
+* `qml.grad` and `qml.jacobian` can now be used with `qjit` when program capture is enabled.
+  [(#2078)](https://github.com/PennyLaneAI/catalyst/pull/2078)
+
 <h3>Breaking changes 💔</h3>
 
 <h3>Deprecations 👋</h3>
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -33,4 +33,4 @@ lxml_html_clean
 --extra-index-url https://test.pypi.org/simple/
 pennylane-lightning-kokkos==0.43.0
 pennylane-lightning==0.43.0
-pennylane==0.44.0.dev17
+pennylane==0.44.0.dev20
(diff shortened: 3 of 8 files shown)


mlir/lib/Quantum/IR/QuantumOps.cpp
```diff
@@ -155,9 +155,11 @@ LogicalResult ExtractOp::canonicalize(ExtractOp extract, mlir::PatternRewriter &
...
         bool staticallyEqual = bothStatic && extract.getIdxAttrAttr() == insert.getIdxAttrAttr();
         bool dynamicallyEqual = bothDynamic && extract.getIdx() == insert.getIdx();
-        // if other users of insert are also `insert`, we are good to go
         bool valid = llvm::all_of(insert.getResult().getUsers(), [&](Operation *op) {
-            return isa<InsertOp>(op) || op == extract.getOperation();
+            if (isa<InsertOp>(op) || isa<ExtractOp>(op)) {
+                return true;
+            }
+            return false;
         });
```
dime10: Do we care what the other users of insert are at all? 🤔

rniczh: No, considering the following IR, no matter what the insert is (child1), we can update the %1 to %0 in child1 when we removing the root and child0.
```
%1 <- insert %0 %qubit1 [index] (root)
%2 <- extract %1 [index] (child0: can be cancelled with %1) 
...
%3 <- insert %1 %qubit2 [i0] (child1)
%4 <- extract %1 [i1] (child2)
```

dime10: Agreed 👍 

So this check can be removed entirely right?

rniczh: Yes!

mlir/lib/Quantum/IR/QuantumOps.cpp
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <llvm/ADT/STLExtras.h>
```
dime10: Why the header for the added lines? Needs to go into the third-party headers though, not the stdlib ones.

rniczh: It's something added by clangd, but redudant, I removed it

PR #2164, opened 2025-10-30
Issue #1621: Fix DataView iterator segfault with axis of size 0

### LINKED ISSUE #1621: Internal issue with `DataView` iterator increment operator when size along at least one axis is zero
We discovered this issue with the `DataView` iterator increment operator when putting together #1598:

https://github.com/PennyLaneAI/catalyst/blob/46d839e57b09b125a7baf25a02b19d7268901426/runtime/include/DataView.hpp#L56-L72

See the comment [here](https://github.com/PennyLaneAI/catalyst/pull/1598#issuecomment-2779178046) for more details.

In short, when the size along at least one axis of the data view is zero, there's an issue with the multidimensional indexing computation that results in the `loc` variable increasing indefinitely until the program segfaults. Specifically, the issue is that `view.sizes[idx] == 0` with `idx == 1` on the first iteration, resulting in an underflow error when subtracting 1 from it (since `view.sizes` is of type `size_t`). We added an assert in #1598 to ensure that the iteration terminates before an underflow error occurs, but the original issue with the multidimensional indexing should be fixed properly.

---

As far as we are aware, there are no user-facing or internal functions that will trigger this error. To reproduce it, you can modify the `NullQubit::Sample()` function to the following

```cpp
void Sample(DataView<double, 2> &samples, size_t)
{
    std::fill(samples.begin(), samples.end(), 0.0);
}
```

and execute the following workload:

```python
import pennylane as qml

@qml.qjit
@qml.qnode(qml.device("null.qubit", wires=0, shots=10))
def circuit():
    return qml.sample()
```

kipawaa:
**Context:**
The `DataView` iterators suffer an underflow of unsigned types that can lead to a segfault when presented an axis of size 0, as discussed in #1621 and [here](https://github.com/PennyLaneAI/catalyst/pull/1598#issuecomment-2779178046).

This change removes the `var - 1` type arithmetic that lead to the underflows, which in turn prevents the segfaults, by adjusting the indexing done in the for loop.

**Benefits:**
No more underflow errors and segfaults, accurate `view.size()` and `std::distance(view.begin(), view.end())` results in the case of an axis of size 0.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
Closes #1621 

[sc-88502]

### DIFF
--- a/runtime/tests/Test_DataView.cpp
+++ b/runtime/tests/Test_DataView.cpp
@@ -1,4 +1,4 @@
-// Copyright 2023-2025 Xanadu Quantum Technologies Inc.
+// Copyright 2025 Xanadu Quantum Technologies Inc.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.


doc/releases/changelog-dev.md
```diff
@@ -23,14 +23,17 @@
   [(#2128)](https://github.com/PennyLaneAI/catalyst/pull/2128)
   [(#2133)](https://github.com/PennyLaneAI/catalyst/pull/2133)
 
+* Fixes the underflow error when an axis of size 0 is present in a `DataView` object.
+  [(#1621)](https://github.com/PennyLaneAI/catalyst/pull/2164)
```
joeycarter: I would actually move this under the "Internal changes" section, since this wasn't a user-facing bug.

dime10: Can you mention how the error would present to the user?
What you wrote is the "technical" reason the error appears, but more important would be to answer how a user have encountered it? What kind of error would they have seen? What kind of program/pattern would they need to run? The technical details can always be obtained from the PR.

See here for some pointers on writing a good changelog entry: https://xanaduhq.slack.com/archives/C030UB33QH3/p1759270941744829

runtime/tests/Test_DataView.cpp
```diff
@@ -0,0 +1,178 @@
...
+#include <catch2/matchers/catch_matchers_string.hpp>
+#include <cstdint>
+#include <cstdio>
+
+#include "ExecutionContext.hpp"
+#include "RuntimeCAPI.h"
+
+#include "DataView.hpp"
+#include "TestUtils.hpp"
+#include "Types.h"
```
joeycarter: Some of these headers are unused so you can remove them:

```suggestion
#include <cstdio>

#include <catch2/catch_approx.hpp>
#include <catch2/catch_test_macros.hpp>
#include <catch2/matchers/catch_matchers_string.hpp>

#include "DataView.hpp"
```

runtime/tests/Test_DataView.cpp
```diff
@@ -0,0 +1,178 @@
...
+    size_t offset = 0;
+    size_t sizes[2] = {0, 10};
+    size_t strides[2] = {0, 0};
+
+    DataView<int, 2> view(data_aligned, offset, sizes, strides);
+
+    auto start = view.begin();
+    auto end = view.end();
+
+    CHECK(std::distance(start, end) == 0);
```
joeycarter: Nice, this is much more concise than what I originally proposed 👍 

You could probably do away with the `start` and `end` variables here and use `view.begin()` and `view.end()` directly in the call to `std::distance()`.

runtime/include/DataView.hpp
joeycarter, line 16: Could you add

```cpp
#include <vector>
```

while we're here? My IDE has been complaining about this for a while 😅

runtime/tests/Test_DataView.cpp
```diff
@@ -0,0 +1,178 @@
...
+    size_t sizes[2] = {10, 0};
+    size_t strides[2] = {0, 0};
+
+    DataView<int, 2> view(data_aligned, offset, sizes, strides);
+
+    auto start = view.begin();
+    auto end = view.end();
+
+    CHECK(std::distance(start, end) == 0);
+}
```
joeycarter, line 180: [Optional] Might be a good idea to add a couple of tests checking `distance` for non-empty arrays of a few different shapes.

kipawaa: Good idea, added in 0d8f9bd3.

runtime/include/DataView.hpp
kipawaa: No problem! added in 896354a.

runtime/tests/Test_DataView.cpp
kipawaa: Thanks again for that idea! The variables were left as a relic from other testing, definitely simpler to remove them. Changed in 0d8f9bd3.

runtime/tests/Test_DataView.cpp
kipawaa: Fixed in 896354a4.

doc/releases/changelog-dev.md
kipawaa: That makes sense, done in 9bcc965f.

runtime/include/DataView.hpp
```diff
@@ -15,6 +15,7 @@
 #pragma once
 
 #include <Exception.hpp>
+#include <vector>
```
joeycarter: Minor point, but our convention is to put system-level/STL includes at the top, followed by project-wide/external-package includes and then local headers last. Have a look at our guidance docs here: https://github.com/PennyLaneAI/guidance-docs/blob/master/development/cpp-dev-guidelines.md#include-file-use.

```suggestion
#include <vector>

#include <Exception.hpp>
```

kipawaa: Good catch, sorry I missed that!

runtime/tests/Test_DataView.cpp
```diff
@@ -0,0 +1,192 @@
+// Copyright 2023-2025 Xanadu Quantum Technologies Inc.
```
dime10: Very minor: new files start in the current year, modifications to existing files update the upper year
```suggestion
// Copyright 2025 Xanadu Quantum Technologies Inc.
```

doc/releases/changelog-dev.md
dime10: I see Joey mentioned this error was not accessible to users, in that case may comment may not apply. But I'd still be curious to know how we discovered/triggered it, which may be worth including.

dime10 (approved):
💯

doc/releases/changelog-dev.md
kipawaa: Thanks for the reference! It's true that the error was not accessible to users, but I've updated the changelog in 3f8697f7 with a link to Joey's discovery of the error. Let me know if that seems more appropriate?

PR #2165, opened 2025-10-30
Support pytree inputs when program capture enabled

albi3ro:
**Context:**

Change broken off from #2078, since it's basically it's own feature.

Just determine the inputs to `from_plxpr` from the plxpr itself. This means they will always match what's expected.

**Benefits:**

Always accurate with arbitrary nested arguments.

**Possible Drawbacks:**

Won't work with dynamically shaped inputs, but we don't support that yet anyway.

**Related GitHub Issues:**

[sc-102817]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -13,6 +13,7 @@
   [(#2154)](https://github.com/PennyLaneAI/catalyst/pull/2154)
 
 * Pytree inputs can now be used when program capture is enabled.
+  [(#2165)](https://github.com/PennyLaneAI/catalyst/pull/2165)
 
 <h3>Breaking changes 💔</h3>
 


doc/releases/changelog-dev.md
```diff
@@ -12,6 +12,8 @@
   :func:`~.qjit` with program capture enabled.
   [(#2154)](https://github.com/PennyLaneAI/catalyst/pull/2154)
 
+* Pytree inputs can now be used when program capture is enabled.
```
albi3ro, line 15:
```suggestion
* Pytree inputs can now be used when program capture is enabled.
  [(#2165)](https://github.com/PennyLaneAI/catalyst/pull/2165)
```

frontend/test/pytest/test_pytree_args.py
```diff
@@ -22,13 +22,15 @@
 import pennylane as qml
 import pytest
 from jax._src.tree_util import tree_flatten
+from pennylane import adjoint, cond, for_loop, grad, qjit
```
paul0403, line 25: 🤩

PR #2169, opened 2025-10-31
Auto detect xdsl passes

mehrdad2m:
**Context:**
We want to be able to use the xdsl interpreter to apply user passes without the need to specify the xdsl plugin manually. 

if a pass uses xdsl implementation, the transform sequence module and the transform itself is marked by adding a and attribute to the operation. this attribute is used later to detect wether we should use the xdsl compiler or not.

**Benefits:**
No need to specify the xdsl plugin path. 

**Possible Drawbacks:**
An extra walk over the program is added to detect the passes. This Could be avoided by keeping a global state when lowering the passes, but it would lose the locality of this information in cases with multiple qnodes that have different transform schedules.

**Related GitHub Issues:**
[sc-102894]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -27,13 +27,16 @@
   This removes the need to pass the `pass_plugins` argument to the `qjit` decorator.
   [(#2169)](https://github.com/PennyLaneAI/catalyst/pull/2169)
 
+* Dynamically allocated wires can now be passed into control flow and subroutines.
+  [(#2130)](https://github.com/PennyLaneAI/catalyst/pull/2130)
+
 <h3>Breaking changes 💔</h3>
 
 <h3>Deprecations 👋</h3>
 
 <h3>Bug fixes 🐛</h3>
 
-* Fixes an issue where a heap-to-stack allocation conversion pass was causing SIGSEGV issues 
+* Fixes an issue where a heap-to-stack allocation conversion pass was causing SIGSEGV issues
   during program execution at runtime.
   [(#2172)](https://github.com/PennyLaneAI/catalyst/pull/2172)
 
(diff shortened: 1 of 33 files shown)


frontend/catalyst/jax_primitives_utils.py
```diff
@@ -359,6 +363,25 @@ def transform_named_sequence_lowering(jax_ctx: mlir.LoweringRuleContext, pipelin
                     dynamic_options={},
                 )
                 target = apply_registered_pass_op.result
+
+                try:
+                    from pennylane.compiler.python_compiler.pass_api import (  # pylint: disable=import-outside-toplevel
```
mudit2812: To keep the line length short, probably better to add the pylint warning suppression to a different line:
```suggestion
                    # pylint: disable=import-outside-toplevel
                    from pennylane.compiler.python_compiler.pass_api import (
```

frontend/catalyst/jax_primitives_utils.py
```diff
@@ -359,6 +363,26 @@ def transform_named_sequence_lowering(jax_ctx: mlir.LoweringRuleContext, pipelin
...
+                    from pennylane.compiler.python_compiler.pass_api import (
+                        is_xdsl_pass,
+                    )
+
+                    if is_xdsl_pass(_pass.name):
+                        uses_xdsl_passes = True
+                        apply_registered_pass_op.operation.attributes["xdsl_pass"] = (
+                            ir.UnitAttr.get()
+                        )
+                except ImportError:
```
mudit2812: If the error you're trying to catch is for when xDSL is not installed, it would be a `ModuleNotFoundError`, not an `ImportError`.

```suggestion
                except ModuleNotFoundError:
```

mudit2812 (commented):
Thanks Mehrdad. Should there be tests for this? I would like for us to start moving away from not prioritizing tests for the unified compiler work

frontend/catalyst/compiler.py
```diff
@@ -478,19 +478,81 @@ def run_from_ir(self, ir: str, module_name: str, workspace: Directory):
...
+
+        try:
+            for op in mlir_module.operation.regions[0].blocks[0].operations:
+                try:
+                    # Check nested modules (look for transform modules)
+                    if hasattr(op, "regions") and len(op.regions) > 0:
+                        for nested_op in op.regions[0].blocks[0].operations:
+                            if hasattr(nested_op, "attributes"):
+                                if has_both_attributes(nested_op.attributes):
+                                    return True
```
joeycarter: I think this should also work (and it avoids an extra level of indentation):

```suggestion
                            if hasattr(nested_op, "attributes") and has_both_attributes(nested_op.attributes):
                                return True
```

frontend/catalyst/compiler.py
```diff
@@ -478,19 +478,81 @@ def run_from_ir(self, ir: str, module_name: str, workspace: Directory):
...
+                    or "uses_xdsl_passes" in getattr(attrs, "keys", lambda: [])()
+                )
+                return has_transform and has_xdsl
+            except (AttributeError, KeyError, TypeError):
+                return False
+
+        try:
+            for op in mlir_module.operation.regions[0].blocks[0].operations:
+                try:
+                    # Check nested modules (look for transform modules)
```
joeycarter: Is there any way we can refactor this section to reduce the number of indented blocks? e.g. by pulling some of it out into a separate function?

doc/releases/changelog-dev.md
```diff
@@ -42,12 +26,11 @@
...
-  [(#2142)](https://github.com/PennyLaneAI/catalyst/pull/2142)	
+  [(#2142)](https://github.com/PennyLaneAI/catalyst/pull/2142)
 
-* Re-work `DataView` to avoid an axis of size 0 possibly triggering a segfault via an underflow
-  error, as discovered in 
-  [this comment](https://github.com/PennyLaneAI/catalyst/pull/1598#issuecomment-2779178046).
-  [(#1621)](https://github.com/PennyLaneAI/catalyst/pull/2164)
+* xDSL passes are now automatically detected when using the `qjit` decorator. 
+  This removes the need to pass the `pass_plugins` argument to the `qjit` decorator.
+  [(#2169)](https://github.com/PennyLaneAI/catalyst/pull/2169)
```
joeycarter: Should we put this in the "Improvements" section? No longer needing to pass certain arguments to `qjit` is a user-facing change, even if xDSL work is still somewhat experimental.

mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/2169/commits/edd26e31215a3ef246a33056448a83a86aa755ee

frontend/catalyst/compiler.py
mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/2169/commits/46d17d2400dea9d96300567d17b76b88039b2784

mudit2812 (approved):
> Something else that might be nice to have is lit tests for the transform module lowered IR but those again has to wait for the https://github.com/PennyLaneAI/catalyst/pull/2169 in pennyLane to be merged first.

This is perfect. I'll approve to unblock for now in that case.

PR #2171, opened 2025-11-04
Switch control flow

### LINKED ISSUE #1702: [Feature Request] Multi-branch cond
Currently the Catalyst version of the Jax conditional [cond_p](https://github.com/PennyLaneAI/catalyst/blob/ba3f7517e6741049bc0f5a18990cbce02e5475c4/frontend/catalyst/jax_primitives.py#L265) only supports two branches. The Jax version can indeed process multiple branches (the front-end for this feature is lax.switch). While it is possible to emulate a N-case switch with a static loop and N 2-case switches, this is a bit of a duct-tape solution.

### LINKED ISSUE #1943: Adding support for switch statements over quantum instructions
Adding support of switch/case statements could greatly enhance the potential use cases of leveraging real-time classical processing and feedforward. This should be doable through the existing implementation of JAX switch expressions.

kipawaa:
**Context:**
Catalyst currently lacks a switch interface, a notable difference in control flow between the jax interface and catalyst. Switches allow for more efficient, non-recursive lowering of distinct cases and can simplify control flow among multiple branches.

Adds index-switch functionality to catalyst, bringing the frontend further inline with jax. The switch is provided as a decorator and is compatible with `@qjit` compilation.

Example:
```python
@qjit
def my_circuit(i, theta):
    @catalyst.switch(i) # initialize a switch on variable i
    def my_switch(angle): # this is the default branch (required)
        qml.RX(angle, wires=0)

    @my_switch.branch(1) # create a branch with case i = 1
    def my_branch(angle):
        qml.RY(angle, wires=0)

    @my_switch.branch(4) # create a branch with case i = 4
    def my_branch_4(angle):
        qml.H(0)

    my_switch(theta) # must invoke the switch
    return qml.probs()
```
```python
>>> my_circuit(4, 0) # case i = 4
[0.5 0.5]
>>> my_circuit(-3, pi) # un-matched cases use default branch
[0. 1.]
```

**Benefits:**
Switches simplify control flow over multiple branches, and allow for more efficient lowering to `IndexSwitchOp` rather than recursively lowering to `ConditionalOp`.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
Closes #1702
Closes #1943

[sc-96414]

### DIFF
--- a/frontend/catalyst/api_extensions/control_flow.py
+++ b/frontend/catalyst/api_extensions/control_flow.py
@@ -1192,7 +1192,7 @@ class SwitchCallable:
             )
         return self._operation
 
-    def branch(self, case):
+    def branch(self, case: int):
         """
         Branch to be run if the switch's case is equivalent to the case provided here.
 
@@ -1203,7 +1203,7 @@ class SwitchCallable:
             A callable decorator that wraps this case of the switch.
         """
 
-        def decorator(branch):
+        def decorator(branch: Callable):
             self.case_to_branch[case] = branch
             return self
 
@@ -1298,7 +1298,7 @@ class SwitchCallable:
         branches.append(self.default_branch)
 
         # wraps trace to allow simple unzipping
-        def _trace(branch_fn):
+        def _trace(branch_fn: Callable):
             _, in_sig, out_sig = trace_function(
                 branch_fn,
                 *(),


frontend/test/pytest/test_conditionals.py
```diff
@@ -212,8 +212,8 @@ def cond_fn():
             cond_fn()
             return measure(wires=0)
 
-        assert circuit(3) == False
-        assert circuit(6) == True
```
rniczh, line 216: Just wondering why these tests need to be changed?

frontend/test/pytest/test_conditionals.py
```diff
@@ -588,7 +588,7 @@ def f(x: int):
             with pytest.raises(ValueError, match="false branch must be provided"):
                 qjit(f)
         else:
-            with pytest.raises(TypeError, match="Please specify an else branch"):
+            with pytest.raises(TypeError, match="requires a consistent return structure"):
```
rniczh, line 591: Doesn't the original error message make more sense here?

frontend/test/pytest/test_switch.py
```diff
@@ -0,0 +1,729 @@
...
+            def my_default():
+                return "default"
+
+            return my_switch()
+
+        assert circuit(-4) == "default"
+        assert circuit(-1) == "default"
+        assert circuit(0) == "first"
+        assert circuit(1) == "default"
+        assert circuit(12) == "default"
```
rniczh, line 76: I'm quite confuse with this design, why `circuit(0)` should be "first" instead of "default". Doesn't each case should specify the case value, except for the default one.

frontend/catalyst/api_extensions/control_flow.py
```diff
@@ -559,6 +559,127 @@ def _decorator(body_fn):
...
+    traced at compile time, but only one will be executed at runtime, depending on the value of
+    `case_var`.  The JAX equivalent of this is the ``jax.lax.switch`` function, but this version
+    allows for arbitrary integer valued cases, does not clamp the index, and is optimized to work
+    with quantum programs in PennyLane.  This version also supports a default branch that is not
+    present in the JAX implementation.
+
+    Values produced within the scope of the branches can be returned to the outside context, but
+    the return type signature of each branch must be identical.
+    If no default branch is assigned, the first branch provided will be assigned as default.
+    The first branch is assigned case 0 by default, but can be overridden.
```
rniczh: What the purpose of assigning 0 as case value for the first branch?

frontend/catalyst/api_extensions/control_flow.py
```diff
@@ -1365,6 +1698,68 @@ def trace_quantum(self, ctx, device, trace, qrp) -> QRegPromise:
         return qrp2
 
 
+class Switch(HybridOp):
+    """TODO"""
+
+    binder = switch_p.bind
+
+    def trace_quantum(self, ctx, device, trace, qrp) -> QRegPromise:
```
rniczh, line 1660: The body of the function here is just exactly do the same thing as `Cond.trace_quantum`, couldn't it just call the `Cond.trace_quantum` function directly instead of copying the whole function?

frontend/test/pytest/test_conditionals.py
kipawaa: After I made other changes in this file the style checker was upset about these lines comparing against boolean literals, so I changed them to assert the values. I can change this back, but since I was updating the file I thought that I might as well include that change.

frontend/test/pytest/test_conditionals.py
kipawaa: The original message is more informative, but it was returned by a helper function that is now used for both conditionals and switches, and it's misleading if it gets returned from a switch. I can add a catch specifically in `cond` if you think that I should preserve this message for conditionals.

frontend/catalyst/api_extensions/control_flow.py
kipawaa: There is a difference in the call to `op.bind_overwrite_classical_tracers`. There is a lot of similarity though, perhaps it would be worth setting up a helper function for the core functionality and leaving just the `bind_overwrite_classical_tracers` as the unique entry in their respective `trace_quantum` methods?

doc/dev/quick_start.rst
```diff
@@ -330,6 +330,7 @@ the compilation of arbitrarily parametrized circuits.
     :nosignatures:
 
     ~catalyst.cond
+    -catalyst.switch
```
joeycarter: This wasn't rendering (need a tilde `~` rather than a hyphen `-`)

```suggestion
    ~catalyst.switch
```

Also just to confirm, the conclusion was that we _do_ want to add this to the Quick Start guide?

doc/dev/quick_start.rst
```diff
@@ -376,6 +377,38 @@ decorator.
...
+
+        @my_switch.default() # optionally specify a default branch
+        def my_default_branch():
+            return 0
+
+        return my_switch() # must invoke the switch
+
+.. warning::
+
+    Switches are not currently compatible with program capture, or autograph.
```
joeycarter:
```suggestion
    Switch functions are not currently compatible with program capture or autograph.
```

frontend/catalyst/api_extensions/control_flow.py
```diff
@@ -559,6 +559,127 @@ def _decorator(body_fn):
...
+    present in the JAX implementation.
+
+    Values produced within the scope of the branches can be returned to the outside context, but
+    the return type signature of each branch must be identical.
+    If no default branch is assigned, the first branch provided will be assigned as default.
+    The first branch is assigned case 0 by default, but can be overridden.
+    Refer to the examples below to learn more about this decorator.
+
+    This form of control flow can also be called from the Python interpreter without needing to use
+    :func: `~.qjit`.
```
joeycarter: This isn't rendering correctly, I think this should fix it:

```suggestion
    :func:`~.qjit`.
```

frontend/catalyst/api_extensions/control_flow.py
```diff
@@ -559,6 +559,127 @@ def _decorator(body_fn):
...
+
+            my_switch()
+
+            return qml.probs()
+    >>> circuit(0)
+    [1. 0.]
+    >>> circuit(1)
+    [0. 1.]
+
+    A default branch that catches any value not otherwise specified can be set via the ``.default``
```
joeycarter:
```suggestion
    A default branch that catches any value not otherwise specified can be set via the ``default``
```

frontend/catalyst/api_extensions/control_flow.py
```diff
@@ -559,6 +559,127 @@ def _decorator(body_fn):
...
+    method, and the initial case can be specified using the `case` parameter
```
joeycarter:
```suggestion
    method, and the initial case can be specified using the `case` parameter:
```

frontend/catalyst/api_extensions/control_flow.py
```diff
@@ -559,6 +559,127 @@ def _decorator(body_fn):
...
+
+            return my_switch() # must invoke the switch
+    >>> foo(6)
+    (1.4+0j)
+    >>> foo(3)
+    (2+0j)
+
+    .. note::
+
+        ``catalyst.switch`` is not supported in program capture mode.
```
joeycarter:
```suggestion
        ``catalyst.switch`` is not supported with PennyLane program capture enabled.
```

frontend/catalyst/api_extensions/control_flow.py
```diff
@@ -559,6 +559,127 @@ def _decorator(body_fn):
...
+        ``catalyst.switch`` also does not support autograph (``qjit(autograph=True)``).
```
joeycarter:
```suggestion
        There is also currently no support for automatic conversion of native Python ``match`` statements to the ``catalyst.switch`` operation when using :func:`~.qjit` with AutoGraph enabled.
```

joeycarter (commented):
Nicely done, thanks @kipawaa! This is an impressive amount of work 🚀 

I'll start with a few comments on the docs and take a closer look at the technical details tomorrow.

frontend/test/pytest/test_conditionals.py
rniczh: If that because of the pylint, I think it's fine 👍  Don't need to change it back, because we still try to meet pylint requirements as much as possible.

doc/dev/quick_start.rst
kipawaa: Isaac said we should add it but make sure to be clear about the differences in capabilities with autograph and capture. He was also curious when we might add those 👀

kipawaa:
@dime10 
> * What does this part do (i.e. when would this body be executed)?
> 
> ```python
>     @catalyst.switch(i) # initialize a switch
>     def my_switch(angle):
>         qml.RX(angle, wires=0)
> ```
The first branch was designated case 0 ***and*** default branch when i first implemented this (the default could later be overridden with `switch.default`). @rniczh pointed out that implicitly assigning a case here was confusing, so we're discussing how it should be changed. This behaviour likely won't stay, but we haven't decided what it should be changed to.
 
> * Is the default case mandatory? With cond we made the else optional if no values are returned from the branches. Alternatively (depending on the answer to the first question) if it is mandatory, the switch could be initialized with the default?

We're also discussing this. I lean towards making the default case mandatory. If we do that, there are two options: we can require it gets declared explicitly with `switch.default`, or we have it assigned with the initial declaration of the switch. Currently chatting with product about these options.

frontend/catalyst/api_extensions/control_flow.py
rniczh: https://github.com/PennyLaneAI/catalyst/pull/2171/commits/199b1c756606e820315a6aef35b64721dc736d84

frontend/test/pytest/test_switch.py
rniczh: https://github.com/PennyLaneAI/catalyst/pull/2171/commits/199b1c756606e820315a6aef35b64721dc736d84

frontend/test/pytest/test_conditionals.py
rniczh: No worries, I think the new one is a better fit here for capturing the error message. The error message only shows `Please specify an else branch if PyTreeDef(None)`, so users still need to investigate it themselves 🤔

```python
  raise TypeError(
-      "Conditional requires a consistent return structure across all branches! "
-      f"Got {tree} and {expected_tree}. Please specify an else branch if PyTreeDef(None) "
-      "was specified."
+      "Control flow requires a consistent return structure across all branches! "
+      f"Got {tree} and {expected_tree}."
  )
```

frontend/catalyst/api_extensions/control_flow.py
kipawaa: Merged the bodies of these in 393fdb5.

rniczh: looks great 👍

frontend/catalyst/api_extensions/control_flow.py
```diff
@@ -1020,6 +1132,227 @@ def __call__(self, *init_state):
...
+
+        >>> circuit(0)
+        [0.5 0.5]
+        >>> circuit(2)
+        [0. 1.]
+        >>> circuit(5)
+        [1. 0.]
+    """
+
+    def __init__(self, case, default_branch, cases=[], branches=[]):
```
joeycarter: This CodeFactor warning on `dangerous default value []` is probably worth addressing; see here for reference: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments.

Would it be possible to change `cases=[], branches=[]` to this?

```python
cases: list = None, branches: list = None
```

paul0403: We can use empty tuples, they are immutable.

joeycarter: That works too :+1:

kipawaa: Updated default values in dc33bd3, added a few short lit tests in db8f4ad. Thanks for the good feedback!

PR #2174, opened 2025-11-05
Update nanobind 2.9

rniczh:
**Context:**

Latest LLVM update https://github.com/PennyLaneAI/catalyst/pull/2122 requires `nanobind>=2.9` 

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-103342]


mehrdad2m (approved):

dime10:
> Should we wait until the current run on main finishes?

Yeah we shouldn't build llvm on PR branches, that will probably max out the cache if there are multiple instances.

PR #2177, opened 2025-11-06
Update for use of a single transform primitive

albi3ro:
**Context:**

See https://github.com/PennyLaneAI/pennylane/pull/8576

Right now each transform has its own primitive, so we have to register each of those primitives in a whole bunch of different locations.  This is hard to manage, and can lead to memory leaks.

So we are just going to deal with a single trnasform primitive with a `transform` kwarg that holds onto the `TransformDispatcher` object.

Update to have a single registration to handle transforms.

**Benefits:**

Less work to extend to new transforms.

**Possible Drawbacks:**

Harder to customize the behaviour for one specific transform.

**Related GitHub Issues:**

[[sc-103382](https://app.shortcut.com/xanaduai/story/103382)]


doc/releases/changelog-dev.md
```diff
@@ -75,6 +75,8 @@
 
 <h3>Internal changes ⚙️</h3>
 
+* Updates to PennyLane's use of a single transform primitive with a `transform` kwarg.
```
albi3ro, line 101:
```suggestion
* Updates to PennyLane's use of a single transform primitive with a `transform` kwarg.
  [(#2177)](https://github.com/PennyLaneAI/catalyst/pull/2177)
```

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -267,108 +268,97 @@ def calling_convention(*args):
...
+        # gds_interpreter = DecompRuleInterpreter(*targs, **tkwargs)
+
+        # def gds_wrapper(*args):
+        #     return gds_interpreter.eval(inner_jaxpr, consts, *args)
+
+        # final_jaxpr = jax.make_jaxpr(gds_wrapper)(*args)
+        # return self.eval(final_jaxpr.jaxpr, consts, *non_const_args)
         return next_eval.eval(inner_jaxpr, consts, *non_const_args)
 
+    catalyst_pass_name = transforms_to_passes[transform][0]
```
mudit2812:
```suggestion
    catalyst_pass_name = transforms_to_passes.get(transform, (None,))[0]
```

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -267,108 +268,97 @@ def calling_convention(*args):
...
-        tkwargs,
-        catalyst_pass_name=pass_name,
-        requires_decomposition=decomposition,
-        pl_plxpr_transform=pl_transform._plxpr_transform,
+    transforms_to_passes[pl_transform] = (pass_name, decomposition)
+
+
+# pylint: disable=too-many-arguments
+@WorkflowInterpreter.register_primitive(transform_prim)
+def handle_transform(
```
paul0403, line 278: Nice, just to confirm my understanding: 
- this `handle_transform` is now the `custom_handler` for the new `pennylane.capture.primitives.transform_prim` single transform primitive, used like all custom handlers in base plxpr interpreter? And so we do not have to register the per-transform primitives one at a time (the `register_primitive(pl_transform._primitive)`)?
- And consequently, the above `register_transform` does not need to be called in catalyst manually anymore, and is only left in here as a internal API for PL to register more transforms not already handled by catalyst here? (e.g. the PPR transforms https://github.com/PennyLaneAI/pennylane/blob/master/pennylane/ftqc/catalyst_pass_aliases.py)

albi3ro: Exactly.  I hope to add a `pass_name` to transforms soon so we will be able to do away with `register_transform`, but I decided it was best to split this into two steps.

No more need to register everything on a per-transform basis.

paul0403 (approved):
All good from catalyst side :+1:

mudit2812 (approved):
I misread the diff. I thought you were removing the `register_transform` function, but you're not. So, no changes should be needed to `python_compiler`

PR #2178, opened 2025-11-06
Starting to better organize tests for program capture

albi3ro:
**Context:**

To better keep track of what is and isn't supported with program capture, I'm trying out adding markers for things that just work the old frontend, and features we still need to get working with program capture.

Adds two markers, and starts applying them to tests.


frontend/test/pytest/test_adjoint.py
```diff
@@ -22,16 +22,17 @@
 import pennylane.numpy as pnp
 import pytest
 from numpy.testing import assert_allclose
-from pennylane import adjoint, cond, for_loop, qjit, while_loop
+from pennylane import adjoint, cond, for_loop, qjit, while_loop, qjit
```
JerryChen97, line 25:
```suggestion
from pennylane import adjoint, cond, for_loop, qjit, while_loop
```
it's already imported before `while_loop`

frontend/test/pytest/test_seeded_qjit.py
```diff
@@ -16,9 +16,18 @@
 import numpy as np
 import pennylane as qml
 import pytest
+from pennylane import cond
 
-from catalyst import CompileError, cond, measure, qjit
+from catalyst import CompileError, measure, qjit
```
JerryChen97, line 21:
```suggestion
from catalyst import CompileError, qjit
```

PR #2185, opened 2025-11-10
RTIO dialect is added to Catalyst

rniczh:
**Context:**

This PR introduces RTIO dialect, a dilaect is designed for FPGA-based quantum control system. The design bridges high-level quantum abstractions with low-level hardware control primitives, enabling precise timing control and asynchronous  execution of quantum pulses.

The IR design introduces an additional layer of abstraction. At the language level, only the semantics regarding timing relationships are preserved. This approach serves two purposes: it ensures generalizations across other arch that adopt the timing semantic, and also maintains flexibility for future arch changes within OQD. For OQD hardware, it will lower to the precise timestamp semantics at the end.

***RTIO Types:***
- `!rtio.event`: Event handle representing pending or completed hardware operations, used for explicit dependency tracking and synchronization
- `!rtio.channel<KIND, [QUALIFIERS], ID>`: Hardware channel type with:
  - `KIND`: Channel kind (e.g., `"dds"` for Direct Digital Synthesizer)
  - `QUALIFIERS`: Optional array of qualifiers for logical channel identification (e.g., ion index, transition index). it is used to distinguish the different logical channels of the same kind. Considering the `ion 0` with `transition 0`, the mapping of this channel should different from the `transition 1` with the same ion. That being said, the qualifiers provide a way for developer to take some note to the certain qubit. But the mapping logic might vary from HW design, so user still need to specify the mapping logic during lowering process.

    The current mapping logic of OQD ion HW is quite trivial:
    `qubit0` x `transition 0` -> `channel 0`
    `qubit0` x `transition 1` -> `channel 1`
    `qubit1` x `transition 0` -> `channel 2`
    `qubit1` x `transition 1` -> `channel 3`

  - `ID`: Channel identifier (`?` for dynamic resolution, or integer for static hardware channel).

***Event-Based API:***
- `rtio.channel`: Creates a channel instance with identity encoded in the result type
- `rtio.qubit_to_channel`: Maps a qubit value to an RTIO channel
- `rtio.pulse`: Generates a pulse event with:
  - Channel operand
  - Duration, frequency, and phase parameters
  - Optional `wait` operand for event dependencies
  - Returns an `!rtio.event` handle
- `rtio.sync`: Synchronization barrier that waits for multiple events and returns a combined event handle
- `rtio.empty`: Creates an empty event for initializing event chains

**RTIO Attributes:**
- `rtio.config<{...}>`: Used to import the dict attr into MLIR. Currently, for OQD, it import the `device_db.json` (which is converted from debug_db.py directly) into MLIR, so we can access those HW specific meta data for compiling.

***TODO:***
- [x] Remove Timeline IR from RTIO, just lowering to the LLVM directly

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-100850]


mlir/tools/quantum-opt/quantum-opt.cpp
```diff
@@ -69,6 +70,7 @@ int main(int argc, char **argv)
     registry.insert<catalyst::mbqc::MBQCDialect>();
     registry.insert<catalyst::mitigation::MitigationDialect>();
     registry.insert<catalyst::ion::IonDialect>();
+    registry.insert<catalyst::rtio::RTIODialect>();
```
mehrdad2m, line 73: Can we also register the dialect in` mlir/tools/quantum-lsp-server/quantum-lsp-server.cpp` and `mlir/lib/Driver/CompilerDriver.cpp`?

mlir/include/RTIO/IR/RTIODialect.td
```diff
@@ -0,0 +1,164 @@
...
+    let summary = "Real-Time I/O dialect for FPGA quantum control";
+    let description = [{
+        The RTIO dialect provides operations for precise timing control
+        and hardware signal generation on FPGAs for quantum computing.
+
+        This dialect supports two levels of abstraction:
+        1. Event-Based IR (high-level): Declarative operations with explicit event dependencies
+
+        // TODO: Do we need the separate Timeline IR for artiq family?
+        2. Timeline IR (low-level): Stateful operations with implicit time cursor
```
mehrdad2m: Are we going to add these in a separate PR?

mlir/lib/RTIO/IR/RTIODialect.cpp
```diff
@@ -0,0 +1,150 @@
...
+#include "RTIO/IR/RTIOOpsTypes.cpp.inc"
+        >();
+
+    addOperations<
+#define GET_OP_LIST
+#include "RTIO/IR/RTIOOps.cpp.inc"
+        >();
+}
+
+// Not support any custom attributes yet, might be supported in the future
```
mehrdad2m: I can see that we set `useDefaultAttributePrinterParser = 0`, but not providing implementations. If we are not supporting custom implementation for `parseAttribute`/`printAttribute` why not using the default methods?

mlir/include/RTIO/IR/RTIODialect.td
```diff
@@ -0,0 +1,164 @@
...
+        ```
+
+        Examples:
+        ```mlir
+
+        // Simple DDS, channel TBD
+        !rtio.channel<"dds", ?>
+
+        // DDS for one two qualifiers 0 and "t0", channel TBD
+        !rtio.channel<"dds", [0, "transition_0"], ?>
```
mehrdad2m, line 77: Can you expand on what is the vision for qualifiers and they will be used?

mlir/include/RTIO/IR/RTIODialect.td
```diff
@@ -0,0 +1,164 @@
...
+        !rtio.channel<KIND, [QUALIFIERS], N>   // Static: with qualifiers + channel N
+        ```
+
+        Examples:
+        ```mlir
+
+        // Simple DDS, channel TBD
+        !rtio.channel<"dds", ?>
+
+        // DDS for one two qualifiers 0 and "t0", channel TBD
```
mehrdad2m:
```suggestion
        // DDS channel with two qualifiers 0 and "t0", channel ID TBD
```

mlir/include/RTIO/IR/RTIODialect.td
```diff
@@ -0,0 +1,164 @@
...
+// Channel type with variadic parameters
+def RTIOChannelType : RTIO_Type<"Channel", "channel"> {
+    let summary = "A hardware I/O channel with logical and physical identification";
+    let description = [{
+        Represents a virtual hardware channel for RTIO operations
+
+        Syntax:
+        ```
+        !rtio.channel<KIND, ?>                 // Dynamic: kind only
+        !rtio.channel<KIND, N>                 // Static: kind + channel N
```
mehrdad2m:
```suggestion
        !rtio.channel<KIND, N>                 // Static: kind + channel ID
```

mlir/include/RTIO/IR/RTIODialect.td
```diff
@@ -0,0 +1,164 @@
...
+        !rtio.channel<KIND, [QUALIFIERS], ?>   // Dynamic: with qualifiers
+        !rtio.channel<KIND, [QUALIFIERS], N>   // Static: with qualifiers + channel N
```
mehrdad2m:
```suggestion
        !rtio.channel<KIND, [QUALIFIERS], N>   // Static: with qualifiers + channel ID
```

mlir/include/RTIO/IR/RTIODialect.td
```diff
@@ -0,0 +1,164 @@
...
+        // === Channel Resolution During Compilation ===
+
+        // Before channel resolution: Dynamic channels
+        !rtio.channel<"dds", [0], ?>
+
+        // After channel resolution: Resolved to hardware channel 1
+        !rtio.channel<"dds", [0], 1>
+
+        // Note: The qualifiers are used provided addtional information to the channel type.
+        // And it will used to distinguish different channels with the same kind.
```
mehrdad2m:
```suggestion
        // Note: The qualifiers are used to provide additional information to the channel type.
        // And they will be used to distinguish different channels with the same kind.
```

mlir/include/RTIO/IR/RTIODialect.td
```diff
@@ -0,0 +1,164 @@
...
+    let description = [{
+        The RTIO dialect provides operations for precise timing control
+        and hardware signal generation on FPGAs for quantum computing.
+
+        This dialect supports two levels of abstraction:
+        1. Event-Based IR (high-level): Declarative operations with explicit event dependencies
+
+        // TODO: Do we need the separate Timeline IR for artiq family?
+        2. Timeline IR (low-level): Stateful operations with implicit time cursor
+        ```
```
mehrdad2m:
```suggestion
```

mlir/include/RTIO/IR/RTIOOps.td
```diff
@@ -0,0 +1,357 @@
...
+        attr-dict `:` type(results)
+    }];
+}
+
+def RTIOQubitToChannelOp : RTIO_Op<"qubit_to_channel"> {
+    let summary = "Map a qubit to an RTIO channel";
+    let description = [{
+        It's a temporary operation that will be lowered to a static `rtio.channel` operation during
+        the channel resolution stage. The purpose of this operation is to allow the qubit from a
+        from a high-level dialect to be mapped to an `!rtio.channel`.
```
mehrdad2m:
```suggestion
        the channel resolution stage. The purpose of this operation is to allow a qubit from a
        high-level dialect to be mapped to an `!rtio.channel`.
```

mlir/include/RTIO/IR/RTIOOps.td
```diff
@@ -0,0 +1,357 @@
...
+include "mlir/IR/BuiltinAttributes.td"
+include "mlir/Interfaces/SideEffectInterfaces.td"
+include "RTIO/IR/RTIODialect.td"
+
+//===----------------------------------------------------------------------===//
+// Event-Based API
+//===----------------------------------------------------------------------===//
+
+def RTIOChannelOp : RTIO_Op<"channel"> {
+    let summary = "Define a static channel";
```
mehrdad2m: Does Static here mean a resolved channel ID? in that case, it seems that dynamic channel is also supported.

mlir/include/RTIO/IR/RTIOOps.td
```diff
@@ -0,0 +1,357 @@
...
+    let arguments = (ins AnyType:$qubit);
+    let results = (outs RTIOChannelType:$channel);
+
+    let assemblyFormat = [{
+        $qubit attr-dict `:` type(operands) `->` type(results)
+    }];
+}
+
+def RTIOPulseOp : RTIO_Op<"pulse"> {
+    let summary = "Generate a event-based pulse on a channel";
```
mehrdad2m:
```suggestion
    let summary = "Generate an event-based pulse on a channel";
```

mlir/include/RTIO/IR/RTIOOps.td
```diff
@@ -0,0 +1,357 @@
...
+        Generate a pulse on the channel. Returns an `!rtio.event` handle that can be used with
+        `rtio.sync` to wait for completion or to be used by other pulse operations with the `wait`
+        operand.
+
+        Parameters:
+        - channel: Target output channel
+        - duration: Pulse duration in seconds (f64)
+        - frequency: frequency in Hz
+        - phase: Phase in radians
+        - wait: Optional events to wait for before starting this pulse (for sequencing)
```
mehrdad2m, line 92: Shouldn't this be handled by the `sync` operation? what would be the difference between having this vs having a sync?

mlir/include/RTIO/IR/RTIOOps.td
```diff
@@ -0,0 +1,357 @@
...
+
+def RTIOOnOp : RTIO_Op<"on"> {
+    let summary = "Turn on a channel output";
+    let description = [{
+        Activates the output of a channel at the current cursor.
+        For DDS channels, this typically enables the RF output switch.
+
+        Example:
+        ```mlir
+        rtio.at %t_start : !rtio.time
```
mehrdad2m: Do we have a `!rtio.time` in the dialect?

mlir/include/RTIO/IR/RTIOOps.td
```diff
@@ -0,0 +1,357 @@
...
+    }];
+
+    let arguments = (ins
+        RTIOChannelType:$channel,
+        F64:$frequency
+    );
+    let assemblyFormat = "$channel `,` $frequency attr-dict `:` type($channel) `,` type($frequency)";
+}
+
+def RTIOSetPhaseOp : RTIO_Op<"set_phase"> {
```
mehrdad2m: shouldn't phase be a property of the pulse rather than the channel? 🤔 I am wondering if we could just model it as a delay before the pulse starts

mlir/tools/quantum-opt/quantum-opt.cpp
mehrdad2m: Also maybe in the CAPI module

mlir/include/RTIO/IR/RTIODialect.td
rniczh: No, it's a question, I just put it here for discussion, but I think for now, nothing need to change.

mlir/include/RTIO/IR/RTIOOps.td
rniczh: For an example, if there are two pulse and one pulse is depend on the other pulse:
```
event0 <- pulse(..., wait=None)
event1 <- pulse(..., wait=event1)
```
It can not just rely on `sync` to establish the event dependency. Since `sync` is sync the events from its source, in the following case, to ensure the event3 is after event0 and event1, it need to sync `event0` and `event1` and pass the `event2` to `event3` with wait operand.
```
event0 <- pulse(..., wait=None)
event1 <- pulse(..., wait=None)
event2 <- sync(event0, event1)
event3 <- pulse(event0, wait=event2)
```

mlir/include/RTIO/IR/RTIOOps.td
rniczh: No, it's typo when developing

mlir/include/RTIO/IR/RTIOOps.td
rniczh: I think model it as `delay` works here. We can remove this one I think.

mlir/include/RTIO/IR/RTIOOps.td
mehrdad2m: Ah I see, Thanks for the explanation

mlir/include/RTIO/IR/RTIOOps.td
mehrdad2m: We can also discuss this on Wednesday to decide how phase should be handled in general. I don't think I have tried lowering an ARTIQ experiment with nonzero phase before to see how they handle it, but it might be worth trying to get more insight.

mlir/include/RTIO/IR/RTIODialect.td
mehrdad2m: I see, yeah we can decide on this later as we develop the lowering. I think if the lowering is simple enough an artiq-specific lowering pass would more ideal than an artiq-specific IR/dialect.

mlir/include/RTIO/IR/RTIOOps.td
rniczh: Yes, it supports the dynamic channel, it's also another typo here, i will fix it in next revision

mlir/include/RTIO/IR/RTIODialect.td
rniczh: Qualifiers are used to distinguish the different logical channels of the same kind.
Consider the `ion 0` with `transition 0`, the channel mapping of this channel should different from the `transition 1` with the same ion. That being said, the qualifiers are more like a hint for developer to take some note to the certain qubit usage and it may vary from HW design. So user still need to specify the mapping logic.

The current mapping logic of OQD ion HW is quite trivial:
`qubit0` x `transition 0` -> `channel 0`
`qubit0` x `transition 1` -> `channel 1`
`qubit1` x `transition 0` -> `channel 2`
`qubit1` x `transition 1` -> `channel 3`

I updated the comment on the td file as well https://github.com/PennyLaneAI/catalyst/pull/2185/commits/c8193334f0d4118d47d126be989ddeff6e4e6287

mlir/include/RTIO/IR/RTIOOps.td
rniczh: https://github.com/PennyLaneAI/catalyst/pull/2185/commits/c0e7234fe9b7bc0e08c2a02f644049d9d0bf32ab

mlir/include/RTIO/IR/RTIOOps.td
rniczh: https://github.com/PennyLaneAI/catalyst/pull/2185/commits/c0e7234fe9b7bc0e08c2a02f644049d9d0bf32ab

mlir/lib/RTIO/IR/RTIODialect.cpp
rniczh: Currently, if want to set `useDefaultAttributePrinterParser` as 1, MLIR require us to at least, create an attribute for them to instance the `parseAttribute/printAttribute` impl. Otherwise, it will show an error:
```
"catalyst::rtio::RTIODialect::parseAttribute(mlir::DialectAsmParser&, mlir::Type) const", referenced from:
      vtable for catalyst::rtio::RTIODialect in libMLIRRTIO.a[2](RTIODialect.cpp.o)
  "catalyst::rtio::RTIODialect::printAttribute(mlir::Attribute, mlir::DialectAsmPrinter&) const", referenced from:
      vtable for catalyst::rtio::RTIODialect in libMLIRRTIO.a[2](RTIODialect.cpp.o)
ld: symbol(s) not found for architecture arm64
```

mlir/tools/quantum-opt/quantum-opt.cpp
rniczh: commits for lsp-server and compilerDriver:
https://github.com/PennyLaneAI/catalyst/pull/2185/commits/4cf83fec25261de44932a16a34ff00403af64db7
https://github.com/PennyLaneAI/catalyst/pull/2185/commits/27a4b981e011d17d71c04ec81a5d9495ef03a1ca

mlir/test/RTIO/VerifierTest.mlir
```diff
@@ -0,0 +1,199 @@
...
+}
+
+// -----
+
+///////////////////////////////
+// Timeline-Based Operations //
+///////////////////////////////
+
+func.func @timeline_now() {
+    %t = rtio.now : i64
```
paul0403: I see you declared the `empty` and `now` ops as `Pure`. Will they get dead code eliminated here in the tests if you don't return them?

paul0403: Or does that even matter, since this is just a smoke test? :thinking:

mlir/include/RTIO/IR/RTIODialect.td
```diff
@@ -0,0 +1,168 @@
...
+            return getQualifiers() && !getQualifiers().empty();
+        }
+
+        size_t getNumQualifiers() const {
+            return getQualifiers() ? getQualifiers().size() : 0;
+        }
+
+        mlir::Attribute getQualifier(size_t index) const {
+            if (!getQualifiers() || index >= getQualifiers().size())
+                return nullptr;
```
paul0403: I feel like this should be an assert if index are out of bounds for an indexing operation

mlir/include/RTIO/IR/RTIODialect.td
```diff
@@ -0,0 +1,168 @@
...
+            return getQualifiers()[index];
+        }
+
+        bool isStatic() const {
+            auto channelId = getChannelId();
+            if (!channelId) {
+                return false;
+            }
+            return channelId.getInt() >= 0;
```
paul0403:
```suggestion
            return channelId.getInt() != ShapedType::kDynamic;
```

there is this thing in the mlir capi btw, I guess it was always meant for dynamic shapes since the only numbers in their types are shapes?

mlir/test/RTIO/VerifierTest.mlir
```diff
@@ -0,0 +1,199 @@
...
+    %ch0 = rtio.channel : !rtio.channel<"dds", 0>
+    %ch0_q1 = rtio.channel : !rtio.channel<"dds", [0], 0>
+    %ch0_q2 = rtio.channel : !rtio.channel<"dds", [0, "t0"], 0>
+    %ch_explict_dyn = rtio.channel : !rtio.channel<"dds", ?>
+    %ch_explict_dyn_q1 = rtio.channel : !rtio.channel<"dds", [0], ?>
+    %ch_explict_dyn_q2 = rtio.channel : !rtio.channel<"dds", [0, "t0"], ?>
+
+    %ch0_implicit_dyn = rtio.channel : !rtio.channel<"dds">
+    %ch_implicit_dyn_q1 = rtio.channel : !rtio.channel<"dds", [0]>
+    %ch_implicit_dyn_q2 = rtio.channel : !rtio.channel<"dds", [0, "t0"]>
```
paul0403, line 43: It's ok to have a channel type without a field for the hardware channel ID? Not even a `?` as placeholder?

mlir/lib/RTIO/IR/RTIODialect.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+    // 3. Parse channelId: `?` or non-negative integer
+    if (succeeded(parser.parseOptionalQuestion())) {
+        channelId = parser.getBuilder().getI64IntegerAttr(-1);
+        return success();
+    }
+
+    int64_t id;
+    if (failed(parser.parseInteger(id)))
+        return failure();
```
paul0403: We put brackets for single statement scopes too

https://github.com/PennyLaneAI/guidance-docs/blob/master/development/cpp-dev-guidelines.md#brace-style

mlir/lib/RTIO/IR/RTIODialect.cpp
paul0403: Is this error why you had to set `useDefaultAttributePrinterParser = 0` (and hence needed to implement the parsing rules `print/parseChannelTypeBody` ?

paul0403: Actually is it the other way around? Did you want to use a custom parser, so you had to set  `useDefaultAttributePrinterParser = 0`, which required the two boilerplate functions `parseAttribute` here, on top of the actual custom parser `parseChannelTypeBody`?

If this is the case, why not just use the default parser? What was the issue there?

paul0403 (commented):
Nice, left some comments on the general dialect organization. 

I'll leave the actual contents of the dialect to you and Mehrdad since you guys are more familiar with it.

mlir/test/RTIO/VerifierTest.mlir
rniczh: Yes, both will be removed, but only after canonicalization? 

`rtio.now` will return the current time cursor, it doesn't has any side effect, and can be eliminated if there is no user to use it. And `rtio.empty` is a virtual event, just for initialization purpose, no need to keep them if no one need to use it.

mlir/test/RTIO/VerifierTest.mlir
rniczh: Yes, `?` just for pretty printing, like `tensor`, `memref`, ... etc, but also can present as nothing. So both `!rtio.channel<"dds", [0, "t0"]>` and `!rtio.channel<"dds", [0, "t0"], ?>` are the same thing

mlir/lib/RTIO/IR/RTIODialect.cpp
rniczh: Done https://github.com/PennyLaneAI/catalyst/pull/2185/commits/6d57469419acb8d292874f02644655d9fb857878

mlir/include/RTIO/IR/RTIODialect.td
rniczh: Done https://github.com/PennyLaneAI/catalyst/pull/2185/commits/bd1c4c4c1ec1ce4116349c00b7ace09abe84f35f

mlir/include/RTIO/IR/RTIODialect.td
rniczh: Done https://github.com/PennyLaneAI/catalyst/pull/2185/commits/c4a3afa2f8ae640d26e43a232101ba2487432ca3

mlir/include/RTIO/IR/RTIOOps.td
rniczh: Now the timeline-based IR has been dropped from the RTIO dialect, we just convert it to LLVM IR at the end to remove the redundancy

mlir/include/RTIO/IR/RTIODialect.td
rniczh: I update the comment, now it only contain the event-based IR.

mlir/lib/RTIO/IR/RTIODialect.cpp
rniczh: Since we need the additional attribute to import the `device_db` to store the metadata for compiling. So now:
1. We set the `useDefaultAttributePrinterParser` to 1
2. Remove custom attribute parsing/printing declaration

mlir/include/RTIO/IR/RTIOOps.td
```diff
@@ -0,0 +1,200 @@
...
+    let summary = "Map a qubit to an RTIO channel";
+    let description = [{
+        It's a temporary operation that will be lowered to a static `rtio.channel` operation during
+        the channel resolution stage. The purpose of this operation is to allow a qubit from a
+        high-level dialect to be mapped to an `!rtio.channel`.
+
+        Example:
+        ```mlir
+        // Map ion qubit to dds channel with dynamic channel id
+        %ch = rtio.qubit_tos_channel %ion_qubit : !ion.qubit -> !rtio.channel<"dds", ?>
```
mehrdad2m:
```suggestion
        %ch = rtio.qubit_to_channel %ion_qubit : !ion.qubit -> !rtio.channel<"dds", ?>
```

mlir/test/RTIO/VerifierTest.mlir
mehrdad2m, line 1: This file is only testing the parsing and not the printing. Could we add some filecheck lines to test the printer as well? Especially for the `rtio.channel` as it has custom parse/print methods.

mlir/lib/RTIO/IR/RTIODialect.cpp
```diff
@@ -0,0 +1,152 @@
...
+                printer << id;
+            }
+        }
+    }
+    else {
+        if (qualifiers && !qualifiers.empty()) {
+            printer << ", ";
+        }
+        printer << "?";
+    }
```
mehrdad2m, line 118:
```suggestion
    if (channelId) {
        int64_t id = channelId.getInt();
        printer << ", ";
        if (id >= 0) {
            printer << id;
        }
        else {
            printer << "?";
        }
    }
    else if (qualifiers && !qualifiers.empty()) {
        printer << ", ?";
    }
```
The logic here seems a bit complicated and I couldn't understand the reasoning behind it. How about this suggestion? Or alternatively if you could add some comments that would be great.

rniczh: Thanks! I think it's because I modify this part multiple times and didn't clean those redundant cond

mlir/test/RTIO/VerifierTest.mlir
rniczh: Added https://github.com/PennyLaneAI/catalyst/pull/2185/commits/e6d0c7c65829d13dc8e7a2aca40909bc6484eaa5

PR #2186, opened 2025-11-11
Improve intermediate IR file management and add support for unified compiler   

mehrdad2m:
**Context:**
We need to enhance Catalyst's intermediate IR file management system to support the unified compiler path that integrates xDSL passes. 

1. New `changed` level for `keep_intermediate`: Adds a new option to save IR files only when the IR is actually modified by a pass, reducing unnecessary file generation.

2. Enhances IR file organization by creating subdirectories for each compilation stage to better organize intermediate files

3. We want the initial unoptimized IR to be the same regardless of whether we are using xdsl passes or not. Therefore we need to dump the initial IR before any xdsl pass is applied. Printing the IR from Python will result in the generic format IR since the python context does not have all the dialects registered. To resolve this, We run `quantum-opt` on the IR before saving to ensure consistent, pretty format output. This introduces an extra call to `quantum_opt` but only when `keep_intermediate` is True

4. Adds callback to save IR after each xDSL passes. Note that an another [PR](https://github.com/PennyLaneAI/catalyst/pull/1978) adds the support on the c++ front

5. Pipeline stage renaming/shuffling: Moved all the quantum compilation passes up to the first stage so all of the quantum-level passes are applied at the same stage. The only exception is `lower-gradients` which currently uses lower level ops. We also renamed the first compilation stage from `enforce-runtime-invariants-pipeline` to `quantum-compilation-pipeline` and the old `quantum-compilation-pipeline` to `gradient-lowering-pipeline`. 

**Benefits:**
Better IR file management. 
E.g. When using `keep_intermediate="changed"` or `keep_intermediate="pass"` with xDSL passes, the output directory structure looks like:

```
workspace/
├── 0_workflow.mlir                           # Initial IR before any transformations
├── 1_AfterQuantumCompilationPass.mlir             # Stage-level summary
├── 2_AfterHLOLoweringPass.mlir               # Stage-level summary
├── 3_AfterGradientLoweringPass.mlir        # Stage-level summary
├── 4_AfterBufferizationPass.mlir            # Stage-level summary
├── 5_AfterMLIRToLLVMDialect.mlir            # Stage-level summary
├── 6_llvm_ir.ll                         # Final LLVM IR
├── workflow.so                                # Compiled shared object
├── workflow.o                                 # Object file
│
├── 0_QuantumCompilationPass/                 # Pass-level details for UserTransformPass (Python interpreter)
│   ├── 1_remove-chained-self-inverse.mlir    
│   ├── 2_xdsl-merge-rotations.mlir          
..
│
├── 1_QuantumCompilationPass/                #  User pass will be shown here if no xdsl pass is provided (Python interpreter)
│   └── 1_InlineNestedModulePass.mlir  # Plus other passes that are in this stage
│
├── 2_HLOLoweringPass/                         
│   ├── 1_StablehloLegalizeToLinalgPass.mlir
...
│
├── 4_BufferizationPass/                       
│   ├── 1_OneShotBufferizePass.mlir
...
│
└── 5_MLIRToLLVMDialect/                      
    ├── 1_SCFToControlFlowPass.mlir
...
```

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-103720]

### DIFF
--- a/.dep-versions
+++ b/.dep-versions
@@ -12,7 +12,7 @@ enzyme=v0.0.203
 # 'doc/requirements.txt'
 # TODO: uncomment and update to latest version of pennylane
 #       after https://github.com/PennyLaneAI/pennylane/pull/8525 is merged.
-# pennylane=0.44.0.dev31
+pennylane @ git+https://github.com/PennyLaneAI/pennylane@bump-jax-api-hashability
 
 # For a custom LQ/LK version, update the package version here and at
 # 'doc/requirements.txt'
--- a/.github/workflows/build-wheel-linux-arm64.yaml
+++ b/.github/workflows/build-wheel-linux-arm64.yaml
@@ -478,6 +478,11 @@ jobs:
       run: |
         python${{ matrix.python_version }} -m pip install oqc-qcaas-client
 
+    - name: Install PennyLane
+      run: |
+        # Install the specific pennylane version required
+        python${{ matrix.python_version }} -m pip install git+https://github.com/PennyLaneAI/pennylane@bump-jax-api-hashability
+
     - name: Install Catalyst
       run: |
         python${{ matrix.python_version }} -m pip install dist/*.whl --extra-index-url https://test.pypi.org/simple
--- a/.github/workflows/build-wheel-linux-x86_64.yaml
+++ b/.github/workflows/build-wheel-linux-x86_64.yaml
@@ -492,6 +492,11 @@ jobs:
         python${{ matrix.python_version }} -m pip install PennyLane-Lightning-Kokkos
         python${{ matrix.python_version }} -m pip install 'amazon-braket-pennylane-plugin>=1.31.0'
 
+    - name: Install PennyLane
+      run: |
+        # Install the specific pennylane version required
+        python${{ matrix.python_version }} -m pip install git+https://github.com/PennyLaneAI/pennylane@bump-jax-api-hashability
+
     - name: Install OQC client
       if: matrix.python_version != '3.13'
       run: |
--- a/.github/workflows/build-wheel-macos-arm64.yaml
+++ b/.github/workflows/build-wheel-macos-arm64.yaml
@@ -473,6 +473,11 @@ jobs:
         python${{ matrix.python_version }} -m pip install PennyLane-Lightning-Kokkos
         python${{ matrix.python_version }} -m pip install 'amazon-braket-pennylane-plugin>1.27.1'
 
+    - name: Install PennyLane
+      run: |
+        # Install the specific pennylane version required
+        python${{ matrix.python_version }} -m pip install git+https://github.com/PennyLaneAI/pennylane@bump-jax-api-hashability
+
     - name: Install OQC client
       if: matrix.python_version != '3.13'
       run: |
--- a/Makefile
+++ b/Makefile
@@ -117,7 +117,7 @@ frontend:
 	# versions of a package with the same version tag (e.g. 0.38-dev0).
 	$(PYTHON) -m pip uninstall -y pennylane
 	$(PYTHON) -m pip install -e . --extra-index-url https://test.pypi.org/simple $(PIP_VERBOSE_FLAG)
-    # TODO: remove after https://github.com/PennyLaneAI/pennylane/pull/8525 is merged.
+	# TODO: remove after https://github.com/PennyLaneAI/pennylane/pull/8525 is merged.
 	$(PYTHON) -m pip install git+https://github.com/PennyLaneAI/pennylane@bump-jax-api-hashability
 	rm -r frontend/pennylane_catalyst.egg-info
 
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -77,6 +77,10 @@
 * Intermediate IR files are now organized into subdirectories for each compilation stage
   when using ``keep_intermediate="changed"`` or ``keep_intermediate="pass"``.
   [(#2186)](https://github.com/PennyLaneAI/catalyst/pull/2186)
+
+* Resource tracking now tracks calls to `SetState` and `SetBasisState`, and can report results
+  that include `qml.StatePrep` operations.
+  [(#2230)](https://github.com/PennyLaneAI/catalyst/pull/2230)
   
 * Remove the hardcoded list of runtime operations in the frontend.
   This will allow arbitrary PL gates to be represented without hyperparameters in MLIR.
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -33,4 +33,4 @@ lxml_html_clean
 --extra-index-url https://test.pypi.org/simple/
 pennylane-lightning-kokkos==0.44.0-dev11
 pennylane-lightning==0.44.0-dev11
-pennylane==0.44.0.dev31
+git+https://github.com/PennyLaneAI/pennylane@bump-jax-api-hashability
--- a/frontend/catalyst/_version.py
+++ b/frontend/catalyst/_version.py
@@ -16,4 +16,4 @@
 Version number (major.minor.patch[-label])
 """
 
-__version__ = "0.14.0-dev43"
+__version__ = "0.14.0-dev45"
--- a/frontend/test/lit/test_decomposition.py
+++ b/frontend/test/lit/test_decomposition.py
@@ -34,6 +34,27 @@ from catalyst.jax_primitives import decomposition_rule
 # pylint: disable=too-many-lines
 
 
+# Helper to skip tests that fail due to PauliRot type annotation issue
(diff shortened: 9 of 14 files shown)


frontend/catalyst/compiler.py
```diff
@@ -349,9 +350,9 @@ def _options_to_cli_flags(options):
     if options.keep_intermediate:
         extra_args += ["--keep-intermediate"]
 
-    if options.keep_intermediate >= KeepIntermediateLevel.PASS:
-        extra_args += ["--save-ir-after-each=pass"]
-        extra_args += ["--dump-module-scope"]
+    if options.keep_intermediate >= KeepIntermediateLevel.CHANGED:
+        level = "pass" if options.keep_intermediate >= KeepIntermediateLevel.PASS else "changed"
+        extra_args += [f"--save-ir-after-each={level}", "--dump-module-scope"]
```
paul0403: Why the `>= KeepIntermediateLevel.PASS`? That is the max level right? :thinking: 

I think it would be more readable to just do 
```python
match options.keep_intermediate:
   case PASS:
      level = "pass"
   case CHANGED:
      level = "changed"
```

mlir/include/Driver/CompilerDriver.h
```diff
@@ -94,16 +94,35 @@ struct CompilerOutput {
     std::string outIR;
     std::string diagnosticMessages;
     PipelineOutputs pipelineOutputs;
-    size_t pipelineCounter = 0;
+    size_t globalPipelineCounter = 0; // Counter for root-level pipeline summary files
+    size_t passCounter = 0;           // Counter for passes within a pipeline folder
+    std::string currentStage = ".";   // Current compilation stage subdirectory
```
paul0403, line 99: Is this variable necessary? :sweat_smile: I feel like having the `"."` literal inlined in the `std::filesystem` lines is more readable, but up to you

frontend/catalyst/compiler.py
```diff
@@ -571,6 +572,46 @@ def is_using_python_compiler(self, mlir_module=None):
...
+            workspace: The workspace directory path
+
+        Returns:
+            Callable or None: The callback function if intermediate saving is enabled, None otherwise
+        """
+        if not (workspace and self.options.keep_intermediate >= KeepIntermediateLevel.CHANGED):
+            return None
+
+        # pylint: disable-next=import-outside-toplevel
+        from xdsl.printer import Printer
```
paul0403, line 603: Can we update the name of this function to reflect that it is xdsl path only?

frontend/catalyst/compiler.py
```diff
@@ -571,6 +572,46 @@ def is_using_python_compiler(self, mlir_module=None):
...
+
+        user_transform_dir = os.path.join(str(workspace), "0_UserTransformPass")
+        os.makedirs(user_transform_dir, exist_ok=True)
+
+        pass_counter = 1
+
+        def save_pass_ir(previous_pass, module, _next_pass=None, **_kwargs):
+            """Callback to save IR after each pass in python_compiler."""
+            nonlocal pass_counter
```
paul0403: `nonlocal` always scares me, especially for these factory functions. Do we have a way to avoid that? :thinking:

mlir/include/Driver/CompilerDriver.h
```diff
@@ -94,16 +94,35 @@ struct CompilerOutput {
...
+        return std::filesystem::path(currentStage) /
+               std::filesystem::path(std::to_string(this->passCounter++) + "_" + pipelineName)
+                   .replace_extension(ext);
+    };
+
+    // Gets the root-level pipeline summary file name
+    std::string nextPipelineSummaryFilename(std::string pipelineName, std::string ext = ".mlir")
+    {
+        return std::filesystem::path(std::to_string(this->globalPipelineCounter) + "_After" +
+                                     pipelineName)
```
paul0403, line 115: Clarifying that the intermediates are always "after" is definitely a nice thing to have! :+1: 

Just one thing though, I think the final llvm ir should still be called "6_llvm_ir.ll", not "6_after_llvm_ir.ll", that sounds pretty weird :sweat_smile:

requirements.txt
```diff
@@ -35,3 +35,4 @@ nbmake
 pennylane-lightning-kokkos
 amazon-braket-pennylane-plugin>1.27.1
 xdsl
+xdsl-jax @ git+https://github.com/xdslproject/xdsl-jax.git@main
```
mlxd: https://pypi.org/project/xdsl-jax/

Any reason we cannot use the above?

paul0403: I think this line was added when they weren't on pypi

paul0403: https://github.com/PennyLaneAI/catalyst/pull/2193

mlir/include/Driver/CompilerDriver.h
mehrdad2m: Well it's mutable. It get updated during lowering by `setStage()`. The "." is just the initial "no stage set yet" value but becomes things like "1_PipelineName" as we progress through pipelines.

frontend/catalyst/compiler.py
mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/2186/commits/5aa77796e9a2ceb8c9fd2d5fad4c73b550b7edc2

frontend/catalyst/compiler.py
mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/2186/commits/11edae72b7346e222271d3fc5db8c91c73fff250

mlir/include/Driver/CompilerDriver.h
mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/2186/commits/70e196a072fcfae8cd264eb94e562e5ed6363075

frontend/catalyst/compiler.py
mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/2186/commits/e5d5da1666036c22f876655c3279c202690e4131

mehrdad2m: yes that's the max. For some reason I thought we are supporting larger numbers to fall back to `KeepIntermediateLevel.PASS` like `keep_intermeidate = 5`

frontend/catalyst/compiler.py
```diff
@@ -582,6 +587,52 @@ def is_using_python_compiler(self, mlir_module=None):
 
         return True
 
+    def _create_xdsl_pass_save_callback(self, workspace):
```
mudit2812, line 590: Does it make sense to decorate this function with `@functools.lru_cache`? Seems to me like we may call it multiple times with the same `workspace`. Let me know if my assumption is incorrect.

mudit2812 (commented):
Looks good to me. You have a soft approval from me, but I think the real approvals should come from the compilation team.

frontend/catalyst/compiler.py
mehrdad2m: We are calling `_create_xdsl_pass_save_callback` only once. What is being called more is the `__call__` method of `SavePassIRCallback`. but since it uses and modifies an outer scope variable (the `counter`) I am not sure if caching it would work 🤔

frontend/catalyst/pipelines.py
```diff
@@ -222,8 +227,12 @@ def get_enforce_runtime_invariants_stage(_options: CompileOptions) -> List[str]:
         # But qnodes targeting other backends may choose to lower
         # this into something else.
         "inline-nested-module",
+        "annotate-function",
```
dime10: I think this new structure looks good! The only thing I was wondering about was this pass, I think it might for gradients but I'm not sure. I couldn't find out yet how the annotations are being used, from what I remember it was for verification (e.g. not differentiating an mcm)

mehrdad2m: Yeah you are right. This pass basically annotate functions that can't have gradients with `hasInvalidGradientOp` attribute and then runs a verifier right after it. Then the verification methods in the `gradient` dialect detect the attribute and raise an error. The name of the pass is definitely very confusing :) 
I think it could also live next to the `lower-gradients` pass. I just moved it to `QuantumCompilation` stage cause I thought we will move everything there eventually. But maybe it make sense to leave it in the `GradientLowering` stage for now to keep the semantic consistency of the stage names?

dime10: Yeah I think keeping the gradients in their own stage is not necessarily a bad thing :)

So we can update the annotate-pass name and move it there

mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/2186/commits/58244ba0d9516844e9fca901a071ce07b3f5a91f

PR #2183, opened 2025-11-10
Add prefix to the xdsl related discardable attributes

mehrdad2m:
**Context:**
MLIR requires discardable attributes to be dialect-prefixed. 

**Description of the Change:**
- Add prefix to the xdsl related attributes 
- Add lit tests to check for the strings in the lowered MLIR

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

PR #2187, opened 2025-11-11
Fix missing dialect prefix in xdsl detection

mehrdad2m:
**Context:**
dialect prefix was missed when fixing the bug in https://github.com/PennyLaneAI/catalyst/pull/2183

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


lazypanda10117 (approved):
Thanks for the quick fix!

PR #2188, opened 2025-11-11
Add a Pauli frame tracking dialect

joeycarter:
**Context:** A *Pauli frame tracker* [1] is a system that tracks gates from the Pauli group in classical electronics instead of physically applying them to qubits on the device. Doing so reduces the number of quantum operations applied to the system, thus reducing the probability of errors. This PR adds a Pauli frame tracking dialect to Catalyst and includes a set of abstractions and operations for interacting with an external Pauli frame tracking library.

[1] Knill, E. *Quantum computing with realistically noisy devices*. Nature **434**, 39-44 (2005). https://doi.org/10.1038/nature03350.

**Description of the Change:** The `PauliFrame` dialect includes the following operations:

* `pauli_frame.init`, which initializes the Pauli record(s) of the target qubit(s) to *I*.
* `pauli_frame.init_qreg`, same as `pauli_frame.init` except it initializes the Pauli records of all the qubits in the target quantum register.
* `pauli_frame.read`, which reads the Pauli record of the target qubit.
* `pauli_frame.update`, which updates the Pauli record(s) of the target qubit(s) given a new Pauli record.
* `pauli_frame.update_with_clifford`, which updates the Pauli record(s) of the target qubit(s) given the conjugation relation between the given Clifford gate and the current Pauli record.
* `pauli_frame.correct_measurement`, which corrects a measurement result given the Pauli record of the target qubit.
* `pauli_frame.flush`, which flush the Pauli record(s) of the target qubit(s) (*flushing* means physically applying the Pauli gates in the record and re-initializing).
* `pauli_frame.set`, an extra op that sets the Pauli record(s) of the target qubit(s).

Note that this PR only adds the dialect itself. Future PRs will add compilation passes to insert these ops into the quantum program IR and to lower them to runtime stubs that call into a (thus far) mocked out Pauli frame tracking runtime library.

[sc-103530]

### DIFF
--- a/frontend/catalyst/_version.py
+++ b/frontend/catalyst/_version.py
@@ -16,4 +16,4 @@
 Version number (major.minor.patch[-label])
 """
 
-__version__ = "0.14.0-dev48"
+__version__ = "0.14.0-dev49"


mlir/include/PauliFrame/IR/PauliFrameDialect.td
```diff
@@ -0,0 +1,115 @@
...
+        interacting with an external Pauli frame tracking library. A *Pauli frame tracker* [1] is a
+        system that tracks gates from the Pauli group in classical electronics instead of applying
+        them to qubits on the device. Doing so reduces the number of quantum operations applied to
+        the system, thus reducing the probability of errors.
+
+        A single Pauli record *Rq* tracks all the Pauli gates that are applied on qubit *q*. Every
+        set of tracked Pauli gates can be reduced to one of the elements in the set {I, X, Z, XZ},
+        hence a single Pauli record can be stored using two bits, encoded as X- and Z-parity bits:
+
+            I = (0, 0),   X = (1, 0),   Z = (0, 1),   XZ = (1, 1)
```
multiphaseCFD, line 38: yep, two bits works fine to record the `X`,`Z` by-product corrections. While, there is a scenario that `Pauli Record` has to combined with `PauliY`,  which would introduce a phase info. Not sure how the phase info would impact the final results for now. But if we would like to add phase info to the record, we might need another two bits to represent the corresponding phase info, i.e., (`1`,`-1`, `i`, `-i`) .

mlir/include/PauliFrame/IR/PauliFrameDialect.td
```diff
@@ -0,0 +1,115 @@
...
+        [1] Knill, E. Quantum computing with realistically noisy devices. Nature 434, 39-44 (2005).
+        https://doi.org/10.1038/nature03350.
+
+        [2] Riesebos, L., et al. Pauli Frames for Quantum Computer Architectures.
+        DAC '17: Proceedings of the 54th Annual Design Automation Conference 76, 1-6 (2017).
+        https://doi.org/10.1145/3061639.3062300.
+
+        > [!IMPORTANT]
+        > The pauli_frame dialect is experimental and will not maintain API stability between releases.
+        > Use at your own risk.
```
multiphaseCFD, line 74: 💯

mlir/lib/PauliFrame/IR/PauliFrameOps.cpp
```diff
@@ -0,0 +1,76 @@
...
+#include "PauliFrame/IR/PauliFrameOps.h"
+
+using namespace mlir;
+using namespace catalyst::pauli_frame;
+
+//===----------------------------------------------------------------------===//
+// PauliFrame op definitions.
+//===----------------------------------------------------------------------===//
+
+#define GET_OP_CLASSES
```
multiphaseCFD, line 29: Just wondering what this macro stands for?

mlir/include/PauliFrame/IR/PauliFrameOps.td
```diff
@@ -0,0 +1,183 @@
...
+// PauliFrame dialect attributes.
+//===----------------------------------------------------------------------===//
+
+
+//===----------------------------------------------------------------------===//
+// PauliFrame dialect operations.
+//===----------------------------------------------------------------------===//
+
+def InitOp : PauliFrame_Op<"init"> {
+    let summary = "Initialize the Pauli record of the target qubit(s) to I";
```
multiphaseCFD: Do we want to accept a `qreg` and initialize all qubits in the container as `I` as well?

mlir/include/PauliFrame/IR/PauliFrameOps.td
```diff
@@ -0,0 +1,183 @@
...
+
+def UpdateOp : PauliFrame_Op<"update"> {
+    let summary = "Update the Pauli record of the target qubit(s)";
+    let description = [{
+        Update the Pauli record of the target qubit(s) given the new Pauli record. The Pauli record
+        is encoded by its X- and Z-parity bits. If multiple target qubits are given as input, the
+        new Pauli record is used to update the Pauli records of all the given qubits.
+    }];
+
+    let arguments = (ins
```
multiphaseCFD, line 126: Why do we need `ins` here for args and `out` for the  results? Is it a syntax rule for the `.td`?

mlir/include/PauliFrame/IR/PauliFrameOps.td
```diff
@@ -0,0 +1,183 @@
...
+
+    let results = (outs
+        Variadic<QubitType>:$out_qubits
+    );
+
+    let assemblyFormat = [{
+        $in_qubits attr-dict `:` type(results)
+    }];
+
+    let hasVerifier = 1;
```
multiphaseCFD, line 163: What is `hasVerifier` for?

multiphaseCFD: Looks like we need to define a `verify()` method for those ops set `hasVerifier` as `1`.

mlir/lib/PauliFrame/IR/PauliFrameOps.cpp
```diff
@@ -0,0 +1,76 @@
...
+#include "PauliFrame/IR/PauliFrameOps.cpp.inc"
+
+//===----------------------------------------------------------------------===//
+// PauliFrame op verifiers.
+//===----------------------------------------------------------------------===//
+
+LogicalResult InitOp::verify()
+{
+    if (getInQubits().size() == 0) {
```
multiphaseCFD: A bit confused to me. Where does the `getInQubits()` come from?

dime10 (commented):
Love it! 😍 

The one thing I wonder is whether the pauli frame ops should generate new qubits 🤔 After all, they don't write to the qubits (they read/write from pauli frame which is a hidden state (no value for it in the ir)?), the only exception is the flush op so there it makes sense.

mlir/lib/PauliFrame/IR/PauliFrameOps.cpp
dime10: It's MLIR internal stuff, this will make it such that the C++ classes for dialect operations generated from the tablegen file are included in the subsequent include call (that you way you can selectively include different elements from a single file).

mlir/include/PauliFrame/IR/PauliFrameOps.td
dime10: It generates the declaration for the verify method (you provide the definition).

mlir/lib/PauliFrame/IR/PauliFrameOps.cpp
dime10: All named inputs/outputs from the tablegen declaration generate `get*` methods, where the name is normalized to CamelCase (edit: brainfart).

mlir/include/PauliFrame/IR/PauliFrameOps.td
joeycarter: Yes this is just TableGen syntax. It's not particularly well documented, but there are some examples here: https://llvm.org/docs/TableGen/

mlir/include/PauliFrame/IR/PauliFrameOps.td
joeycarter: :100: That might prove useful, I'll add it.

mlir/include/PauliFrame/IR/PauliFrameOps.td
joeycarter: Exactly that :+1:

mlir/lib/PauliFrame/IR/PauliFrameOps.cpp
joeycarter: I think you mean CamelCase :wink: 

But yes, that's exactly what's happening here, these methods are auto-generated by the MLIR TableGen infrastructure.

mlir/include/PauliFrame/IR/PauliFrameDialect.td
joeycarter: That's a fair point. The Riesebos paper didn't mention this and operates using only two bits per record. It's not clear if that's because they don't consider Y gates in the Pauli record or because they just ignore phase information. Let me think more on this.

mlir/lib/PauliFrame/IR/PauliFrameOps.cpp
multiphaseCFD: > All named inputs/outputs from the tablegen declaration generate get* methods,

Thanks @dime10 !

mlir/include/PauliFrame/IR/PauliFrameOps.td
joeycarter: e61ffc634

joeycarter:
> > The reason I opted for this design was to easily replace Pauli gates with calls to the Pauli frame tracker.
> 
> Replacing (or erasing) a quantum gate is actually quite easy, you would just call reaplaceAllUses on the output qubits with the input qubits.
> 
> > We'll also need to consider commutation relations at some point, so returning an updated qubit value helps keep track of the operation order (although we're still working out the mechanics of how this will work).
> 
> That's a good point, if the pauli frame op consumes the same qubit value as another gate, there is no defined ordering between the two (at least on the SSA graph), the ordering would only be defined when considering the absolute ordering in the block's op list. MLIR passes won't mess with this ordering if the op is effectful, but our transforms might if we don't pay attention. In this case your current design might be safer then 👍

Yes I agree, it might not be strictly necessary to return the qubit, but it's safer for the time being, given our current knowledge of the system. In that case I propose we leave it as.

mlir/include/PauliFrame/IR/PauliFrameDialect.td
joeycarter: I was thinking about this more, and my sense is that given that the purpose of the Pauli tracker is largely to correct measurement outcomes, which are not affected by phase information, we don't need to track the phase. Certainly for simulators we would want the phase to be correct, but for the purposes of execution on hardware I think it's safe to ignore. What do you think?

multiphaseCFD: I agree, this seems reasonable. Not a blocker, but it would be safer to ask opinions from the algo team.

joeycarter: This is probably more a question for the HW/Architecture teams. Let's run with this as is since this is the formulation found in the literature, and since I think it's suitable for prototyping. We can always change it if need be.

mlir/include/PauliFrame/IR/PauliFrameOps.td
```diff
@@ -0,0 +1,242 @@
...
+    let summary = "Update the Pauli record(s) of the target qubit(s)";
+    let description = [{
+        Update the Pauli record(s) of the target qubit(s) given the new Pauli record. The Pauli
+        record is encoded by its X- and Z-parity bits. If multiple target qubits are given as input,
+        the new Pauli record is used to update the Pauli records of all the given qubits.
+    }];
+
+    let arguments = (ins
+        I1:$x_parity,
+        I1:$z_parity,
```
joeycarter: TBD: Should the x-/z-parity bit inputs be operands or attributes? (i.e. should we allow them to be specified at runtime, in which case they should be operands, or will they always be known at compile time, in which case they should be attributes). Making it generic to allow both causes issues having a combination of optional operands and variadic qubit inputs.

mlir/include/PauliFrame/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+add_subdirectory(IR)
+# add_subdirectory(Transforms)
```
mlxd: Anything to be done here?

joeycarter: Nope. I can remove it, but we'll have to add it back in again and uncomment as soon as we start adding compilation passes (in a future PR).

mlir/include/PauliFrame/IR/PauliFrameOps.td
lillian542: Discussed offline a bit, but just noting that we anticipate that registering the MBQC byproduct corrections on the pauli tracker will be conditional on the MCM results we get during runtime, so it seems useful to have them as operands. 

E.g. in mock MLIR, things like

```
%m1, %1 = quantum.measure_x %0
%m2, %2 = quantum.measure_x %1
%m3, %3 = quantum.measure_y %2
%m4, %4 = quantum.measure_x %3
%x = %m1 ^ %m3
%z = %m2 ^ %m4
%7 = pauli_tracker.update([%x, %z])
```

Then again, an alternative is to conditionally call updates, i.e. instead of `%7 = pauli_tracker.update([%x, %z])`, we have

```
%7 = scf.if %x -> !quantum.bit {
    %out_q = pauli_frame.update [1, 0] %4 : !quantum.bit
    scf.yield %out_q : !quantum.bit
} else {
    scf.yield %4 : !quantum.bit
}
%8 = scf.if %z -> !quantum.bit {
    %out_q  = pauli_frame.update [1, 0] %7 :  !quantum.bit
    scf.yield %out_q : !quantum.bit
} else {
    scf.yield %7 : !quantum.bit
}
```
Then I guess we can use attributes.

_Or_ we could pass the additional XOR responsibility for determining MBQC byproduct corrections on to a specialized MBQC pauli tracker, and have some separate `pauli_frame.register_mbqc_byproduct` that takes measurements (runtime values) and an operation name (known at compile time).

I don't have much of a preference between these with the current information 😅

joeycarter: I'm leaning now towards the second option; that is, to have the x-/z-parity bit inputs as _attributes_ of the `update` op. I think we will always know what these bits ought to be at compile time, since they mainly (exclusively?) get dropped in place of Pauli gates already in the IR.

Moreover, this structure is already in place for the byproduct ops after the `convert_to_mbqc` pass:

```mlir
%0 = scf.if %x -> (!quantum.bit) {
  %1 = quantum.custom "PauliX"() %q : !quantum.bit
  scf.yield %1 : !quantum.bit
} else {
  scf.yield %q : !quantum.bit
}
%1 = scf.if %z -> (!quantum.bit) {
  %2 = quantum.custom "PauliZ"() %0 : !quantum.bit
  scf.yield %2 : !quantum.bit
} else {
  scf.yield %0 : !quantum.bit
}
```

so it should just be a matter of replacing the `quantum.custom` ops for the Pauli gates with the corresponding `pauli_frame.update` ops.

joeycarter: 6f96ee2d8

mlir/include/PauliFrame/CMakeLists.txt
mlxd: Sounds good --- let's pull this out, and re-add later (there may be another one too, if not mistaken)

joeycarter: Done 091594235

mlir/include/PauliFrame/IR/PauliFrameOps.td
```diff
@@ -0,0 +1,242 @@
...
+    );
+
+    let assemblyFormat = [{
+        `[` $clifford_gate `]` $in_qubits attr-dict `:` type(results)
+    }];
+
+    let hasVerifier = 1;
+}
+
+def CorrectMeasurementOp : PauliFrame_Op<"correct_measurement"> {
```
dime10, line 166: How does this op work, do we not need knowledge of the measurement basis as well as the pauli frame to correct the output bit?

mlir/include/PauliFrame/IR/PauliFrameOps.td
```diff
@@ -0,0 +1,242 @@
...
+def FlushOp : PauliFrame_Op<"flush"> {
+    let summary = "Flush the Pauli record of the target qubit";
+    let description = [{
+        Flush the Pauli record of the target qubit. *Flushing* the Pauli record of a qubit refers to
+        physically applying the Pauli gates stored in the record on that qubit and then resetting
+        its record to I, as part of the Pauli frame tracking protocol for non-Clifford gates.
+
+        This op does not physically apply the Pauli gates, since these operations must be specified
+        explicitly in the IR. Instead, it returns the X- and Z- parity bits to conditionally apply
+        these gates based on the returned values.
```
dime10, line 196: So `flush` is just a `read`+`init` right?

mlir/include/PauliFrame/IR/PauliFrameOps.td
```diff
@@ -0,0 +1,242 @@
...
+    let results = (outs
+        QuregType:$out_qreg
+    );
+
+    let assemblyFormat = [{
+        $in_qreg attr-dict `:` type(results)
+    }];
+}
+
+def ReadOp : PauliFrame_Op<"read"> {
```
dime10, line 96: Just for my understanding, what would be the use case of a read?

mlir/include/PauliFrame/IR/PauliFrameOps.td
joeycarter: Yep, exactly that. It combines the two ops into one, meaning we only need one call into the process running the Pauli frame tracker, so in principle it should reduce latency.

mlir/include/PauliFrame/IR/PauliFrameOps.td
joeycarter: At the end of an execution block, we need to read the Pauli frame of the target qubit and physically apply the Pauli gates, conditional on the returned parity bits. For example, if we start with a program

```mlir
%q = ... // a !quantum.bit
%0 = quantum.custom "X" %q : !quantum.bit
%1 = quantum.custom "Y" %0 : !quantum.bit
%2 = quantum.custom "Z" %1 : !quantum.bit
// measurements, etc. ...
```

The `read` op would be used as follows:

```mlir
%q = ... // a !quantum.bit
%0 = pauli_frame.init %q : !quantum.bit                  // init, Rq = (0, 0)
%1 = pauli_frame.update [1, 0] %0 : !quantum.bit         // X,    Rq = (0, 0) ^ (1, 0) = (1, 0)
%2 = pauli_frame.update [1, 1] %1 : !quantum.bit         // Y,    Rq = (1, 0) ^ (1, 1) = (0, 1)
%3 = pauli_frame.update [0, 1] %2 : !quantum.bit         // Z,    Rq = (0, 1) ^ (0, 1) = (0, 0)

// Done, apply gates in Pauli record
%x, %z, %4 = pauli_frame.read %3 : i1, i1, !quantum.bit  // runtime returns %x = 0, %z = 0
%5 = scf.if %x -> !quantum.bit {
    %out_q = quantum.custom "X" %4 : !quantum.bit
    scf.yield %out_q : !quantum.bit
} else {
    scf.yield %4 : !quantum.bit
}
%6 = scf.if %z -> !quantum.bit {
    %out_q = quantum.custom "Z" %5 : !quantum.bit
    scf.yield %out_q : !quantum.bit
} else {
    scf.yield %5 : !quantum.bit
}

// measurements, etc. ...
```

dime10: Wouldn't that be a use case for the flush though? Presumably after applying the gates in the record you'd want the record reset no?

mlir/include/PauliFrame/IR/PauliFrameOps.td
joeycarter: My understanding is that in the "traditional" formulation of the Pauli frame tracker, which we're prototyping here, that we can only correct measurements in the computational (or _Z_) basis in this way. To do measurements in another basis you would have to flush the Pauli records and physically apply the Pauli gates before performing the measurement in another basis.

Granted, this is going to cause headaches for MBQC, but for the purposes of the prototype we're building it's not so important since we can prototype it at the layer above MBQC.

mlir/include/PauliFrame/IR/PauliFrameOps.td
joeycarter: True, in this case we could use a `flush` instead of a `read`. Although, since we're at the end of the execution block, it doesn't matter here since we won't need to read from or write to the Pauli record again, so it saves us a step by not having to re-initialize. This is also still a prototype, so I'd like to have the ability to query the Pauli record without re-initializing for the sake of flexibility as we figure the details out.

mlir/include/PauliFrame/IR/PauliFrameOps.td
dime10: Got it. I guess I'm mostly familiar with it in the PBC setting, where I think you would need the measurement basis (x, y, z) to post-correct the outcome of PPMs (since you're trying to determine whether the pauli frame commutes or anti-commutes with the ppm).

mlir/include/PauliFrame/IR/PauliFrameOps.td
dime10: Sounds good. From the docstrings I just couldn't figure out why you'd need this operation, whereas the set op explicitly stated that it wasn't technically needed 👍

PR #2190, opened 2025-11-12
Add support for xdsl passes to mlir_opt property of qjit

mehrdad2m:
**Context:**
Currently when calling a `circuit.mlir-opt`, we invoke a call to `quantum_opt` which would not apply any xdsl passes if there is any which would cause failure in MLIR parsing since the xdsl pass would not be registered. 

We need to add support for xdsl so it apply them first before calling `quantum_opt`
**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-103772]

### DIFF
--- a/frontend/catalyst/compiler.py
+++ b/frontend/catalyst/compiler.py
@@ -379,8 +379,6 @@ def to_mlir_opt(
     *args, stdin=None, options: Optional[CompileOptions] = None, using_python_compiler=False
 ):
     """echo ${input} | catalyst --tool=opt *args *opts -"""
-    if stdin is None:
-        return None
     # Check if we need to use Python compiler for xDSL passes
     if using_python_compiler:
         # Use Python compiler path for xDSL passes


frontend/catalyst/compiler.py
```diff
@@ -375,8 +375,21 @@ def to_llvmir(*args, stdin=None, options: Optional[CompileOptions] = None):
     return _catalyst(*opts, *args, stdin=stdin)
 
 
-def to_mlir_opt(*args, stdin=None, options: Optional[CompileOptions] = None):
+def to_mlir_opt(
+    *args, stdin=None, options: Optional[CompileOptions] = None, using_python_compiler=False
+):
     """echo ${input} | catalyst --tool=opt *args *opts -"""
+    if stdin is None:
+        return None
```
paul0403: What is this extra branch for?

frontend/catalyst/jit.py
```diff
@@ -582,11 +582,14 @@ def mlir_opt(self):
         """Obtain the MLIR representation after optimization"""
         if not self.mlir_module:
             return None
-
+        using_python_compiler = self.compiler.is_using_python_compiler(self.mlir_module)
         stdin = self.mlir_module.operation.get_asm(
-            enable_debug_info=self.compile_options.use_nameloc
+            print_generic_op_form=using_python_compiler,
```
paul0403, line 587: xdsl can only support generic op form printing now? :thinking:

frontend/catalyst/compiler.py
mehrdad2m: I added it first but after some changes, it is actually redundant now. I forgot to remove it. Thanks for mentioning that

frontend/catalyst/jit.py
mehrdad2m: Actually xdsl can support custom formats. The problem is that a lot of stableHLO ops (specifically the ones with custom print/parse methods in C++) that are added don't have assembly format defined. For this reason, inputing a custom formatted IR would crash xdsl. This would be resolved after we add all the custom assembly implementations but it is not yet prioritized.

frontend/catalyst/compiler.py
mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/2190/commits/da8daafd60c9cac682c8e0113b67ada5ef6fb6f9

PR #2191, opened 2025-11-13
Add a fixture that checks and skip tests if xdsl is not installed 

mehrdad2m:
**Context:**
Pytest tests that use the python compiler should skip if xdsl is not installed since xdsl is not a hard dependency for Catalyst. 
 
**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/frontend/test/pytest/test_jit_behaviour.py
+++ b/frontend/test/pytest/test_jit_behaviour.py
@@ -901,14 +901,11 @@ class TestDefaultAvailableIR:
         assert g.mlir_opt
         assert "__catalyst__qis" in g.mlir_opt
 
-    @pytest.mark.usefixtures("use_capture")
+    @pytest.mark.usefixtures("use_capture", "requires_xdsl")
     def test_mlir_opt_using_xdsl_passes(self, backend):
         """Test mlir opt using xDSL passes."""
-        try:
-            # pylint: disable-next=import-outside-toplevel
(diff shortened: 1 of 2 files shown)


frontend/test/pytest/test_jit_behaviour.py
```diff
@@ -904,8 +904,11 @@ def g(x: float):
     @pytest.mark.usefixtures("use_capture")
     def test_mlir_opt_using_xdsl_passes(self, backend):
         """Test mlir opt using xDSL passes."""
-        # pylint: disable-next=import-outside-toplevel
-        from pennylane.compiler.python_compiler.transforms import iterative_cancel_inverses_pass
+        try:
+            # pylint: disable-next=import-outside-toplevel
+            from pennylane.compiler.python_compiler.transforms import iterative_cancel_inverses_pass
+        except ModuleNotFoundError:
+            pytest.skip("xdsl is not installed, skipping test")
```
paul0403: I feel like it would be a good idea to turn this into a fixture, as I imagine there's will be more tests that use xdsl in the future.

mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/2191/commits/0355743f68fe3e0e0b66187995dede5946f4d531

paul0403 (approved):
Awesome :100:

PR #2192, opened 2025-11-13
Adding PPR canonicalization pass

lazypanda10117:
**Context:** We sometimes have operations like `qec.ppr["I"](0)` in the MLIR. Here, we write a canonicalization pass to remove such PPRs consisting only of "I"s.

### DIFF
--- a/mlir/test/QEC/CanonicalizationTest.mlir
+++ b/mlir/test/QEC/CanonicalizationTest.mlir
@@ -14,17 +14,17 @@
 
 // RUN: quantum-opt --split-input-file -verify-diagnostics --canonicalize %s | FileCheck %s
 
-func.func @test_ppr_canonicalize_single_identity(%q1 : !quantum.bit, %q2 : !quantum.bit) {
+func.func @test_ppr_canonicalize_single_identity(%q1 : !quantum.bit, %q2 : !quantum.bit) -> (!quantum.bit, !quantum.bit) {
     %0 = qec.ppr ["I"](4) %q1 : !quantum.bit
     %1 = qec.ppr ["I"](4) %q2 : !quantum.bit
-    return
+    return %0, %1 : !quantum.bit, !quantum.bit
     // CHECK-NOT: qec.ppr ["I"](4)
 }
 
 // -----
 
-func.func @test_ppr_canonicalize_multiple_identity(%q1 : !quantum.bit, %q2 : !quantum.bit, %q3 : !quantum.bit) {
+func.func @test_ppr_canonicalize_multiple_identity(%q1 : !quantum.bit, %q2 : !quantum.bit, %q3 : !quantum.bit) -> (!quantum.bit, !quantum.bit, !quantum.bit) {
     %out_qubits:3 = qec.ppr ["I", "I", "I"](4) %q1, %q2, %q3: !quantum.bit, !quantum.bit, !quantum.bit
-    return
+    return %out_qubits#0, %out_qubits#1, %out_qubits#2 : !quantum.bit, !quantum.bit, !quantum.bit
     // CHECK-NOT: qec.ppr ["I", "I", "I"](4)
 }


mlir/lib/QEC/IR/QECDialect.cpp
```diff
@@ -131,6 +131,21 @@ LogicalResult FabricateOp::verify()
...
+
+    bool allIdentity = llvm::all_of(pauliProduct, [](mlir::Attribute attr) {
+        auto pauliStr = llvm::cast<mlir::StringAttr>(attr);
+        return pauliStr.getValue() == "I";
+    });
+
+    if (allIdentity) {
+        rewriter.replaceOp(op, op.getInQubits());
+    }
+    return mlir::success();
```
paul0403:
```suggestion
    if (allIdentity) {
        rewriter.replaceOp(op, op.getInQubits());
        return mlir::success();
    }
    return mlir::failure();
```

paul0403: Canonicalization is a pattern rewrite. It should return success when it did some rewrite.

mlir/test/QEC/CanonicalizePPR.mlir
```diff
@@ -0,0 +1,30 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// RUN: quantum-opt --split-input-file -verify-diagnostics --canonicalize %s | FileCheck %s
+
+func.func @test_ppr_canonicalize_single_identity(%q1 : !quantum.bit, %q2 : !quantum.bit) {
+    %0 = qec.ppr ["I"](4) %q1 : !quantum.bit
+    %1 = qec.ppr ["I"](4) %q2 : !quantum.bit
+    return
+    // CHECK-NOT: qec.ppr ["I"](4)
```
paul0403: Let's return the qubit values, `--canonicalize` will do dead code elimination iirc so without a user the ppr ops are removed not from your canonicalization pattern but from DCE.

doc/releases/changelog-dev.md
```diff
@@ -144,6 +144,10 @@
   for the qec dialect.
   [(#2189)](https://github.com/PennyLaneAI/catalyst/pull/2189)
 
+  * Added a canonicalization pass for `qec.ppr` to remove any PPRs consisting only
```
```suggestion
  * Added a canonicalization pattern for `qec.ppr` to remove any PPRs consisting only
```

mlir/test/QEC/CanonicalizePPR.mlir
paul0403: e.g. 
```
paul.wang@L7450-H298SW3:~/catalyst_new/catalyst$ cat hi.mlir 
func.func @hi() {
    %phi = arith.constant 0.3 : f64
    return
}
paul.wang@L7450-H298SW3:~/catalyst_new/catalyst$ ./mlir/build/bin/quantum-opt --canonicalize hi.mlir 
module {
  func.func @hi() {
    return
  }
}
```

mlir/test/QEC/CanonicalizationTest.mlir
```diff
@@ -0,0 +1,30 @@
+// Copyright 2025 Xanadu Quantum Technologies Inc.
```
paul0403, line 1: Also, can we call this file `CanonicalizationTest.mlir` like the others?

mlir/test/QEC/CanonicalizePPR.mlir
paul0403: This by the way is probably also the reason why your test passed even though your canonicalization method returned the opposite thing :sweat_smile: so two errors cancelled out

lazypanda10117: oh no, let me double check this.

lazypanda10117: It's fixed. Thanks for catching this.

mlir/test/QEC/CanonicalizationTest.mlir
lazypanda10117: Done.

PR #2195, opened 2025-11-14
No longer absorb initial StatePrep for null.qubit

jzaia18:
**Context:**
The `null.qubit` device currently removes `StatePrep` operations that happen at the beginning of the circuit. This is intended as a feature to speed up simulators, but `null.qubit` doesn't actually simulate any gates. This can cause issues where the IRs do not match the expected value.

Updates `null.qubit` to no longer skip initial state preps.

**Benefits:**
`null.qubit` behaves in a more straightforward and expected way.

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-104048]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -4,6 +4,9 @@
 
 <h3>Improvements 🛠</h3>
 
+* ``null.qubit`` no longer skips initial `qml.StatePrep` operations.
+  [(#2195)](https://github.com/PennyLaneAI/catalyst/pull/2195)
+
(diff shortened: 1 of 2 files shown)


dime10:
Are you sure it "removes StatePrep at the beginning of a circuit"? Afaik that flag instructs the decomposition to skip prep ops, not remove them from the program (essentially it declares the device supports them natively).

Or are you looking for StatePrep ops to be decomposed into elementary gates when used with null qubit?

jzaia18:
Closing since we've chosen to take a different approach forward

PR #2198, opened 2025-11-17
Fix one-shot doesn't take pass-pipeline

rniczh:
**Context:**

`cancel_inverses` does not apply with `one-shot` enabled

```python
import pennylane as qml
import catalyst
from catalyst.passes import cancel_inverses

@qml.qjit(keep_intermediate='pass')
@cancel_inverses
@qml.qnode(dev, shots=100, mcm_method='single-branch-statistics')
def circ():
    qml.H(1)
    qml.H(1)
    return qml.sample(catalyst.measure(0))

circ()
print(circ.mlir)
```

pass `pass-pipeline` to `one-shot`

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -109,6 +109,10 @@
   // ... use %4
   ```
 
+* The pass pipeline is correctly registered to the transform named sequence of the
+  one-shot qnode when `one-shot` mcm method is used.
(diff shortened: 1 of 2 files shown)


sengthai (approved):
Thanks @rniczh  and @paul0403! 🫡

PR #2203, opened 2025-11-17
Support Dynamic Qubit Allocation for Resource Tracking

jzaia18:
**Context:**
Resource tracking does not currently explicitly support dynamic qubit allocation. This can cause issues when using it in practice.

Adds explicit support for dynamic qubit allocation to resource tracking by
* Tracking the *total* number of allocations made
* Tracking the current number of wires
* Tracking the maximum concurrent number of wires
* Tracking wire depths, even for deallocated wires

**Benefits:**
Resource tracking now supports dynamic qubit allocation, enabling more workflows to gather specs via `qml.specs` with `level="device"`.

**Possible Drawbacks:**
Slightly more complex and higher overhead resource tracking.

**Related GitHub Issues:**
[sc-104151]

### DIFF
--- a/runtime/lib/backend/null_qubit/ResourceTracker.hpp
+++ b/runtime/lib/backend/null_qubit/ResourceTracker.hpp
@@ -64,12 +64,12 @@ struct ResourceTracker final {
                 RT_FAIL_IF(
                     curr_depth == wire_depths_.end(),
                     ("Wire index " + std::to_string(i) + " is not an allocated wire").c_str());
-                    max_depth = std::max(max_depth, curr_depth->second);
-                }
-                for (const auto &i : controlled_wires) {
-                    auto curr_depth = wire_depths_.find(i);
-                    RT_FAIL_IF(curr_depth == wire_depths_.end(),
-                    ("Control wire index " + std::to_string(i) + " is not an allocated wire")
+                max_depth = std::max(max_depth, curr_depth->second);
+            }
+            for (const auto &i : controlled_wires) {
+                auto curr_depth = wire_depths_.find(i);
+                RT_FAIL_IF(curr_depth == wire_depths_.end(),
+                           ("Control wire index " + std::to_string(i) + " is not an allocated wire")
                                .c_str());
                 max_depth = std::max(max_depth, curr_depth->second);
             }


runtime/lib/backend/null_qubit/ResourceTracker.hpp
```diff
@@ -55,18 +59,19 @@ struct ResourceTracker final {
...
-                       ("Wire index " + std::to_string(i) + " exceeds allocated wires").c_str());
-        }
 
-        gate_types_[name]++;
-        gate_sizes_[total_wires]++;
         if (compute_depth_) {
+            std::vector<QubitIdType> combined_wires = {};
+            combined_wires.insert(combined_wires.end(), wires.begin(), wires.end());
+            combined_wires.insert(combined_wires.end(), controlled_wires.begin(),
+                                  controlled_wires.end());
```
joeycarter: Is it possible to avoid copying the elements of `wires` and `controlled_wires` into a new vector? Or can we iterate over the two original vectors separately below?

Or, since we're using C++20, we could be fancy and use [`std::span`](https://en.cppreference.com/w/cpp/container/span.html) and [`std::views::join`](https://en.cppreference.com/w/cpp/ranges/join_view.html), e.g.

```cpp
#include <iostream>
#include <vector>
#include <ranges>
#include <span>
#include <array>

int main() {
    std::vector<int> v1 = {1, 2, 3};
    std::vector<int> v2 = {4, 5, 6};

    // 1. Create spans (lightweight, non-owning views) of the vectors
    // 2. Put them in a temporary array
    // 3. Use views::join to flatten the array of spans into one sequence
    auto combined_view = std::array{std::span{v1}, std::span{v2}} 
                       | std::views::join;

    for (const auto& elem : combined_view) {
        std::cout << elem << " ";
    }

    return 0;
}
```

Output:

```
1 2 3 4 5 6 
```

runtime/lib/backend/null_qubit/NullQubit.hpp
```diff
@@ -100,11 +100,12 @@ struct NullQubit final : public Catalyst::Runtime::QuantumDevice {
...
     {
-        num_qubits_++; // next_id
+        QubitIdType new_qubit = this->qubit_manager.Allocate(num_qubits_);
         if (this->track_resources_) {
-            this->resource_tracker_.SetMaxWires(static_cast<std::size_t>(num_qubits_));
+            this->resource_tracker_.AllocateQubit(new_qubit);
         }
-        return this->qubit_manager.Allocate(num_qubits_);
+        num_qubits_++; // next_id
+        return new_qubit;
```
dime10, line 108: Isn't this a change in behaviour? `qubit_manager.Allocate` was called with the incremented `num_qubits_`, now it's without the increment

jzaia18: Correct. It turns out we actually have to do this, the code that drives this assumes the 1st allocated qubit has ID 0, which causes errors unless we switch the order. (This was always an issue, just previously there were no runtime checks that actually looked at the chosen qubit identifiers)

dime10: Thanks

runtime/lib/backend/null_qubit/ResourceTracker.hpp
jzaia18: I actually can't get this example to compile with clang version 14 (apparently 15 wouldn't work either), does Catalyst assume a clang version 16 or greater? If not, we may not be able to use this approach (but I can still prevent the copy and instead use 2 for loops)

jzaia18: Nvm, had a chat with some other members of performance team. I think it makes more sense just to iterate the 2 vectors separately. The need to combine them at all is vestigial

jzaia18: [f1cb350](https://github.com/PennyLaneAI/catalyst/pull/2203/commits/f1cb350c6667567c9d040bd67834a01538cc7ef4)

joeycarter: Looks good, thanks. :+1:

dime10 (approved):
Nice update :)

PR #2204, opened 2025-11-18
Lower ion to rtio

rniczh:
**Context:**

Convert ion dialect to rtio dialect

There are several steps to convert the ion dialect to rtio dialect:

0. Prepare the enviorment:
   - Import device db json to rtio.config
   - Collect ION infos
   - Create a kernel function, which is just cloned from the original qnode function
   - Drop the pulse with transition 0, that's because that can't be applied on the current device, and it will be dropped in the current ARTIQ's compilation flow as well
   - Initialize memref map: Construct mapping from `qreg alloc` and `qreg extract` to `memref`. In the later conversion, we use the mapping to construct the channel for `rtio.pulse`. It will create a global memref that contains index from 0 to N to represent the qubits.

1. parallel protocol to rtio:
   Converts `ion.parallelprotocol` regions into flat RTIO operations with proper synchronization. We rely on the data dependency of qubit to introduce the unrealized conversion cast between `ion.qubit` and `rtio.event`. That's the way how we currently build the relationship between events. Later, those unrealized conversion cast will be removed via event propagation.
   
   Input mlir:
   ```mlir
    %0, %1 = ion.parallelprotocol(%q0, %q1) {
      ^bb0(%arg0, %arg1):
      %p0 = rtio.pulse(...) : !rtio.event
      %p1 = rtio.pulse(...) : !rtio.event
      ion.yield %arg0, %arg1
    }
   ```
   
   ```mlir
    %event0 = unrealized_conversion_cast %q0 : !ion.qubit -> !rtio.event
    %event1 = unrealized_conversion_cast %q1 : !ion.qubit -> !rtio.event
    %p0 = rtio.pulse(..., wait = %inputSync) : !rtio.event
    %p1 = rtio.pulse(..., wait = %inputSync) : !rtio.event
    %sync = rtio.sync %p0, %p1 : !rtio.event
    %0 = unrealized_conversion_cast %sync : !rtio.event -> !ion.qubit
    %1 = unrealized_conversion_cast %sync : !rtio.event -> !ion.qubit
    ```
2. `ion.pulse` to `rtio.pulse`:
    Converts `ion.pulse` to `rtio.pulse` with calculated frequency.

    **Frequency Calculation:**
    ```cpp
    frequency = (reference_energy + detuning - raman1_frequency) / (2 * pi)
    ```
    
    **Steps:**
    - Extract pulse parameters (duration, beam, phase)
    - Calculate frequency from transition index and detuning
    - Create `rtio.channel` via `rtio.qubit_to_channel`
    - Generate `rtio.pulse` operation

3. Resolve channel mapping
    Resolves dynamic channel references to static channel IDs.

    ```
    memref.global (constants) -> memref.get_global -> memref.load -> qubit_to_channel
    ```
    
    ```mlir
    // Before
    %ch = rtio.qubit_to_channel %qubit : !ion.qubit -> !rtio.channel<"dds", ?>
    
    // After
    %ch = rtio.channel "dds" { channel_id = N } : !rtio.channel<"dds">
    ```

4. propagate events
    Resolves `unrealized_conversion_cast` and connect the ops with event type

    **Steps:**
    - Trace backward to collect all the source events.
    - Create sync from collected events
    - If reaches `quantum.AllocOp`: create `rtio.empty`

5. Remove all other dialects:
    Removes unused quantum/ion/memref/linalg dialect after conversion.

TODOs:
-  [x] Add lit tests

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-100851]


mlir/lib/Ion/Transforms/ion-to-rtio.cpp
```diff
@@ -0,0 +1,1139 @@
...
+///   - scf.if
+///   - ion.parallelprotocol
+///   - unrealized_conversion_cast
+///   - quantum.extract
+///   - quantum.insert
+template <TraceMode ModeT, typename CallbackT>
+auto traceValueWithCallback(Value value, CallbackT &&callback)
+{
+    WalkResult walkResult = WalkResult::advance();
+    std::queue<Value> visited;
```
mehrdad2m: very nitpicky suggestion: I usually prefer `worklist` instead of `visited` since we are popping from it.

doc/releases/changelog-dev.md
```diff
@@ -2,8 +2,9 @@
 
 <h3>New features since last release</h3>
 
-* RTIO dialect is added to bypass the compilation flow from OpenAPL to ARTIQ’s LLVM IR. It is introduced to bridge the gap between ION dialect and ARTIQ’s LLVM IR. The design philosophy of RTIO dialect is primarily event-based. Every operation is asynchronous; sync behaviour occurs only via `rtio.sync` or `wait operand` in event operation.
+* RTIO dialect is added to bypass the compilation flow from OpenAPL to ARTIQ’s LLVM IR. It is introduced to bridge the gap between ION dialect and ARTIQ’s LLVM IR. The design philosophy of RTIO dialect is primarily event-based. Every operation is asynchronous; sync behaviour occurs only via `rtio.sync` or `wait operand` in event operation. And we now support the compiling from ION dialect to RTIO dilalect.
```
paul0403: I don't think any OQD related stuff go under new features, since that is not part of the core catalyst features. We usually put them under internal changes.

mlir/include/RTIO/IR/RTIOOps.td
```diff
@@ -51,7 +51,7 @@ def RTIOChannelOp : RTIO_Op<"channel"> {
     }];
 }
 
-def RTIOQubitToChannelOp : RTIO_Op<"qubit_to_channel"> {
+def RTIOQubitToChannelOp : RTIO_Op<"qubit_to_channel", [Pure]> {
```
paul0403, line 54: Can you briefly explain why you added these pure traits? :thinking:

rniczh: `rtio.channel` and `rtio.qubit_to_channel` are side-effect free, the are solely represent the channel, and the actual behaviour are executed until the certain event operation that it connect to. So if there are no users of these operations, it's definitely fine to remove them.

doc/releases/changelog-dev.md
rniczh: Sure https://github.com/PennyLaneAI/catalyst/pull/2204/commits/7e7ee45d6ecd56642c01787d73c53287dab836a1

mlir/lib/Ion/Transforms/ion-to-rtio.cpp
```diff
@@ -0,0 +1,1135 @@
...
+
+        if constexpr (std::is_same_v<std::invoke_result_t<CallbackT, Value>, WalkResult>) {
+            if (callback(value).wasInterrupted()) {
+                walkResult = WalkResult::interrupt();
+                continue;
+            }
+        }
+        else {
+            callback(value);
+        }
```
mehrdad2m: Why do we need this? Seems like all the callback defined in this file always return a WalkResult?

mlir/lib/Ion/Transforms/ion-to-rtio.cpp
```diff
@@ -0,0 +1,1135 @@
...
+            }
+            else if constexpr (ModeT == TraceMode::Event) {
+                visited.push(qubit);
+                // only trace qreg if it defined op is also come from insert op
+                if (llvm::isa_and_present<quantum::InsertOp>(inQreg.getDefiningOp())) {
+                    visited.push(inQreg);
+                }
+            }
+        }
+    }
```
mehrdad2m: Here again, I don't see why this is needed. Especially that the return value is not used anywhere. Would it make sense to return a `LogicalResult` to help exist gracefully and get rid of the `llvm::reportFatalInternalError`?

mlir/lib/Ion/Transforms/ion-to-rtio.cpp
```diff
@@ -0,0 +1,1135 @@
...
+                continue;
+            }
+            else if (auto parallelProtocolOp = dyn_cast<ion::ParallelProtocolOp>(parentOp)) {
+                unsigned argIndex = arg.getArgNumber();
+                Value inQubit = parallelProtocolOp.getInQubits()[argIndex];
+                visited.push(inQubit);
+                continue;
+            }
+            parentOp->emitError("Unsupported parent operation for block argument: ") << value;
+            llvm::reportFatalInternalError("Unsupported block argument");
```
mehrdad2m: This seems to be a bit aggressive since it would abort compilation. I would rather exit gracefully if possible.

mlir/lib/Ion/Transforms/ion-to-rtio.cpp
```diff
@@ -0,0 +1,1135 @@
...
+{
+    if (events.size() == 1) {
+        return events.front();
+    }
+    auto eventType = rtio::EventType::get(rewriter.getContext());
+    return rewriter.create<rtio::RTIOSyncOp>(rewriter.getUnknownLoc(), eventType, events);
+}
+
+// Helper class to store ion information
+class IonInfo {
```
mehrdad2m: This class could be useful in general and is not tied to this pass. Can we move it to separate file to make this file smaller?

mlir/lib/Ion/Transforms/ion-to-rtio.cpp
```diff
@@ -0,0 +1,1135 @@
...
+            return transitions[index];
+        }
+        return std::nullopt;
+    }
+};
+
+} // namespace
+
+//===----------------------------------------------------------------------===//
+// Conversion Patterns
```
mehrdad2m: We usually have this convention of putting patterns in a separate file (like `IonToRTIOPatterns.cpp`) This would make sense here as well to maintain consistency and also making the file size smaller and manageable.

mlir/lib/Ion/Transforms/ion-to-rtio.cpp
```diff
@@ -0,0 +1,1135 @@
...
+    }
+    return success();
+}
+
+//===----------------------------------------------------------------------===//
+// JSON to MLIR Attribute Conversion
+//===----------------------------------------------------------------------===//
+
+/// Convert a JSON value to an MLIR Attribute
+Attribute jsonToAttribute(MLIRContext *ctx, const llvm::json::Value &json)
```
mehrdad2m: Can we also move this method out of this file since it is general util?

mlir/lib/Ion/Transforms/ion-to-rtio.cpp
rniczh: Sure. https://github.com/PennyLaneAI/catalyst/pull/2204/commits/e19c29ccc75384b35f5cfeadfcfcd9b6e048e524

mlir/lib/Ion/Transforms/ion-to-rtio.cpp
rniczh: Good point, back then, when I developing the helper function, I just want to cover all the possible cases we might need. But turn out for now, we can narrow the scope to ensure the callback function always need to declare with return type `WalkResult`. I fixed it in this commit:
https://github.com/PennyLaneAI/catalyst/pull/2204/commits/c368eff39ba3423e025ccc90d07d32d93dcd475a

mlir/lib/Ion/Transforms/ion-to-rtio.cpp
rniczh: Sure https://github.com/PennyLaneAI/catalyst/pull/2204/commits/81db1094c9c01b2c208efe4444dc2162c533eb10

mlir/lib/Ion/Transforms/ion-to-rtio.cpp
rniczh: https://github.com/PennyLaneAI/catalyst/pull/2204/commits/81db1094c9c01b2c208efe4444dc2162c533eb10

mlir/lib/Ion/Transforms/ion-to-rtio.cpp
rniczh: https://github.com/PennyLaneAI/catalyst/pull/2204/commits/436beaae316f985a3700fce43f185d54990b6205

mlir/lib/Ion/Transforms/ion-to-rtio.cpp
rniczh: no problem https://github.com/PennyLaneAI/catalyst/pull/2204/commits/820ec1994cc021d522f3640f4d1db61b41040812

mlir/lib/Ion/Transforms/ion-to-rtio.cpp
rniczh: Sure, I moved those utils and patterns to separate files https://github.com/PennyLaneAI/catalyst/pull/2204/commits/8c210ca9a6f68ce12d1ea7a100704a53a234f770

paul0403 (approved):

mlir/lib/Ion/Transforms/ion-to-rtio.cpp
```diff
@@ -0,0 +1,438 @@
...
+
+        // collect all ion information for calculating frequency when converting ion.pulse
+        SmallVector<IonInfo> ionInfos = getIonInfos();
+        if (ionInfos.empty()) {
+            getOperation()->emitError("Failed to get ion information");
+            return signalPassFailure();
+        }
+
+        // currently, we only support one ion information
+        assert(ionInfos.size() == 1 && "only one ion information is allowed");
```
mehrdad2m: Minor suggestion: I generally would prefer to only use asserts for for invariants that are purely internal and cannot be violated by user IR. In this case returning with `emitError` seems more appropriate. What do you think?

mlir/lib/Ion/Transforms/ion-to-rtio.cpp
```diff
@@ -0,0 +1,438 @@
...
+        // check if there is only one qnode function
+        func::FuncOp qnodeFunc = nullptr;
+        int qnodeCounts = 0;
+        module.walk([&](func::FuncOp funcOp) {
+            if (funcOp->hasAttr("qnode")) {
+                qnodeFunc = funcOp;
+                qnodeCounts++;
+            }
+        });
+        assert(qnodeCounts == 1 && "only one qnode function is allowed");
```
mehrdad2m: Same here

mlir/lib/Ion/Transforms/ion-to-rtio.cpp
rniczh: https://github.com/PennyLaneAI/catalyst/pull/2204/commits/dcc2ae5527fb1eaed712fc64384e87b7c789b90e
Thanks! I removed those asserts, they should return `signalPassFailure()`.
```
if (...) {
    getOperation()->emitError("...");
    return signalPassFailure();
}
```

mlir/lib/Ion/Transforms/ion-to-rtio.cpp
rniczh: Done https://github.com/PennyLaneAI/catalyst/pull/2204/commits/dcc2ae5527fb1eaed712fc64384e87b7c789b90e

PR #2206, opened 2025-11-18
Use transform pass name

albi3ro:
**Context:**

In #2149  I'm integrating transforms with passes names with the infrastructure so that we can replace the Catalyst `Pass` instracture with a unified transform handling ecosystem.

In order to test this integration,  I'm just switching the passes over to use the transform instructure.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


frontend/test/pytest/test_peephole_optimizations.py
```diff
@@ -353,6 +310,7 @@ def f():
 
         return f()
 
+    print(test_merge_ppr_ppm_workflow.mlir)
```
albi3ro:
```suggestion
```

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -1082,34 +1065,31 @@ def loop(i):
         . . .
 
     """
-
-    if fn.mlir_module is not None:
```
albi3ro, line 1188: Just switching to using a guard clause here. No actual code changes, but made it easier to scan the source code and see what was happening.

albi3ro:
> Did you accidentally commit the intermediates? 😅

looks like yes ...

PR #2208, opened 2025-11-19
Create xDSL universe for the unified compiler

mudit2812:
Duplicate of https://github.com/PennyLaneAI/pennylane/pull/8372

[sc-100654]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -368,6 +368,10 @@
   contains function declarations to external APIs, the names of these declarations must
   stay unchanged. This change enables quantum passes to generate calls to external APIs.
 
+* Fixes a bug where Catalyst was incorrectly raising an error about a missing shots parameter
+  on devices that support analytical execution.
+  [(#2281)](https://github.com/PennyLaneAI/catalyst/pull/2281)
+
 <h3>Internal changes ⚙️</h3>
 
 * Integration tests for `qml.specs` have been updated to match the new output format introduced
--- a/frontend/catalyst/_version.py
+++ b/frontend/catalyst/_version.py
@@ -16,4 +16,4 @@
 Version number (major.minor.patch[-label])
 """
 
-__version__ = "0.14.0-dev54"
+__version__ = "0.14.0-dev55"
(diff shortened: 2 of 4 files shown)


frontend/catalyst/python_interface/__init__.py
```diff
@@ -13,6 +13,15 @@
...
 
+# pylint: disable=wrong-import-position
+from importlib.util import find_spec
+
+if not (find_spec("xdsl") and find_spec("xdsl_jax")):  # pragma: no cover
+    raise RuntimeError(
+        "Using the Unified compiler framework requires xDSL and xDSL-JAX to be installed. "
+        "They can be installed by executing 'python -m pip install xdsl xdsl-jax'."
+    )
+
```
mudit2812: Everything failing. Need to remove this.

frontend/catalyst/python_interface/xdsl_universe.py
```diff
@@ -0,0 +1,57 @@
...
+try:
+    from xdsl.passes import ModulePass
+    from xdsl.universe import Universe
+except (ImportError, ModuleNotFoundError):
+    xdsl_available = False  # pragma: no cover
+
+# We must check that xDSL is installed because we're adding an entry point to
+# PennyLane that references this file, and we must ensure that PennyLane can
+# be installed in environments where xDSL is not installed.
+XDSL_UNIVERSE = None
```
superlopuh, line 27: Just seeing this now, but this could be called 
```suggestion
CATALYST_XDSL_UNIVERSE = None
```
and it would work just as well, as long as it was in sync with the entry point in setup.py. This might make the tests a bit less awkward if importing XDSL_UNIVERSE directly. Of course it works great as-is, glad to see it used here!

dime10: Thanks Sasha!

mudit2812: > This might make the tests a bit less awkward if importing XDSL_UNIVERSE directly

Ah, that makes sense. The testing right now is minimal and the motivation behind this PR was strictly to expose our dialects and passes to xDSL's tools (xdsl-gui, xdsl-opt), so it shouldn't matter too much. I'll track this is a minor improvement. Thanks for the suggestion 🙂

PR #2214, opened 2025-11-19
feat: add `ConstructCircuitDAG` director pattern

andrijapau:
This utility is used to traverse the IR and then use the DAGBuilder to generate a DAG to be visualized.

[sc-103458]

### DIFF
--- a/frontend/test/pytest/python_interface/visualization/test_construct_circuit_dag.py
+++ b/frontend/test/pytest/python_interface/visualization/test_construct_circuit_dag.py
@@ -98,7 +98,7 @@ class FakeDAGBuilder(DAGBuilder):
 def test_dependency_injection():
     """Tests that relevant dependencies are injected."""
 
-    mock_dag_builder = Mock(DAGBuilder) 
+    mock_dag_builder = Mock(DAGBuilder)
     utility = ConstructCircuitDAG(mock_dag_builder)
     assert utility.dag_builder is mock_dag_builder
 


frontend/catalyst/python_interface/visualization/circuit_dag_analysis_pass.py
```diff
@@ -0,0 +1,131 @@
...
+    ) -> None:
+        """Generic handler for unitary gates and quantum state preparation operations."""
+        pass
+
+    # =============================================
+    # 4. QUANTUM MEASUREMENT & OBSERVABLE HANDLERS
+    # =============================================
+
+    @visit_op.register
+    def _visit_terminal_state_op(self, op: quantum.StateOp) -> None:
```
andrijapau:
```suggestion
    def _visit_state_op(self, op: quantum.StateOp) -> None:
```

frontend/catalyst/python_interface/visualization/circuit_dag_analysis_pass.py
```diff
@@ -0,0 +1,131 @@
...
+        """Generic handler for unitary gates and quantum state preparation operations."""
+        pass
+
+    # =============================================
+    # 4. QUANTUM MEASUREMENT & OBSERVABLE HANDLERS
+    # =============================================
+
+    @visit_op.register
+    def _visit_state_op(self, op: quantum.StateOp) -> None:
+        """Handler for the terminal StateOp, which retrieves the final state vector."""
```
andrijapau:
```suggestion
        """Handler for the terminal StateOp."""
```

frontend/catalyst/python_interface/visualization/circuit_dag_analysis_pass.py
```diff
@@ -0,0 +1,131 @@
...
+    def _visit_state_op(self, op: quantum.StateOp) -> None:
+        """Handler for the terminal StateOp."""
+        pass
+
+    @visit_op.register
+    def _visit_statistical_measurement_ops(
+        self,
+        op: quantum.ExpvalOp | quantum.VarianceOp | quantum.ProbsOp | quantum.SampleOp,
+    ) -> None:
+        """Handler for statistical measurement operations (e.g., Expval, Sample)."""
```
andrijapau:
```suggestion
        """Handler for statistical measurement operations."""
```

frontend/catalyst/python_interface/visualization/circuit_dag_analysis_pass.py
```diff
@@ -0,0 +1,131 @@
...
+
+    @visit_op.register
+    def _visit_while_op(self, op: scf.WhileOp) -> None:
+        """Handle an xDSL WhileOp operation (Loop cluster creation)."""
+        pass
+
+    @visit_op.register
+    def _visit_if_op(self, op: scf.IfOp) -> None:
+        """Handle an xDSL IfOp operation (Conditional cluster creation)."""
+        pass
```
andrijapau:
```suggestion
    @visit_op.register
    def _visit_for_op(self, op: scf.ForOp) -> None:
        """Handle an xDSL ForOp operation."""
        pass

    @visit_op.register
    def _visit_while_op(self, op: scf.WhileOp) -> None:
        """Handle an xDSL WhileOp operation."""
        pass

    @visit_op.register
    def _visit_if_op(self, op: scf.IfOp) -> None:
        """Handle an xDSL IfOp operation."""
        pass
```

frontend/catalyst/python_interface/visualization/circuit_dag_analysis_pass.py
```diff
@@ -0,0 +1,131 @@
...
+            self.visit_op(op)
+
+    # =======================
+    # 2. HIERARCHY TRAVERSAL
+    # =======================
+    # These methods navigate the recursive IR hierarchy (Op -> Region -> Block -> Op).
+
+    @visit_op.register
+    def visit_region(self, region: Region) -> None:
+        """Visit an xDSL Region operation, delegating traversal to its Blocks."""
```
andrijapau:
```suggestion
        """Visit an xDSL Region operation."""
```

frontend/catalyst/python_interface/visualization/circuit_dag_analysis_pass.py
```diff
@@ -0,0 +1,131 @@
...
+            | quantum.MultiRZOp
+            | quantum.SetStateOp
+            | quantum.SetBasisStateOp
+        ),
+    ) -> None:
+        """Generic handler for unitary gates and quantum state preparation operations."""
+        pass
+
+    # =============================================
+    # 4. QUANTUM MEASUREMENT & OBSERVABLE HANDLERS
```
andrijapau:
```suggestion
    # 4. QUANTUM MEASUREMENT HANDLERS
```

frontend/test/pytest/python_interface/visualization/test_construct_circuit_dag.py
```diff
@@ -0,0 +1,123 @@
...
+from catalyst.python_interface.visualization.dag_builder import DAGBuilder
+from xdsl.dialects import test
+from xdsl.dialects.builtin import ModuleOp
+from xdsl.ir.core import Block, Region
+
+
+class FakeDAGBuilder(DAGBuilder):
+    """
+    A concrete implementation of DAGBuilder used ONLY for testing.
+    It stores all graph manipulation calls in simple Python dictionaries
```
```suggestion
    It stores all graph manipulation calls in data structures
```

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -0,0 +1,78 @@
...
+
+from xdsl.dialects import builtin
+from xdsl.ir import Block, Operation, Region
+
+from catalyst.python_interface.visualization.dag_builder import DAGBuilder
+
+
+class ConstructCircuitDAG:
+    """A tool that traverses an xDSL module and constructs a Directed Acyclic Graph (DAG)
+    of it's quantum program using an injected DAGBuilder instance. This tool does not mutate the xDSL module.
```
mudit2812, line 29: Not sure what injection really means here.
```suggestion
    of it's quantum program using a ``DAGBuilder`` instance. This does not mutate the xDSL module.
```

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+
+    def __init__(self, dag_builder: DAGBuilder) -> None:
+        """Initialize the utility by injecting the DAG builder dependency.
+
+        Args:
+            dag_builder (DAGBuilder): The concrete builder instance used for graph construction.
+        """
```
mudit2812: Init methods shouldn't have docstrings. Any documentation needed for initialization must be in the class's main docstring.
```suggestion
```

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
andrijapau: Was following this terminology: https://en.wikipedia.org/wiki/Dependency_injection

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -0,0 +1,78 @@
...
+
+        """
+        for op in module.ops:
+            self._visit(op)
+
+    # =======================
+    # 2. IR TRAVERSAL
+    # =======================
+
+    @_visit.register
```
mudit2812: I would move the main `_visit` entry point from the "CORE DISPATCH" section to the "IR TRAVERSAL" section.

andrijapau: Fair enough

andrijapau: https://github.com/PennyLaneAI/catalyst/pull/2214/commits/40b1eb4885996f505dfdeeec1719c788946aa2f7

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
mudit2812: Thanks for the reference, sounds good. Resolving.

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -0,0 +1,69 @@
...
+    def _region(self, region: Region) -> None:
+        """Visit an xDSL Region operation."""
+        for block in region.blocks:
+            self._visit(block)
+
+    @_visit.register
+    def _block(self, block: Block) -> None:
+        """Visit an xDSL Block operation, dispatching handling for each contained Operation."""
+        for op in block.ops:
+            self._visit(op)
```
mudit2812: Instead of having a `singledispatch` for regions, blocks, and operations, does it make sense to have a separate visitor for regions and blocks, i.e, `_visit_block`, `_visit_region`, and then another for operations (`_visit_op`). Then, `_visit_op` can be the `singledispatch` with a default implementation and registrations for all operations for which we would want to have custom implementations.

andrijapau: I like that idea. I can try it out!

andrijapau: Seems to work well https://github.com/PennyLaneAI/catalyst/pull/2214/commits/0460cac9ad28ce6e0e17382b6cdf74ef00ab131c

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -0,0 +1,72 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Contains the ConstructCircuitDAG tool for constructing a DAG from an xDSL module."""
+
+from functools import singledispatchmethod
+from typing import Any
```
mehrdad2m:
```suggestion
```

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def _visit_region(self, region: Region) -> None:
+        """Visit an xDSL Region operation."""
+        for block in region.blocks:
+            self._visit_block(block)
+
+    def _visit_block(self, block: Block) -> None:
+        """Visit an xDSL Block operation, dispatching handling for each contained Operation."""
+        for op in block.ops:
+            self._visit_operation(op)
```
mudit2812, line 71: Does it make sense for the default implementations of these methods to add nodes/clusters to the dag_builder? Eg, for `_visit_operation`, the default could be something along the lines of
```python
def _visit_operation(self, operation):
    if not op.regions:
        node_id = ...
        self.dag_builder.add_node(...)

    else:
        cluster_id = ...
        self.dag_builder.add_cluster(...)
        self._cluster_stack.append(cluster_id)

        for r in op.regions:
            self._visit_region(r)

        self._cluster_stack.pop()
```
And something similar for regions and blocks as well

mudit2812: And you might need special cases for single-block regions and single-region ops as well, but it seems like a sensible default.

andrijapau: But do we want to add all operations as nodes / clusters?

andrijapau: I'd rather be very deliberate what gets visualized by dispatching. 🤔 What do you think?

mudit2812: Ah, right. Yeah sounds good. For which ops do we want to add nodes? Gates, measurements, MCMs, control flow. What about arithmetic ops? `quantum.extract`/`quantum.insert`? cc @AntonNI8

andrijapau: As of right now, I would say no. Correct me if I am wrong @AntonNI8.

> Gates, measurements, MCMs, control flow 

The list you outlined above is the current expectation for MVP. Later on we can imagine adding classical operation visualization as well.

AntonNI8: @andrijapau has it right here, no need to worry about these insertion and extraction operations for now.

frontend/test/pytest/python_interface/visualization/test_construct_circuit_dag.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def to_string(self) -> str:
+        return "graph"
+
+
+@pytest.mark.unit
+def test_dependency_injection():
+    """Tests that relevant dependencies are injected."""
+
+    dag_builder = FakeDAGBuilder()
```
mehrdad2m: Why not using the `Mock(DAGBuilder)` here too instead of the manually created `FakeDAGBuilder()`

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -0,0 +1,72 @@
...
+        """
+        for op in module.ops:
+            self._visit_operation(op)
+
+    # =============
+    # IR TRAVERSAL
+    # =============
+
+    @singledispatchmethod
+    def _visit_operation(self, operation: Operation) -> None:
```
mehrdad2m, line 58: How would we insert nodes into the DAG when visiting an operation? Do we need to traverse the DAG for each visit? Not sure if it make sene but I am wondering if it helps to have some sort of context/scope tracking like a stack to associate a node to say the current `cluster_id`.

mehrdad2m (commented):
Thanks @andrijapau, Looks very good. Just some comments:

frontend/test/pytest/python_interface/visualization/test_construct_circuit_dag.py
andrijapau: Good point, oversight 😓 https://github.com/PennyLaneAI/catalyst/pull/2214/commits/3632ebe2bbda9d02499d770e33909321a419f660

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
andrijapau: Good question. Yes, I will be doing exactly that. Internally I will have a `cluster_stack` state variable and everytime I build a cluster, I append to said stack (and pop when required). Then each node just gets added to the top cluster on the stack.

Hopefully that makes sense 😅 .

mudit2812 (approved):

Issue #2114, opened 2025-10-10
Test suite maintenance regarding qjit device capabilities


paul0403:
One of the purposes of [the `QJITDevice` class](https://github.com/PennyLaneAI/catalyst/blob/659c69a878c218a7dd9c13e7309e916159deef11/frontend/catalyst/device/qjit_device.py#L295) is to compute the preprocessing transform program depending on the device capabilities, i.e. what qnode-level transforms (aka `qml.transfroms.blah`) need to happen before reaching the mlir layer. For example, for devices without analytical mode execution, the `measurement_from_samples/counts` qnode transform is added here.

The code around here has recently been refactored https://github.com/PennyLaneAI/catalyst/pull/2089 (to account for whether the user has provided shots in the program, as well as just general clean up). One remaining function, `get_device_capabilities` , is not actually used in the codebase anymore. However, it was not deleted, because various tests are still using this internal function directly. 

See https://github.com/PennyLaneAI/catalyst/pull/2089#discussion_r2421256201 for details.

The tests should properly construct custom capability sets for their respective purposes, instead of piggy-backing off lightning.
As an alternative, the `get_device_capabilities` function can be demoted to a test util, and decoupled from the actual code base.

PR #2215, opened 2025-11-20
[Prototype] Remove the hardcoded list of runtime operations in the frontend

maliasadi:
**Context:**
The frontend no longer maintains a hardcoded list of runtime operations, allowing arbitrary PennyLane gates with Quantum dialect-compatible hyperparameters to be captured and represented in MLIR. Users of the legacy compilation pipeline are unaffected, as Catalyst continues to decompose unsupported gates  based on device capabilities before lowering to MLIR. Gates that cannot be represented as MLIR operators will temporarily  raise a `CompileError` during program capture. The long-term solution will integrate the new decomposition framework  with capture-enabled compilation.

- Remove `RUNTIME_OPERATIONS` from `qjit_device.py`
- Add `is_lowering_compatible` to check if an operation can be lowered to MLIR using primitives.
- Update quantum operations capture logic in `from_plxpr` 

**Benefits:**
- Lower and represent all gates in the device toml file as long as they can be represented as `quantum.custom` op
- Support IR representation of custom gates with compatible `hyperparameters` at MLIR
- Users can decompose their custom ops via `qml.transforms.decompose` for gates w/o runtime support, or just rely on the device decomposition to deal with unsupported ops

**Possible Drawbacks:**
- Device developers are responsible to maintain the device toml file based on the QIR QIS 

**Related GitHub Issues:**
[sc-102160]

### DIFF
--- a/frontend/catalyst/_version.py
+++ b/frontend/catalyst/_version.py
@@ -16,4 +16,4 @@
 Version number (major.minor.patch[-label])
 """
 
-__version__ = "0.14.0-dev49"
+__version__ = "0.14.0-dev50"


frontend/catalyst/device/op_support.py
```diff
@@ -41,6 +41,28 @@ def is_supported(op: Operator, capabilities: DeviceCapabilities) -> bool:
...
+def is_lowering_compatible(op: Operator) -> bool:
+    """Check if an operation can be lowered to MLIR using JAX primitives."""
+    # Exceptions for operations that are not quantum instructions but are allowed
+    # via custom lowering rules.
+    # TODO: Revisit this as more explicit ops will be added to Catalyst Compiler.
+    if isinstance(op, (qml.Snapshot, qml.PCPhase, qml.MultiRZ)):
+        return True
+
+    # Accepted hyperparameters for quantum instructions bind calls
+    _accepted_hyperparams = {
```
maliasadi, line 53: @albi3ro Here's the minimum set of hyperparams that we always need to check for compatibility with the `quantum.custom` MLIR Op.

frontend/catalyst/device/op_support.py
```diff
@@ -41,6 +41,28 @@ def is_supported(op: Operator, capabilities: DeviceCapabilities) -> bool:
     return op.name in capabilities.operations
 
 
+def is_lowering_compatible(op: Operator) -> bool:
```
dime10, line 44: I'm not convinced this function is sufficient to guarantee compatibility with the quantum dialect. Looking at the [structure of an operator](https://github.com/PennyLaneAI/pennylane/blob/77531f6e5ad02706f1faf9fc4525d766a1873f58/pennylane/operation.py#L396-L398):
>  Operators are uniquely defined by their **name**, the **wires** they act on, their (trainable) **parameters**,
    and their (non-trainable) **hyperparameters**. The trainable parameters can be tensors of any
    supported auto-differentiation framework.

Name and wires are always supported by `quantum.custom`. 

Hyperparameters are being checked below, although I don't know if the check is not too generic (i.e. I don't know if _all_ hyperparameters are supported by _all_ operation types).

The thing that's missing entirely is verifying parameters. Pennylane only requires that they be tensor-like, so this could be an arbitrary sequence of arbitrary tensors. This cannot be mapped to the quantum dialect at the moment. Some special ops support certain tensors (like 2D complex for quantum.unitary), but for the generic quantum.custom we would only support a sequence of scalar floats (or perhaps a single tensor that is flattened into a sequence of floats, but this could get highly inefficient if the tensor is large).

dime10, line 44: Do we need to consider operations like quantum.measure at all, since its signature even has a return value 🤔

frontend/catalyst/device/op_support.py
```diff
@@ -41,6 +41,28 @@ def is_supported(op: Operator, capabilities: DeviceCapabilities) -> bool:
...
+
+    # Accepted hyperparameters for quantum instructions bind calls
+    _accepted_hyperparams = {
+        "base",
+        "n_wires",
+        "num_wires",
+        "control_wires",
+        "control_values",
+        "work_wires",
+        "work_wire_type",
```
dime10, line 60: Do we handle work wires when mapping to mlir?

frontend/catalyst/device/qjit_device.py
```diff
@@ -109,11 +71,9 @@
...
 RUNTIME_MPS = ["ExpectationMP", "SampleMP", "VarianceMP", "CountsMP", "StateMP", "ProbabilityMP"]
 
-# The runtime interface does not care about specific gate properties, so set them all to True.
-RUNTIME_OPERATIONS = {
-    op: OperatorProperties(invertible=True, controllable=True, differentiable=True)
-    for op in RUNTIME_OPERATIONS
-}
+# A list of custom operations supported by the Catalyst compiler.
+# This is useful especially for testing a device with custom operations.
+CUSTOM_OPERATIONS = {}
```
paul0403, line 76: Is there a point to keep this empty set, considering its entire purpose is to be unioned with something else?

Or is it just for tests? In that case possibly related: #2114

frontend/catalyst/device/qjit_device.py
```diff
@@ -199,6 +159,13 @@ def extract_backend_info(device: qml.devices.QubitDevice) -> BackendInfo:
     return BackendInfo(dname, device_name, device_lpath, device_kwargs)
 
 
+def union_operations(
+    a: Dict[str, OperatorProperties], b: Dict[str, OperatorProperties]
+) -> Dict[str, OperatorProperties]:
+    """Union of two sets of operator properties"""
+    return {**a, **b}
```
paul0403, line 166: Any reason why not just 
```
>>> x
{1: 2}
>>> y
{3: 4}
>>> x.update(y)
>>> x
{1: 2, 3: 4}
```
?

PR #2219, opened 2025-11-20
Support multiple executions for runtime resource tracking

jzaia18:
**Context:**
Runtime resource tracking currently writes out resource data at the moment that all allocated qubits are released. This causes issues when executing functions that allocate and deallocate multiple times, such as this illustrative example:

```python
import pennylane as qml
from pennylane.compiler.python_compiler.transforms import split_non_commuting_pass

qml.capture.enable()

@qml.qjit(keep_intermediate=True)
@split_non_commuting_pass
@qml.qnode(dev)
def circuit():
    qml.Hadamard(wires=0)
    return qml.expval(qml.PauliX(0)), qml.expval(qml.PauliY(0))

print(qml.specs(circuit, level="device")())
```

Updates `NullQubit` to instead write out resource information in its destructor, ensuring all resource data is the sum of data collected for ***all*** executions.

**Benefits:**
Prevents crashes with an ambiguous error message, allows total resource counts to be collected from split workflows

**Possible Drawbacks:**
The aggregated data is not possible to disaggregate if this would be desired in the future.

**Related GitHub Issues:**
[sc-104344]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -52,6 +52,9 @@
 
 <h3>Breaking changes 💔</h3>
 
+* The plxpr transform `pl_map_wires` has been removed along with its test.
+  [(#2220)](https://github.com/PennyLaneAI/catalyst/pull/2220)
+
 * (Compiler integrators only) The versions of LLVM/Enzyme/stablehlo used by Catalyst have been
   updated. Enzyme now targets `v0.0.203` with the build target `EnzymeStatic-22`, and the nanobind
   requirement for the latest LLVM has been updated to version 2.9.
@@ -203,6 +206,7 @@
 This release contains contributions from (in alphabetical order):
 
 Ali Asadi,
+Yushao Chen,
 Sengthai Heng,
 Jeffrey Kam,
 Christina Lee,
--- a/frontend/catalyst/from_plxpr/from_plxpr.py
+++ b/frontend/catalyst/from_plxpr/from_plxpr.py
@@ -29,7 +29,6 @@ from pennylane.capture import PlxprInterpreter, qnode_prim
 from pennylane.capture.expand_transforms import ExpandTransformsInterpreter
 from pennylane.capture.primitives import jacobian_prim as pl_jac_prim
 from pennylane.capture.primitives import transform_prim
-from pennylane.ops.functions.map_wires import _map_wires_transform as pl_map_wires
 from pennylane.transforms import cancel_inverses as pl_cancel_inverses
 from pennylane.transforms import commute_controlled as pl_commute_controlled
 from pennylane.transforms import decompose as pl_decompose
@@ -266,7 +265,6 @@ transforms_to_passes = {
     pl_cancel_inverses: ("cancel-inverses", False),
     pl_commute_controlled: (None, False),
     pl_decompose: (None, False),
-    pl_map_wires: (None, False),
     pl_merge_amplitude_embedding: (None, True),
     pl_merge_rotations: ("merge-rotations", False),
     pl_single_qubit_fusion: (None, False),
--- a/frontend/test/pytest/from_plxpr/test_capture_integration.py
+++ b/frontend/test/pytest/from_plxpr/test_capture_integration.py
@@ -1372,37 +1372,6 @@ class TestCapture:
 
         assert jnp.allclose(circuit(1.5, 2.5, 3.5), capture_result)
 
-    def test_transform_map_wires_workflow(self, backend):
-        """Test the integration for a circuit with a 'map_wires' transform."""
-
-        # Capture enabled
-
-        qml.capture.enable()
-
-        @qjit(target="mlir")
-        @partial(qml.map_wires, wire_map={0: 1})
-        @qml.qnode(qml.device(backend, wires=2))
-        def captured_circuit(x):
-            qml.RX(x, 0)
-            return qml.expval(qml.PauliZ(0))
-
-        capture_result = captured_circuit(1.5)
-
-        assert is_wire_mapped(captured_circuit.mlir)
-
-        qml.capture.disable()
-
-        # Capture disabled
-
-        @qjit
-        @partial(qml.map_wires, wire_map={0: 1})
-        @qml.qnode(qml.device(backend, wires=2))
-        def circuit(x):
-            qml.RX(x, 0)
-            return qml.expval(qml.PauliZ(0))
-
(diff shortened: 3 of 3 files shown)


runtime/lib/backend/null_qubit/NullQubit.hpp
```diff
@@ -79,7 +79,16 @@ struct NullQubit final : public Catalyst::Runtime::QuantumDevice {
             this->resource_tracker_.SetComputeDepth(device_kwargs["compute_depth"] == "True");
         }
     }
-    ~NullQubit() {} // LCOV_EXCL_LINE
+    ~NullQubit()
+    {
+        // We always want to gather resources that were used for an *entire* execution end-to-end
+        // A device is guaranteed to live as long as its ExecutionContext, so it's destructor is a
```
multiphaseCFD: `its destructor`?

runtime/lib/backend/null_qubit/NullQubit.hpp
```diff
@@ -154,9 +163,6 @@ struct NullQubit final : public Catalyst::Runtime::QuantumDevice {
         for (auto q : qubits) {
             this->ReleaseQubit(q);
         }
```
multiphaseCFD, line 156: Would you like to update the docstr here?

runtime/lib/backend/null_qubit/NullQubit.hpp
```diff
@@ -79,7 +79,16 @@ struct NullQubit final : public Catalyst::Runtime::QuantumDevice {
             this->resource_tracker_.SetComputeDepth(device_kwargs["compute_depth"] == "True");
         }
     }
-    ~NullQubit() {} // LCOV_EXCL_LINE
+    ~NullQubit()
```
mlxd, line 82: Should this be virtual?
UPDATE: Missed the  `final` -- all good

runtime/tests/Test_NullQubit.cpp
```diff
@@ -938,11 +938,19 @@ TEST_CASE("Test NullQubit device resource tracking integration", "[NullQubit]")
...
 
-    // Open the file of resource data
+    // Check no prior steps actually wrote the resource file
     std::ifstream resource_file_r(RESOURCES_FILENAME);
+    if (resource_file_r.is_open()) {
+        FAIL("Resource file should not exist before releasing qubits"); // LCOV_EXCL_LINE
+        std::remove(RESOURCES_FILENAME.c_str());                        // LCOV_EXCL_LINE
+    }
+
+    sim.reset(); // Destroy the device to trigger writing the resource data
```
paul0403, line 950: Wait, there's a `reset` method on null qubit??

jzaia18: It's on the `std::unique_ptr`

paul0403: Oh it's a dot not an arrow... I was blind

runtime/tests/Test_NullQubit.cpp
```diff
@@ -938,11 +938,19 @@ TEST_CASE("Test NullQubit device resource tracking integration", "[NullQubit]")
...
     sim->MatrixOperation({}, {Qs[0]}, true, {Qs[1], Qs[2]});
 
-    // Releasing all qubits should write out all tracked information
     sim->ReleaseQubits(Qs);
 
-    // Open the file of resource data
+    // Check no prior steps actually wrote the resource file
     std::ifstream resource_file_r(RESOURCES_FILENAME);
+    if (resource_file_r.is_open()) {
+        FAIL("Resource file should not exist before releasing qubits"); // LCOV_EXCL_LINE
```
paul0403:
```suggestion
        FAIL("Resource file should not exist before releasing the device"); // LCOV_EXCL_LINE
```
?

runtime/tests/Test_NullQubit.cpp
paul0403: would this better communicate the intent of "I'm calling the destructor"?
```C 
{
   unique_ptr<NullQubit> sim;
   // do stuff
   // check no resource tracking results were dumped
}

// check dumped resource tracking results
```

But if you prefer the `reset()` call I'm also fine with that

jzaia18: I think the `reset()` is slightly cleaner but I don't have any strong opinions, anyone else want to weigh in?

runtime/tests/Test_NullQubit.cpp
paul0403: Do you not need to update the error msg itself?

Or is this not a "check I failed with this error msg, but just check that I failed, and this msg is just a comment"?

runtime/tests/Test_NullQubit.cpp
paul0403: I don't have strong opinions, if you want `reset` then let's go with that 👍

runtime/tests/Test_NullQubit.cpp
jzaia18: This is like an `assert(False)` and then it logs that message as the reason for failing the test

jzaia18: But I did just realize that the 2 lines are flipped, will fix

doc/releases/changelog-dev.md
```diff
@@ -130,6 +130,9 @@
 
 <h3>Internal changes ⚙️</h3>
 
+* Resource tracking now writes out at device destruction instead of qubit deallocation
```
multiphaseCFD: Please add notes on the operations collected with this update, specifically for the split-non-commuting pass.

runtime/lib/backend/null_qubit/NullQubit.hpp
jzaia18: Good catch!

doc/releases/changelog-dev.md
paul0403: Good point! We can say something like "for multiple qnodes being executed on the same device, for instance those produced by `qml.transforms.split_non_commuting`, ..."

runtime/lib/backend/null_qubit/NullQubit.hpp
jzaia18: :open_mouth:

paul0403 (approved):
All good from my side, just pending resolution of Shuli's comments 👍

multiphaseCFD (approved):
Nice one! Thanks @jzaia18 !

PR #2224, opened 2025-11-21
merge_rotations PPR support

kipawaa:
**Context:**
With the addition of QEC and PPRs, passes need to be updated to accommodate them.

Adds a merge rule for PPRs to the `merge_rotations` pass, which merges PPRs of the form 

$$PPR(\theta) \cdot PPR(\theta') \to PPR(\theta + \theta'). $$

The rule also simplifies the case 

$$PPR(\theta) \cdot PPR(-\theta) = I \to \textnormal{no instruction}$$

 and 

$$PPR(\pi - \theta) \cdot PPR(\theta) \to \textnormal{no instruction}.$$

Note: PPRs use a different angle convention than `qml.PauliRot`, so $PPR(\pi) = I$.

**Benefits:**
Simplifies mlir output with regards to PPRs, adds expected functionality to the `merge_rotations` pass.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A

[sc-104292]
[sc-104293]

### DIFF
--- a/frontend/test/pytest/test_peephole_optimizations.py
+++ b/frontend/test/pytest/test_peephole_optimizations.py
@@ -496,10 +496,6 @@ def test_commute_ppr_and_merge_ppr_ppm_with_max_pauli_size():
 @pytest.mark.usefixtures("use_capture")
 def test_merge_rotation_ppr():
     """Test that the merge_rotation pass correctly merges PPRs."""
-    if not qml.capture.enabled():
-        pytest.skip("capture only test")
-
-    qml.capture.enable()  # capture is needed for PauliRot -> qec.ppr
 
     my_pipeline = [("pipe", ["enforce-runtime-invariants-pipeline"])]
 
@@ -508,7 +504,6 @@ def test_merge_rotation_ppr():
         @qml.transforms.merge_rotations  # have to use qml to be capture-compatible
         @qml.qnode(qml.device("lightning.qubit", wires=3))
         def circuit():
-            # equivalent to a Hadamard gate
             qml.PauliRot(np.pi / 2, pauli_word="XYZ", wires=[0, 1, 2])
             qml.PauliRot(np.pi / 2, pauli_word="XYZ", wires=[0, 1, 2])
 
@@ -516,13 +511,11 @@ def test_merge_rotation_ppr():
 
     ir = test_merge_rotation_ppr_workflow.mlir
     ir_opt = test_merge_rotation_ppr_workflow.mlir_opt
-    print(ir_opt)
+
     assert 'transform.apply_registered_pass "merge-rotations"' in ir
     assert "qec.ppr" in ir
     assert 'qec.ppr ["X", "Y", "Z"](4)' not in ir_opt
     assert 'qec.ppr ["X", "Y", "Z"](2)' in ir_opt
-
-    qml.capture.disable()
 
 
 def test_clifford_to_ppm():


mlir/include/QEC/Transforms/Passes.td
```diff
@@ -85,6 +85,14 @@ def CommutePPRPass : Pass<"commute-ppr"> {
     let options = [MaxPauliSizeOption];
 }
 
+def MergePPRPass : Pass<"merge-ppr"> {
+    let summary = "Merge like PPRotation operations.";
+
+    let dependentDialects = [ "catalyst::qec::QECDialect" ];
+
+    let options = [MaxPauliSizeOption];
```
lazypanda10117: I am not sure if we need the `MaxPauliSizeOption` in general for this merge rotation pass?

mlir/test/QEC/MergePPR.mlir
```diff
@@ -0,0 +1,336 @@
...
+    %0 = qec.ppr ["Z"](8) %q1: !quantum.bit
+    %1 = qec.ppr ["X"](-4) %0: !quantum.bit
+    %2 = qec.ppr ["Z"](-8) %1: !quantum.bit
+    %3 = qec.ppr ["X"](4) %2: !quantum.bit
+    func.return
+}
+
+// CHECK-LABEL: cancel_correct_references
+func.func public @cancel_correct_references(%q1: !quantum.bit) {
+    // CHECK: %[[in:[a-zA-Z0-9_]+]] = qec.ppr ["Y"](2)
```
lazypanda10117: I think usually we can just do something simpler like this `[[q1:%.+]]`

mlir/test/QEC/MergePPR.mlir
```diff
@@ -0,0 +1,336 @@
...
+    %2 = qec.ppr ["Z"](8) %1: !quantum.bit
+    %3 = qec.ppr ["X"](8) %2: !quantum.bit
+    %4 = qec.ppr ["Y"](2) %3: !quantum.bit
+    %5 = qec.ppr ["Z"](2) %4: !quantum.bit
+    func.return
+}
+
+// CHECK-LABEL: merge_correct_references
+func.func public @merge_correct_references(%q1: !quantum.bit) {
+    // CHECK: %[[in:[a-zA-Z0-9_]+]] = qec.ppr ["Y"](4)
```
lazypanda10117: Same comment as below.

mlir/lib/QEC/Transforms/merge_ppr.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include "mlir/Pass/Pass.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+#include "QEC/IR/QECDialect.h"
+#include "QEC/Transforms/Patterns.h"
+
+using namespace llvm;
+using namespace mlir;
+using namespace catalyst;
+using namespace catalyst::qec;
```
lazypanda10117: I am a little confused why we need this `catalyst::qec` namespace here.

doc/releases/changelog-dev.md
```diff
@@ -189,6 +189,10 @@
   of identities.
   [(#2192)](https://github.com/PennyLaneAI/catalyst/pull/2192)
 
+  * Added a `merge-ppr` pass that allows merging of PPRs with equivalent angles, and cancelling of
+  PPRs with opposite angles.
+  [(#2224)](https://github.com/PennyLaneAI/catalyst/pull/2224)
```
lazypanda10117, line 267: We probably want to add a python interface to call this mlir pass from the frontend. For example, take a look at `ppr_to_ppm()` in `builtin_passes.py`.

mlir/lib/QEC/Transforms/MergePPR.cpp
```diff
@@ -0,0 +1,110 @@
...
+        if (exceedPauliSizeLimit(op.getPauliProduct().size(), MAX_PAULI_SIZE)) {
+            return failure();
+        }
+
+        // check same pauli strings
+        if (op.getPauliProduct() != prevOp.getPauliProduct()) {
+            return failure();
+        }
+
+        // TODO this will need to update when we allow arbitrary angles
```
lazypanda10117: Probably don't want to leave the TODO's here but track them in a story.  Tracking in [sc-104451] for now.

doc/releases/changelog-dev.md
```diff
@@ -189,6 +189,10 @@
   of identities.
   [(#2192)](https://github.com/PennyLaneAI/catalyst/pull/2192)
 
+  * Added a `merge-ppr` pass that allows merging of PPRs with equivalent angles, and cancelling of
```
lazypanda10117: Also, a note about the naming, `merge_ppr` is very similar to `merge_ppr_ppm`, which makes it hard to distinguish the purpose of these passes. @sengthai 

Maybe something like `merge_and_cancel_ppr` or something a little more distinct.

mlir/lib/QEC/Transforms/MergePPR.cpp
```diff
@@ -0,0 +1,110 @@
...
+        // NOTE: a bit unorthodox, but we find the *second* PPR in a pair, since it's easier to
+        // look backwards by checking inQubits.getDefiningOp
+        ValueRange inQubits = op.getInQubits();
+        auto definingOp = inQubits[0].getDefiningOp();
+
+        if (!definingOp) {
+            return failure();
+        }
+
+        auto prevOp = cast<PPRotationOp>(definingOp);
```
lazypanda10117: I don't think we should use cast here when we are not sure of the type? Use a checking cast like `dyn_cast` might make more sense, and return a failure if the defining op is not of type `PPRotationOp`.

lazypanda10117: On a related note, there should be a lit test to check when the MLIR has `qec.ppr` ioperations nterleaved with other operations, like `qec.ppm`.

mlir/lib/QEC/Transforms/MergePPR.cpp
```diff
@@ -0,0 +1,110 @@
...
+        // for now, the angles must match (otherwise they can't merge or cancel)
+        int16_t opRotation = static_cast<int16_t>(op.getRotationKind());
+        int16_t prevOpRotation = static_cast<int16_t>(prevOp.getRotationKind());
+
+        // cancel inverse operations
+        if (opRotation == -prevOpRotation) {
+            // erase in reverse to avoid use issues
+            ValueRange originalQubits = prevOp.getInQubits();
+            rewriter.replaceOp(op, originalQubits);
+            rewriter.replaceOp(prevOp, originalQubits);
```
lazypanda10117: I am confused why we want to replace the `prevOp` with `originalQubits`? It seems to be `eraseOp` is more appropriate. Same for the case in line 92-93.

lazypanda10117 (commented):
Thanks River! Great PR :) Just have some minor comments

doc/releases/changelog-dev.md
lazypanda10117: We don't need to do this now because this will be part of the merge-rotation pass

doc/releases/changelog-dev.md
lazypanda10117: Same here, this is no longer an issue since we are extending the merge-rotation pass.

mlir/lib/QEC/Transforms/MergePPR.cpp
lazypanda10117: We should also add some integration test in pytest to make sure the `merge-rotation` pass works with PPRs.

mlir/lib/QEC/Transforms/MergePPR.cpp
kipawaa: makes sense, removed in [84fcac7](https://github.com/PennyLaneAI/catalyst/pull/2224/commits/84fcac7095487412ce138c278c3996b7660695ec).

mlir/include/QEC/Transforms/Passes.td
kipawaa: I was mostly trying to follow the convention set in `CommutePPR` and `MergePPRIntoPPM`. Removed in [25e3eb0](https://github.com/PennyLaneAI/catalyst/pull/2224/commits/25e3eb0af7daffa51f7bf2910707d6cfd144cc2d) when adding to `Quantum/Transforms/MergeRotationPatterns`.

mlir/lib/QEC/Transforms/MergePPR.cpp
kipawaa: `eraseOp` is exclusively for deleting operations with _no uses_. I may be wrong, but in the scenario
```
%0 = qec.ppr ["X"](4) %q1
%1 = qec.ppr ["X"](-4) %0
%2 = (some operation) %1
```
where we want to cancel the operations assigned to `%0` and `%1`, `%2`'s assignment counts as a use of `%1`, so `%1` can't be erased. Instead, we replace it with `%q1`, the input qubits to the first operation of the two being cancelled.

Let me know if I'm mistaken in my understanding of how this works! I saw this pattern in `Quantum/Transforms/CancelInversesPatterns.cpp` and it seemed like a good way to handle this.

lazypanda10117: Oh sorry, I wasn't clear in my previous comment. What I want to say is that we can probably use `eraseOp` for  `rewriter.replaceOp(prevOp, originalQubits);` this line only. But I think the use of replaceOp for `op` is necessary.

mlir/test/QEC/MergePPR.mlir
kipawaa: I had a misunderstanding about the regex matching, so I wasn't aware that worked. Thanks for the tip, it's much nicer! Updated in 62f095eb

mlir/test/QEC/MergePPR.mlir
kipawaa: changed in 62f095eb

mlir/lib/QEC/Transforms/MergePPR.cpp
kipawaa: That makes sense! I hadn't considered that, but it's a nice simplification. Added in [8200f4b](https://github.com/PennyLaneAI/catalyst/pull/2224/commits/8200f4b6ddf0e4e04416ba819a6bf2eb8df1b532)

mlir/lib/QEC/Transforms/MergePPR.cpp
kipawaa: changed casting and added a lit test in [38cc27e](https://github.com/PennyLaneAI/catalyst/pull/2224/commits/38cc27e7164a3370ea4bfe6000699327efdfa66e). No integration test since capture etc. are making things difficult.

mlir/lib/QEC/Transforms/MergePPR.cpp
lazypanda10117: awesome, thanks!

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
```diff
@@ -320,6 +322,68 @@ struct MergeRotationsRewritePattern : public OpRewritePattern<OpType> {
     }
 };
 
+struct MergePPRRewritePattern : public OpRewritePattern<PPRotationOp> {
```
sengthai, line 325: We can merge the angle unless the Pauli strings (e.g, "XY") are the same as well.
For example, we cannot perform on %2 and %3 since they have different Paulis.
```mlir
    %1, %2 = qec.ppr ["XZ"](8) %q, %q1: !quantum.bit, !quantum.bit
    %3, %4 = qec.ppr ["YZ"](8) %1, %2: !quantum.bit, !quantum.bit
```

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
```diff
@@ -320,6 +322,68 @@ struct MergeRotationsRewritePattern : public OpRewritePattern<OpType> {
...
+
+        if (!definingOp) {
+            return failure();
+        }
+
+        auto prevOp = dyn_cast<PPRotationOp>(definingOp);
+
+        if (!prevOp) {
+            return failure();
+        }
```
sengthai, line 343: It seems we look at the `second PPR` in a pair based on the first operand of `current PPR` - this is a good start. However, we need to ensure that the `second PPR`'s operands are dominated by the `current PPR`. This would avoid the case like this ---`current PPR` is on `qec.ppr ["YY"](8) %3, %q4` where it acts on different qubits of `qec.ppr ["YY"](8) %1, %Q3`.  
```mlir
%1, %2 = qec.ppr ["XX"](8) %Q0, %Q1: !quantum.bit, !quantum.bit
%3, %4 = qec.ppr ["YY"](8) %1, %Q3: !quantum.bit, !quantum.bit
%5, %6 = qec.ppr ["YY"](8) %3, %q4: !quantum.bit, !quantum.bit
```

sengthai (changes requested):
Thanks for the PR @kipawaa! 👍 
I have a few pieces of feedback that we should address.

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
kipawaa: I believe this is handled correctly and tested for [here](https://github.com/PennyLaneAI/catalyst/pull/2224/files#diff-be53ba335c6d103d2dd6c30f47bfad7ecbabc3520ab675a7d759af50902785dfR818-R829). Let me know if I'm misunderstanding.

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
kipawaa: Good catch! I've added a lit test and fixed this in c03fea2.

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
sengthai: Oh nice!

doc/releases/changelog-dev.md
```diff
@@ -261,6 +261,11 @@
   of identities.
   [(#2192)](https://github.com/PennyLaneAI/catalyst/pull/2192)
 
+  * Added support for PPRs to the `merge_rotations` pass to merge PPRs with
```
```suggestion
  * Added support for PPRs to the :func:`~.passes.merge_rotations` pass to merge PPRs with
```

frontend/test/pytest/test_peephole_optimizations.py
```diff
@@ -493,6 +493,38 @@ def g():
     assert ppm_specs_output["g_0"]["max_weight_pi8"] == 1
 
 
+@pytest.mark.usefixtures("use_capture")
+def test_merge_rotation_ppr():
+    """Test that the merge_rotation pass correctly merges PPRs."""
+    if not qml.capture.enabled():
```
lazypanda10117: I don't think you need this here, `@pytest.mark.usefixtures("use_capture")` should be sufficient?

lazypanda10117: `@pytest.mark.usefixtures("use_capture")` i think this decorator already helps you set up all the capture stuff, so you don't have to manually call `qml.capture.enable()` yourself.

frontend/test/pytest/test_peephole_optimizations.py
```diff
@@ -493,6 +493,38 @@ def g():
...
+    qml.capture.enable()  # capture is needed for PauliRot -> qec.ppr
+
+    my_pipeline = [("pipe", ["enforce-runtime-invariants-pipeline"])]
+
+    @qml.qjit(pipelines=my_pipeline, target="mlir")
+    def test_merge_rotation_ppr_workflow():
+        @qml.transforms.merge_rotations  # have to use qml to be capture-compatible
+        @qml.qnode(qml.device("lightning.qubit", wires=3))
+        def circuit():
+            # equivalent to a Hadamard gate
```
lazypanda10117: what do you mean by equivalent to a Hadamard gate here?

frontend/test/pytest/test_peephole_optimizations.py
```diff
@@ -493,6 +493,38 @@ def g():
...
+            qml.PauliRot(np.pi / 2, pauli_word="XYZ", wires=[0, 1, 2])
+            qml.PauliRot(np.pi / 2, pauli_word="XYZ", wires=[0, 1, 2])
+
+        return circuit()
+
+    ir = test_merge_rotation_ppr_workflow.mlir
+    ir_opt = test_merge_rotation_ppr_workflow.mlir_opt
+    print(ir_opt)
```
lazypanda10117: I don't think we want to print the ir_opt here

frontend/test/pytest/test_peephole_optimizations.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2024 Xanadu Quantum Technologies Inc.
+# Copyright 2024-2025 Xanadu Quantum Technologies Inc.
```
lazypanda10117, line 1: For the copyright stuff, I am actually not sure how we handle this. Do we have to update the year when modifying existing files? If we do have to update, do we provide a range like this, or do we just change the year directly?

lazypanda10117 (approved):
Thanks! I think this looks good now 🚀 Just some minor notes to be addressed

frontend/test/pytest/test_peephole_optimizations.py
kipawaa: @dime10 told me [here](https://github.com/PennyLaneAI/catalyst/pull/2164#discussion_r2482199344) that we update the range on edits, place a single year on new files, so that's what I've been doing 🤷

lazypanda10117: ohh, didn't know that. Thanks!

frontend/test/pytest/test_peephole_optimizations.py
kipawaa: Another relic of my testing 😅 removed in 1906ca07.

frontend/test/pytest/test_peephole_optimizations.py
kipawaa: Woops! Relic of the example I copied this from when struggling with all of the capture/qml/catalyst interactions. Removed in 1906ca07

frontend/test/pytest/test_peephole_optimizations.py
kipawaa: I completely forgot about that, I was only thinking of the mark! Removed the unnecessary setup in 1906ca07.

PR #2227, opened 2025-11-21
Add adjoint lowering pattern for PPR op

paul0403:
**Context:**
When a user places a PPR op in an adjoint qfunc (see #2145 ), the `--adjoint-lowering` pass will not correctly distribute the adjoint into the region onto PPRs, because the pass does not recognize PPRs.

Add pattern for PPR ops in adjoint lowering pass.

**Benefits:**
Correct behavior

**Possible Drawbacks:**
None

### DIFF
--- a/frontend/catalyst/_version.py
+++ b/frontend/catalyst/_version.py
@@ -16,4 +16,4 @@
 Version number (major.minor.patch[-label])
 """
 
-__version__ = "0.14.0-dev37"
+__version__ = "0.14.0-dev38"
--- a/frontend/test/pytest/from_plxpr/test_decompose_transform.py
+++ b/frontend/test/pytest/from_plxpr/test_decompose_transform.py
@@ -229,6 +229,9 @@ class TestGraphDecomposition:
         resources = qml.specs(with_qjit, level="device")(x, y, z)["resources"].gate_types
         assert resources == expected_resources
 
+    @pytest.mark.skip(
+        reason="inconsistent type and error msg across gcc/clang on arm/x86 for undefined symbols"
+    )
     @pytest.mark.usefixtures("use_capture_dgraph")
     def test_gateset_with_rotxzx(self):
         """Test the runtime raises an error if RotXZX is not decomposed."""
@@ -244,7 +247,7 @@ class TestGraphDecomposition:
 
         with pytest.raises(
             OSError,
-            match="undefined symbol: __catalyst__qis__RotXZX",
+            match="undefined symbol",  # ___catalyst__qis__RotXZX
         ):
             qml.qjit(circuit)()
 


mlir/lib/Quantum/Transforms/AdjointPatterns.cpp
```diff
@@ -34,6 +34,7 @@
 #include "mlir/Transforms/DialectConversion.h"
 
 #include "Catalyst/IR/CatalystOps.h"
+#include "QEC/IR/QECDialect.h"
```
paul0403, line 37: changelog

mlir/test/QEC/AdjointTest.mlir
```diff
@@ -0,0 +1,33 @@
+// Copyright 2023 Xanadu Quantum Technologies Inc.
```
paul0403:
```suggestion
// Copyright 2025 Xanadu Quantum Technologies Inc.
```

mlir/lib/Quantum/Utils/QuantumSplitting.cpp
```diff
@@ -204,6 +205,9 @@ void AugmentedCircuitGenerator::generate(Region &region, OpBuilder &builder)
         else if (isa<QuantumDialect>(op.getDialect())) {
             // Any quantum op other than a parametrized gate/insert/extract is ignored.
         }
+        else if (isa<qec::PPRotationOp>(op)) {
```
lazypanda10117, line 208: Just a comment: I suppose when PPR supports arbitrary angle later, we would need to cache this as well?

paul0403: ~~Not really, this was a confusing name. The only op that is a `ParametrizedGate` is unitary~~ :sweat_smile:

paul0403: And this isn't really a "cache" actually.

The input IR with adjoint region looks something like this:
```mlir
func.func(){
  ... stuff
  q.adjoint {
     q.gate
  }
  ... more stuff
}
```

And the output should look like 
```mlir
func.func(){
  ... stuff
  q.gate {adj}
  ... more stuff
}
```

This isn't just a simple use update of SSA values (like merge rotations). This is moving ops across region boundaries. Therefore this needs to be an actual cloning of the ops, from the inner adjoint region to the outer circuit function region.

Now, for a parametrized gate, the classical parameters can be cloned as-is, but the quantum gates need to be cloned in reverse order. The main visitor only deals with the quantum gates. The classical parts are pre-cloned in this "cache". Because the only `ParametrizedGate` is just unitary, the cache is just cloning the matrix, and then the visitor is obtaining the adjoint of the matrix, by creating ops that compute the transpose-conjugate. 

Which means when there's no classical parameters, nothing needs to be pre-cloned. 

In the specific case you're thinking, we can still apply the same general logic as what I did here. It would literally be changing the one-line where I resolve the adjoint in the visitor method:
```
- clone.setRotationKind(ppr.getRotationKind() * (-1)); // what I currently have
+ clone.setAngle(ppr.getAngle() * (-1));  // with arbitrary angle
```

lazypanda10117 (approved):
Looks good to me :) Thanks!

mlir/lib/Quantum/Utils/QuantumSplitting.cpp
lazypanda10117: thanks, this is very helpful : )

sengthai (approved):
Thanks @paul0403 for the PR! 🛠️

PR #2230, opened 2025-11-25
Resource tracking can now handle statepreps

jzaia18:
**Context:**
The `quantum.set_state` and `quantum.set_basis_state` instructions are not currently tracked via runtime resource tracking.

* Adds 2 new functions, `SetState` and `SetBasisState` to the ResourceTracker class, allowing these instructions to be tracked.
* Reports uses of these functions as `StatePrep` and `BasisState` operators respectively

**Benefits:**
Device-level `qml.specs` for qjit'd circuits now reports the number of initial `StatePrep` operators.

**Possible Drawbacks:**

**Related GitHub Issues:**
https://github.com/PennyLaneAI/pennylane/issues/8650
[sc-104048]

### DIFF
--- a/.dep-versions
+++ b/.dep-versions
@@ -12,7 +12,7 @@ enzyme=v0.0.203
 # 'doc/requirements.txt'
 # TODO: uncomment and update to latest version of pennylane
 #       after https://github.com/PennyLaneAI/pennylane/pull/8525 is merged.
-# pennylane=0.44.0.dev31
+pennylane @ git+https://github.com/PennyLaneAI/pennylane@bump-jax-api-hashability
 
 # For a custom LQ/LK version, update the package version here and at
 # 'doc/requirements.txt'
--- a/.github/workflows/build-wheel-linux-arm64.yaml
+++ b/.github/workflows/build-wheel-linux-arm64.yaml
@@ -478,6 +478,11 @@ jobs:
       run: |
         python${{ matrix.python_version }} -m pip install oqc-qcaas-client
 
+    - name: Install PennyLane
+      run: |
+        # Install the specific pennylane version required
+        python${{ matrix.python_version }} -m pip install git+https://github.com/PennyLaneAI/pennylane@bump-jax-api-hashability
+
     - name: Install Catalyst
       run: |
         python${{ matrix.python_version }} -m pip install dist/*.whl --extra-index-url https://test.pypi.org/simple
--- a/.github/workflows/build-wheel-linux-x86_64.yaml
+++ b/.github/workflows/build-wheel-linux-x86_64.yaml
@@ -492,6 +492,11 @@ jobs:
         python${{ matrix.python_version }} -m pip install PennyLane-Lightning-Kokkos
         python${{ matrix.python_version }} -m pip install 'amazon-braket-pennylane-plugin>=1.31.0'
 
+    - name: Install PennyLane
+      run: |
+        # Install the specific pennylane version required
+        python${{ matrix.python_version }} -m pip install git+https://github.com/PennyLaneAI/pennylane@bump-jax-api-hashability
+
     - name: Install OQC client
       if: matrix.python_version != '3.13'
       run: |
--- a/.github/workflows/build-wheel-macos-arm64.yaml
+++ b/.github/workflows/build-wheel-macos-arm64.yaml
@@ -473,6 +473,11 @@ jobs:
         python${{ matrix.python_version }} -m pip install PennyLane-Lightning-Kokkos
         python${{ matrix.python_version }} -m pip install 'amazon-braket-pennylane-plugin>1.27.1'
 
+    - name: Install PennyLane
+      run: |
+        # Install the specific pennylane version required
+        python${{ matrix.python_version }} -m pip install git+https://github.com/PennyLaneAI/pennylane@bump-jax-api-hashability
+
     - name: Install OQC client
       if: matrix.python_version != '3.13'
       run: |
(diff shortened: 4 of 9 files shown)


doc/releases/changelog-dev.md
```diff
@@ -7,6 +7,10 @@
 
 <h3>Improvements 🛠</h3>
 
+* Resource tracking now tracks calls to `SetState` and `SetBasisState`, and can report results
+  that include StatePrep operations.
```
```suggestion
  that include `qml.StatePrep` operations.
```

runtime/lib/backend/null_qubit/NullQubit.hpp
```diff
@@ -224,7 +224,12 @@ struct NullQubit final : public Catalyst::Runtime::QuantumDevice {
      * @param state The state vector data (ignored)
      * @param wires The qubits to prepare (ignored)
      */
-    void SetState(DataView<std::complex<double>, 1> &, std::vector<QubitIdType> &) {}
+    void SetState(DataView<std::complex<double>, 1> &, std::vector<QubitIdType> &wires)
```
maliasadi, line 227: We chatted about making these `wires` arguments constant. Will this be fixed in a follow-up PR?

runtime/lib/backend/null_qubit/ResourceTracker.hpp
```diff
@@ -53,11 +53,13 @@ struct ResourceTracker final {
      * @param name The name of the operation being applied
      * @param wires The wires the operation is being applied to
      * @param controlled_wires The control wires the operation is being applied to
+     * @param impacts_depth Whether the operation impacts the circuit depth
```
maliasadi:
```suggestion
     * @param impacts_depth Whether the operation impacts the circuit depth. Defaults to true.
```

runtime/lib/backend/null_qubit/ResourceTracker.hpp
```diff
@@ -318,6 +320,30 @@ struct ResourceTracker final {
...
     }
 
+    /**
+     * @brief Records a state preparation operation for resource tracking
+     *
+     * Note that as a non-unitary operation, this does not impact circuit depth.
+     *
+     * @param wires The target wires the operation acts upon
+     */
+    void SetState(const std::vector<QubitIdType> &wires)
```
maliasadi, line 326: Just to confirm: Will we record these set-state ops with Lightning devices?

Like, we are expecting to see these ops for a device with [initial_state_prep = true](https://github.com/PennyLaneAI/pennylane-lightning/blob/0a033201252a0b4375848214d868f25c800bd065/pennylane_lightning/lightning_qubit/lightning_qubit.toml#L131C1-L131C26)

runtime/lib/backend/null_qubit/ResourceTracker.hpp
```diff
@@ -318,6 +320,30 @@ struct ResourceTracker final {
...
+    {
+        RecordOperation("StatePrep", wires, {}, false);
+    }
+
+    /**
+     * @brief Records a basis state preparation operation for resource tracking
+     *
+     * Note that as a non-unitary operation, this does not impact circuit depth.
```
maliasadi: Is this compatible with specs without qjit?

runtime/lib/backend/null_qubit/NullQubit.hpp
jzaia18: If we want to make this change I think it'd be best to do in a follow-up PR, it'd require changing the function signature in `QuantumDevice` and probably a few other places as well.

runtime/lib/backend/null_qubit/ResourceTracker.hpp
jzaia18: > with Lightning devices?

Assuming you mean when we use `qml.specs(..., level="device")`. Yes, these initial statepreps still end up calling the `SetState` function and will be counted.

runtime/lib/backend/null_qubit/ResourceTracker.hpp
jzaia18: Could you clarify the question? Nothing within Catalyst should be used if we're running specs without qjit

dime10 (approved):

runtime/lib/backend/null_qubit/NullQubit.hpp
maliasadi: Just wanted to check if this is still on your TODO list :smile: A separate PR makes total sense.

runtime/lib/backend/null_qubit/ResourceTracker.hpp
maliasadi: >      * Note that as a non-unitary operation, this does not impact circuit depth.

I'm just checking here if the depth of the circuit without QJIT will be the same as with QJIT in the `qml.specs` report :thinking:

jzaia18: Just checked and it seems that without `qjit` it does count `StatePrep` toward depth. Do you think it makes more sense to just match the existing behaviour in PL in that case?

maliasadi: We care about consistency of behaviour in these reports at the device level.
cc: @anthayes92

jzaia18: Assuming you meant to tag @AntonNI8 instead? I spoke with him earlier today, we agreed to match the behaviour of counting this gate towards depth.

PR #2232, opened 2025-11-25
Add new operation in QEC dialect support arbitrary-angle PPR

sengthai:
**Context:**
Currently, the `qec.ppr` in the QEC dialect has an angle value expressed as a fraction of π. Here, we add a new op to support arbitrary angles in radians.

- Add new `qec.ppr.arbitrary` op

**Example**
```
        %const = arith.constant 0.14 : f64
        %result = qec.ppr.arbitrary ["X", "Z"](%const) %q0, %q1 : !quantum.bit, !quantum.bit
```

[[sc-104449]]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -212,6 +212,10 @@
   one-shot qnode when `one-shot` mcm method is used.
   [(#2198)](https://github.com/PennyLaneAI/catalyst/pull/2198)
 
+* Fixed a bug where `qml.StatePrep` and `qml.BasisState` might be pushed after other
+  gates, overwriting their effects.
+  [(#2239)](https://github.com/PennyLaneAI/catalyst/pull/2239)
+
 <h3>Internal changes ⚙️</h3>
 
 * Resource tracking now writes out at device destruction time instead of qubit deallocation
--- a/frontend/catalyst/jax_tracer.py
+++ b/frontend/catalyst/jax_tracer.py
@@ -237,7 +237,7 @@ KNOWN_NAMED_OBS = (qml.Identity, qml.PauliX, qml.PauliY, qml.PauliZ, qml.Hadamar
 # Take care when adding primitives to this set in order to avoid introducing a quadratic number of
 # edges to the jaxpr equation graph in ``sort_eqns()``. Each equation with a primitive in this set
 # is constrained to occur before all subsequent equations in the quantum operations trace.
-FORCED_ORDER_PRIMITIVES = {device_init_p, gphase_p}
+FORCED_ORDER_PRIMITIVES = {device_init_p, gphase_p, set_basis_state_p, set_state_p}
 
 PAULI_NAMED_MAP = {
     "I": "Identity",
--- a/frontend/test/lit/test_skip_initial_state_prep.py
+++ b/frontend/test/lit/test_skip_initial_state_prep.py
@@ -73,3 +73,49 @@ def state_prep_example_double():
 #       CHECK:   quantum.set_state
 #   CHECK-NOT:   quantum.set_state
 print(state_prep_example_double.mlir)
+
+
+@qjit(target="mlir")
+@qml.qnode(qml.device("lightning.qubit", wires=2))
+def state_prep_trotter():
+    """
+    Test state prep used with Trotter Product.
+    https://github.com/PennyLaneAI/catalyst/issues/2235
+    """
+    H = qml.Hamiltonian([1, 1], [qml.PauliZ(0), qml.PauliZ(1)])
+    qml.StatePrep([0, 0, 0, 1], wires=[0, 1])
+    qml.Hadamard(wires=2)
+    qml.ctrl(
+        qml.TrotterProduct(H, 1, n=3, order=2),
+        control=2,
+    )
+    return qml.probs(wires=[2])
+
+
+# CHECK-LABEL: func.func public @state_prep_trotter
+# CHECK: quantum.set_state
+# CHECK: quantum.custom "Hadamard"
+print(state_prep_trotter.mlir)
+
+
+@qjit(target="mlir")
+@qml.qnode(qml.device("lightning.qubit", wires=2))
+def basis_state_trotter():
+    """
+    Test basis state used with Trotter Product.
+    https://github.com/PennyLaneAI/catalyst/issues/2235
+    """
+    H = qml.Hamiltonian([1, 1], [qml.PauliZ(0), qml.PauliZ(1)])
+    qml.BasisState([1, 1], wires=[0, 1])
+    qml.Hadamard(wires=2)
+    qml.ctrl(
+        qml.TrotterProduct(H, 1, n=3, order=2),
+        control=2,
+    )
+    return qml.probs(wires=[2])
+
+
+# CHECK-LABEL: func.func public @basis_state_trotter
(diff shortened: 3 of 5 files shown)


mlir/include/QEC/IR/QECDialect.td
```diff
@@ -216,12 +216,15 @@ def PPRotationOp : QEC_Op<"ppr", [QECOpInterface, AttrSizedOperandSegments]> {
...
+        4. A condition (optional) for the operation to be applied only if the condition is true.
         
         The operation returns the same number of qubits as input.
         
         Example:
         ```mlir
-        %result = qec.ppr ["X", "I", "Z"](4) %q0, %q1, %q2 : !quantum.bit, !quantum.bit, !quantum.bit
+        %c0 = arith.constant true : i1
+        %Q_0:3 = qec.ppr ["X", "I", "Z"](4) %q0, %q1, %q2 : !quantum.bit, !quantum.bit, !quantum.bit
+        %Q_1:3 = qec.ppr ["X", "I", "Z"](4) %Q_0[0], %Q_0[1], %Q_0[2] cond(%c0) : !quantum.bit, !quantum.bit, !quantum.bit
```
lazypanda10117, line 227: Does it not support condition before?

mlir/include/QEC/IR/QECDialect.td
```diff
@@ -486,7 +580,10 @@ def PPMeasurementOp : QEC_Op<"ppm", [QECOpInterface, AttrSizedOperandSegments]>
         };
         void setRotationKind(uint16_t attrValue){
             setRotationSign(static_cast<int16_t>(attrValue) < 0 ? -1 : 1);
-        }
+        };
+        bool isArbitrary(){
```
lazypanda10117: I don't think the isArbitrary method is necessary for PPRs as We can easily infer this from the type of the operation (e,g, `qec.ppr.arbitrary`) and I don't think it makes sense to have this for PPM.

mlir/include/QEC/IR/QECDialect.td
```diff
@@ -344,10 +347,101 @@ def PPRotationOp : QEC_Op<"ppr", [QECOpInterface, AttrSizedOperandSegments]> {
...
+
+    let assemblyFormat = [{
+        $pauli_product `(` $arbitrary_angle `)` $in_qubits attr-dict (`cond` `(` $condition^ `)`)? `:` type($out_qubits)
+    }];
+
+    let hasVerifier = 1;
+    let hasCanonicalizeMethod = 1;
+
+    code extraBaseClassDeclaration = [{
+        uint16_t getRotationKind(){
```
lazypanda10117: I would suggest rewriting the interface because these methods do not make sense for PPM and Arbitrary PPR. We can clean this up here or at another PR.

mlir/include/QEC/IR/QECDialect.td
```diff
@@ -344,10 +347,101 @@ def PPRotationOp : QEC_Op<"ppr", [QECOpInterface, AttrSizedOperandSegments]> {
...
+            return 0;  // Arbitrary rotations are not characterized by a rotation kind.
+        };
+        void setRotationKind(uint16_t attrValue){
+            // Convert rotation kind to angle: rotation_kind N means angle = π/N.
+            if (attrValue == 0) {
+                // Invalid rotation kind, set to 0 angle as fallback.
+                // setArbitraryAngle(::llvm::APFloat(0.0));
+            } else {
+                // setArbitraryAngle(::llvm::APFloat(M_PI / static_cast<double>(attrValue)));
+            }
```
lazypanda10117: This function `setRotationKind()` currently does nothing?

mlir/include/QEC/IR/QECDialect.td
```diff
@@ -344,10 +347,101 @@ def PPRotationOp : QEC_Op<"ppr", [QECOpInterface, AttrSizedOperandSegments]> {
         bool isClifford(){
             return hasPiOverTwoRotation() || hasPiOverFourRotation();  // π/2 and π/4 rotations are Clifford
         };
+        bool isArbitrary(){
```
lazypanda10117: Same comment as below.

mlir/include/QEC/IR/QECDialect.td
sengthai: it supports condition before, but I forgot to add that example in docstring.

mlir/include/QEC/IR/QECDialect.td
sengthai: I agree!

mlir/include/QEC/IR/QECDialect.td
sengthai: Nice spot! I've debugged on that, and turned those lines off. it should be :
```c++
            if (attrValue == 0) {
                // Invalid rotation kind, set to 0 angle as fallback.
                setArbitraryAngle(::llvm::APFloat(0.0));
            } else {
                setArbitraryAngle(::llvm::APFloat(M_PI / static_cast<double>(attrValue)));
            }
```

mlir/include/QEC/IR/QECDialect.td
sengthai: actually it be used for interface

mlir/include/QEC/IR/QECDialect.td
sengthai: The interface of both for arbitrary and non-arbitrary ppr, should have this `isArbitrary` for a generic function.

mlir/include/QEC/IR/QECDialect.td
sengthai: oh that's the old code, will update soon.

mlir/include/QEC/IR/QECDialect.td
lazypanda10117: Yea I get it, but what I am trying to say is that this should not be part of the interface. Or at least, PPM should not care about this (i.e, PPM should not use this interface).

sengthai: it makes sense. I removed that function from ppm.

doc/releases/changelog-dev.md
```diff
@@ -115,6 +115,19 @@
 * Dynamically allocated wires can now be passed into control flow and subroutines.
   [(#2130)](https://github.com/PennyLaneAI/catalyst/pull/2130)
 
+* Catalyst now supports arbitrary angle Pauli product rotations in the QEC dialect.
```
rniczh, line 118: I'm wondering why we can't reuse the original `qec.ppr` operation, but introduce one more operation to express the arbitrary angle?

sengthai: As we discussed offline, we will research more about the custom attribute in MLIR, then we will take into account this. For now, we keep adding a new instruction for arbitrary-angle ppr.

dime10: I think it makes sense. They actually represent something quite different, both in terms of cost and semantically. The existing ppr op represents a class of low-level logical operations (pi/8, pi/4, pi/2 rotations), that map to the surface code relatively easily. The rotation angle here identifies the instruction kind and is not considered data. The arbitrary angle PPR operation is a higher level instruction where the angle is a _parameter_ to the operation, i.e. data that can come from the program. It's implementation is much more costly.

sengthai: well explanation! Thanks @dime10!

PR #2233, opened 2025-11-25
Adding frontend support for arbitrary angle PauliRot

lazypanda10117:
**Context:** This is to add support for arbitrary PPRs in the frontend. Users should now be able to call `qml.PauliRot(0.24, "X", wires=[0])` and have it lowered to the corresponding arbitrary angle ppr operation at the MLIR level, namely `qec.ppr.arbitrary`. It should also now support dynamic angle PPRs like in the following example:

```
qml.capture.enable()

@qml.qjit(target="mlir")
def circuit(x: float):
    qml.PauliRot(x, pauli_word="Z", wires=0)
    return qml.probs()
```

**Benefits:** Supports arbitrary angle PPR at the MLIR level, potentially improves performance by deferring discretization to a later stage.

**Related GitHub PR:** This PR contains the frontend changes to support https://github.com/PennyLaneAI/catalyst/pull/2232

[sc-104450]

### DIFF
--- a/frontend/test/lit/test_pauli_rot_and_measure.py
+++ b/frontend/test/lit/test_pauli_rot_and_measure.py
@@ -58,7 +58,7 @@ def test_arbitrary_angle_pauli_rotations():
     qml.capture.enable()
     dev = qml.device("null.qubit", wires=1)
 
-    pipeline = [("pipe", ["quantum-compilation-pipeline"])]
+    pipeline = [("pipe", ["quantum-compilation-stage"])]
 
     @qjit(pipelines=pipeline, target="mlir")
     @qml.qnode(device=dev)
@@ -79,7 +79,7 @@ def test_dynamic_angle_pauli_rotations():
     qml.capture.enable()
     dev = qml.device("null.qubit", wires=1)
 
-    pipeline = [("pipe", ["quantum-compilation-pipeline"])]
+    pipeline = [("pipe", ["quantum-compilation-stage"])]
 
     @qjit(pipelines=pipeline, target="mlir")
     @qml.qnode(device=dev)
@@ -126,7 +126,7 @@ def test_arbitrary_angle_multi_qubit_pauli_rotations():
     qml.capture.enable()
     dev = qml.device("null.qubit", wires=3)
 
-    pipeline = [("pipe", ["quantum-compilation-pipeline"])]
+    pipeline = [("pipe", ["quantum-compilation-stage"])]
 
     @qjit(pipelines=pipeline, target="mlir")
     @qml.qnode(device=dev)
@@ -150,7 +150,7 @@ def test_dynamic_angle_multi_qubit_pauli_rotations():
     qml.capture.enable()
     dev = qml.device("null.qubit", wires=3)
 
-    pipeline = [("pipe", ["quantum-compilation-pipeline"])]
+    pipeline = [("pipe", ["quantum-compilation-stage"])]
 
     @qjit(pipelines=pipeline, target="mlir")
     @qml.qnode(device=dev)
--- a/frontend/test/pytest/test_pauli_rot_and_measure.py
+++ b/frontend/test/pytest/test_pauli_rot_and_measure.py
@@ -42,7 +42,7 @@ def test_pauli_rot_to_ppr():
 @pytest.mark.usefixtures("use_capture")
 def test_pauli_rot_with_arbitrary_angle_to_ppr():
     """Test that Pauli rotation for arbitrary angle."""
-    pipe = [("pipe", ["quantum-compilation-pipeline"])]
+    pipe = [("pipe", ["quantum-compilation-stage"])]
 
     @qjit(pipelines=pipe, target="mlir")
     def test_pauli_rot_with_arbitrary_angle_to_ppr_workflow():
@@ -60,7 +60,7 @@ def test_pauli_rot_with_arbitrary_angle_to_ppr():
 @pytest.mark.usefixtures("use_capture")
 def test_pauli_rot_with_dynamic_angle_to_ppr():
     """Test that Pauli rotation for dynamic angle."""
-    pipe = [("pipe", ["quantum-compilation-pipeline"])]
+    pipe = [("pipe", ["quantum-compilation-stage"])]
 
     @qjit(pipelines=pipe, target="mlir")
     def test_pauli_rot_with_dynamic_angle_to_ppr_workflow():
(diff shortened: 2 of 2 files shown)


frontend/catalyst/from_plxpr/qfunc_interpreter.py
```diff
@@ -592,11 +541,34 @@ def _pauli_rot_bind_call(*invals, op, qubits_len, params_len, ctrl_len, adjoint,
...
-        *wires, theta=theta, pauli_word=pauli_word, qubits_len=qubits_len, adjoint=adjoint
-    )
+    theta = params[0]  # This is a float or a tracer object
+
+    allowed_angles = [np.pi / 4, np.pi / 2, np.pi, -np.pi / 4, -np.pi / 2, -np.pi]
+    is_arbitrary = False
+    if isinstance(theta, jax._src.interpreters.partial_eval.DynamicJaxprTracer) or not np.any(
+        np.isclose(theta, allowed_angles)
+    ):
+        is_arbitrary = True
```
sengthai: How about assigning to `is_arbitrary` directly?

frontend/catalyst/jax_primitives.py
```diff
@@ -1435,8 +1437,10 @@ def _unitary_lowering(
 # pauli rot operation
 #
 @pauli_rot_p.def_abstract_eval
-def _pauli_rot_abstract_eval(*qubits, theta=None, pauli_word=None, qubits_len=0, adjoint=False):
-    qubits = qubits[:qubits_len]
+def _pauli_rot_abstract_eval(
+    *qubits_or_params, pauli_word=None, theta=0, qubits_len=0, params_len=0, adjoint=False
+):
+    qubits = qubits_or_params[:qubits_len]
```
sengthai: Just wondering if the `qubits_or_params` contains `qubits` and `angle`?

frontend/catalyst/from_plxpr/qfunc_interpreter.py
lazypanda10117: what do you mean by that?

lazypanda10117: like `is_arbitrary = isinstance(theta, jax._src.interpreters.partial_eval.DynamicJaxprTracer) or not np.any(np.isclose(theta, allowed_angles))`?

frontend/catalyst/jax_primitives.py
lazypanda10117: For this case, no. For the arbitrary case, yes.

frontend/catalyst/from_plxpr/qfunc_interpreter.py
sengthai: Yes that's right. 🤔

frontend/catalyst/jax_primitives.py
sengthai: If no, should it be named `qubits` instead of `qubits_or_params`?

frontend/catalyst/from_plxpr/qfunc_interpreter.py
lazypanda10117: Fixed :)

frontend/catalyst/jax_primitives.py
lazypanda10117: Renamed

mlir/include/QEC/IR/QECDialect.td
```diff
@@ -344,10 +347,101 @@ def PPRotationOp : QEC_Op<"ppr", [QECOpInterface, AttrSizedOperandSegments]> {
...
             return hasPiOverTwoRotation() || hasPiOverFourRotation();  // π/2 and π/4 rotations are Clifford
         };
+        bool isArbitrary(){
+            return false;
+        };
     }];
     let extraClassDeclaration = extraBaseClassDeclaration;
 }
 
+def PPRArbitraryRotationOp : QEC_Op<"ppr.arbitrary", [QECOpInterface, AttrSizedOperandSegments]> {
```
dime10: Just a note that the name contains "rotation" twice 😅

lazypanda10117: Oops you're right, I will change it to `PPRotationArbitraryOp`

frontend/test/lit/test_pauli_rot_and_measure.py
```diff
@@ -53,6 +53,48 @@ def circuit():
...
+    """Test dynamic angle PauliRot"""
+    qml.capture.enable()
+    dev = qml.device("null.qubit", wires=1)
+
+    pipeline = [("pipe", ["enforce-runtime-invariants-pipeline"])]
+
+    @qjit(pipelines=pipeline, target="mlir")
+    @qml.qnode(device=dev)
+    def circuit(x: float):
+        qml.PauliRot(x, "X", wires=0)
```
sengthai, line 87: Could you include multi-qubit PauliRot as well?

sengthai (approved):
Nice work! 🚀 I have one request in the comment below. @lazypanda10117

frontend/test/lit/test_pauli_rot_and_measure.py
lazypanda10117: Done!

frontend/catalyst/from_plxpr/qfunc_interpreter.py
```diff
@@ -353,6 +355,49 @@ def _pcphase_bind_call(*invals, op, qubits_len, params_len, ctrl_len, adjoint, h
...
+    # invals are the input wires
+    wires = invals[:qubits_len]
+    params = invals[qubits_len : qubits_len + params_len]
+    pauli_word = hyperparameters["pauli_word"]
+    theta = params[0]  # This is a float or a tracer object
+
+    allowed_angles = [np.pi / 4, np.pi / 2, np.pi, -np.pi / 4, -np.pi / 2, -np.pi]
+    is_arbitrary = isinstance(
+        theta, jax._src.interpreters.partial_eval.DynamicJaxprTracer
+    ) or not np.any(np.isclose(theta, allowed_angles))
```
kipawaa, line 370: Before, these were the only acceptable angles to provide to `qml.PauliRot`. Now, if you supply an angle like `4 * np.pi` it will be accepted, but the mlir will generate a `qec.ppr.arbitrary` op, which seems counter-intuitive. I think some sort of handling to either disallow angles beyond $[-\pi, \pi]$ or to take the angle mod $\pi$ would be more appropriate.

lazypanda10117: ~~Good point, lemme add a check for this~~
Update: On second thought, I don't think we should restrict the arbitrary angle to only `[-pi,pi]` since `PauliRot` itself can take on arbitrary angles. This can be handled downstream. For example, the RZ decomposition that we will call later can take in arbitrary angles larger than `pi`. Doing `mod pi` unnecessarily can lead to precision loss, so we should probably do this only when necessary. @kipawaa

kipawaa: Discussed internally, we'll leave this for now and possibly implement changes at a later time.

PR #2234, opened 2025-11-25
 feat: add cluster visualization for `WhileOp`, `IfOp`, `ForOp` 

andrijapau:
This PR adds basic cluster visualization for these control flow operators.

⚠️ The cluster labels are not descriptive at the moment but a future PR should improve upon them (not required for MVP).

```python
@qml.qjit(autograph=True, target="mlir")
@qml.qnode(dev)
def my_workflow(x):
    if x == 2:
        qml.X(0)
    elif x == 3:
        qml.Y(0)
    else:
        qml.Z(0)

    for i in range(3):
        qml.H(0)

    counter = 0 
    while counter < 5:
        qml.RX(0.5, 0)
```
Screenshot 2025-12-10 at 3 05 55 PM

[sc-103462]
[sc-103464]
[sc-103465]

### DIFF
--- a/frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
+++ b/frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
@@ -254,6 +254,5 @@ def _flatten_if_op(op: scf.IfOp) -> list[Region]:
         return flattened_op
 
     # No more nested IfOps, therefore append final region
-    # with no SSAValue
     flattened_op.append(else_region)
     return flattened_op


frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -69,6 +69,72 @@ def _visit_block(self, block: Block) -> None:
...
+    @_visit_operation.register
+    def _for_op(self, operation: scf.ForOp) -> None:
+        """Handle an xDSL ForOp operation."""
+        cluster_id = f"cluster_{id(operation)}"
+        self.dag_builder.add_cluster(
+            cluster_id,
+            node_label="for ...",
+            cluster_id=self._cluster_stack[-1],
+        )
+        self._cluster_stack.append(cluster_id)
```
mudit2812: Let me know what you think, but the `_cluster_stack` seems like something that should be part of the DAGBuilder

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -69,6 +69,72 @@ def _visit_block(self, block: Block) -> None:
...
+            self._cluster_stack.append(cluster_id)
+
+            # Go recursively into the branch to process internals
+            self._visit_region(branch)
+
+            # Pop branch cluster after processing to ensure
+            # logical branches are treated as 'parallel'
+            self._cluster_stack.pop()
+
+        self._cluster_stack.pop()
```
mudit2812: So the IfOp is a cluster, which contains 2 clusters for the condition regions?

andrijapau: Yes

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
andrijapau: Yeah, I'm starting to see a trend that I'm always just popping the last cluster off the stack to use as the parent. I think we can move it to the builder.

mudit2812: Honestly, the more I think about it, it doesn't really matter where it lives. I'll leave it up to your discretion, but with [this suggestion](https://github.com/PennyLaneAI/catalyst/pull/2214/files#r2566148503), it might make sense to just keep it in the director class.

frontend/test/pytest/python_interface/visualization/test_construct_circuit_dag.py
```diff
@@ -217,7 +217,7 @@ def my_workflow():
         cluster_labels = {info["label"] for info in graph_clusters.values()}
         for expected_name in expected_cluster_labels:
             assert expected_name in cluster_labels
-
+
```
```suggestion
```

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
andrijapau: I'm going to leave it in the director class

frontend/test/pytest/python_interface/visualization/test_construct_circuit_dag.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 """Unit tests for the ConstructCircuitDAG utility."""
 
+import re
```
```suggestion
```

frontend/test/pytest/python_interface/visualization/test_construct_circuit_dag.py
```diff
@@ -264,7 +264,7 @@ def my_workflow():
 
         utility = ConstructCircuitDAG(FakeDAGBuilder())
         utility.construct(module)
-
+
```
```suggestion
```

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -85,6 +85,97 @@ def _visit_block(self, block: Block) -> None:
...
+            uid,
+            node_label="for loop",
+            label="",
+            cluster_uid=self._cluster_uid_stack[-1],
+        )
+        self._cluster_uid_stack.append(uid)
+        self._cluster_uid_counter += 1
+
+        for region in operation.regions:
+            self._visit_region(region)
```
mudit2812:
```suggestion
        self._visit_region(operation.regions[0])
```

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -85,6 +85,97 @@ def _visit_block(self, block: Block) -> None:
...
+        num_regions = len(flattened_if_op)
+        for i, (condition_ssa, region) in enumerate(flattened_if_op):
+
+            def _get_conditional_branch_label(i):
+                if i == 0:
+                    return "if"
+                elif i == num_regions - 1:
+                    return "else"
+                else:
+                    return "elif"
```
mudit2812: I'd prefer if this function was defined outside the class's definition.

mudit2812: Definitely no reason for it to be defined for each loop iteration.

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -85,6 +85,97 @@ def _visit_block(self, block: Block) -> None:
...
+            label="",
+            cluster_uid=self._cluster_uid_stack[-1],
+        )
+        self._cluster_uid_stack.append(uid)
+        self._cluster_uid_counter += 1
+
+        for region in operation.regions:
+            self._visit_region(region)
+
+        self._cluster_uid_stack.pop()
```
mudit2812, line 126: Shouldn't there be a cluster for each of the regions of the while loop similar to `IfOp`?

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
mudit2812: It might not even make sense for this to be a function at all (optional)

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -85,6 +85,97 @@ def _visit_block(self, block: Block) -> None:
...
+                label="",
+                style="dashed",
+                penwidth=1,
+                cluster_uid=self._cluster_uid_stack[-1],
+            )
+            self._cluster_uid_stack.append(uid)
+            self._cluster_uid_counter += 1
+
+            # Go recursively into the branch to process internals
+            self._visit_region(region)
```
mudit2812, line 163: For if-elif-else ops, won't this also visit all the inner IfOps that you already flattened?

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -85,6 +85,97 @@ def _visit_block(self, block: Block) -> None:
...
+            label="conditional",
+            labeljust="l",
+            cluster_uid=self._cluster_uid_stack[-1],
+        )
+        self._cluster_uid_stack.append(uid)
+        self._cluster_uid_counter += 1
+
+        # Loop through each branch and visualize as a cluster
+        num_regions = len(flattened_if_op)
+        for i, (condition_ssa, region) in enumerate(flattened_if_op):
```
mudit2812: `condition_ssa` is not being used, so it will not render. Same for the predicate for `WhileOp`s as well. Is that future work?

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -138,3 +229,28 @@ def _func_return(self, operation: func.ReturnOp) -> None:
...
+
+    # Peak into else region to see if there's another IfOp
+    else_block: Block = else_region.block
+    # Completely relies on the structure that the second last operation
+    # will be an IfOp (seems to hold true)
+    if isinstance(else_block.ops.last.prev_op, scf.IfOp):
+        # Recursively flatten any IfOps found in said block
+        nested_flattened_op = _flatten_if_op(else_block.ops.last.prev_op)
+        flattened_op.extend(nested_flattened_op)
+        return flattened_op
```
mudit2812: Shouldn't you be checking if the _first_ op in the `else_block` is an `IfOp` rather than the last [non-`scf.YielfOp`] op?

frontend/test/pytest/python_interface/visualization/test_construct_circuit_dag.py
```diff
@@ -287,3 +287,256 @@ def my_workflow():
...
+        assert clusters["cluster2"]["cluster_label"] == "conditional"
+        assert clusters["cluster2"]["parent_cluster_uid"] == "cluster1"
+
+        # Check three clusters live within conditional
+        assert clusters["cluster3"]["node_label"] == "if"
+        assert clusters["cluster3"]["parent_cluster_uid"] == "cluster2"
+        assert clusters["cluster4"]["node_label"] == "elif"
+        assert clusters["cluster4"]["parent_cluster_uid"] == "cluster2"
+        assert clusters["cluster5"]["node_label"] == "else"
+        assert clusters["cluster5"]["parent_cluster_uid"] == "cluster2"
```
mudit2812: I suspect the `else` cluster will have another cluster inside for the inner `IfOp`.

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
andrijapau: lol. true. 🤦🏼 

Here's a fix: https://github.com/PennyLaneAI/catalyst/pull/2234/commits/20efb546245ae6786a46122f3b82af3ad7650ed3

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
andrijapau: Yes, down the line. But the MVP will not have any predicate processing.

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
andrijapau: I don't think so. Here's an example of a module with if/elif/else,

```
%6 = scf.if %5 -> (!quantum.reg) {
        %7 = "stablehlo.constant"() <{value = dense<0> : tensor<i64>}> : () -> tensor<i64>
        %8 = tensor.extract %7[] : tensor<i64>
        %9 = quantum.extract %4[%8] : !quantum.reg -> !quantum.bit
        %10 = quantum.custom "Hadamard"() %9 : !quantum.bit
        %11 = tensor.extract %7[] : tensor<i64>
        %12 = quantum.insert %4[%11], %10 : !quantum.reg, !quantum.bit
        scf.yield %12 : !quantum.reg
      } else {
        %13 = stablehlo.compare, %arg0, %arg1 : (tensor<i64>, tensor<i64>) -> tensor<i1>
        %14 = tensor.extract %13[] : tensor<i1>
        %15 = scf.if %14 -> (!quantum.reg) {
          %16 = "stablehlo.constant"() <{value = dense<0> : tensor<i64>}> : () -> tensor<i64>
          %17 = tensor.extract %16[] : tensor<i64>
          %18 = quantum.extract %4[%17] : !quantum.reg -> !quantum.bit
          %19 = quantum.custom "PauliX"() %18 : !quantum.bit
          %20 = tensor.extract %16[] : tensor<i64>
          %21 = quantum.insert %4[%20], %19 : !quantum.reg, !quantum.bit
          scf.yield %21 : !quantum.reg
        } else {
          %22 = "stablehlo.constant"() <{value = dense<0> : tensor<i64>}> : () -> tensor<i64>
          %23 = tensor.extract %22[] : tensor<i64>
          %24 = quantum.extract %4[%23] : !quantum.reg -> !quantum.bit
          %25 = quantum.custom "PauliY"() %24 : !quantum.bit
          %26 = tensor.extract %22[] : tensor<i64>
          %27 = quantum.insert %4[%26], %25 : !quantum.reg, !quantum.bit
          scf.yield %27 : !quantum.reg
        }
        scf.yield %15 : !quantum.reg
      }
```

frontend/test/pytest/python_interface/visualization/test_construct_circuit_dag.py
andrijapau: Doesn't look like it, these are the clusters for that example (sorry for the huge chunk of text),

```
{'cluster0': {'attrs': {'fontname': 'Helvetica',
                        'label': 'qjit',
                        'penwidth': 2,
                        'shape': 'rectangle',
                        'style': 'solid'},
              'cluster_label': 'qjit',
              'cluster_uid': None,
              'node_label': None,
              'uid': 'cluster0'},
 'cluster1': {'attrs': {'fontname': 'Helvetica',
                        'label': 'my_workflow',
                        'penwidth': 2,
                        'shape': 'rectangle',
                        'style': 'solid'},
              'cluster_label': 'my_workflow',
              'cluster_uid': 'cluster0',
              'node_label': None,
              'uid': 'cluster1'},
 'cluster2': {'attrs': {'fontname': 'Helvetica',
                        'label': 'conditional',
                        'labeljust': 'l',
                        'penwidth': 2,
                        'shape': 'rectangle',
                        'style': 'solid'},
              'cluster_label': 'conditional',
              'cluster_uid': 'cluster1',
              'node_label': '',
              'uid': 'cluster2'},
 'cluster3': {'attrs': {'fontname': 'Helvetica',
                        'label': '',
                        'penwidth': 1,
                        'shape': 'rectangle',
                        'style': 'dashed'},
              'cluster_label': '',
              'cluster_uid': 'cluster2',
              'node_label': 'if',
              'uid': 'cluster3'},
 'cluster4': {'attrs': {'fontname': 'Helvetica',
                        'label': '',
                        'penwidth': 1,
                        'shape': 'rectangle',
                        'style': 'dashed'},
              'cluster_label': '',
              'cluster_uid': 'cluster2',
              'node_label': 'elif',
              'uid': 'cluster4'},
 'cluster5': {'attrs': {'fontname': 'Helvetica',
                        'label': '',
                        'penwidth': 1,
                        'shape': 'rectangle',
                        'style': 'dashed'},
              'cluster_label': '',
              'cluster_uid': 'cluster2',
              'node_label': 'else',
              'uid': 'cluster5'},
 'cluster6': {'attrs': {'fontname': 'Helvetica',
                        'label': 'setup',
                        'penwidth': 2,
                        'shape': 'rectangle',
                        'style': 'solid'},
              'cluster_label': 'setup',
              'cluster_uid': 'cluster0',
              'node_label': None,
              'uid': 'cluster6'},
 'cluster7': {'attrs': {'fontname': 'Helvetica',
                        'label': 'teardown',
                        'penwidth': 2,
                        'shape': 'rectangle',
                        'style': 'solid'},
              'cluster_label': 'teardown',
              'cluster_uid': 'cluster0',
              'node_label': None,
              'uid': 'cluster7'}}
```

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
andrijapau: I don't think we need that for the MVP. We only want to visualize the quantum operations within the while loop. Am I overlooking something?

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
andrijapau: That's a good question, short answer is no.

The job of `_flatten_if_op` is to flatten this nested tree of `IfOp`s into a flat list of regions (or logical branches) containing the internal operations. All of the nested `IfOp` have been handled (not queued through the dispatcher) by the `_flatten_if_op` function and `_visit_region` will only be visualizing the "internal" operations of each region. 😅 

Hopefully that makes some sense.

frontend/test/pytest/python_interface/visualization/test_construct_circuit_dag.py
andrijapau: Note that setup and teardown are from the FuncOp PR and it's a known "issue". I need to make a PR to remove adding them as clusters. 😅

frontend/test/pytest/python_interface/visualization/test_construct_circuit_dag.py
```diff
@@ -287,3 +287,256 @@ def my_workflow():
...
+                qml.Z(0)
+
+        args = (1, 2)
+        module = my_workflow(*args)
+
+        utility = ConstructCircuitDAG(FakeDAGBuilder())
+        utility.construct(module)
+
+        clusters = utility.dag_builder.clusters
+        clusters = utility.dag_builder.clusters
```
mehrdad2m, line 510:
```suggestion
```

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -138,3 +225,28 @@ def _func_return(self, operation: func.ReturnOp) -> None:
...
+    # Save condition SSA in case we want to visualize it eventually
+    flattened_op: list[tuple[SSAValue | None, Region]] = [(condition_ssa, then_region)]
+
+    # Peak into else region to see if there's another IfOp
+    else_block: Block = else_region.block
+    # Completely relies on the structure that the second last operation
+    # will be an IfOp (seems to hold true)
+    if isinstance(else_block.ops.last.prev_op, scf.IfOp):
+        # Recursively flatten any IfOps found in said block
+        nested_flattened_op = _flatten_if_op(else_block.ops.last.prev_op)
```
mehrdad2m: This is very specific. Is there any other way to achieve this without this assumption?

mehrdad2m: I actually was running this example and I feel like it is working because it violates this assumption:
```python
@xdsl_from_qjit
@qml.qjit(autograph=True, target="mlir")
@qml.qnode(dev)
def my_workflow(x):
    if x == 1:
        qml.X(0)
    elif x == 2:
        qml.Y(0)
    else:
        qml.Z(0)
        if x == 3:
            qml.Z(0)
        elif x == 4:
            qml.Z(0)
        else:
            qml.Z(0)
``` 
When running this, I only see one cluster (I assume there should be two right?)

mehrdad2m (commented):
Thanks @andrijapau, looks good. I only have some concerns regarding the flattening logic.

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -138,3 +225,28 @@ def _func_return(self, operation: func.ReturnOp) -> None:
             # If we hit a func.return operation we know we are leaving
             # the FuncOp's scope and so we can pop the ID off the stack.
             self._cluster_uid_stack.pop()
+
+
+def _flatten_if_op(op: scf.IfOp) -> list[tuple[SSAValue | None, Region]]:
+    """Recursively flattens a nested IfOp (if/elif/else chains)."""
+
+    condition_ssa: SSAValue = op.operands[0]
```
mehrdad2m: Why do we return this? It seems to be not used anywhere.

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
andrijapau: Ah good counter example. Let me take a look.

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
andrijapau: I can take it out for now as the MVP won't display any predicates. Would that be better?

CC @mudit2812 too as you had the same question.

I kept it there as it will probably be needed down the line.

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
andrijapau: > When running this, I only see one cluster (I assume there should be two right?)

Yeah, I think there should be two "conditional" clusters each with three sub-clusters for if/elif/else

andrijapau: Note, I changed the circuit so each region has a unique operation,
```
@qml.qjit(autograph=True, target="mlir")
@qml.qnode(dev)
def my_workflow(x):
    if x == 1:
        qml.X(0)
    elif x == 2:
        qml.Y(0)
    else:
        qml.Z(0)
        if x == 3:
            qml.H(0)
        elif x == 4:
            qml.RY(0, 0)
        else:
            qml.RZ(0, 0)
```

image

To me this diagram would make? All I did was check if any quantum operations are present - if so, don't flatten. 😬 Feels hacky but works.

andrijapau: But now things like this,
```
@qml.qjit(autograph=True, target="mlir")
@qml.qnode(dev)
def my_workflow(x):
    if x == 1:
        qml.X(0)
    elif x == 2:
        qml.Y(0)
    else:
        for i in range(3):
            qml.Z(0)
        if x == 3:
            qml.H(0)
        elif x == 4:
            qml.RY(0, 0)
        else:
            qml.RZ(0, 0)
```
don't work properly either 😓. I'd have to recursively check inside the else region if there's anything that needs to be visualized.

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
mudit2812: Nope, you answered my question here as well as in [this response](https://github.com/PennyLaneAI/catalyst/pull/2234#discussion_r2600773932). Resolving.

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
mudit2812: Interesting. I half expected the predicates to be computed outside the first `IfOp`, which is why I thought the nested ifs would be the first ops inside the outer if op's regions. Regardless, for if-elif-else, you're right that the last non-yield op will be IfOp. Resolving :)

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
mudit2812: You're right. Thanks for the detailed response :). Resolving

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
mudit2812: I think it's worth keeping the option open to just visualize if-elif-else as if-else(if-else) for the MVP. Thanks for catching this @mehrdad2m , this seems like a pretty significant edge case.

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
andrijapau: I decided to take it out. [KISS](https://en.wikipedia.org/wiki/KISS_principle). 

https://github.com/PennyLaneAI/catalyst/pull/2234/commits/c2bd854cd21a214e00a1c4daf5ecc22df584c70e

andrijapau:
> I will approve to not block the PR, I think I looks good as long as the unresolved conversation is fixed.

Thanks, I think the fix is already in place. I posted in slack [here](https://xanaduhq.slack.com/archives/C08MQQ53U15/p1765306050897719) to see product's opinion.

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
andrijapau: Should be fixed now, feel free to take a look @mudit2812

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -227,26 +224,36 @@ def _func_return(self, operation: func.ReturnOp) -> None:
...
+        if isinstance(op, scf.IfOp):
+            nested_if_op = op
+            # No need to walk this op as this will be
+            # recursively handled down below
+            continue
+        for internal_op in op.walk():
+            if type(internal_op) in quantum.Quantum.operations:
+                has_quantum_ops = True
+                # No need to check anything else
+                break
```
mudit2812, line 248: Why are you doing a walk and not just this:
```suggestion
        if type(op) in quantum.Quantum.operations:
            has_quantum_ops = True
            # No need to check anything else
            break
```

andrijapau: For scenarios where quantum operations are nested inside other operations,
```
def my_workflow(x):
    if x == 1:
        qml.X(0)
    elif x == 2:
        qml.Y(0)
    else:
        for i in range(3): <----- this wouldn't get rendered
            qml.X(0)
        if x == 3:
            qml.H(0)
        elif x == 4:
            qml.RY(0, 0)
        else:
            qml.RZ(0, 0)
```

mudit2812: Good point. Resolving.

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
andrijapau: Resolving as approvals were given.

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
andrijapau: Resolving as approvals were given.

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -138,3 +222,38 @@ def _func_return(self, operation: func.ReturnOp) -> None:
...
+
+    if nested_if_op and not has_quantum_ops:
+        # Recursively flatten any IfOps found in said block
+        nested_flattened_op: list[Region] = _flatten_if_op(nested_if_op)
+        flattened_op.extend(nested_flattened_op)
+        return flattened_op
+
+    # No more nested IfOps, therefore append final region
+    # with no SSAValue
```
andrijapau:
```suggestion
    # No more nested IfOps, therefore append final region
```

PR #2238, opened 2025-11-26
Implement new `mlir_specs` Python function to get resource counts from MLIR passes

jzaia18:
**Context:**
There is currently no way to view the impact to runtime costs of a given MLIR pass from the python frontend of PennyLane.

Creates a new `mlir_specs` function, in addition to a `specs_collect` function which uses xDSL to allow inspection of compilation passes written in MLIR.

**Benefits:**
Allows users to evaluate the impact on the IR of various MLIR passes.

**Possible Drawbacks:**
Not integrated with `qml.specs()`, this will be handled in a followup PR.

**Related GitHub Issues:**
[sc-103510]
Migrated from https://github.com/PennyLaneAI/pennylane/pull/8660

See also the frontend integration for this PR: https://github.com/PennyLaneAI/pennylane/pull/8606

### DIFF
--- a/frontend/catalyst/_version.py
+++ b/frontend/catalyst/_version.py
@@ -16,4 +16,4 @@
 Version number (major.minor.patch[-label])
 """
 
-__version__ = "0.14.0-dev52"
+__version__ = "0.14.0-dev53"


frontend/catalyst/python_interface/pass_api/compiler_transform.py
```diff
@@ -53,6 +53,9 @@ def compiler_transform(module_pass: ModulePass) -> PassDispatcher:
     """Wrapper function to register xDSL passes to use with QJIT-ed workflows."""
     dispatcher = PassDispatcher(module_pass)
 
+    # TODO: This is a temporary workaround and shouldn't be needed long-term
```
jzaia18: This is needed to determine whether a given pass is a tape transform or an MLIR pass, which is needed for the `qml.specs` integration. Once [this PennyLane PR](https://github.com/PennyLaneAI/pennylane/pull/8539) is merged, we no longer need to use this workaround.

frontend/catalyst/python_interface/inspection/specs_collector.py
```diff
@@ -0,0 +1,544 @@
...
+### Subroutine calls
+############################################################
+
+
+@handle_resource.register
+def _(
+    xdsl_op: func.CallOp,
+) -> tuple[ResourceType, str]:
+    # If these types are matched, parse them with the extra specs handler
+    return ResourceType.FUNC_CALL, xdsl_op.callee.string_value()
```
dime10, line 278: Does this mean a func call is just counted as a "func" type resource, or a we recursively traversing call-graphs to propagate its resources up to the parent? 
(looks awesome btw!)

jzaia18: We do both! I gather up the number of calls to each other function and then at the end add in the used resources of each subroutine (recursively).

dime10: Nice!

frontend/catalyst/python_interface/inspection/xdsl_conversion.py
```diff
@@ -372,3 +422,46 @@ def xdsl_to_qml_measurement(op, *args, **kwargs) -> MeasurementProcess | Operato
...
+        A string representing the PennyLane measurement.
+    """
+
+    if op.name == "quantum.measure":
+        gate_name = "MidMeasure"
+
+    elif op.name == "quantum.compbasis":
+        # Defines a pseudo-observable to represent measurements in the computational basis
+        # Used within e.g. `probs()`
+        gate_name = f"{len(op.qubits)} wires"
```
jzaia18: TODO: Would be better if this said "all wires" instead of "0 wires" to mean all wires

frontend/catalyst/python_interface/inspection/specs.py
```diff
@@ -0,0 +1,101 @@
...
+    from catalyst.jit import QJIT
+
+
+class StopCompilation(Exception):
+    """Custom exception to stop compilation early when the desired specs level is reached."""
+
+
+def mlir_specs(
+    qnode: QJIT, level: int | tuple[int] | list[int] | Literal["all"], *args, **kwargs
+) -> ResourcesResult | dict[str, ResourcesResult]:
```
mudit2812, line 36: I was confused initially because this signature implies that we're doing away with the wrapper-style implementation. However, this will be "private" and eventually accessed through `qml.specs`, so I see why you've implemented it this way.

frontend/catalyst/python_interface/inspection/specs.py
```diff
@@ -0,0 +1,101 @@
...
+    """Compute the specs used for a circuit at the level of an MLIR pass.
```
mudit2812, line 37: In the case where there are both tape transforms and MLIR/xDSL transforms in the transform program, is the case that the `<mlir_specs>level` argument will be equal to `<specs>level - num_tape_transforms`? I assume yes based on the description of `level` in the docstring. Just checking.

jzaia18: Yes, that's correct. And from this function's POV the 1st level is always 0, that translation happens within `qml.specs` itself.

frontend/catalyst/python_interface/inspection/specs.py
```diff
@@ -0,0 +1,101 @@
...
+            pass_name if pass_level else "Before MLIR Passes",
+        )
+
+        if max_level is not None and pass_level >= max_level:
+            raise StopCompilation("Stopping compilation after reaching max specs level.")
+
+    mlir_module = get_mlir_module(qnode, args, kwargs)
+    try:
+        Compiler.run(mlir_module, callback=_specs_callback)
+    except StopCompilation:
```
mudit2812, line 84: This is pretty neat!

frontend/catalyst/python_interface/inspection/specs.py
```diff
@@ -0,0 +1,101 @@
...
+    elif isinstance(level, (tuple, list)):
+        max_level = max(level)
+    elif not isinstance(level, int):
+        raise ValueError("The `level` argument must be an int, a tuple/list of ints, or 'all'.")
+
+    def _specs_callback(previous_pass, module, next_pass, pass_level=0):
+        """Callback function for gathering circuit specs."""
+
+        pass_instance = previous_pass if previous_pass else next_pass
+        ops = specs_collect(module)
```
mudit2812: Might be worth renaming `ops` since it is a `ResourcesResult` instance. The name implies it is some sort of sequence of operations. Maybe `result`?
```suggestion
        result = specs_collect(module)
```

frontend/catalyst/python_interface/inspection/xdsl_conversion.py
```diff
@@ -334,6 +341,49 @@ def xdsl_to_qml_op(op) -> Operator:
...
+        "quantum.gphase": "GlobalPhase",
+        "quantum.multirz": "MultiRZ",
+        "quantum.set_basis_state": "BasisState",
+        "quantum.set_state": "StatePrep",
+        "quantum.unitary": "QubitUnitary",
+    }
+
+    if op.name == "quantum.custom":
+        gate_cls = resolve_gate(op.properties.get("gate_name").data)
+        gate_name = gate_cls.__name__
```
mudit2812: This should be good enough right? Let me know if I'm missing something.
```suggestion
        gate_name = op.properties.get("gate_name").data
```

frontend/catalyst/python_interface/inspection/xdsl_conversion.py
```diff
@@ -334,6 +341,49 @@ def xdsl_to_qml_op(op) -> Operator:
     return _apply_adjoint_and_ctrls(gate, op)
 
 
+def xdsl_to_qml_op_type(op, adjoint_mode: bool) -> str:
```
mudit2812: `adjoint_mode` is not in the args list of the docstring.

mudit2812: Also, I expected the return type of this function to be `type[qml.operation.Operator]` based on the name. Maybe the following name makes more sense?
```suggestion
def xdsl_to_qml_op_name(op, adjoint_mode: bool) -> str:
```

frontend/catalyst/python_interface/inspection/xdsl_conversion.py
```diff
@@ -372,3 +422,50 @@ def xdsl_to_qml_measurement(op, *args, **kwargs) -> MeasurementProcess | Operato
...
+        ops_list = [xdsl_to_qml_measurement_type(term.owner) for term in op.terms]
+        gate_name = f"Hamiltonian({', '.join(ops_list)})"
+
+    elif op.name == "quantum.tensor":
+        ops_list = [xdsl_to_qml_measurement_type(operand.owner) for operand in op.operands]
+        gate_name = " @ ".join(ops_list)
+
+    elif op.name == "quantum.namedobs":
+        gate_cls = resolve_gate(op.type.data.value)
+        gate_name = gate_cls.__name__
```
mudit2812: Same comment as earlier. Isn't the following fine?
```suggestion
        gate_name = op.type.data.value
```

frontend/catalyst/python_interface/inspection/xdsl_conversion.py
```diff
@@ -372,3 +422,50 @@ def xdsl_to_qml_measurement(op, *args, **kwargs) -> MeasurementProcess | Operato
 
         case _:
             raise NotImplementedError(f"Unsupported measurement/observable: {op.name}")
+
+
+def xdsl_to_qml_measurement_type(op, obs_op=None) -> str:
```
```suggestion
def xdsl_to_qml_measurement_name(op, obs_op=None) -> str:
```

mudit2812: And `obs_op` is missing from the docstring.

frontend/catalyst/python_interface/inspection/specs_collector.py
mudit2812, line 1: This looks awesome!

frontend/test/pytest/python_interface/inspection/test_mlir_specs.py
```diff
@@ -0,0 +1,589 @@
...
+"""Unit test module for the mlir_specs function in the Python Compiler inspection module."""
+
+
+import pytest
+
+pytestmark = pytest.mark.external
+
+pytest.importorskip("xdsl")
+pytest.importorskip("catalyst")
+pytest.importorskip("jax")
```
mudit2812:
```suggestion
pytestmark = pytest.mark.requires_xdsl
```

frontend/catalyst/python_interface/inspection/xdsl_conversion.py
jzaia18: Not sure why I didn't just do that to begin with tbh haha. 0e4cd8d31

frontend/catalyst/python_interface/inspection/specs.py
jzaia18: 0e4cd8d31

frontend/test/pytest/python_interface/inspection/test_mlir_specs.py
```diff
@@ -0,0 +1,589 @@
...
+# pylint: disable=wrong-import-position
+import jax.numpy as jnp
+import pennylane as qml
+
+# pylint: disable=wrong-import-position
+import catalyst
+from catalyst.passes.xdsl_plugin import getXDSLPluginAbsolutePath
+from catalyst.python_interface.inspection import ResourcesResult, mlir_specs
+
+# pylint: disable=implicit-str-concat, unnecessary-lambda
```
mudit2812: You can get rid of the wrong-import-position warnings after the `importorskip` removal :)

frontend/test/pytest/python_interface/inspection/test_mlir_specs.py
```diff
@@ -0,0 +1,584 @@
...
+
+
+# TODO: Add a test that calls another function as a subroutine
+
+
+@pytest.mark.usefixtures("use_capture")
+class TestMLIRSpecsWithPLXPR(TestMLIRSpecs):
+    """Unit tests for the mlir_specs function in the Python Compiler inspection module with plxpr enabled."""
+
+    use_plxpr = True
```
mudit2812: Is this a TODO?

jzaia18: No, it inherits all the tests from the parent class. But I should document this better

mudit2812: Oh I just realized what you've done. Couldn't you also use a fixture to toggle `use_plxpr` with just the base `TestMLIRSpecs` class?

mudit2812: There is a `use_both_frontend` fixture that is relevant.

jzaia18: I remember trying this and realizing that there were a few issues. I think one might have been that for some tests I need them to behave slightly differently based on whether or not they're using capture and there was no way to retrieve this info from the fixture(?). Either that or it had something to do with the capture enabling/disabling at a weird time between tests that broke things. Let me try this again and see if its still an issue

jzaia18: Just realizing I could always use `qml.capture.enabled()` to check, so that probably isn't it

jzaia18: Okay this seems to work, no clue what I was doing wrong before ¯\\_ (ツ)_/¯. Fixed in 02bbdfc2

mudit2812 (commented):
Looks great @jzaia18 . I don't see any major blockers. The only suggestion I have is to potentially break up the content in `specs_collector` to smaller functions, specifically the `_collect_region` function. It seems to me like it could be broken down into 2 functions that:
* Handles operations with regions (control flow, adjoint)
* Handles the various `ResourceType`s for non-region ops.

frontend/test/pytest/python_interface/inspection/test_mlir_specs.py
```diff
@@ -0,0 +1,592 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Unit test module for the mlir_specs function in the Python Compiler inspection module."""
+
+import jax.numpy as jnp
+import pennylane as qml
+import pytest
+
+import catalyst
+from catalyst.passes.xdsl_plugin import getXDSLPluginAbsolutePath
```
mudit2812: Btw we no longer need to specify the pass plugin when using xdsl passes. I would suggest doing one of the two things:
* Remove the usage of the pass plugin entirely
* Parametrize the tests so that they run both with and without specifying the pass plugin.

I'm in favour of the first option, since we may just get rid of the plugin entirely in the future, and w're already doing that for other tests in [this PR](https://github.com/PennyLaneAI/pennylane/pull/8664).

I would consider at least one of the above changes to be blocking.

jzaia18: Removed in [3947e05](https://github.com/PennyLaneAI/catalyst/pull/2238/commits/3947e059f7d8c9b10d01449ffa4abc06dd4e348e)

frontend/catalyst/python_interface/inspection/specs_collector.py
```diff
@@ -0,0 +1,552 @@
...
+
+        self.classical_instructions: dict[str, int] = defaultdict(int)
+        self.function_calls: dict[str, int] = defaultdict(int)
+        self._unresolved_function_calls: dict[str, int] = defaultdict(int)
+
+        self.device_name = None
+        self.num_allocs = 0  # The total number of distinct qubits allocated in this region
+
+    def merge_with(self, other: "ResourcesResult", method: str = "sum") -> None:
+        """Merge another ResourcesResult into this one."""
```
andrijapau, line 129: Optional but do we want to add some validation on the type for other?

frontend/catalyst/python_interface/inspection/specs_collector.py
```diff
@@ -0,0 +1,552 @@
...
+            self.function_calls[name] = merge_func(self.function_calls[name], count)
+        for name, count in other._unresolved_function_calls.items():
+            self._unresolved_function_calls[name] = merge_func(
+                self._unresolved_function_calls[name], count
+            )
+
+        self.device_name = self.device_name or other.device_name
+        self.num_allocs = merge_func(self.num_allocs, other.num_allocs)
+
+    def multiply_by_scalar(self, scalar: int) -> None:
```
andrijapau, line 161: Any reason you didn't go for modifying the mul dunders?

frontend/catalyst/python_interface/inspection/specs_collector.py
```diff
@@ -0,0 +1,552 @@
...
+
+            resources.merge_with(used_resources)
+            continue
+
+        _collect_operation(resources, op, adjoint_mode=adjoint_mode)
+
+    return resources
+
+
+def specs_collect(module) -> ResourcesResult:
```
andrijapau: + `from xdsl.dialects.builtin import ModuleOp`
```suggestion
def specs_collect(module: ModuleOp) -> ResourcesResult:
```

frontend/catalyst/python_interface/inspection/specs_collector.py
jzaia18: I didn't think to tbh, but I also this I prefer the explicit function call for readability

frontend/catalyst/python_interface/inspection/specs_collector.py
jzaia18: [5bf82d8](https://github.com/PennyLaneAI/catalyst/pull/2238/commits/5bf82d8a320ac4f0641caae59b16d5c01b49d805)

frontend/catalyst/python_interface/inspection/specs_collector.py
jzaia18: [5bf82d8](https://github.com/PennyLaneAI/catalyst/pull/2238/commits/5bf82d8a320ac4f0641caae59b16d5c01b49d805)

frontend/test/pytest/python_interface/inspection/test_mlir_specs.py
```diff
@@ -0,0 +1,592 @@
...
+            measurements={"probs(all wires)": 1},
+            num_allocs=2,
+            function_calls={"extra_function": 1},
+        )
+
+        res = mlir_specs(circ, level=0)
+        assert resources_equal(res, expected)
+
+
+# TODO: In the future, it would be good to add unit tests for specs_collector instead of just
```
andrijapau, line 627: :)

andrijapau: One can only dream

jzaia18:
Awesome! Thanks for the reviews. Will wait to merge this until after the migration PR gets merged to main.

PR #2241, opened 2025-11-27
Rename annotate-function pass and move it to the gradient dialect

mehrdad2m:
**Context:**
The `annotate-functions `pass marks functions that cannot be automatically differentiated. It scans functions for operations incompatible with automatic gradients (quantum measurements, certain custom calls, and callbacks without custom derivatives) and annotates those functions with a `invalidGradientOperation`. It also propagates the annotation up the call graph, so if a function calls an annotated function, it gets annotated too. Later, when verifiers for gradient operations are run they will fail with a clear error message. 
Currently this pass is in the quantum dialect and grouped with other quantum compilation passes. 

This PR improves the naming of the pass. Move it to the gradient dialect and schedule within the `lower-gradients` compilation stage 

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


mlir/include/Gradient/Transforms/annotate_invalid_gradient_functions.h
```diff
@@ -14,9 +14,9 @@
 #pragma once
 
 namespace catalyst {
-namespace quantum {
+namespace gradient {
 
-static constexpr const char *hasInvalidGradientOp = "catalyst.invalidGradientOperation";
+static constexpr const char *hasInvalidGradientOp = "gradient.invalidGradientOperation";
```
dime10: This is minor, but I would almost be inclined to leave the attribute with a catalyst prefix (independent of the cpp namespace), because that it makes it clearer what library it's from (and gradient is already in the suffix)

mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/2241/commits/0a73958c5eff6de107c61ca2ab6f2007e9d4930b

PR #2244, opened 2025-11-28
Do not rename external func decls during --inline-nested-modules

paul0403:
**Context:**
One of the first steps in the compilation pipeline is `--inline-nested-modules`. This pass lifts the per-qnode module up into the global module of the qjit.

Because (usually) the symbol table of a module is isolated from above, multiple modules in parallel may contain the same names inside. Therefore when inlining child modules into the global module, we need to rename them to unique names. 

However, there's a small problem. Each qnode module has their own specific transform sequence, so the quantum passes (applied during `--apply-transform-sequence`) have to be resolved before `--inline-nested-module`. If a pass generates a function declaration to an external API (e.g. runtime functions), that name must not be altered.

Do not perform renaming on external function declarations from within the qnode modules during `--inline-nested-modules`.
Only inline the first occurrence of such func decls from the qnode modules into the global qjit module.

**Benefits:**
Quantum passes that generate calls to other APIs can work.  
The flipside is also true! API developers do not have to maintain multiple aliases for their API functions (e.g. gridsynth pass #2140 )

[sc-105020]

### DIFF
--- a/mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp
+++ b/mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp
@@ -240,8 +240,8 @@ LogicalResult RenameFunctionsPattern::matchAndRewrite(Operation *child,
                 // We should not rename external function declarations, as they can be
                 // names required by other APIs.
                 // We record these external func decls during the rename pattern.
-                // Then during the actual inlining stage, only the first occurrence of the per-module
-                // func decls of these external decls should be inlined.
+                // Then during the actual inlining stage, only the first occurrence of the
+                // per-module func decls of these external decls should be inlined.
                 if (auto f = dyn_cast<func::FuncOp>(op)) {
                     if (f.isExternal()) {
                         _externalFuncDeclNames->insert(f.getName());


mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp
```diff
@@ -466,6 +537,31 @@ struct InlineNestedSymbolTablePass : PassWrapper<InlineNestedSymbolTablePass, Op
...
+
+        RewritePatternSet updateCalleeToExternalAPINames(context);
+        updateCalleeToExternalAPINames.add<UpdateCalleeToExternalAPINamesPattern>(
+            context, uniqued_names_to_APIName);
+        run = _stopAfterStep >= 5 || _stopAfterStep == 0;
+        if (run && failed(applyPatternsGreedily(
+                       symbolTable, std::move(updateCalleeToExternalAPINames), config))) {
+            signalPassFailure();
+        }
+
```
dime10: Wow there is a lot of invocations of the greedy pattern applicator in this pass. I think this may indicate a misuse of the infrastructure.
Patterns are meant to be small rewrites that are independent, composable, and repeatable (typically).
If this is not the case for a pass, and instead we want to apply a _specific sequence_ of rewrites _one time_, then we should do so directly which will be much more efficient.

I'm not saying this is 100% the case here, but it definitely looks suspicious.

mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp
```diff
@@ -228,8 +230,18 @@ LogicalResult RenameFunctionsPattern::matchAndRewrite(Operation *child,
...
+                }
+
+                // We should not rename external function declarations, as they can be
+                // names required by other APIs.
+                // During inlining, on-the-fly they still need to be renamed, otherwise module
+                // verifier complains. So we save the original external API name and rename
+                // them back after inlining is finished.
+                if (auto f = dyn_cast<FunctionOpInterface>(op); f.isExternal()) {
+                    f->setAttr("original_external_API_name", rewriter.getStringAttr(f.getName()));
+                }
```
dime10: Thanks for catching this Paul!

Approach seems ok, although perhaps a bit hacky. Have you considered wether it's possible to directly deduplicate the external functions while we're inlining the module? (instead of rename -> restore)

paul0403: I tried this, but it was a bit difficult...

The main issue is how Erick structured this pass into multiple independent stages, each with its own pattern (aka your other comment above). One of the stages/patterns is responsible for the renaming. My very first commit on this PR is actually just skipping the renaming for external function declarations.

But then, one of the next stages/patterns updates the call ops to call those newly name-uniqued declarations. This would now involve passing some helper containers across different rewrite pattern struct! Because I would need the renaming pattern struct to record a list of external names, so that the call update pattern can know to leave those untouched.

This might be a problem inside the pass code, so it's not that big of a dealbreaker. But there's actually something more serious at the IR level. Consider this IR (which is the structure that will be added by Gridsynth pass on a qjitted module with two qnodes)
```mlir
module @global {

  module @local0 {
    func.func private @f()
    func.func @main() {
      func.call @f() : () -> ()
      return
    }
  }

  module @local1 {
    func.func private @f()
    func.func @main() {
      func.call @f() : () -> ()
      return
    }
  }

}
```

When inlining the two `f` declarations, if we do not rename them to `f_0` and `f_1`, then the inlining itself will fail! The inlining of the second `f` will say it clashes with an existing `f` in the parent module (result of inlining the first `f`). 

So what about not inling the second `f`, but just discard it? Well, that discard can't even happen, because regions are isolated from above, so if we try to discard it, its call in its module will say it is calling an undefined symbol! So we have to inline that second call first (no renaming required: the callee and the inlined first decl are both `f` in this approach). But now this would start to involve big refactoring of Erick's original structure, since now we are inling call ops at the renaming stage.

I very much agree that this structure is quite weird. This might actually be one of these situations where a raw `runOnOperation()` with raw logic is more suited then rewrite patterns. But I don't want to spend too much time refactoring, since 
- (a) gridsynth is hard-blocked by this (unless we are good with attaching `_0`s to all of the runtime solver function names, which I think obviously we shouldn't)
- (b) the new quantum scope abstraction proposal might make this entire pass obsolete anyway

dime10: Thank you for the in-depth analysis!

> So what about not inling the second f, but just discard it? Well, that discard can't even happen, because regions are isolated from above, so if we try to discard it, its call in its module will say it is calling an undefined symbol!

This I don't quite follow. If `f` already exists in the scope we are inlining into (from inlining the first module), why do we have to inline the second `f` in order to inline the second module? I would imagine that it's kind of like our `getOrCreateFunction` helpers, where we would only inline if it doesn't already exist. Unless we don't have granular control over the inlining, i.e. we are not doing the inlining ourself but have to do the entire module at once.

> so if we try to discard it, its call in its module will say it is calling an undefined symbol!

By discard you mean deleting it before we inline? Generally we should be allowed to create invalid IR as an intermediate step (in fact I don't think you can do a lot of meaningful rewrites without this), and the only requirement is that the IR is legal after a pass is run no? 

> But I don't want to spend too much time refactoring

Absolutely, if this is too complicated then I'm happy to proceed with the do-undo approach :)

paul0403: I got it to work, now we don't have to rename and undo rename via a discardable in-between attribute :partying_face: https://github.com/PennyLaneAI/catalyst/pull/2244/commits/d5b7c32aa10e9b4fec485e15a50ce7755d2f26e2

doc/releases/changelog-dev.md
```diff
@@ -244,6 +244,9 @@
   gates, overwriting their effects.
   [(#2239)](https://github.com/PennyLaneAI/catalyst/pull/2239)
 
+* The `--inline-nested-module` pass no longer renames external function declarations.
```
dime10, line 266: Usually I jump over these, but I think typically you'd want more context in the release notes. Why is this relevant, where would a user or developer have encountered this, what is the implications, etc. Doesn't have to be long (succinct is good!), but it's more important to highlight the _context_ over the _thing_ that changed.

mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp
```diff
@@ -228,8 +233,22 @@ LogicalResult RenameFunctionsPattern::matchAndRewrite(Operation *child,
...
                     continue;
+                }
+
+                // We should not rename external function declarations, as they can be
+                // names required by other APIs.
+                // We record these external func decls during the rename pattern.
+                // Then during the actual inlining stage, only the first occurance of the per-module
+                // func decls of these external decls should be inlined.
+                if (isa<func::FuncOp>(op)) {
+                    auto f = cast<func::FuncOp>(op);
```
dime10:
```suggestion
                if (auto f = dyn_cast<func::FuncOp>(op)) {
```

mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp
```diff
@@ -183,14 +185,17 @@ struct AnnotateWithFullyQualifiedName : public OpInterfaceRewritePattern<SymbolO
...
+                           llvm::SmallSet<StringRef, 8> *externalFuncDeclNames)
+        : RewritePattern(MatchAnyOpTypeTag(), 1, context), _symbolTables(symbolTables),
+          _externalFuncDeclNames(externalFuncDeclNames)
     {
     }
 
     LogicalResult matchAndRewrite(Operation *op, PatternRewriter &rewriter) const override;
 
     SmallVector<Operation *> *_symbolTables;
+    llvm::SmallSet<StringRef, 8> *_externalFuncDeclNames;
```
dime10, line 198: This is just me being paranoid, but since we've had so many incidents with object lifetimes I want to confirm with you that the string storage lives for as long as this set does?

mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp
```diff
@@ -249,7 +268,14 @@ LogicalResult RenameFunctionsPattern::matchAndRewrite(Operation *child,
...
     /// This overload constructs a pattern that matches any operation type.
-    InlineNestedModule(MLIRContext *context) : RewritePattern(MatchAnyOpTypeTag(), 1, context) {}
+    InlineNestedModule(MLIRContext *context,
+                       const llvm::SmallSet<StringRef, 8> &externalFuncDeclNames)
+        : RewritePattern(MatchAnyOpTypeTag(), 1, context),
+          _externalFuncDeclNames(externalFuncDeclNames)
+    {
+    }
+
+    mutable llvm::SmallSet<StringRef, 8> alreadyInlinedFuncDeclNames;
```
dime10, line 286: Interesting, didn't even know this mechanism existed. Are you comfortable breaking the constness of the mlir rewrite function?

mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp
```diff
@@ -261,6 +287,27 @@ struct InlineNestedModule : public RewritePattern {
         }
 
         auto parent = op->getParentOp();
+        assert(parent->hasTrait<OpTrait::SymbolTable>() &&
+               "the direct parent of a qnode module must be a module op");
+
+        // Look for the func decls in the current qnode module
+        // If it is a recorded external func decl, erase it if it already has been inlined.
+        SmallVector<Operation *> _erasureWorklist;
+        op->walk([&](func::FuncOp f) {
```
dime10: Is op a module? I'm not sure we actually need to walk it recursively because I've never seen function definitions being nested in other functions 🤔 I simple iteration might suffice

mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp
```diff
@@ -272,6 +319,8 @@ struct InlineNestedModule : public RewritePattern {
 
         return success();
     }
+
+    llvm::SmallSet<StringRef, 8> _externalFuncDeclNames;
```
dime10: Can this be in the same location as the other datastructure? Throws me off that they are split 😅

dime10 (commented):
Nice work Paul!

mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp
paul0403: I tried the version below, but turns out `catalyst.callback_call` can actually be dyn casted to funcop (I don't know why), causing segfaults on some later methods that do not exist on the callback call op. That's why I opted for the explicit `isa` check.

But that might be specific only to the old hacky design? Let me try this now.

dime10: Oh what the heck, that should not be possible 👀

paul0403: My guess is because the callback op is [declared with `FunctionOpInterface`](https://github.com/PennyLaneAI/catalyst/blob/fd2e4beb5eed7e9d2dbce0df1c86dcac1782d92c/mlir/include/Catalyst/IR/CatalystOps.td#L118)? But that is also weird, if both B and C "inherit" from A, casting between B and C should still not be possible...

paul0403: Unless dyn_cast is much more loose than I originally thought. But well, let me try it with the new non hacky version now

dime10: Yeah, and also it's just an interface, and many ops inherit from it (and are supposed to) :O

paul0403: It might also just be I was dyn casting to the interface directly instead of funcop :sweat_smile: but I don't remember exactly. 

This is probably what happened just based on how much sense it makes haha

dime10: Oh yeah, if you were casting to the interface then that makes sense 😆

mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp
paul0403: I guess all I needed was a set to record some names. I can create this set outside the struct and pass it in as reference :+1:

mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2244/commits/f4d6216d19ae4ae5d2c6fec0994a2b51101e935f

mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2244/commits/541ff1de489d005c0b845e620c0b6012ce11cced

mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2244/commits/ee3b2f7ff16fc54d59028135b71f85e613ca129a

mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp
paul0403: This is just the struct's internal handle to the set (hence the pointer), I created this set outside the struct in the global "main" function of the pass where the pattern struct is called:
```C
        llvm::SmallSet<StringRef, 8> externalFuncDeclNames;
        renameFunctions.add<RenameFunctionsPattern>(context, &symbolTables, &externalFuncDeclNames);

        // ...stuff...
        inlineNested.add<InlineNestedModule>(context, externalFuncDeclNames,
                                             &alreadyInlinedFuncDeclNames);
```

I need the rename pattern to record which names are external func decls, so the subsequent inline pattern knows to handle them specially. Note that the rename pattern needs to write to the set so it's passed in as a pointer, but the inline pattern only needs to read it so it's passed in as const reference.

So the global `llvm::SmallSet<StringRef, 8> externalFuncDeclNames` is the one that controls the lifetime. And there is only ever one set! The local variables are just handles, aliasing the global one.

mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp
paul0403: `op` is the qnode module. Yeah come to think of it direct iteration makes more sense, [this is also what Erick chose in other places in this pass](https://github.com/PennyLaneAI/catalyst/blob/5e3813a243082e40daae48d6243e2761005a4189/mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp#L223) as well.

paul0403: https://github.com/PennyLaneAI/catalyst/pull/2244/commits/5ef407dca17e267985a27336def0ca218c4ba7ff

doc/releases/changelog-dev.md
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2244/commits/8eaf2a5f74f92b4ef39ea88b602fdc94330db6d6 added some context here

mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp
paul0403: Discussed offline, using mutable now

mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp
paul0403: Discussed offline, using `region.getOps<OpType>()` now

mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp
dime10: We realized the function symbols are stored as string attributes which are stored in the mlir context (forever essentially), so there is no lifetime concern here :)

mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp
```diff
@@ -228,8 +233,21 @@ LogicalResult RenameFunctionsPattern::matchAndRewrite(Operation *child,
...
             for (auto &op : block) {
-                if (!isa<SymbolOpInterface>(op))
+                if (!isa<SymbolOpInterface>(op)) {
                     continue;
+                }
+
+                // We should not rename external function declarations, as they can be
+                // names required by other APIs.
+                // We record these external func decls during the rename pattern.
+                // Then during the actual inlining stage, only the first occurance of the per-module
```
josephleekl:
```suggestion
                // Then during the actual inlining stage, only the first occurrence of the per-module
```
😆

paul0403: I trip up on this word so much

josephleekl (approved):
Thanks!

PR #2249, opened 2025-11-29
Update pennylane version

JerryChen97:
**Context:**
After two bumps of jax version in both Catalyst and Pennylane, here we update pennylane's pinned version back to the latest dev

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


paul0403:
Please make sure that all temporary workarounds are reverted in the following places:
- .dep-versions
- requirements.txt and doc/requirements.txt
- makefile
- CI scripts, both regular and wheels

PR #2250, opened 2025-12-01
Remove `setRotationKind` and `getRotationKind` from QEC interface

sengthai:
**Context:**
Previously, `setRotationKind` and `getRotationKind` were implemented for hooks in the QEC interface, but this should not be in the `QECOpInterface` because that interface also applies `PPMeasurementOp`, where `RotationKind` is not available.

[[sc-104780]]

### DIFF
--- a/mlir/lib/QEC/Transforms/TLayerReduction.cpp
+++ b/mlir/lib/QEC/Transforms/TLayerReduction.cpp
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "QEC/IR/QECDialect.h"
 #define DEBUG_TYPE "reduce-t-depth"
 
 #include <vector>
@@ -21,6 +20,7 @@
 #include "mlir/IR/PatternMatch.h"
 #include "mlir/Pass/Pass.h"
 
+#include "QEC/IR/QECOps.h"
 #include "QEC/Utils/PauliStringWrapper.h"
 #include "QEC/Utils/QECLayer.h"
 #include "QEC/Utils/QECOpUtils.h"


mlir/lib/QEC/Transforms/TLayerReduction.cpp
```diff
@@ -63,6 +64,8 @@ std::pair<bool, QECOpInterface> checkCommutationAndFindMerge(QECOpInterface rhsO
         auto normalizedOps =
             normalizePPROps(lhsOp, rhsOp, lhsOp.getInQubits(), rhsOpInQubitsFromLhsOp);
 
+        // TODO: Handle PPRotationArbitraryOp properly
```
sengthai, line 68: Arbitrary-angle PPR will be handled in the following PR.

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
```diff
@@ -137,16 +137,19 @@ void constructReverseCNOTLadder(SmallVector<Value> &qubits, ConversionPatternRew
 void constructKernelOperation(SmallVector<Value> &qubits, Value &measResult, QECOpInterface op,
                               ConversionPatternRewriter &rewriter)
 {
-    if (isa<PPMeasurementOp>(op)) {
+    if (isa<PPMeasurementOp>(op.getOperation())) {
```
rniczh: similar, why not just `isa<PPMeasurementOp>(op)`?

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
```diff
@@ -137,16 +137,19 @@ void constructReverseCNOTLadder(SmallVector<Value> &qubits, ConversionPatternRew
...
         auto measOp = buildMeasurementOp(qubits[0], rewriter);
         measResult = measOp.getMres();
         qubits[0] = measOp.getOutQubit();
     }
-    else {
-        int16_t signedRk = static_cast<int16_t>(op.getRotationKind());
+    else if (auto pprOp = dyn_cast<PPRotationOp>(op.getOperation())) {
```
rniczh, line 145: same

mlir/lib/QEC/Transforms/TLayerReduction.cpp
```diff
@@ -132,8 +135,11 @@ void moveOpToLayer(QECOpInterface rhsOp, QECLayer &rhsLayer, QECOpInterface merg
 // then just remove the `rhsOp` from the rhsLayer.
 void mergePPR(QECOpInterface rhsOp, QECLayer &rhsLayer, QECOpInterface mergeOp, IRRewriter &writer)
 {
-    int16_t signedRk = static_cast<int16_t>(mergeOp.getRotationKind());
-    mergeOp.setRotationKind(static_cast<uint16_t>(signedRk / 2));
+    auto mergeOpPprOp = dyn_cast<PPRotationOp>(mergeOp.getOperation());
```
rniczh, line 139: same

mlir/lib/QEC/Transforms/CountPPMSpecs.cpp
```diff
@@ -129,8 +129,8 @@ struct CountPPMSpecsPass : public impl::CountPPMSpecsPassBase<CountPPMSpecsPass>
         return true;
     }
 
-    bool isPPR(QECOpInterface op) { return isa<qec::PPRotationOp>(op); }
-    bool isPPM(QECOpInterface op) { return isa<qec::PPMeasurementOp>(op); }
+    bool isPPR(QECOpInterface op) { return isa<qec::PPRotationOp>(op.getOperation()); }
+    bool isPPM(QECOpInterface op) { return isa<qec::PPMeasurementOp>(op.getOperation()); }
```
rniczh: Doesn't MLIR already allow you to operate on an Interface without calling `.getOperation()`

rniczh (commented):
I forgot to post this comment, it should go along with https://github.com/PennyLaneAI/catalyst/pull/2250#pullrequestreview-3526873358

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
sengthai: Good point. For `isa`, we can pass `op`. However, I'm also wondering why `op` cannot pass to `dyn_cast`.

mlir/lib/QEC/Transforms/CountPPMSpecs.cpp
sengthai: You are right. I used op.getOperation() here because at that time multiple-level interface.

sengthai: I modified in here. [b01acda](https://github.com/PennyLaneAI/catalyst/pull/2250/commits/b01acda693b0f35063d9acabc524ba52cf9e79c2)

mlir/lib/QEC/Transforms/TLayerReduction.cpp
sengthai: dyn_cast needs `Operation *` type, and since that is QECOpInterface, we need to get the operation from it.

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
dime10: Curious if you found the answer to this? Normally a specific op class can automatically decay into the generic operation class.

Edit: Just read your other comment, I suppose it's because it's not an op but an interface class...

mlir/lib/QEC/Transforms/TLayerReduction.cpp
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "QEC/IR/QECDialect.h"
```
lazypanda10117: Need to update this to `QEC/IR/QECOps.h` after the refactoring update

mlir/lib/QEC/Transforms/PPRToMBQC.cpp
rniczh: @dime10 

`isa<...>` will go into the `class_of` to get the operation in `Operation.h`
```cpp
/// Interfaces.
  template <typename T>
  static std::enable_if_t<std::is_base_of<OpState, T>::value, bool>
  classof(const T *op) {
    return classof(const_cast<T *>(op)->getOperation());
  }
```
Through
```cpp
struct CastInfo<T, ::mlir::Operation>
    : public NullableValueCastFailed<T>,
      public DefaultDoCastIfPossible<T, ::mlir::Operation &,
                                     CastInfo<T, ::mlir::Operation>> {
  // ...
  static bool isPossible(::mlir::Operation &val) { return T::classof(&val); }
  static T doCast(::mlir::Operation &val) { return T(&val); }
};
```

`dyn_cast` requires to do the `cast_convert_val` after they ensured the op can be converted via the `class_of`  method. But `cast_convert_val` just the cast. I think this is the reason of why it cannot be converted to the specific op.

```cpp
template <class To, class FromTy> struct cast_convert_val<To, FromTy, FromTy> {
  // If it's a reference, switch to a pointer to do the cast and then deref it.
  static typename cast_retty<To, FromTy>::ret_type doit(const FromTy &Val) {
    return *(std::remove_reference_t<typename cast_retty<To, FromTy>::ret_type>
                 *)&const_cast<FromTy &>(Val);
  }
};
```

PR #2256, opened 2025-12-02
Add error if transform applied inside a qnode with program capture.

albi3ro:
**Context:**

We currently don't support applying transforms to a subcircuit. We need to explicitly error out on this case.

For:
```
@qml.qjit
@qml.qnode(qml.device('lightning.qubit', wires=1))
def c():
    qml.X(0)
    qml.X(0)
    return qml.probs()

c(), print(c.mlir)
```
we would have gotten:
```
TypeError: No ir_type_handler for aval type: <class 'pennylane.measurements.capture_measurements._get_abstract_measurement.<locals>.AbstractMeasurement'>
```

Now we get:
```
NotImplementedError: transforms cannot currently be applied inside a QNode.
```
Which is much more helpful and informative.

`QFuncToPlxprInterpreter` now raises an error if it encounters a transform.

**Benefits:**

Improved usability.

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-104320]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -71,6 +71,7 @@
 <h3>Improvements 🛠</h3>
 
 * An error is now raised if a transform is applied inside a QNode when program capture is enabled.
+  [(#2256)](https://github.com/PennyLaneAI/catalyst/pull/2256)
 
 * A new ``"changed"`` option has been added to the ``keep_intermediate`` parameter of 
   :func:`~.qjit`. This option saves intermediate IR files after each pass,


doc/releases/changelog-dev.md
```diff
@@ -70,6 +70,8 @@
 
 <h3>Improvements 🛠</h3>
 
+* An error is now raised if a transform is applied inside a QNode when program capture is enabled.
```
albi3ro, line 73:
```suggestion
* An error is now raised if a transform is applied inside a QNode when program capture is enabled.
  [(#2256)](https://github.com/PennyLaneAI/catalyst/pull/2256)
```

astralcai (commented):
Potentially this error message could be more informative as well.
```python
@qml.transforms.decompose
def circuit():
    qml.QFT([0, 1, 2, 3])
```
```pycon
>>> circuit()
RuntimeError: No queuing context available to append operation to.
```

PR #2258, opened 2025-12-02
merge_rotations arbitrary angle ppr support

kipawaa:
**Context:**
We've added arbitrary angle PPRs, but `merge_rotations` ignores them.

Adds support for arbitrary angle `qec.ppr.arbitrary` pprs to the `merge_rotations` pass.
The added rule merges PPRs as follows:

```
PPR(theta) * PPR(theta') -> PPR(theta + theta')
```

The rule supports conditions, re-ordering of Pauli words and qubits, and arbitrary sized PPRs.

**Benefits:**
`merge_rotations` simplifies QEC workflows as expected.

[sc-105961]

### DIFF
--- a/frontend/test/pytest/test_peephole_optimizations.py
+++ b/frontend/test/pytest/test_peephole_optimizations.py
@@ -535,6 +535,7 @@ def test_merge_rotation_arbitrary_angle_ppr():
     @qml.qjit(pipelines=my_pipeline, target="mlir")
     def test_merge_rotation_ppr_workflow():
         @qml.transforms.merge_rotations
+        @to_ppr_alias
         @qml.qnode(qml.device("lightning.qubit", wires=2))
         def circuit(x, y):
             qml.PauliRot(x, pauli_word="ZY", wires=[0, 1])
@@ -546,11 +547,11 @@ def test_merge_rotation_arbitrary_angle_ppr():
     ir_opt = test_merge_rotation_ppr_workflow.mlir_opt
 
     assert 'transform.apply_registered_pass "merge-rotations"' in ir
-    assert "qec.ppr.arbitrary" in ir
+    assert "qec.ppr.arbitrary" not in ir
     assert "arith.addf" not in ir
 
     assert "arith.addf" in ir_opt
-    assert 'qec.ppr.arbitrary ["Z", "Y"]' in ir_opt
+    assert ir_opt.count('qec.ppr.arbitrary ["Z", "Y"]') == 1
 
 
 def test_clifford_to_ppm():


mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
```diff
@@ -322,6 +322,24 @@ struct MergeRotationsRewritePattern : public OpRewritePattern<OpType> {
...
+    for (auto [qubit, pauli] : llvm::zip(parentOp.getOutQubits(), parentOp.getPauliProduct())) {
+        qubit_to_pauli[hash_value(qubit)] = pauli;
+    }
+
+    // check pairings
+    for (auto [qubit, pauli] : llvm::zip(op.getInQubits(), op.getPauliProduct())) {
+        if (qubit_to_pauli[hash_value(qubit)] != pauli) {
+            return false;
+        }
+    }
```
rniczh: I think you can use `Value` directly

```suggestion
    llvm::DenseMap<Value, Attribute> qubit_to_pauli;
    for (auto [qubit, pauli] : llvm::zip(parentOp.getOutQubits(), parentOp.getPauliProduct())) {
        qubit_to_pauli[qubit] = pauli;
    }

    // check pairings
    for (auto [qubit, pauli] : llvm::zip(op.getInQubits(), op.getPauliProduct())) {
        if (qubit_to_pauli[qubit] != pauli) {
            return false;
        }
    }
```

kipawaa: When I try to use `Value` like this I get an error: `the specified hash does not meet the Hash requirements`. Am I doing something wrong? [The docs](https://mlir.llvm.org/doxygen/classmlir_1_1Value.html#a34d1d75ed959007d8db5bd317d101d35) seem to suggest using `hash_value()`:

rniczh: Oh...Good point, you have to use `DenseMap`

rniczh: Edited

kipawaa: Cool! I added locally first 🙃 but it's in dbf09e68. Didn't know this was an option, thanks for showing me!

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
```diff
@@ -396,6 +414,66 @@ struct MergePPRRewritePattern : public OpRewritePattern<PPRotationOp> {
...
 };
 
+struct MergePPRArbitraryRewritePattern : public OpRewritePattern<PPRotationArbitraryOp> {
+    using OpRewritePattern::OpRewritePattern;
+
+    LogicalResult matchAndRewrite(PPRotationArbitraryOp op,
+                                  PatternRewriter &rewriter) const override
+    {
+        ValueRange inQubits = op.getInQubits();
+        auto definingOp = inQubits[0].getDefiningOp();
```
paul0403: We use `auto` only when the type is immediately obvious from RHS. If not, we should be explicit about the types.

```suggestion
        Operation *definingOp = inQubits[0].getDefiningOp();
```

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
```diff
@@ -322,6 +322,24 @@ struct MergeRotationsRewritePattern : public OpRewritePattern<OpType> {
...
+    }
+
+    // check pairings
+    for (auto [qubit, pauli] : llvm::zip(op.getInQubits(), op.getPauliProduct())) {
+        if (qubit_to_pauli[qubit] != pauli) {
+            return false;
+        }
+    }
+
+    return true;
```
paul0403: I don't think you even need to use a map, since you can get the index of the result
```C
bool matchingQubitsAndPaulis(PPR op, PPR parent){
   for (auto [inQubit, pauli]  : llvm::zip(op.getInQubits(), op.getPauliProduct()) ){
      if (parent.getPauliProduct()[cast<mlir::OpResult>(inQubit).getResultNumber()] != pauli){
         return false;
      }
   }
   return true;
}
```

https://mlir.llvm.org/doxygen/classmlir_1_1OpResult.html#ad8714eb4af7b6172d88a25bd6442b00b

But up to you. Either is fine.

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
```diff
@@ -396,6 +414,66 @@ struct MergePPRRewritePattern : public OpRewritePattern<PPRotationOp> {
...
+        if (op.getCondition() != prevOp.getCondition()) {
+            return failure();
+        }
+
+        auto opRotation = op.getArbitraryAngle();
+        auto prevOpRotation = prevOp.getArbitraryAngle();
+
+        if (!opRotation || !prevOpRotation) {
+            return failure();
+        }
```
paul0403: What is this block for? Is the `$arbitrary_angle` field on `PPRotationArbitraryOp` optional? I wouldn't think so because that would kinda defeat the point of the op? :sweat_smile:

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
```diff
@@ -396,6 +414,66 @@ struct MergePPRRewritePattern : public OpRewritePattern<PPRotationOp> {
...
+
+        // create merged op
+        auto loc = op.getLoc();
+        mlir::Value newAngleOp =
```
paul0403: Like here, `auto` would be fine! Since the type is obvious from `create<OpTy>` on the right hand side.

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
```diff
@@ -396,6 +414,66 @@ struct MergePPRRewritePattern : public OpRewritePattern<PPRotationOp> {
...
+            rewriter.create<arith::AddFOp>(loc, opRotation, prevOpRotation).getResult();
+
+        auto mergeOp = rewriter.create<PPRotationArbitraryOp>(
+            loc, op.getOutQubits().getTypes(), op.getPauliProduct(), newAngleOp,
+            prevOp.getInQubits(), op.getCondition());
```
paul0403: So your new op is getting the pauli product string from the second PPR and the operand qubit values from the first PPR... wouldn't that cause a mismatch in the case that your helper was intended for? i.e. for this case 
```mlir
%0, %1 = qec.ppr ["X", "Y"](2) %q1, %q2
%2, %3 = qec.ppr ["Y", "X"](2) %1, %0
``` 

Your new op created here would be 
```mlir
qec.ppr ["Y", "X"](<new_angle>) %q1, %q2
```
which is wrong?

Now I'm thinking your helper probably needs the map. Not just that it needs it, but it needs to return the map in order to correctly wire the qubit values here :thinking:

mlir/test/Quantum/MergeRotationsTest.mlir
```diff
@@ -941,3 +941,195 @@ func.func public @dont_merge_conditionals(%q1: !quantum.bit, %q2: !quantum.bit,
...
+// -----
+
+// re-arranging qubits is ok as long as the pauli words are re-arranged too
+
+// CHECK-LABEL: mix_and_match
+func.func public @mix_and_match(%q0: !quantum.bit, %q1: !quantum.bit, %0: f64, %1: f64) {
+    // CHECK: [[angle:%.+]] = arith.addf
+    // CHECK: qec.ppr.arbitrary [{{.+}}]([[angle]])
+    %2:2 = qec.ppr.arbitrary ["Z", "Y"](%0) %q0, %q1: !quantum.bit, !quantum.bit
+    %3:2 = qec.ppr.arbitrary ["Y", "Z"](%1) %2#1, %2#0: !quantum.bit, !quantum.bit
```
paul0403: Like here in the test, the result is probably wrong. It probably gave 
```mlir
   qec.ppr.arbitrary ["Y", "Z"](<addOp>) %q0, %q1
```

It only passed the test because you were checking against an arbitrary string pattern.

paul0403 (changes requested):

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
kipawaa: good point, removed in [28ce871](https://github.com/PennyLaneAI/catalyst/pull/2258/commits/28ce8714df8cfa03b56cf2ba9989860cca809192)

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
kipawaa: Good to know! I've updated my typing to be more explicit in [8e1f850](https://github.com/PennyLaneAI/catalyst/pull/2258/commits/8e1f850bc70ad4f864fb6e52af51a6a49e3d810e)

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
kipawaa: that makes sense, thanks for clarifying!

mlir/test/Quantum/MergeRotationsTest.mlir
kipawaa: hmmm, good catch. I've updated the test in 2fc4a226 for what I think would be good behaviour, but haven't found a good way to make that happen quite yet.

kipawaa: I've re-written and added more tests for this in [831620e](https://github.com/PennyLaneAI/catalyst/pull/2258/commits/831620e313638f2184d39bc9d9a3f58d99063f0a), that correspond to the new implementation add in [9820611](https://github.com/PennyLaneAI/catalyst/pull/2258/commits/98206113329e44cd8ae5c985a2b9ffc1ea6e0df5). Thanks again for catching this!

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
kipawaa: I've re-implemented with a single map that allows checking the Pauli strings and assigning the qubits to the new `mergeOp` in [9820611](https://github.com/PennyLaneAI/catalyst/pull/2258/commits/98206113329e44cd8ae5c985a2b9ffc1ea6e0df5).

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
kipawaa: Similar to comments below, went with an inverse permutation map to handle all of this. I think it's nicely readable, and can be used to accomplish everything needed rather simply. [9820611](https://github.com/PennyLaneAI/catalyst/pull/2258/commits/98206113329e44cd8ae5c985a2b9ffc1ea6e0df5)

kipawaa: I did use this technique to simplify the map creation in [0afae00](https://github.com/PennyLaneAI/catalyst/pull/2258/commits/0afae006e7f0ef1db0630509efd0dd0f9e00ef35) though, thanks for this tip!

mlir/test/Quantum/MergeRotationsTest.mlir
```diff
@@ -941,3 +941,221 @@ func.func public @dont_merge_conditionals(%q1: !quantum.bit, %q2: !quantum.bit,
...
+// re-arranging qubits is ok as long as the pauli words are re-arranged too
+
+// CHECK-LABEL: merge_permutations
+func.func public @merge_permutations(%z0: !quantum.bit, %y0: !quantum.bit, %0: f64, %1: f64, %2: f64, %3: f64) {
+    // CHECK-DAG: [[angle:%.+]] = arith.addf
+    // CHECK: qec.ppr.arbitrary ["Y", "Z"]([[angle]]) %arg1, %arg0
+    %z1, %y1 = qec.ppr.arbitrary ["Z", "Y"](%1) %z0, %y0: !quantum.bit, !quantum.bit
+    %y2, %z2 = qec.ppr.arbitrary ["Y", "Z"](%2) %y1, %z1: !quantum.bit, !quantum.bit
+    func.return
+}
```
paul0403:
```suggestion
func.func public @merge_permutations(%z0: !quantum.bit, %y0: !quantum.bit, %0: f64, %1: f64) {
    // CHECK-DAG: [[angle:%.+]] = arith.addf
    // CHECK: qec.ppr.arbitrary ["Y", "Z"]([[angle]]) %arg1, %arg0
    %z1, %y1 = qec.ppr.arbitrary ["Z", "Y"](%0) %z0, %y0: !quantum.bit, !quantum.bit
    %y2, %z2 = qec.ppr.arbitrary ["Y", "Z"](%1) %y1, %z1: !quantum.bit, !quantum.bit
    func.return
}
```

frontend/test/pytest/test_peephole_optimizations.py
```diff
@@ -525,6 +525,33 @@ def circuit():
...
 
+@pytest.mark.usefixtures("use_capture")
+def test_merge_rotation_arbitrary_angle_ppr():
+    """Test that the merge_rotation pass correctly merges arbtirary angle PPRs."""
+
+    my_pipeline = [("pipe", ["quantum-compilation-stage"])]
+
+    @qml.qjit(pipelines=my_pipeline, target="mlir")
+    def test_merge_rotation_ppr_workflow():
+        @qml.transforms.merge_rotations  # have to use qml to be capture-compatible
```
paul0403: No need for the comment here : )

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
```diff
@@ -396,6 +396,86 @@ struct MergePPRRewritePattern : public OpRewritePattern<PPRotationOp> {
...
+        // When two rotations have permuted Pauli strings, we can still merge them, we just need to
+        // correctly re-map the inputs. This map stores the index of a qubit in parentOp's out
+        // qubits at the index it appears in op's in qubits.
+        SmallVector<unsigned> inverse_permutation;
+        for (auto qubit : opInQubits) {
+            inverse_permutation.push_back(cast<mlir::OpResult>(qubit).getResultNumber());
+        }
+
+        // check Pauli + qubit pairings
+        mlir::ArrayAttr opPauliProduct = op.getPauliProduct();
```
paul0403: No need for all the `mlir::`, the namespace has already been opened.

frontend/test/pytest/test_peephole_optimizations.py
kipawaa: okay! Removed :) [ebec260](https://github.com/PennyLaneAI/catalyst/pull/2258/commits/ebec260eb192d669c9bea817a1fdcb29b5870241)

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
kipawaa: I removed it from most, but I get warnings about ambiguity with `Value` due to `llvm::Value` also being defined here so I kept those. [ebec260](https://github.com/PennyLaneAI/catalyst/pull/2258/commits/ebec260eb192d669c9bea817a1fdcb29b5870241)

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
kipawaa: switched to auto [ebec260](https://github.com/PennyLaneAI/catalyst/pull/2258/commits/ebec260eb192d669c9bea817a1fdcb29b5870241)

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
```diff
@@ -396,6 +396,86 @@ struct MergePPRRewritePattern : public OpRewritePattern<PPRotationOp> {
...
+            return failure();
+        }
+
+        auto parentOp = dyn_cast<PPRotationArbitraryOp>(definingOp);
+        if (!parentOp) {
+            return failure();
+        }
+
+        // verify that parentOp is parent of all qubits
+        for (mlir::Value qubit : opInQubits) {
```
paul0403, line 417:
```suggestion
        for (Value qubit : opInQubits) {
```

paul0403: ctrl+f tells me there's 5 such places left :wink:

lazypanda10117:
> Nice! @sengthai @lazypanda10117 either of you has time to give a second approval?

I had a quick look and it looks good. It shouldn't conflict with my PR changes. Will do a proper review later today.

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
kipawaa: Thanks for discussing with me! Cleaned up namespaces in [3f869d3](https://github.com/PennyLaneAI/catalyst/pull/2258/commits/3f869d382490e3d33692de9c6bad27c0c44c4295)

frontend/test/pytest/test_peephole_optimizations.py
```diff
@@ -525,6 +525,33 @@ def circuit():
...
+@pytest.mark.usefixtures("use_capture")
+def test_merge_rotation_arbitrary_angle_ppr():
+    """Test that the merge_rotation pass correctly merges arbtirary angle PPRs."""
+
+    my_pipeline = [("pipe", ["quantum-compilation-stage"])]
+
+    @qml.qjit(pipelines=my_pipeline, target="mlir")
+    def test_merge_rotation_ppr_workflow():
+        @qml.transforms.merge_rotations
+        @qml.qnode(qml.device("lightning.qubit", wires=2))
```
lazypanda10117, line 539: With the new changes in `https://github.com/PennyLaneAI/catalyst/pull/2284`, you would need to add the `@to_ppr` pass before `merge_rotations`.

mlir/test/Quantum/MergeRotationsTest.mlir
```diff
@@ -941,3 +941,221 @@ func.func public @dont_merge_conditionals(%q1: !quantum.bit, %q2: !quantum.bit,
...
+    %5:3 = qec.ppr.arbitrary ["X", "Z", "Y"](%2) %4#0, %4#1, %4#2: !quantum.bit, !quantum.bit, !quantum.bit
+    func.return
+}
+
+// -----
+
+// merge through other ops
+
+// CHECK-LABEL: merge_through
+func.func public @merge_through(%q0: !quantum.bit, %q1: !quantum.bit, %0: f64, %1: f64) -> !quantum.bit {
```
lazypanda10117, line 1034: Helps to mention that this merges through because the Hadamard operates on a different qubit.

lazypanda10117 (approved):
Looks good, thanks @kipawaa 🍾

frontend/test/pytest/test_peephole_optimizations.py
lazypanda10117: Approved, but please do not merge until #2284 is merged, as that may introduce new errors.

mlir/test/Quantum/MergeRotationsTest.mlir
sengthai, line 1: Can you add a test for an arbitrary PPR where the angle value is known?

kipawaa: Good idea! Added some tests for this in [d25b5fb](https://github.com/PennyLaneAI/catalyst/pull/2258/commits/d25b5fb92e395504da0efd293c6bd940ad40db15)

mlir/test/Quantum/MergeRotationsTest.mlir
kipawaa: Added some clarification in [45395b2](https://github.com/PennyLaneAI/catalyst/pull/2258/commits/45395b2686e903be09da3cad3c96b7cd05759e64)

frontend/test/pytest/test_peephole_optimizations.py
kipawaa: I'll wait to fix this until after your PR is merged so that I can properly verify everything works as expected

kipawaa: Updated [here](https://github.com/PennyLaneAI/catalyst/pull/2258/commits/4f33565cf237f97ac219c070cfbc97ed3df97040)! Thanks for your help :)

Issue #1762, opened 2025-05-22
Pipelines should have one source of truth


paul0403:
Currently we have two pipeline registries: 
https://github.com/PennyLaneAI/catalyst/blob/main/frontend/catalyst/pipelines.py: the frontend pipeline is used by default qjit program flow
https://github.com/PennyLaneAI/catalyst/blob/main/mlir/lib/Driver/Pipelines.cpp: the cpp pipeline is used by cli and things like having stage names in the pipeline being passed into the `pipeline` kwarg on qjit, e.g. in the oqd, mbqc, ppm pytests. 

This seems redundant, not to mention error-prone to develop and maintain. 

With more and more custom pipelines emerging and people start thinking about pipeline organization more (e.g. #1758 ), it would be good to have pipeline registration in just one place.

rturrado:
Hi all,

I have an initial [commit](https://github.com/PennyLaneAI/catalyst/compare/main...rturrado:catalyst:1762) that may serve as a proof of concept.

Explanation of the change:
- Set `pipelineList` in `mlir/include/DefaultPipelines.h` as the only source of truth for pipeline and pass names.
  This pipeline list is a vector of `PipelineInfo`, each of which is a pair of pipeline name and pass names.
  The pass names are just vectors of strings too.
- Update `mlir/lib/Driver/Pipelines.cpp` and `frontend/catalyst/pipelines.py` to use `DefaultPipelines.h`, via a few accessor functions.
  These were the 2 points where information was being duplicated at the moment.
  In order for the Python code to use the C++ code,  add a `default_pipelines` nanobind module to `frontend/catalyst`.

TODO:
- Add the rest of the pipelines and pass names to `pipelineList`, together with their corresponding accessor functions.
  This change will require accessor functions to receive, at least, an `async_qnodes` flag, to determine:
  1. if some passes are added or not to the pipeline (e.g., `qnode-to-async-lowering`), and
  2. the name for some of the passes (e.g., `one-shot-bufferize`).
- Update the code in `frontend/catalyst/pipelines.py` and `mlir/lib/Driver/Pipelines.cpp` to use those accessor functions.

PR #2259, opened 2025-12-02
Issue #1762: Pipelines should have one source of truth

rturrado:
**Context:**
Currently we have two pipeline registries: [the frontend pipeline](https://github.com/PennyLaneAI/catalyst/blob/main/frontend/catalyst/pipelines.py) is used by default qjit program flow, and [the CPP pipeline](https://github.com/PennyLaneAI/catalyst/blob/main/mlir/lib/Driver/Pipelines.cpp) is used by CLI and things like having stage names in the pipeline being passed into the `pipeline` kwarg on qjit, e.g. in the oqd, mbqc, ppm pytests.

This seems redundant, not to mention error-prone when developing and maintaining.

With more and more custom pipelines emerging and people start thinking about pipeline organization more (e.g. [#1758](https://github.com/PennyLaneAI/catalyst/pull/1758)), it would be good to have pipeline registration in just one place.

- Set `pipelineList` in `mlir/include/DefaultPipelines.h` as the only source of truth for pipeline and pass names.
  This pipeline list is a vector of `PipelineInfo`, each of which is a pair of pipeline name and pass names.
  The pass names are just vectors of strings too.
- Update `mlir/lib/Driver/Pipelines.cpp` and `frontend/catalyst/pipelines.py` to use `DefaultPipelines.h`, via a few accessor functions.
  These were the 2 points where information was being duplicated at the moment.
  In order for the Python code to use the C++ code,  add a `default_pipelines` nanobind module to `frontend/catalyst`.

**Benefits:**
The definition of the pipelines and passes, used by C++ and Python code, is defined just in one point.

**Possible Drawbacks:**
- We introduce `mlir/include/Driver/DefaultPipelines.h` as a dependency to the `frontend`. It's not a big deal but it is ugly to depend on files outside the project.
- I don't think this PR solves the two pipeline registration issue. It just proposes a solution for the duplication of data. But I fear that the issue complains about the duplication of the processes of registration too.
- On top of this last point, I think that the CPP pipeline code could be refactored a lot. I assume this is going to be accomplished in a near future work.

**Related GitHub Issues:**
#1762


rturrado:
@dime10 yes, yes, thanks!

rturrado:
Sorry. I closed the PR by error. I have been rebasing my 1762 branch to upstream/main. I still have to update it to add the PBC passes. Then accomplish the requested changes in the review.

rturrado:
> Sorry. I closed the PR by error. I have been rebasing my 1762 branch to upstream/main. I still have to update it to add the PBC passes. Then accomplish the requested changes in the review.

Current status for this PR:
- Rebased to upstream/main.
- Contains the new PBC passes.
- TODO: move the the nanobind module from the frontend to the compiler.

I have found some inconsistencies between `pipelines.py` and `Pipelines.cpp`, the current definitions of pipelines and passes in the `frontend` and `catalyst`, respectively:
- For `bufferization-pipeline`:
  - `eliminate-empty-tensors` and `promote-buffers-to-stack` passes are disabled in the `pipelines.py` and enabled in `Pipelines.cpp`. At the moment, I have left them enabled in `DefaultPipelines.h`.
- For `llvm-dialect-lowering-pipeline`:
  - `qnode-to-async-lowering`, `async-func-to-async-runtime`, `async-to-async-runtime`, and `convert-async-to-llvm` passes can be enabled in `pipelines.py` when an `async_qnodes` flag is true. Those passes are not present in `Pipelines.cpp`. `DefaultPipelines.h` allows adding them by passing a true `asyncQNodes` flag to `getLLVMDialectLoweringStage`.
  - `add-exception-handling` pass can be enabled in `pipelines.py` when an `async_qnodes` flag is true. That pass is always enabled in `Pipelines.cpp`. `DefaultPipelines.h` allows adding it by passing a true `asyncQNodes` flag to `getLLVMDialectLoweringStage`.
  - `convert-catalyst-to-llvm` pass goes before `convert-pbc-to-llvm` pass in `default_pipelines.py`, while the order is the opposite in `Pipelines.cpp`. `DefaultPipelines.h` uses the `pipelines.py` order. Otherwise, a lot of errors like the one below are raised when running the tests, making all MLIR to LLVM lowering tests fail:

```
grad.mul:26:3: error: failed to legalize operation 'gradient.custom_grad' that was explicitly marked illegal
  "gradient.custom_grad"() {callee = @callback_matrix_multiply_callback_125154215013664, forward = @matrix_multiply_fwd.fwd, reverse = @matrix_multiply_bwd.rev} : () -> ()
  ^
grad.mul:26:3: note: see current operation: "gradient.custom_grad"() <{callee = @callback_matrix_multiply_callback_125154215013664, forward = @matrix_multiply_fwd.fwd, reverse = @matrix_multiply_bwd.rev}> : () -> ()
While processing 'CatalystConversionPass' pass Failed to lower MLIR module
```

rturrado:
@dime10 I've now moved the `default_pipelines` nanobind module from the frontend to the compiler. The code is passing `make test` locally. However, I'm still not sure if this is what you had in mind.

- There is now a `DefaultPipelines` subfolder both within `mlir/include/Driver` and `mlir/lib/Driver`.

- `mlir/include/Driver/DefaultPipelines` holds:
  - `DefaultPipelines.h`, the new one-source-of-truth. This file is used both from the frontend (through the `default_pipelines` nanobind module) and from the compiler (through the Driver code). **Note** however that the `DefaultPipelines.h` doesn't have any dependencies with the rest of the Catalyst code. This was already intended in the original implementation to minimise coupling between the frontend and the compiler.

- `mlir/lib/Driver/DefaultPipelines` holds:
  - `DefaultPipelines.cpp` with the `NB_MODULE` code, and a
  - `CMakeLists.txt` with a `project(DefaultPipelines)` definition. This is needed when `cmake` is invoked from `setup.py` using `mlir/lib/Driver/DefaultPipelines` as root folder. **Note** that, if we were to use `mlir` as root folder from `setup.py`, the whole of the compiler would be compiled; while the nanobind module is really needed.

dime10:
Thanks for discovering (and fixing!) those inconsistencies in the default pipeline! In general I would say `pipelines.py` is probably the more up-to-date one, because that's what we use more often. Let me have a look in detail:

> - eliminate-empty-tensors and promote-buffers-to-stack passes are disabled in the pipelines.py and enabled in Pipelines.cpp. At the moment, I have left them enabled in DefaultPipelines.h.

I would say we should comment them out in `DefaultPipelines.h` as well, with the [same](https://github.com/PennyLaneAI/catalyst/blob/41b3e6e701f3f6b24962f514136f31b152b2355d/frontend/catalyst/pipelines.py#L319-L321) [notes](https://github.com/PennyLaneAI/catalyst/blob/41b3e6e701f3f6b24962f514136f31b152b2355d/frontend/catalyst/pipelines.py#L305-L308) from the frontend.

> - For llvm-dialect-lowering-pipeline:

Everything looks good to me here :)

rturrado:
> > * eliminate-empty-tensors and promote-buffers-to-stack passes are disabled in the pipelines.py and enabled in Pipelines.cpp. At the moment, I have left them enabled in DefaultPipelines.h.
> 
> I would say we should comment them out in `DefaultPipelines.h` as well, with the [same](https://github.com/PennyLaneAI/catalyst/blob/41b3e6e701f3f6b24962f514136f31b152b2355d/frontend/catalyst/pipelines.py#L319-L321) [notes](https://github.com/PennyLaneAI/catalyst/blob/41b3e6e701f3f6b24962f514136f31b152b2355d/frontend/catalyst/pipelines.py#L305-L308) from the frontend.

Perfect, thanks! I'll address this in the following commits.

I will also revert all the changes in the frontend because this PR shouldn't really affect the frontend, should it?

And I think there may be room for a couple of follow-up PRs, but I will leave that completely to you:
1. `setup.py` is now invoking `cmake` from `frontend` to build the `wrapper` extension, but I think it should suffice to invoke it from `frontend/catalyst/utils`. However, that would probably change the way `wrapper` is imported (e.g., from something like `from catalyst.utils import wrapper` to `from catalyst import wrapper`). And I don't know if you would be happy with that change.
2. I would also revisit the names of the passes. At the moment, you can find a mix of verbs and nouns, e.g., `lower-mitigation` and `adjoint-lowering`, which, in my opinion, calls for more consistency.

dime10:
Thanks @rturrado, I was actually in the process of testing out some changes, in particular I'm trying not to invoke the DefaultPipelines CMake script from setup.py but directly in the MLIR build. The nanobind module would then be discovered in the same way the compiler CLI and other artifacts are from the `mlir` and `runtime` systems. I can open a patch PR against your branch, or just open a follow up PR if you prefer :)

Overall your PR looks amazing btw 🎉

rturrado:
> Thanks @rturrado, I was actually in the process of testing out some changes, in particular I'm trying not to invoke the DefaultPipelines CMake script from setup.py but directly in the MLIR build. The nanobind module would then be discovered in the same way the compiler CLI and other artifacts are from the `mlir` and `runtime` systems.

Ah, yes, good. That is something I hadn't fully understood from your comments above. I should have asked for clarification. Anyway, I am very glad you are addressing it now.

> I can open a patch PR against your branch, or just open a follow up PR if you prefer :)

I leave that up to you. Whatever you decide is best.

> Overall your PR looks amazing btw 🎉

Many thanks! Appreciated.

.pylintrc
```diff
@@ -2,7 +2,7 @@
 # A comma-separated list of package or module names from where C extensions may
 # be loaded. Extensions are loading into the active Python interpreter and may
 # run arbitrary code
-extension-pkg-allow-list=catalyst.utils.wrapper
+extension-pkg-allow-list=catalyst.utils.wrapper,catalyst.default_pipelines
```
dime10, line 5: We might need to update this? I don't remember exactly how this setting works

dime10 (approved):
There was another pass added on main that wasn't added to the C++ definition (see merge commit), so it's a good thing this PR is finally going in :) 

I think we can ignore codefactor, this pr doesn't modify anything in the flagged files 🤔 

Changelog should be added though, this is probably an internal change because the `pipelines.py` still works the same for users.

EDIT: Ah, looks like we need to ensure nanobind is available for the mlir build step in CI, makes sense!

dime10:
Alright looks like we're good to go, the last failure from the mac wheels is coming from main 👌

.pylintrc
rturrado: I had added `catalyst.default_pipelines` after having added it to the list of Python extensions in `setup.py`. With the current code, it may be not necessary, but I have to check it.

rturrado:
> There was another pass added on main that wasn't added to the C++ definition (see merge commit), so it's a good thing this PR is finally going in :)

Great!

> I think we can ignore codefactor, this pr doesn't modify anything in the flagged files 🤔

OK.

> Changelog should be added though, this is probably an internal change because the `pipelines.py` still works the same for users.

I can add that if you haven't already.

.pylintrc
rturrado: I've just removed it. Because `default_pipelines` is discovered at runtime, it wouldn't be locatable at linting time.

mehrdad2m (approved):
Looks Great Thanks @rturrado!

rturrado:
> Looks Great Thanks @rturrado!

Thanks to you guys, a pleasure!

PR #2261, opened 2025-12-03
Remove the local transform dialect

mehrdad2m:
**Context:**
The local  transform dialect was created in https://github.com/PennyLaneAI/pennylane/pull/7956 because xdsl was not up to date with Catalyst's LLVM version which updated the transform dialect to accept dictionaries as options. Since xDSL has now updated their LLVM version, we can remove the local dialect.
 
Remove the local shim for the transform dialect and use the xdsl upstream one.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-103798]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -379,6 +379,10 @@
   [(#2245)](https://github.com/PennyLaneAI/catalyst/pull/2245)
   [(#2254)](https://github.com/PennyLaneAI/catalyst/pull/2254)
 
+  * Removed the `catalyst.python_interface.dialects.transform` module in favor of
+  using the `xdsl.dialects.transform` module directly.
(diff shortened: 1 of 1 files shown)


frontend/test/pytest/python_interface/dialects/test_transform_dialect.py
```diff
@@ -23,32 +23,34 @@
...
     compiler_transform,
 )
 
 
 def test_dict_options():
-    """Test ApplyRegisteredPassOp constructor with dict options."""
+    """Test ApplyRegisteredPassOp constructor with DictionaryAttr options."""
     target = create_ssa_value(AnyOpType())
-    options = {"option1": 1, "option2": True}
+    options = DictionaryAttr(
```
mehrdad2m, line 43: Note that this is because xDSL's __init__ method doesn't handle dictionaries.

paul0403 (approved):

PR #2262, opened 2025-12-03
Automatic qubit management works with gradients

paul0403:
**Context:**
In automatic qubit management mode, a `quantum.num_qubits` operation is inserted before terminal measurement ops to query their shape. This breaks gradient transforms because these expect the cloned classical parts of the circuit  to be free of quantum ops, but the num qubits ops are not removed there.

Add `MemoryRead` to `quantum.num_qubits` op: if it has no users, DCE (and canonicalizer) can remove it. 

**Benefits:**
Automatic qubit management works with gradients

[sc-105179]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -247,6 +247,12 @@
   gates, overwriting their effects.
   [(#2239)](https://github.com/PennyLaneAI/catalyst/pull/2239)
 
+* Fixed a bug where `quantum.num_qubits` operations are not properly removed during classical
+  processing of gradient transforms.
+  [(#2262)](https://github.com/PennyLaneAI/catalyst/pull/2262)
+
+  This fix enables automatic qubit management to be used with gradients.
+
 <h3>Internal changes ⚙️</h3>
 
 * Resource tracking now writes out at device destruction time instead of qubit deallocation


frontend/catalyst/jax_tracer.py
```diff
@@ -42,6 +42,7 @@
     ExpectationMP,
     MeasurementProcess,
     ProbabilityMP,
+    SampleMP,
```
paul0403, line 45: changelog

mlir/include/Quantum/IR/QuantumOps.td
```diff
@@ -125,7 +125,7 @@ def DeviceReleaseOp : Quantum_Op<"device_release"> {
     }];
 }
 
-def NumQubitsOp : Quantum_Op<"num_qubits"> {
+def NumQubitsOp : Quantum_Op<"num_qubits", [NoMemoryEffect]> {
```
dime10: I don't think this is correct. It definitely has a side-effect since it reads from a hidden state (a read). With this in place I think mlir would CSE two instances of this when they might not at all produce the same value (alloc in between, also a (write) side-effect).

frontend/catalyst/jax_tracer.py
```diff
@@ -1077,23 +1078,26 @@ def trace_quantum_measurements(
...
 
             if device.wires is None:
-                d_wires = num_qubits_p.bind()
+                # Automatic qubit management, need to query num qubits for shape-ful measurements
+                # Note: 2 ifs instead of `and` because the next case uses `is_dynamic_wires`, which
+                # asserts that it is not seeing `None`
+                if type(output) in [SampleMP, CountsMP, ProbabilityMP, StateMP]:
+                    d_wires = num_qubits_p.bind()
             elif catalyst.device.qjit_device.is_dynamic_wires(device.wires):
                 d_wires = num_qubits_p.bind()
```
dime10: I'm slightly hesitant. You're right currently that the variable `d_wires` or rather `nqubits` is only used for those MPs (which is actually most of them, only expval/var is excluded). But I wonder if it isn't more trouble than it's worth if the code below changes (another implicit assumption, might not be obvious why `nqubits` is None in that case). You also didn't apply this to the branch right below which should have the same constraints.

I'd say for simplicity there is no harm in always generating if we can just DCE it.

mlir/include/Quantum/IR/QuantumOps.td
paul0403: You are right! I completely forgot about the read!

Added mem read effects https://github.com/PennyLaneAI/catalyst/pull/2262/commits/c6e8cc6b40900e77d4bf6e780e481f0be1661711

Note that the test has an alloc in between the two numqubits op since mlir still cse two ops with memread effect if they don't have another side-effect-ful op in between (which makes sense)

frontend/catalyst/jax_tracer.py
paul0403: reverted https://github.com/PennyLaneAI/catalyst/pull/2262/commits/6f30fdc9acc8d9b6473956deec66e4651fb182d7

mlir/include/Quantum/IR/QuantumOps.td
dime10: Nice!

dime10 (approved):

frontend/test/pytest/test_automatic_qubit_management.py
```diff
@@ -151,3 +152,19 @@ def circuit():
...
+    dev = qml.device(backend)
+
+    @qjit
+    @qml.qnode(dev)
+    def circuit(params):
+        qml.RX(params[0], wires=1)
+        qml.RY(params[1], wires=2)
+        return qml.expval(qml.PauliZ(0))
+
+    assert jnp.allclose(jacobian(circuit)(jnp.array([0.37, 0.42])), jnp.zeros(2))
```
mlxd, line 170: I presume we cannot test this without a device execution?

mlxd (approved):
Single query, non-blocking.
If we need the dev exec, then we can merge. If we can avoid it, let's lock in another follow-up backlog item to do it more fine-grained.

frontend/test/pytest/test_automatic_qubit_management.py
paul0403: This is just the end-to-end test, the unit test is already added as the mlir lit test.

PR #2269, opened 2025-12-05
Add pass that applies Clifford+T -> Pauli frame protocols

joeycarter:
**Context:** In #2188, we added a new `PauliFrame` dialect to Catalyst to represent programs using Pauli frame tracking protocols. This PR adds a new compilation pass, `to-pauli-frame`, which applies these Pauli frame tracking protocols to a Clifford+T program.

**Description of the Change:** Applying the `to-pauli-frame` pass applies the following transformations:

1. Initialize the Pauli record of the qubits in a register by inserting a `pauli_frame.init_qreg` op immediately after `quantum.alloc` ops, taking the `!quantum.reg` as its input operand. A similar pattern that inserts a `pauli_frame.init` op immediately after a single-qubit allocation op `quantum.alloc_qb` is also applied.
2. Converts Pauli gates to their equivalent `pauli_frame.update` ops, e.g. it replaces a Pauli X gate with a `pauli_frame.update [1, 0]` op, where the attributes in the square brackets are the [x, z] parity bits.
3. Inserts `pauli_frame.update_with_clifford` ops immediately before Clifford ops (H, S, and CNOT).
4. Flushes the Pauli record immediately before non-Clifford ops (T). _Flushing_ applies the `pauli_frame.flush` op, which returns the x- and z-parity bits for the given qubit, and applies the X and Z gates to that qubit, conditional on the returned parity bits.
5. Corrects single-qubit mid-circuit measurements by inserting `pauli_frame.correct_measurement` ops immediately after `quantum.measure` ops.
6. Flushes the Pauli records of all relevant qubits immediately before applying terminal measurement processes.

**Benefits:** This is a proof-of-concept prototype that allows us to express quantum programs in Catalyst IR using Pauli frame tracking protocols.

**Possible Drawbacks:** This prototype is by no means production grade and the final version of this compilation pass used for utility-scale workloads will need updating for both functional purposes (e.g. to communicate with a remote process running the Pauli frame tracking process) and scalability purposes (e.g. flushing the Pauli frame for each qubit before performing terminal measurement processes is not optimal, as it requires many calls into the Pauli frame process to retrieve the Pauli records, and as it currently inlines all of these operations into the main region that performs the quantum execution). The scalability issues might be avoided by carefully defining a new operation that corrects the measurement results in an array of samples in one go.

Another detail that we have not considered in this prototype is handling the Pauli byproduct corrections in a program's MBQC representation. This is a more challenging task because of the state-teleportation that occurs in MBQC, and it will need to be addressed in a future PR.  

-----

[sc-103533]

### DIFF
--- a/mlir/lib/PauliFrame/Transforms/CliffordTToPauliFramePatterns.cpp
+++ b/mlir/lib/PauliFrame/Transforms/CliffordTToPauliFramePatterns.cpp
@@ -174,7 +174,7 @@ LogicalResult convertCliffordGate(CustomOp op, PatternRewriter &rewriter, Cliffo
     UpdateWithCliffordOp updateOp =
         rewriter.create<UpdateWithCliffordOp>(loc, outQubitTypes, gate, inQubits);
 
-    op->setOperands(updateOp->getResults());
+    rewriter.modifyOpInPlace(op, [&] { op->setOperands(updateOp->getResults()); });
     return success();
 }
 
@@ -224,7 +224,7 @@ LogicalResult convertNonCliffordGate(CustomOp op, PatternRewriter &rewriter)
 
     auto pauliZOutQubit = insertPauliOpsAfterFlush(rewriter, loc, flushOp);
 
-    op->setOperands(pauliZOutQubit);
+    rewriter.modifyOpInPlace(op, [&] { op->setOperands(pauliZOutQubit); });
 
     return success();
 }
@@ -455,12 +455,12 @@ struct FlushBeforeMeasurementProcessPattern : public OpRewritePattern<Measuremen
                 auto flushOp = rewriter.create<FlushOp>(
                     loc, rewriter.getI1Type(), rewriter.getI1Type(), qubit.getType(), qubit);
                 auto pauliZOutQubit = insertPauliOpsAfterFlush(rewriter, loc, flushOp);
-                obsOp->setOperand(idx, pauliZOutQubit);
+                rewriter.modifyOpInPlace(obsOp, [&] { obsOp->setOperand(idx, pauliZOutQubit); });
             }
             else {
                 // Get output qubit of Pauli Z op after the flush
                 auto pauliZOutQubit = getOutputQubitOfPauliZAfterFlush(flushOp.value());
-                obsOp->setOperand(idx, pauliZOutQubit);
+                rewriter.modifyOpInPlace(obsOp, [&] { obsOp->setOperand(idx, pauliZOutQubit); });
             }
         };
 


joeycarter:
Can't reproduce the failing lit test locally—will need to investigate.

mlir/lib/PauliFrame/Transforms/CliffordTToPauliFramePatterns.cpp
```diff
@@ -0,0 +1,447 @@
...
+    else if (gateName == "H" || gateName == "Hadamard")
+        return GateEnum::H;
+    else if (gateName == "S")
+        return GateEnum::S;
+    else if (gateName == "T")
+        return GateEnum::T;
+    else if (gateName == "CNOT")
+        return GateEnum::CNOT;
+    else
+        return GateEnum::Unknown;
```
rniczh: I think you can try using `llvm::StringSwitch`, https://llvm.org/doxygen/classllvm_1_1StringSwitch.html
```suggestion
    return llvm::StringSwitch<GateEnum>(op.getGateName())
        .Cases("Identity", "I", GateEnum::I)
        .Cases("PauliX", "X", GateEnum::X)
        .Cases("PauliY", "Y", GateEnum::Y)
        .Cases("PauliZ", "Z", GateEnum::Z)
        .Cases("H", "Hadamard", GateEnum::H)
        .Case("S", GateEnum::S)
        .Case("T", GateEnum::T)
        .Case("CNOT", GateEnum::CNOT)
        .Default(GateEnum::Unknown);
```

mlir/lib/PauliFrame/Transforms/CliffordTToPauliFramePatterns.cpp
```diff
@@ -0,0 +1,447 @@
...
+        case GateEnum::T:
+            return convertNonCliffordGate(op, rewriter);
+        case GateEnum::Unknown: {
+            op->emitError() << "Unsupported gate: '" << op.getGateName()
+                            << "'. Only Clifford+T gates are supported for Pauli frame conversion: "
+                            << "I, X, Y, Z, H, S, S†, T, T†, and CNOT";
+            return failure();
+        }
+        }
+        return success();
```
rniczh: If it's covered, this line is deadcode right? Maybe we can just call the unreachable if this pattern intends to cover the gate enum?

```suggestion
        llvm_unreachable("unhandled gate");
```

mlir/lib/PauliFrame/Transforms/CliffordTToPauliFramePatterns.cpp
```diff
@@ -0,0 +1,447 @@
...
+#include <llvm/Support/Casting.h>
+#include <llvm/Support/Debug.h>
+#include <llvm/Support/LogicalResult.h>
+#include <llvm/Support/raw_ostream.h>
+#include <mlir/Dialect/SCF/IR/SCF.h>
+#include <mlir/IR/Builders.h>
+#include <mlir/IR/Location.h>
+#include <mlir/IR/PatternMatch.h>
+#include <mlir/IR/Value.h>
+#include <mlir/IR/ValueRange.h>
```
rniczh: Pretty minor thing, the coding convention of `include` for llvm and mlir should be "..." instead of <...>

```suggestion
#include "llvm/Support/Casting.h"
#include "llvm/Support/Debug.h"
#include "llvm/Support/LogicalResult.h"
#include "llvm/Support/raw_ostream.h"

#include "mlir/Dialect/SCF/IR/SCF.h"
#include "mlir/IR/Builders.h"
#include "mlir/IR/Location.h"
#include "mlir/IR/PatternMatch.h"
#include "mlir/IR/Value.h"
#include "mlir/IR/ValueRange.h"
```

mlir/lib/PauliFrame/Transforms/CliffordTToPauliFramePatterns.cpp
```diff
@@ -0,0 +1,447 @@
...
+        LLVM_DEBUG(llvm::dbgs() << "Applying Pauli frame protocol to flush Pauli record before "
+                                   "terminal measurement process: "
+                                << op << "\n");
+        auto loc = op->getLoc();
+
+        auto obs = op.getObs();
+        if (!obs) {
+            op.emitError() << "Failed to flush Pauli record before terminal measurement process";
+            return failure();
+        }
```
rniczh: So the obs might be pointing to nullptr?

mlir/lib/PauliFrame/Transforms/CliffordTToPauliFramePatterns.cpp
```diff
@@ -0,0 +1,447 @@
...
+            return convertPauliGate(op, rewriter, false, false);
+        case GateEnum::X:
+            return convertPauliGate(op, rewriter, true, false);
+        case GateEnum::Y:
+            return convertPauliGate(op, rewriter, true, true);
+        case GateEnum::Z:
+            return convertPauliGate(op, rewriter, false, true);
+        case GateEnum::H:
+            return convertCliffordGate(op, rewriter, CliffordGate::Hadamard);
+        case GateEnum::S:
```
rniczh, line 256: What happen if the given gate is adjoint? does the conversion still sound?

mlir/lib/PauliFrame/Transforms/CliffordTToPauliFramePatterns.cpp
```diff
@@ -0,0 +1,447 @@
...
+
+        // Helper function to insert the flush operations per qubit operand of the observable op
+        auto insertFlushOpsPerQubit = [&](unsigned int idx, const Value qubit) {
+            auto flushOp = rewriter.create<FlushOp>(loc, rewriter.getI1Type(), rewriter.getI1Type(),
+                                                    qubit.getType(), qubit);
+            auto pauliZOutQubit = insertPauliOpsAfterFlush(rewriter, loc, flushOp);
+            obsOp->setOperand(idx, pauliZOutQubit);
+        };
+
+        if (auto compBasisOp = dyn_cast<ComputationalBasisOp>(obsOp)) {
```
rniczh, line 467: I'm not sure if this is the case that two measurement processes rely on one observable, would it flush the qubit twice?

mlir/lib/PauliFrame/Transforms/CliffordTToPauliFramePatterns.cpp
joeycarter: Thanks 🙂

mlir/lib/PauliFrame/Transforms/CliffordTToPauliFramePatterns.cpp
joeycarter: Nice, wasn't aware that LLVM had a `StringSwitch`. Thanks!

mlir/lib/PauliFrame/Transforms/CliffordTToPauliFramePatterns.cpp
joeycarter: Yes, good call. 👍

mlir/lib/PauliFrame/Transforms/CliffordTToPauliFramePatterns.cpp
joeycarter: Yes, this is correct as is; it's explained in the documentation for the `convertCliffordGate` function:

> Note that since H = H†, CNOT = CNOT†, and since the Pauli conjugation relations for S and S† are equivalent up to a global phase, we need not consider the adjoint parameter of the quantum gate.

Let me know if you'd like me to add a comment here to reiterate this point.

mlir/lib/PauliFrame/Transforms/CliffordTToPauliFramePatterns.cpp
joeycarter: Good catch. I believe in our IR we don't normally have two measurement processes relying on the same observable, but we _do_ support having multiple observables on the same qubit, e.g.

```python
@catalyst.qjit(keep_intermediate=True)
@qml.qnode(dev, shots=100)
def circuit():
    qml.Hadamard(0)
    return qml.probs(wires=[0]), qml.sample(wires=[0])
```

gives

```mlir
func.func public @circuit() -> (tensor<2xf64>, tensor<100x1xi64>) attributes {...} {
  %c100_i64 = arith.constant 100 : i64
  quantum.device shots(%c100_i64) ["", "", ""]
  %out_qubits = quantum.custom "Hadamard"() %1 : !quantum.bit
  %2 = quantum.compbasis qubits %out_qubits : !quantum.obs
  %3 = quantum.probs %2 : tensor<2xf64>
  %4 = quantum.compbasis qubits %out_qubits : !quantum.obs
  %5 = quantum.sample %4 : tensor<100x1xf64>
  %6 = stablehlo.convert %5 : (tensor<100x1xf64>) -> tensor<100x1xi64>
  %7 = quantum.insert %0[ 0], %out_qubits : !quantum.reg, !quantum.bit
  quantum.dealloc %7 : !quantum.reg
  quantum.device_release
  return %3, %6 : tensor<2xf64>, tensor<100x1xi64>
}
```

in which case, you're correct, this would result in two separate `flush` ops on `%out_qubits`.

This isn't _wrong_, per se; the second flush would always return x-/z-parity bits (0, 0), since flushing also reinitialized the Pauli record to (0, 0), so the Pauli X and Z gates wouldn't be applied again. However, this is obviously inefficient to flush twice and should be fixed. Let me come back to this, I think instead we should check if a qubit SSA value is already used in a `flush` op, and if so, skip it.

mlir/lib/PauliFrame/Transforms/CliffordTToPauliFramePatterns.cpp
rniczh: Thanks for explanation

mlir/lib/PauliFrame/Transforms/CliffordTToPauliFramePatterns.cpp
joeycarter: Realistically, no, `getObs()` should always return a value given the measurement processes we're supporting. This check should really be an `assert` rather than an op error. I'll update.

joeycarter: 75f819b17

mlir/lib/PauliFrame/Transforms/CliffordTToPauliFramePatterns.cpp
joeycarter: WIP: b1e968464

Still need to add a few more test cases and polish up.

joeycarter: Done: e0849d263

doc/releases/changelog-dev.md
```diff
@@ -488,6 +488,10 @@
   operations for interacting with an external Pauli frame tracking library.
   [(#2188)](https://github.com/PennyLaneAI/catalyst/pull/2188)
 
+* A new `to-pauli-frame` compilation pass has been added, which applies the Pauli frame protocols to
```
```suggestion
* A new `to-pauli-frame` compilation pass has been added, which applies the Pauli frame tracking protocols to
```

mlir/lib/PauliFrame/Transforms/CliffordTToPauliFramePatterns.cpp
```diff
@@ -0,0 +1,502 @@
...
+    LLVM_DEBUG(llvm::dbgs() << "Applying Pauli frame protocol to Clifford gate: "
+                            << op.getGateName() << "\n");
+    auto loc = op->getLoc();
+    auto outQubitTypes = op.getOutQubits().getTypes();
+    auto inQubits = op.getInQubits();
+
+    UpdateWithCliffordOp updateOp =
+        rewriter.create<UpdateWithCliffordOp>(loc, outQubitTypes, gate, inQubits);
+
+    op->setOperands(updateOp->getResults());
```
dime10: do we need to notify the rewriter of the mutation here?

mlir/lib/PauliFrame/Transforms/CliffordTToPauliFramePatterns.cpp
```diff
@@ -0,0 +1,502 @@
...
+ */
+LogicalResult convertNonCliffordGate(CustomOp op, PatternRewriter &rewriter)
+{
+    LLVM_DEBUG(llvm::dbgs() << "Applying Pauli frame protocol to non-Clifford gate: "
+                            << op.getGateName() << "\n");
+    auto loc = op->getLoc();
+    auto outQubitTypes = op.getOutQubits().getTypes();
+
+    if (outQubitTypes.size() > 1) {
+        op->emitError() << "Only single-qubit non-Clifford gates are supported";
```
dime10: Why is this? Can we not just flush all the qubits involved in the gate?

mlir/lib/PauliFrame/Transforms/CliffordTToPauliFramePatterns.cpp
```diff
@@ -0,0 +1,502 @@
...
+            return convertCliffordGate(op, rewriter, CliffordGate::S);
+        case GateEnum::CNOT:
+            return convertCliffordGate(op, rewriter, CliffordGate::CNOT);
+        case GateEnum::T:
+            return convertNonCliffordGate(op, rewriter);
+        case GateEnum::Unknown: {
+            op->emitError() << "Unsupported gate: '" << op.getGateName()
+                            << "'. Only Clifford+T gates are supported for Pauli frame conversion: "
+                            << "I, X, Y, Z, H, S, S†, T, T†, and CNOT";
+            return failure();
```
dime10: I'm not sure if that's your intention, but this will not stop the compilation process. The error will be printed, and a match failure returned (which skips the pattern), but otherwise everything will go on as normal.

dime10 (approved):
nice 👍

mlir/lib/PauliFrame/Transforms/CliffordTToPauliFramePatterns.cpp
joeycarter: This is more just a sanity check, since the only non-Clifford gate that's supported in this pass is the `T` gate. I've changed to an assert in 5ac3e7aa8.

mlir/lib/PauliFrame/Transforms/CliffordTToPauliFramePatterns.cpp
joeycarter: Interesting, I'm not sure why it was working without notifying here :thinking:. Anyway I updated this line and the other `setOperands()` to use `rewriter.modifyOpInPlace()`: 9007baed1

mlir/lib/PauliFrame/Transforms/CliffordTToPauliFramePatterns.cpp
joeycarter: Good catch, thanks. Instead of returning `failure()` here I've let it fall back to `llvm_unreachable` to abort: b94fd23c8

There's probably a better way to do this, e.g. if it comes across an unsupported gate then in principle it could flush the Pauli record for the qubits acted upon by that gate, keep the unsupported gate in the IR and carry on. For the sake of the prototype, where we can assume the initial program representation is in Clifford+T, I think it's fastest, easiest and safest to just abort.

mlir/lib/PauliFrame/Transforms/CliffordTToPauliFramePatterns.cpp
dime10: So the functions do work on their own (as in they execute the mutation), but it's more about triggering any registered callbacks, updating internal bookkeeping (e.g. for worklist algorithms, although I don't think the walk applicator uses one), that sort of thing. Relevant docs:
https://mlir.llvm.org/docs/PatternRewriter/#restrictions
https://mlir.llvm.org/docs/PatternRewriter/#pattern-rewriter
>As noted above, _all_ IR mutations, including creations, are required to be performed via the PatternRewriter class. This is required because the underlying pattern driver may have state that would be invalidated when a mutation takes place.

mlir/lib/PauliFrame/Transforms/CliffordTToPauliFramePatterns.cpp
dime10: Yes I think in general it's best to design passes such that they don't error if they can't do their thing, unless it's absolutely required. Regarding the failure mechanism, unfortunately we are quite constrained (probably intentionally, but I'm not sure of the reasons), within MLIR rewrite patterns:
- for one, LLVM disables exceptions across the board, so we cannot raise them where we wish
- then we have asserts, but those could be disabled via build flags, and already have a different purpose anyway
- MLIR does provide a `signalPassFailure` function, which aborts the current pipeline, and in our setup will abort the compiler, but it can only be used from within a pass, not a pattern

I'm not sure we've ever figured out the best (or intended) way to handle this 😅

PR #2134, opened 2025-10-21
Bump jax to 0.7.1

rniczh:
**Context:**

**Description of the Change:**

1. API change of `lower_jaxpr_to_fun`, update it in catalyst accordingly
2. API change of `jaxpr_subcomp`, update it in catalyst accordingly
3. `_jaxpr_replicas` has been moved to  `jax._src.interpreters.pxla` in jax
4. After jax 0.7.1+ `mlir.multi_broadcast_in_dim` has changed, and break the dynamic shape, patch it to fix the dynamic shape issue

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

PR #2271, opened 2025-12-05
Add forward capability with python 3.14

rniczh:
**Context:**

About the JAX's version, now catalyst upgraded to JAX 0.7.1 https://github.com/PennyLaneAI/catalyst/pull/2134 and after JAX 0.7.1, JAX started to ship python 3.14 wheel https://docs.jax.dev/en/latest/changelog.html#jax-0-7-1-august-20-2025. This PR adds the forward capability with python 3.14.

Add forward capability with python 3.14.

Some packages still need conditionally install the python version below 3.14.
```
pennylane-lightning-kokkos; python_version < "3.14"
amazon-braket-pennylane-plugin>1.27.1; python_version < "3.14"
```

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -292,6 +292,10 @@
 
 <h3>Bug fixes 🐛</h3>
 
+* Update `catalyst.python_interface.inspection.xdsl_conversion.get_mlir_module` to
+  no longer explicitly use the xDSL pass plugin.
+  [(#2303)](https://github.com/PennyLaneAI/catalyst/pull/2303)
+
 * Added missing `detensorize-function-boundary` and `symbol-dce` passes to `Pipelines.cpp`.
   [(#2266)](https://github.com/PennyLaneAI/catalyst/pull/2266)
 
@@ -521,6 +525,7 @@ Joseph Lee,
 Mehrdad Malekmohammadi,
 River McCubbin,
 Lee J. O'Riordan,
+Andrija Paurevic,
 Roberto Turrado,
 Paul Haochen Wang,
 Jake Zaia,
--- a/frontend/catalyst/python_interface/inspection/xdsl_conversion.py
+++ b/frontend/catalyst/python_interface/inspection/xdsl_conversion.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 
 import inspect
 from collections.abc import Callable
+from copy import deepcopy
 from typing import TYPE_CHECKING
 
 from pennylane import ops
@@ -32,7 +33,6 @@ from xdsl.dialects.tensor import ExtractOp as TensorExtractOp
 from xdsl.ir import SSAValue
 
 from catalyst.jit import QJIT, qjit
-from catalyst.passes.xdsl_plugin import getXDSLPluginAbsolutePath
 
 from ..dialects.quantum import (
     CustomOp,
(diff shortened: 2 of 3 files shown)


.github/workflows/build-wheel-macos-arm64.yaml
```diff
@@ -166,9 +166,9 @@ jobs:
         lookup-only: True
 
     - name: Setup Python version
-      # There are multiple Python versions installed on the GitHub image, 3.11 - 3.12 is already
+      # There are multiple Python versions installed on the GitHub image, 3.11 - 3.14 is already
       # available under /Library/Frameworks/Python.framework/Versions/, but homebrew also provides
-      # 3.11 and 3.12. Make sure to consistently use the system versions.
+      # 3.11 and 3.14. Make sure to consistently use the system versions.
```
dime10: Did you verify this was the case on the image or did you routinely update these versions?

.github/workflows/constants.yaml
```diff
@@ -80,12 +80,12 @@ jobs:
       - name: Python versions
         id: python_versions
         run: |
-          echo 'python_versions=["3.11", "3.12"]' >> $GITHUB_OUTPUT
+          echo 'python_versions=["3.11", "3.12", "3.13", "3.14"]' >> $GITHUB_OUTPUT
```
dime10: How come 3.13 is being added now?

dime10: These might be the build versions, in which case the goal is to never go beyond 12 (well at some point we might, when 3.12 is end-of-life, but for now we won't).

.github/workflows/build-wheel-macos-arm64.yaml
rniczh: I just added to test if they can be installed from GitHub

EDITED:
And yes, they are in the image

.github/workflows/constants.yaml
rniczh: Yes, build version should only be 3.11 and 3.12, just fixed it https://github.com/PennyLaneAI/catalyst/pull/2271/commits/bf72a489da2ede0527a0b885aa87263d29030314

requirements.txt
```diff
@@ -32,5 +32,7 @@ pytest-mock
 nbmake
 
 # optional rt/test dependencies
-pennylane-lightning-kokkos
-amazon-braket-pennylane-plugin>1.27.1
+pennylane-lightning-kokkos; python_version < "3.14"
```
JerryChen97:
```suggestion
pennylane-lightning-kokkos
```

.github/workflows/build-wheel-linux-arm64.yaml
```diff
@@ -474,7 +474,7 @@ jobs:
         python${{ matrix.python_version }} -m pip install 'amazon-braket-pennylane-plugin>=1.31.0'
 
     - name: Install OQC client
-      if: matrix.python_version != '3.13'
+      if: ${{ !contains(fromJSON('["3.13", "3.14"]'), matrix.python_version) }}
```
dime10: I wonder what the issue is, OQC doesn't seem to ship version-specific wheels: https://pypi.org/project/oqc-qcaas-client/#files

Oh wait, is this related to our build of the OpenQASM-based plugins?

dime10: On a separate note, is there no way to do a less-than comparison (to avoid updating these versions every time? 😅)

dime10:
> changelog?

Yes I think we should advertise that 3.14 is officially supported now :)

.github/workflows/build-wheel-linux-arm64.yaml
rniczh: Thanks! I updated the checking logic https://github.com/PennyLaneAI/catalyst/pull/2271/commits/011ae84a10e67c48e916910594d6938a9e4352e2

JerryChen97 (approved):
LGTM!

PR #2272, opened 2025-12-05
Add integration tests for pass-by-pass specs

jzaia18:
**Context:**
https://github.com/PennyLaneAI/pennylane/pull/8606 adds pass-by-pass specs to PennyLane.

Adds integration tests of the new pass-by-pass specs.

**Benefits:**
Will proactively alert devs if any changes to Catalyst or the unified compiler would cause issues within `qml.specs`.

**Possible Drawbacks:**
Because the unit tests are placed here, changes to the `qml.specs` frontend may require changes to Catalyst as well

**Related GitHub Issues:**
[sc-102889]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -419,6 +419,9 @@
 
 <h3>Internal changes ⚙️</h3>
 
+* Update references to `TransformProgram` with `CompilePipeline`
+  [(#2314)](https://github.com/PennyLaneAI/catalyst/pull/2314)
+
 * xDSL and xDSL-JAX are now dependencies of Catalyst.
   [(#2282)](https://github.com/PennyLaneAI/catalyst/pull/2282)
 
--- a/frontend/catalyst/device/qjit_device.py
+++ b/frontend/catalyst/device/qjit_device.py
@@ -28,6 +28,7 @@ from typing import Any, Dict, Optional
 
 import pennylane as qml
 from jax.interpreters.partial_eval import DynamicJaxprTracer
+from pennylane import CompilePipeline
 from pennylane.devices.capabilities import (
     DeviceCapabilities,
     ExecutionCondition,
@@ -38,7 +39,6 @@ from pennylane.transforms import (
     split_non_commuting,
     split_to_single_terms,
 )
-from pennylane.transforms.core import TransformProgram
 
 from catalyst.device.decomposition import (
     catalyst_decompose,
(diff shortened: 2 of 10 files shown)


jzaia18:
Note that this won't pass CIs until the main PR is merged: https://github.com/PennyLaneAI/pennylane/pull/8606. I would recommend running this locally (can confirm that at the time of writing, all specs tests pass)

jzaia18:
> Thanks Jake. I might have missed it because there are tests with and without capture, but I don't think I saw any tests for tape transforms + MLIR/xDSL transforms together

Good catch. I added a test case explicitly for this, and also expanded many of the existing test cases to use both frontends wherever possible

frontend/test/pytest/test_specs.py
```diff
@@ -284,6 +284,69 @@ def test_marker(self, simple_circuit):
...
+
+        simple_circuit = qml.transforms.cancel_inverses(
+            simple_circuit
+        )  # Has to be applied as a tape transform
+        simple_circuit = qml.transforms.undo_swaps(
+            simple_circuit
+        )  # No actual swaps to undo, but forces normal tape transform
+        simple_circuit = qml.transforms.merge_rotations(
+            simple_circuit
+        )  # Can be applied as an MLIR pass
```
mudit2812, line 298: How is merge_rotations being applied as an MLIR without capture enabled? If capture is disabled, catalyst passes need to be applied using the ones from `catalyst.passes`.

jzaia18: Is this still the case? When printing the raw MLIR I can see the pass being applied in there. I think there was a PR from Christina that fixed this a few weeks ago but I don't precisely remember

mudit2812: You're right, it also works with capture disabled. I forgot about that improvement 😅

mudit2812: In that case why not use both frontends?

jzaia18: `undo_swaps` doesn't work with capture

jzaia18: But if you have an example of a pass which doesn't have an MLIR implementation and works with both frontends we can use that one instead

mudit2812: As far as I know, any transform that returns a single tape and has a null processing function should work (at least for a program with no control flow). It's interesting that `undo_swaps` doesn't work. Looking at the implementation, it might be because of the use of `map_wires`? I'm not sure

Issue #2194, opened 2025-11-13
`sortTopologically` incorrectly moves nullary operations (Breaks Order)


sengthai:
Currently, both `commute_ppr` and `merge_ppr_ppm` use the `sortTopologically` function to sort a range of operations in said block in topological order. 

However, this approach will lead to a problem where one of nullary operation (e.g. `quantum.device_release`) is reorder incorrectly (should not move). 

Example:  `test.mlir`
```mlir
func.func @test_commute_1(%q1 : !quantum.bit){
    %100 = arith.constant 100 : i64
    quantum.device shots(%100) ["librtd_null_qubit.dylib", "NullQubit", "{'track_resources': False}"]
    %0 = qec.ppr ["Z"](4) %q1 : !quantum.bit
    %1 = qec.ppr ["Z"](8) %0 : !quantum.bit
    %2 = qec.ppr ["Z"](4) %1 : !quantum.bit
    quantum.device_release
    func.return
}
```
After perform `quantum-opt --commute-ppr test.mlir`, we got:

```mlir
  func.func @test_commute_1(%arg0: !quantum.bit) {
    %0 = qec.ppr ["Z"](8) %arg0 : !quantum.bit
    quantum.device_release
    %1 = qec.ppr ["Z"](8) %0 : !quantum.bit
    %2 = qec.ppr ["Z"](8) %1 : !quantum.bit
    return
  }
```

`quantum.device_release` should not be moved to the top.

**Proposed solution:**
- set the list of operation that needed to order in `isOperandReady` in `sortTopologically`. 

**This is not only solve the problem, but also reduce the sorting time as well.**

paul0403:
https://github.com/llvm/llvm-project/blob/3d4156700ef2ea678c3dabf796bc8622ac0d7ecb/mlir/lib/Analysis/TopologicalSortUtils.cpp#L102 this convenience version only takes in a block and will sort everything in the block

https://github.com/llvm/llvm-project/blob/3d4156700ef2ea678c3dabf796bc8622ac0d7ecb/mlir/lib/Analysis/TopologicalSortUtils.cpp#L56 this underlying version seems to be able to explicitly take in a list of ops to sort

PR #2273, opened 2025-12-08
Lower RTIO to LLVM IR

rniczh:
**Context:**

The conversion from RTIO to LLVM IR can be divided into three steps:
1. Schedule for finding feasible groups to meet the constraints of hardware
2. Extract frequency setting for each group (chain of frequency setting ahead of time)
3. Generate ARTIQ’s primitives in LLVM IR

**Scheduling**

The purpose of schedule step is to further distinguish which regions must be executed serially based on HW constraints derived from ARTIQ. Here we employ BFS to identify the first feasible group and repeat it with the remaining RTIO operation to find all feasible groups. 

Currently, a major constraint is that events sharing the same channel must have a unique frequency setting. In other words, if a group contains the same channel but with different frequencies, it’s considered an illegal group.

**Frequency Setting**

For every feasible group, we execute their frequency setting serially before the group itself starts to run. This ensures that the corresponding channel has been updated to the correct frequency via SPI register and I/O updates on the device before the group execution begins. The IR graph shown here illustrates this intermediate state. Normally, the process lowers directly to LLVM IR, so this step represents a transient state within the IR, which will not appear in the final output of the pass, since the result should already be in LLVM IR.

And there is only one SPI bus available at the device. So each time you can only update one frequency. We also adopt a simple optimization: If that channel has already been updated to a certain frequency then we don’t need to do the same frequency setting again.

**Lowering to LLVM IR**

Converting from RTIO to LLVM IR is straight forward:

- `rtio.empty`
   Just call `now_mu()` to get the current timestamp of ARTIQ and return it

- `rtio.pulse`
   Involve three steps in `rtio.pulse`, we split each pulse to three steps during the frequency setting part.
   1. Setting Frequency to SPI
   2. Slack delay to prevent underflow (since setting the frequency to SPI take time on CPU)
       Because frequency setting takes time, it involves the setting from CPU to update value to SPI register, needing to do the delay on the cursor at the same time to ensure no underflow issue happens on the device.
   3. TTL ON + Delay + TTL OFF

   step 1 + 2:
   ```
    tail call fastcc void @at_mu(i64 %13)
    tail call fastcc void @_rtio_set_frequency(i32 5120, double 0x41721EAC006AA488, double 0.000000e+00, double 1.000000e+00)
    %16 = tail call fastcc i64 @now_mu()
    tail call fastcc void @at_mu(i64 %13)
    tail call fastcc void @delay_mu(i64 10000)
    %17 = tail call fastcc i64 @now_mu()
    %18 = icmp sgt i64 %16, %17
    %19 = select i1 %18, i64 %16, i64 %17
    tail call fastcc void @at_mu(i64 %19)
    %20 = tail call fastcc i64 @now_mu()
    ```
   
    step 3:
    ```
    %9 = tail call fastcc i64 @__rtio_sec_to_mu(double 0x3E8646DB8A138F8C)
    tail call fastcc void @at_mu(i64 %8)
    tail call fastcc void @rtio_output(i32 4864, i32 1)
    tail call fastcc void @delay_mu(i64 %9)
    tail call fastcc void @rtio_output(i32 4864, i32 0)
    %10 = tail call fastcc i64 @now_mu()
   ```
    
- `rtio.sync`
    We sync the events by collecting their timestamps and find the maximum value from them and jump to there. And transfer the timestamp through `now_mu()` as well
    ```
    %46 = icmp sgt i64 %44, %45
    %47 = select i1 %46, i64 %44, i64 %45
    tail call fastcc void @at_mu(i64 %47)
    %48 = tail call fastcc i64 @now_mu()
    ```

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-100852]

### DIFF
--- a/mlir/lib/RTIO/Transforms/RTIOEventToARTIQ.cpp
+++ b/mlir/lib/RTIO/Transforms/RTIOEventToARTIQ.cpp
@@ -243,65 +243,75 @@ class PulseScheduler {
                             DenseMap<int32_t, rtio::RTIOPulseOp> &channelBoundary,
                             SmallVector<rtio::RTIOPulseOp> &boundaryConsumers)
     {
+        if (channelPulses.size() > 1 && !channelBoundary.empty()) {
+            return createSyncEvent(channelLastPulse, channelBoundary, boundaryConsumers);
+        }
+        return collectNextEvents(channelLastPulse, channelBoundary, boundaryConsumers);
+    }
+
+    SmallVector<Value> createSyncEvent(DenseMap<int32_t, rtio::RTIOPulseOp> &channelLastPulse,
+                                       DenseMap<int32_t, rtio::RTIOPulseOp> &channelBoundary,
+                                       SmallVector<rtio::RTIOPulseOp> &boundaryConsumers)
+    {
+        // Collect events to sync
+        SmallVector<Value> eventsToSync;
+        for (auto &entry : channelLastPulse) {
+            rtio::RTIOPulseOp pulse = entry.second;
+            eventsToSync.push_back(pulse.getEvent());
+        }
+
+        auto anyPulse = channelLastPulse.begin()->second;
+        builder.setInsertionPointAfter(anyPulse);
+
+        auto eventType = rtio::EventType::get(builder.getContext());
+        Value syncEvent =
+            builder.create<rtio::RTIOSyncOp>(anyPulse.getLoc(), eventType, eventsToSync);
+
+        // Update boundaries and consumers
+        for (auto &[_, pulse] : channelBoundary) {
+            pulse.setWait(syncEvent);
+        }
+        for (auto pulse : boundaryConsumers) {
+            pulse.setWait(syncEvent);
+        }
+        for (auto &entry : channelLastPulse) {
+            rtio::RTIOPulseOp pulse = entry.second;
+            for (auto user : pulseConsumers[pulse]) {
+                auto userChannel = extractChannelId(user.getChannel());
+                if (!channelBoundary.count(userChannel) || channelBoundary[userChannel] != user) {
+                    if (user.getWait() == pulse.getEvent()) {
+                        user.setWait(syncEvent);
+                    }
+                }
+            }
+        }
+
+        return {syncEvent};
+    }
+
+    SmallVector<Value> collectNextEvents(DenseMap<int32_t, rtio::RTIOPulseOp> &channelLastPulse,
+                                         DenseMap<int32_t, rtio::RTIOPulseOp> &channelBoundary,
+                                         SmallVector<rtio::RTIOPulseOp> &boundaryConsumers)
+    {
         SmallVector<Value> nextEvents;
 
-        if (channelPulses.size() > 1 && !channelBoundary.empty()) {
-            // Collect events to sync
-            SmallVector<Value> eventsToSync;
-            for (auto &entry : channelLastPulse) {
-                rtio::RTIOPulseOp pulse = entry.second;
-                eventsToSync.push_back(pulse.getEvent());
-            }
-
-            auto anyPulse = channelLastPulse.begin()->second;
-            builder.setInsertionPointAfter(anyPulse);
-
-            auto eventType = rtio::EventType::get(builder.getContext());
-            Value syncEvent =
-                builder.create<rtio::RTIOSyncOp>(anyPulse.getLoc(), eventType, eventsToSync);
-
-            // Update boundaries and consumers
-            for (auto &[_, pulse] : channelBoundary) {
-                pulse.setWait(syncEvent);
-            }
-            for (auto pulse : boundaryConsumers) {
-                pulse.setWait(syncEvent);
-            }
-            for (auto &entry : channelLastPulse) {
-                rtio::RTIOPulseOp pulse = entry.second;
-                for (auto user : pulseConsumers[pulse]) {
-                    auto userChannel = extractChannelId(user.getChannel());
-                    if (!channelBoundary.count(userChannel) ||
-                        channelBoundary[userChannel] != user) {
-                        if (user.getWait() == pulse.getEvent()) {
-                            user.setWait(syncEvent);
-                        }
-                    }
-                }
-            }
-
-            nextEvents.push_back(syncEvent);
+        for (auto &entry : channelBoundary) {
+            rtio::RTIOPulseOp pulse = entry.second;
+            nextEvents.push_back(pulse.getWait());
         }
-        else {
-            // No sync needed
-            for (auto &entry : channelBoundary) {
-                rtio::RTIOPulseOp pulse = entry.second;
-                nextEvents.push_back(pulse.getWait());
+        if (!boundaryConsumers.empty() && !channelLastPulse.empty()) {
+            rtio::RTIOPulseOp firstPulse = channelLastPulse.begin()->second;
+            Value lastEvent = firstPulse.getEvent();
+            for (auto pulse : boundaryConsumers) {
+                pulse.setWait(lastEvent);
             }
-            if (!boundaryConsumers.empty() && !channelLastPulse.empty()) {
-                rtio::RTIOPulseOp firstPulse = channelLastPulse.begin()->second;
-                Value lastEvent = firstPulse.getEvent();
-                for (auto pulse : boundaryConsumers) {
-                    pulse.setWait(lastEvent);
-                }
-                nextEvents.push_back(lastEvent);
-            }
-            for (auto &entry : channelLastPulse) {
-                rtio::RTIOPulseOp pulse = entry.second;
-                for (auto *user : pulse.getEvent().getUsers()) {
-                    if (auto syncOp = dyn_cast<rtio::RTIOSyncOp>(user)) {
-                        nextEvents.push_back(syncOp.getSyncEvent());
-                    }
+            nextEvents.push_back(lastEvent);
+        }
+        for (auto &entry : channelLastPulse) {
+            rtio::RTIOPulseOp pulse = entry.second;
+            for (auto *user : pulse.getEvent().getUsers()) {
+                if (auto syncOp = dyn_cast<rtio::RTIOSyncOp>(user)) {
+                    nextEvents.push_back(syncOp.getSyncEvent());
                 }
             }
         }


mlir/lib/RTIO/Transforms/ARTIQRuntimeBuilder.hpp
```diff
@@ -0,0 +1,400 @@
...
+
+    // TTL operations
+    void ttlOn(Value channelAddr) { rtioOutput(channelAddr, constI32(1)); }
+
+    void ttlOff(Value channelAddr) { rtioOutput(channelAddr, constI32(0)); }
+
+    // Constant creation helpers
+    Value constI32(int32_t val)
+    {
+        return builder.create<arith::ConstantOp>(getLoc(), builder.getI32IntegerAttr(val));
```
mehrdad2m, line 185: Are we relying on the default `MLIRToLLVMDialectConversion` pipeline to lower this to llvmIR or would there be a custom pipeline needed?

mlir/tools/quantum-opt/quantum-opt.cpp
```diff
@@ -19,10 +19,13 @@
 #include "llvm/Support/CommandLine.h"
 #include "llvm/Support/raw_ostream.h"
 
+#include "mlir/Conversion/ControlFlowToLLVM/ControlFlowToLLVM.h"
+#include "mlir/Conversion/FuncToLLVM/ConvertFuncToLLVM.h"
```
mehrdad2m: Why do we need to modify `quantum-opt`?

mlir/tools/quantum-opt/CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ set(LIBS
     ExternalStablehloLib
     MLIROptLib
     MLIRRegisterAllDialects
+    MLIRRegisterAllExtensions
```
mehrdad2m: Can you also explain why we need this?

mlir/tools/quantum-opt/quantum-opt.cpp
mehrdad2m: I mean my question is what new thing are we adding now that requires these two files.

mlir/lib/RTIO/Transforms/RTIOEventToARTIQPatterns.cpp
```diff
@@ -0,0 +1,288 @@
...
+
+        Value newTime = artiq.nowMu();
+        rewriter.replaceOp(op, newTime);
+        return success();
+    }
+
+    LogicalResult lowerSlackPulse(RTIOPulseOp op, ConversionPatternRewriter &rewriter,
+                                  ARTIQRuntimeBuilder &artiq) const
+    {
+        artiq.delayMu(artiq.constI64(ARTIQHardwareConfig::freqSetSlackDelay));
```
mehrdad2m, line 80: How did you determine what this delay should be? is this the same number that OQD uses on their compiler?

mlir/lib/RTIO/Transforms/ARTIQRuntimeBuilder.hpp
rniczh: Yes, we will still rely on the original MLIR to LLVM conversion pipeline to lower. But we will need to modify the original path to ensure it only generate object file for link.

mlir/lib/RTIO/Transforms/RTIOEventToARTIQPatterns.cpp
rniczh: Yes, and this slack value is only for frequency setting, so it's fixed.

mlir/tools/quantum-opt/quantum-opt.cpp
rniczh: No, we don't need them, they are redundant, I will remove them

rniczh: Done https://github.com/PennyLaneAI/catalyst/pull/2273/commits/65f33887c6fbe112ec332be91406b736cee172ed

mlir/tools/quantum-opt/CMakeLists.txt
rniczh: same as https://github.com/PennyLaneAI/catalyst/pull/2273#discussion_r2604221447
https://github.com/PennyLaneAI/catalyst/pull/2273/commits/65f33887c6fbe112ec332be91406b736cee172ed

mlir/lib/RTIO/Transforms/RTIOEventToARTIQPatterns.cpp
rniczh: This is the code in artiq bare metal compiler. Currently for frequency setting, they guarantee that always has 10μs delay to avoid underflow
```
ast.Expr(
    value=ast.Call(
        func=ast.Name(id="delay", ctx=ast.Load()),
        args=[ast.Constant(value=1e-5)],
        keywords=[],
    )
),
```

mlir/tools/quantum-opt/CMakeLists.txt
mehrdad2m: `MLIRRegisterAllExtensions` still is remaining in the `mlir/tools/quantum-opt/CMakeLists.txt`

mehrdad2m (commented):
Note that at the time of the review, there is no tests added. We have to add tests to check the generated IR

mlir/tools/quantum-opt/CMakeLists.txt
rniczh: Thanks, I was missing to include that file change https://github.com/PennyLaneAI/catalyst/pull/2273/commits/4eed6d5810ba8504313e15b9587a41adcd2c5a86

mlir/lib/RTIO/Transforms/RTIOEventToARTIQPatterns.cpp
```diff
@@ -0,0 +1,288 @@
...
+        }
+
+        Operation *defOp = from.getDefiningOp();
+        if (!defOp) {
+            return false;
+        }
+
+        // TODO shouldn't use recursive call here
+        if (auto pulse = dyn_cast<RTIOPulseOp>(defOp)) {
+            return canReach(target, pulse.getWait(), visited);
```
mehrdad2m: A bit hard to verify without tests, but the recursive call here looks a bit scary to me. Would it be too much work to address the TODO?

rniczh: No, I think it's pretty straightforward to do that. Let me to address this TODO

rniczh: Add I also added the lit test https://github.com/PennyLaneAI/catalyst/pull/2273/commits/6b1c4ea3e642fd1d4ee38d76522540985b4d4762

rniczh: Fixed, addressed this TODO https://github.com/PennyLaneAI/catalyst/pull/2273/commits/26d2279def80ea74fed5f0919378d5c6287f05c8

mehrdad2m: Thanks for adding the test. Given that we have pretty much all the pieces of the puzzle, would it make sense to also have some integration lit tests that lowers an actual circuit and checks the lowered IR?

rniczh: @mehrdad2m  The circuit you mentioned here is in IR level or from the python?

rniczh: Hmm...I was thinking how to handle those json and toml files. If in the IR level, I could probably place them at the same folder initially and run the pipeline from the specific IR. 
But if considering the test from python level, I think I would need to open an additional PR (end2end pipeline) to add the integrated test.

mlir/lib/RTIO/Transforms/RTIOEventToARTIQPatterns.cpp
mehrdad2m: Nice, I would like move these hardcoded numbers out of the code sometime. Not in this PR though, but I think it would be nice to get them from a config file or something.

mlir/lib/RTIO/Transforms/RTIOEventToARTIQPatterns.cpp
mehrdad2m: I meant from python. Sure, I am fine to have them as a separate PR.

mlir/include/RTIO/Transforms/ARTIQRuntimeBuilder.h
```diff
@@ -0,0 +1,429 @@
...
+        auto module = getModule();
+        if (auto func = module.lookupSymbol<LLVM::LLVMFuncOp>(name)) {
+            return func;
+        }
+
+        OpBuilder::InsertionGuard guard(builder);
+        builder.setInsertionPointToStart(module.getBody());
+        return builder.create<LLVM::LLVMFuncOp>(contextOp->getLoc(), name, funcTy,
+                                                LLVM::Linkage::External);
+    }
```
paul0403, line 223: Any issues with using the existing `ensureFunctionDeclaration`? 

https://github.com/PennyLaneAI/catalyst/blob/main/mlir/include/Catalyst/Utils/EnsureFunctionDeclaration.h
https://github.com/PennyLaneAI/catalyst/blob/main/mlir/lib/Catalyst/Utils/EnsureFunctionDeclaration.cpp

This util is used in all `--convert-dialect-to-llvm` I believe

mlir/lib/RTIO/Transforms/ARTIQRuntimeBuilder.hpp
```diff
@@ -0,0 +1,429 @@
+// Copyright 2025 Xanadu Quantum Technologies Inc.
```
paul0403, line 1: Nitpick: I feel like since this file has code instead of just declarations, it would make more sense to make it a `.hpp` instead of `.h`

mlir/lib/RTIO/Transforms/RTIOEventToARTIQPatterns.cpp
```diff
@@ -0,0 +1,301 @@
...
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/LLVMIR/LLVMDialect.h"
+#include "mlir/IR/BuiltinOps.h"
+#include "mlir/Transforms/DialectConversion.h"
+
+#include "RTIO/IR/RTIOOps.h"
+#include "RTIO/Transforms/ARTIQRuntimeBuilder.h"
+#include "RTIO/Transforms/Patterns.h"
+#include "Utils.h"
```
paul0403: Let's move things like `Utils.h` to the `include` folder!

mlir/lib/RTIO/Transforms/RTIOEventToARTIQ.cpp
```diff
@@ -0,0 +1,553 @@
...
+//===----------------------------------------------------------------------===//
+// Pass Implementation
+//===----------------------------------------------------------------------===//
+
+struct RTIOEventToARTIQPass : public impl::RTIOEventToARTIQPassBase<RTIOEventToARTIQPass> {
+    using RTIOEventToARTIQPassBase::RTIOEventToARTIQPassBase;
+
+    void runOnOperation() override
+    {
+        ModuleOp module = getOperation();
```
paul0403, line 447: `module` is a keyword in cpp!

Let's call it `mod` or `moduleOp` (lowercase) or something

mlir/lib/RTIO/Transforms/RTIOEventToARTIQPatterns.cpp
```diff
@@ -0,0 +1,301 @@
...
+// Conversion Patterns
+//===----------------------------------------------------------------------===//
+
+struct PulseOpLowering : public OpConversionPattern<RTIOPulseOp> {
+    using OpConversionPattern::OpConversionPattern;
+
+    LogicalResult matchAndRewrite(RTIOPulseOp op, OpAdaptor adaptor,
+                                  ConversionPatternRewriter &rewriter) const override
+    {
+        ARTIQRuntimeBuilder artiq(rewriter, op);
```
paul0403, line 42: As far as I understand it, the `ARTIQRuntimeBuilder` class is to be interacted only lazily right? In other words, there is no global state (except the IR it sees) that this builder class needs to keep track of? And different instances of this class do not need to talk to each other?

If that is the case, can we maybe avoid creating one `ARTIQRuntimeBuilder` instance per pattern class; actually not even just one instance per pattern, but once instance per pattern _invocation_? Because the builder class looks quite big. 

How about making a single `ARTIQRuntimeBuilder` instance in `runOnOperation()` of the overall `RTIOEventToARTIQPass`, since you already have a `OpBuilder builder(ctx);` there, and send it into the patterns via the `populateBlahBlahPattern()` functions? :thinking:

mlir/lib/RTIO/Transforms/RTIOEventToARTIQ.cpp
paul0403: Same with all other places where you used `module` as a variable name

mlir/lib/RTIO/Transforms/RTIOEventToARTIQ.cpp
```diff
@@ -0,0 +1,553 @@
...
+        DenseMap<int32_t, rtio::RTIOPulseOp> channelBoundary;
+        SmallVector<rtio::RTIOPulseOp> boundaryConsumers;
+
+        // Initial
+        for (auto pulse : consumers) {
+            if (processedPulses.contains(pulse)) {
+                continue;
+            }
+
+            int32_t channel = extractChannelId(pulse.getChannel());
```
paul0403, line 149: Is channel ID guaranteed to be static?

I guess at the user level, this means circuits with dynamic user indices are not supported?
```python
@qjit
@qnode(...)
def circuit(i):
   qml.RX(3.14, wires=i)
```

What would the error message be in these cases? Can you try? 

What I am trying to avoid is these errors giving random mlir messages, like "can't cast `Value` to `int32_t`" or something (although I guess that would be an error during `make dialects`, but you get my point)

mlir/lib/RTIO/Transforms/RTIOEventToARTIQ.cpp
```diff
@@ -0,0 +1,553 @@
...
+        }
+
+        // Lowering to LLVM
+        if (failed(lowerToLLVM(module))) {
+            return signalPassFailure();
+        }
+    }
+
+  private:
+    static bool sameChannelSameFrequency(RTIOPulseOp ref, RTIOPulseOp candidate)
```
paul0403, line 498: This can take in const refs right? Since it's just reading to compare

mlir/lib/RTIO/Transforms/RTIOEventToARTIQ.cpp
```diff
@@ -0,0 +1,553 @@
...
+
+        // Record group
+        recordGroup(channelPulses);
+
+        // Create sync and update dependencies
+        return createSyncAndUpdateDeps(channelPulses, channelLastPulse, channelBoundary,
+                                       boundaryConsumers);
+    }
+
+    bool canJoinGroup(rtio::RTIOPulseOp cand,
```
paul0403, line 177: Same here, this can be const ref pulse op?

mlir/lib/RTIO/Transforms/RTIOEventToARTIQ.cpp
```diff
@@ -0,0 +1,553 @@
...
+            return ref.getFrequency() == candidate.getFrequency();
+        }
+        return true;
+    }
+
+    static void sortAllBlocks(ModuleOp module)
+    {
+        module.walk([](func::FuncOp funcOp) {
+            for (auto &block : funcOp.getBody()) {
+                sortTopologically(&block);
```
paul0403, line 510: We recently discovered that this `sortTopologically` helper will even reorder ops that have side effects! That is, this sorting does not consider side effects at all https://github.com/PennyLaneAI/catalyst/issues/2194

Given the nature of rtio, I would imagine implicit ordering of ops from side effects are actually very important. We did also discover that there's a version of the util that can take in an explicit list of ops to sort. I wonder if it's safer to use that version.

What is the purpose of this sort here? Is it necessary in the first place?

mlir/lib/RTIO/Transforms/Utils.hpp
```diff
@@ -0,0 +1,74 @@
...
+        int64_t channelId = channelIdAPInt.getSExtValue();
+        int32_t addr = static_cast<int32_t>((channelId + channelBase) << 8);
+        return builder.create<mlir::arith::ConstantOp>(loc, builder.getI32IntegerAttr(addr));
+    }
+
+    // Dynamic channel
+    mlir::Value offset =
+        builder.create<mlir::arith::ConstantOp>(loc, builder.getI32IntegerAttr(channelBase));
+    mlir::Value sum = builder.create<mlir::arith::AddIOp>(loc, channelValue, offset);
+    mlir::Value shift = builder.create<mlir::arith::ConstantOp>(loc, builder.getI32IntegerAttr(8));
```
paul0403: Hmm, I thought we could only handle static channels. Or am I misunderstanding what this is?

mlir/lib/RTIO/Transforms/RTIOEventToARTIQPatterns.cpp
```diff
@@ -0,0 +1,301 @@
...
+        }
+        else {
+            rewriter.replaceOpWithNewOp<RTIOSyncOp>(op, op.getType(), requiredEvents);
+        }
+        return success();
+    }
+
+  private:
+    // Check if target is reachable from 'from' by traversing event dependencies.
+    static bool canReach(Value target, Value from)
```
paul0403, line 235: const ref?

paul0403 (commented):
Really amazing! Thanks Hongsheng! :100:

mlir/include/RTIO/Transforms/ARTIQRuntimeBuilder.h
rniczh: wow, I didn't notice that we already have an ensure function

paul0403: Yes! That one even looks (almost) exactly the same as yours :laughing:

mlir/lib/RTIO/Transforms/RTIOEventToARTIQ.cpp
rniczh: Currently, it's not supported by ion pipeline, it will run into an legalization error when converting to ion pulses. So at that point, I think we currently might not need to worry about that. 

```
catalyst.utils.exceptions.CompileError: catalyst failed with error code 1: Failed to run pipeline: oqd_pipeline
Compilation failed:
circuit:23:12: error: Impossible to determine the original qubit because of dynamism.
      %5 = "quantum.custom"(%4, %3) {gate_name = "RX", operandSegmentSizes = array<i32: 1, 1, 0, 0>, resultSegmentSizes = array<i32: 1, 0>} : (f64, !quantum.bit) -> !quantum.bit
           ^
circuit:23:12: note: see current operation: %6 = "quantum.custom"(%2, %5) <{gate_name = "RX", operandSegmentSizes = array<i32: 1, 1, 0, 0>, resultSegmentSizes = array<i32: 1, 0>}> : (f64, !quantum.bit) -> !quantum.bit
circuit:23:12: error: failed to legalize operation 'quantum.custom' that was explicitly marked illegal
      %5 = "quantum.custom"(%4, %3) {gate_name = "RX", operandSegmentSizes = array<i32: 1, 1, 0, 0>, resultSegmentSizes = array<i32: 1, 0>} : (f64, !quantum.bit) -> !quantum.bit
           ^
circuit:23:12: note: see current operation: %6 = "quantum.custom"(%2, %5) <{gate_name = "RX", operandSegmentSizes = array<i32: 1, 1, 0, 0>, resultSegmentSizes = array<i32: 1, 0>}> : (f64, !quantum.bit) -> !quantum.bit
While processing 'GatesToPulsesPass' pass While processing 'mlir::detail::OpToOpPassAdaptor' pass Failed to lower MLIR module
```

Btw, If we want to support the variable channel, it need to do some change during the grouping schedule, it need to ensure no other pulses occupies the device at the same time.

mlir/lib/RTIO/Transforms/RTIOEventToARTIQ.cpp
rniczh: Since all RTIO ops are async by design, if it need the order, then it should be connected. If not connected, it can be placed at anywhere.

The purpose of this sorting is to fix the dominate issue when operations are inserted to the graph.

mlir/lib/RTIO/Transforms/Utils.hpp
rniczh: Yes, I found that during development but forgot to delete it. Right now it only have static channels, so this part won't be covered.

mlir/lib/RTIO/Transforms/RTIOEventToARTIQ.cpp
paul0403: Ah this makes sense, thanks for explaining!

mlir/lib/RTIO/Transforms/ARTIQRuntimeBuilder.hpp
dime10: I'm not sure that convention has anything to do with whether there is code in the header or not, rather `.hpp` is used to (explicitly) indicate a C++ header. Unfortunately, this convention is not universal and some projects use `.h` even for C++ projects. In MLIR, this is the case for anything in the "public" includes of the project, whereas `.hpp` is used for "private" headers colocated with the source files.

mlir/lib/RTIO/Transforms/RTIOEventToARTIQ.cpp
paul0403: Nice, so it is already a meaningful error message. No action needed then!

mlir/lib/RTIO/Transforms/RTIOEventToARTIQPatterns.cpp
dime10: Please read https://mlir.llvm.org/docs/Rationale/UsageOfConst/

mlir/lib/RTIO/Transforms/RTIOEventToARTIQPatterns.cpp
dime10: Somewhat related to my comment above, but please consider whether helper code is actually reusable and would benefit from making it "public", or whether it's very local to the current module and there is no reason to move it into the public includes.

mlir/lib/RTIO/Transforms/RTIOEventToARTIQPatterns.cpp
paul0403: Wow, I never knew about this! Thanks David for pointing this out!

>  you should never see a const method on Operation, should never see the type const Value, and you shouldn’t feel bad about this

It somehow read my mind (about feeling bad about `const`)

mlir/lib/RTIO/Transforms/RTIOEventToARTIQPatterns.cpp
paul0403: I see, given these guidelines I think the most sensible thing is to make these two files (currently `include/.../ARTIQRuntimeBuilder.h` and `lib/.../Utils.h`) both `.hpp` files next to the `lib` folder cpp source files?

mlir/include/RTIO/Transforms/ARTIQRuntimeBuilder.h
rniczh: I used that util https://github.com/PennyLaneAI/catalyst/pull/2273/commits/473555656544a4954bb1d2425d069497db412a16

mlir/lib/RTIO/Transforms/RTIOEventToARTIQPatterns.cpp
paul0403: I just realized this builder is sensitive to the op being matched. All good then.

mlir/lib/RTIO/Transforms/RTIOEventToARTIQ.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2273/commits/cbd628de244e872e2dd8a0356817fbd13f69ea80

mlir/lib/RTIO/Transforms/Utils.hpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2273/commits/52221ab20989426de0167366d0fd54acdef76e1f

PR #2274, opened 2025-12-08
Miscellaneous fixes to the unified compiler

mudit2812:
* Update MLIR graph tests so that they are skipped if at least one of the following conditions is not met:
  * The Python `graphviz` package is not installed
  * The system-level `graphviz` package is not installed
* Update a test for the xDSL Quantum dialect that was causing the `--force-flaky` workflow to fail
* Update version of PennyLane in `.dep-versions`
* Move changelog entry for `qml.specs` integration with xDSL to the correct section

### DIFF
--- a/frontend/test/pytest/python_interface/inspection/test_mlir_graph.py
+++ b/frontend/test/pytest/python_interface/inspection/test_mlir_graph.py
@@ -23,7 +23,7 @@ pytestmark = pytest.mark.xdsl
 xdsl = pytest.importorskip("xdsl")
 graphviz = pytest.importorskip("graphviz")
 
-if run(["which", "dot"], check=False).returncode != 0:
+if run(["/usr/bin/which", "dot"], check=False).returncode != 0:
     pytest.skip(reason="Graphviz isn't installed.")
 
 
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -33,4 +33,4 @@ lxml_html_clean
 --extra-index-url https://test.pypi.org/simple/
 pennylane-lightning-kokkos==0.44.0-dev16
 pennylane-lightning==0.44.0-dev16
-pennylane==0.44.0-dev44
+pennylane==0.44.0-dev48


frontend/test/pytest/python_interface/inspection/test_mlir_graph.py
```diff
@@ -22,6 +23,10 @@
 xdsl = pytest.importorskip("xdsl")
 graphviz = pytest.importorskip("graphviz")
 
+if run(["which", "dot"], check=False).returncode != 0:
```
mudit2812: There is a CodeFactor warning here because `which` is not being specified an an absolute executable path. But, `which` is a built-in command in UNIX systems. I'm not really sure how I'm supposed to use the absolute path for `which`. It seems reasonable to ignore this warning IMO, but this is generally a security vulnerability so I wanted to note this.

dime10: @rashidnhm what do you think?

.dep-versions
```diff
@@ -10,7 +10,7 @@ enzyme=v0.0.203
 
 # For a custom PL version, update the package version here and at
 # 'doc/requirements.txt'
-pennylane=0.44.0-dev44
+pennylane=0.44.0-dev48
```
paul0403, line 13: Can you update the one in `doc/requirements.txt` too? :pray:

frontend/test/pytest/python_interface/inspection/test_mlir_graph.py
rashidnhm: I'm not seeing any crazy risks ... But let me check how much of a hassle it is to use absolute path

rashidnhm: Is this intended to run on Windows (native) ever? If not, we might be able to just get away with `/usr/bin/which` for the first argument (should be valid within WSL as well).

mudit2812: @rashidnhm I don't see wheel build workflows for windows, so I would assume no. I'll update the command to use `/usr/bin/which` 🎉

paul0403: Yeah catalyst doesn't have windows. I think it's even explicitly said so in the docs.

PR #2275, opened 2025-12-08
Fix wheels from nanobind update

paul0403:
**Context:**
Nanobind updated an hour ago, included is an update on how they handle their `tsl::robin-map` dependency.
https://github.com/wjakob/nanobind/pull/1212

Anti-pin against the broken nanobind version.

**Benefits:**
Wheels work.

### DIFF
--- a/.github/workflows/constants.yaml
+++ b/.github/workflows/constants.yaml
@@ -68,7 +68,6 @@ jobs:
       - name: LLVM version
         id: llvm_version
         run: echo "llvm_version=$(grep llvm .dep-versions | awk -F '=' '{ print $2 }')" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -7,7 +7,7 @@ pip>=22.3
 # this bug was fixed in 2.0.1 (https://github.com/numpy/numpy/pull/26995)
 numpy!=2.0.0
 # llvm requires nanobind 2.9 or higher
-nanobind>=2.9
+nanobind==2.9.2
```
dime10: If we know the bounds then I think that's better (semantically more meaningful)
```suggestion
# nanobind 2.10 is not compatible with ... / broke X
nanobind>=2.9,<2.10
```

paul0403: Makes sense

paul0403: well, pinning to 2.9.2 doesn't solve the original error :melting_face:

PR #2278, opened 2025-12-08
Remove xdsl-plugin entry point from setup.py

mehrdad2m:
**Context:**
We need to remove xdsl entry point form Catalyst as we no longer need to use the plugin route to apply xdsl passes.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-103076]

### DIFF
--- a/frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_diagonalize_measurements.py
+++ b/frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_diagonalize_measurements.py
@@ -24,7 +24,6 @@ xdsl = pytest.importorskip("xdsl")
 
 import pennylane as qml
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -147,7 +147,6 @@ def parse_dep_versions():
         "cuda_quantum.ops = catalyst.api_extensions",
         "cuda_quantum.qjit = catalyst.third_party.cuda:cudaqjit",
     ],
-    "catalyst.passes_resolution": ["catalyst_xdsl_plugin.passes = catalyst.passes.xdsl_plugin"],
```
mudit2812, line 150: Idk why but I thought this would've broken the callbacks. Glad to see that's not the case :)

mehrdad2m: yeah it was actually very minimally used in the code base

PR #2279, opened 2025-12-09
Integrate vjp with program capture

albi3ro:
**Context:**

Catalyst follow on from: https://github.com/PennyLaneAI/pennylane/pull/8736

Making it so that `qml.vjp` can be captured and lowered with program capture.

Just translates and adds a lowering for the to-be-added `vjp` primitive.

**Benefits:**

`qml.vjp` works with program capture.

**Possible Drawbacks:**

**Related GitHub Issues:**

[[sc-101549](https://app.shortcut.com/xanaduai/story/101549)]


.github/workflows/check-catalyst.yaml
```diff
@@ -477,6 +477,10 @@ jobs:
         sudo apt-get install -y graphviz
         make frontend
 
+    - name: Install PennyLane branch
+      run: |
+        pip install --no-deps --force git+https://github.com/PennyLaneAI/pennylane@capture-vjp
+
```
albi3ro:
```suggestion
```

.github/workflows/check-catalyst.yaml
```diff
@@ -565,6 +569,10 @@ jobs:
         python3 -m pip install -r requirements.txt
         make frontend
 
+    - name: Install PennyLane branch
+      run: |
+        pip install --no-deps --force git+https://github.com/PennyLaneAI/pennylane@capture-vjp
+
```
albi3ro:
```suggestion
```

.github/workflows/check-catalyst.yaml
```diff
@@ -628,6 +636,10 @@ jobs:
...
```
albi3ro:
```suggestion
```

frontend/catalyst/api_extensions/differentiation.py
```diff
@@ -546,13 +547,14 @@ def f(x):
     (Array([0.09983342, 0.04      , 0.02      ], dtype=float64),
      (Array([-0.43750208,  0.07      ], dtype=float64),))
     """
+    if qml.capture.enabled():
+        return qml.vjp(f, params, cotangents, method=method, h=h, argnums=argnums)
```
mudit2812, line 551: This is funny. With capture, `catalyst.vjp` calls `qml.vjp`, and without capture, `qml.vjp` calls `catalyst.vjp`

frontend/test/pytest/test_autograph.py
```diff
@@ -396,19 +396,26 @@ def fn(x: float):
         assert check_cache(inner)
         assert fn(3) == tuple([jax.numpy.array(2.0), jax.numpy.array(6.0)])
 
+    @pytest.mark.usefixtures("use_both_frontend")
     @pytest.mark.parametrize("vjp_func", [vjp, qml.vjp])
     def test_vjp_wrapper(self, vjp_func):
         """Test conversion is happening succesfully on functions wrapped with 'vjp'."""
 
+        if qml.capture.enabled() and vjp_func == vjp:  # pylint: disable=comparison-with-callable
+            pytest.skip("program capture needs qml.vjp")
```
mudit2812: Is this necessary? `catalyst.vjp` now dispatches to `qml.vjp` right?

mudit2812 (approved):
Non-blocking question

frontend/catalyst/api_extensions/differentiation.py
dime10: Makes perfect sense 🙆‍♂️

frontend/test/pytest/test_autograph.py
albi3ro:
```suggestion
```

albi3ro: So turns out capture autograph won't propogate through `catalyst.vjp`... but I think that should be ok.

frontend/test/pytest/test_autograph.py
```diff
@@ -396,19 +396,24 @@ def fn(x: float):
         assert check_cache(inner)
         assert fn(3) == tuple([jax.numpy.array(2.0), jax.numpy.array(6.0)])
 
+    @pytest.mark.usefixtures("use_both_frontend")
     @pytest.mark.parametrize("vjp_func", [vjp, qml.vjp])
     def test_vjp_wrapper(self, vjp_func):
         """Test conversion is happening succesfully on functions wrapped with 'vjp'."""
 
+
```
albi3ro, line 406:
```suggestion
        if qml.capture.enabled() and vjp_func == vjp:  # pylint: disable=comparison-with-callable
            pytest.skip("program capture needs qml.vjp")
```

mudit2812: Weird that this causes a failure. Any ideas on why?

frontend/test/pytest/test_autograph.py
dime10: Propagating autograph is relatively straightforward, the issue is just it has to be manually done for most decorators. It should be done though, we've had lots of issues opened just because X operator wasn't propagating autograph.

There is two mechanism in catalyst for this:
 - propagating through the decorator call directly: https://github.com/PennyLaneAI/catalyst/blob/5ab3a66d671e9f48f881de9b6a26f7988fda378e/frontend/catalyst/autograph/ag_primitives.py#L547
 - propagating through an already decorated function: https://github.com/PennyLaneAI/catalyst/blob/5ab3a66d671e9f48f881de9b6a26f7988fda378e/frontend/catalyst/autograph/ag_primitives.py#L576 (we set it up such that all decorators return a common base class which makes this quite straightforward)

frontend/test/pytest/test_autograph.py
```diff
@@ -396,19 +396,25 @@ def fn(x: float):
         assert check_cache(inner)
         assert fn(3) == tuple([jax.numpy.array(2.0), jax.numpy.array(6.0)])
 
+    @pytest.mark.usefixtures("use_both_frontend")
     @pytest.mark.parametrize("vjp_func", [vjp, qml.vjp])
     def test_vjp_wrapper(self, vjp_func):
         """Test conversion is happening succesfully on functions wrapped with 'vjp'."""
 
+        if qml.capture.enabled() and vjp_func == vjp:  # pylint: disable=comparison-with-callable
+            pytest.skip("program capture needs qml.vjp")
```
mudit2812: I would prefer using xfail here. It makes catching fixed issues much easier
```suggestion
            pytest.xfail("program capture autograph doesn't work with catalyst.vjp")
```

frontend/test/pytest/test_autograph.py
albi3ro: The problem is that we would have to add reference to `catalyst.vjp` in `pennylane/capture/autograph/ag_primitives.py`. We can probably go back into PL and do that, but seems not particularly worth it.

dime10: Got it, yeah not worth it.

frontend/catalyst/api_extensions/differentiation.py
```diff
@@ -564,7 +566,10 @@ def check_is_iterable(x, hint):
         grad_params = _check_grad_params(method, scalar_out, h, argnums, len(args_flatten), in_tree)
 
         args_argnums = tuple(params[i] for i in grad_params.argnums)
-        _, in_tree = tree_flatten(args_argnums)
```
dime10, line 567: PyTree shenanigans get really confusing once your dealing with vjp/jvps 😅 I'm not even sure the current implementation works sensibly (e.g. should the output pytree really depend on the argnums? I think to do it "properly", since the vjp shapes depend on the primal arguments, so should the vjp pytrees, but anyway...). 
Jax doesn't have argnums for vjp and always returns a tuple, but given that we do have it I think this is a reasonable extension to apply to the outer most tuple of the vjps 👍

dime10 (approved):
Thanks @albi3ro !

frontend/catalyst/api_extensions/differentiation.py
albi3ro: Yeah... how `argnums` should interact with `vjp` and what the shapes should need to be caused me a multiple headaches.  But ended up settling on this behaviour to match grad and jacobian.

PR #2281, opened 2025-12-09
Fix a logic during checking whether a device needs shots

paul0403:
**Context:**
There was an issue with the way we check whether a device requires shots or not, surfaced by the Qrack device.

Use `==` and `is` for checking the enum directly, instead of `hasattr` (which checks enum class attributes and hence is always `True`).

### DIFF
--- a/frontend/catalyst/device/qjit_device.py
+++ b/frontend/catalyst/device/qjit_device.py
@@ -640,7 +640,6 @@ def _requires_shots(capabilities):
     Checks if a device capabilities requires shots.
 
     A device requires shots if all of its MPs are finite shots only.
-    Shots is absolutely required if all MPs require it.
     """
     return all(
         ExecutionCondition.FINITE_SHOTS_ONLY in MP_conditions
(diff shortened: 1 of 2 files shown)


dime10 (approved):
Approved pending changelog (and test?)

frontend/catalyst/device/qjit_device.py
```diff
@@ -636,13 +640,9 @@ def _requires_shots(capabilities):
     Checks if a device capabilities requires shots.
 
     A device requires shots if all of its MPs are finite shots only.
-    If any of the MPs support modes other than finite shots, shots is not absolutely required.
+    Shots is absolutely required if all MPs require it.
```
joeycarter: Is this sentence redundant given the one above it?

paul0403: That is very true :sweat_smile:

paul0403: https://github.com/PennyLaneAI/catalyst/pull/2281/commits/e4ff095988474009d52b7fbc872a6d7b5976c324

PR #2282, opened 2025-12-09
Add `xdsl` and `xdsl-jax` as dependencies of Catalyst

mudit2812:
* Add `xdsl` and `xdsl-jax` to the requirements in `setup.py`
* Remove use of `pytest.importorskip("xdsl")` from the tests
* Add `XDSL_TESTS` argument to the `Makefile` for `make pytest`. `ON` by default
* Remove separate steps for xDSL tests in the GitHub workflows. Instead, xDSL tests are skipped by setting `XDSL_TESTS=OFF` or explicitly using the `-m "not xdsl"` pytest marker.
  * Question for reviewers: Should the `check-pl-compat.yml` workflow run tests that require graphviz to be installed?

[sc-105557]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -117,6 +117,8 @@
   spurred from `A Game of Surface Codes (arXiv1808.02892) <https://arxiv.org/pdf/1808.02892>`_.
   [(#2145)](https://github.com/PennyLaneAI/catalyst/pull/2145)
   [(#2233)](https://github.com/PennyLaneAI/catalyst/pull/2233)
+  [(#2284)](https://github.com/PennyLaneAI/catalyst/pull/2284)
+
 
   :class:`~.PauliRot` and :func:`~.pauli_measure` can be manipulated with Catalyst's existing passes
   for PPR-PPM compilation, which includes :func:`catalyst.passes.to_ppr`,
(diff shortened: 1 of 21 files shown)


setup.py
```diff
@@ -120,6 +132,14 @@ def parse_dep_versions():
 else:
     lightning_dep = f"pennylane-lightning>={lq_min_release}"
     kokkos_dep = ""
+if xdsl_version is not None:
```
dime10: I'm not sure we actually need to duplicate all this logic for the xdsl dependency.
For some context, the dep-versions file wasn't really meant as a general dependency management system for Catalyst, but to put the versions/hashes of _some_ dependencies in an easily accessible place for the build system. For instance, the llvm hash is needed when checking out its submodule, or labelling CI caches. Since the JAX version was tied to the LLVM version they are kept together (and there was a script that updated them all together). 

For general Python package dependencies, I think it's better to just follow the usual conventions, and we can have the dependency listed directly in the requirements section of setup.py. We also do this with diastatic-malt for example, which is needed for the autograph feature.

A somewhat related question: do we need to (or want to) hard-pin xdsl?

mudit2812: Thanks @dime10. @mlxd can chime in here, but the hard pin is there because xDSL, being a research project, evolves extremely rapidly with no promise of deprecation cycles before breaking changes. They have releases every week, and recently they've been pushing new features even in patch releases. Additionally, their processes for release and development are likely not as robust as some of our other dependencies.

dime10: Sounds reasonable then 👍

.github/workflows/check-pl-compat.yaml
mudit2812, line 1: To graphviz or not to graphviz?

PR #2284, opened 2025-12-10
Lowering qml.PauliRot to quantum dialect

lazypanda10117:
**Context:** Previously, we bind the paulirot operation directly to the QEC dialect's `qec.ppr` or `qec.ppr.arbitrary` operations. This has undesired side effects, such as incompatibility with the existing graph decomposition system. Now, instead of binding directly to the `qec.ppr` operation, we bind to a new `quantum.paulirot` operation in the Quantum dialect. Then, when a user needs to lower further to the QEC dialect, they can call the existing `to_ppr` pass to lower these `quantum.paulirot` operations into the corresponding `qec.ppr` or `qec.ppr.arbitrary` operations.

This two stage lowering approach helps unblock the graph decomposition issue, allowing us to use graph decomposition with the new decomposition rules introduced in https://github.com/PennyLaneAI/pennylane/pull/8700 and https://github.com/PennyLaneAI/pennylane/pull/8704.

**Benefits:** Fixes graph decomposition issues and enables new decomposition rules.

[[sc-105502]]

### DIFF
--- a/.github/workflows/build-wheel-linux-arm64.yaml
+++ b/.github/workflows/build-wheel-linux-arm64.yaml
@@ -474,9 +474,11 @@ jobs:
         python${{ matrix.python_version }} -m pip install 'amazon-braket-pennylane-plugin>=1.31.0'
 
     - name: Install OQC client
-      if: matrix.python_version != '3.13'
       run: |
-        python${{ matrix.python_version }} -m pip install oqc-qcaas-client
+        python_minor=$(echo "${{ matrix.python_version }}" | cut -d. -f2)
+        if [ "$python_minor" -lt 13 ]; then
+          python${{ matrix.python_version }} -m pip install oqc-qcaas-client
+        fi
 
     - name: Install Catalyst
       run: |
--- a/.github/workflows/build-wheel-linux-x86_64.yaml
+++ b/.github/workflows/build-wheel-linux-x86_64.yaml
@@ -493,9 +493,11 @@ jobs:
         python${{ matrix.python_version }} -m pip install 'amazon-braket-pennylane-plugin>=1.31.0'
 
     - name: Install OQC client
-      if: matrix.python_version != '3.13'
       run: |
-        python${{ matrix.python_version }} -m pip install oqc-qcaas-client
+        python_minor=$(echo "${{ matrix.python_version }}" | cut -d. -f2)
+        if [ "$python_minor" -lt 13 ]; then
+          python${{ matrix.python_version }} -m pip install oqc-qcaas-client
+        fi
 
     - name: Install Catalyst
       run: |
--- a/.github/workflows/build-wheel-macos-arm64.yaml
+++ b/.github/workflows/build-wheel-macos-arm64.yaml
@@ -166,9 +166,9 @@ jobs:
         lookup-only: True
 
     - name: Setup Python version
-      # There are multiple Python versions installed on the GitHub image, 3.11 - 3.12 is already
+      # There are multiple Python versions installed on the GitHub image, 3.11 - 3.14 is already
       # available under /Library/Frameworks/Python.framework/Versions/, but homebrew also provides
-      # 3.11 and 3.12. Make sure to consistently use the system versions.
+      # 3.11 - 3.14. Make sure to consistently use the system versions.
       run: |
         echo /Library/Frameworks/Python.framework/Versions/${{ matrix.python_version }}/bin >> $GITHUB_PATH
 
@@ -245,9 +245,9 @@ jobs:
         ref: ${{ inputs.branch || github.ref }}
 
     - name: Setup Python version
-      # There are multiple Python versions installed on the GitHub image, 3.11 - 3.12 is already
+      # There are multiple Python versions installed on the GitHub image, 3.11 - 3.14 is already
       # available under /Library/Frameworks/Python.framework/Versions/, but homebrew also provides
-      # 3.11 and 3.12. Make sure to consistently use the system versions.
+      # 3.11 - 3.14. Make sure to consistently use the system versions.
       run: |
         echo /Library/Frameworks/Python.framework/Versions/${{ matrix.python_version }}/bin >> $GITHUB_PATH
 
@@ -458,9 +458,9 @@ jobs:
         path: ${{ github.workspace }}/standalone_plugin_wheel/wheel
 
     - name: Setup Python version
-      # There are multiple Python versions installed on the GitHub image, 3.11 - 3.12 is already
+      # There are multiple Python versions installed on the GitHub image, 3.11 - 3.14 is already
       # available under /Library/Frameworks/Python.framework/Versions/, but homebrew also provides
-      # 3.11 and 3.12. Make sure to consistently use the system versions.
+      # 3.11 - 3.14. Make sure to consistently use the system versions.
       run: |
         echo /Library/Frameworks/Python.framework/Versions/${{ matrix.python_version }}/bin >> $GITHUB_PATH
 
@@ -474,9 +474,11 @@ jobs:
         python${{ matrix.python_version }} -m pip install 'amazon-braket-pennylane-plugin>1.27.1'
 
     - name: Install OQC client
-      if: matrix.python_version != '3.13'
       run: |
-        python${{ matrix.python_version }} -m pip install oqc-qcaas-client
+        python_minor=$(echo "${{ matrix.python_version }}" | cut -d. -f2)
+        if [ "$python_minor" -lt 13 ]; then
+          python${{ matrix.python_version }} -m pip install oqc-qcaas-client
+        fi
 
     - name: Install Catalyst
       run: |
--- a/.github/workflows/constants.yaml
+++ b/.github/workflows/constants.yaml
@@ -85,7 +85,7 @@ jobs:
       - name: Python test versions
         id: python_test_versions
         run: |
-          echo 'python_test_versions=["3.11", "3.12", "3.13"]' >> $GITHUB_OUTPUT
+          echo 'python_test_versions=["3.11", "3.12", "3.13", "3.14"]' >> $GITHUB_OUTPUT
 
       - name: Primary Python version
         id: primary_python_version
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -387,10 +387,13 @@
 
 <h3>Internal changes ⚙️</h3>
 
-* RTIO dialect is added to bypass the compilation flow from OpenAPL to ARTIQ’s LLVM IR. It is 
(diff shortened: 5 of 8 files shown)


mlir/lib/QEC/Transforms/ToPPR.cpp
```diff
@@ -233,6 +242,63 @@ LogicalResult convertMeasureOpToPPM(MeasureOp op, StringRef axis,
     return success();
 }
 
+LogicalResult convertPauliRotGate(PauliRotOp op, ConversionPatternRewriter &rewriter)
+{
+    auto loc = op.getLoc();
+
+    auto angleValue = op.getAngle();
+    auto pauliProduct = op.getPauliProduct();
```
kipawaa, line 242: I think `auto` is over-used here, and should be replaced with explicit types. Paul's comment on my PR was

> We use auto only when the type is immediately obvious from RHS. If not, we should be explicit about the types.

Logically this looks great!

frontend/catalyst/from_plxpr/decompose.py
```diff
@@ -233,16 +233,31 @@ def interpret_measurement(self, measurement: "qml.measurement.MeasurementProcess
                     # In this case, we fall back to using the COMPILER_OPS_FOR_DECOMPOSITION
                     # dictionary to get the number of wires.
                     num_wires, num_params = COMPILER_OPS_FOR_DECOMPOSITION[op.op.name]
+                    pauli_word = op.op.params.get("pauli_word", None)
+
+                    if op.op.name == "PauliRot":
+                        num_wires = len(pauli_word)
+                    elif num_wires == -1:
+                        num_wires = op_num_wires
```
kipawaa: How does this affect other gates with an arbitrary number of input wires (eg. `MultiRZ` or `GlobalPhase`)? Before we would have passed `num_wires = -1` to `_create_decomposition_rule`, but now it would be `None`, does this change anything? `op.name` PauliMeasure should also be considered here. Maybe checking `num_wires not None` can fix this?

frontend/catalyst/jax_primitives.py
```diff
@@ -439,9 +438,15 @@ def decomposition_rule(func=None, *, is_qreg=True, num_params=0):
...
 
+    if pauli_word is not None:
+        func = functools.partial(func, pauli_word=pauli_word)
+
     @functools.wraps(func)
     def wrapper(*args, **kwargs):
-        jaxpr = jax.make_jaxpr(func)(*args, **kwargs)
+        if pauli_word is not None:
+            jaxpr = jax.make_jaxpr(func)(theta=args[0], wires=args[1], **kwargs)
+        else:
```
kipawaa: Why does this need a special case for theta and wires? Something like multi-rz can already pass through here similarly without a special case.

1_Afterpipe.mlir
kipawaa, line 1: Was this mistakenly included?

mlir/lib/QEC/Utils/QECOpUtils.cpp
```diff
@@ -97,5 +101,91 @@ bool commutes(QECOpInterface rhsOp, QECOpInterface lhsOp)
...
+    if (auto addOp = dyn_cast<arith::AddFOp>(defOp)) {
+        double sum = 0.0;
+        for (auto operand : addOp.getOperands()) {
+            auto operandVal = resolveConstantValue(operand);
+            if (!operandVal.has_value())
+                return std::nullopt;
+            sum += operandVal.value();
+        }
+        return sum;
+    }
```
kipawaa: Does this need to be a for loop? `AddFOp` only has two operands (LHS and RHS), and resolving each recursively should be sufficient to calculate the sum. Not that this is incorrect, I just think that it would be simpler that way.

mlir/lib/QEC/Utils/QECOpUtils.cpp
```diff
@@ -97,5 +101,91 @@ bool commutes(QECOpInterface rhsOp, QECOpInterface lhsOp)
     return true;
 }
 
+// Recursively resolve the constant parameter of a value and returns std::nullopt if not a constant.
+std::optional<double> resolveConstantValue(Value value)
```
kipawaa, line 126: I think adding tests for the various input types handled here would be a good idea, since they all have nuance, and it will be simpler to catch any errors with resolving these values here than it will be downstream in passes.

mlir/lib/QEC/Transforms/ToPPR.cpp
dime10: I would agree. This is a matter of style/preference of course, but generally explicit types are much faster and easier to read imo (especially without an IDE!). The exceptions I would use are:
- if the type is already obvious from the right hand side (like a cast)
- if the type is unwieldy and would hurt readability rather than help it
- if auto is required for a language feature (I think structured binding is an example of this)

frontend/test/lit/test_pauli_rot_and_measure.py
```diff
@@ -29,6 +29,27 @@
 from catalyst import qjit
 
 
+def test_pauli_rot_lowering():
+    """Test that qml.PauliRot is lowered to quantum.paulirot."""
+    qml.capture.enable()
```
sengthai, line 36: can we use `@pytest.mark.usefixtures("use_capture")` here?

1_Afterpipe.mlir
lazypanda10117: oops, deleted.

mlir/lib/QEC/Utils/QECOpUtils.cpp
lazypanda10117: We can add proper unit tests for this after we separate this into a more general util function as we've discussed. We can open a new PR for this and put this in quantum utils.

frontend/catalyst/from_plxpr/decompose.py
lazypanda10117: Thanks River, good catch. I will fix this.

mlir/include/Quantum/IR/QuantumOps.td
```diff
@@ -564,6 +564,43 @@ def CustomOp : UnitaryGate_Op<"custom", [DifferentiableGate, NoMemoryEffect,
...
+def PauliWord : TypedArrayAttrBase<StrAttr, "A product of Pauli operators, aka a Pauli word.">;
+
+def PauliRotOp : UnitaryGate_Op<"paulirot", [DifferentiableGate, NoMemoryEffect,
+                                         AttrSizedOperandSegments, AttrSizedResultSegments]> {
+    let summary = "Apply a Pauli Product Rotation";
+    let description = [{
+        The `quantum.paulirot` operation applies a rotation around a Pauli product
+        operator to the state-vector. 
+        The arguments are the rotation angle `angle`, a string representing the
+        Pauli product operator, and a set of qubits the operation acts on.
```
sengthai, line 576: I think we should note in here that this op is not able to be executed in the moment and needs to be decomposed.

mlir/lib/QEC/Transforms/ToPPR.cpp
```diff
@@ -13,11 +13,20 @@
...
 
 #include "QEC/IR/QECOps.h"
 #include "QEC/Transforms/Patterns.h"
+#include "QEC/Utils/QECOpUtils.h"
 #include "Quantum/IR/QuantumOps.h"
+#include <cmath>
+#include <iostream>
+#include <llvm/Support/Debug.h>
+#include <llvm/Support/LogicalResult.h>
+#include <optional>
```
sengthai: i think we can remove these headers.

mlir/lib/QEC/Utils/QECOpUtils.cpp
lazypanda10117: It just cleaner to write it this way without having to explicitly check each op has value imo.

mlir/lib/QEC/Transforms/ToPPR.cpp
```diff
@@ -13,11 +13,20 @@
 // limitations under the License.
 
 #include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/IR/BuiltinAttributes.h"
 #include "mlir/IR/TypeRange.h"
```
sengthai: Could you check if we use these headers? 🤔

mlir/lib/QEC/Utils/QECOpUtils.cpp
```diff
@@ -97,5 +101,91 @@ bool commutes(QECOpInterface rhsOp, QECOpInterface lhsOp)
...
+        else if (auto denseAttr = dyn_cast<DenseFPElementsAttr>(valueAttr)) {
+            if (denseAttr.isSplat() || denseAttr.getNumElements() == 1) {
+                return denseAttr.getSplatValue<APFloat>().convertToDouble();
+            }
+        }
+        else if (auto denseAttr = dyn_cast<DenseIntElementsAttr>(valueAttr)) {
+            if (denseAttr.isSplat() || denseAttr.getNumElements() == 1) {
+                return static_cast<double>(denseAttr.getSplatValue<APInt>().getSExtValue());
+            }
+        }
```
sengthai: These conditions seem the same as those implemented in the Stablehlo case. How about making the lambda or normal functions for that? I think lambda is okay.

sengthai: This might resolve CodeFactor as well.

frontend/test/lit/test_pauli_rot_and_measure.py
lazypanda10117: Cannot, that is a fixture for pytest only. This is a lit test.

mlir/lib/QEC/Transforms/ToPPR.cpp
lazypanda10117: Oh right, i forgot to clean those up. Thanks.

mlir/lib/QEC/Utils/QECOpUtils.cpp
kipawaa: I feel that something like
```cpp
auto lhs = resolveConstantValue(addOp.getOperand(0));
auto rhs = resolveConstantValue(addOp.getOperand(1));
if (!lhs.has_value() || !rhs.has_value()) {
    return std::nullopt;
}
return lhs.value() + rhs.value();
```
might read more clearly, since we're examining an `add` op, but this isn't a big sticking point for me so I can mark this as resolved

mlir/lib/QEC/Utils/QECOpUtils.cpp
lazypanda10117: Updated.

mlir/lib/QEC/Utils/QECOpUtils.cpp
lazypanda10117: Fixed.

PR #2285, opened 2025-12-10
fix: use standard cluster label rather than "label" nodes

andrijapau:
As discussed with product, we've simplified cluster visualization by removing the information label nodes and relying on the standard cluster labels.

This improves both rendering performance and readability by preventing the illogical layout issues previously caused by treating labels as separate nodes.

[sc-105624]


doc/releases/changelog-dev.md
```diff
@@ -8,6 +8,7 @@
   [(#2214)](https://github.com/PennyLaneAI/catalyst/pull/2214)
   [(#2246)](https://github.com/PennyLaneAI/catalyst/pull/2246)
   [(#2231)](https://github.com/PennyLaneAI/catalyst/pull/2231)
+  [(#)](https://github.com/PennyLaneAI/catalyst/pull/)
```
andrijapau:
```suggestion
  [(#2285)](https://github.com/PennyLaneAI/catalyst/pull/2285)
```

PR #2286, opened 2025-12-10
Adapt MBQC test to follow new kwargs structure for captured programs

mlxd:
**Context:** Fixes cond measure test failures due to changes introduced by https://github.com/PennyLaneAI/pennylane/pull/8290

**Description of the Change:** Moves parameter data from calling conditional measure to inside the partial branches.

**Benefits:** Fixes [LLL](https://github.com/PennyLaneAI/catalyst/actions/runs/20095498237) failures.

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/frontend/test/lit/test_pauli_rot_and_measure.py
+++ b/frontend/test/lit/test_pauli_rot_and_measure.py
@@ -401,7 +401,7 @@ def test_pauli_rot_and_measure_with_cond():
     # CHECK: [[q0:%.+]] = qec.ppr ["Z"](4)
     # CHECK: qec.ppm ["Z"]
     # CHECK: scf.if
-    # CHECK: qec.ppr ["Z"](4)
+    # CHECK: qec.ppr.arbitrary ["Z"]
     # CHECK: scf.yield
     # CHECK: else
     # CHECK: scf.yield


.dep-versions
```diff
@@ -10,7 +10,7 @@ enzyme=v0.0.203
 
 # For a custom PL version, update the package version here and at
 # 'doc/requirements.txt'
```
JerryChen97, line 12: doc/requirements.txt also needs the same adjustment such that the doc CI runs correctly

mlxd: Forgot that, good catch.

JerryChen97:
Is `frontend/test/pytest/test_measurements_shots_results.py::TestOtherMeasurements::test_multiple_return_values[capture-lightning.kokkos]` the same thing to fix or might be some other stuff?

doc/releases/changelog-dev.md
```diff
@@ -290,6 +290,9 @@
 
 <h3>Bug fixes 🐛</h3>
 
+* Updated tests and PennyLane dependency pin to follow changes introduced by [PennyLaneAI/pennylane#8290](https://github.com/PennyLaneAI/pennylane/pull/8290).
+  [(#2286)](https://github.com/PennyLaneAI/catalyst/pull/2286)
```
paul0403, line 297: We don't do changelog entries for these "updating tests because PL changed something in the same development cycle". Note that changelog is a log for users to understand what changed wrt the previous release, not a day-to-day development log. (@dime10 )

mlxd: I disagree with this sentiment --- let's chat offline about the purpose of the CL :)

JerryChen97:
@mlxd @paul0403 What's our plan here? Note that LLL has been failing for days

lazypanda10117 (approved):
Looks good. Thanks.

.dep-versions
```diff
@@ -10,7 +10,7 @@ enzyme=v0.0.203
...
-pennylane=0.44.0-dev50
+pennylane=0.44.0-dev51
```
lazypanda10117: Need to update this to the latest PL version.

frontend/test/lit/test_pauli_rot_and_measure.py
```diff
@@ -392,9 +394,9 @@ def test_pauli_rot_and_measure_with_cond():
     @qjit(pipelines=pipeline, target="mlir")
     @qml.qnode(device=dev)
     def circuit():
-        qml.PauliRot(np.pi / 2, "Z", wires=0)
+        partial(qml.PauliRot, pauli_word="Z")(np.pi / 2, wires=0)
```
lazypanda10117: It actually doesn't affect normal usage of PauliRot. Only when it is used in conjunction with `qml.cond`.

JerryChen97: I'll revert this

.dep-versions
JerryChen97: I'll bump PL first and then update here

PR #2289, opened 2025-12-11
Fix xDSL transform interpreter to add options support for MLIR passes

mudit2812:
**Context:**
With the unified compiler pipeline, MLIR passes were being scheduled, but without any pass options. In the process of fixing this bug, I realized that pass options weren't being passed correctly even for xDSL passes. Other low hanging fruit related to the changes in this PR was also addressed.

* Added a `get_pyval_from_xdsl_attr` utility function that converts an xDSL attribute into its corresponding Python value
* Allow transmitting pass options to MLIR passes correctly.
  * Add support for ints, floats, bools, strings, dictionaries and arrays as pass options
  * Add support for arrays/dicts of the above as options.
* Update `QuantumParser` to also load the LLVM dialect. This was needed because I realized that Catalyst uses the `llvm.linkage` attribute as part of the qnode function, which means that we can never parse MLIR produced by Catalyst unless `allow_unregistered=True`.
* `compiler_transform` now also sets the `pass_name` attribute of the underlying `TransformDispatcher`, allowing using xDSL passes as decorators even with program capture disabled.

**Benefits:**
* MLIR passes with options work correctly with the unified compiler.
* Catalyst generated MLIR strings can be more generally parsed into xDSL modules without using `allow_unregistered=True`.
* xDSL transform decorators can be used with capture disabled.

[sc-105740]


frontend/catalyst/python_interface/pass_api/transform_interpreter.py
```diff
@@ -143,3 +145,36 @@ def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
...
+            case builtin.FloatAttr:
+                val = val.value.data
+            case builtin.StringAttr:
+                val = val.data
+            case _:
+                raise ValueError(f"Unsupported pass option {opt}: {val}.")
+
+        options.append(f"{opt}={val}")
+
+    cli_pass = f"builtin.module({pass_name}{{{" ".join(options)}}})"
```
josephleekl: I think i think this line has some issues? the `"` pairing are messed up?

mudit2812: Fixed in [0e614fc](https://github.com/PennyLaneAI/catalyst/pull/2289/commits/0e614fc59b77914dcabc4db575161c907de1c9ec)

frontend/catalyst/python_interface/pass_api/transform_interpreter.py
```diff
@@ -99,10 +102,9 @@ def run_apply_registered_pass_op(
         # ---- Catalyst path ----
         buffer = io.StringIO()
         Printer(stream=buffer, print_generic_format=True).print_op(module)
-
-        schedule = f"--{pass_name}"
+        schedule = _create_schedule([op])
```
mudit2812, line 106: The reason why we wrap the pass op in a list is so that in the future, we can update the interpreter to apply a sequence of consecutive MLIR passes rather than having to do a roundtrip dump-parse-dump-parse for each MLIR pass.

frontend/catalyst/python_interface/pass_api/transform_interpreter.py
```diff
@@ -143,3 +145,76 @@ def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
...
+        if not pass_options:
+            schedule.append(f"--{pass_name}")
+            continue
+
+        cli_options = []
+        for opt, val in pass_options.items():
+            cli_options.append(f"{opt}={_get_cli_option_from_attr(val)}")
+        cli_options = " ".join(cli_options)
+
+        cli_pass = f"--{pass_name}={cli_options}"
```
sengthai, line 182: I'd recommend using two backticks for the pass option `--test-pass="list-opt={a=1 b=2}"`to improve clarity.

frontend/test/pytest/python_interface/pass_api/test_transform_interpreter.py
```diff
@@ -0,0 +1,362 @@
...
+    """Get an xDSL attribute corresponding to a Python value."""
+    attr = None
+
+    match val:
+        case bool():
+            attr = builtin.IntegerAttr(int(val), 1)
+        case int():
+            attr = builtin.IntegerAttr(val, 64)
+        case float():
+            attr = builtin.FloatAttr(val, 64)
```
sengthai, line 52: I wonder if the operands or attributes in MLIR ops are `i8` or `I8Attr`; does xDSL handle type conversion?

frontend/catalyst/python_interface/pass_api/transform_interpreter.py
mudit2812: I was doing that initially, but because of how `subprocess.run` parses the command, an error gets raised when invoking `quantum-opt`. Without your suggestion, `subprocess.run` interprets the pass as the following, and that works correctly with `subprocess.run`:
```bash
quantum-opt "--test-pass=list-option={a=1 b=2}"
```

I agree that what you're suggesting is more readable, but it doesn't work with `subprocess.run` sadly :(

frontend/test/pytest/python_interface/pass_api/test_transform_interpreter.py
mudit2812: I'm not sure I understand what you're asking. But regardless, this function is just a utility for testing, so it probably doesn't matter right now.

sengthai: ohh it is okay now.

frontend/catalyst/python_interface/pass_api/transform_interpreter.py
sengthai: oh i see, how about `quantum-opt "--test-pass="list-option={a=1 b=2}""`? It seems to work in the terminal, but it looks a little messy. :3

frontend/test/pytest/python_interface/pass_api/test_transform_interpreter.py
```diff
@@ -0,0 +1,362 @@
...
+    def test_pass_basic_options(self):
+        """Test that passes with basic options (int, float, bool, string) are parsed correctly."""
+        pass_op = create_apply_registered_pass_op(
+            "test-pass",
+            options={"int-opt": 1, "float-opt": 1.5, "bool-opt": False, "str-opt": "test_string"},
+        )
+        schedule = _create_schedule(pass_ops=[pass_op])
+        assert len(schedule) == 1
+        assert (
+            schedule[0] == "--test-pass=int-opt=1 float-opt=1.5 bool-opt=false str-opt=test_string"
```
andrijapau: I thought it would need to be `--test-pass="int-opt=1 float-opt=1.5 bool-opt=false str-opt=test_string"` to prevent it from being incorrectly split?

frontend/test/pytest/python_interface/pass_api/test_transform_interpreter.py
```diff
@@ -0,0 +1,362 @@
...
+        pass_op = create_apply_registered_pass_op("test-pass")
+        schedule = _create_schedule(pass_ops=[pass_op])
+        assert len(schedule) == 1
+        assert schedule[0] == "--test-pass"
+
+    def test_pass_basic_options(self):
+        """Test that passes with basic options (int, float, bool, string) are parsed correctly."""
+        pass_op = create_apply_registered_pass_op(
+            "test-pass",
+            options={"int-opt": 1, "float-opt": 1.5, "bool-opt": False, "str-opt": "test_string"},
```
andrijapau: What if the string had a space in it?

frontend/test/pytest/python_interface/pass_api/test_transform_interpreter.py
mudit2812: That's what I tried initially but that failed. "--test-pass=arg1=val1 arg2=val2" works correctly as well. When `subprocess.run` sees spaces, it automatically adds double quotes around the whole expression.

frontend/test/pytest/python_interface/pass_api/test_transform_interpreter.py
mudit2812: I'll test this and tag you again. Thanks for thinking of this :)

frontend/test/pytest/python_interface/pass_api/test_transform_interpreter.py
andrijapau: Ohh cool. Resolving.

frontend/test/pytest/python_interface/pass_api/test_transform_interpreter.py
andrijapau: This is where the hypothesis tool would come in handy

mudit2812: Fixed [8caa90f](https://github.com/PennyLaneAI/catalyst/pull/2289/commits/8caa90fcb5ebeaa8001f91561d096369d8acadf9)

PR #2290, opened 2025-12-11
feat: add support for dynamic wire connectivity

andrijapau:
This PR accomplishes a few things,

- Graph now uses polyline edges by default.
- Adds static / dynamic connectivity logic in a `_connect` helper method.
- Connects the device node to the rest of the graph.
- Updates some utility functions in xdsl_conversion to allow for processing of dynamic wires in operators (using the argX from the IR as wire labels). You can also use pythonic operators in wires or params like `H(i % num_qubits)`

Here's an example, 
```python
@qml.qjit(autograph=True, target="mlir")
@qml.qnode(dev)
def my_workflow(x, y):
    qml.X(0)
    qml.Y(1)
    qml.Z(2)
    qml.H(x)
    qml.S(0)
    qml.T(2)
    qml.H(y)
```
Screenshot 2025-12-22 at 11 43 26 AM

[sc-103466]

### DIFF
--- a/frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
+++ b/frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
@@ -514,9 +514,8 @@ class ConstructCircuitDAG:
 
             # If we just exited a conditional (and are not in a nested one currently)
             # We need to connect to everything seen so far as all branches are a possibility.
-            exited_conditional_cluster = (
-                "conditional" in self._last_cluster_uid
-                and not any("conditional" is s for s in self._cluster_uid_stack)
+            exited_conditional_cluster = "conditional" in self._last_cluster_uid and not any(
+                "conditional" in s for s in self._cluster_uid_stack
             )
             if exited_conditional_cluster:
                 return all_active


frontend/catalyst/python_interface/inspection/xdsl_conversion.py
```diff
@@ -208,6 +218,8 @@ def count_static_loop_iterations(for_op: ForOp) -> int:
     lower_bound = resolve_constant_params(for_op.lb)
     upper_bound = resolve_constant_params(for_op.ub)
     step = resolve_constant_params(for_op.step)
+    if not all(isinstance(x, int) for x in [lower_bound, upper_bound,step]):
+        raise NotImplementedError
```
andrijapau: Broke `pytest frontend/test/pytest/python_interface/inspection/test_mlir_specs.py::TestMLIRSpecs::test_dynamic_for_loop` because now dynamic loop variables are caught as strings.

andrijapau: @jzaia18 this is the line of code we talked about.

frontend/catalyst/python_interface/inspection/xdsl_conversion.py
```diff
@@ -302,41 +339,44 @@ def xdsl_to_qml_op(op) -> Operator:
...
-            )
-
-        case "quantum.custom":
-            gate_cls = resolve_gate(op.properties.get("gate_name").data)
-            gate = gate_cls(*ssa_to_qml_params(op), wires=ssa_to_qml_wires(op))
-
-        case _:
-            raise NotImplementedError(f"Unsupported gate: {op.name}")
+    # Pause capture *only if active* so we can allow strings (dynamic wires) as allowed wires
+    with conditional_pause(capture.pause):
```
andrijapau, line 379: Horrible diff but all I did was put all of these in the pause context manager.

frontend/catalyst/python_interface/inspection/xdsl_conversion.py
```diff
@@ -158,17 +159,26 @@ def _apply_adjoint_and_ctrls(qml_op: Operator, xdsl_op) -> Operator:
 
 
 # pylint: disable=too-many-return-statements
-def resolve_constant_params(ssa: SSAValue) -> float | int:
+def resolve_constant_params(ssa: SSAValue) -> float | int | str:
```
andrijapau, line 175: This should probably be renamed to `resolve_params`

frontend/catalyst/python_interface/inspection/collector.py
```diff
@@ -71,6 +71,13 @@ def _(self, xdsl_meas: StateOp) -> MeasurementProcess:
     @handle.register
     def _(self, xdsl_meas_op: ExpvalOp | VarianceOp | ProbsOp | SampleOp) -> MeasurementProcess:
         obs_op = xdsl_meas_op.obs.owner
+
+        if isinstance(xdsl_meas_op, (ProbsOp, SampleOp)):
+            # TODO: This doesn't logically make sense, but quantum.compbasis
+            # is obs_op and function below just pulls out the static wires
+            wires = xdsl_to_qml_measurement(obs_op)
+            return xdsl_to_qml_measurement(xdsl_meas_op, wires=None if wires == [] else wires)
```
andrijapau, line 79: After modifying `xdsl_conversion.py` this also needed to be updated as it broke some tests.

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -199,6 +185,9 @@ def _measurements(
                 meas = xdsl_to_qml_measurement(op, xdsl_to_qml_measurement(obs_op))
                 prev_wires = meas.wires.labels
 
+                if any(isinstance(w, str) for w in meas.obs.wires):
+                    prev_wires = self._wire_to_node_uids.keys()
```
andrijapau: Special handling as the measurement's wires here are the observable's wires.

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -363,9 +358,10 @@ def _device_init(self, operation: quantum.DeviceInitOp) -> None:
             fillcolor="grey",
             color="black",
             penwidth=2,
-            shape="rectangle",
         )
         self._node_uid_counter += 1
+        self._dynamic_node_uids.add(node_id)
+        self._wire_to_node_uids["device"].add(node_id)
```
andrijapau: Allows us to connect the device node as the "source" of all edges

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -455,7 +531,7 @@ def _operator(op: Operator) -> str:
     else:
         wires_str = f"[{', '.join(map(str, wires))}]"
     # Using <...> lets us use ports (https://graphviz.org/doc/info/shapes.html#record)
-    return f"<name> {op.name}|<wire> {wires_str}"
+    return f"<name> {str(op).split('(')[0]}|<wire> {wires_str}"
```
andrijapau: Bit of a "hacky" way to extract the simplified name from an operator instance. Any better ideas?

frontend/catalyst/python_interface/visualization/pydot_dag_builder.py
```diff
@@ -171,11 +175,11 @@ def add_edge(self, from_uid: str, to_uid: str, **attrs: Any) -> None:
             ValueError: Destination is not found in the graph.
 
         """
-        if from_uid == to_uid:
+        if from_uid.split(":")[0] == to_uid.split(":")[0]:
```
andrijapau, line 178: Allow [ports](https://graphviz.org/docs/attr-types/portPos/) to be used

frontend/catalyst/python_interface/inspection/xdsl_conversion.py
```diff
@@ -58,6 +60,18 @@
...
+def conditional_pause(pause):
+    """Will only pause capture if it is enabled."""
+    if capture.enabled():
+        return pause()
+
+    @contextlib.contextmanager
+    def dont_do_anything():
+        yield
+
+    return dont_do_anything()
```
andrijapau, line 72: I would personally use this work around until the feature is finished and then worry about updating capture.pause to have the same behaviour. Thoughts?

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -363,9 +358,10 @@ def _device_init(self, operation: quantum.DeviceInitOp) -> None:
             fillcolor="grey",
             color="black",
             penwidth=2,
-            shape="rectangle",
         )
         self._node_uid_counter += 1
+        self._dynamic_node_uids.add(node_id)
```
andrijapau:
```suggestion
        # NOTE: Consider the device node as dynamic
        # as any wire can source from it
        self._dynamic_node_uids.add(node_id)
```

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -405,6 +403,86 @@ def _func_return(self, operation: func.ReturnOp) -> None:
...
+            all_prev_uids = set().union(*self._wire_to_node_uids.values())
+
+            # Filter out all dynamic uids first
+            only_dynamic_uids = all_prev_uids.issubset(self._dynamic_node_uids)
+            static_uids = set()
+            if not only_dynamic_uids:
+                static_uids = all_prev_uids - self._dynamic_node_uids
+
+            # If static nodes exist, connect only to them to avoid
+            # direct dynamic-to-dynamic links
```
andrijapau:
```suggestion
            # unnecessary dynamic-to-dynamic connections
```

frontend/test/pytest/python_interface/visualization/test_construct_circuit_dag.py
```diff
@@ -1533,4 +2092,4 @@ def my_workflow():
         assert clusters["cluster2"]["label"] == "adjoint"
         assert clusters["cluster2"]["parent_cluster_uid"] == "cluster1"
         assert "Hadamard" in nodes["node1"]["label"]
-        assert nodes["node1"]["parent_cluster_uid"] == "cluster2"
+        assert nodes["node1"]["parent_cluster_uid"] == "cluster2"
```
```suggestion
        assert nodes["node1"]["parent_cluster_uid"] == "cluster2"

```

frontend/catalyst/python_interface/inspection/xdsl_conversion.py
```diff
@@ -158,17 +172,33 @@ def _apply_adjoint_and_ctrls(qml_op: Operator, xdsl_op) -> Operator:
...
+                x = resolve_constant_params(op.operands[0])
+                y = resolve_constant_params(op.operands[1])
+                return f"({x} % {y})"
+            raise NotImplementedError(f"Function call to {op.callee} not supported")
+
+        case "tensor.from_elements":
+            return resolve_constant_params(op.operands[0])
+
+        case "arith.index_cast":
+            return resolve_constant_params(op.operands[0])
```
sengthai, line 201: I would suggest using this. 
```suggestion
        case "tensor.from_elements" | "arith.index_cast":
            return resolve_constant_params(op.operands[0])
```

frontend/catalyst/python_interface/inspection/xdsl_conversion.py
```diff
@@ -394,34 +470,38 @@ def xdsl_to_qml_measurement(op, *args, **kwargs) -> MeasurementProcess | Operato
...
-            return _extract(op, "qubits", resolve_constant_wire)
-
-        case (
-            "quantum.state" | "quantum.probs" | "quantum.sample" | "quantum.expval" | "quantum.var"
-        ):
-            return resolve_measurement(op.name)(*args, **kwargs)
-
-        case _:
-            raise NotImplementedError(f"Unsupported measurement/observable: {op.name}")
+    with conditional_pause(capture.pause):
```
sengthai, line 474: Just curious, what and how conditional pause works? 🤔

frontend/catalyst/python_interface/inspection/xdsl_conversion.py
```diff
@@ -223,14 +269,41 @@ def dispatch_wires_extract(op: ExtractOp):
...
     if isinstance(ssa, IntegerAttr):  # Catalyst
         return ssa.value.data
 
     op = ssa.owner
 
+    if isinstance(op, Block):
+        arg_name = list(compress(op.args, map(lambda arg: arg is ssa, op.args)))[0]
+        return arg_name.name_hint
+
     match op:
```
sengthai, line 283: I think using `Guarded Patterns` (`case _ if`) can be slower than `Structural Patterns`. I would recommend separating this `match` into two. One handles the case with `op.name`, and another handles the object type. 

Also, we can group the cases together: `stablehlo.add`, `tensor.from_elements`, and `stablehlo.convert`

frontend/catalyst/python_interface/inspection/xdsl_conversion.py
andrijapau: It's defined at the top of the file. Basically, it will only pause capture if it is enabled in the first place. :)

frontend/catalyst/python_interface/inspection/xdsl_conversion.py
andrijapau: Thanks!

I think those are both great ideas, but I believe it is beyond the scope of the PR. I can create a story to track this improvement though and we can get to it in Q1 (lots of focus on improving the internals of the python_interface module). What do you think?

frontend/catalyst/python_interface/inspection/xdsl_conversion.py
andrijapau: To keep the conversation in an isolated place, I will respond [here](https://github.com/PennyLaneAI/catalyst/pull/2290#discussion_r2632478576).

frontend/catalyst/python_interface/inspection/xdsl_conversion.py
sengthai: Creating a story to track this sounds good to me!

sengthai: Feel free to resolve this when the story is created.

andrijapau: Done [sc-106427], resolving!

sengthai (approved):
LGTM~ 👍

frontend/catalyst/python_interface/inspection/collector.py
mudit2812: Future work but maybe we can update `xdsl_to_qml_measurement` into `xdsl_to_qml_measurement` and `xdsl_to_qml_obs_or_wires`?

frontend/catalyst/python_interface/inspection/xdsl_conversion.py
mudit2812: Why make `pause` an argument?

frontend/catalyst/python_interface/inspection/xdsl_conversion.py
```diff
@@ -158,17 +172,33 @@ def _apply_adjoint_and_ctrls(qml_op: Operator, xdsl_op) -> Operator:
 
 
 # pylint: disable=too-many-return-statements
-def resolve_constant_params(ssa: SSAValue) -> float | int:
+def resolve_constant_params(ssa: SSAValue) -> float | int | str:
     """Resolve a constant parameter SSA value to a Python float or int."""
     op = ssa.owner
 
+    if isinstance(op, Block):
+        arg_name = list(compress(op.args, map(lambda arg: arg is ssa, op.args)))[0]
```
mudit2812: This should work, right?
```suggestion
        arg_name = next(compress(op.args, map(lambda arg: arg is ssa, op.args)))
```

frontend/catalyst/python_interface/inspection/xdsl_conversion.py
```diff
@@ -223,14 +269,41 @@ def dispatch_wires_extract(op: ExtractOp):
...
-def resolve_constant_wire(ssa: SSAValue) -> float | int:
+def resolve_constant_wire(ssa: SSAValue) -> float | int | str:
     """Resolve the wire for the given SSA qubit."""
     if isinstance(ssa, IntegerAttr):  # Catalyst
         return ssa.value.data
 
     op = ssa.owner
 
+    if isinstance(op, Block):
+        arg_name = list(compress(op.args, map(lambda arg: arg is ssa, op.args)))[0]
```
```suggestion
        arg_name = next(compress(op.args, map(lambda arg: arg is ssa, op.args)))
```

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -128,23 +132,14 @@ def _gate_op(
...
-            self._wire_to_node_uids[wire] = {node_uid}
+        self._connect(qml_op, node_uid)
 
     @_visit_operation.register
     def _projective_measure_op(self, op: quantum.MeasureOp) -> None:
         """Handler for the single-qubit projective measurement operation."""
 
         # Create PennyLane instance
-        meas = xdsl_to_qml_measurement(op)
+        meas: MeasurementProcess = xdsl_to_qml_measurement(op)
```
mudit2812: Won't this be a MidMeasure (operator, not mp)?

mudit2812: Future task: converting xDSL `MeasureOp` to PL `MidMeasure` should be done by `xdsl_to_qml_op`, not `xdsl_to_qml_measurement`

andrijapau: > Won't this be a MidMeasure (operator, not mp)?

Good catch! https://github.com/PennyLaneAI/catalyst/pull/2290/commits/1deb9872a1d8cb57eb48247b307754ee3de2baa6

andrijapau: > Future task: converting xDSL MeasureOp to PL MidMeasure should be done by xdsl_to_qml_op, not xdsl_to_qml_measurement

https://app.shortcut.com/xanaduai/story/106497/converting-xdsl-measureop-to-pl-midmeasure-should-be-done-by-xdsl-to-qml-op-not-xdsl-to-qml-measurement

frontend/catalyst/python_interface/inspection/xdsl_conversion.py
```diff
@@ -223,14 +269,41 @@ def dispatch_wires_extract(op: ExtractOp):
...
+    if isinstance(op, Block):
+        arg_name = next(compress(op.args, map(lambda arg: arg is ssa, op.args)))
+        return arg_name.name_hint
+
     match op:
+        case _ if op.name == "func.call":
+            if op.callee.string_value() == "remainder":
+                x = resolve_constant_params(op.operands[0])
+                y = resolve_constant_params(op.operands[1])
+                return f"({x} % {y})"
```
mudit2812, line 288: What's this about?

frontend/catalyst/python_interface/inspection/xdsl_conversion.py
```diff
@@ -223,14 +269,41 @@ def dispatch_wires_extract(op: ExtractOp):
...
+        case _ if op.name == "stablehlo.reshape":
+            return resolve_constant_wire(op.operands[0])
+
+        case _ if op.name == "stablehlo.add":
+            x, y = (resolve_constant_wire(op.operands[0]), resolve_constant_wire(op.operands[1]))
+            return f"({x} + {y})"
+
+        case _ if op.name == "stablehlo.subtract":
+            x, y = (resolve_constant_wire(op.operands[0]), resolve_constant_wire(op.operands[1]))
+            return f"({x} - {y})"
```
mudit2812, line 300: This is a way to have symbolic representation of wires that have been math'd?

mudit2812: I'm concerned that we're making one function to do many different things and would prefer if there was a better separation of responsibilities.

frontend/catalyst/python_interface/inspection/xdsl_conversion.py
andrijapau: So this allows operators like `qml.H(x % 3)` to render properly. It was inspired by Lee's [circuit](https://gist.github.com/mlxd/9cc97102073dd7e815d836ef6c419c61).

frontend/test/pytest/python_interface/visualization/test_construct_circuit_dag.py
```diff
@@ -842,6 +842,213 @@ def my_circuit():
...
+
+        # Construct DAG
+        utility = ConstructCircuitDAG(FakeDAGBuilder())
+        utility.construct(module)
+
+        nodes = utility.dag_builder.nodes
+        assert len(nodes) == 3  # Device node + H + X
+
+        assert nodes["node1"]["label"] == f"<name> Hadamard|<wire> [arg0]"
+        assert nodes["node2"]["label"] == f"<name> PauliX|<wire> [arg1]"
```
mudit2812, line 996: Shouldn't the edges in the DAG be tested? We can't assume that `x` will always be different to `y`, so there will have to be an edge connecting the two gates.

frontend/catalyst/python_interface/inspection/xdsl_conversion.py
andrijapau: > This is a way to have symbolic representation of wires that have been math'd?

Yes, exactly. So operators like `qml.H(x + y)` will render correctly.

andrijapau: > I'm concerned that we're making one function to do many different things and would prefer if there was a better separation of responsibilities.

I agree. Although, I don't know how much can be done before EOY. 😬

frontend/test/pytest/python_interface/visualization/test_construct_circuit_dag.py
mudit2812: I just saw you have tests for the edges in a different class. You can ignore this comment.

andrijapau: I guess adding a test for edges wouldn't hurt but it's in the `TestCreateDynamicOperatorNodes` class which deliberately tests the creation of nodes, not connectivity. Happy to add it though if you want.

frontend/catalyst/python_interface/inspection/xdsl_conversion.py
mudit2812: I would consider this to be a change within the scope of this PR. The change was made to accommodate dynamic wires. I would really prefer to avoid technical debt where possible. But, if you find this to take too long, then we can revisit later.

mudit2812:
> Sounds good! Would be great if you could provide an example that would meet your concerns

Very simple pseudocode but something along the lines of:
```python
def circuit(x):
    <gates>
    <control flow>
    <gates>
    <measurements>
```

frontend/test/pytest/python_interface/visualization/test_construct_circuit_dag.py
andrijapau: > I just saw you have tests for the edges in a different class. You can ignore this comment.

:) resolving

mudit2812 (commented):
Looks great. I just realized there is one case regarding control flow that is not covered at all, which is the case where there are no quantum operations inside control flow. For loops, it seems to make sense that we won't need to render them at all, but for conditionals, what if some branches have gates but others don't? How will that render?

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -219,12 +214,29 @@ def _measurements(
...
-            if (dyn_node_uids := self._wire_to_node_uids["dyn_wire"]) < all_prev_uids:
-                all_prev_uids -= dyn_node_uids
+            # wires = [] means connect to all active wires
+            if "device" in self._wire_to_node_uids and len(self._wire_to_node_uids) == 1:
+                # Case 1: Only the device node in past history
+                all_prev_uids = self._wire_to_node_uids["device"]
+            else:
+                # Case 2: Wire map is "device" + other stuff
+                device_node_uid: set[str] = self._wire_to_node_uids.get("device", set())
+                all_active = set().union(*self._wire_to_node_uids.values()) - device_node_uid
```
mudit2812, line 225: ?
```suggestion
                all_active = set(self._wire_to_node_uids.values()) - device_node_uid
```

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -219,12 +214,29 @@ def _measurements(
...
+
+                # If we just exited a conditional (and are not in a nested one currently)
+                # We need to connect to everything seen so far as all branches are a possibility.
+                exited_conditional_cluster = (
+                    "conditional" in self._last_cluster_uid
```
mudit2812: Nice to have for the future but maybe it makes sense to have a enum for the different cluster/node types

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -219,12 +214,29 @@ def _measurements(
...
+                    and sum(1 for s in self._cluster_uid_stack if "conditional" in s) == 0
```
mudit2812:
```suggestion
                    and not any("conditional" in s for s in self._cluster_uid_stack)
```

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -219,12 +214,29 @@ def _measurements(
...
+                )
+                if exited_conditional_cluster:
+                    all_prev_uids = all_active
+                else:
+                    # Otherwise, just connect to static nodes as they block dynamic
+                    # node connections
+                    static_nodes = all_active - self._dynamic_node_uids
+                    all_prev_uids = static_nodes if static_nodes else all_active
```
mudit2812:
```suggestion
                    all_prev_uids = static_nodes or all_active
```

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
andrijapau: That's a good idea! [sc-106741]

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -310,7 +319,11 @@ def _if_op(self, operation: scf.IfOp):
         self._cluster_uid_counter += 1
 
         # Save wires state before all of the branches
-        wire_map_before = self._wire_to_node_uids.copy()
+        wire_map_before = deepcopy(self._wire_to_node_uids)
+
+        # If we have a dynamic wire pop the device
+        # if "dyn_wire" in wire_map_before:
+        #    wire_map_before.pop("device", set())
```
andrijapau:
```suggestion
```

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -307,6 +321,10 @@ def _if_op(self, operation: scf.IfOp):
 
         # Save wires state before all of the branches
         wire_map_before = deepcopy(self._wire_to_node_uids)
+
+        # If we have a dynamic wire pop the device
+        # if "dyn_wire" in wire_map_before:
+        #    wire_map_before.pop("device", set())
```
mudit2812:
```suggestion
```

andrijapau: Yeah sorry, I just removed that :)

frontend/catalyst/python_interface/visualization/construct_circuit_dag.py
```diff
@@ -445,69 +483,131 @@ def _connect(self, node: Operator | MeasurementProcess, node_uid: str, **edge_at
...
+                if not self._wire_to_node_uids:
+                    return device_node_uid
+
             all_active = set().union(*self._wire_to_node_uids.values())
+
+            # If we just exited a conditional (and are not in a nested one currently)
+            # We need to connect to everything seen so far as all branches are a possibility.
+            exited_conditional_cluster = (
+                "conditional" in self._last_cluster_uid
+                and sum(1 for s in self._cluster_uid_stack if "conditional" in s) == 0
```
mudit2812:
```suggestion
                and not any("conditional" is s for s in self._cluster_uid_stack)
```

mudit2812 (approved):

PR #1455, opened 2025-01-13
Fix wheel upload issues

dime10:
This PR fixes two issues preventing wheel upload to PyPI:
- removes the deprecated `provides` kwarg in setup.py
- reduces the file size of the wheel due to duplicate packaging of the catalyst cli

Here is the successful build & upload of all wheels based on this branch:
https://github.com/PennyLaneAI/catalyst/actions/runs/12754030162

PR #2291, opened 2025-12-11
Small fixes around path detections

dime10:
Includes:
- Fix cleanup of wheel artifacts
  - now includes an `include` directory
  - `bin` directory was moved a while ago one level up
- Fix misleading `package_root` variable
  - now it points to the actual package root, not the `utils` directory
  - since we're using `os.join`, we should use it for all path separators
- Remove obsolete `get_bin_dir` function
  - only used in a test, not even to get the `bin` dir but the `lib` dir
  - add catalyst dialects build dir to the default lib paths
- Add cli path fallback for rare wheel setups
 
  The catalyst binary used to be shipped right within the package, under
catalyst/bin. This is [no longer the case](https://github.com/PennyLaneAI/catalyst/pull/1455), since pip will install the
binary into the dedicated binary location for Python packages.
However, we may still encounter a configuration where the binary is
located in a bin folder next to the catalyst package folder,
specifically:
  - after `make wheel` we'll have frontend/bin next to frontend/catalyst,
  although we rarely directly test this configuration since it's more
  of a setup step to packaging the wheel.
  - when trying to emulate the above configuration for a quick
  frontend-only editable install, e.g. by extracting wheel contents
  over a checked-out catalyst repo.


kipawaa:
Is there a reason to prefer `os.path.join` over `Pathlib.Path` here? Many path operations are simplified by pathlib, and it is generally considered the preferred module for path handling unless os is specifically needed.

dime10:
> Is there a reason to prefer `os.path.join` over `Pathlib.Path` here? Many path operations are simplified by pathlib, and it is generally considered the preferred module for path handling unless os is specifically needed.

It's just what the majority of the file was already using, I don't think there is a big difference in this case.

EDIT: sorry, didn't realize you were assuming it's a Path object in a bunch of places downstream

paul0403 (commented):
Nice! Remember to update `precompile_decomposition_rules.py` to `os.path` from `pathlib` and then remove the unused package!

frontend/catalyst/utils/precompile_decomposition_rules.py
```diff
@@ -214,7 +214,7 @@ def precompile_decomp_rules(decomp_file_path: Path = BYTECODE_FILE_PATH):
     Args:
         decomp_file_path (Path): path to compile rules to.
     """
-    decomp_file_path.parent.mkdir(parents=True, exist_ok=True)
+    Path(decomp_file_path).parent.mkdir(parents=True, exist_ok=True)
```
dime10, line 217: @kipawaa The last commit aligns on the string type for the global, while leaving your code to use Path, although if you prefer one could probably make the case to make the global a Path (even though the other globals aren't). 
(If we do revert this, since it's a global, and you're using it as a default value in signatures, I did want to make sure that `Path` objects are actually immutable, but I think this is the case.)

frontend/test/pytest/test_precompile_decomp_rules.py
```diff
@@ -127,14 +129,14 @@ def fake_op_decomp(string):
 
 def test_bytecode_file():
     """Test that the bytecode file is generated correctly."""
-    BYTECODE_FILE_PATH.unlink(missing_ok=True)
+    Path(BYTECODE_FILE_PATH).unlink(missing_ok=True)
 
     precompile_decomp_rules()
 
-    assert BYTECODE_FILE_PATH.exists()
+    assert Path(BYTECODE_FILE_PATH).exists()
```
dime10: While updating the code I did find this test code, which modifies the user installation right? As in, it deletes a file from disk to then test its creation. Technically, if the test failed, the global state wouldn't be restored to what it is was before the test, which is a rule we normally try to follow using fixtures and finalizers, but in this case I don't see how you could restore the prior state since we are testing the thing that would restore it? 😅 
My only idea is to move the file instead of delete it, which is easier to restore afterwards if something goes wrong?

frontend/catalyst/utils/precompile_decomposition_rules.py
paul0403: https://docs.python.org/3/library/pathlib.html#general-properties

> Paths are immutable and [hashable](https://docs.python.org/3/glossary.html#term-hashable).

kipawaa: I'm generally in favour of `Path`s, but consistency is valuable and this looks good to me :)

frontend/test/pytest/test_precompile_decomp_rules.py
kipawaa: I considered this when writing the test as well, but didn't consider moving the file! I think that's a great idea.

dime10: Hopefully this does the trick, should be pretty thorough: https://github.com/PennyLaneAI/catalyst/pull/2291/commits/c4b8665412a3ae3590488ca40f1d81b657a9ff49

frontend/test/lit/test_mlir_plugin.py
```diff
@@ -90,8 +90,7 @@ def module():
 """
 
 ext = "so" if platform.system() == "Linux" else "dylib"
-plugin_path = get_lib_path("catalyst", "") + f"/StandalonePlugin.{ext}"
-plugin = Path(plugin_path)
+plugin = Path(get_lib_path("catalyst", "CATALYST_LIB_DIR")) / f"StandalonePlugin.{ext}"
```
dime10, line 93: Looks like there was bug hiding here previously (somehow)

paul0403 (approved):

kipawaa (approved):
This is much nicer than it was, thanks @dime10 !

frontend/test/pytest/test_precompile_decomp_rules.py
kipawaa: Much better, thank you!

PR #2293, opened 2025-12-11
Update special treatment for new PauliRot rule in graph decomp

paul0403:
**Context:**
Some decomposition test cases have been disabled previously due to incompatibility with the new `qml.PauliRot` lowering rules. With the new implementation and fixes in https://github.com/PennyLaneAI/catalyst/pull/2284, we try re-enabling the test cases, but encounter various issues in the frontend pytest and lit tests.

Currently, 5 pytests in `pytest/from_plxpr/test_decompose_transform.py` and 2 lit tests in `lit/test_decomposition.py` still fail. 

**Benefits:** Reenabling tests that shouldn't have been skipped.

**Related GitHub Issues:**
https://github.com/PennyLaneAI/catalyst/pull/2284
https://github.com/PennyLaneAI/pennylane/pull/8765

### DIFF
--- a/frontend/catalyst/from_plxpr/decompose.py
+++ b/frontend/catalyst/from_plxpr/decompose.py
@@ -255,7 +255,7 @@ class DecompRuleInterpreter(qml.capture.PlxprInterpreter):
                         "Adjoint",
                         "Controlled",
                         "TemporaryAND",
-                        "ChangeBasisOp",
+                        "ChangeOpBasis",
                         "Prod",
                     )
                 ):  # pragma: no cover


frontend/test/lit/test_decomposition.py
```diff
@@ -855,75 +856,6 @@ def circuit_16(x: float):
...
-
-    @qml.qjit(target="mlir")
-    @partial(
-        qml.transforms.decompose,
-        gate_set={"RX", "RZ", "H", "CZ"},
-    )
-    @qml.qnode(qml.device("lightning.qubit", wires=2))
-    # SKIP-CHECK-DAG: %0 = transform.apply_registered_pass "decompose-lowering"
-    # SKIP-CHECK{LITERAL}: func.func public @circuit_17() -> tensor<f64> attributes {decompose_gatesets
-    def circuit_17():
```
paul0403, line 872: This test somehow is flaky, in the sense that the graph decomp solution is not deterministic. Sometimes it contains a ppr op (despite ppr not being in the target gate set).

frontend/test/lit/test_decomposition.py
```diff
@@ -855,75 +856,6 @@ def circuit_16(x: float):
...
-
-    @qml.qjit(autograph=True, target="mlir")
-    @partial(
-        qml.transforms.decompose,
-        gate_set={"RX", "RY", "CNOT", "GlobalPhase"},
-    )
-    @qml.qnode(qml.device("lightning.qubit", wires=4))
-    # SKIP-CHECK: %0 = transform.apply_registered_pass "decompose-lowering"
-    # SKIP-CHECK: func.func public @circuit_18(%arg0: tensor<3xf64>) -> tensor<f64> attributes {decompose_gatesets
-    def circuit_18():
```
paul0403, line 906: ditto

frontend/test/lit/test_decomposition.py
paul0403: @astralcai

frontend/test/pytest/from_plxpr/test_decompose_transform.py
```diff
@@ -173,9 +173,6 @@ def circuit():
         resources = qml.specs(with_qjit, level="device")()["resources"].gate_types
         assert resources == expected_resources
 
-    @pytest.mark.xfail(
```
paul0403: Note that there are 5 other tests in this file skipped to push out jax update, due to the same paulirot issue.

paul0403: The remaining 5 pytests fail with the following error inside `--decompose-lowering` pass:
```
circuit_18_1/1_QuantumCompilationStage/1_SplitMultipleTapesPass.mlir:322:25: error: 'tensor.from_elements' op failed to verify that operand types match result element type
      %out_qubits_9:2 = quantum.multirz(%extracted_8) %8, %9 : !quantum.bit, !quantum.bit
                        ^
circuit_18_1/1_QuantumCompilationStage/1_SplitMultipleTapesPass.mlir:322:25: note: see current operation: %21 = "tensor.from_elements"(%17, %18) : (i64, i64) -> tensor<1xi64>

```

Apparantly multirz had some issues, and the new pauli rot rule is the first rule that contains a multirz in the body so this was never hit.

paul0403: Actually no, here's the list of their corresponding errors for the 5 remaining skipped pytests (all were passing as of the previous stable release):
- `test_ftqc_rotxzx`: pennylane has no module named `ftqc`
- `test_multi_qubits`: aforementioned tensor.from_elements error
- `test_adjoint`: `ValueError: Could not capture _OperatorNode(op=TemporaryAND, num_work_wire_not_available=1, work_wire_dependent=False, min_work_wires=0) without the number of wires.`. After adding `"TemporaryAND", "ChangeOpBasis", "Prod"` to the special list (@lazypanda10117 ), we discovered that the root cause is still the aforementioned tensor.from_elements error
- `test_ctrl`: aforementioned tensor.from_elements error
- `test_template_qft`: aforementioned tensor.from_elements error

PR #2296, opened 2025-12-12
Revert Catalyst ppm/ppr pass docstrings 

isaacdevlugt:
**Context:** https://github.com/PennyLaneAI/pennylane/pull/8762 Adds wrappers for transforms in PennyLane via `pass_name` that allow for using Catalyst's Pauli-based computation compilation passes (`to_ppr`, `commute_ppr`, `merge_ppr_ppm`, `ppr_to_ppm`, `ppm_compilation`, and `reduce_t_depth`). Includes docstrings for those transforms for the corresponding use cases of those transforms.

**Benefits:** Clear documented separation between PL frontend and Catalyst frontend passes.

**Possible Drawbacks:** Two frontends that clash, may confuse users.

**Related GitHub Issues:**

[sc-99855](https://app.shortcut.com/xanaduai/story/99855/amend-documentation-of-all-ppm-ppr-passes-in-catalyst-passes-to-show-measurement-outputs-and-manipulation-of-pl-operations)

### DIFF
--- a/doc/releases/changelog-0.14.0.md
+++ b/doc/releases/changelog-0.14.0.md
@@ -2,7 +2,7 @@
 
 <h3>New features since last release</h3>
 
-* Programs compiled with ``qjit`` can now be visualized with :func:`~.draw_graph`, allowing for 
+* Programs compiled with ``qjit`` can now be visualized with :func:`~.draw_graph`, allowing for
   sequentially analyzing impacts of compilation passes on structured and dynamic programs.
   [(#2213)](https://github.com/PennyLaneAI/catalyst/pull/2213)
   [(#2229)](https://github.com/PennyLaneAI/catalyst/pull/2229)
@@ -17,6 +17,7 @@
   [(#2243)](https://github.com/PennyLaneAI/catalyst/pull/2243)
   [(#2340)](https://github.com/PennyLaneAI/catalyst/pull/2340)
   [(#2357)](https://github.com/PennyLaneAI/catalyst/pull/2357)
+  [(#2363)](https://github.com/PennyLaneAI/catalyst/pull/2363)
 
   Consider the following circuit:
 
@@ -33,8 +34,8 @@
   def circuit(x, y):
       qml.X(0)
       qml.Y(1)
-      qml.H(x) 
-      
+      qml.H(x)
+
       for i in range(3):
           qml.S(0)
           qml.RX(0.1, wires=1)
@@ -51,7 +52,7 @@
       return qml.expval(qml.Z(y))
   ```
 
-  The circuit structure (``for`` loop and conditional branches) along with the dynamicism can be 
+  The circuit structure (``for`` loop and conditional branches) along with the dynamicism can be
   succinctly represented with :func:`~.draw_graph`.
 
   ```pycon
@@ -61,10 +62,10 @@
   ```
   ![Graphical representation of circuit with dynamicism and structure](../../doc/_static/catalyst-draw-graph-changelog-0.14-example.png)
 
-  By default, all compilation passes specified will be applied and visualized. However, 
-  :func:`~.draw_graph` can be used with the ``level`` argument to inspect compilation pass impacts, 
-  where the ``level`` value denotes the cumulative set of applied compilation transforms (in the 
-  order they appear) to be applied and visualized. With ``level=1``, drawing the above circuit will 
+  By default, all compilation passes specified will be applied and visualized. However,
+  :func:`~.draw_graph` can be used with the ``level`` argument to inspect compilation pass impacts,
+  where the ``level`` value denotes the cumulative set of applied compilation transforms (in the
+  order they appear) to be applied and visualized. With ``level=1``, drawing the above circuit will
   apply the ``merge_rotation`` transform only:
 
   ```pycon
@@ -76,16 +77,16 @@
   The :func:`~.draw_graph` function visualizes QNodes in a similar manner as
   `view-op-graph in traditional MLIR <https://mlir.llvm.org/docs/Passes/#-view-op-graph>`_,
   which leverages `Graphviz <https://graphviz.org/download/>`_ to show data-flow in compiled
-  programs. As such, use of :func:`~.draw_graph` requires installation of 
+  programs. As such, use of :func:`~.draw_graph` requires installation of
   `Graphviz <https://graphviz.org/download/>`_ and the
-  `pydot <https://pypi.org/project/pydot/>`_ software package. Please consult the links provided for 
-  installation instructions. Additionally, it is recommended to use :func:`~.draw_graph` with 
+  `pydot <https://pypi.org/project/pydot/>`_ software package. Please consult the links provided for
+  installation instructions. Additionally, it is recommended to use :func:`~.draw_graph` with
   PennyLane's program capture enabled (see :func:`qml.capture.enable <pennylane.capture.enable>`).
 
   [(#2298)](https://github.com/PennyLaneAI/catalyst/pull/2298)
   [(#2290)](https://github.com/PennyLaneAI/catalyst/pull/2290)
   [(#2309)](https://github.com/PennyLaneAI/catalyst/pull/2309)
-  
+
 * A new `gridsynth` pass is added to support Clifford+T decomopsition. This pass discretizes `RZ` and `PhaseShift` gates to either the Clifford+T basis or to the PPR basis. The pass also supports decomposing single-qubit arbitrary angle PPR in the Z basis.
   [(#2140)](https://github.com/PennyLaneAI/catalyst/pull/2140)
   [(#2166)](https://github.com/PennyLaneAI/catalyst/pull/2166)
@@ -226,6 +227,7 @@
   [(#2233)](https://github.com/PennyLaneAI/catalyst/pull/2233)
   [(#2284)](https://github.com/PennyLaneAI/catalyst/pull/2284)
   [(#2336)](https://github.com/PennyLaneAI/catalyst/pull/2336)
+  [(#2360)](https://github.com/PennyLaneAI/catalyst/pull/2360)
 
   :class:`~.PauliRot` and :func:`~.pauli_measure` can be manipulated with Catalyst's existing passes
   for PPR-PPM compilation, which includes :func:`catalyst.passes.to_ppr`,
@@ -283,13 +285,13 @@
   }
   ```
 
-* A new transform called :func:`~catalyst.passes.decompose_arbitrary_ppr` pass has been added, 
-  which decomposes abitrary-angle Pauli product rotations (PPRs) as outlined in Figure 13(d) from 
+* A new transform called :func:`~catalyst.passes.decompose_arbitrary_ppr` pass has been added,
+  which decomposes abitrary-angle Pauli product rotations (PPRs) as outlined in Figure 13(d) from
   `arXiv:2211.15465 <https://arxiv.org/abs/2211.15465>`_.
   [(#2304)](https://github.com/PennyLaneAI/catalyst/pull/2304)
   [(#2354)](https://github.com/PennyLaneAI/catalyst/pull/2354)
 
-  The ``decompose_arbitrary_ppr`` will decompose an arbitrary-angle PPR into a collection of PPRs, 
+  The ``decompose_arbitrary_ppr`` will decompose an arbitrary-angle PPR into a collection of PPRs,
   PPMs, and a single-qubit arbitrary PPR in the ``Z`` basis:
 
   ```python
@@ -430,9 +432,19 @@
   conform with the name of the corresponding transform in PennyLane.
   [(#2201)](https://github.com/PennyLaneAI/catalyst/pull/2201)
 
+* The `to-ppr` pass now automatically runs canonicalization patterns after converting quantum
+  operations to Pauli Product Rotation (PPR) form. This removes identity Pauli rotations
+  (e.g., `["I", "I", "I"]`) automatically, simplifying the resulting IR.
+  [(#2367)](https://github.com/PennyLaneAI/catalyst/pull/2367)
+
 <h3>Deprecations 👋</h3>
 
 <h3>Bug fixes 🐛</h3>
+
+* Fixed a bug in the `gradient.value_and_grad` verifier that incorrectly validated gradient result
+  types by matching from the tail of callee arguments, ignoring `diffArgIndices`. This caused false
+  verification errors when differentiating a subset of arguments with different types.
+  [(#2349)](https://github.com/PennyLaneAI/catalyst/pull/2349)
 
 * Updated the type support for callbacks allowing use of unsigned integers.
   [(#2330)](https://github.com/PennyLaneAI/catalyst/pull/2330)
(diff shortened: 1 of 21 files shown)


doc/releases/changelog-dev.md
```diff
@@ -112,6 +112,7 @@
   spurred from `A Game of Surface Codes (arXiv1808.02892) <https://arxiv.org/pdf/1808.02892>`_.
   [(#2145)](https://github.com/PennyLaneAI/catalyst/pull/2145)
   [(#2233)](https://github.com/PennyLaneAI/catalyst/pull/2233)
+  [()]()
```
```suggestion
  [(#2296)](https://github.com/PennyLaneAI/catalyst/pull/2296)
```

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -572,44 +577,43 @@ def to_ppr(qnode):
...
-        'to-ppr (MLIR-1)': Resources(
-            num_wires=2,
-            num_gates=7,
-            gate_types=defaultdict(<class 'int'>, {'PPR-pi/4-w1': 5, 'PPR-pi/4-w2': 1, 'PPR-pi/8-w1': 1}),
-            gate_sizes=defaultdict(<class 'int'>, {1: 6, 2: 1}),
-            depth=None,
-            shots=Shots(total_shots=None, shot_vector=())
-        )
-    }
+            return measure(1)
```
lazypanda10117: I am not sure how you want to handle the documentation here. This line will fail after turning capture on.

isaacdevlugt: yep that's fine :)

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -536,10 +536,15 @@ def circuit():
 
 
 def to_ppr(qnode):
-    R"""
-    A quantum compilation pass that converts Clifford+T gates into Pauli Product Rotation (PPR)
+    R"""A quantum compilation pass that converts Clifford+T gates into Pauli Product Rotation (PPR)
```
lazypanda10117, line 634: With #2284 , `to_ppr` also lowers `PauliRot` into PPR.

frontend/catalyst/passes/builtin_passes.py
lazypanda10117: I am still a little concerned about the documentation around this. This is not a code pattern that should be encouraged, especially in an official documentation. We should always return a `MeasurementProcess`. 
However, we do not have proper lowering of Measurement processes into PPMs (this needs actual investigation, it is not as simple as a direct lowering). Without this, the `merge_ppr_ppm` pass will not work properly, as it requires the final operations to be PPMs in order for PPRs to merge over it and be removed. 

We may want to capture this info somewhere.

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -950,53 +950,60 @@ def ppm_compilation(
...
 
     .. code-block:: python
 
         import pennylane as qml
-        from functools import partial
+        from catalyst import qjit, measure
+        from catalyst.passes import ppm_compilation
 
-        qml.capture.enable()
+        pipeline = [("pipe", ["enforce-runtime-invariants-pipeline"])]
```
lazypanda10117: pipeline = [("pipe", ["quantum-compilation-stage"])]

change this and other occurrences.

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -1018,8 +1025,7 @@ def circuit():
 
 
 def ppm_specs(fn):
```
lazypanda10117, line 1149: This is completely supplanted by the new `qml.spec` feature.  Might want to direct users to use `qml.spec` here.

lazypanda10117 (approved):
Thanks @isaacdevlugt 💯 Some minor feedbacks. Main concern is still around the usage of `measure(0)` as the final operation.

frontend/catalyst/passes/builtin_passes.py
dime10: I would recommend not returning anything! That doesn't mean that the body of the circuit doesn't have any measurements (including final, i.e. the last, ones), but I don't think example programs for the purpose of PBC compilation need to have explicit return values (especially since they are as of yet still non-executable).

lazypanda10117: I agree, but I guess there is a small issue (?) when we don't have explicit return values. Graph decomposition does not work when we don't a final return value, and i guess something like this
```
@to_ppr
@qml.transforms.decompose(...)
def circuit()
```
is a common usecase (e.g, in the XAS workflow, but the XAS workflow has a final return value).

dime10: > Graph decomposition does not work when we don't a final return value

🫨  What the heck, why does it care about return values? I thought it's only supposed to touch gates 🤔 

In that case we might have no choice. Could also return an integer (e.g. `return 0`), maybe that's slightly less controversial than the mcm

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -536,10 +536,15 @@ def circuit():
...
-    R"""
-    A quantum compilation pass that converts Clifford+T gates into Pauli Product Rotation (PPR)
+    R"""A quantum compilation pass that converts Clifford+T gates into Pauli Product Rotation (PPR)
     gates.
 
+    .. note::
+
+        For improved integration with the PennyLane frontend, including inspectability with
+        :func:`pennylane.specs`, please use :func:`pennylane.transforms.to_ppr` with program capture
+        enabled (:func:`pennylane.capture.enable`).
```
```suggestion
        :func:`pennylane.specs`, please use :func:`pennylane.transforms.to_ppr`.
```

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -620,82 +624,94 @@ def commute_ppr(qnode=None, *, max_pauli_size=0):
     :math:`\exp(-{iP\tfrac{\pi}{4}})`, past non-Clifford PPRs gates,
     :math:`\exp(-{iP\tfrac{\pi}{8}})`, where :math:`P` is a Pauli word.
 
+    .. note::
+
+        For improved integration with the PennyLane frontend, including inspectability with
+        :func:`pennylane.specs`, please use :func:`pennylane.transforms.commute_ppr` with program
+        capture enabled (:func:`pennylane.capture.enable`).
```
```suggestion
        :func:`pennylane.specs`, please use :func:`pennylane.transforms.commute_ppr`.
```

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -710,74 +726,85 @@ def merge_ppr_ppm(qnode=None, *, max_pauli_size=0):
...
-    the `Compilation Hub <https://pennylane.ai/compilation/pauli-product-measurement>`_.
-
     .. note::
 
-        The circuits that generated from this pass are currently not executable on any backend.
-        This pass is only for analysis with the ``null.qubit`` device and potential future execution
-        when a suitable backend is available.
+        For improved integration with the PennyLane frontend, including inspectability with
+        :func:`pennylane.specs`, please use :func:`pennylane.transforms.merge_ppr_ppm` with program
+        capture enabled (:func:`pennylane.capture.enable`).
```
```suggestion
        :func:`pennylane.specs`, please use :func:`pennylane.transforms.merge_ppr_ppm`.
```

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -791,96 +818,76 @@ def ppr_to_ppm(qnode=None, *, decompose_method="pauli-corrected", avoid_y_measur
     A quantum compilation pass that decomposes Pauli product rotations (PPRs),
     :math:`P(\theta) = \exp(-iP\theta)`, into Pauli product measurements (PPMs).
 
+    .. note::
+
+        For improved integration with the PennyLane frontend, including inspectability with
+        :func:`pennylane.specs`, please use :func:`pennylane.transforms.ppr_to_ppm` with program
+        capture enabled (:func:`pennylane.capture.enable`).
```
```suggestion
        :func:`pennylane.specs`, please use :func:`pennylane.transforms.ppr_to_ppm`.
```

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -904,6 +911,12 @@ def ppm_compilation(
     A quantum compilation pass that transforms Clifford+T gates into Pauli product measurements
     (PPMs).
 
+    .. note::
+
+        For improved integration with the PennyLane frontend, including inspectability with
+        :func:`pennylane.specs`, please use :func:`pennylane.transforms.ppm_compilation` with
+        program capture enabled (:func:`pennylane.capture.enable`).
```
```suggestion
        :func:`pennylane.specs`, please use :func:`pennylane.transforms.ppm_compilation`.
```

frontend/catalyst/passes/builtin_passes.py
isaacdevlugt:
```suggestion
            return 0
```

This okay?

frontend/catalyst/passes/builtin_passes.py
isaacdevlugt: There's this note: 

```
    .. note::

        It is recommended to use :func:`pennylane.specs` instead of ``ppm_specs`` to retrieve
        resource counts of PPR-PPM workflows.
```

isaacdevlugt: (line 1036)

frontend/catalyst/passes/builtin_passes.py
lazypanda10117: I think it needs to be an actual measurement process. The way graph decomp works is that it has its own decompose interpreter, and it needs to collect all the decomposition rules for operations within the circuit. However, to figure out which operations are in the circuit, it relies on the assumptipn that measurement process is the end of a circuit, so seeing a measurement process means we have seen all the  existing operations (see the logic in 'interpret_measurement'). Maybe @maliasadi can confirm here.

doc/releases/changelog-dev.md
```diff
@@ -294,14 +294,12 @@
 
 <h3>Bug fixes 🐛</h3>
 
-<<<<<<< mlxd/fix_cond_measure_test
```
paul0403: Remember to fix git conflicts!

frontend/catalyst/passes/builtin_passes.py
paul0403: The measurement contraint for graph decomp has been resolved since then https://github.com/PennyLaneAI/catalyst/pull/2312

But I'm curious as to why is graph decomp involved here at all. Isn't this just the `to-ppr` pass?

paul0403: I just tried, on latest rc branch, with

```python
import pennylane as qml

qml.capture.enable()  # works both with or without capture enabled

ppm_passes = [("PPM", ["to-ppr"])]
@qjit(pipelines=ppm_passes, target="mlir")
def circuit():
    qml.H(0)
    qml.CNOT([0, 1])
    qml.T(0)
    return

print(circuit.mlir)
print(circuit.mlir_opt)
```

I get the correct quantum dialect mlir and the ppr dialect mlir.

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -668,44 +673,43 @@ def to_ppr(qnode):
...
     .. code-block:: python
 
         import pennylane as qml
+        from catalyst import qjit, measure
 
-        qml.capture.enable()
+        ppm_passes = [("PPM", ["to-ppr"])]
 
-        @qml.qjit(target="mlir")
-        @qml.transforms.to_ppr
```
paul0403, line 680: Why are we reverting this? Isn't the overall goal to use pennylane as frontend, and for catlayst to not have its own frontend?

paul0403: Because I just tried and the following works fine

```python
import pennylane as qml


qml.capture.enable()

p = [("my_pipe", ["quantum-compilation-stage"])]  # halt pipeline after quantum passes so it doesn't reach lowering

@qml.qjit(pipelines=p, target="mlir")
@qml.transforms.to_ppr
def circuit():
    qml.H(0)
    qml.CNOT([0, 1])
    qml.T(0)
    return

print(circuit.mlir)
print(circuit.mlir_opt)
```

paul0403: I believe this is the usage pattern that we want to advertise, right? Everything goes with `qml.blah`. The only thing we need to note to the user is that ppr passes in catalyst can only be run within the compiler, and there is no execution of ppr ops, so we need to halt the pipeline before lowering.

In fact, I would even argue we added these wrappers in PL https://github.com/PennyLaneAI/pennylane/pull/8762 precisely because we want to use `qml.transforms.to_ppr`, instead of `catalyst.passes.to_ppr`

paul0403: Discussed offline, we will keep catalyst docs to reflect catalyst frontend. The `qml.transforms.to_ppr` usage pattern is included in the PL docs.

doc/releases/changelog-0.14.0.md
```diff
@@ -224,10 +224,10 @@
   spurred from [A Game of Surface Codes (arXiv1808.02892)](https://arxiv.org/pdf/1808.02892).
   [(#2145)](https://github.com/PennyLaneAI/catalyst/pull/2145)
   [(#2233)](https://github.com/PennyLaneAI/catalyst/pull/2233)
+  [(#2296)](https://github.com/PennyLaneAI/catalyst/pull/2296)
```
isaacdevlugt:
```suggestion
```

covered in https://github.com/PennyLaneAI/catalyst/pull/2351

frontend/catalyst/passes/builtin_passes.py
lazypanda10117: thanks!

frontend/catalyst/passes/builtin_passes.py
lazypanda10117: @paul0403 Re Graph Decomposition: Returning empty works with specific PPM/PPR passes always work, but when you want to use it with graph decomposition, then returning empty wouldn't work because of the aforementioned issue (before the fix you mentioned). The example use case I have in mind is when users write custom decomposition rules to PPM/PPR, and later apply a pass like `ppr_to_ppm` on the decomposed circuit, then this would be a problem.

paul0403: Can you post a code example? @lazypanda10117

isaacdevlugt: @lazypanda10117 any ideas here? Otherwise this PR should be merged soon

lazypanda10117: This should now be fixed as part of https://github.com/PennyLaneAI/catalyst/pull/2312, as Paul mentioned above. 
What i meant was that previously, if we have a circuit like this, it will not decompose the Hadamard gate
```python
qml.capture.enable()

@qml.qjit(pipelines=[("pipe", ["quantum-compilation-stage"])], target="mlir")
@qml.transforms.decompose(gate_set=[qml.RZ, qml.RX, qml.GlobalPhase, qml.Z])
def circuit():
    qml.H(0)
    return

print(circuit.mlir_opt)

>>>    ...
>>>    %out_qubits = quantum.custom "Hadamard"() %1 : !quantum.bit
>>>    ...
```

It will only work if we return a measurement process like following:
```python
qml.capture.enable()

@qml.qjit(pipelines=[("pipe", ["quantum-compilation-stage"])], target="mlir")
@qml.transforms.decompose(gate_set=[qml.RZ, qml.RX, qml.GlobalPhase, qml.Z])
def circuit():
    qml.H(0)
    return qml.expval(qml.Z(0))

print(circuit.mlir_opt)

>>>    ...
>>>    %out_qubits = quantum.custom "RZ"(%cst) %3 : !quantum.bit
>>>    %out_qubits_2 = quantum.custom "RX"(%cst) %out_qubits : !quantum.bit
>>>    %out_qubits_3 = quantum.custom "RZ"(%cst) %out_qubits_2 : !quantum.bit
>>>.   ....
```

Feel free to resolve the comment as you see fit.

isaacdevlugt: @paul0403 what do you think? I think it's ok to resolve this

paul0403: Yes it's ok to resolve this now, the issue has been fixed

PR #2299, opened 2025-12-15
New OQD end-to-end pipeline is added to Catalyst

rniczh:
**Context:**

This pipeline enables Catalyst to compile circuit for ARTIQ-based quantum device. When `device_db` configuration is provided, the compilation follows the ARTIQ route:
- `ions-decomposition`: Decompose quantum operations for trapped ion systems
- `gates-to-pulses`: Convert gate operations to pulse sequences
- `convert-ion-to-rtio`: Convert ion operations to RTIO dialect
- `convert-rtio-event-to-artiq`: Lower RTIO event to ARTIQ's primitives
- `llvm-dialect-lowering-stage`: Lower to LLVM IR
- `emit-artiq-runtime`: Generate ARTIQ runtime entry point as wrapper for ARITQ device to execute

The final stage compiles LLVM IR to an ELF binary targeting the ARTIQ device (ARM Cortex-A9). Due to current limitations where Catalyst's internal LLVM build does not include the corresponding ARM backend, the compilation will use `llc` to compile `.ll` to object file, and use `ld.lld` to compile to `.elf`

Note: This PR only covers LLVM IR generation; the `compile_to_artiq` function itself is not included.

***Example:***
Replace the artiq configs to your own env setting:
```python
import os

import numpy as np
import pennylane as qml

from catalyst import QJIT, CompileOptions
from catalyst.third_party.oqd import OQDDevice, OQDDevicePipeline

OQD_PIPELINES = OQDDevicePipeline(
    os.path.join("calibration_data", "device.toml"),
    os.path.join("calibration_data", "qubit.toml"),
    os.path.join("calibration_data", "gate.toml"),
    os.path.join("device_db", "device_db.json"),
)


def test_rx_gate():
    """Test RX gate with ARTIQ linking done in user code."""
    oqd_dev = OQDDevice(
        backend="default",
        shots=4,
        wires=1
    )
    qml.capture.enable()

    # Compile to LLVM IR only
    @qml.qnode(oqd_dev)
    def circuit():
        x = np.pi / 2
        qml.RX(x, wires=0)
        return qml.counts(wires=0)
        
    compiled_circuit = QJIT(circuit, CompileOptions(link=False, pipelines=OQD_PIPELINES))

    # Compile to ARTIQ ELF (using `compile_to_artiq` function from third-party repo to compile to it)
    artiq_config = {
        "kernel_ld": "/path/to/kernel.ld",
        "llc_path": "/path/to/llc",
        "lld_path": "/path/to/ld.lld",
    }

    output_elf_path = compile_to_artiq(compiled_circuit, artiq_config)
    print(f"ARTIQ ELF file generated: {output_elf_path}")

test_rx_gate()
```

The result will store to `circuit.elf`

```
[ARTIQ] Generated ELF: circuit.elf
```

And you can run it on artiq device:

```shell
artiq_run  --device-db device_db.py circuit.elf
```

**Benefits:**

**Possible Drawbacks:**
~It causes the OQD specific compile stuffs steps into the original catalyst compiler driver's pipepline~

**Related GitHub Issues:**
[sc-100853]


mlir/lib/Driver/CompilerDriver.cpp
```diff
@@ -805,20 +806,85 @@ LogicalResult QuantumDriverMain(const CompilerOptions &options, CompilerOutput &
...
+
+            std::string lldCmd = lldPath +
+                                 " -shared --eh-frame-hdr -m armelf_linux_eabi "
+                                 "--target2=rel -T " +
+                                 options.artiqKernelLd + " " + options.getObjectFile() + " -o " +
+                                 options.getElfFile() + " 2>&1";
+
+            CO_MSG(options, Verbosity::All, "[ARTIQ] Linking ELF: " << lldCmd << "\n");
+
+            int lldResult = std::system(lldCmd.c_str());
```
rniczh: OS command Injection

rniczh: If we want to use LLVM C++ API to compile it to object file directly without `llc` steps in, we need to add ARM target when building LLVM
```
-DLLVM_TARGETS_TO_BUILD="host;ARM" 
```
And on more thing should be noted is that the we also need to build `lld` to get the linker as well.

Need `lld` in LLVM_PROJECTS, for clarification: 
`LLVM_ENABLE_LLD=ON` is different, this is about using lld as the linker during LLVM’s own build, not necessarily building the ld.lld binary
```
LLVM_PROJECTS ?= mlir;lld
```

rniczh: Done https://github.com/PennyLaneAI/catalyst/pull/2299/commits/ffa587565473fef3f7559654206825be0ce87889

frontend/catalyst/third_party/oqd/oqd_compile.py
```diff
@@ -0,0 +1,229 @@
...
+    Returns:
+        str: Path to the generated binary file
+    """
+    # Get LLVM IR text and write to file
+    llvm_ir_text = circuit.qir
+    circuit_name = getattr(circuit, "__name__", "circuit")
+    llvm_ir_path = os.path.join(str(circuit.workspace), f"{circuit_name}.ll")
+    with open(llvm_ir_path, "w", encoding="utf-8") as f:
+        f.write(llvm_ir_text)
+    print(f"LLVM IR file written to: {llvm_ir_path}")
```
paul0403: Only print when verbose? Or should we always print?

frontend/catalyst/third_party/oqd/oqd_compile.py
```diff
@@ -0,0 +1,229 @@
...
+
+    # Link to ARTIQ's binary
+    if output_elf_name is None:
```
paul0403: Why do we need this condition? :thinking: Is it for the user to specify a path they want themselves?

frontend/catalyst/third_party/oqd/oqd_compile.py
```diff
@@ -0,0 +1,229 @@
...
+        verbose: Whether to print verbose output
+
+    Raises:
+        RuntimeError: If compilation fails
+        FileNotFoundError: If llc is not found
+    """
+    llc_args = [
+        llc_cmd,
+        "-mtriple=armv7-unknown-linux-gnueabihf",
+        "-mcpu=cortex-a9",
```
paul0403: Does this constaint the type of the host machine? i.e. does this mean an OQD script can only be run on these kinds of systems?

frontend/catalyst/third_party/oqd/oqd_compile.py
```diff
@@ -0,0 +1,229 @@
...
+    Raises:
+        RuntimeError: If linking fails
+        FileNotFoundError: If ld.lld is not found
+    """
+    lld_args = [
+        lld_cmd,
+        "-shared",
+        "--eh-frame-hdr",
+        "-m",
+        "armelf_linux_eabi",
```
paul0403: Same here

frontend/catalyst/third_party/oqd/oqd_compile.py
```diff
@@ -0,0 +1,229 @@
...
+        str(kernel_ld),
+        str(object_file),
+        "-o",
+        str(output_elf_path),
+    ]
+
+    if verbose:
+        print(f"[ARTIQ] Linking ELF: {' '.join(lld_args)}")
+
+    try:
```
paul0403: This try-except block seems identical for llc and lld. Maybe we can factor them out into a helper function? But I'm also fine with it if you leave it as is :+1:

mlir/lib/RTIO/Transforms/RTIOEmitARTIQRuntime.cpp
```diff
@@ -0,0 +1,164 @@
...
+        ModuleOp module = getOperation();
+        MLIRContext *ctx = &getContext();
+        OpBuilder builder(ctx);
+
+        // Check if __modinit__ already exists (the entry point of ARTIQ device)
+        if (module.lookupSymbol<LLVM::LLVMFuncOp>(ARTIQRuntime::modinit)) {
+            return;
+        }
+
+        // Find the kernel function e(could be LLVM func or func.func)
```
paul0403:
```suggestion
        // Find the kernel function (could be LLVM func or func.func)
```

mlir/lib/RTIO/Transforms/RTIOEmitARTIQRuntime.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+        // Create ARTIQ runtime wrapper
+        if (failed(emitARTIQRuntimeForLLVMFunc(module, builder, llvmKernelFunc))) {
+            return signalPassFailure();
+        }
+    }
+
+  private:
+    /// Emit ARTIQ runtime wrapper for LLVM dialect kernel function
+    LogicalResult emitARTIQRuntimeForLLVMFunc(ModuleOp module, OpBuilder &builder,
```
paul0403: `module` is a keyword in cpp, let's avoid it as a variable name maybe

paul0403 (commented):
Thanks Hongsheng, looks really nice! I think this is the minimally intrusive way we can do it without altering too much of the base jit flow!

I left some comments but all are minor. Another thing is can you add your PR description's example as an end-to-end pytest in the frontend oqd tests? Obviously no need for execution, just check that the compiled ELF exist (maybe check a little bit of its content as well)? You may need to do things like `subprocess.run("which llc", shell=True)` in the test, but I think that's possible.

frontend/catalyst/third_party/oqd/oqd_compile.py
rniczh: So the function `compile_to_artiq` is solely compile to ARTIQ and this is the hardware that ARTIQ employed. Well, it might be chnaged if the OQD target to different FPGA board (except the ARTIQ).

frontend/catalyst/third_party/oqd/oqd_compile.py
rniczh: https://github.com/PennyLaneAI/catalyst/pull/2299#discussion_r2699841178

frontend/catalyst/third_party/oqd/__init__.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2024 Xanadu Quantum Technologies Inc.
+# Copyright 2024-2026 Xanadu Quantum Technologies Inc.
```
mehrdad2m: Why the copyright change? 😄

frontend/catalyst/third_party/oqd/oqd_device.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2024 Xanadu Quantum Technologies Inc.
+# Copyright 2024-2026 Xanadu Quantum Technologies Inc.
```
mehrdad2m: Same here

mlir/lib/RTIO/Transforms/RTIOEmitARTIQRuntime.cpp
rniczh: Thanks for pointing out!

frontend/catalyst/third_party/oqd/oqd_compile.py
```diff
@@ -0,0 +1,229 @@
...
+import subprocess
+from pathlib import Path
+from typing import Optional
+
+
+def compile_to_artiq(circuit, artiq_config, output_elf_name=None, verbose=True):
+    """Compile a qjit-compiled circuit to ARTIQ's binary.
+
+    This function takes a circuit compiled with target="llvmir", writes the LLVM IR
+    to a file, and links it to an ARTIQ'se binary.
```
mehrdad2m:
```suggestion
    to a file, and links it to an ARTIQ's binary.
```

frontend/catalyst/compiler.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2022-2023 Xanadu Quantum Technologies Inc.
+# Copyright 2022-2026 Xanadu Quantum Technologies Inc.
```
mehrdad2m: same here

frontend/catalyst/jit.py
```diff
@@ -546,6 +546,7 @@ def __init__(self, fn, compile_options):
         self.mlir_module = None
         self.out_type = None
         self.overwrite_ir = None
+        # Use cwd for workspace if keep_intermediate is set
```
mehrdad2m:
```suggestion
```
This comment seems redundant since the code is very verbose itself

frontend/catalyst/jit.py
```diff
@@ -621,16 +626,18 @@ def aot_compile(self):
         self.workspace = self._get_workspace()
 
         # TODO: awkward, refactor or redesign the target feature
-        if self.compile_options.target in ("jaxpr", "mlir", "binary"):
+        if self.compile_options.target in ("jaxpr", "mlir", "binary", "llvmir"):
```
mehrdad2m:
```suggestion
        if self.compile_options.target in ("jaxpr", "mlir", "llvmir", "binary"):
```
just to be consistent with the order in the other lists 😄

mehrdad2m (commented):
Thanks @rniczh, I did a first round of review with some minor comments. However, I would like to play around with a bit more and maybe more review on Monday.

frontend/catalyst/third_party/oqd/oqd_compile.py
rniczh: Yes, it's also allow user to specify a path so that they can choose the destiantation.

For example:

```python
output_elf_path = compile_to_artiq(circuit, oqd_dev.artiq_config, "./result.elf")
```

frontend/catalyst/third_party/oqd/oqd_compile.py
rniczh: Hmm... if always intend to print the log to let user know what going on. I think that verbose is not needed.
I just removed it in this commit: https://github.com/PennyLaneAI/catalyst/pull/2299/commits/1fbcd7a09adc458d8a178ad701ed53acd98261e2

mlir/lib/RTIO/Transforms/RTIOEmitARTIQRuntime.cpp
rniczh: Done https://github.com/PennyLaneAI/catalyst/pull/2299/commits/9f95ddd6d294ebc00d86762c2a3b70b75df9153f

frontend/catalyst/third_party/oqd/oqd_device.py
rniczh: Done https://github.com/PennyLaneAI/catalyst/pull/2299/commits/9f95ddd6d294ebc00d86762c2a3b70b75df9153f

frontend/catalyst/third_party/oqd/__init__.py
rniczh: Done https://github.com/PennyLaneAI/catalyst/pull/2299/commits/9f95ddd6d294ebc00d86762c2a3b70b75df9153f

frontend/catalyst/compiler.py
rniczh: Done  https://github.com/PennyLaneAI/catalyst/pull/2299/commits/9f95ddd6d294ebc00d86762c2a3b70b75df9153f

frontend/catalyst/third_party/oqd/oqd_compile.py
rniczh: Done https://github.com/PennyLaneAI/catalyst/pull/2299/commits/f13f1efbc231ca4d80fb0748f22899a202c70b76

mlir/Makefile
```diff
@@ -36,8 +36,9 @@ USE_SANITIZER_NAMES=""
 USE_SANITIZER_FLAGS=""
 endif
 
-LLVM_PROJECTS ?= mlir
-LLVM_TARGETS ?= check-mlir llvm-symbolizer
+LLVM_PROJECTS ?= mlir;lld
```
paul0403: Is there a way to install lld when `ENABLE_OQD=ON`?

paul0403: Like here https://github.com/PennyLaneAI/catalyst/blob/main/Makefile#L192, in `make pytest`, in the oqd portion? Because this test is the only place where we need it right? Can we just add `apt-get install lld` here?

rniczh: I don't think placing `apt-get install lld` is suitable here, since mac user cannot use this command to install lld

paul0403 (approved):

frontend/catalyst/compiler.py
mehrdad2m:
```suggestion
# Copyright 2022-2023 Xanadu Quantum Technologies Inc.
```

frontend/catalyst/jit.py
```diff
@@ -603,6 +603,10 @@ def __call__(self, *args, **kwargs):
 
         requires_promotion = self.jit_compile(args, **kwargs)
 
+        # For llvm-ir target, compilation is complete, no execution needed
+        if self.compile_options.target == "llvmir":
+            return None
```
mehrdad2m: Can we add a test for this? I think it make sense to add coverage at least for the core files that are modified.

frontend/catalyst/jit.py
```diff
@@ -656,6 +662,15 @@ def jit_compile(self, args, **kwargs):
             bool: whether the provided arguments will require promotion to be used with the compiled
                   function
         """
+        if self.compile_options.target == "llvmir":
+            if self.mlir_module is not None:
```
mehrdad2m: Same here, it would be nice to have some tests for this as well.

frontend/catalyst/jit.py
```diff
@@ -820,6 +836,9 @@ def compile(self):
         else:
             shared_object, llvm_ir = self.compiler.run(self.mlir_module, self.workspace)
 
+        if self.compile_options.target == "llvmir":
+            return None, llvm_ir
```
mehrdad2m, line 831: And here

frontend/catalyst/jit.py
rniczh: Thanks, I added the llvm ir test here: https://github.com/PennyLaneAI/catalyst/pull/2299/commits/e9e40923007496a9b70b7a97d0167906805c1eb6

frontend/catalyst/jit.py
rniczh: Done https://github.com/PennyLaneAI/catalyst/pull/2299/commits/e9e40923007496a9b70b7a97d0167906805c1eb6

frontend/catalyst/jit.py
rniczh: Done https://github.com/PennyLaneAI/catalyst/pull/2299/commits/e9e40923007496a9b70b7a97d0167906805c1eb6

mehrdad2m (approved):
🥳

frontend/catalyst/jit.py
dime10: This is not inline with how the existing `target` feature works, which could be confusing. `target` so far only limits AOT compilation, but when the function is invoked compilation is done all the way in order to produce results. We should either change this behaviour globally, or remove these lines.

Do you have a need to call the qjit function if you don't want results? Note that `jit_compile` can be invoked directly if that's all you need (or AOT mode can used via type-hints).

frontend/test/test_oqd/oqd/test_oqd_artiq_llvmir.py
```diff
@@ -0,0 +1,94 @@
...
+        @qjit(pipelines=oqd_pipelines, target="llvmir")
+        @qml.set_shots(4)
+        @qml.qnode(oqd_dev)
+        def circuit():
+            x = np.pi / 2
+            qml.RX(x, wires=0)
+            return qml.counts(all_outcomes=True)
+
+        # Get the LLVM IR
+        llvm_ir = circuit.qir
```
dime10: I think it's time we renamed this attribute `llvmir` 😅

frontend/catalyst/jit.py
```diff
@@ -656,6 +662,15 @@ def jit_compile(self, args, **kwargs):
             bool: whether the provided arguments will require promotion to be used with the compiled
                   function
         """
+        if self.compile_options.target == "llvmir":
+            if self.mlir_module is not None:
+                return False
```
dime10: I don't understand this branch, if the mlir module is there, you don't do anything (even though you want the llvmir)?

frontend/catalyst/jit.py
```diff
@@ -621,16 +625,18 @@ def aot_compile(self):
...
                 self.user_sig or ()
             )
 
-        if self.compile_options.target in ("mlir", "binary"):
+        if self.compile_options.target in ("mlir", "llvmir", "binary"):
             self.mlir_module = self.generate_ir()
 
-        if self.compile_options.target in ("binary",):
+        if self.compile_options.target in ("llvmir", "binary"):
             self.compiled_function, _ = self.compile()
```
dime10: Don't you want to store the result of the compilation (the llvmir I think)?

frontend/catalyst/jit.py
dime10, line 1: I think the current changes are poking quite a few holes in the QJIT class, which was to designed to have very well-defined, discrete stages arranged in sequence, and which can be replaced at will as long as the input/output contract between each stage is fulfilled.

Is there a need here to split the compile stage either at the llvmir or object file generation, leaving the linking/executable generation to a separate stage? Would that prevent all these special branches?

Or can the feature get away with adding llmvir to the AOT pipeline only, without affecting the rest?

frontend/catalyst/jit.py
dime10: Or is it because accessing the `.qir` attribute recompiles starting with the mlir_module anyway, in which case `target="mlir"` would be sufficient since you're not getting the llvmir from the first compilation anyway?

frontend/catalyst/jit.py
mehrdad2m: > Is there a need here to split the compile stage either at the llvmir or object file generation, leaving the linking/executable generation to a separate stage? Would that prevent all these special branches?

I would love to see this happen. I didn't raise this because I thought maybe it's too much of a change in the core function for this PR, but I think that could help minimize the special branching. 

> Or can the feature get away with adding llmvir to the AOT pipeline only, without affecting the rest?

I think AOT should be enough for testing this, but is there a reason we don't want the `target` option in JIT mode? Personally, I would prefer to unify the behaviour.

frontend/test/test_oqd/oqd/test_oqd_artiq_llvmir.py
rniczh: Done, i changed all the `.qir` usage in the latest revision to `.llvmir`

frontend/catalyst/jit.py
rniczh: I think we only need to use the AOT for OQD. So I removed this line.

frontend/catalyst/jit.py
rniczh: I need the result of the compilation, the llvm ir, I cached the llvm ir before calling the `.llvmir`. https://github.com/PennyLaneAI/catalyst/pull/2299/changes#diff-648249b9fca379a495b52fc5c000fb7c4425daa46b6044df3ddf4515b3bc7b0aR637

frontend/catalyst/jit.py
rniczh: Beacuse I want to stop at the llvm ir when compiling with OQD, but as what you said, this behaviour not aligned with other `targets`, so I added a `link` option to qjit to determine whether we need to link to shared object. How do you think?
https://github.com/PennyLaneAI/catalyst/pull/2299/changes#diff-648249b9fca379a495b52fc5c000fb7c4425daa46b6044df3ddf4515b3bc7b0aR81

frontend/catalyst/jit.py
rniczh: @dime10  I added `link` option to the compiler to stop at llvm ir before it linking to shared object rather than specializing the `llvmir`. But I don't know whether this is a good idea to add this option.

dime10: > I think AOT should be enough for testing this, but is there a reason we don't want the target option in JIT mode? Personally, I would prefer to unify the behaviour.

So one reason is that JIT is only invoked from a call, and if you're calling the function it's to get results no?

frontend/test/test_oqd/oqd/test_oqd_artiq_llvmir.py
dime10: Make sure to mention this as a breaking change in the changelog :)

frontend/catalyst/compiler.py
```diff
@@ -484,16 +485,22 @@ def run_from_ir(self, ir: str, module_name: str, workspace: Directory):
         else:
             out_IR = None
 
+        if not self.options.link:
+            output = output_ir_name if os.path.exists(output_ir_name) else None
+            if os.path.exists(tmp_infile_name):
+                os.remove(tmp_infile_name)
+            return output, out_IR
```
dime10: Does it make sense to return both the output IR as a string and as a file?

frontend/catalyst/compiler.py
```diff
@@ -484,16 +485,22 @@ def run_from_ir(self, ir: str, module_name: str, workspace: Directory):
         else:
             out_IR = None
 
+        if not self.options.link:
+            output = output_ir_name if os.path.exists(output_ir_name) else None
```
dime10: What controls whether the LLVM IR is generated? For some configuration this will be None right?

frontend/catalyst/jit.py
dime10: Thanks, that's definitely less intrusive!

frontend/catalyst/compiler.py
rniczh: No, IR is clearly enough, it was reundnat. And we actually don't need to check whether the file exists. Fixed in https://github.com/PennyLaneAI/catalyst/pull/2299/commits/4c04388268bf7111c47776c1badb82af8c477bf7

```python
if not self.options.link:
            if os.path.exists(tmp_infile_name):
                os.remove(tmp_infile_name)
            return None, out_IR
```

frontend/catalyst/jit.py
mehrdad2m: > So one reason is that JIT is only invoked from a call, and if you're calling the function it's to get results no?

Oh yeah that's practically true. You don't usually need the target in jit mode. What I meant is while unifying the behaviour wouldn't practically be useful but it is not hurtful either and actually would make the UI more readable. I think the fact that we allow something like:
```python
@qjit(target='mlir')
@qml.qnode(dev)
def f(x):
  qml.RY(x, wires=0)
  ...
f(1.0)
```
And qjit argument has no effect becasue we are in jit mode is very confusing.

frontend/catalyst/compiler.py
rniczh: Yes, but even it fails to lower to llvm ir or lower to llvm is not set, once the `link` option is not set, It's still cannot link whatsoever. And https://github.com/PennyLaneAI/catalyst/pull/2299#discussion_r2718805066 we just need to return out_ir for the case we don't want to link to object file.

frontend/catalyst/jit.py
dime10: Would you prefer that to raise an error then? Because if the compilation isn't forced then there is nothing to execute.
Or just return nothing? (I think this last option would be much more dangerous though)

frontend/catalyst/compiler.py
dime10: Sounds good, then I think we can delete it if it exists though (same as cleanup below)

frontend/catalyst/compiler.py
dime10: Yeah the mechanism is safe I think, but how would you get it to actually produce the llvmir here?

rniczh: The compiler option `lower_to_llvm` defaults to True in CompileOptions (defined in `pipelines.py`). Unless explicitly disabled, it remains True.
If a user disables it, for example:
```
QJIT(circuit, CompileOptions(..., lower_to_llvm=False))
```
This will trigger an assertion at the beginning of `run_from_ir`:
```
assert (    
          self.options.lower_to_llvm, ...
```

frontend/catalyst/compiler.py
rniczh: Done https://github.com/PennyLaneAI/catalyst/pull/2299/commits/4e8b98251e1bf1605921cc53ed201897159bfcd1

frontend/catalyst/compiler.py
dime10: Awesome, thanks!

frontend/catalyst/compiler.py
```diff
@@ -484,16 +485,23 @@ def run_from_ir(self, ir: str, module_name: str, workspace: Directory):
...
             out_IR = None
 
+        if not self.options.link:
+            if os.path.exists(tmp_infile_name):
+                os.remove(tmp_infile_name)
+            if os.path.exists(output_ir_name):
+                os.remove(output_ir_name)
+            return None, out_IR
+
         output = LinkerDriver.run(output_object_name, options=self.options)
```
dime10:
```suggestion
        if self.options.link:
            output = LinkerDriver.run(output_object_name, options=self.options)
```

frontend/catalyst/compiler.py
```diff
@@ -484,16 +485,23 @@ def run_from_ir(self, ir: str, module_name: str, workspace: Directory):
...
-        output_object_name = str(pathlib.Path(output).absolute())
+        output = str(pathlib.Path(output).absolute())
```
dime10:
```suggestion
            output = str(pathlib.Path(output).absolute())
        else:
            output = None
```

dime10 (approved):
Little cleanup to avoid duplicate code

frontend/catalyst/jit.py
mehrdad2m: Yeah I think raising an error (or at least a warning) make sense since we are providing an implicit `target` by calling the function and yet explicitly providing a different one.

rniczh: I do agree to raise an error if nothing to execute. Just quickly submitted a commmit: https://github.com/PennyLaneAI/catalyst/pull/2299/commits/7ec3d5f209859fa546729e5d5df1603b1a57ce7d

So there are two checks to be added:
1. Only insert compiled function to `fn_cache` if `link` option is set.
```python
 if self.compile_options.link:
                self.fn_cache.insert(self.compiled_function, args, self.out_treedef, self.workspace)
 ```
 2. Ensure the compiled function exists at the begining fo the `run` function.
 ```python
 if self.compiled_function is None:
            raise CompileError(
                "Cannot execute function: no compiled function available. "
                "The function must be compiled before execution."
            )
```

So it's expected to get the error if user try to run the circuit without `link`
```
catalyst.utils.exceptions.CompileError: Cannot execute function: no compiled function available. The function must be compiled before execution.
```

dime10: >Yeah I think raising an error (or at least a warning) make sense since we are providing an implicit target by calling the function and yet explicitly providing a different one.

Sounds fine by me, this dates all the way back so can't be sure, but I think reasoning was JIT compiling (i.e. calling the function) overrides whatever previous compilation you did, unless it matches what you already have exactly. This is the same reasoning that providing a different signature than what was specified for AOT compilation recompiles with the new signature regardless (although it does raise a warning I believe).

dime10: To me calling a jitted function is an implicit command to compile the function (that's the whole point of jit after all). I would find it weird that the framework refuses to compile it for me when I ask to run the program (and compilation is a prerequisite to running it).

dime10: We might want to consult product if there is some uncertainty about desired behaviour here.

mehrdad2m: yeah I guess that's a fair point. Consulting product is a definitely good idea. I am fine with keepings things as before though for now if we don't want to block this PR since it's a rather low priority item.

doc/releases/changelog-dev.md
```diff
@@ -2,6 +2,55 @@
...
+  from catalyst.third_party.oqd import OQDDevice, OQDDevicePipeline, compile_to_artiq
+
+  OQD_PIPELINES = OQDDevicePipeline(
+      os.path.join("calibration_data", "device.toml"),
+      os.path.join("calibration_data", "qubit.toml"),
+      os.path.join("calibration_data", "gate.toml"),
+      os.path.join("device_db", "device_db.json"),
+  )
+
+  def test_rx_gate():
```
dime10: Changelog entry should not contain a pytest-style test, instead show the code directly.

doc/releases/changelog-dev.md
```diff
@@ -2,6 +2,55 @@
...
+      )
+      qml.capture.enable()
+
+      @qjit(pipelines=OQD_PIPELINES, target=llvmir)
+      @qml.qnode(oqd_dev)
+      def circuit():
+          qml.RX(0.5, wires=0)
+          return qml.counts(wires=0)
+
+      output_elf_path = compile_to_artiq(circuit, oqd_dev.artiq_config)
```
dime10: Are there any outputs to be observed?

frontend/test/test_oqd/oqd/test_oqd_device.py
```diff
@@ -69,6 +69,8 @@ def test_get_c_interface(self):
     def test_unsupported_one_shot_device(self):
         """Test unsupported device edge case."""
 
+        qml.capture.disable()
```
dime10: Was capture even enabled here?

frontend/catalyst/jit.py
```diff
@@ -78,6 +78,7 @@ def qjit(
     autograph_include=(),
     async_qnodes=False,
     target="binary",
+    link=True,
```
dime10: I thought we agreed to remove this as a user option

frontend/catalyst/jit.py
```diff
@@ -838,6 +850,12 @@ def run(self, args, kwargs):
         Returns:
             Any: results of the execution arranged into the original function's output PyTrees
         """
+        if self.compiled_function is None:
+            raise CompileError(
+                "Cannot execute function: no compiled function available. "
+                "The function must be compiled before execution."
+            )
```
dime10: I'm still not convinced of this error pattern. If a user saw this, what would they have to do to fix it?

dime10 (changes requested):

frontend/test/test_oqd/oqd/test_oqd_device.py
rniczh: I thought I removed it, let me try it again. So this test passed in my local device, but failed on CI, and just figured it out that this is due to the qml.capture be enabled on the CI (for this test)

dime10: Oh weird

frontend/catalyst/jit.py
rniczh: Just removed it in the latest revision, thanks!

doc/releases/changelog-dev.md
rniczh: Expected output is added: https://github.com/PennyLaneAI/catalyst/pull/2299/commits/5e48c1407bf190aca18a45485bd513caf11a5dd3

frontend/catalyst/jit.py
rniczh: I added one more error pattern for the case that llvm ir is selected as the target

https://github.com/PennyLaneAI/catalyst/pull/2299/changes#diff-648249b9fca379a495b52fc5c000fb7c4425daa46b6044df3ddf4515b3bc7b0aR850

EDIT: The fallback branch isn't reachable, so I removed that case and kept only the llvm ir check.
https://github.com/PennyLaneAI/catalyst/pull/2299/commits/40003bf0270ad92b78276ca223971dff181d99da

doc/releases/changelog-dev.md
rniczh: Done, thanks!

frontend/test/test_oqd/oqd/test_oqd_device.py
rniczh: Now, I probably know the reason. So now it works, I guess this is beause I originally used enable capture in oqd artiq test, but now I removed the enable capture in oqd artiq test, turned to use use_capture instead.

frontend/catalyst/jit.py
dime10: Thanks, this is relatively clean given the circumstances 👌

dime10 (approved):
Alright I'll approve the qjit updates (note I haven't reviewed the OQD portions) because the changes are kept fairly small, however there is change in contract / behaviour of QJIT class in the following ways:
- `compile` is no longer guaranteed to output a `CompiledFunction` (wrapping binary code) which it was guaranteed to do until now
- `__call__` / `jit_compile` will no longer force compilation of the function down to binary if the target is set as `"llvmir"`, as a result `run` will raise an error for this case
  - this is different than the other target levels (`"jaxpr"`, `"mlir"`) which only control aot compilation, while calling the function will force the full compilation pipeline so the program can be executed
  - once the "llvmir" target is set there is no way to compile the program for execution, except by mutating the compile options in place or defining a new qjit object

The inconsistency could be solved by limiting the change (e.g. the internal link option) to take effect in AOT mode only. Alternatively, one could expand the target option to affect both AOT and and JIT mode for all targets, but I still think this comes with downsides (see the last bullet point above).
However, the change in the contract structure of QJIT (main idea: QJIT is broken down into discrete stages, each of which produces well defined results that can feed into the next stage) might only be resolvable by breaking the `compile` step up such that we have separate compile and link stages.

If @mehrdad2m is happy with the current state we can merge this PR but I think these two points will be added as technical debt.

rniczh:
@mehrdad2m Sure, I reverted it in this commit: https://github.com/PennyLaneAI/catalyst/pull/2299/commits/67dba0996fdd989719a7c90b2e848facd0dd921d

frontend/catalyst/jit.py
```diff
@@ -820,10 +828,12 @@ def compile(self):
         else:
             shared_object, llvm_ir = self.compiler.run(self.mlir_module, self.workspace)
 
+        if not self.compile_options.link:
```
mehrdad2m, line 830: Wouldn't this still crash the `jit_compile` path if we have `target=llvmir`? 🤔

rniczh: Yeah, should remove those setting as well. So the current goal is to remove all LLVM IR specializations in the pipeline. For now, we're reverting to only adding the link setting. Since end-to-end is not possible without disabling link, when using OQD's end-to-end, we need to change to actively construct using QJIT.

Changed: https://github.com/PennyLaneAI/catalyst/pull/2299/commits/9d80502ae78899f46a79a5c8d1fa031bccd50258

Therefore, in terms of usage, we're changing the approach to first create a qnode with `@qnode`, then use `QJIT` with the current compile options to achieve this.

```python
oqd_pipelines = _get_oqd_pipelines()

@qml.set_shots(4)
@qml.qnode(oqd_dev)
def circuit():
    x = np.pi / 2
    qml.RX(x, wires=0)
    return qml.counts(all_outcomes=True)

# Get the LLVM IR
compiled_circuit = QJIT(circuit, CompileOptions(link=False, pipelines=oqd_pipelines))
llvm_ir = compiled_circuit.llvmir
```

mehrdad2m: Thanks @rniczh . I think this looks good for now. This could be improved if we divide `compile()` into `compile()` and `link()` like @dime10 suggested. But we don't need to address this in this PR.

rniczh:
> Did we remove a test for `compile_to_artiq`? Codecov is not happy about it, and it seems to be not used anywhere except in documentation.

We had the `ARM` target and `lld` installed for OQD pipeline at the beginning (in `mlir/CMakeLists.txt`, but after discussing with @paul0403, we decided to remove them (less intrusive). So that means we can't have the test for the `compile_to_artiq` function.

rniczh:
> > > Did we remove a test for `compile_to_artiq`? Codecov is not happy about it, and it seems to be not used anywhere except in documentation.
> > 
> > 
> > We had the `ARM` target and `lld` installed for OQD pipeline at the beginning (in `mlir/CMakeLists.txt`, but after discussing with @paul0403, we decided to remove them (less intrusive). So that means we can't have the test for the `compile_to_artiq` function.
> > EDIT: so this happend in this commit: [354ad5c](https://github.com/PennyLaneAI/catalyst/commit/354ad5ce2f4089522a15c0409650532895df2429)
> 
> Ah I see, thanks for the context. This makes me wonder if this file should live somewhere else like in the [oqd-catalyst-integration-test](https://github.com/OpenQuantumDesign/oqd-catalyst-integration-test/tree/main) repo. Especially that it is not relevant for any general user. Having a completely untested file like this make me a little uncomfortable. What do you think? We can maybe discuss this tomorrow in our standup.

As we discussed in offline. We removed the `compile_to_artiq` function from this PR and moved to the third-party OQD repo. So this PR only reponsible for generating llvm ir.

mehrdad2m (approved):
Nice! Thanks for the great work! 🚀

PR #2301, opened 2025-12-16
fix: update `get_mlir_module`

andrijapau:
[sc-105948]


jzaia18 (commented):
Looking good! Nice find on the error. Can you confirm the code snippet from before works with this?

andrijapau:
> Why not target `main`?

True, I can rebase.

PR #2305, opened 2025-12-16
fix: add `use_both_frontends` fixture to all tests

andrijapau:
This feature needs to work with both frontends (i.e. with and without capture). This unfortunately affected a few tests (tests were written too specific to the legacy frontend).

⚠️ Order of operations in the IR (if operators are "parallel") are not guaranteed to be the same with or without capture. That's why some tests needed to be updated so they are agnostic to the order they appear.

[sc-105774]

### DIFF
--- a/frontend/catalyst/python_interface/inspection/xdsl_conversion.py
+++ b/frontend/catalyst/python_interface/inspection/xdsl_conversion.py
@@ -177,7 +177,7 @@ def resolve_constant_params(ssa: SSAValue) -> float | int | str:
     op = ssa.owner
 
     if isinstance(op, Block):
-        arg_name = list(compress(op.args, map(lambda arg: arg is ssa, op.args)))[0]
+        arg_name = next(compress(op.args, map(lambda arg: arg is ssa, op.args)))
         return arg_name.name_hint
 
     if isinstance(op, TensorExtractOp):
@@ -277,7 +277,7 @@ def resolve_constant_wire(ssa: SSAValue) -> float | int | str:
     op = ssa.owner
 
     if isinstance(op, Block):
-        arg_name = list(compress(op.args, map(lambda arg: arg is ssa, op.args)))[0]
+        arg_name = next(compress(op.args, map(lambda arg: arg is ssa, op.args)))
         return arg_name.name_hint
 
     match op:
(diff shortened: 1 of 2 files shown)


frontend/test/pytest/python_interface/visualization/test_construct_circuit_dag.py
```diff
@@ -1275,23 +1349,40 @@ def my_workflow():
         # node0 -> NullQubit
 
         # Check all nodes
-        assert "PauliX" in nodes["node1"]["label"]
-        assert "PauliY" in nodes["node2"]["label"]
-        assert "PauliZ" in nodes["node3"]["label"]
-        assert "Hadamard" in nodes["node4"]["label"]
-        assert "S" in nodes["node5"]["label"]
-        assert "T" in nodes["node6"]["label"]
+        # NOTE: Order of nodes isn't the same with/without capture
```
sengthai, line 1352: what is the reason that the nodes are not the same for both cases? 🤔

andrijapau: I don't know the details of it. I think it just has to do with how it is captured 🤷🏼.

andrijapau: However, it is only "parallel" operators that are being shuffled around so it isn't a big deal :)

frontend/test/pytest/python_interface/visualization/test_construct_circuit_dag.py
```diff
@@ -1275,23 +1349,40 @@ def my_workflow():
...
+            assert "Hadamard" in nodes["node4"]["label"]
+            assert "S" in nodes["node5"]["label"]
+            assert "T" in nodes["node6"]["label"]
+        else:
+            assert "PauliX" in nodes["node1"]["label"]
+            assert "PauliY" in nodes["node2"]["label"]
+            assert "PauliZ" in nodes["node3"]["label"]
+            assert "Hadamard" in nodes["node4"]["label"]
+            assert "S" in nodes["node5"]["label"]
+            assert "T" in nodes["node6"]["label"]
```
sengthai: It seems only `PauliZ` and `PauliY` are parallel. So, how about hoisting the other out of the if-else?

```python
        assert "PauliX" in nodes["node1"]["label"]
        if qml.capture.enabled():
            assert "PauliZ" in nodes["node2"]["label"]
            assert "PauliY" in nodes["node3"]["label"]
        else:
            assert "PauliY" in nodes["node2"]["label"]
            assert "PauliZ" in nodes["node3"]["label"]
        assert "Hadamard" in nodes["node4"]["label"]
        assert "S" in nodes["node5"]["label"]
        assert "T" in nodes["node6"]["label"]
```

PR #2311, opened 2025-12-18
Recognize and Merge Permuted PPRs in merge-rotations Pass

kipawaa:
**Context:**
`merge-rotations` does not currently merge PPRs with permuted qubits+Pauli words for example

```llvm
%0:2 = qec.ppr ["X", "Y"](4) %q0, %q1
%1:2 = qec.ppr ["Y", "X"](4) %0#1, %0#0
```
even though these are compatible PPRs.

Merges PPRs with permuted qubits+Pauli words. The example above would be merged into
```llvm
%0:2 = qec.ppr ["Y", "X"](2) %q1, %q0
```

**Benefits:**
`merge-rotations` functionality is expanded.

[[sc-106063]]

### DIFF
--- a/mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
+++ b/mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
@@ -382,6 +382,11 @@ struct MergePPRRewritePattern : public OpRewritePattern<OpType> {
             return failure();
         }
 
+        // same number of non-identity qubits
+        if (opNonIdentityIndices.size() != getNonIdentityIndices(parentOp).size()) {
+            return failure();
+        }
+
         // When two rotations have permuted Pauli strings, we can still merge them, we just need to
         // correctly re-map the inputs. This map stores the index of a qubit in parentOp's out
         // qubits at the index it appears in op's in qubits.
--- a/mlir/test/Quantum/MergeRotationsTest.mlir
+++ b/mlir/test/Quantum/MergeRotationsTest.mlir
@@ -1049,6 +1049,15 @@ func.func public @identity_agnostic_sizing(%q0: !quantum.bit, %q1: !quantum.bit,
     func.return
 }
 
+func.func public @dont_merge_superset(%q0: !quantum.bit, %q1: !quantum.bit) {
+    // CHECK: qec.ppr ["X", "Y"](4)
+    // CHECK: qec.ppr ["X"](4)
+    %0:2 = qec.ppr ["X", "Y"](4) %q0, %q1 : !quantum.bit, !quantum.bit
+    %1 = qec.ppr ["X"](4) %0#0 : !quantum.bit
+
+    func.return
+}
+
 
 // Arbitrary Angle PPR Tests
 
@@ -1327,3 +1336,4 @@ func.func public @merge_const_var(%q0: !quantum.bit, %0: f64) {
     %3 = qec.ppr.arbitrary ["Z"](%1) %2: !quantum.bit
     func.return
 }
+


mlir/test/Quantum/MergeRotationsTest.mlir
sengthai, line 1: I think we should add a test like this one as well. 
```MLIR
qec.ppr ["Z", "I", "Y"](4) Q0, Q1, Q2
qec.ppr ["Y", "Z", "I"](4) Q2, Q0, Q3
```

sengthai (approved):
Well done! 👍

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
```diff
@@ -321,105 +321,85 @@ struct MergeRotationsRewritePattern : public OpRewritePattern<OpType> {
...
 
             return success();
         }
 
-        if (opRotation != prevOpRotation) {
-            return failure();
-        }
+        auto mergeOp = rewriter.create<PPRotationOp>(loc, parentOp.getOutQubits().getTypes(),
+                                                     op.getPauliProduct(), newAngle, newInQubits,
+                                                     op.getCondition());
```
sengthai: Nice! Alternatively, we also can build using this constructor:

```C++
auto mergeOp = rewriter.create<PPRotationOp>(loc, op.getPauliProduct(), newAngle, newInQubits);
```

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
```diff
@@ -321,105 +321,85 @@ struct MergeRotationsRewritePattern : public OpRewritePattern<OpType> {
...
-            rewriter.replaceOp(op, prevOp);
-        }
-        else {
-            ValueRange originalQubits = prevOp.getInQubits();
-            rewriter.replaceOp(op, originalQubits);
-            rewriter.eraseOp(prevOp);
-        }
+        auto mergeOp = rewriter.create<PPRotationArbitraryOp>(
+            loc, parentOp.getOutQubits().getTypes(), op.getPauliProduct(), newAngle, newInQubits,
+            op.getCondition());
```
sengthai: Same here:
```CPP
        auto mergeOp = rewriter.create<PPRotationArbitraryOp>(loc, op.getPauliProduct(), newAngle, newInQubits);
```

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
kipawaa: does this consider conditions?

sengthai: you are right. we need to consider condition as well

mlir/test/Quantum/MergeRotationsTest.mlir
```diff
@@ -943,6 +943,63 @@ func.func public @dont_merge_conditionals(%q1: !quantum.bit, %q2: !quantum.bit,
...
+// -----
+
+// check permutations with duplicate Pauli symbols
+
+// CHECK-LABEL: merge_permutations_with_duplicates
+func.func public @merge_permutations_with_duplicates(%q0: !quantum.bit, %q1: !quantum.bit, %q2: !quantum.bit) {
+    // CHECK: ([[q0:%.+]]: !quantum.bit, [[q1:%.+]]: !quantum.bit, [[q2:%.+]]: !quantum.bit)
+    // CHECK-NOT: qec.ppr ["X", "Y", "X"]
+    // CHECK-NOT: qec.ppr ["Y", "X", "X"](8)
+    // CHECK: qec.ppr ["Y", "X", "X"](4) [[q1]], [[q2]], [[q0]]
```
dime10, line 968: Might be worth checking that the results appear in the expected order as well?  For example in the return op

kipawaa: I'm not sure I understand exactly, do you mean check that the positions of the results of the operation correspond correctly to the op's representation of the Pauli word like in [this test](https://github.com/PennyLaneAI/catalyst/pull/2311/commits/4dee6e675e978b358c48d41d56f96e6666c83047)?

dime10: Not necessarily the pauli word, but since we're replacing the original op, there will be new values created in the IR. Right now the test doesn't have any uses on the op that's replaced, so it would be good to add a user (like the return op), and then check that the values have been replaced correctly, in the intended order (especially since this is a transform that reorders values, even though I can see it's only the inputs that are  -intended- to be reordered).

kipawaa: [da00eb6](https://github.com/PennyLaneAI/catalyst/pull/2311/commits/da00eb61a83e489b6833e47acf1cd2e67bc3e595) introduces a test where the return op uses the results of the merged op. Since the merged op should always adopt the ordering of the last ppr of the pprs being merged, the order of the results should be preserved. Does this seem appropriate?

mlir/test/Quantum/MergeRotationsTest.mlir
kipawaa: Good catch, I hadn't considered this case! ~Originally added in [e4169fb](https://github.com/PennyLaneAI/catalyst/pull/2311/commits/e4169fbdcc972c5dee5dc3f6584248b2ae579bf8) but it got modified a bit afterwards due to codefactor issues.~ I made a mistake in the test, fixing now

kipawaa: Should be correct as of [4feb1a6](https://github.com/PennyLaneAI/catalyst/pull/2311/commits/4feb1a654e4cfc1239d7a09bed6f4b55a7882893).

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
```diff
@@ -321,154 +321,146 @@ struct MergeRotationsRewritePattern : public OpRewritePattern<OpType> {
...
         ValueRange opInQubits = op.getInQubits();
+        ValueRange opOutQubits = op.getOutQubits();
+        ArrayAttr opPauliProduct = op.getPauliProduct();
+        Value opCondition = op.getCondition();
+        SmallVector<int16_t> opNonIdentityIndices =
+            getNonIdentityIndicesAndReplaceIdentityQubitUses(op, rewriter);
 
-        Operation *definingOp = opInQubits[0].getDefiningOp();
+        // get parent op
+        Operation *definingOp = opInQubits[opNonIdentityIndices[0]].getDefiningOp();
```
sengthai, line 370: If all Pauli words are "I", then `opNonIdentityIndices` will be empty. Therefore, it will cause an out-of-bounds index error.

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
```diff
@@ -321,154 +321,146 @@ struct MergeRotationsRewritePattern : public OpRewritePattern<OpType> {
...
-            if (qubit.getDefiningOp() != parentOp) {
-                return failure();
-            }
-        }
-        ValueRange parentOpOutQubits = parentOp.getOutQubits();
-        if (parentOpOutQubits.size() != opInQubits.size()) {
+        // collect info on parent op
+        ValueRange parentOpInQubits = parentOp.getInQubits();
+        ArrayAttr parentOpPauliProduct = parentOp.getPauliProduct();
+        SmallVector<int16_t> parentOpNonIdentityIndices =
```
sengthai: I wonder if this `parentOpNonIdentityIndices` variable will be used in the code?

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
kipawaa: That makes sense. fixed in c7094ce12 (leaving the cleanup to canonicalization).

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
kipawaa: Good point, the variable isn't used but the call is necessary to replace the uses of the identity qubits. Removed the variable in 0921ea198

mlir/test/Quantum/MergeRotationsTest.mlir
dime10: Looks great!

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
```diff
@@ -321,154 +321,148 @@ struct MergeRotationsRewritePattern : public OpRewritePattern<OpType> {
...
-        if (parentOpOutQubits.size() != opInQubits.size()) {
+        // collect info on parent op
+        ValueRange parentOpInQubits = parentOp.getInQubits();
+        ArrayAttr parentOpPauliProduct = parentOp.getPauliProduct();
+        getNonIdentityIndicesAndReplaceIdentityQubitUses(parentOp, rewriter);
+
+        // check compatible conditionals
+        if (opCondition != parentOp.getCondition()) {
             return failure();
         }
```
dime10: Apparently this is illegal MLIR code, all IR mutations must come after a guaranteed match success:
https://mlir.llvm.org/docs/PatternRewriter/#matchandrewrite-implementation

Don't know how serious it is to violate it, but we probably shouldn't

kipawaa: That definitely make sense, I could see that leading to some nasty bugs! Amended in 592a50c0b. That helper function probably should have been two separate helpers anyways 😅

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
```diff
@@ -321,154 +321,154 @@ struct MergeRotationsRewritePattern : public OpRewritePattern<OpType> {
...
+        Operation *definingOp = opInQubits[opNonIdentityIndices[0]].getDefiningOp();
         if (!definingOp) {
             return failure();
         }
-
-        auto parentOp = dyn_cast<PPRotationArbitraryOp>(definingOp);
+        auto parentOp = dyn_cast<ParentOpType>(definingOp);
         if (!parentOp) {
             return failure();
         }
```
dime10, line 374: Little trick: `dyn_cast_or_null` saves you one branch since it propagates null values through

mlir/lib/Quantum/Transforms/MergeRotationsPatterns.cpp
```diff
@@ -321,154 +321,149 @@ struct MergeRotationsRewritePattern : public OpRewritePattern<OpType> {
...
+        SmallVector<Attribute> newPauliProductItems;
+        for (int16_t i : opNonIdentityIndices) {
+            // check Pauli + qubit pairings
+            if (opPauliProduct[i] != parentOpPauliProduct[inversePermutation[i]]) {
+                return failure();
+            }
+
+            newInQubitsItems.push_back(parentOpInQubits[inversePermutation[i]]);
+            newPauliProductItems.push_back(parentOpPauliProduct[inversePermutation[i]]);
         }
```
dime10, line 420: Maybe the code below wouldn't allow this or error out, but I don't think we are checking that the Pauli words are actually the same? Only that the child is present in the parent, so the parent's Pauli word could be a superset. Maybe I missed it though 🤔

kipawaa: I think you're right, sorry I missed that! I've (re-)added an explicit size check in 4bf819565, and a test for this

dime10 (approved):

PR #2315, opened 2025-12-19
Support for new arbitrary PPR and PauliRot in `mlir_specs` and enhance the specs collector

sengthai:
Here we support handling `qec.ppr.arbitrary` and `quantum.paulirot` ops using `mlir_specs`. Also, enhance the analysis of the estimated count in the for-loop.

### DIFF
--- a/frontend/test/pytest/python_interface/inspection/test_mlir_specs.py
+++ b/frontend/test/pytest/python_interface/inspection/test_mlir_specs.py
@@ -258,6 +258,39 @@ class TestMLIRSpecs:
         ):
             mlir_specs(simple_circuit, level=[0, 3])
 
+    def test_not_qnode(self):
+        """Test that a malformed QNode raises an error."""
+
+        def not_a_qnode():
+            pass
+
+        with pytest.raises(
+            ValueError,
+            match="The provided `qnode` argument does not appear to be a valid QJIT "
+            "compiled QNode.",
+        ):
+            mlir_specs(not_a_qnode, level=0)
+
+    def test_malformed_qnode(self):
+        """Test that a QNode without measurements can still be collected."""
+
+        dev = qml.device("lightning.qubit", wires=1)
+
+        @qml.qjit
+        @qml.qnode(dev)
+        def circ(wire):
+            qml.X(0)
+            qml.X(wire)
+
+        res = mlir_specs(circ, 0, 1)
+        expected = make_static_resources(
+            operations={"PauliX": {1: 2}},
+            measurements={},
+            num_allocs=1,
+        )
+
+        assert resources_equal(res, expected)
+
     @pytest.mark.parametrize(
         "pl_ctrl_flow, iters, autograph",
         [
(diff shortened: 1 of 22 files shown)


frontend/test/pytest/python_interface/inspection/test_mlir_specs.py
```diff
@@ -36,7 +36,10 @@ def resources_equal(
 
         # actual.device_name == expected.device_name TODO: Don't worry about this one for now
         assert actual.num_allocs == expected.num_allocs
-
+        print("-------------------------------- actual --------------------------------")
+        print(actual.operations)
+        print("-------------------------------- expected --------------------------------")
+        print(expected.operations)
```
jzaia18:
```suggestion
```

jzaia18 (approved):
Looks good, just 1 comment

frontend/catalyst/python_interface/inspection/specs_collector.py
```diff
@@ -260,6 +270,11 @@ def _(xdsl_op: PPRotationOp) -> tuple[ResourceType, str]:
     return ResourceType.QEC, s
 
 
+@handle_resource.register
+def _(xdsl_op: PPRotationArbitraryOp) -> tuple[ResourceType, str]:
+    return ResourceType.QEC, "PPR-arbitrary"
```
dime10: Suggestion: This "arbitrary" naming is somewhat suboptimal (in clarity, and it's unwieldy). For the reporting should we do something like this? 😃 (unless this name needs to be usable as a dictionary key)
```suggestion
    return ResourceType.QEC, "PPR(ϕ)"
```

frontend/catalyst/python_interface/inspection/specs_collector.py
```diff
@@ -430,19 +446,23 @@ def _collect_region(
                 cond_warning=cond_warning,
                 adjoint_mode=adjoint_mode,
             )
-            try:
-                iters = count_static_loop_iterations(op)
+            if not "estimated_iterations" in op.attributes:
```
dime10: nitpick: In general, I would favour the unnegated version of the check since it's simpler, and the attribute case seems like the special case here (which means adding more special cases is more complicated since you have to negate all those cases, whereas if you don't negate the special cases it's as easy as adding another branch).

frontend/catalyst/python_interface/inspection/specs_collector.py
jzaia18: The name is used as a dictionary-key in the main specs output. Maybe "PPR-phi" if you wanted to do something like this?

dime10: ah yeah that's though then 😅

sengthai: Thanks for pointing that out! Then, since it used a dic-key, I will use `PPR-Phi` instead.

frontend/catalyst/python_interface/inspection/specs_collector.py
sengthai: I absolutely agree with that!

PR #2316, opened 2025-12-19
Translate and lower the jvp

albi3ro:

Now that https://github.com/PennyLaneAI/pennylane/pull/8788 is in, we need to be able to translate and lower the jvp primitive.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-106503]


frontend/catalyst/api_extensions/differentiation.py
```diff
@@ -21,7 +21,7 @@
 import copy
 import functools
 import numbers
-from typing import Callable, Iterable, List, Optional, Union
+from typing import Callable, List, Optional, Sequence, Union
```
mudit2812: This is not really related to your PR, but `Callable` and `Sequence` should be getting imported from `collections.abc`, and we shouldn't be importing `typing.List` at all since lower-case `list` can be used as type hints

frontend/test/pytest/test_adjoint.py
```diff
@@ -888,7 +888,7 @@ def test_queue_category(self):
     def test_queue_category_None(self):
         """Test that the queue category `None` for some observables carries over."""
         op = adjoint(qml.Hermitian([[1, 0], [0, 1]], wires=0))
-        assert op._queue_category is None  # pylint: disable=protected-access
+        assert op._queue_category == "_ops"  # pylint: disable=protected-access
```
mudit2812: This is because of the change that David W. added?

mudit2812 (approved):

frontend/catalyst/api_extensions/differentiation.py
```diff
@@ -442,12 +442,12 @@ def workflow(primals, tangents):
...
-        if not isinstance(x, Iterable):
-            raise ValueError(f"vjp '{hint}' argument must be an iterable, not {type(x)}")
+    def check_is_Sequence(x, hint):
+        if not isinstance(x, Sequence):
+            raise ValueError(f"vjp '{hint}' argument must be a Sequence, not {type(x)}")
 
-    check_is_iterable(params, "params")
-    check_is_iterable(tangents, "tangents")
+    check_is_Sequence(params, "params")
+    check_is_Sequence(tangents, "tangents")
```
dwierichs: I think the naming convention for functions overrules the naming convention for objects whose name is part of the function name, so I then the `S` should be lowercase instead.

frontend/catalyst/api_extensions/differentiation.py
```diff
@@ -442,12 +442,12 @@ def workflow(primals, tangents):
     (Array(0.78766064, dtype=float64), Array(-0.7011436, dtype=float64))
     """
 
-    def check_is_iterable(x, hint):
-        if not isinstance(x, Iterable):
-            raise ValueError(f"vjp '{hint}' argument must be an iterable, not {type(x)}")
+    def check_is_Sequence(x, hint):
+        if not isinstance(x, Sequence):
+            raise ValueError(f"vjp '{hint}' argument must be a Sequence, not {type(x)}")
```
dwierichs:
```suggestion
            raise ValueError(f"jvp '{hint}' argument must be a Sequence, not {type(x)}")
```

frontend/catalyst/from_plxpr/from_plxpr.py
dwierichs:
```suggestion
    """Translate a jvp equation."""
```

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -252,6 +253,18 @@ def handle_vjp(self, *args, jaxpr, **kwargs):
     return pl_vjp_prim.bind(*new_args, jaxpr=new_j, **kwargs)
 
 
+@WorkflowInterpreter.register_primitive(pl_jvp_prim)
+def handle_jvp(self, *args, jaxpr, **kwargs):
+    """Translate a grad equation."""
```
dwierichs:
```suggestion
    """Translate a vjp equation."""
```

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -252,6 +253,18 @@ def handle_vjp(self, *args, jaxpr, **kwargs):
...
+    f = partial(copy(self).eval, jaxpr, [])
+    new_jaxpr = jax.make_jaxpr(f)(*args[: len(jaxpr.invars)])
+
+    new_args = (*new_jaxpr.consts, *args)
+    j = new_jaxpr.jaxpr
+    new_j = j.replace(constvars=(), invars=j.constvars + j.invars)
```
dwierichs, line 264: There is quite similar logic [in the PL primitive](https://github.com/PennyLaneAI/pennylane/blob/a78becfba8a612fbd7a66a60cc3a53bb9859793a/pennylane/_grad/jvp.py#L117). Probably it's not worth the hassle, but just want to confirm that we can't unify this logic/have it in one place?

frontend/catalyst/jax_primitives.py
```diff
@@ -882,6 +883,40 @@ def _jvp_lowering(ctx, *args, jaxpr, fn, grad_params):
     ).results
 
 
+def _capture_jvp_lowering(ctx, *args, jaxpr, fn, method, argnums, h):
```
dwierichs, line 886: I can't speak much to this implementation, so my review will not be very thorough for this part of the stack, sorry 😬

frontend/test/pytest/test_adjoint.py
dwierichs: Fixed in #2465

frontend/test/pytest/test_jvpvjp.py
```diff
@@ -62,7 +62,7 @@ def f(x):
     ct = jnp.array(1.0)
     res, f_vjp = jax.vjp(f, x)
     expected = tuple([res, f_vjp(ct)])
-    result = C_vjp(f, x, ct)
+    result = C_vjp(f, (x,), ct)
```
dwierichs, line 65: How was this found? Why do we change it here?

frontend/test/pytest/test_jvpvjp.py
```diff
@@ -183,9 +184,10 @@ def J_workflow():
     res_jax, tree_jax = jax.tree_util.tree_flatten(r1)
     res_cat, tree_cat = jax.tree_util.tree_flatten(r2)
     assert tree_jax == tree_cat
-    assert_allclose(res_jax, res_cat)
+    assert_allclose(res_jax, res_cat, atol=5e-7)
```
dwierichs, line 187: Does the precision differ per frontend, or why do we have to add this here? :o

frontend/test/pytest/test_jvpvjp.py
```diff
@@ -396,15 +405,15 @@ def f(x):
...
+        return qml.jvp(f, [0.3], [1.1], method=diff_method, argnums=[0])
 
     catalyst_res = workflow()
     jax_res = J_jvp(f, [0.3], [1.1])
 
     catalyst_res_flatten, tree_cat = jax.tree_util.tree_flatten(catalyst_res)
     jax_res_flatten, tree_jax = jax.tree_util.tree_flatten(jax_res)
     assert tree_cat == tree_jax
-    assert_allclose(catalyst_res_flatten, jax_res_flatten, rtol=1e-6)
+    assert_allclose(catalyst_res_flatten, jax_res_flatten, rtol=1e-6, atol=5e-6)
```
dwierichs, line 416: Same question as above re: frontend-dependent precision.

dwierichs (changes requested):
Just some very small comments/questions, looks almost good to go from my side 👍 🎉

frontend/catalyst/from_plxpr/from_plxpr.py
albi3ro: I think extracting these couple simple lines will just make it more complicated, since then you have to go track down the definition.  If these functions start getting too complicated, it's an option, but I think they are still simple enough.

frontend/test/pytest/test_jvpvjp.py
albi3ro: Something like `jnp.array([0, 1])` is an `Iterable` but not a `Sequence`.  I changed it since what we want for the inputs is really a `Sequence`.  We want it to have a fixed order, length, and be indexable. 

For example, `jnp.array(4.0)` does not have a length and is not indexable. I think this just cleans up a rather ambiguous interface.

frontend/test/pytest/test_jvpvjp.py
albi3ro: It's small enough I'm just chalking it up to stochastisticity. Might be to a different execution pathway when capture is enabled but not using qjit.

frontend/test/pytest/test_jvpvjp.py
albi3ro: Probably capture-execution slight differences.

frontend/catalyst/jax_primitives.py
albi3ro: I've been taking a rather practical approach of just seeing if it works or not. But basically this is just slight modifications to the the `_jvp_lowering` for the catalyst version.

frontend/test/pytest/test_jvpvjp.py
dwierichs: Nice 👍 So if we didn't make this change, the validation that you are modifying in this PR would error out? Then I'm happy :)

albi3ro: Yep.  And with program capture we are needing it to be a full `Sequence` as well, since we are doing `*params`.

frontend/test/pytest/test_jvpvjp.py
dwierichs: Ahh, it's using finite diff sometimes, got it.

frontend/catalyst/api_extensions/differentiation.py
albi3ro: done locally.

frontend/catalyst/api_extensions/differentiation.py
albi3ro: done locally.

frontend/catalyst/api_extensions/differentiation.py
albi3ro: fixed locally.

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -242,7 +243,7 @@ def handle_grad(self, *args, jaxpr, n_consts, **kwargs):
 
 @WorkflowInterpreter.register_primitive(pl_vjp_prim)
 def handle_vjp(self, *args, jaxpr, **kwargs):
-    """Translate a grad equation."""
+    """Translate a jvp equation."""
```
dwierichs:
```suggestion
    """Translate a vjp equation."""
```

dwierichs (approved):
Nice, thanks @albi3ro 💯

PR #2320, opened 2025-12-21
[qref 1] Add reference semantics quantum dialect definition

paul0403:
**Context:**
There will be 3 PRs on reference semantics quantum dialect.
1. This one. It adds the dialect and ops themselves.
2. A conversion from reference semantics quantum dialect to value semantics quantum dialect (aka the "regular" quantum dialect)
3. A conversion from value dialect to reference dialect. 

Ideally there will be a fourth PR:
4. Lowering rules from plxpr primitives to reference semantics quantum dialect. 

But depending on time/complexity/workload/priority, we might need to drop some of them this quarter.

Like the value semantics dialect, we "extract" a qubit from an allocated register and an index. This is called the `GetOp` in the reference semantics dialect.
Unlike qubit SSA values, a qubit reference SSA value is allowed to be used multiple times.
The operands of gates and measurements will be these qubit reference values.

For example, in the following circuit, gates and observable ops take in the qubit reference they're acting on, and do not produce qubit values.
Notice that qubit reference values are reusable.
```mlir
func.func @expval_circuit() -> f64 {
    %a = qref.alloc(2) : !qref.reg<2>
    %q0 = qref.get %a[0] : !qref.reg<2> -> !qref.bit
    %q1 = qref.get %a[1] : !qref.reg<2> -> !qref.bit
    qref.custom "Hadamard"() %q0 : !qref.bit
    qref.custom "CNOT"() %q0, %q1 : !qref.bit, !qref.bit
    qref.custom "Hadamard"() %q0 : !qref.bit
    %obs = qref.namedobs %q1 [ PauliX] : !quantum.obs
    %expval = quantum.expval %obs : f64
    qref.dealloc %a : !qref.reg<2>
    return %expval : f64
}
```

We can also do this with loops. Notice, that the loop does not have to take in the "global qreg value" anymore. 

```mlir
func.func @circuit_with_loop(%nqubits: i64) -> f64 {
    %a = qref.alloc(%nqubits) : !qref.reg<?>

    %start = arith.constant 0 : index
    %step = arith.constant 1 : index
    %stop = index.casts %nqubits : i64 to index

    scf.for %i = %start to %stop step %step {
        %int = index.casts %i : index to i64
        %this_q = qref.get %a[%int] : !qref.reg<?>, i64 -> !qref.bit
        qref.custom "Hadamard"() %this_q : !qref.bit
        scf.yield
    }

    %q0 = qref.get %a[0] : !qref.reg<?> -> !qref.bit
    %obs = qref.namedobs %q0 [ PauliX] : !quantum.obs
    %expval = quantum.expval %obs : f64
    qref.dealloc %a : !qref.reg<?>
    return %expval : f64
}
```

Here's the list of all operations in the regular quantum dialect that uses `QubitType` or `QuregType` values as inputs/outputs:
- [x] AllocOp 
- [x] DeallocOp
- [x] AllocQubitOp 
- [x] DeallocQubitOp
- [x] ExtractOp: reference semantics counterpart is `GetOp`
- [x] InsertOp (not meaningful in reference semantics, no counterpart needed)
- [x] SetStateOp
- [x] SetBasisStateOp
- [x] CustomOp
- [x] PauliRotOp
- [x] GlobalPhaseOp
- [x] MultiRZOp
- [x] PCPhaseOp
- [x] QubitUnitaryOp
- [x] ComputationalBasisOp
- [x] NamedObsOp
- [x] HermitianOp
- [x] MeasureOp
- [x] AdjointOp
- [x] YieldOp: no counterpart in reference semantics, as quantum region ops do not return new qreg values

[sc-105532]


mlir/include/CMakeLists.txt
```diff
@@ -7,5 +7,6 @@ add_subdirectory(Mitigation)
 add_subdirectory(PauliFrame)
 add_subdirectory(QEC)
 add_subdirectory(Quantum)
+add_subdirectory(RefQuantum)
```
dime10: Just a note on naming, I would put quantum first before the qualifier, because that is the primary identity which you typically would want to know first. That way the two dialects should also always appear together

mlir/include/RefQuantum/IR/RefQuantumOps.td
```diff
@@ -0,0 +1,368 @@
...
+//===----------------------------------------------------------------------===//
+// RefQuantum dialect operations.
+//===----------------------------------------------------------------------===//
+
+// -----
+
+class Memory_Op<string mnemonic, list<Trait> traits = []> : RefQuantum_Op<mnemonic, traits>;
+
+// TODO: (dynamic) alloc and deallocs
+// Note that extracts and inserts are not a thing in reference semantics
```
dime10: That's interesting question, I don't know the conclusion is necessarily obvious and might warrant some discussion.

1) There is a few different ways we could design the reference semantic dialect. 
    a. Right now it seems like operations directly take in integer labels to reference qubits. While that is what the PennyLane _language_ does, I'm not sure I'm a fan for a _compiler_ IR. Even with reference semantics, I think qubit references should still come from a resource allocation, rather than being side-effected into existence by merely appearing in an operation.
    b. One alternative is what we see in [this paper](https://arxiv.org/pdf/2101.11030) where registers are allocated and the register reference is passed to operations alongside a (static or dynamic) indexing value:
    image
    Advantage here is that the index is always accessible straight from the op.
    c. To reduce complexity in the gate ops, one could instead use extract operations for registers, and/or single qubit allocs, in order to pass qubit references into operations instead. (In fact I believe the ops in the paper support both b. and c.). Example:
    ```mlir
    %q0 = q.alloc
    %r = q.allocreg 5
    %q2 = q.extract %r[1]
    q.CX %q0, %q1
    ```
    d. Maybe others..

2) The other thing worth thinking about is how these designs will interact with the new register features we are planning, such that qubits can be moved around / swapped with respect to the algorithmic indices, and also combined into arbitrary collections more easily.

joeycarter:
Amazing! :star_struck: 

I had a few quick thoughts on the IR (mostly on naming conventions) based on a flyby reading of the example:

1. Could we rename the dialect from `ref_quantum` to something like `qref`? For brevity and also to mirror the MLIR `memref` dialect.
2. Would it make sense to include the size of the allocated resource in the type (in a similar way that memref does it)? e.g.
    ```mlir
    %a = qref.alloc() : !qref.qref<2>
    ```
    for static allocs and
    ```mlir
    %d = arith.constant 2 : i64
    %a = qref.alloc(%d) : !qref.qref<?>
    ```
    for dynamic allocs.
3. `ref_quantum.make_reference` sounds similar to [`memref.load`](https://mlir.llvm.org/docs/Dialects/MemRef/#memrefload-memrefloadop). Maybe for brevity we could call it `qref.load`?
4. Ignore this one if you think the names will clash, but could we rename the type `!ref_quantum.qubit_ref` to `!qref.bit`? Or similar. This is probably beyond the scope of this PR to add them here, but we have more qubits types now in the main quantum dialect :blush:

mlir/include/RefQuantum/Transforms/Patterns.h
```diff
@@ -0,0 +1,25 @@
...
+// limitations under the License.
+
+#pragma once
+
+#include "mlir/IR/PatternMatch.h"
+
+namespace catalyst {
+namespace ref_quantum {
+
+void populateRQHelloWorldPatterns(mlir::RewritePatternSet &patterns);
```
paul0403: Reminder to self: remove pass boilerplate

paul0403:
> > > ref_quantum.make_reference sounds similar to [memref.load](https://mlir.llvm.org/docs/Dialects/MemRef/#memrefload-memrefloadop). Maybe for brevity we could call it qref.load?
> > 
> > 
> > So I was explicitly trying to avoid the word "load", because it too heavily implies something about memory, whereas we are not dealing with that aspect at all. The `make_reference` op is solely just for creating a SSA value that represents a qubit reference. It has no memory cannotations whatsoever. Also, "load" seems to suggest that "I am reading a value from a memory location", but the result SSA value of this op is explicitly not a qubit value!
> > I did think about following memref's structure, i.e. `ref_quantum.alloc` will allocate a "qref", and we can index individual qubit addresses into the qref by `ref_quantum.load` (which is essentially a getelementptr op in llvm), but in the end I decided to steer the wording away from anything memory-related.
> 
> Got it. So just to confirm, with memrefs, when we have e.g.
> 
> ```mlir
> %buffer = memref.alloc() : memref<8xf32>
> %val = memref.load %buffer[0] : memref<8xf32>
> ```
> 
> then `%val` is an SSA value of type `f32` (i.e. the element type of the memref), but with the quantum reference semantics dialect,
> 
> ```mlir
> %0 = qref.alloc() : !qref.qref<8>
> %1 = qref.make_reference %0[0] : !qref.qref<8> -> !qref.qubit_ref
> ```
> 
> then `%1` is **not** the `!quantum.bit` value loaded from memory, but rather just a reference to it.
> 
> > > Ignore this one if you think the names will clash, but could we rename the type !ref_quantum.qubit_ref to !qref.bit? Or similar. This is probably beyond the scope of this PR to add them here, but we have more qubits types now in the main quantum dialect 😊
> > 
> > 
> > Hmm, I want the type to explicitly indicate that this is a reference 🤔
> 
> I guess this is where the design philosophy of `qref` and `memref` diverge. While `memref` works more like a buffer of values, `qref` works more like a buffer of references (if I've understood correctly). In that case `!qref.qubit_ref` is probably fine.

Exactly! In my mind I don't really think of qubit SSA values like regular classical values (aka numbers). I've always viewed qubit values to essentially represent an "in-between quantum state" between 2 gates, which is not really load/write-able to and from memory due to no cloning theorem. Really, qubit SSA values are just there in the value semantics IR to chain up def-use of gates. 

So when we say "a reference to a qubit" here, what we really mean is a qubit **address**, or in plain PL wording, the "wire label". A SSA value of type `!qref.qubit_ref` is more like an address, or a pointer, although I really, really do not want to use these two words in the IR!

Or, it might help to think of them like FPGA: all the registers have their own "name", and gates just take in the name of their target registers. Essentially, that is what I'm going for here. The `make_reference`op is my way to generate a "name" for a "register".

paul0403:
In other words, while `memref.alloc` is allocating a segment of memory to hold some classical values (numbers), and returning a reference to that segment of memory as its op result, our `qref.alloc` is allocating a bunch of qubit addresses, instead of quantum values (e.g. state vector between two gates,etc; although, again, the `quantum.bit` value in core quantum dialect are only symbolic, and does not represent any concrete program values during execution, so my analogy is just a mental model).

mlir/include/QRef/IR/QRefInterfaces.td
```diff
@@ -0,0 +1,163 @@
...
+
+    let methods = [
+        InterfaceMethod<
+            "Return all operands which are considered gate parameters.",
+            "mlir::ValueRange", "getAllParams"
+        >,
+        InterfaceMethod<
+            "Return the param operand at the requested index.",
+            "mlir::Value", "getParam", (ins "size_t":$idx), /*methodBody=*/[{}],
+            /*defaultImplementation=*/[{ return mlir::cast<ConcreteOp>($_op).getAllParams()[idx]; }]
```
rniczh: It would be safer to validate the `[idx]` and raise an error properly? rather than has an out-of-bound access.

mlir/include/QRef/IR/QRefInterfaces.td
```diff
@@ -0,0 +1,163 @@
...
+        The actions do not have to be unitary. For example, the SetState operations also falls
+        under this class.
+    }];
+
+    let cppNamespace = "::catalyst::qref";
+
+    let methods = [
+        InterfaceMethod<
+            "Return all operands which are considered input qubits (including controls).",
+            "std::vector<mlir::Value>", "getQubitOperands"
```
rniczh, line 60: Just relalized we return the `std::vector` in quantum ops instead of `ValueRange`. Is there any reason for us to use the `std::vector` as return type but not `ValueRange`?

mlir/include/QRef/IR/QRefOps.td
```diff
@@ -0,0 +1,655 @@
...
+}
+
+def PauliRotOp : UnitaryGate_Op<"paulirot", [ParametrizedGate, AttrSizedOperandSegments]> {
+    let summary = "Apply a Pauli Product Rotation";
+    let description = [{
+        The `qref.paulirot` operation applies a rotation around a Pauli product
+        operator to the state-vector.
+        The arguments are the rotation angle `angle`, a string representing the
+        Pauli product operator, and a set of qubits the operation acts on.
+        Note that this operation is currently not excutable. There isn't a valid
```
rniczh: typo

```suggestion
        Note that this operation is currently not executable. There isn't a valid
```

mlir/include/QRef/IR/QRefOps.td
```diff
@@ -0,0 +1,655 @@
...
+        QRef_Op<mnemonic, traits # [Pure]>;
+
+def ComputationalBasisOp : Observable_Op<"compbasis", [AttrSizedOperandSegments]> {
+    let summary = "Define a pseudo-observable of the computational basis for use in measurements";
+    let description = [{
+        The `qref.compbasis` operation defines a quantum observable to be used by other
+        operations such as measurement processes. The specific observable defined here is a
+        "pseudo" observable to represent measurements in the computational basis.
+        The only arguments are either the list of qubit references to measure, or the quantum
+        register refrence to measure. When the argument is a register, the measurement occurs
```
rniczh: typo

```suggestion
        register reference to measure. When the argument is a register, the measurement occurs
```

mlir/include/QRef/IR/QRefOps.td
```diff
@@ -0,0 +1,655 @@
...
+            return getAdjoint();
+        }
+
+        void setAdjointFlag(bool adjoint) {
+            if (adjoint) {
+               (*this)->setAttr("adjoint", mlir::UnitAttr::get(this->getContext()));
+            } else {
+               (*this)->removeAttr("adjoint");
+            }
+        };
```
rniczh:
```suggestion
        }
```

rniczh (commented):
Thanks @paul0403 Just some minor comments from my end.

mlir/include/QRef/IR/QRefTypes.td
```diff
@@ -0,0 +1,59 @@
...
+//===----------------------------------------------------------------------===//
+// QRef dialect types.
+//===----------------------------------------------------------------------===//
+
+class QRef_Type<string name, string typeMnemonic, list<Trait> traits = []>
+        : TypeDef<QRefDialect, name, traits> {
+    let mnemonic = typeMnemonic;
+}
+
+def QRefQuregType : QRef_Type<"Qureg", "reg"> {
```
dime10: How about `QubitRefType` and `QuregRefType`? A bit cleaner I would say

(I think we should try to avoid turning everything into a "Q" salad as best we can 😅)

mlir/include/QRef/Transforms/Patterns.h
```diff
@@ -0,0 +1,25 @@
...
+// limitations under the License.
+
+#pragma once
+
+#include "mlir/IR/PatternMatch.h"
+
+namespace catalyst {
+namespace qref {
+
+void populateRQHelloWorldPatterns(mlir::RewritePatternSet &patterns);
```
dime10: Are we merging these in? 😅

mlir/lib/Driver/CMakeLists.txt
```diff
@@ -34,6 +34,8 @@ set(LIBS
     catalyst-transforms
     MLIRQuantum
     quantum-transforms
+    MLIRQRef
+    ref-quantum-transforms
```
dime10: out of date maybe?
```suggestion
    qref-transforms
```

If not, `quantum` should come before `ref` for better sorting and visual filtering

mlir/include/QRef/IR/QRefOps.td
```diff
@@ -0,0 +1,655 @@
...
+
+    let arguments = (ins
+        1DTensorOf<[Complex<F64>]>:$in_state,
+        Variadic<QRefQubitType>:$qubits
+    );
+
+    let assemblyFormat = [{
+        `(` $in_state `)` $qubits attr-dict `:` type(operands)
+    }];
+
```
dime10:
```suggestion
```

mlir/include/QRef/IR/QRefOps.td
```diff
@@ -0,0 +1,655 @@
...
+    let assemblyFormat = [{
+        `(` $basis_state`)` $qubits attr-dict `:` type(operands)
+    }];
+}
+
+
+def CustomOp : UnitaryGate_Op<"custom", [ParametrizedGate, AttrSizedOperandSegments]> {
+    let summary = "A generic quantum gate on n qubits with m floating point parameters.";
+    let description = [{
+    }];
```
dime10, line 303: Should we add a description here?

dime10 (commented):
Thanks, great work 💯 Haven't reviewed the C++ but tablegen-wise I would say it looks ready :)

mlir/include/CMakeLists.txt
dime10: oops, just found out I never submitted the first two comments from forever ago 🤦

mlir/include/QRef/IR/QRefInterfaces.td
```diff
@@ -0,0 +1,163 @@
...
+            return $_op->emitError() <<
+                "number of controlling qubits in input (" <<
+                gate.getCtrlQubitOperands().size() << ") " <<
+                "and controlling values (" <<
+                gate.getCtrlValueOperands().size() <<
+                ") must be the same";
+        }
+
+        // STL methods to check duplicates will all complain about `mlir::Value` not having a
+        // comparison method defined, since they all use map/set, which is hash-based
```
mehrdad2m: You are right that it needs a comparison method but I don't think map/set is hash-based. AFAIK, they are tree-based.

mlir/include/QRef/IR/QRefInterfaces.td
```diff
@@ -0,0 +1,163 @@
...
+        // So we just do it manually
+        std::vector<mlir::Value> qubitOperands = gate.getQubitOperands();
+        for (size_t i=0; i < qubitOperands.size(); i++) {
+            for (size_t j=i+1; j < qubitOperands.size(); j++) {
+                if (qubitOperands[i] == qubitOperands[j]) {
+                    return $_op->emitError() << "all qubits on a quantum gate must be " <<
+                    "distinct (including controls)";
+                }
+            }
+        }
```
mehrdad2m, line 130: LLVM has some data structures we can use instead of the STL ones like `SmallPtrSet` or `SmallDenseSet`. something like:
```suggestion
        llvm::SmallPtrSet<mlir::Value, 4> seenQubits;

        for (mlir::Value qubit : gate.getQubitOperands()) {
            if (!seenQubits.insert(qubit).second) {
                return $_op->emitError() << "all qubits on a quantum gate must be " << 
                "distinct (including controls)";
            }
        }

        return mlir::success();
```

mlir/include/QRef/IR/QRefOps.td
```diff
@@ -0,0 +1,655 @@
...
+    );
+
+    let assemblyFormat = [{
+        attr-dict `:` type(results)
+    }];
+
+    let builders = [
+        OpBuilder<(ins),[{
+            AllocQubitOp::build($_builder, $_state,
+            /*qubit=*/ $_builder.getType<QubitType>()
```
mehrdad2m, line 78:
```suggestion
            /*qubit=*/ $_builder.getType<QRefQubitType>()
```

mlir/include/QRef/IR/QRefInterfaces.td
```diff
@@ -0,0 +1,163 @@
...
+    let cppNamespace = "::catalyst::qref";
+
+    let methods = [
+        InterfaceMethod<
+            "Return all operands which are considered gate parameters.",
+            "mlir::ValueRange", "getAllParams"
+        >,
+        InterfaceMethod<
+            "Return the param operand at the requested index.",
+            "mlir::Value", "getParam", (ins "size_t":$idx), /*methodBody=*/[{}],
```
mehrdad2m, line 152: I noticed this is different that the counterpart in value semantic dialect. is there a specific reason for this?

mlir/include/QRef/IR/QRefOps.td
```diff
@@ -0,0 +1,655 @@
...
+    }];
+
+    let hasVerifier = 1;
+}
+
+// -----
+
+class Measurement_Op<string mnemonic, list<Trait> traits = []> :
+        QRef_Op<mnemonic, traits>;
+
```
mehrdad2m:
```suggestion
```
I don't think we need this as other observables are not added right?

mlir/include/QRef/IR/QRefTypes.td
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        bool isDynamic() const {
+            return !isStatic();
+        }
+    }];
+}
+
+def QRefQubitType : QRef_Type<"Qubit", "bit"> {
+    let summary = "A reference to a qubit.";
```
mehrdad2m, line 56: What about `QubitLevel` and `QubitRole` that @joeycarter added recently? don't we need to add them here as well?

mlir/include/QRef/Transforms/Passes.td
```diff
@@ -0,0 +1,29 @@
...
+
+include "mlir/Pass/PassBase.td"
+
+def RQHelloWorldPass : Pass<"rq-hw"> {
+    let summary = "QRef Hello world!";
+
+    // let dependentDialects = [
+    //    "scf::SCFDialect",
+    //    "catalyst::quantum::QuantumDialect"
+    // ];
```
mehrdad2m:
```suggestion
```

mlir/lib/QRef/Transforms/HelloWorldPatterns.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include "Quantum/IR/QuantumOps.h"
+
+using namespace mlir;
+
+namespace {
+
+struct RQHelloWorldPattern : public OpRewritePattern<catalyst::quantum::CustomOp> {
+    using OpRewritePattern::OpRewritePattern;
+
+    LogicalResult matchAndRewrite(catalyst::quantum::CustomOp op,
```
mehrdad2m:
```suggestion
    LogicalResult matchAndRewrite(catalyst::qref::CustomOp op,
```
And also in the rest of this file

mlir/include/QRef/IR/QRefTypes.td
joeycarter: It should be safe to leave these new qubit types out for now, they're still experimental and might change in the future. We can add them in later if we ever need them in the reference-semantics dialect.

mlir/include/CMakeLists.txt
paul0403: I'm resolving the ones from forever ago since they are discussed in the meeting we had with Joey

mlir/include/QRef/IR/QRefInterfaces.td
paul0403: Good point! https://github.com/PennyLaneAI/catalyst/pull/2320/commits/14bdbeab42693fe2c339c88de9eccd9028ec10a6

I also addressed some of your other minor comments here

mlir/include/QRef/Transforms/Patterns.h
paul0403: AH I completely forgot about it haha

mlir/lib/Driver/CMakeLists.txt
paul0403: yep, these were the "pass" part of the boilerplate, I forgot about it after first setting it up so it still has the old name. I'll remove it.

mlir/include/QRef/IR/QRefTypes.td
paul0403: @mehrdad2m mind if I resolve this?

mlir/include/QRef/IR/QRefInterfaces.td
paul0403: I'll just update my comment to say 
```
        // STL methods to check duplicates will all complain about `mlir::Value` not having a
        // comparison method defined
```
?

paul0403: Oh, I saw your suggestion in the next comment!

mlir/include/QRef/IR/QRefInterfaces.td
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2320/commits/b23934cc8b7bf465fdab1a2705b70af2adcf89f5

mlir/include/QRef/IR/QRefInterfaces.td
paul0403: I don't actually know, but this was also the case in the regular quantum dialect. @dime10 do you know why we use `std::vector<Value>` here instead of `ValueRange`?

mlir/include/QRef/IR/QRefOps.td
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2320/commits/208bda81ac91286e9c4c91cc5d73066a1a9104cb

mlir/include/QRef/IR/QRefOps.td
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2320/commits/208bda81ac91286e9c4c91cc5d73066a1a9104cb

mlir/include/QRef/IR/QRefInterfaces.td
dime10: I think the issue is that this aggregates qubit values from different ODS operands, which are not consecutive. Ranges are just a start and length though.

mlir/include/QRef/IR/QRefInterfaces.td
paul0403: It's not really a difference, I just wanted to add an indexed version for the `getAllParams()` and forgot about this other one, so it looks like I replaced this one with `getParam(i)` :sweat_smile: 

Added back https://github.com/PennyLaneAI/catalyst/pull/2320/commits/eaa61045f369fbaf2f5e8d02d17fb764acff717f

mlir/include/QRef/IR/QRefOps.td
paul0403: So, when you do something like 

```
class QRef_Type<string name, string typeMnemonic, list<Trait> traits = []>
        : TypeDef<QRefDialect, name, traits> {
    let mnemonic = typeMnemonic;
}  

def QRefQuregType : QRef_Type<"Qureg", "reg"> 
```

to define a type in tablegen, here's what happens:

- `QRefQuregType`: the name of the type in tablegen op defs, i.e. ops can take in it as 
```
    let arguments = (ins
        QRefQuregType:$qreg
    );
```

- `"Qureg"`: the "name". This is the class name in C++, with tablegen appending a `"Type"`, so in C++ you would use this type as `QuregType`
- "reg": the "typeMnemonic", i.e. the name in IR.

paul0403: See [the definition for the base `TypeDef` tablegen class](https://github.com/llvm/llvm-project/blob/20bb230380c614854fd86c8a2ae8d57ecdf8d4ad/mlir/include/mlir/IR/AttrTypeBase.td#L298)

mlir/include/QRef/IR/QRefTypes.td
paul0403: Do you mean the tablegen name, or the cpp name, or both?

mlir/include/QRef/Transforms/Patterns.h
paul0403: Removed hello world pass https://github.com/PennyLaneAI/catalyst/pull/2320/commits/7e2b46230f14aab16d2fcad13ad2406d2e704c1b

mlir/include/QRef/Transforms/Passes.td
paul0403: Removed, this was part of the pass boilerplate, which is not relevant in this PR as there are no passes https://github.com/PennyLaneAI/catalyst/pull/2320/commits/7e2b46230f14aab16d2fcad13ad2406d2e704c1b

mlir/include/QRef/IR/QRefOps.td
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2320/commits/f8efef08939f87fc18ae92dd62ba5ce14b6a4c96

mlir/include/QRef/IR/QRefTypes.td
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2320/commits/9e25712591acd68d9a4ebb7a75d267413a2c6f9b

mlir/include/QRef/IR/QRefInterfaces.td
```diff
@@ -0,0 +1,167 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef QREF_INTERFACES
+#define QREF_INTERFACES
+
+include "mlir/IR/OpBase.td"
+
+def QuantumRegion : OpInterface<"QuantumRegion"> {
```
dime10, line 20: I would actually like to move away from this concept/definition of a quantum region, since we're not going to have a single register anymore. But if you want to mirror the quantum dialect for now that's fine too

mlir/include/QRef/IR/QRefOps.td
```diff
@@ -0,0 +1,667 @@
...
+    let results = (outs
+        QubitRefType:$qubit
+    );
+
+    let assemblyFormat = [{
+        attr-dict `:` type(results)
+    }];
+
+    let builders = [
+        OpBuilder<(ins),[{
```
dime10, line 76: Just curious, but what does default builder look like that this custom one improves?

mlir/lib/QRef/IR/QRefOps.cpp
```diff
@@ -0,0 +1,235 @@
...
+//===----------------------------------------------------------------------===//
+
+static const mlir::StringSet<> hermitianOps = {"Hadamard", "PauliX", "PauliY", "PauliZ", "CNOT",
+                                               "CY",       "CZ",     "SWAP",   "Toffoli"};
+static const mlir::StringSet<> rotationsOps = {"RX",  "RY",  "RZ",  "PhaseShift",
+                                               "CRX", "CRY", "CRZ", "ControlledPhaseShift"};
+
+LogicalResult CustomOp::canonicalize(CustomOp op, mlir::PatternRewriter &rewriter)
+{
+    if (op.getAdjoint()) {
```
dime10, line 59: I do wonder whether we should automatically (because canonicalizations happen very frequently) optimize the adjoint of operations away, because:
- quantum compilation is often a bit more deliberate from the user side
- a user may also expect to see an adjointed op when inspecting a circuit (because that's what they wrote)
- the adjoint carries semantic meaning

A more controllable approach would simply wrap this is a pass. (out of scope of course)

mlir/test/QRef/UnitTests.mlir
dime10, line 1: What is this file testing?

dime10 (approved):
Thanks Paul 💯

mlir/include/QRef/IR/QRefInterfaces.td
paul0403: Oh, I didn't know about this! I guess the only op that really uses it the the adjoint op, since the other two regioned-thing in PL, namely subroutine and control, both dispatch to other entities in MLIR. 

But for now I'll just keep the mirroring. I anticipate the "can only have one register as input" verifier might need to be removed as I move to the adjoint op in the semantic conversion PR, but I want to deal with it over there as the requirements become a bit more concrete.

paul0403: My current plan is just let the adjoint op see all qubit/qreg references from above via regular closure, like control flow. In that case, the adjoint op itself will simply have no arguments, and this `QuantumRegion` interface can be removed. But again, I'll need to try this out in the semantics conversion PR. Here in the definition PR, just mirroring is enough imo.

mlir/lib/QRef/IR/QRefOps.cpp
paul0403: So, this was proposed by Romain when we first started to implement merge rotations (for OQD). At the time we didn't want to deal with all the adjoint possibilities in the merge rotation pass, so we just decided to canonicalize the adjoint away on the gate itself.

But of course, now that I think about it, pulling it out into its own pass makes the most sense. The merge rotation pass can just call that pass before merging, if needed. 

But I guess it's for another PR. I'll open an issue.

paul0403: https://github.com/PennyLaneAI/catalyst/issues/2516

mlir/test/QRef/UnitTests.mlir
paul0403: Just parsing (I guess it's called "roundtrip"?)

mlir/include/QRef/IR/QRefOps.td
paul0403: This is copied from the value dialect one, but this one doesn't need the op's result type in the `create()` call. It's because this op's result is completely determined with no variability, and it has no operands, and no operand/result segment sizes for mandatory attributes, so in the `create()` call with this builder we don't need to pass in anything.

Image

mlir/lib/QRef/IR/QRefOps.cpp
dime10: Thanks!

mlir/include/QRef/IR/QRefInterfaces.td
dime10: Sounds good, I'll leave it up to you how you want handle it in this PR, just thought I'd mention that the concept of an op/region accepting exactly 1 register doesn't make too much sense anymore^^

doc/releases/changelog-dev.md
```diff
@@ -270,6 +270,34 @@
...
+      qref.custom "CNOT"() %q0, %q1 : !qref.bit, !qref.bit
+      qref.custom "Hadamard"() %q0 : !qref.bit
+      %obs = qref.namedobs %q1 [ PauliX] : !quantum.obs
+      %expval = quantum.expval %obs : f64
+      qref.dealloc %a : !qref.reg<2>
+      return %expval : f64
+  }
+  ```
+
+  Notice that qubit reference values are reusable.
```
joeycarter, line 299: Just a thought, this documentation would probably be better suited in the dialect description in the TableGen rather than in the changelog.

doc/releases/changelog-dev.md
```diff
@@ -346,7 +374,7 @@
 * The upstream MLIR `Test` dialect is now available via the `catalyst` command line tool.
   [(#2417)](https://github.com/PennyLaneAI/catalyst/pull/2417)
 
-* Removing some previously-added guardrails that were in place due to a bug in dynamic allocation 
+* Removing some previously-added guardrails that were in place due to a bug in dynamic allocation
```
joeycarter: This caught my eye since you removed the trailing whitespace; this shouldn't be hyphenated. 🙂 

```suggestion
* Removing some previously added guardrails that were in place due to a bug in dynamic allocation
```

mlir/include/QRef/IR/QRefDialect.h
```diff
@@ -0,0 +1,25 @@
+// Copyright 2025 Xanadu Quantum Technologies Inc.
```
joeycarter:
```suggestion
// Copyright 2026 Xanadu Quantum Technologies Inc.
```

mlir/include/QRef/IR/QRefDialect.td
```diff
@@ -0,0 +1,59 @@
+// Copyright 2025 Xanadu Quantum Technologies Inc.
```
joeycarter:
```suggestion
// Copyright 2026 Xanadu Quantum Technologies Inc.
```

mlir/include/QRef/IR/QRefInterfaces.h
```diff
@@ -0,0 +1,25 @@
+// Copyright 2025 Xanadu Quantum Technologies Inc.
```
joeycarter:
```suggestion
// Copyright 2026 Xanadu Quantum Technologies Inc.
```

mlir/include/QRef/IR/QRefInterfaces.td
```diff
@@ -0,0 +1,167 @@
+// Copyright 2025 Xanadu Quantum Technologies Inc.
```
joeycarter:
```suggestion
// Copyright 2026 Xanadu Quantum Technologies Inc.
```

mlir/include/QRef/IR/QRefInterfaces.td
```diff
@@ -0,0 +1,167 @@
...
+            "mlir::Value", "getRegisterOperand", (ins), /*methodBody=*/[{
+                return $_op->getOperand(0);
+            }]
+        >
+    ];
+
+    let verify = [{
+
+        if ($_op->getNumOperands() != 1)
+            return $_op->emitOpError("must have exactly one operand (quantum register)");
```
joeycarter:
```suggestion
        if ($_op->getNumOperands() != 1) {
            return $_op->emitOpError("must have exactly one operand (a quantum register)");
        }
```

mlir/include/QRef/IR/QRefInterfaces.td
```diff
@@ -0,0 +1,167 @@
...
+                "and controlling values (" <<
+                gate.getCtrlValueOperands().size() <<
+                ") must be the same";
+        }
+
+        llvm::SmallPtrSet<mlir::Value, 4> seenQubits;
+        for (mlir::Value qubit : gate.getQubitOperands()) {
+            if (!seenQubits.insert(qubit).second) {
+                return $_op->emitError() << "all qubits on a quantum gate must be " <<
+                "distinct (including controls)";
```
joeycarter: [Minor] This avoids splitting the string.

```suggestion
                return $_op->emitError() <<
                    "all qubits on a quantum gate must be distinct (including controls)";
```

mlir/include/QRef/IR/QRefOps.h
```diff
@@ -0,0 +1,37 @@
+// Copyright 2025 Xanadu Quantum Technologies Inc.
```
joeycarter:
```suggestion
// Copyright 2026 Xanadu Quantum Technologies Inc.
```

mlir/include/QRef/IR/QRefOps.td
```diff
@@ -0,0 +1,667 @@
+// Copyright 2025 Xanadu Quantum Technologies Inc.
```
joeycarter:
```suggestion
// Copyright 2026 Xanadu Quantum Technologies Inc.
```

mlir/lib/QRef/IR/QRefDialect.cpp
```diff
@@ -0,0 +1,87 @@
+// Copyright 2025 Xanadu Quantum Technologies Inc.
```
joeycarter:
```suggestion
// Copyright 2026 Xanadu Quantum Technologies Inc.
```

mlir/lib/QRef/IR/QRefInterfaces.cpp
```diff
@@ -0,0 +1,24 @@
+// Copyright 2025 Xanadu Quantum Technologies Inc.
```
joeycarter:
```suggestion
// Copyright 2026 Xanadu Quantum Technologies Inc.
```

mlir/lib/QRef/IR/QRefOps.cpp
```diff
@@ -0,0 +1,235 @@
+// Copyright 2025 Xanadu Quantum Technologies Inc.
```
joeycarter:
```suggestion
// Copyright 2026 Xanadu Quantum Technologies Inc.
```

mlir/test/QRef/UnitTests.mlir
```diff
@@ -0,0 +1,314 @@
+// Copyright 2025 Xanadu Quantum Technologies Inc.
```
joeycarter:
```suggestion
// Copyright 2026 Xanadu Quantum Technologies Inc.
```

mlir/test/QRef/VerifierTests.mlir
```diff
@@ -0,0 +1,211 @@
+// Copyright 2025 Xanadu Quantum Technologies Inc.
```
joeycarter:
```suggestion
// Copyright 2026 Xanadu Quantum Technologies Inc.
```

mlir/unittests/QRef/InterfaceTest.cpp
```diff
@@ -0,0 +1,295 @@
+// Copyright 2025 Xanadu Quantum Technologies Inc.
```
joeycarter:
```suggestion
// Copyright 2026 Xanadu Quantum Technologies Inc.
```

joeycarter (approved):
Fantastic, awesome work @paul0403! Kudos, this must have been a big effort 🏆 

I only had a few minor comments and suggestions, but happy to approve 💯

doc/releases/changelog-dev.md
paul0403: How about both?

paul0403: https://github.com/PennyLaneAI/catalyst/pull/2320/commits/ea0e6c15a80db157af2e916ca9d7241aec4c68a8 anyway, added here

PR #2322, opened 2025-12-22
Make all xDSL passes visible to the compiler

mudit2812:
**Context:**
Currently, trying to apply xDSL passes using `@catalyst.apply_pass("my-pass")` or `@qml.transform(pass_name="my-pass")` doesn't work for the passes in `catalyst.python_interface.transforms` because their files do not get collected unless something from that module is imported. This PR makes sure that the global xDSL pass registry is up-to-date before QJIT workflows are lowered to MLIR.

* Import `XDSL_UNIVERSE` when lowering transforms programs to `transform.named_sequence`. The `XDSL_UNIVERSE` is a data structure that collects all dialects and transforms into a single entry point. This makes sure that the transforms get imported, and consequently, registered.
* While fixing this change, I realized that the filecheck tests for the unified compiler were not executing at all. So, this PR also updates `requirements.txt` to install filecheck
* Additionally, this caused new test failures to be detected, which are resolved too.

**Benefits:**

**Possible Drawbacks:**
* This only works for transforms that are present in `catalyst/python_interface/transforms/__init__.py`. But, there are probably not any other cases in which we would want to use a transform that has not been directly or indirectly imported.

**Related GitHub Issues:**

[sc-106692]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -124,6 +124,9 @@
     into phase polynomials and re-synthesizes them into ``CNOT`` and ``RZ`` operators again.
     [(#2294)](https://github.com/PennyLaneAI/catalyst/pull/2294)
 
+  * The `catalyst.python_interface.pass_api.PassDispatcher` now has a more lightweight implementation.
+    [(#2324)](https://github.com/PennyLaneAI/catalyst/pull/2324)
+
   * The global xDSL pass registry is now explicitly refreshed before compiling workflows decorated with
     :func:`catalyst.qjit`.
     [(#2322)](https://github.com/PennyLaneAI/catalyst/pull/2322)
--- a/frontend/catalyst/_version.py
+++ b/frontend/catalyst/_version.py
@@ -16,4 +16,4 @@
 Version number (major.minor.patch[-label])
 """
 
-__version__ = "0.14.0-dev64"
+__version__ = "0.14.0-dev65"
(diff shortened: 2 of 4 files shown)


frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_parity_synth.py
```diff
@@ -479,7 +478,7 @@ def circuit(x: float, y: float, z: float):
             return qml.state()
 
         raw_circuit = qml.qjit(circuit)
-        compiled_circuit = qml.qjit(parity_synth_pass(circuit), target="mlir")
+        compiled_circuit = qml.qjit(parity_synth_pass(circuit))
```
mudit2812, line 481: This change was made because there was a jit-compilation warning being raised because we were trying to execute a circuit that had not been compiled all the way and required recompilation.

frontend/test/pytest/python_interface/conftest.py
```diff
@@ -104,6 +105,11 @@ def run_filecheck():
             generic program string back to an xDSL module. ``False`` by default.
     """
     if not deps_available:
+        warn(
+            "The filecheck Python package must be installed to use the 'run_filecheck' "
+            "fixture. Install it using 'python -m pip install filecheck'.",
+            UserWarning,
+        )
```
mudit2812, line 112: This warning was added so that filecheck being missing is slightly more visible.

sengthai (approved):
Nice! Thanks @mudit2812

PR #2323, opened 2025-12-22
Bump xDSL to 0.56.1

mudit2812:
Bump `xdsl` version to 0.56.1, and `xdsl-jax` version to 0.2.0

Only necessary update was to make all `irdl_options` be defined using tuples instead of lists.

[sc-106690]


mudit2812:
While there are two approvals, I will still consider this PR to be blocked, as I want at least one approval from the compilation team.

PR #2326, opened 2025-12-29
Fix MultiRZ graph decomp

paul0403:
**Context:**
Some graph decomp tests were disabled earlier to push out Jax update.
We wish to restore these functionality.

The relavant core PL change is that, with the new PauliRot decomp rules, the graph could now encounter previously unencountered gate paths. Consider the following example (all examples here assume capture and decomp graph enabled):
```python
@partial(
    qml.transforms.decompose,
    gate_set={"RY", "RX", "CNOT", "Hadamard", "GlobalPhase"},
)
@qml.qnode(qml.device("lightning.qubit", wires=4))
def circuit():
    qml.SingleExcitationPlus(0.5, wires=[0, 1])
    return qml.expval(qml.Z(0))

``` 

Under the current stable versions (v0.43 and v0.13), the reached gate path does not have PauliRot gates. However, on latest, the reached gate path includes a paulirot, which itself includes a mulitrz:
```mlir
// starting IR, aka "0.mlir"
// One of the decomp rule functions is decomposing pauli rot into multirz
    func.func public @_pauli_rot_decomposition.partial(%arg0: !quantum.reg, %arg1: tensor<1xf64>, %arg2: tensor<1xi64>) -> !quantum.reg attributes {llvm.linkage = #llvm.linkage<internal>} {
      ...
      %out_qubits = quantum.multirz(%extracted_0) %2 : !quantum.bit
      ...
    }

// Then, another decomp rule function to decompose multi rz into regular rot gates (since that's the requested gateset from user)
    func.func public @_multi_rz_decomposition(%arg0: !quantum.reg, %arg1: tensor<1xf64>, %arg2: tensor<1xi64>) -> !quantum.reg attributes {llvm.linkage = #llvm.linkage<internal>, num_wires = 2 : i64, target_gate = "MultiRZ"} {
      ...
      %out_qubits = quantum.custom "RZ"(%extracted_0) %2 : !quantum.bit
      ...
    }
```

This brings about a previously unencoutnered problem: multirz is special, in the sense that it can take a variable number of qubit input values. For different number of input qubits, the decomposition rule function is of course different (e.g. multirz on 3 qubits will need a for loop, whereas multirz on one qubit does not). So there needs to be one decomp rule function per multirz-versioned-by-num-wires.

The way these versioned-by-num-wires decomp rule functions are created in the frontend is [controlled by a `requires_copy` argument](https://github.com/PennyLaneAI/catalyst/blob/b9b15733cbd447b7b6aa4f401db59b79d07a8610/frontend/catalyst/from_plxpr/decompose.py#L340). However, looking at [how we determine whether copies are required](https://github.com/PennyLaneAI/catalyst/blob/b9b15733cbd447b7b6aa4f401db59b79d07a8610/frontend/catalyst/from_plxpr/decompose.py#L236), we see the logic is buggy. `num_wires` will be `-1` in the static registry `COMPILER_OPS_FOR_DECOMPOSITION` for gates on variable number of wires (multirz, gphase, PPR/M), but this is overwritten by the `num_wires = op_num_wires` line above. 

The purpose of this line is to tell the `_create_decomposition_rule` helper function below the number of wires on this specific instance of multirz gate (so it knows how to create the signature of the decomp rule func, etc.), so that line should stay. However, this line overwrites the `-1` in the static registry, but the check later `requires_copy=num_wires == -1` is still depending on the static registry info. This means for multirz (and other variadic gates), `requires_copy` is always `False` in this buggy logic.

This is the source of the original error inside the `--decompose-lowering` mlir pass: it is complaining about 
```
circuit_18_1/1_QuantumCompilationStage/1_SplitMultipleTapesPass.mlir:322:25: error: 'tensor.from_elements' op failed to verify that operand types match result element type
      %out_qubits_9:2 = quantum.multirz(%extracted_8) %8, %9 : !quantum.bit, !quantum.bit
                        ^
circuit_18_1/1_QuantumCompilationStage/1_SplitMultipleTapesPass.mlir:322:25: note: see current operation: %21 = "tensor.from_elements"(%17, %18) : (i64, i64) -> tensor<1xi64>
```
In the above example (with the single excitation), multiple pauli rot rules created multiple multirz gates, some using one qubit and some using 2 qubits. But because we did not make any num-qubits-versioned copies of the decomp rule function, all multirzs are using the same rule. In this particular error message, the rule function takes one qubit, but the visited multitz op takes in 2. This causes a type mismatch when [creating the tensor argument to call the decomp rule function with (to prep for the inlining later)](https://github.com/PennyLaneAI/catalyst/blob/b9b15733cbd447b7b6aa4f401db59b79d07a8610/mlir/lib/Quantum/Transforms/DecomposeLoweringImpl.hpp#L341) .

Turn on `requires_copy` for all 4 variadic gates (multirz, gphase, PPR/M). 

**Benefits:**

**Possible Drawbacks:**
Yet another band-aid on the graph decomp...

**Related GitHub Issues:**

[sc-105502]

### DIFF
--- a/.dep-versions
+++ b/.dep-versions
@@ -10,8 +10,8 @@ enzyme=v0.0.203
 
 # For a custom PL version, update the package version here and at
 # 'doc/requirements.txt'
-pennylane=0.44.0-dev62
+pennylane=0.44.0-dev72
 
 # For a custom LQ/LK version, update the package version here and at
 # 'doc/requirements.txt'
-lightning=0.44.0-dev16
+lightning=0.44.0-dev26
--- a/.github/workflows/check-pl-compat.yaml
+++ b/.github/workflows/check-pl-compat.yaml
@@ -44,7 +44,7 @@ jobs:
         # was built and uploaded to TestPyPI, so that the tests are the compatible ones
         # to be run with the latest rc/rc/rc packges on TestPyPI.
         # This would be the last autobump-and-upload commit made on the rc branch.
-        git checkout v0.13.0-rc
+        git checkout v0.14.0-rc
         git checkout "$(git log -1 --author="ringo-but-quantum" --pretty=format:%H)"
 
     - name: Set up Python
@@ -139,7 +139,7 @@ jobs:
         # Thus this one shouldn't have no-deps: it needs to pull in all PL dependencies
         jax_version=$(grep jax .dep-versions | awk -F '=' '{ print $2 }')
         pip install jax==$jax_version
-        pip install --pre -U --index-url https://test.pypi.org/simple/ pennylane-catalyst'<=0.13.0'
+        pip install --pre -U --index-url https://test.pypi.org/simple/ pennylane-catalyst'<=0.14.0'
 
     - name: Install PennyLane-Lightning (stable)
       if: ${{ inputs.lightning == 'stable' }}
@@ -159,8 +159,8 @@ jobs:
     - name: Install PennyLane-Lightning (release-candidate)
       if: ${{ inputs.lightning == 'release-candidate' }}
       run: |
-        pip install --no-deps --pre -U --index-url https://test.pypi.org/simple/ pennylane-lightning'<=0.43.0'
-        pip install --no-deps --pre -U --index-url https://test.pypi.org/simple/ pennylane-lightning-kokkos'<=0.43.0'
+        pip install --no-deps --pre -U --index-url https://test.pypi.org/simple/ pennylane-lightning'<=0.44.0'
+        pip install --no-deps --pre -U --index-url https://test.pypi.org/simple/ pennylane-lightning-kokkos'<=0.44.0'
 
     # PennyLane doesn't update its dev version number on every commit like Lightning, so we need to
     # force the package to be re-installed. First, handle potential dependency changes, then force
@@ -178,7 +178,7 @@ jobs:
     - name: Install PennyLane (release-candidate)
       if: ${{ inputs.pennylane == 'release-candidate' }}
       run: |
-        pip install --no-deps --pre -U --index-url https://test.pypi.org/simple/ pennylane'<=0.43.0'
+        pip install --no-deps --pre -U --index-url https://test.pypi.org/simple/ pennylane'<=0.44.0'
 
     - name: Add Frontend Dependencies to PATH
       if: ${{ inputs.catalyst == 'latest' }}
--- a/doc/dev/release_notes.rst
+++ b/doc/dev/release_notes.rst
@@ -3,7 +3,7 @@ Release notes
 
 This page contains the release notes for Catalyst.
 
-.. mdinclude:: ../releases/changelog-dev.md
+.. mdinclude:: ../releases/changelog-0.14.0.md
 
 .. mdinclude:: ../releases/changelog-0.13.0.md
 
--- a/doc/releases/changelog-0.13.0.md
+++ b/doc/releases/changelog-0.13.0.md
@@ -1,4 +1,4 @@
-# Release 0.13.0 (current release)
+# Release 0.13.0
 
 <h3>New features since last release</h3>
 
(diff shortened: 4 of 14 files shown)


frontend/catalyst/from_plxpr/decompose.py
```diff
@@ -228,12 +228,15 @@ def cleanup(self):
                 elif num_wires == -1 and op_num_wires is not None:
                     num_wires = op_num_wires
 
+                requires_copy = num_wires == -1
```
dime10: Can you just move this check up before overwriting the num_wires, instead of hardcoding the gates for which we need a copy?

paul0403: Makes sense :sweat_smile:

frontend/test/pytest/from_plxpr/test_decompose_transform.py
```diff
@@ -251,9 +251,6 @@ def circuit():
         ):
             qml.qjit(circuit)()
 
-    @pytest.mark.xfail(
-        reason="paulirot issue to be resolved at https://github.com/PennyLaneAI/catalyst/pull/2233"
-    )
     @pytest.mark.usefixtures("use_capture_dgraph")
     def test_ftqc_rotxzx(self):
         """Test that FTQC RotXZX decomposition works with from_plxpr."""
```
paul0403, line 256: Unrelated, but for me locally when I run this pytest file with `-n auto`, I get 
```
pennylane has no module named ftqc
```
:sweat_smile: 

Doesn't show up when just running the tests serially

frontend/test/pytest/from_plxpr/test_decompose_transform.py
```diff
@@ -384,8 +375,8 @@ def circuit():
         resources = qml.specs(with_qjit, level="device")()["resources"].gate_types
         assert resources == expected_resources
 
-    @pytest.mark.xfail(
-        reason="paulirot issue to be resolved at https://github.com/PennyLaneAI/catalyst/pull/2233"
+    @pytest.mark.skip(
```
paul0403, line 378: I changed this from xfail to skip, because these tests now actually sometimes pass (which would be reported as pytest failure due to strict xfail)!

This is the non-deterministic decomp graph solution issue. We can wait for @astralcai to come back and deal with this separately.

lazypanda10117 (approved):
Thanks @paul0403! I did the `num_wires` fix but didn't realize it could affect the `requires_copy` flag. Thanks for catching this 💯

frontend/test/pytest/from_plxpr/test_decompose_transform.py
maliasadi: what? :thinking:

frontend/test/pytest/from_plxpr/test_decompose_transform.py
maliasadi: Sounds good! Could you create/post the SC ticket related to this fix here?

maliasadi (approved):

frontend/test/pytest/from_plxpr/test_decompose_transform.py
paul0403: I don't know if Astral has already created a story for the non deterministic graph solution issue... I don't want to duplictae stories. Can you help me search for it? :sweat_smile: I think you're way more familiar with the graph decomp stories than me haha

frontend/test/pytest/from_plxpr/test_decompose_transform.py
paul0403: Also CI is run with `-n auto` and CI is fine... very strange, but probably just something with my setup

frontend/test/pytest/from_plxpr/test_decompose_transform.py
paul0403: For this PR itself, I linked to Jeffery's story with the list of all the issues for the quantum.paulirot op

PR #2327, opened 2025-12-30
Test suite maintainence regarding kept intermediates

paul0403:
**Context:**
By default, when `keep_intermediate` is turned on, [the workspace is the current working directory](https://github.com/PennyLaneAI/catalyst/blob/ed3920d4e705f8a5f06e263df66e92f433c5e796/frontend/catalyst/jit.py#L549).

This of course makes sense from a user perspective, but in CI, when multiple tests are run in parallel, this could lead to multiple tests competing for the same name (e.g. multiple tests have their qnode named `workflow`). In the worst case, these OS level race conditions could lead to pytest worker crashes. 

All tests that keep intermediates should keep their intermediates in a temporary directory.
To do so, we manually overwrite the `QJIT.use_cwd_for_workspace` flag. 
Since the compilation needs to happen after overwriting the flag, which can only happen after the `QJIT.__init__` (where aot compile also happens), these tests all need to be converted to jit compile mode.

I also updated some tests as small cleanups.

**Benefits:**
Safer CI.

**Possible Drawbacks:**
How should we broadcast this new rule (that tests keeping intermediates should do so in a temporary directory)? Ideally we should add a fixture, but it's a bit tricky...

### DIFF
--- a/frontend/test/pytest/test_compiler.py
+++ b/frontend/test/pytest/test_compiler.py
@@ -67,11 +67,14 @@ class TestCompilerOptions:
 
         @qjit(verbose=verbose, logfile=logfile, keep_intermediate=keep_intermediate)
         @qml.qnode(qml.device(backend, wires=1))
-        def workflow():
-            qml.PauliX(wires=0)
+        def workflow(x):
+            qml.RX(x, wires=0)
             return qml.state()
 
-        workflow()
+        # Create tmp workspaces for intermediates to avoid CI race conditions
+        workflow.use_cwd_for_workspace = False
+        workflow.jit_compile((1.2,))
+
         capture_result = capsys.readouterr()
         capture = capture_result.out + capture_result.err
         assert ("[SYSTEM]" in capture) if verbose else ("[SYSTEM]" not in capture)


paul0403:
I added this to release milestone because I think the random failure (that this would hopefully fix) will cause a lot of headaches during the release process.

dime10 (approved):

frontend/test/pytest/test_compiler.py
kipawaa, line 83: Also needed here

frontend/test/pytest/test_debug.py
```diff
@@ -350,26 +353,26 @@ def f(x: float):
...
                 "BufferizationStage",
-                "%2 = arith.mulf %1, %1 : f64",
+                "%2 = arith.mulf %1, %1 : f64\n",
                 "%t = arith.mulf %1, %1 : f64\n" + "    %2 = arith.mulf %t, %1 : f64\n",
             ),
             (
                 "MLIRToLLVMDialectConversion",
-                "%5 = llvm.fmul %4, %4  : f64\n",
-                "%t = llvm.fmul %4, %4  : f64\n" + "    %5 = llvm.fmul %t, %4  : f64\n",
+                "%7 = llvm.fmul %6, %6 : f64\n",
```
paul0403, line 374: Just a note on this change, one of the llvm updates introduced (more) poison values into the llvm dialect IR, so things got pushed down.

frontend/test/pytest/test_compiler.py
paul0403: Nice catch! https://github.com/PennyLaneAI/catalyst/pull/2327/commits/19e3c0e01d049b469b81eb84d5cd65aca8b7f1c9

PR #2329, opened 2026-01-02
Refactor the quantum dialect for improved modularity

paul0403:
**Context:**
The enums and traits of quantum dialect are defined in `QuantumOps.td`.
If other dialects want to use them (e.g. observable value types), they have to include the entire `QuantumOps.td`, which might include things that they don't want.

For example, the reference semantics dialect will have its own "custom op", so it cannot include `QuantumOps.td`: the two custom ops will clash.

Create "include files" for quantum dialect's enums and traits. The quantum-dialect types have also been refactored and placed into a separate file scope.

**Benefits:**
Better modularity.

[sc-107669]

### DIFF
--- a/mlir/include/Quantum/IR/QuantumTypes.td
+++ b/mlir/include/Quantum/IR/QuantumTypes.td
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#ifndef QUANTUM_TYPES
+#define QUANTUM_TYPES
+
 include "mlir/IR/AttrTypeBase.td"
 
 include "Quantum/IR/QuantumDialect.td"
@@ -37,3 +40,5 @@ def ObservableType : Quantum_Type<"Observable", "obs"> {
 def ResultType : Quantum_Type<"Result", "res"> {
     let summary = "A quantum measurement result.";
 }
+
+#endif // QUANTUM_TYPES


mlir/include/Quantum/IR/QuantumEnums.td
```diff
@@ -0,0 +1,46 @@
...
+}
+
+//===----------------------------------------------------------------------===//
+// Quantum dialect attributes.
+//===----------------------------------------------------------------------===//
+
+def NamedObservableAttr : EnumAttr<QuantumDialect, NamedObservable, "named_observable">;
+
+
+#endif // QUANTUM_OPS
```
joeycarter:
```suggestion
#endif // QUANTUM_ENUMS
```

mlir/include/Quantum/IR/QuantumOps.td
```diff
@@ -15,45 +15,15 @@
 #ifndef QUANTUM_OPS
 #define QUANTUM_OPS
 
-include "mlir/IR/EnumAttr.td"
+// include "mlir/IR/EnumAttr.td"
```
joeycarter: I assume you were going to remove this line, but just a reminder 🙂 

```suggestion
```

mlir/include/Quantum/IR/QuantumAttrDefs.td
```diff
@@ -0,0 +1,46 @@
...
+    ]> {
+    let cppNamespace = "catalyst::quantum";
+    let genSpecializedAttr = 0;
+}
+
+//===----------------------------------------------------------------------===//
+// Quantum dialect attributes.
+//===----------------------------------------------------------------------===//
+
+def NamedObservableAttr : EnumAttr<QuantumDialect, NamedObservable, "named_observable">;
```
joeycarter, line 39: This might be misleading for people working on this file in the future, suggesting that all quantum dialect attributes should go here, rather than just attributes to represent the enums. Maybe a simple fix is just to remove the comment banner.

joeycarter (commented):
Makes sense to me, happy to approve once these few minor points are resolved 🚀 

Also don't forget the changelog 😉

mlir/include/Quantum/IR/QuantumAttrDefs.td
paul0403: Any thoughts on putting "attributes" into dialect.td?

joeycarter: Personally I like having the enum object and its corresponding attribute defined one after the other; it makes navigating the code easier.

Maybe in the "Quantum dialect enums" banner above we just change it to "Quantum dialect enums and enum attributes", and group them all together there.

paul0403:
@joeycarter I addressed your comments!

For changelog, since today is the last day of feature freeze, and the odds of getting this merged in today seems quite low, how about I wait until after the rc branch and the next release's changelog is created?

paul0403:
> Looks good to me! Thanks @paul0403 👍 It makes me think we should isolate QEC's enum as well.

At the end of the day it's a per-dialect decision. I don't think upstream mlir has a rule/guidance to separate out enums for all dialects. But for the core quantum one, since it is a "base" dialect, I think it would make more sense to have the headers to be more modular. 

And also this was prompted by a concrete issue (described in PR description) :laughing: .

mlir/include/Quantum/IR/QuantumDialect.td
```diff
@@ -73,6 +73,13 @@ def ResultType : Quantum_Type<"Result", "res"> {
     let summary = "A quantum measurement result.";
 }
 
+//===----------------------------------------------------------------------===//
+// Quantum dialect traits.
+//===----------------------------------------------------------------------===//
+
+def Unitary : NativeOpTrait<"UnitaryTrait">;
+def Hermitian : NativeOpTrait<"HermitianTrait">;
+
```
joeycarter, line 59: Since we're adding the trait TableGen declarations to `QuantumDialect.td`, would it be better to move the C++ declarations to `QuantumDialect.h` instead of creating a new header? Would that block your upcoming work in any way?

joeycarter:
Following discussions with @paul0403, I've also refactored the quantum type declarations into separate files. I've also renamed a few things to follow a few conventions I was able to identify in some of MLIR's built-in dialects (although even there there is no strict adherence to any fully established convention).

joeycarter (approved):

PR #2330, opened 2026-01-02
Fix missing type support for unsigned integers in callbacks

mlxd:
**Context:** The use of `jax.numpy.uintX` types in Catalyst programs causes a failure with callbacks due to missing conversion support to the appropriate MLIR types.

**Description of the Change:** This PR adds the explicit unsigned conversions.

**Benefits:** Ensures we can compile and pass unsigned values to and from the program

**Possible Drawbacks:** None

**Related GitHub Issues:**

### DIFF
--- a/doc/releases/changelog-0.14.0.md
+++ b/doc/releases/changelog-0.14.0.md
@@ -434,6 +434,11 @@
 * Updated the type support for callbacks allowing use of unsigned integers.
   [(#2330)](https://github.com/PennyLaneAI/catalyst/pull/2330)
 
+* Fixed a bug where the `qec.ppr` op attribute `rotation_kind` was not correctly constrained to
+  be one of ±1, ±2, ±4, or ±8.
+  Also, for Identity Pauli product, the `rotation_kind` was correctly set to 1, instead of 0.
+  [(#2344)](https://github.com/PennyLaneAI/catalyst/pull/2344)
+
 * Updated tests and PennyLane dependency pin to follow changes introduced by
   [PennyLaneAI/pennylane#8290](https://github.com/PennyLaneAI/pennylane/pull/8290).
   [(#2286)](https://github.com/PennyLaneAI/catalyst/pull/2286)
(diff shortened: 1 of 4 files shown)


mlxd:
[sc-107008]

frontend/catalyst/utils/types.py
```diff
@@ -61,14 +62,11 @@ def convert_numpy_dtype_to_mlir(dtp):
...
-    elif dtp == np.dtype(np.int16):
-        return ir.IntegerType.get_signless(16)
-    elif dtp == np.dtype(np.int32):
-        return ir.IntegerType.get_signless(32)
-    elif dtp == np.dtype(np.int64):
-        return ir.IntegerType.get_signless(64)
+    elif np.issubdtype(dtp, np.signedinteger):
+        return ir.IntegerType.get_signless(8 * num_bytes)
+    elif np.issubdtype(dtp, np.unsignedinteger):
+        return ir.IntegerType.get_unsigned(8 * num_bytes)
```
dime10, line 68: The previous code only allowed up to 64 bits, any chance the new code could allow more now? (or even non-powers of two)

mlxd: The types are all under the 1 byte to 8 byte widths (8,16,32,64), as the limits follow the C conventions for dtypes of ints. It should be safe to do this as numpy won't allow creating a non 1-8 byte sized integer, with no non-power-of-2 allowed.

mlxd: https://numpy.org/doc/stable/reference/arrays.scalars.html#sized-aliases

dime10 (changes requested):
Happy to approve, wanna add a changelog entry?
I'm also wondering if we'd want an integration test on this or not, since there could be more issues throughout the pipeline (first time we're using unsigned values) so it would be good to see this working at every level.

frontend/test/pytest/test_callback.py
```diff
@@ -193,20 +193,37 @@ def cir(x):
...
+        jnp.uint8,
+        jnp.uint16,
+        jnp.uint32,
+        jnp.uint64,
+    ],
+)
+def test_identity_types_cast_shaped_array(arg, dtype):
+    """Test callback with arguments and return values of given types."""
+
+    arg_cast = jnp.array(dtype(arg))
```
mlxd:
```suggestion
    arg_cast = jnp.array(arg, dtype=dtype)
```

PR #2334, opened 2026-01-05
Install RC PL and Lightning when on RC branch

paul0403:
**Context:**
In `setup.py`, we install RC PL and Lightning packages. This only occurs for the rc branch.
For PRs targeting RC branch as base, locally `make frontend` will now automatically install RC PL and Lightning! 

Note that starting from the previous release, we sync all versioning from testpypi. There will be no more git-based installations.

Note that when merging rc branch back to main after the release, this PR must be reverted for .dep_versions to take back control of the versions.

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -30,7 +30,14 @@ matplotlib==3.10.0
 lxml_html_clean
 
 # Pre-install PL development wheels
---extra-index-url https://test.pypi.org/simple/
-pennylane-lightning-kokkos==0.44.0-dev26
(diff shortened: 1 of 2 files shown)


paul0403:
But in that scenario, `setup.py` also needs to be reverted when merging rc branch back to main. So it's the same clunkiness but in a different file no?

paul0403:
I just hardcode the needed rc versions in setup.py. Since it will be reverted when merging back to `main` anyway, I didn't bother with the whole fstring song and dance.

It will fail until lightning uploads rc

PR #2336, opened 2026-01-05
Fix arbitrary angle PPRs for negative angles

lazypanda10117:
**Context:** Fixing a bug in the handling of negative angles for arbitrary angle PPRs.

https://github.com/PennyLaneAI/catalyst/issues/2341


sengthai (approved):
Nice!

PR #2338, opened 2026-01-06
A few fixes in rcrcrc action

paul0403:
**Context:**
We fix a few things regarding the rcrcrc script, and xdsl dependency.

1. Adds explicit `rc0` lower bound to rcrcrc pip install
2. Adds apt install graphviz to rcrcrc
3. Remove xdsl from `requirements.txt`, since it is already in `setup.py`.

### DIFF
--- a/.github/workflows/check-pl-compat.yaml
+++ b/.github/workflows/check-pl-compat.yaml
@@ -193,9 +193,6 @@ jobs:
 
     - name: Run Frontend Tests
       run: |
-        pip list | grep -i penny
-        pip list | grep -i xdsl
-        pip list | grep -i jax
         make pytest
 
     - name: Run Frontend Tests (Kokkos)


.github/workflows/check-pl-compat.yaml
```diff
@@ -191,6 +192,8 @@ jobs:
 
     - name: Run Frontend Tests
       run: |
+        pip list | grep -i penny
+        pip list | grep -i xdsl
```
mudit2812:
```suggestion
```

.github/workflows/check-pl-compat.yaml
```diff
@@ -137,9 +137,9 @@ jobs:
       run: |
         # This is the first for the three rc installs
         # Thus this one shouldn't have no-deps: it needs to pull in all PL dependencies
-        jax_version=$(grep jax .dep-versions | awk -F '=' '{ print $2 }')
-        pip install jax==$jax_version
-        pip install --pre -U --index-url https://test.pypi.org/simple/ pennylane-catalyst'<=0.14.0'
+        # And also this one needs to be --extra-index-url because it needs to see regular PyPI
+        sudo apt-get install -y graphviz
+        pip install --pre -U --extra-index-url https://test.pypi.org/simple/ pennylane-catalyst'>=0.14.0rc0,<=0.14.0'
```
mudit2812: Isn't it a good idea to always use `--extra-index-url` instead of `--index-url`?

paul0403: I think it depends, [Alex made a point during the last release](https://github.com/PennyLaneAI/catalyst/pull/2101#discussion_r2412366448) that if we are only installing rc packages, we should only look at testpypi.

The logic of the rcrcrc script is:

0. pip install requirements.txt (for basic things like numpy)
1. Install catalyst rc package. This pulls in special catalyst dependencies in setup.py, like jax and xdsl, as well as dependencies of PL itself, since PL itself is in catalyst's setup.py (versioned by `.dep-versions` file). So this needs to see regular pypi and hence it needs `--extra-index-url`
2. Install PL and lightning rc package. This step should not see regular pypi. This step has `--no-deps` since its only purpose is to overwrite the PL and lightning dev versions installed in the above step. This was Alex's point. So this step uses `--index-url`.

I think this system is robust enough. The only breaking possibility is when a PL rc package has a new dependency/version not included in a dev version, but I think it's safe to assume that no dependency updates will happen during feature freeze :laughing:

.github/workflows/check-pl-compat.yaml
paul0403: removed https://github.com/PennyLaneAI/catalyst/pull/2338/commits/3235166ce1e3686221a4122f9400847f842fdd65

.github/workflows/check-pl-compat.yaml
jzaia18: This might pull in pre-release versions of some of the dependencies of Catalyst, such as [this numpy prerelease](https://test.pypi.org/project/numpy/2.4.0.dev0/). We run into this problem in the benchmark suite CIs as well. Just a comment, there's not much we can do about it and (at least for now) it's benign.

paul0403: Oh no :melting_face: I guess we cross that bridge when we come to it...

dime10: There should be a way around this, some ideas that might work:
- `-U` may not be needed if Catalyst wasn't already installed. If we remove the flag, and the dependencies are already installed, they might not be upgraded to pre-release.
- `--pre` is not needed if you explicitly specify a pre-release version constraint (at least I know it's not needed with `==X.X.X-preY`, not sure about lower bounds). If we can remove it dependencies won't be pre-release.

paul0403: I think you are right, we needed those flags before because we were not specifying `rc` in the bounds, but now that we are sandwhiching between rc and release version, the only acceptable versions are the rc versions, so we don't need `--pre`. 

I also remove `-U` because this is the first time this script installs catalyst. Let's try...

paul0403: It worked, we no longer need `--pre` here

mudit2812: Thanks for the explanation! Resolving.

mudit2812 (approved):

PR #2344, opened 2026-01-06
Enhance `qec.ppr` op attribute constraints and fix related bugs

### LINKED ISSUE #2343: `rotation_kind` in the `PPRotationOp` operation is missing a constraint on its legal values
The `rotation_kind` in the `PPRotationOp` is defined as a I16Attr but there is no limitation on the values. Since only 2, 4, and 8 are the only relevant values, We should add a verifier or a constraint (preferred) so any invalid IR raises an error. 
Currently, merge rotation pass is already implicitly assuming these values to be powers of 2, so if someone tries a goofy value,  merge rotation would incorrectly change it to a valid IR and continue without raising an error. e.g.
```
%0 = qec.ppr ["Y"](9) %q1: !quantum.bit
%1 = qec.ppr ["Y"](9) %0: !quantum.bit
```
would turn into
```
%0 = qec.ppr ["Y"](4) %arg0 : !quantum.bit 
```
After merge-rotation pass.

sengthai:
- Added a constraint for the `rotation_kind` attribute to ensure it is limited to ±1, ±2, ±4, or ±8.
- Corrected the default value for the `rotation_kind` in the Identity Pauli product from 0 to 1.

**Related GitHub Issues:**
close #2343

### DIFF
--- a/mlir/test/QEC/SmokeTest.mlir
+++ b/mlir/test/QEC/SmokeTest.mlir
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// RUN: quantum-opt %s | FileCheck %s
+// RUN: quantum-opt %s --split-input-file --verify-diagnostics | FileCheck %s
 
 func.func @foo(%q1 : !quantum.bit, %q2 : !quantum.bit) {
     qec.ppr ["X", "Z"] (4) %q1, %q2 : !quantum.bit, !quantum.bit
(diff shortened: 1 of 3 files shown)


lazypanda10117 (approved):
Looks good! thanks @sengthai

mlir/test/QEC/SmokeTest.mlir
```diff
@@ -101,3 +101,11 @@ func.func @arbitrary(%q1 : !quantum.bit, %q2 : !quantum.bit) {
     %2:2 = qec.ppr.arbitrary ["X", "Z"](%const_1) %1#0, %1#1 cond(%c0) : !quantum.bit, !quantum.bit
     func.return
 }
+
+// -----
+
+func.func @baz_error(%q1 : !quantum.bit, %q2 : !quantum.bit) {
+    // expected-error@below {{'qec.ppr' op attribute 'rotation_kind' failed to satisfy constraint: 16-bit signless integer attribute whose value is ±1, ±2, ±4, or ±8}}
+    %0, %1 = qec.ppr ["X", "Z"] (16) %q1, %q2 : !quantum.bit, !quantum.bit
```
paul0403, line 109: Can we add a test that actually uses the bit shift? aka one in the range of 1~8

Issue #2251, opened 2025-12-01
`qjit(value_and_grad)` fails when `argnums=0` in multi-arg function


ddeunghoon:
### Summary
`qjit(value_and_grad(...))` looks like fails to compile for a multi-argument function when differentiating w.r.t. the first argument (`argnums=0`).  
Single-argument functions work, and differentiating w.r.t. the second argument (`argnums=1`) works. Swapping the argument order and adjusting `argnums` also works, which suggests an `argnums`/result mapping issue in the generated `gradient.value_and_grad` op.

### Environment
- PennyLane Catalyst: 0.13.0
- JAX: 0.6.2
- Device: CudaDevice(id=0)

### Reproduction steps
1. Run the script below: `debug_value_and_grad_mapping.py`
2. Observe:
   - `f1(params)` (single arg) succeeds for both f64 and f32
   - `f2(params, seed)` fails only when `argnums=0`
   - `f2(params, seed)` succeeds when `argnums=1`
   - Swapping argument order (`f2_swapped(seed, params)`) and using `argnums=1` succeeds

### Expected behavior
For `f2(params, seed)` with `argnums=0`, the compiled `value_and_grad` should return:
- value: scalar
- grad: same shape/dtype as `params` (e.g., `(476,)`, `float64`)

### Actual behavior

```python
# debug_value_and_grad_mapping.py
import os, glob, tempfile, time
import jax
import jax.numpy as jnp
import catalyst
from catalyst import qjit, value_and_grad

def block(x):
    try: x.block_until_ready()
    except Exception: pass
    return x

def read_text(path):
    with open(path, "r", encoding="utf-8", errors="ignore") as f:
        return f.read()

def extract_context(text, needle, lines=16):
    if needle not in text:
        return None
    parts = text.splitlines()
    idxs = [i for i, ln in enumerate(parts) if needle in ln]
    out = []
    for i in idxs[:3]:
        lo, hi = max(0, i - lines), min(len(parts), i + lines)
        out.append("\n".join(parts[lo:hi]))
    return "\n\n---\n\n".join(out)

def find_mlir_with_token(search_roots, token="gradient.value_and_grad"):
    hits = []
    for root in search_roots:
        mlirs = glob.glob(os.path.join(root, "**", "*.mlir"), recursive=True)
        for p in sorted(mlirs):
            try:
                t = read_text(p)
            except Exception:
                continue
            if token in t:
                hits.append((p, t))
    return hits

[...]

PR #2349, opened 2026-01-06
Fix gradient type validation when differentiating a subset of arguments

sengthai:
**Context:**
The `gradient.value_and_grad` operation's verifier incorrectly validated gradient result types by matching from the tail of callee arguments, ignoring the `diffArgIndices` attribute. This caused false verification errors when differentiating a subset of arguments with different types.

**Related GitHub Issues:**
close https://github.com/PennyLaneAI/catalyst/issues/2251
[[sc-105011]]

### DIFF
--- a/mlir/test/Gradient/VerifierTest.mlir
+++ b/mlir/test/Gradient/VerifierTest.mlir
@@ -489,3 +489,27 @@ func.func private @multi_arg_diff_types2(%arg0: tensor<3xf64>, %arg1: tensor<2xf
 // expected-error@+1 {{result types do not match gradient 0 (for arg 0) expected type 'tensor<3xf64>' but got 'tensor<2xf32>'}}
 gradient.value_and_grad "fd" @multi_arg_diff_types2(%t0, %t1) {diffArgIndices = dense<0> : tensor<1xi64>, resultSegmentSizes = array<i32: 1, 1>} : (tensor<3xf64>, tensor<2xf32>) -> (tensor<f64>, tensor<2xf32>)
 
+// -----
+
+func.func private @multi_diff_args(%arg0: tensor<3xf64>, %arg1: tensor<2xf64>) -> tensor<f64>
+
+%f0 = arith.constant 0.0 : f64
+%t0 = tensor.from_elements %f0, %f0, %f0 : tensor<3xf64>
+%t1 = tensor.from_elements %f0, %f0 : tensor<2xf64>
+
+// This should pass: 2 diff args -> 2 gradients, each matching its arg type
+gradient.value_and_grad "fd" @multi_diff_args(%t0, %t1) {diffArgIndices = dense<[0, 1]> : tensor<2xi64>, resultSegmentSizes = array<i32: 1, 2>} : (tensor<3xf64>, tensor<2xf64>) -> (tensor<f64>, tensor<3xf64>, tensor<2xf64>)
+
+// -----
+
+// Test value_and_grad differentiating only the second argument (not the first).
+
+func.func private @diff_second_arg(%arg0: tensor<3xf64>, %arg1: tensor<2xf64>) -> tensor<f64>
+
+%f0 = arith.constant 0.0 : f64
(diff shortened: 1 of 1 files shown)


mlir/lib/Gradient/IR/GradientOps.cpp
```diff
@@ -220,18 +220,15 @@ LogicalResult ValueAndGradOp::verifySymbolUses(SymbolTableCollection &symbolTabl
...
-        // grad type and shape should match the callee's argument's type and shape
-        // match from the tail because constant inputs will have types in the front
-        auto calleeInputType =
-            callee.getFunctionType().getInput(callee.getFunctionType().getNumInputs() - 1 - i);
-        auto gradRtype = grad_types[grad_types.size() - 1 - i];
+    // Each gradient result type must match the type of its corresponding
+    // differentiated operand (indexed by diffArgIndices).
+    for (size_t i = 0; i < diffArgIndices.size(); i++) {
+        auto calleeInputType = callee.getFunctionType().getInput(diffArgIndices[i]);
+        auto gradRtype = grad_types[i];
```
dime10, line 227: Can you test a function with multiple results as well? If I remember correctly there are `num_diff_args x num_results` many gradients, but it looks like here we assume there are only `num_diff_args` many gradients. 

The gradient type would normally also not be exactly the argument type, but a [combination of the arg and result type](https://github.com/PennyLaneAI/catalyst/blob/27ff184a568ecff61d8ad6c5cfd5f789ce21f642/frontend/catalyst/utils/calculate_grad_shape.py#L93). But I guess since this is the ValueAndGradOp, which we don't define for Jacobians (whereas the GradOp is), we can expect scalar outputs only in which case the gradient type is trivially the argument type.

sengthai: > If I remember correctly there are num_diff_args x num_results many gradients, but it looks like here we assume there are only num_diff_args many gradients.

Good point about this. However, `ValueAndGradOp` is different where the frontend enforces scalar output only(`scalar_out=True`) in [here](https://github.com/PennyLaneAI/catalyst/blob/f4a946bc28f68afdc8a25dd6203a278211342fea/frontend/catalyst/api_extensions/differentiation.py#L300). So, the `num_results` is always 1. 

I had added tests covering multiple diff args in [392ebca](https://github.com/PennyLaneAI/catalyst/pull/2349/commits/392ebcaa19617ab21c28e994db5ec04b3ff26b09)

dime10: Just because the output values have to be scalars doesn't mean there is only one output though!

sengthai: Umm, you have a point, but I'm not quite sure if I understand this well. 
Look at this line:
`if not (len(jaxpr.out_avals) == 1 and jaxpr.out_avals[0].shape == ()):` in frontend [here](https://github.com/PennyLaneAI/catalyst/blob/f4a946bc28f68afdc8a25dd6203a278211342fea/frontend/catalyst/api_extensions/differentiation.py#L823),
it seems to me that we enforce exactly one output if it is scalar output.  🤔

paul0403: Will vjp and jvp ever dispatch to `ValueAndGradOp`? Because if it's just `grad` and `value_and_grad`, I think the line Sengthai linked to does suggest that they are constrained to return a single scalar.

dime10: 😮 Good point, you might be right!

My bad, it looks like we indeed restrict catalyst.grad (and thus value_and_grad) to _single_ scalar outputs only. All good then 👍

paul0403: Since this check in the frontend has been untouched for 2 years, it must have been the case that `grad` and `value_and_grad` has only allowed a single scalar for all those 2 years! (`value_and_grad` sends `scalar_out=True` in its `GradCallable`). 

I wasn't here when the gradient stuff was first added, but it was probably intentional to clearly separate functionality between `grad` and the vjps?

PR #2351, opened 2026-01-07
v0.14.0 changelog

isaacdevlugt:
**Context:** Changelog editing for v0.14 release

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:** [sc-99379 (epic)](https://app.shortcut.com/xanaduai/epic/99379?group_by=none&hide_subtasks=false&vc_group_by=day&ct_workflow=all&cf_workflow=500000005)

### DIFF
--- a/doc/dev/sharp_bits.rst
+++ b/doc/dev/sharp_bits.rst
@@ -1302,7 +1302,7 @@ Currently, however, this is not the case for the following functionalities.
     ignored in Catalyst. The reason is that the only supported mode is to always allocate in the
     zero state.
 
-  - Related to the above point, in PennyLane, dynamic wire allocations do not 
+  - Related to the above point, in PennyLane, dynamic wire allocations do not
     increase the total number of wires used in the circuit. This is because
     PennyLane treats the number of wires during device initialization (the 
     ``qml.device("...", wires=N)``) as the device capacity. Briefly, when 
@@ -1331,4 +1331,40 @@ Currently, however, this is not the case for the following functionalities.
         return qml.probs(wires=[0])
 
   >>> print(circuit())
-    NotImplementedError: Dynamically allocated wires cannot be used in quantum adjoints yet.
+  NotImplementedError: Dynamically allocated wires cannot be used in quantum adjoints yet.
+
+  - Usage of ``qml.allocate()`` with Catalyst prohibits returning any terminal measurement
+    *except for* ``qml.probs``. This is due to a bug in Lightning.
+    Other measurement types (``qml.sample``, ``qml.expval``, ``qml.var``, etc.)
+    will be supported in the future after the underlying bug is fixed.
+
+  .. code-block:: python
+
+    qml.capture.enable()
+
+    @qjit
+    @qml.qnode(qml.device("lightning.qubit", wires=3, shots=10))
+    def circuit():
+        with qml.allocate(2) as qs:
+            qml.X(qs[1])
+        return qml.sample(wires=[0, 1])
+
+  >>> circuit()
+  CompileError: Only probability measurements (qml.probs) are currently supported
+  when dynamic allocations are present in the program. Other measurement
+  types (qml.sample, qml.expval, qml.var, ...etc) will be supported
+  in a future release after the underlying bug is fixed.
+
+  .. code-block:: python
+
+    qml.capture.enable()
+
+    @qml.qjit
+    @qml.qnode(qml.device("lightning.qubit", wires=1))
+    def circuit():
+        with qml.allocate(1) as q:
+            qml.X(q[0])
+        return qml.probs(wires=[0])
+
+  >>> circuit()
+  Array([1., 0.], dtype=float64)
--- a/frontend/catalyst/from_plxpr/qfunc_interpreter.py
+++ b/frontend/catalyst/from_plxpr/qfunc_interpreter.py
@@ -254,6 +254,20 @@ class PLxPRToQuantumJaxprInterpreter(PlxprInterpreter):
                         """
                     )
                 )
+            # Only probs measurements are currently supported with dynamic allocation
+            # due to a bug in Lightning's PartialSample implementation
+            # TODO: Remove this once the bug is fixed
+            if not isinstance(measurement, qml.measurements.ProbabilityMP):
+                raise CompileError(
+                    textwrap.dedent(
+                        """
+                        Only probability measurements (qml.probs) are currently supported
+                        when dynamic allocations are present in the program. Other measurement
+                        types (qml.sample, qml.expval, qml.var, ...etc) will be supported
+                        in a future release after the underlying bug is fixed.
+                        """
+                    )
+                )
 
         if type(measurement) not in measurement_map:
             raise NotImplementedError(
--- a/frontend/test/pytest/test_dynamic_qubit_allocation.py
+++ b/frontend/test/pytest/test_dynamic_qubit_allocation.py
@@ -611,5 +611,39 @@ def test_unsupported_adjoint(backend):
             return qml.probs(wires=[0, 1])
 
 
+@pytest.mark.usefixtures("use_capture")
+@pytest.mark.parametrize(
+    "measurement_fn, shots",
+    [
+        (lambda: qml.expval(qml.Z(0)), None),
+        (lambda: qml.var(qml.Z(0)), None),
+        (lambda: qml.sample(wires=[0]), 10),
+    ],
+)
+def test_non_probs_measurement_with_dynamic_wires(backend, measurement_fn, shots):
+    """
+    Test that an error is raised when using non-probs measurements with dynamic wire allocations.
+    """
+
+    with pytest.raises(
+        CompileError,
+        match=textwrap.dedent(
+            """
+            Only probability measurements \\(qml.probs\\) are currently supported
+            when dynamic allocations are present in the program. Other measurement
+            types \\(qml.sample, qml.expval, qml.var, ...etc\\) will be supported
+            in a future release after the underlying bug is fixed.
+            """
+        ),
+    ):
+
+        @qjit
+        @qml.qnode(qml.device(backend, wires=1, shots=shots))
+        def circuit():
+            with qml.allocate(1) as q:
+                qml.X(q[0])
+            return measurement_fn()
+
+
 if __name__ == "__main__":
     pytest.main(["-x", __file__])


doc/releases/changelog-0.14.0.md
```diff
@@ -262,114 +300,163 @@
...
   [(#2165)](https://github.com/PennyLaneAI/catalyst/pull/2165)
 
-* `qml.grad` and `qml.jacobian` can now be used with `qjit` when program capture is enabled.
+* ``qml.grad`` and ``qml.jacobian`` can now be used with ``qjit`` when PennyLane's program capture 
+  is enabled.
   [(#2078)](https://github.com/PennyLaneAI/catalyst/pull/2078)
 
-* Dynamically allocated wires can now be passed into control flow and subroutines.
+* Dynamically allocated wires (:func:`pennylane.allocate`) can now be passed into control flow 
+  blocks and subroutines.
```
isaacdevlugt, line 352: Is ``subroutine`` not supposed to be user-facing? 🤔 I was going to put an example in this changelog entry that was a copy-paste of a test that Paul wrote, but realized that ``subroutine`` is from ``catalyst.jax_primitives``

doc/releases/changelog-0.14.0.md
```diff
@@ -423,31 +512,34 @@
...
 
-* Update `catalyst.python_interface.inspection.xdsl_conversion.get_mlir_module` to
-  no longer explicitly use the xDSL pass plugin.
+* Updated ``catalyst.python_interface.inspection.xdsl_conversion.get_mlir_module`` to
+  no longer explicitly use the xDSL pass plugin, which was causing an issue with the new ``specs``
+  functionality introduced in this release. 
   [(#2303)](https://github.com/PennyLaneAI/catalyst/pull/2303)
 
-* Added missing `detensorize-function-boundary` and `symbol-dce` passes to `Pipelines.cpp`.
+* Added missing ``detensorize-function-boundary`` and ``symbol-dce`` passes to ``Pipelines.cpp``.
```
isaacdevlugt: What's the benefit of this?

doc/releases/changelog-0.14.0.md
```diff
@@ -494,13 +589,12 @@
...
-  moving nullary operations. This also improves the compilation time by reducing
-  the sort function, by explicitly passing the operations that need to be sorted.
+* Fixed an issue with :func:`~.passes.commute_ppr` and :func:`~.passes.merge_ppr_ppm` where they
+  were incorrectly moving nullary operations. This also improves the compilation time by reducing
+  the sort function by explicitly passing the operations that need to be sorted.
   [(#2200)](https://github.com/PennyLaneAI/catalyst/pull/2200)
 
-* The pass pipeline is correctly registered to the transform named sequence of the
-  one-shot qnode when `one-shot` mcm method is used.
+* Fixed a bug that was causing compilation passes to not apply when using ``mcm_method="one-shot"``.
```
isaacdevlugt, line 620: I think this is correct but let me know otherwise

doc/releases/changelog-0.14.0.md
```diff
@@ -494,51 +475,161 @@
...
+    [(#2183)](https://github.com/PennyLaneAI/catalyst/pull/2183)
+
+  * The ``mlir_opt`` property now correctly handles xDSL passes by automatically
+    detecting when the Python compiler is being used and routing through it appropriately.
+    [(#2190)](https://github.com/PennyLaneAI/catalyst/pull/2190)
+
+  * A new experimental ``parity_synth_pass`` compiler pass has been added to
+    ``catalyst.python_interface.transforms``. This pass groups ``CNOT`` and ``RZ`` operators
+    into phase polynomials and re-synthesizes them into ``CNOT`` and ``RZ`` operators again.
+    [(#2294)](https://github.com/PennyLaneAI/catalyst/pull/2294)
```
isaacdevlugt: Should we add an example here?

doc/releases/changelog-0.14.0.md
paul0403: I think `@subroutine` is user-facing!

We probably just forgot to expose it at top-level :sweat_smile:

doc/releases/changelog-0.14.0.md
paul0403: Yes! This is correct (and much better wording)

doc/releases/changelog-0.14.0.md
```diff
@@ -650,39 +768,21 @@
...
-  `--pass-pipeline`.
+* Updated the MLIR Plugins documentation stating that plugins require adding passes via 
+  ``--pass-pipeline``.
   [(#2168)](https://github.com/PennyLaneAI/catalyst/pull/2168)
 
-* Typos in docstring for `PPRotationArbitraryOp` and `PPRRotationOp` have been corrected.
+* Typos in the docstrings for ``PPRotationArbitraryOp`` and ``PPRRotationOp`` have been corrected.
   [(#2297)](https://github.com/PennyLaneAI/catalyst/pull/2297)
 
 <h3>Contributors ✍️</h3>
```
isaacdevlugt, line 806: TODO: carve out dedicated section for external contributions / community contributions

doc/releases/changelog-0.14.0.md
```diff
@@ -227,15 +163,15 @@
   [(#2284)](https://github.com/PennyLaneAI/catalyst/pull/2284)
```
isaacdevlugt, line 163:
```suggestion
  [(#2284)](https://github.com/PennyLaneAI/catalyst/pull/2284)
  [(#2296)](https://github.com/PennyLaneAI/catalyst/pull/2296)
```

doc/releases/changelog-0.14.0.md
paul0403: The two catalyst default pipelines registered (one in https://github.com/PennyLaneAI/catalyst/blob/main/frontend/catalyst/pipelines.py, one in https://github.com/PennyLaneAI/catalyst/blob/main/mlir/lib/Driver/Pipelines.cpp) now agree with each other.

doc/releases/changelog-0.14.0.md
dime10: No it never left the prototype stage, so it's not documented or properly tested. It is an item on the IR improvements proposal though.

doc/releases/changelog-0.14.0.md
dime10: Just fixing an inconsistency. If you were to run the Catalyst compiler from the commandline (not from Python), it would have been missing those two passes. Should definitely reword this though

doc/releases/changelog-0.14.0.md
paul0403: > No it never left the prototype stage

:astonished:

doc/releases/changelog-0.14.0.md
```diff
@@ -225,17 +161,18 @@
...
-  :func:`catalyst.passes.ppm_compilation`. For clear and inspectable results, use ``target="mlir"``
-  in the ``qjit`` decorator, ensure that PennyLane's program capture is enabled,
-  :func:`pennylane.capture.enable`, and call the Catalyst passes from the PennyLane frontend (e.g.,
-  ``qml.transforms.ppr_to_ppm`` instead of from ``catalyst.passes.``).
+  :func:`catalyst.passes.ppr_to_ppm`, :func:`catalyst.passes.reduce_t_depth`, 
+  :func:`catalyst.passes.decompose_arbitrary_ppr` and :func:`catalyst.passes.ppm_compilation`.
+  
+  For most use cases, it is recommended to ensure that PennyLane's program capture is enabled, 
+  :func:`pennylane.capture.enable`, and to call the above compilation passes from the PennyLane 
+  frontend (e.g., ``qml.transforms.ppr_to_ppm`` instead of from ``catalyst.passes.ppr_to_ppm``).
```
dime10: Just curious, is there a difference between the two?

doc/releases/changelog-0.14.0.md
```diff
@@ -244,7 +181,7 @@
 
   qml.capture.enable()
 
-  @qjit(target="mlir")
+  @qml.qjit(target="mlir")
   @partial(qml.transforms.ppm_compilation, decompose_method="auto-corrected")
```
dime10: We really need to get rid of this partial everywhere 😅  (ignore)

doc/releases/changelog-0.14.0.md
```diff
@@ -265,129 +202,179 @@
 
       ppm = qml.pauli_measure(pauli_word="YYZ", wires=[0, 2, 1])
 
-      return
+      return qml.expval(qml.Z(0))
```
dime10: There was a reason we didn't put a measurement process in the example no, because it's not really integrated into the PBC framework yet?

doc/releases/changelog-0.14.0.md
```diff
@@ -265,129 +202,179 @@
...
-      gate_types=defaultdict(<class 'int'>, {'PPM-w3': 2, 'PPM-w2': 4, 'PPM-w1': 4, 'PPR-pi/2-w1': 4}),
-      gate_sizes=defaultdict(<class 'int'>, {3: 2, 2: 4, 1: 8}),
-      depth=None,
-      shots=Shots(total_shots=None, shot_vector=())
-    )
-  }
+  >>> print(qml.specs(circuit, level=2)())
+  Device: null.qubit
+  Device wires: 3
+  Shots: Shots(total=None)
```
dime10: Wow I didn't know this printed so nicely, is that new?

doc/releases/changelog-0.14.0.md
```diff
@@ -265,129 +202,179 @@
...
+      PPR-pi/2: 6
+      qec.fabricate: 1
+
+    Measurements:
+      expval(PauliZ): 1
   ```
 
-* A new transform called :func:`~catalyst.passes.decompose_arbitrary_ppr` pass has been added, 
-  which decomposes abitrary-angle Pauli product rotations (PPRs) as outlined in Figure 13(d) from 
+  As the circuits generated by these pass are currently not executable on any backend, ensure that
```
dime10:
```suggestion
  As the circuits generated by these passes are currently not executable on any backend, ensure that 
```

doc/releases/changelog-0.14.0.md
```diff
@@ -265,129 +202,179 @@
...
+    Measurements:
+      expval(PauliZ): 1
   ```
 
-* A new transform called :func:`~catalyst.passes.decompose_arbitrary_ppr` pass has been added, 
-  which decomposes abitrary-angle Pauli product rotations (PPRs) as outlined in Figure 13(d) from 
+  As the circuits generated by these pass are currently not executable on any backend, ensure that 
+  ``target="mlir"`` is in the ``qjit`` decorator for clear and inspectable results. Compiled 
+  circuits expressed in terms of :class:`~.PauliRot` and :func:`~.pauli_measure` for Pauli-based 
+  computation can only be analyzed with the ``null.qubit`` device.
```
dime10: How come this doesn't work with any device?

doc/releases/changelog-0.14.0.md
```diff
@@ -265,129 +202,179 @@
...
-* Resource tracking now tracks calls to `SetState` and `SetBasisState`, and can report results
-  that include `qml.StatePrep` operations.
+* Resource tracking with :func:`pennylane.specs` now includes ``qml.StatePrep`` operations.
   [(#2230)](https://github.com/PennyLaneAI/catalyst/pull/2230)
 
-* `qml.PCPhase` can be compiled and executed with capture enabled.
-  [(#2226)](https://github.com/PennyLaneAI/catalyst/pull/2226)
-
-* Resource tracking now supports dynamic qubit allocation
+* Resource tracking with :func:`pennylane.specs` now tracks dynamic wire allocation
```
dime10: nit: Combine with the above?

doc/releases/changelog-0.14.0.md
```diff
@@ -265,129 +202,179 @@
...
 
-* The ``decompose-lowering`` MLIR pass now supports ``qml.MultiRZ``
-  with an arbitrary number of wires. This decomposition is performed
-  at MLIR when both capture and graph-decomposition are enabled.
+* The decomposition of ``qml.MultiRZ`` operations with an arbitrary number of wires is now supported
+  at the MLIR level with graph-based decompositions enabled and PennyLane's program capture enabled. 
   [(#2160)](https://github.com/PennyLaneAI/catalyst/pull/2160)
 
-* A new option ``use_nameloc`` has been added to :func:`~.qjit` that embeds variable names
+* A new option called ``use_nameloc`` has been added to :func:`~.qjit` that embeds variable names
```
dime10: Now that I think about it, we should be able to provide a better variable name 🤔 Maybe something like `embed_var_names`?

doc/releases/changelog-0.14.0.md
```diff
@@ -265,129 +202,179 @@
...
   [(#2054)](https://github.com/PennyLaneAI/catalyst/pull/2054)
 
-* Passes registered under `qml.transform` can now take in options when used with
-  :func:`~.qjit` with program capture enabled.
+* Compilation passes registered in PennyLane via ``@qml.transform`` can now take in optional keyword
+  arguments when used with :func:`~.qjit` and when PennyLane's program capture is enabled.
   [(#2154)](https://github.com/PennyLaneAI/catalyst/pull/2154)
 
-* Pytree inputs can now be used when program capture is enabled.
+* Pytree inputs can now be used when PennyLane's program capture is enabled.
```
dime10, line 317: I think it would also be nice to regroup the program capture improvements, since there are quite a few.

doc/releases/changelog-0.14.0.md
```diff
@@ -402,10 +389,11 @@
   update their code to use the new names.
```
dime10: I think the wording here is a bit misleading, as the renaming by itself wouldn't really make sense. Instead, the pipelines were restructured, such that existing quantum passes in the `quantum-compilation-pipeline` are now run together with user provided quantum passes (previously run in `enforce-runtime-invariants`). This new combined pipeline is now run first. A new gradient pipeline was split out from the old quantum-compilation-pipeline.

doc/releases/changelog-0.14.0.md
```diff
@@ -402,10 +389,11 @@
...
   [(#2186)](https://github.com/PennyLaneAI/catalyst/pull/2186)
 
-  * The ``pipeline``  and "passes" postfixes in the compilation stage names have been changed to ``stage``.
+* The ``pipeline``  and ``"passes"`` postfixes in the compilation stage names have been changed to 
+  ``stage``.
```
dime10:
```suggestion
  ``stage`` for clarity.
```

doc/releases/changelog-0.14.0.md
```diff
@@ -402,10 +389,11 @@
...
   [(#2230)](https://github.com/PennyLaneAI/catalyst/pull/2230)
 
-* The plxpr transform `pl_map_wires` has been removed along with its test.
+* The jaxpr transform ``pl_map_wires`` has been removed along with its test.
```
dime10: Was this ever user facing? I would put it in internal

doc/releases/changelog-0.14.0.md
```diff
@@ -426,58 +414,78 @@
   - The Enzyme version has been updated to
```
dime10, line 517: remove 'with the build target `EnzymeStatic-22`' from above, TMI

dime10: Actually that whole sentence can be removed since the versions are already listed here

dime10: Also I don't think we need to inform people that nanobind has gone up in minimum version, since that change was to the build system only.
The reason we inform people of the llvm version is because if they integrate with Catalyst at the MLIR level they might have a conflicting LLVM version for their project.

doc/releases/changelog-0.14.0.md
```diff
@@ -426,58 +414,78 @@
...
+  prevented an instance from being constructed.
+  [(#8456)](https://github.com/PennyLaneAI/pennylane/pull/8456)
+
+* Fixed a bug where the ``qec.ppr`` op attribute ``rotation_kind`` was not correctly constrained to
+  be one of ``±1``, ``±2``, ``±4``, or ``±8``. Also, for the identity Pauli product, the 
+  ``rotation_kind`` was correctly set to ``1``, instead of ``0``.
   [(#2344)](https://github.com/PennyLaneAI/catalyst/pull/2344)
 
-* Updated tests and PennyLane dependency pin to follow changes introduced by
+* Updated tests and the PennyLane dependency pin to follow changes introduced by
```
dime10: I don't think we need to inform users every time we update the PL-dev pin during a development cycle. In fact, the linked PR does not contain any changes to the released package, so we should remove this entry.

doc/releases/changelog-0.14.0.md
```diff
@@ -426,58 +414,78 @@
...
   [PennyLaneAI/pennylane#8290](https://github.com/PennyLaneAI/pennylane/pull/8290).
   [(#2286)](https://github.com/PennyLaneAI/catalyst/pull/2286)
 
-* Update `catalyst.python_interface.inspection.xdsl_conversion.get_mlir_module` to
-  no longer explicitly use the xDSL pass plugin.
+* Updated ``catalyst.python_interface.inspection.xdsl_conversion.get_mlir_module`` to
+  no longer explicitly use the xDSL pass plugin, which was causing an issue with the new ``specs``
+  functionality introduced in this release.
```
dime10: Just add PR link to existing specs entry since it says it's from this release.

doc/releases/changelog-0.14.0.md
```diff
@@ -497,51 +505,161 @@
...
 <h3>Internal changes ⚙️</h3>
 
-* `DecompRuleInterpreter` now solves the graph and adds decompositions rules in the `cleanup` method
-  instead of during the first call to `interpret_measurement`.
+* Catalyst now features a unified compilation framework, which will enable users and developers to 
+  design and implement compilation passes in Python *in addition to* C++, acting on the same 
+  Catalyst IR. The Python interface relies on the `xDSL library <https://xdsl.dev/>` to represent 
+  and manipulate programs (analogous to the MLIR library in C++). As a result, transformations can 
+  be quickly prototyped, easily debugged, and dynamically integrated into Catalyst without changes 
+  to the compiled Catalyst package.
```
dime10: Why is this an internal change, I would have thought this is something we want to advertise widely?

doc/releases/changelog-0.14.0.md
dime10: Then it can't be internal. Internal changes are things that shouldn't affect users in any way but we include them anyway because they are code changes to the released package.
An example would indicate there is an actual user-facing use to this.

doc/releases/changelog-0.14.0.md
```diff
@@ -497,51 +505,161 @@
   // ... use %4
   ```
 
-* Fixes :func:`~.passes.commute_ppr` and :func:`~.passes.merge_ppr_ppm` incorrectly
-  moving nullary operations. This also improves the compilation time by reducing
-  the sort function, by explicitly passing the operations that need to be sorted.
+* Fixed an issue with :func:`~.passes.commute_ppr` and :func:`~.passes.merge_ppr_ppm` where they
+  were incorrectly moving nullary operations. This also improves the compilation time by reducing
```
dime10: What are nullary operations?

doc/releases/changelog-0.14.0.md
```diff
@@ -426,58 +414,78 @@
...
+  modified by ``adjoint`` or ``ctrl``.
   [(##2158)](https://github.com/PennyLaneAI/catalyst/pull/2158)
 
-* Fixes the translation of a workflow with different transforms applied to different qnodes.
+* Fixed an issue with the translation of a workflow with different transforms applied to different 
+  QNodes, which was causing transforms to act beyond the code they are intended to be applied to.
   [(#2167)](https://github.com/PennyLaneAI/catalyst/pull/2167)
 
-* Fix canonicalization of eliminating redundant `quantum.insert` and `quantum.extract` pairs.
+* Fixed canonicalization of eliminating redundant ``quantum.insert`` and ``quantum.extract`` pairs.
```
dime10:
```suggestion
* Fixed canonicalization of redundant ``quantum.insert`` and ``quantum.extract`` pairs.
```

dime10 (changes requested):
Thanks Isaac, the main sections at the top are already in excellent shape 🥇 

A few things I noticed further down:

doc/releases/changelog-0.14.0.md
paul0403: It does! (I just tried with lightning)

doc/releases/changelog-0.14.0.md
isaacdevlugt: In terms of functionalitY? Nope. The PL transform is just replacing the ``PipelineWrapper`` thing in Catalyst. So, however that was functioning before, this should be the exact same

doc/releases/changelog-0.14.0.md
isaacdevlugt: should be doable with this release!

doc/releases/changelog-0.14.0.md
isaacdevlugt: We can return `qml.measure`... but that's not the way we want to do things going forward 😅. Idk, having empty returns looks/feels very strange.

doc/releases/changelog-0.14.0.md
isaacdevlugt: yes!

doc/releases/changelog-0.14.0.md
isaacdevlugt: good point!

doc/releases/changelog-0.14.0.md
isaacdevlugt: Kind of, but internal change is a-ok with me

doc/releases/changelog-0.14.0.md
isaacdevlugt:
```suggestion
* Running the Catalyst compiler from the command line no longer misses the ``detensorize-function-boundary`` and ``symbol-dce`` passes.
```

How's this?

doc/releases/changelog-0.14.0.md
isaacdevlugt: I assume operations that don't do anything?

doc/releases/changelog-0.14.0.md
isaacdevlugt: Not yet, I would argue. For a couple reasons: 

- no documentation for this module 💀 
- no features yet that require the user to know about xDSL directly. All features in the release so far can be strictly explained in a way that doesn't require knowledge of the python interface to the compiler. 

To me, at the moment, the python interface to the compiler is still very much internal until there is documentation and a better API for creating transforms.

doc/releases/changelog-0.14.0.md
isaacdevlugt: nvm, this will go in the blog

doc/releases/changelog-0.14.0.md
dime10: I mean for the user applying the transform, using `qml.transforms.ppr_to_ppm` vs `catalyst.passes.ppr_to_ppm` is the same? What is the recommendation based on then?

doc/releases/changelog-0.14.0.md
dime10: What is strange about it if you are only interested in manipulating the circuit rather than compute any values?^^

doc/releases/changelog-0.14.0.md
paul0403: Isaac and I chatted about this today, we decided that `qml.transforms.ppr_to_ppm` will reside in the PL docs https://github.com/PennyLaneAI/pennylane/pull/8783, and catalyst docs will just have the catalyst frontend, aka `catalyst.passes.ppr_to_ppm`.

Originally I was also on the side of "we should advertise `qml.blah`", but what swayed me to the other side was that, even though we have things like control flow intergrated in plxpr path now, we still show `catalyst.for_loop` on the documentation.

paul0403: I think regarding documentation, eventually what should happen is:
- PL documentation is for frontend
- catalyst documentation is purely for compiler (i.e. just the dialects) (maybe runtime, but the future of runtime is not clear now)

But that requires capture work / maybe the new direct lowering of plxpr to ref dialect to complete. In the meantime, maybe it's still a good idea to document catalyst's own frontend.

doc/releases/changelog-0.14.0.md
dime10: Hmm, documentation is definitely a good point. Just weird to have a whole new feature / module in the section that is meant for refactors or smaller changes without measurable user impact.

doc/releases/changelog-0.14.0.md
paul0403: Also, https://github.com/PennyLaneAI/catalyst/pull/2296

doc/releases/changelog-0.14.0.md
dime10: I think we need a different term here

doc/releases/changelog-0.14.0.md
isaacdevlugt: > mean for the user applying the transform, using qml.transforms.ppr_to_ppm vs catalyst.passes.ppr_to_ppm is the same? What is the recommendation based on then?

Better integration with the program capture pipeline, mostly 😅. But, also to move people to the PL frontend for the reasons Paul said. The PL documentation should be the source of truth for users (where possible), catalyst documentation should eventually just be for compiler stuff

doc/releases/changelog-0.14.0.md
isaacdevlugt: It just looks weird is all. QNodes have to return something, and the user has to write a QNode to use the transforms. So we're forcing / encouraging them to write something that we don't really allow 🤔

doc/releases/changelog-0.14.0.md
```diff
@@ -396,16 +381,18 @@
...
-* The MLIR pipeline ``enforce-runtime-invariants-pipeline`` has been renamed to
-  ``quantum-compilation-pipeline`` and the old ``quantum-compilation-pipeline`` has been renamed to
-  ``gradient-lowering-pipeline``. Users who referenced these pipeline names directly would need to
-  update their code to use the new names.
+* The MLIR pipeline ``enforce-runtime-invariants-pipeline`` has been restructured and a new pipeline
+  called ``quantum-compilation-pipeline`` has been added. Existing quantum compilation passes in the 
+  ``quantum-compilation-pipeline`` are now run together with user-provided quantum compilation 
+  passes (previously run in ``enforce-runtime-invariants``). This new combined pipeline now runs 
+  first.
   [(#2186)](https://github.com/PennyLaneAI/catalyst/pull/2186)
```
isaacdevlugt: @dime10 how's this?

doc/releases/changelog-0.14.0.md
isaacdevlugt: How's that?

doc/releases/changelog-0.14.0.md
isaacdevlugt: Let's just remove the word!

doc/releases/changelog-0.14.0.md
isaacdevlugt: How about improvements then?

doc/releases/changelog-0.14.0.md
```diff
@@ -426,58 +409,69 @@
...
-  Also, for Identity Pauli product, the `rotation_kind` was correctly set to 1, instead of 0.
-  [(#2344)](https://github.com/PennyLaneAI/catalyst/pull/2344)
+* Fixed a bug in the ``catalyst.python_interface.pass_api.TranformInterpreterPass`` pass that 
+  prevented pass options from being used correctly.
+  [(#2289)](https://github.com/PennyLaneAI/catalyst/pull/2289)
 
-* Updated tests and PennyLane dependency pin to follow changes introduced by
-  [PennyLaneAI/pennylane#8290](https://github.com/PennyLaneAI/pennylane/pull/8290).
-  [(#2286)](https://github.com/PennyLaneAI/catalyst/pull/2286)
+* The experimental xDSL :func:`~catalyst.python_interface.transforms.diagonalize_measurements`
```
lazypanda10117, line 550: Curious why we have Pennylane PR in the Catalyst changelog?

doc/releases/changelog-0.14.0.md
```diff
@@ -426,58 +409,69 @@
...
-* The pass `remove-chained-self-inverse` has been renamed to `cancel-inverses`, to better
+* The ``remove-chained-self-inverse`` pass has been renamed to ``cancel-inverses`` to better
   conform with the name of the corresponding transform in PennyLane.
   [(#2201)](https://github.com/PennyLaneAI/catalyst/pull/2201)
 
 <h3>Deprecations 👋</h3>
 
+No deprecations have been made in this release.
+
 <h3>Bug fixes 🐛</h3>
```
lazypanda10117, line 534: * Fixed a bug in the `merge_ppr_ppm` that was causing an iteration out-of-bound error.
[(#2359)](https://github.com/PennyLaneAI/catalyst/pull/2359)

doc/releases/changelog-0.14.0.md
dime10: Sounds good, I'm a bit confused because it sounds like Paul is saying the catalyst docs (i.e. this changelog) should use the catalyst functions

> Isaac and I chatted about this today, we decided that qml.transforms.ppr_to_ppm will reside in the PL docs https://github.com/PennyLaneAI/pennylane/pull/8783, and catalyst docs will just have the catalyst frontend, aka catalyst.passes.ppr_to_ppm.

but I'll leave it to you to figure out if you want to make any edits here.

doc/releases/changelog-0.14.0.md
dime10: But the PPR commutation is not going to work properly, and it is being run in this example.

doc/releases/changelog-0.14.0.md
dime10: Not sure what your question is, I'm just saying there is too much info here and we should remove this sentence: 
```
Enzyme now targets `v0.0.203` with the build target `EnzymeStatic-22`, and the nanobind
  requirement for the latest LLVM has been updated to version 2.9.
```

(can't suggest it directly as it's out of range)

doc/releases/changelog-0.14.0.md
dime10: Almost as weird since it's large and entirely new to catalyst, but i'll let you make the call.

doc/releases/changelog-0.14.0.md
dime10: I think it's because at some point we moved all the xDSL code to Catalyst, and the PRs that had gone in for it this development cycle are now listed here 🤔

doc/releases/changelog-0.14.0.md
dime10: Better!

doc/releases/changelog-0.14.0.md
```diff
@@ -396,24 +381,22 @@
...
 
-* The MLIR pipeline ``enforce-runtime-invariants-pipeline`` has been renamed to
-  ``quantum-compilation-pipeline`` and the old ``quantum-compilation-pipeline`` has been renamed to
-  ``gradient-lowering-pipeline``. Users who referenced these pipeline names directly would need to
-  update their code to use the new names.
+* The MLIR pipeline ``enforce-runtime-invariants-pipeline`` has been restructured and a new pipeline
+  called ``quantum-compilation-pipeline`` has been added. Existing quantum compilation passes in the 
+  ``quantum-compilation-pipeline`` are now run together with user-provided quantum compilation 
+  passes (previously run in ``enforce-runtime-invariants``). This new combined pipeline now runs 
+  first.
```
dime10: To be precise:
```suggestion
* The standard Catalyst pipelines have been restructured, such that default and user QNode passes
  are run together in the first pipeline. For this purpose, the old ``quantum-compilation-pipeline``
  and ``enforce-runtime-invariants-pipeline`` have been merged into a single
  ``quantum-compilation-pipeline``, while a new ``gradient-lowering-pipeline`` has been split out
  from the old ``quantum-compilation-pipeline``. 
```

doc/releases/changelog-0.14.0.md
isaacdevlugt: > it sounds like Paul is saying the catalyst docs (i.e. this changelog) should use the catalyst functions

Oh true! Didn't realize this. I'll update the doc example.

doc/releases/changelog-0.14.0.md
isaacdevlugt: Yeah I removed the sentence haha

doc/releases/changelog-0.14.0.md
isaacdevlugt:
```suggestion
* Fixed a bug in the :func:`catalyst.passes.merge_ppr_ppm` that was causing an iteration out-of-bound error.
  [(#2359)](https://github.com/PennyLaneAI/catalyst/pull/2359)
```

doc/releases/changelog-0.14.0.md
isaacdevlugt: @mudit2812 can confirm

doc/releases/changelog-0.14.0.md
isaacdevlugt: Sounds like there is no ideal example here. I suppose in cases where ``commute`` PPR is being invoked (either directly or via ``ppm_compilation``), the best option is to return ``qml.measure`` (or ``catalyst.measure`` here) even though it's not a pattern we want to support longterm. I'm okay with that. How about you?

doc/releases/changelog-0.14.0.md
isaacdevlugt: I'll go with improvements! The PennyLane changelog has new features in the improvements section, they're just more minor or things that we don't feel are necessary to give a lot of visibility to (e.g., program capture features). I think that precedent is fine to apply here.

doc/releases/changelog-0.14.0.md
dime10: Returning `qml.measure` is fine by me 🙂 (although I have absolutely no issue returning nothing either since these workflows are not about computing anything but compiling instructions)

lazypanda10117: We cannot return `qml.measure`, it is not supported with capture enabled.

dime10: Right, we have gone over this before 😅 Returning nothing is the only reasonable thing at the moment

isaacdevlugt: I have a dozen things to keep track of right now sorry 😅 

For examples with ppm_compilation and commute_ppr, I'll return nothing!

doc/releases/changelog-0.14.0.md
mudit2812: Yep, exactly what @dime10 said. Most of the entries under the "Unified Compiler" project in the changelog will link to pull requests in PennyLane.

doc/releases/changelog-0.14.0.md
lazypanda10117: and also for `merge_ppr_ppm`

doc/releases/changelog-0.14.0.md
isaacdevlugt: @dime10 this should be tracked somewhere else 😅. You ok if I resolve this?

doc/releases/changelog-0.14.0.md
```diff
@@ -694,6 +782,7 @@ This release contains contributions from (in alphabetical order):
 Ali Asadi,
 Joey Carter,
 Yushao Chen,
+Isaac De Vlugt,
 Sengthai Heng,
```
isaacdevlugt, line 814:
```suggestion
Sengthai Heng,
David Ittah,
```

‼️

doc/releases/changelog-0.14.0.md
dime10: Fair 👍

doc/releases/changelog-0.14.0.md
```diff
@@ -263,157 +190,318 @@
...
+  Shots: Shots(total=None)
+  Level: 2
+
+  Resource specifications:
+    Total wire allocations: 18
+    Total gates: 52
+    Circuit depth: Not computed
+
+    Gate types:
+      MultiRZ: 7
```
dime10: Sorry to come back to this, but the output is confusing. Why are there MultiRZ gates in here when the input is already a PBC circuit?

dime10: The other question is what is the intention of this example, what do we want to show?  The use of specs is not really clear to me.

dime10: If the intention is to show that paulirot and pauli_measure can now be used to express PBC circuits directly from the frontend, I would suggest to at least add a pauli_measure to the example, and remove the expval (for the same reason discussed in the other example).
Then the question is what to we want to show. One option is just show the program in the qec dialect, but the example also uses PBC passes, so do we want to show the effect of those?

dime10 (approved):

Issue #2356, opened 2026-01-07
qml.pauli_measure + qml.PauliRot crash on single pauli measurement


kipawaa:
Combining `decompose-arbitrary-ppr` and `merge-ppr-ppm` crashes when applying a 3+ qubit arbitrary PPR or pi / 2 PPR before a single gate PPM:
```python
qml.capture.enable()

pipeline = [("pipe", ["quantum-compilation-stage"])]

@qml.qjit(pipelines=pipeline, target="mlir")
@qml.transform(pass_name="merge-ppr-ppm")
@qml.transform(pass_name="decompose-arbitrary-ppr")
@qml.transform(pass_name="to-ppr")
@qml.qnode(qml.device("lightning.qubit", wires=3))
def circuit():
    qml.PauliRot(pi / 8, "XYZ", wires=[0, 1, 2]) # passes on angles pi/2, pi/4, fails on others (including pi), pass on "X", "XY", fails on larger pauli words
    qml.pauli_measure("Z", wires=0) # <- fails with "X", "Y", "Z", passes with larger pauli words
```
crashes with output:
```shell
Traceback (most recent call last):
  File "catalyst/frontend/catalyst/compiler.py", line 296, in _catalyst
    result = subprocess.run(cmd, input=stdin, check=True, capture_output=True, text=True)
  File ".pyenv/versions/3.13.11/lib/python3.13/subprocess.py", line 577, in run
    raise CalledProcessError(retcode, process.args,
                             output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['catalyst/frontend/catalyst/utils/../../../mlir/build/bin/catalyst', '--tool', 'opt', '--catalyst-pipeline', 'pipe(quantum-compilation-stage),', '-']' died with <Signals.SIGABRT: 6>.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "Test/python.py", line 18, in <module>
    print(circuit.mlir_opt)
          ^^^^^^^^^^^^^^^^
  File "catalyst/frontend/catalyst/jit.py", line 590, in mlir_opt
    return to_mlir_opt(
        stdin=stdin, options=self.compile_options, using_python_compiler=using_python_compiler
    )
  File "catalyst/frontend/catalyst/compiler.py", line 402, in to_mlir_opt
    return _quantum_opt(*opts, *args, stdin=stdin)
  File "catalyst/frontend/catalyst/compiler.py", line 308, in _quantum_opt
[...]

PR #2359, opened 2026-01-07
Fix for Pauli string iteration out-of-bound

lazypanda10117:
**Context:** It was recently reported that  the `merge_ppr_ppm` pass does not work with `decompose_arbitrary_ppr` on certain edge cases. The cause of this is due to a mishandling of pauli word in `merge_ppr_ppm`, where we delete qubits from a reference while iterating its corresponding (unmodified) pauli word.

**Related GitHub Issues:**
Closes https://github.com/PennyLaneAI/catalyst/issues/2356

[[sc-107377]]


kipawaa (approved):

kipawaa:
Thanks for fixing this so quickly! Great job! 🚀

PR #2360, opened 2026-01-07
Include `to_ppr` pass in some passes' docstring

sengthai:
**Context:**
 Improve the code docstring in builtin_passes.py


sengthai:
> I know there's another pr where we change to use "pass_name"=...
> 
> Is that something we want to do here as well?

I think that is different. And, either using pass with "pass_name" or from `qml.transforms` are the same. Here, the code does not work if we don't perform `to_ppr`.

lazypanda10117 (approved):
Thanks @sengthai  💯

PR #2361, opened 2026-01-07
Fix extraneous register inserts with numpy wires

dime10:
Fixes an issue with extraneous insert/extracts when using numpy wires.

Reason is the detection for tracer wires being too loose in the QregPromise.

[sc-102815]


paul0403:
It also just occurred to me that this kind of "python typing error" will never happen again if we do the reference semantics to value semantics conversion (i.e. `QRegPromise` now) in mlir! Let's hope that day will come...

Issue #2353, opened 2026-01-07
qml.PauliRot Identity Lowering/Canonicalization


kipawaa:
Currently, a `qml.PauliRot` with identity gates is lowered to `qec.ppr` with identity gates, and in some cases can be made more complex via `decompose-arbitrary-ppr`. Rather than performing numerous operations on these gates, we should update canonicalization or paulirot lowering rules to remove identity operations.

Example:
```python
qml.capture.enable()

pipeline = [("pipe", ["quantum-compilation-stage"])]

@qml.qjit(pipelines=pipeline, target="mlir")
@qml.transform(pass_name="decompose-arbitrary-ppr")
@qml.transform(pass_name="to-ppr")
@qml.qnode(qml.device("lightning.qubit", wires=3))
def circuit():
    qml.PauliRot(0.2, "III", wires=[0, 1, 2])
```
generates
```mlir
...
    %4 = quantum.alloc_qb : !quantum.bit
    %5 = qec.prepare  plus %4 : !quantum.bit
    %mres, %out_qubits:4 = qec.ppm ["I", "I", "I", "Z"] %1, %2, %3, %5 : !quantum.bit, !quantum.bit, !quantum.bit, !quantum.bit
    %6 = qec.ppr ["X"](2) %out_qubits#3 cond(%mres) : !quantum.bit
    %7 = qec.ppr.arbitrary ["Z"](%cst) %6 : !quantum.bit
    %mres_0, %out_qubits_1 = qec.ppm ["X"] %7 : !quantum.bit
    %8:3 = qec.ppr ["I", "I", "I"](2) %out_qubits#0, %out_qubits#1, %out_qubits#2 cond(%mres_0) : !quantum.bit, !quantum.bit, !quantum.bit
    quantum.dealloc_qb %out_qubits_1 : !quantum.bit
...
```

PR #2367, opened 2026-01-08
Apply canonicalization to `to-ppr` pass to remove identity PPR operations

sengthai:
**Context:**
When converting operations like `qml.PauliRot(0.2, "III", wires=[0, 1, 2])` using the `to-ppr` pass, the resulting PPR operation with all-identity Pauli strings is effectively a no-op (just a global phase). Previously, users had to manually add a `canonicalize` pass to remove these operations.

This change integrates the canonicalization patterns for `PPRotationOp` and `PPRotationArbitraryOp` directly into the `to-ppr` pass.

**Related GitHub Issues:**
close https://github.com/PennyLaneAI/catalyst/issues/2353

### DIFF
--- a/mlir/lib/QEC/Transforms/to_ppr.cpp
+++ b/mlir/lib/QEC/Transforms/to_ppr.cpp
@@ -19,10 +19,11 @@
 #include "mlir/Pass/Pass.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 
-#include "QEC/IR/QECDialect.h"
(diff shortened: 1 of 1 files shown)


kipawaa (approved):
Looks great! Thanks for adding this in time, I think this will be much more intuitive for users, and generally a great improvement 🏆

dime10:
For the future I don't think this needs to go into the RC since it's an improvement, not a bug fix :)

PR #2369, opened 2026-01-08
Add hierarchical qubit types

joeycarter:
**Context:** Catalyst currently only supports a single, universal qubit type in its MLIR program representation: `!quantum.bit`. In order to express FTQC workloads in the compiler, we require a more expressive type system to represent a program that is successively lowered from the user program to a hardware-agnostic QEC layer and to a hardware-specific error-corrected MBQC physical layer. We also anticipate that the definitions of these new types will simplify the implementation of a number of other components in the FTQC software stack, especially those that operate at multiple levels of qubit-type abstraction, such as a Pauli frame tracking system.

**Description of the Change:** This change introduces the following qubit *levels*:

* `!quantum.bit<abstract>`
* `!quantum.bit<logical>`
* `!quantum.bit<qec>`
* `!quantum.bit<physical>`

The `abstract` level is the default and is equivalent to `!quantum.bit`. In addition to the qubit level, the following qubit *roles* have been defined to specialize the `qec` and `physical` qubit types: `null`, `data`, `xcheck` and `zcheck`. These roles are represented in the IR as, for example:

* `!quantum.bit<physical, null>`
* `!quantum.bit<physical, data>`
* `!quantum.bit<physical, xcheck>`
* `!quantum.bit<physical, zcheck>`
* etc.

The `null` role is the default for all levels and can be omitted. The purpose of the role parameter is to distinguish data qubits and syndrome qubits to extract X and Z corrections in CSS QEC codes by way of stabilizer measurements. As such, the IR restricts these role parameters to the `qec` and `physical` qubit types; `abstract` and `logical` qubits currently do not permit specifying a role other than `null`.

To allow for correct type parsing while continuing to use the custom assembly format that quantum operations use, all quantum ops that consume and return qubit values are now restricted to input and output qubits of the same type. This restrictions is accomplished by adding `AllTypesMatch<["in_qubit", "out_qubit"]>` traits to the appropriate quantum operations. This restriction means that operations such as

```mlir
%0 = ... : !quantum.bit<logical>
%1 = ... : !quantum.bit<physical>
%2 = quantum.custom "..."() %0 : !quantum.bit<logical>
%2 = quantum.custom "..."() %1 : !quantum.bit<physical>
```

are permitted, but operations that mix types, such as

```mlir
%0 = ... : !quantum.bit<logical>
%1 = ... : !quantum.bit<physical>
%2 = quantum.custom "..."() %0 : !quantum.bit<qec>
%3, %4 = quantum.custom "..."() %0, %1 :  !quantum.bit<logical>, !quantum.bit<physical>
```

are not.

**Benefits:** More expressive IR for FTQC workloads; simplifies implementation of compilation passes that convert between levels of abstractions.

**Possible Drawbacks:** The details of the compilation passes that convert a program to the QEC and physical layers are still being established, therefore the requirements for these new qubit types may change as these future passes develop.

[sc-107427], [sc-107428], [sc-107429], [sc-107430], [sc-107433]


mlir/include/Quantum/IR/QuantumInterfaces.td
```diff
@@ -79,11 +79,26 @@ def QuantumOperation : OpInterface<"QuantumOperation"> {
...
-        if (gate.getQubitOperands().size() != gate.getQubitResults().size())
+        if (operands.size() != results.size()) {
             return $_op->emitError() <<
-                "number of qubits in input (" << gate.getQubitOperands().size() << ") " <<
-                "and output (" << gate.getQubitResults().size() << ") must be the same";
+                "number of qubits in input (" << operands.size() << ") " <<
+                "and output (" << results.size() << ") must be the same";
+        }
+
+        if (!operands.empty()) {
```
dime10, line 91: Is this not duplicating the property verified by the `AllTypesMatch` declaration?

dime10 (approved):
This is a nice PR! 💯

mlir/include/Quantum/IR/QuantumInterfaces.td
joeycarter: I thought so too at first but `AllTypesMatch` doesn't appear to catch cases involving a single variadic input operand where the same input types are given but with different parameters. For example, removing this additional verifier block means that ops like

```mlir
%0 = ... : !quantum.bit<logical>
%1 = ... : !quantum.bit<physical>
%2, %3 = quantum.custom ""() %0, %1 : !quantum.bit<logical>, !quantum.bit<physical>
```

pass verification.

The `AllTypesMatch` declaration, on the other hand, means that we cannot instantiate ops that mix input and output qubit types, like

```mlir
%0 = ... : !quantum.bit<logical>
%1 = quantum.custom ""() %0 : !quantum.bit<physical>  // illegal
```

dime10: Good to know!

mlir/test/Quantum/QubitTypesTest.mlir
```diff
@@ -0,0 +1,354 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// RUN: quantum-opt %s --split-input-file --verify-diagnostics
+
+//===----------------------------------------------------------------------===//
+// Smoke tests: Can we use each qubit type?
+//===----------------------------------------------------------------------===//
```
lillian542, line 19: Just to improve my understanding of this part of the code base, what do these tests do exactly? Does something parse these `func`s and then we confirm it doesn't fall over? What would one of these tests failing look like?

lillian542 (approved):
Looks great @joeycarter, thanks!

mlir/test/Quantum/QubitTypesTest.mlir
joeycarter: That's it exactly. What happens here is only to test that our compiler tool can parse the ops and types. Suppose we wrote

```mlir
func.func @foo(%0 : !quantum.bit) {
    %1 = quantum.custom ""() %0 : !quantum.bit<magical>
    return
}
```

This would result in the error:

```
test/Quantum/QubitTypesTest.mlir:30:48: error: unexpected error: expected one of [abstract, logical, qec, physical] for Qubit levels in the hierarchical qubit representation, got: magical
    %1 = quantum.custom ""() %0 : !quantum.bit<magical>
                                               ^
test/Quantum/QubitTypesTest.mlir:30:55: error: unexpected error: failed to parse QubitType parameter 'level' which is to be a `QubitLevel`
    %1 = quantum.custom ""() %0 : !quantum.bit<magical>
                                                      ^
```

We're not checking the output here, only that we can run without error given this input MLIR. The command that is run on this input is at the top of the file:

```mlir
// RUN: quantum-opt %s --split-input-file --verify-diagnostics
```

where `%s` is the name of the current file.

PR #2374, opened 2026-01-09
Bump Pennylane and Lightning versions prior to the release (v0.14) 

paul0403:
**Context:**
Bump the PennyLane and Lightning minimum versions in preparation for the Catalyst 0.14.0 release. These changes ensure we can build the wheels after the Lightning release and before the core PennyLane release.

**Do not merge in this PR into the RC branch until the new version of Lightning has been released**


doc/requirements.txt
```diff
@@ -30,14 +30,7 @@ matplotlib==3.10.0
...
-# pennylane-lightning-kokkos==0.44.0-dev26
-# pennylane-lightning==0.44.0-dev26
-# pennylane==0.44.0-dev72
-
-# If targeting RC branch, we must install latest RC packages from TestPyPI
-# Revert when merging RC branch back to main, and uncomment the above regular requirements
---pre --extra-index-url https://test.pypi.org/simple/
-pennylane-lightning-kokkos>=0.44.0rc0,<=0.44.0
-pennylane-lightning>=0.44.0rc0,<=0.44.0
-pennylane>=0.44.0rc0,<=0.44.0
```
JerryChen97, line 43: Should we also keep these and just comment out so that next release reusable?

.dep-versions
```diff
@@ -10,8 +10,8 @@ enzyme=v0.0.203
 
 # For a custom PL version, update the package version here and at
 # 'doc/requirements.txt'
-pennylane=0.44.0-dev72
+# pennylane=0.44.0-dev72
 
 # For a custom LQ/LK version, update the package version here and at
 # 'doc/requirements.txt'
-lightning=0.44.0-dev26
+# lightning=0.44.0-dev26
```
jzaia18, line 17: Any reason we can't just set this to 0.44.0 instead of injecting custom install steps?

jzaia18: nvm, I just realized that this isn't where we install from in the first place

jzaia18 (approved):
Looks good. I think it would also be wise to leave some of this commented out rather than removing it (as Yushao already mentioned) but this is not a blocker

doc/requirements.txt
paul0403: No, I'll add the rc branch specific hacks to release guide

PR #2388, opened 2026-01-13
fix SSS: Install `graphviz` for `stable` Catalyst

andrijapau:
Fixes S/S/S as `graphviz` installation was added to both `latest` and `rc` checks but `stable` was missed.

Tested here: github.com/PennyLaneAI/catalyst/actions/runs/20969778272 💚


andrijapau:
Does this need a changelog @paul0403

PR #2392, opened 2026-01-13
Create workflow for verifying xDSL dialects against their MLIR versions

mudit2812:
Added a workflow that verifies parity between xDSL and MLIR dialects. The workflow does the following:
* Builds the dialects - this is needed to install Catalyst from source
* Creates xDSL dialects using the `create_xdsl_dialects.py` script (more details below)
* Saves the created dialect files as an artifact
* Compares the xDSL dialects in Catalyst against the newly created dialects using the `verify_xdsl_dialects.py` script (more details below)
* It can be triggered manually, and will also run on a weekly schedule.

Two scripts were added to support this workflow:
* `create_xdsl_dialects.py`: This script creates xDSL dialect files for the dialects that we are verifying:
  * It uses `llvm-tblgen` to generate a JSON file. `llvm-tblgen` is available to us because we download the cached LLVM build, which contains the executable
  * It strips some unnecessary fields from the JSON. In particular, it cleans the `Dialect` field, which may sometimes contain multiple dialects, and removes the assembly format, since we won't be validating that anyway.
  * It uses `xdsl-tblgen` to use the cleaned up JSON to create Python files for the dialects. These files are not usable out of the box, so some changes are needed.
  * Using regex, details about the constraints of all fields (other than `UnitAttr` fields, I'll explain later) are removed.
    * All non-`UnitAttr` fields are updated such that their constraint is `AnyAttr()`
    * `UnitAttr` fields are updated to use the standardized format of `opt_prop_def(UnitAttr)` or `opt_attr_def(UnitAttr)`. This is needed so that the verification is simpler.

* `verify_xdsl_dialects.py`: This script uses the dialects generated above to verify the xDSL dialects in Catalyst.
  * It checks that attributes and operations of the same name are in the dialects
  * It checks that all fields are correctly defined: optional fields are optional, and variadic fields are variadic
  * For fields that are supposed to be `UnitAttr`s, it checks that they are defined as *optional* properties or attributes

This workflow caught some issues with the Catalyst and QEC dialects, which are also fixed in this PR.

**Miscellaneous changes:**
* Update all workflows to use the `.yml` suffix instead of `.yaml`. This is already standardized across PL and Lightning.

I added the wheels build label so that all workflows run. I just want to make sure that the `.yaml`->`.yml` change didn't screw anything up. I'll remove the wheel builds once they pass at least once.

[sc-98908]


mudit2812:
@mehrdad2m I'm wondering if running the workflow on every unified compiler PR makes sense. I feel like it might be overkill, especially since we have very few PRs that update dialects. Maybe it makes more sense to run it on a schedule instead. What should the frequency of the schedule be in that case? I would say at least once a week.

.github/workflows/xdsl_scripts/create_xdsl_dialects.py
```diff
@@ -0,0 +1,221 @@
...
+
+    loaded = loads(json_str)
+
+    dialects = loaded["!instanceof"]["Dialect"]
+    new_dialects = [d for d in dialects if dialect_name in d]
+    loaded["!instanceof"]["Dialect"] = new_dialects
+
+    for val in loaded.values():
+        if isinstance(val, dict):
+            _ = val.pop("assemblyFormat", None)
```
mehrdad2m, line 120: Is this because we don't fully support assembly formats and it would result in a lot of mismatches or because the `llvm-tblgen` can not capture them correctly?

.github/workflows/xdsl_scripts/verify_xdsl_dialects.py
```diff
@@ -0,0 +1,321 @@
...
+    OperandDef,
+    OptionalDef,
+    PropertyDef,
+    RegionDef,
+    ResultDef,
+)
+
+from catalyst.python_interface import dialects as xdialects
+
+all_dialects = {
```
mehrdad2m, line 40: Is there a way to not hardcode this list for future ease?

.github/workflows/verify-xdsl-dialects.yml
```diff
@@ -0,0 +1,215 @@
...
+  cancel-in-progress: true
+
+jobs:
+  constants:
+    name: Set Constants
+    if: >-
+      ${{
+        ( github.event_name == 'pull_request'
+        && github.event.pull_request.draft == false
+        && contains(github.event.pull_request.labels.*.name, 'unified compiler') )
```
mehrdad2m: Doesn't this mean this workflow should be run in this PR? I don't see anything run in the CI 🤔

.github/workflows/verify-xdsl-dialects.yml
```diff
@@ -0,0 +1,215 @@
...
+    - name: Restore CCache on feature branches
+      id: restore-ccache
+      if: ${{ github.ref != 'refs/heads/main' }}
+      uses: actions/cache/restore@v4
+      with:
+        path: .ccache
+        key: ${{ runner.os }}-ccache-${{ github.run_id }}
+        restore-keys: ${{ runner.os }}-ccache-
+
+    - name: Build MLIR Dialects
```
mehrdad2m, line 108: Some steps here seems unnecessary. My understanding is that the scripts would only need the llvm-build directory (to access the `llvm-tblgen` too) and the Catalyst root directory (to access the dialect source files for comparison. I am a bit confused on why we need to build anything to do this comparison 🤔

.github/workflows/xdsl_scripts/create_xdsl_dialects.py
mudit2812: It's because we don't fully support assembly formats. This might not be as much of a problem anymore since I added the below irdl_options to all operations, which was the main issue with why assembly formats were not usable. 
```python
irdl_options = (
    AttrSizedOperandSegments(as_property=True),
    AttrSizedResultSegments(as_property=True),
)
```

That being said, assembly formats should now be tested properly with the round-trip testing updates in #2410 , so I don't really think we need to change anything here

.github/workflows/xdsl_scripts/verify_xdsl_dialects.py
mudit2812: I could update the script to import all the dialects that are in `catalyst.python_interface.dialects` (minus stablehlo). That should be a good way to future proof.

.github/workflows/verify-xdsl-dialects.yml
mudit2812: Oh, I removed the pull_request trigger based on [this discussion](https://github.com/PennyLaneAI/catalyst/pull/2392#issuecomment-3770077386). So this can be removed.

.github/workflows/verify-xdsl-dialects.yml
mudit2812: I need to be able to import catalyst so that I compare the autogenerated dialects against the ones in catalyst, but I'm not able to import catalyst without `make dialects`. I get an `ImportError` about `mlir_quantum`. Just to give a bit more detail about all the different stuff going on in the workflow, here is a summary of all the steps that involve restoring caches or building MLIR stuff:
* I need to restore the cache for the LLVM build, because the LLVM build contains the `llvm-tblgen` executable needed for the script. This is in the `compare-dialects` job.
* I need to restore the cached LLVM source code, because I need the include files there to feed to `llvm-tblgen`. This is in the `compare-dialects` job.
* I need to build the MLIR dialects using `make dialects`, because without that, catalyst cannot be imported. To build the dialects, I also need to restore the cached LLVM, StableHLO, and Enzyme builds. All of this is done in the `quantum` job.
  * However, it is not clear to me if I need to also restore the cached LLVM, StableHLO, and Enzyme source code in the `quantum` job. I will try removing those steps and see if anything fails.
* Finally, I need to be able to install Catalyst in the `compare-dialects` job so that I can import the xDSL dialects from it for comparison against the autogenerated dialects.

mudit2812: Looks like the cached source code is necessary for building the dialects. Log for reference [here](https://github.com/PennyLaneAI/catalyst/actions/runs/21263984258/job/61198414191#step:9:60).

.github/workflows/verify-xdsl-dialects.yml
mudit2812, line 1: @rashidnhm this is the only new YAML file

.github/workflows/verify-xdsl-dialects.yml
```diff
@@ -0,0 +1,208 @@
...
+    name: Quantum Dialects Build
+    needs: [constants]
+    runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        compiler: ${{ fromJson(needs.constants.outputs.compilers) }}
+
+    steps:
+    - name: Checkout Catalyst repo
+      uses: actions/checkout@v4
```
rashidnhm:
```suggestion
      uses: actions/checkout@v6
```

.github/workflows/verify-xdsl-dialects.yml
```diff
@@ -0,0 +1,208 @@
...
+
+    - name: Set up Python
+      uses: actions/setup-python@v5
```
rashidnhm:
```suggestion
      uses: actions/setup-python@v6
```

.github/workflows/verify-xdsl-dialects.yml
```diff
@@ -0,0 +1,208 @@
...
+      with:
+        python-version: ${{ needs.constants.outputs.primary_python_version }}
+
+    - name: Install Deps
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y python3 python3-pip cmake ninja-build ccache clang lld
+        python --version | grep ${{ needs.constants.outputs.primary_python_version }}
+        python -m pip install numpy nanobind pybind11
```
rashidnhm, line 46:
```suggestion
    - name: Install Deps
      env:
          EXPECTED_PYTHON_VERSION: ${{ needs.constants.outputs.primary_python_version }}
      run: |
        sudo apt-get update
        sudo apt-get install -y cmake ninja-build ccache clang lld
        python --version | grep $EXPECTED_PYTHON_VERSION
        python -m pip install numpy nanobind pybind11
```

.github/workflows/verify-xdsl-dialects.yml
```diff
@@ -0,0 +1,208 @@
...
+        C_COMPILER=$(which ${{ needs.constants.outputs[format('c_compiler.{0}', matrix.compiler)] }}) \
+        CXX_COMPILER=$(which ${{ needs.constants.outputs[format('cxx_compiler.{0}', matrix.compiler)] }}) \
+        LLVM_BUILD_DIR="$(pwd)/llvm-build" \
+        STABLEHLO_BUILD_DIR="$(pwd)/stablehlo-build" \
+        ENZYME_BUILD_DIR="$(pwd)/enzyme-build" \
+        DIALECTS_BUILD_DIR="$(pwd)/quantum-build" \
+        make dialects
+
+    - name: Upload Quantum Build Artifact
+      uses: actions/upload-artifact@v4
```
rashidnhm:
```suggestion
      uses: actions/upload-artifact@v6
```

`actions/download-artifact` would need to match this

.github/workflows/verify-xdsl-dialects.yml
```diff
@@ -0,0 +1,208 @@
...
+      uses: actions/cache/restore@v4
+      with:
+        path: enzyme-build
+        key: ${{ runner.os }}-enzyme-${{ needs.constants.outputs.llvm_version }}-${{ needs.constants.outputs.enzyme_version }}-ci-build-${{ matrix.compiler }}
+        fail-on-cache-miss: true
+
+    - name: Restore CCache on feature branches
+      id: restore-ccache
+      if: ${{ github.ref != 'refs/heads/main' }}
+      uses: actions/cache/restore@v4
```
rashidnhm, line 102: This has a `@v5` but I think we'd need to update the versions creating this to v5 so no unexpected behavior occurs.

.github/workflows/verify-xdsl-dialects.yml
```diff
@@ -0,0 +1,208 @@
...
+      with:
+        path: .ccache
+        key: ${{ runner.os }}-ccache-${{ github.run_id }}
+        restore-keys: ${{ runner.os }}-ccache-
```
rashidnhm, line 106: I don't think I get the logic here:

- We create the ccache only when running on main and use the run_id (unique) within the key
- And only restore on non main branch, where the primary key will always miss.

Might make sense to save it under `${{ runner.os }}-ccache`

.github/workflows/verify-xdsl-dialects.yml
mudit2812: Is this just so that the `grep` causes a non-zero exit code if the incorrect python version is installed?

rashidnhm: Generally better to not use env var to pass to bash. In this case it's pretty safe I guess but it can get flagged for code injection by a lot of scanners.

.github/workflows/verify-xdsl-dialects.yml
mudit2812: That will probably have to wait until the actions that create the cache are updated.

.github/workflows/verify-xdsl-dialects.yml
mudit2812: I've basically just copied what is in `check-catalyst.yml` for this job, so I'll let someone from the compilation team take over the discussion about how this should be handled. I don't think I still have a good understanding of the ccache-related steps

dime10: ccache is not needed unless this is a performance critical workflow (e.g. required workflows on PRs)

dime10: > We create the ccache only when running on main and use the run_id (unique) within the key
And only restore on non main branch, where the primary key will always miss.

Regarding your question @rashidnhm, this is a strategy to "update" a cache by always fetching the latest and always writing back at the end. Because GH doesn't support writing to an existing entry, I believe this is the closest thing there is at the moment.
The primary key will not miss btw, it's used as a prefix and will match more specific keys

Issue #2397, opened 2026-01-15
mcm_method='one-shot' is not supported with a dynamic number of shots


dime10:
Possibly related to a similar issue with dynamic shapes in #1949, the dynamic one-shot transform does not work when the number of shots is a tracer value. The underlying issue is the vmap transform not being able to handle a dynamic batch size. 
```py
import pennylane as qml

@qml.qjit
def main(shots: int):

    @qml.qnode(qml.device("lightning.qubit", wires=1), mcm_method="one-shot")
    def f():
        qml.H(0)
        return qml.sample()

    return qml.set_shots(shots)(f)()

main(10)
```
```
  File "catalyst/frontend/catalyst/api_extensions/function_maps.py", line 368, in _get_batch_size
    if not batch_size:

jax.errors.TracerBoolConversionError: Attempted boolean conversion of traced array with shape int64[].
```

paul0403:
PL fixed their issue, we can now use one shot with a dynamic number of shots for both counts and samples  https://github.com/PennyLaneAI/catalyst/pull/2786

PR #2398, opened 2026-01-15
No silent fallback to single-branch-statistics when one-shot is unsupported

dime10:
Dynamic-one shot was made the default execution method for the finite-shots setting when MCMs are present in the program (#2019). However, for several unsupported cases of the method a silent fallback to single-branch statistics was implemented.

To reduce potential confusion for users from differing behaviour under various circumstances for the same user-provided configuration, errors will instead be raised clearly indicating what the unsupported feature is and suggesting alternatives (including using single-branch statistics). To goal is to provide a consistent experience and to require the user to explicitly select single-branch statistics if they require it (hopefully making them aware of its implications).

Three automatic fallbacks remain in place:
- when the shots are finite but dynamic, as this is currently unsupported (see #2397)
  - hopefully the need for this fallback can be eliminated soon, although we could choose to raise an error for this even now if we want to
- when the device does not support one-shot
  - this is an edge-case for the CUDAQ integration and the OQD device, I thought it best to just leave the existing behaviour in place without additional hassle
- when measurements_from_samples/counts is explicitly applied by the user, which is not supported with one-shot
  - Technically I would like to remove this fallback as well, but the issue is that it would need to be checked for _after_ we have checked whether MCMs are even present in the program. Currently this check is done beforehand however and I don't want to raise an error about the transforms when we wouldn't use one-shot anyway because no MCMs were found on the tape.

---

Besides the change in fallback behaviour, the code structure was simplified to localize the logic better and some redundancy removed. The ideal flow is roughly: validate user configuration -> determine defaults -> check restrictions.

Also, the `None` sentinel value for analytic execution is preserved for the Python logic instead of moving to a more ambiguous `0` value. 

Additionally, a (slightly) better error message was added for issue #2397.

[sc-108205]

### DIFF
--- a/frontend/test/pytest/test_mid_circuit_measurement.py
+++ b/frontend/test/pytest/test_mid_circuit_measurement.py
@@ -30,7 +30,6 @@ from catalyst import CompileError, cond, grad
 from catalyst import jvp as C_jvp
 from catalyst import qjit, value_and_grad
 from catalyst import vjp as C_vjp
-from catalyst.third_party.oqd import OQDDevice
 
 # TODO: add tests with other measurement processes (e.g. qml.sample, qml.probs, ...)
 
@@ -1003,16 +1002,6 @@ class TestDynamicOneShotIntegration:
         res_cat, tree_cat = tree_flatten(r2)
         assert tree_jax == tree_cat
         assert np.allclose(res_jax, res_cat)
-
-    def test_unsupported_one_shot_device(self):
-        """Test unsupported device edge case."""
-
-        @qml.qnode(OQDDevice(backend="default", wires=1), shots=10, mcm_method="one-shot")
-        def circuit():
-            return qml.sample()
-
-        with pytest.raises(ValueError, match="'one-shot' is not supported in the chosen device"):
-            qjit(circuit)
 
 
 def sample_to_counts(results, meas_obj):
--- a/frontend/test/test_oqd/oqd/test_oqd_device.py
+++ b/frontend/test/test_oqd/oqd/test_oqd_device.py
@@ -17,6 +17,7 @@
 import pennylane as qml
 import pytest
 
+from catalyst import qjit
 from catalyst.third_party.oqd import OQDDevice
 
 
@@ -64,6 +65,16 @@ class TestOQDDevice:
         dev = OQDDevice(backend="default", wires=8)
         name, _ = dev.get_c_interface()
         assert name == "oqd"
+
+    def test_unsupported_one_shot_device(self):
+        """Test unsupported device edge case."""
+
+        @qml.qnode(OQDDevice(backend="default", wires=1), shots=10, mcm_method="one-shot")
+        def circuit():
+            return qml.sample()
+
+        with pytest.raises(ValueError, match="'one-shot' is not supported in the chosen device"):
+            qjit(circuit)
 
 
 if __name__ == "__main__":


frontend/catalyst/qfunc.py
```diff
@@ -65,49 +66,49 @@ class OutputContext:
...
-    if mcm_config.mcm_method is None:
-        updated_values["mcm_method"] = "one-shot"
-    if mcm_config.mcm_method == "deferred":
+
+def _resolve_mcm_config(qnode):
+    """Helper function for resolving and validating that the mcm_config is valid for executing."""
+    total_shots = qnode._shots.total_shots
+    analytic_shots = total_shots is None
+    finite_shots = not analytic_shots
+    static_shots = finite_shots and isinstance(total_shots, Integral)
```
paul0403, line 78: Like we found during the other typing issue (with the extra inserts), do you think it would be better to do the same thing here, i.e. check against tracer types directly?

dime10: Integral includes numpy actually, although we could also check against tracers that's true.

I went with this I think just because logically, I'm asking "is the object is a concrete Python int", not "is the thing is a tracer" (which was the question being asked in the case you mention).

frontend/catalyst/qfunc.py
```diff
@@ -474,52 +439,57 @@ def _validate_one_shot_measurements(
...
+    device_has_static_wires = wires is not None and not is_dynamic_wires(wires)
 
     # Raise an error if there are no mid-circuit measurements, it will fallback to
     # single-branch-statistics
     if (
         not any(isinstance(op, MidCircuitMeasure) for op in tape.operations)
         and user_specified_mcm_method is None
     ):
-        raise ValueError("No need to run one-shot mode when there are no mid-circuit measurements.")
+        raise FallbackToSingleBranch("No mid-circuit measurements present.")
```
rniczh: It appears that if no MCMs are present, the `configure_mcm_...` function returns `None`. , so the check `if one_shot_results is not None:` will evaluate to false and skip the subsequent logic. Given this flow, is it safe to assume we don't need to manually revert the `mcm_method` back to `single-branch-...`?

```python
def configure_mcm_and_try_one_shot(qnode, args, kwargs, pass_pipeline=None):
    ...
    try:
        return Function(
            dynamic_one_shot(qnode, mcm_config=mcm_config, pass_pipeline=pass_pipeline)
        )(*args, **kwargs)
    except FallbackToSingleBranch:
        return None

...

one_shot_results = configure_mcm_and_try_one_shot(new_qnode, args, kwargs, pass_pipeline)
# If we ran one-shot execution, we are done. Otherwise, continue execution as normal.
if one_shot_results is not None:
    return one_shot_results
```

dime10: Thanks for the question! You're talking about this line that was there before right? 
```py
mcm_config = replace(mcm_config, mcm_method="single-branch-statistics")
```

The thing is, the `mcm_config` object was not returned from the try one-shot method, so I don't think updating it was doing anything to begin with 🤔

frontend/catalyst/qfunc.py
```diff
@@ -65,49 +66,49 @@ class OutputContext:
     num_mcm: int
 
 
-def _resolve_mcm_config(mcm_config, shots):
-    """Helper function for resolving and validating that the mcm_config is valid for executing."""
-    updated_values = {}
+class FallbackToSingleBranch(Exception):
```
mlxd: Is this using exceptions for logical program flow?

frontend/catalyst/qfunc.py
rniczh: Thanks! You are right, that was used to skip the following condition previously, apprently we don't that anymore
`if mcm_config.mcm_method == "one-shot":`

frontend/catalyst/qfunc.py
dime10: Yes, but this is existing behaviour. This PR makes it more obvious/explicit. Unfortunately I don't know an immediate alternative to this.

mlxd: Gotcha :+1: 

I'm in favour of being explicit --- we can think about this in future designs. Any thoughts on naming the exception to have an `Ex`, `Error` or `Exception`  suffix ?

dime10: Don't mind an Error suffix if you want it, I omitted it because I think it reads nicer, and that semantically this is not intended to be an error

mlxd: Fine either way: just thought it may make it easier to understand the behaviour, as I rarely have seen exceptions not labelled as such in the wild.

paul0403 (approved):

frontend/catalyst/qfunc.py
dime10: https://github.com/PennyLaneAI/catalyst/pull/2398/commits/d93b3a361c825ac0f611ab0ef9e53abe2500d77a

PR #2399, opened 2026-01-15
Eliminate warnings from the test suite

dime10:
Typically, these can reveal incorrect test design (e.g. incorrect feature use, test logic issues), or even an issue with the feature (e.g. users seeing warnings that shouldn't be there). At best, they're just an oversight that can drown out other important warnings.

Warnings left for later:
- Executing a program capture QNode, raised over 200 times across 12 files, to be addressed in [sc 108254](https://app.shortcut.com/xanaduai/story/108254)
  ```
  pennylane/workflow/_capture_qnode.py:180: UserWarning: Executing PennyLane programs with capture enabled should be done inside ``qml.qjit``. Native execution of captured programs is an unmaintained experimental feature.
  ```
  - if we want we can also silence this globally in the meantime

Warnings removed:
- deprecated device shots
- deprecated `argnum` kwarg
- use PL device that supports arbitrary snapshots for comparison
- qml.specs warnings (both for MPs and for program args)
- graph decomp warnings
- mismatched signature, recompiling..
- qml.counts missing `all_outcomes` with capture
- not differentiating any parameters
- missing DebugInfo object with JAX

PL deprecations are now promoted to errors

[sc-108262]

### DIFF
--- a/frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_measurements_from_samples.py
+++ b/frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_measurements_from_samples.py
@@ -845,7 +845,6 @@ class TestMeasurementsFromSamplesIntegration:
             qml.transforms.decompose,
             gate_set={"X", "Y", "Z", "S", "H", "CNOT", "RZ", "RY", "GlobalPhase"},
         )
-        @qml.set_shots(1000)
         @qml.qnode(dev, shots=1000)
         def circuit():
             qml.CRX(0.1, wires=[0, 1])


frontend/test/pytest/from_plxpr/test_capture_integration.py
```diff
@@ -1097,7 +1097,7 @@ def test_transform_cancel_inverses_workflow(self, backend):
 
         qml.capture.enable()
 
-        @qjit(target="mlir")
+        @qjit
```
paul0403, line 1100: What warning was triggered here with `target="mlir"`?

frontend/test/pytest/test_callback.py
```diff
@@ -877,8 +877,7 @@ def wrapper(x):
         qml.RX(param, wires=0)
         return qml.expval(qml.PauliZ(0))
 
-    @jax.jit
-    @qml.grad
+    @partial(qml.grad, argnums=0)
```
paul0403, line 880: Does this change the intent of the test? i.e. it's no longer jitting

dime10: Great question! I can't know for sure, but I don't see any reason the comparison should be jitted. After providing an argnum, that function as written raises an error actually, so I'm thinking the test never really worked as intended to begin with? 🤔

frontend/test/pytest/from_plxpr/test_capture_integration.py
dime10: It's just unnecessary. I think author just blindly copied some things from other tests. The reason it's unnecessary is because we invoke the function right after anyway. 

A test should either be designed to check the MLIR without executing something (in which case `target="mlir"` can speed things up), or checking computed results (in which case the flag is pointless).

frontend/test/pytest/test_callback.py
paul0403: Ah I see, this was just the non qjit reference run. In that case I'm not super worried.

Also, without the argnums, what was it differentiating against? Was the `np.allclose` passing because it was comparing empty and empty? :sweat_smile:

dime10: Let me double check 😄

dime10: The output is indeed empty for the non-catalyst case!

paul0403: Oh god, does that mean qjit case is also empty?? :scream:

dime10: I don't understand how that passed the allclose check :O

WTF:

```
>>> np.allclose((), (0.08715757724135882,))
True
```

paul0403: Ok I tried, qjit case is good. The check was passing because
```python
(Pdb) import numpy
(Pdb) numpy.allclose((), 42)
True

```

paul0403: https://numpy.org/doc/2.3/reference/generated/numpy.allclose.html

> The comparison of a and b uses standard broadcasting, which means that a and b need not have the same shape in order for allclose(a, b) to evaluate to True.

dime10: Well, this is exactly why warnings are important 😅

dime10: Not sure if there is something we can do about the allclose issue more generally 🤔

paul0403: Now I'm wondering how many other tests we have are fake-passing because of this allclose business

frontend/test/pytest/test_braket_local_devices.py
```diff
@@ -731,7 +731,7 @@ def compiled(x: float):
         def interpreted(x):
             device = qml.device("default.qubit", wires=1)
             g = qml.QNode(f, device, diff_method="finite-diff")
-            h = qml.grad(g, argnum=0)
+            h = qml.grad(g, argnums=0)
```
andrijapau, line 700: Are `PennyLaneDeprecationWarnings` not raised as errors in Catalyst? 🤔

dime10: They should be raised, this fixes that warning :)

JerryChen97: I don't think so. At least not as errors when I committed in Catalyst last time.

andrijapau: >  They should be raised, this fixes that warning :)

Right but this was deprecated last release - it should have been fixed then? Typically we promote them to errors

dime10: Ah yes, there are no warnings as errors in the Catalyst test suite (sorry, missed the error part of your question)

andrijapau: Is this something we want to turn on - specifically for `PennyLaneDeprecationWarning`? It's a simple change to `pytest.ini`

dime10: We probably should at the very least turn PL deprecations into errors in our config, thanks for bringing it up :)

dime10: https://github.com/PennyLaneAI/catalyst/pull/2399/commits/2c8cf22aec61f6b856ef664051a958d7e0e9a448

frontend/test/pytest.ini
```diff
@@ -2,3 +2,5 @@
 markers =
     xdsl: marks tests for xdsl testing (select with '-m "xdsl"')
 xfail_strict=true
+filterwarnings =
+    error::pennylane.exceptions.PennyLaneDeprecationWarning
```
sengthai, line 6: 👍

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_measurements_from_samples.py
```diff
@@ -843,7 +843,7 @@ def test_integrate_with_decompose(self):
         @measurements_from_samples_pass
         @partial(
             qml.transforms.decompose,
-            gate_set={"X", "Y", "Z", "S", "H", "CNOT", "RZ", "GlobalPhase"},
+            gate_set={"X", "Y", "Z", "S", "H", "CNOT", "RZ", "RY", "GlobalPhase"},
         )
         @qml.set_shots(1000)
```
sengthai: It seems we set shots two times here. 

```suggestion
```

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_measurements_from_samples.py
```diff
@@ -843,7 +843,7 @@ def test_integrate_with_decompose(self):
         @measurements_from_samples_pass
         @partial(
             qml.transforms.decompose,
-            gate_set={"X", "Y", "Z", "S", "H", "CNOT", "RZ", "GlobalPhase"},
+            gate_set={"X", "Y", "Z", "S", "H", "CNOT", "RZ", "RY", "GlobalPhase"},
```
sengthai, line 846: As I know, RY is not supported in the MBQC gate set. How do we come up with this? 🤔

sengthai (approved):
Thanks @dime10 for removing warnings and cleaning it up. 💯 
I left a few comments below.

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_measurements_from_samples.py
dime10: @multiphaseCFD told me to add this gate so that we don't get a warning from the graph decomposition that RY doesn't define a decomposition. 
My understanding is that it's ok because the test is just about verifying that the transform works together with the decomp transform, and we don't really care what we're decomposing / decomposing to. If that's not the case, and the gate set was purposefully chosen, then the test author needs to verify whether the example chosen can actually be decomposed successfully to the chosen gate set. If the example is appropriate, then something must be wrong in the decomposition feature as the user shouldn't see warnings if nothing is wrong with their code.

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_measurements_from_samples.py
dime10: Good catch!

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_measurements_from_samples.py
sengthai: Ahh I see the points. And since this test aims to verify `measurements_from_samples_pass`, it is okay to set this RY then. 👍

dime10: I had a quick look and I am a bit confused why it can't find a decomposition without RY. Still investigating maybe there is a bug...

PR #2401, opened 2026-01-15
`draw_graph` tests are skipped if any dependencies are not installed

mudit2812:
The fixtures for skipping tests if any dependencies for `catalyst.draw_graph` are not installed is not complete. This PR fixes that.

[sc-108317]

### DIFF
--- a/frontend/test/pytest/python_interface/inspection/test_dag_builder.py
+++ b/frontend/test/pytest/python_interface/inspection/test_dag_builder.py
@@ -101,7 +101,7 @@ def test_abc_cannot_be_instantiated():
 def test_incomplete_subclass():
     """Tests that an incomplete subclass will fail"""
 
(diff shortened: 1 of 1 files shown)


frontend/test/pytest/python_interface/inspection/test_draw_unified_compiler.py
```diff
@@ -14,21 +14,32 @@
...
 from catalyst.python_interface.transforms import (
     iterative_cancel_inverses_pass,
     merge_rotations_pass,
 )
 
+pytestmark = pytest.mark.xdsl
+
+
+@pytest.fixture(scope="function")
+def skip_no_graph_deps():
```
mudit2812, line 34: I added a fixture to skip the tests instead of using `importorskip` so that the normal draw tests aren't skipped if deps for `draw_graph` aren't installed.

PR #2405, opened 2026-01-16
Add verifiers to quantum.paulirot op

paul0403:
**Context:**
As I was porting quantum dialect to reference semantics, I noticed that the pauli rot op is missing some verifiers.

Add 2 verifiers on `quantum.paulirot`:
1. That the length of the pauliword is the same as the number of qubits
2. That all the pauliwords are legal  

**Benefits:**
Correctness


lazypanda10117 (approved):
Thanks @paul0403! Looks good to me.

PR #2406, opened 2026-01-16
Update Unified Compiler Cookbook to work with PennyLane 0.44+ and Catalyst 0.14+

mudit2812:
This PR updates the cookbook after the 0.44/0.14 release and the increment to the xDSL version to 0.56.1 to use the most up-to-date APIs. Additionally, some language could be improved, which is also done in this PR.

[sc-108515]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -41,9 +41,13 @@
 
 <h3>Documentation 📝</h3>
 
+* Updated the Unified Compiler Cookbook to be compatible with the latest versions of PennyLane and Catalyst.
+  [(#2406)](https://github.com/PennyLaneAI/catalyst/pull/2406)
+
 <h3>Contributors ✍️</h3>
 
 This release contains contributions from (in alphabetical order):
 Ali Asadi
 Sengthai Heng,
-Jeffrey Kam.
+Jeffrey Kam,
+Mudit Pandey.


mehrdad2m (approved):

frontend/catalyst/python_interface/doc/unified_compiler_cookbook.rst
```diff
@@ -2,25 +2,19 @@ Unified Compiler Cookbook
...
-PyPI, so it needs to be installed from GitHub by running the following:
-
-.. code-block:: bash
-
-    pip install git+https://github.com/xdslproject/xdsl-jax.git
+    pennylane==0.45.0.dev5
+    pennylane-lightning==0.44.dev26
+    pennylane-catalyst==0.15.0.dev9
+    xdsl==0.56.1
+    xdsl-jax==0.2.0
```
andrijapau, line 17: Should we rewrite this to be a command you can just run to install these?

mudit2812: I don't think that makes sense because people might want to try it out with newer versions. I see the point of this to just be a reference for which versions are _guaranteed_ to work. For example, the cookbook still worked fine with these versions but without the other changes I made. I just got one deprecation warning.

andrijapau: Fair enough

PR #2408, opened 2026-01-19
Stop ignoring unified compiler coverage

mudit2812:
* Update `ignore` clause in `.codecov.yml` to only ignore `catalyst.python_interface.dialects.stablehlo`
* Fix bug with region ops in `ParitySynth`. See here for more info
* Add minimal testing to get 100% coverage for `python_interface`. Updates regarding this:
  * The coverage for all utilities inside the `inspection` module is much lower than other submodules. Those will be addressed in a separate PR
  * Updates to the transform module may be deferred similarly

Note: I'm only adding minimal testing because the purpose of this PR is strictly to get coverage to 100%. That way, we will at least begin reporting code coverage for unified compiler PRs that have new additions. There is a separate ongoing effort to improve testing retroactively.

[sc-98265]


mudit2812:
@mehrdad2m I think the reason why patch coverage is passing is because we didn't have any reports for the patch coverage to compare against since the whole submodule was being omitted. If you look at [the coverage report](https://app.codecov.io/gh/PennyLaneAI/catalyst/pull/2408/indirect-changes), there are a lot of missing lines in the `python_interface` submodule. I'll add minimal testing for all those lines. I'll add more details to the PR description about why I'm only adding minimal testing.

Will re-request review once coverage is good

frontend/catalyst/python_interface/xdsl_extras/traits.py
```diff
@@ -164,15 +164,8 @@ def verify(self, op: Operation) -> None:
...
-                f"'{op.name}': if an operand is non-scalar, then there must be at "
-                "least one non-scalar result"
-            )
-
         # If an operand is non-scalar, then all results must be non-scalar
-        if len(result_mappable_types) != len(op.results):
+        if not result_mappable_types or len(result_mappable_types) != len(op.results):
             raise VerifyException(
                 f"'{op.name}': if an operand is non-scalar, then all results must be non-scalar"
             )
```
mudit2812, line 171: I feel like it was weird to have one error say that "at least one result must be non-scalar" and then also have an error to say "all results must be non-scalar", so I just merged them into one. @mehrdad2m let me know what you think

frontend/catalyst/python_interface/transforms/quantum/parity_synth.py
```diff
@@ -313,7 +308,8 @@ def match_and_rewrite(self, matchedOp: Operation, rewriter: pattern_rewriter.Pat
                             self.rewrite_phase_polynomial(rewriter)
                             # Rewrite regions of this operation; Creating a new PatternRewriter
                             # so its matched operation is `op`, not `matchedOp`
-                            self.match_and_rewrite(op, pattern_rewriter.PatternRewriter(op))
+                            # self.match_and_rewrite(op, pattern_rewriter.PatternRewriter(op))
+                            self.match_and_rewrite(op, rewriter)
```
mudit2812, line 310: Creating a new pattern rewriter rather than reusing the same one as before was causing the worklist of the original rewriter to not be updated. This was causing issues because any operations that were erased by the new rewriter would still be in the worklist of the original rewriter, which led to errors when there were nested regions in the module being transformed, such as with control flow. @dwierichs this was my suggestion in the original PR, so apologies for that.

frontend/catalyst/python_interface/xdsl_extras/traits.py
mehrdad2m: Yeah that's fine by me. I'll update the xdsl-jax PR as well.

frontend/catalyst/python_interface/transforms/mbqc/convert_to_mbqc_formalism.py
```diff
@@ -498,7 +498,7 @@ def _queue_measurements(
                 return self._rotxzx_measurements(graph_qubit_dict, params)
             case "CNOT":
                 return self._cnot_measurements(graph_qubit_dict)
-            case _:
+            case _:  # pragma: no cover
```
mehrdad2m, line 501: Just to confirm, are these pargmas for cases that we actually don't care about the coverage or is it temporarily there to be covered later?

mehrdad2m (approved):
Thanks @mudit2812!

frontend/catalyst/python_interface/transforms/mbqc/convert_to_mbqc_formalism.py
mudit2812: These are for unreachable branches. The only way to reach that branch is if someone directly calls the helper with incorrect values, but the pass will never call it with inputs that lead to that branch.

andrijapau (approved):

PR #2410, opened 2026-01-19
Add MLIR parsing and pretty-printing to round-trip tests

mudit2812:
**Context:**
This PR updates the round trip testing strategy to also send the printed program to MLIR before parsing it back to xDSL. This improves verification of xDSL-MLIR parity as well.

* Added 2 new arguments to `run_filecheck`:
  * `to_mlir`: Whether to parse to MLIR when `roundtrip=True`. `True` by default.
  * `pretty_print`: If `True`, when `roundtrip=True`, the program will be pretty printed during the round trip
* This change caught an issue in the quantum dialect: `GlobalPhaseOp`'s `assembly_format` was incorrect, which is updated in this PR.
* Lit tests for all dialects were updated based on issues found with the updated round-trip testing. These issues came up because parsing MLIR with `quantum-opt` automatically runs verification on the module, which was failing for some of the tests
* Parametrized all lit tests to do the round trip with both pretty printing and generic printing
  * This caused the lit test for the Catalyst dialect to fail because xDSL doesn't support assembly formats for attributes. This test was xfailed.
* StableHLO lit tests set `to_mlir` to False because it doesn't work perfectly. This is fine in my opinion because StableHLO is being migrated to `xdsl-jax` anyway.

**Benefits:**
Better validation of parity between xDSL and MLIR dialects

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-98396] [sc-99388]

### DIFF
--- a/frontend/test/pytest/python_interface/dialects/test_mbqc_dialect.py
+++ b/frontend/test/pytest/python_interface/dialects/test_mbqc_dialect.py
@@ -94,7 +94,7 @@ def test_assembly_format(run_filecheck):
     %graph_reg = mbqc.graph_state_prep (%adj_matrix : tensor<6xi1>) [init "Hadamard", entangle "CZ"] : !quantum.reg
     """
 
-    run_filecheck(program, roundtrip=True)
+    run_filecheck(program, roundtrip=True, verify=True)
 
 
 class TestMeasureInBasisOp:
--- a/frontend/test/pytest/python_interface/dialects/test_mbqc_dialect.py
+++ b/frontend/test/pytest/python_interface/dialects/test_mbqc_dialect.py
@@ -60,7 +60,10 @@ def test_all_attributes_names(attr):
     assert attr.name == expected_name
 
 
-def test_assembly_format(run_filecheck):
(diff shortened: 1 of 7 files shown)


frontend/test/pytest/python_interface/dialects/test_mbqc_dialect.py
```diff
@@ -60,7 +60,10 @@ def test_all_attributes_names(attr):
     assert attr.name == expected_name
 
 
-def test_assembly_format(run_filecheck):
+@pytest.mark.parametrize(
+    "pretty_print", [pytest.param(True, id="pretty_print"), pytest.param(False, id="generic_print")]
+)
```
mudit2812, line 65: Using `pytest.param` here just so that in the logs the test shows up as `test_assembly_format[pretty_print]` rather than `test_assembly_format[True]`

mehrdad2m (approved):
Looks great! Thanks @mudit2812.

andrijapau (approved):
Thanks for the quick chat explaining the dialect changes. LGTM! 🎉

PR #2414, opened 2026-01-19
Fix PPR angle conversion when lowering to LLVM runtime calls

sengthai:
This PR fixes the angle conversion factor when lowering `qec.ppr` and `qec.ppr.arbitrary` operations to the LLVM runtime function `__catalyst__qis__PauliRot`. The PPR rotation angle is now correctly multiplied  by 2 to match the PauliRot convention (`PauliRot(φ) == PPR(φ/2)`

[[sc-98156]]

### DIFF
--- a/mlir/lib/Quantum/Transforms/ConversionPatterns.cpp
+++ b/mlir/lib/Quantum/Transforms/ConversionPatterns.cpp
@@ -1225,11 +1225,15 @@ template <typename T> struct PPRotationBasedPattern : public OpConversionPattern
             double theta = 2 * (llvm::numbers::pi / static_cast<double>(rotationKind));
             thetaValue = rewriter.create<LLVM::ConstantOp>(loc, rewriter.getF64FloatAttr(theta));
         }
-        else {
+        else if constexpr (std::is_same_v<T, PPRotationArbitraryOp>) {
             // multiply by 2 to get the rotation angle
             thetaValue = rewriter.create<LLVM::FMulOp>(
                 loc, adaptor.getArbitraryAngle(),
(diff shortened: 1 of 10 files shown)


mlir/lib/Quantum/Transforms/ConversionPatterns.cpp
```diff
@@ -1216,17 +1216,20 @@ template <typename T> struct PPRotationBasedPattern : public OpConversionPattern
             rewriter, op, qirName, qirSignature);
 
         // Get the rotation angle based on the op type
+        // Since the qml.PauliRot(phi) == PPR(phi/2), this rotation_kind is multiplied by 2.
```
rniczh, line 1219: can you add more description here to explain what happend? for example, why multiply the value with 2

lazypanda10117: Its the way how qec.ppr is represented in the IR. This is documented in the corresponding QEC pass i think.

rniczh: Just read the lightning code, looks good to me now, Thanks @paul0403

paul0403: For reference, if you follow the program flow, eventually it reaches this `ApplyPauliRot` kernel inside lightning https://github.com/PennyLaneAI/pennylane-lightning/blob/4739a0f6f42a39ae8c32da08ce00be942a604144/pennylane_lightning/core/simulators/lightning_qubit/gates/cpu_kernels/GateImplementationsLM.hpp#L577

This function predates all the PPR stuff, and was used for `qml.PauliRot`, whose angle regarding the factor of 2 is like `qml.RX`. So the PPR angle is half of what the expects.

PR #2415, opened 2026-01-19
[LLVM UPDATE] Update LLVM related dependencies Q1 2026

mehrdad2m:
**Context:**
Updating LLVM related dependencies for Q1 2025. After the update we will point the following versions:

```
jax=0.7.1 (no change)
stablehlo=v1.13.7 -> llvm=8f264586d7521b0e305ca7bb78825aa3382ffef7 
enzyme=v0.0.238
```

- Replaced rewriter/builder.create<Op>() with Op::create(rewriter/builder, ...) for operations. (https://github.com/PennyLaneAI/catalyst/pull/2416)

- Updated the "how to write a pass" tutorial to use the updated creation method. (https://github.com/PennyLaneAI/catalyst/pull/2416)

-  Changed integer literals in `InsertValueOp::create` calls to `SmallVector<int64_t>` because MLIR introduced `DenseI64ArrayAttr` overloads, creating ambiguity between `ArrayRef<int64_t>` and `DenseI64ArrayAttr` when passing integer literals. (https://github.com/PennyLaneAI/catalyst/pull/2416)

- Patched `mlir/llvm-project/mlir/tools/mlir-tblgen/OpPythonBindingGen.cpp` to avoid the crash when importing the generated python files because JAX's LLVM version is older and doesn’t expose `__class_getitem__` which causes an error complaining about indexing. see https://github.com/PennyLaneAI/catalyst/pull/2444

- Remove the xfail for a vmap+grad test that is passing now with the update and move another two xfails inside the tests since they are now segfaulting at runtime. see https://github.com/PennyLaneAI/catalyst/pull/2445

- Changed `std::nullopt` to `{}` for `ValueRange` default arguments because MLIR removed the `ArrayRef(std::nullopt_t)` constructor, requiring empty initializer lists instead. (https://github.com/PennyLaneAI/catalyst/pull/2415/commits/6b4f79066dff56e3bcdec0e47a4b8ad38980009e)

- Added `exactNone` argument to `Arith_DivSIOp` pattern because the operation now requires an explicit `fastmath` attribute parameter in tablegen. (https://github.com/PennyLaneAI/catalyst/pull/2415/commits/279cefc64cc3d5123725f43265e27a75b3005432)

- Fixed `BufferizableOpInterfaceImpl.cpp`: Added explicit cast from `TensorType` to `TensorLikeType` and `BaseMemRefType` to `BufferLikeType` because MLIR's bufferization framework now uses the `TensorLikeType`  and `BufferLikeType` interface to support generic tensor/buffer types. (https://github.com/PennyLaneAI/catalyst/pull/2415/commits/35b5690bb6e3ffbf6623bab470f74616bf5783f1)

- Fixed `StringSwitch::Cases` usage in to use `std::initializer_list` syntax because LLVM deprecated the old `Cases` method that took multiple string arguments in favor of a single `std::initializer_list` parameter. (https://github.com/PennyLaneAI/catalyst/pull/2415/commits/48f97d84accdf9b36dd1e8429311a06a30d44058)

- Fixed `inferCanonicalRankReducedResultType` call from 5 arguments to 3 arguments (https://github.com/PennyLaneAI/catalyst/pull/2415/commits/87cf94bd3d6cc7c16a9b617f40b93162a36653f9) 

- Added `StablehloBroadcastLowering` to `CMakeLists.txt` because StableHLO split broadcast lowering functionality into a separate library (https://github.com/PennyLaneAI/catalyst/pull/2415/commits/fad13a88d0f03ec2e18d4fb02f2b21f634291fff)

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-107552] [sc-107551]

### DIFF
--- a/mlir/lib/QEC/Transforms/DecomposeCliffordPPR.cpp
+++ b/mlir/lib/QEC/Transforms/DecomposeCliffordPPR.cpp
@@ -82,7 +82,7 @@ PPRotationOp decompose_pi_over_four_flattening(bool avoidPauliYMeasure, PPRotati
         PPMeasurementOp::create(rewriter, loc, pauliX, ppmPZ.getOutQubits().back(), measResult);
 
     // FIXME: Check global phase on this decomposition
-  
+
     auto cond = arith::XOrIOp::create(rewriter, loc, ppmPZ.getMres(), ppmX.getMres());
 
     SmallVector<Value> outPZQubits = ppmPZ.getOutQubits();


mlir/include/Ion/IR/IonOps.td
```diff
@@ -213,7 +213,7 @@ def ParallelProtocolOp : Ion_Op<"parallelprotocol", [SingleBlockImplicitTerminat
 
     let builders = [
       OpBuilder<(ins
-        CArg<"mlir::ValueRange", "std::nullopt">:$in_qubits,
+        CArg<"mlir::ValueRange", "{}">:$in_qubits,
```
paul0403, line 216: They finally unified the way to init empty value ranges! :finally:

mlir/lib/PauliFrame/Transforms/CliffordTToPauliFramePatterns.cpp
```diff
@@ -69,11 +69,11 @@ enum class GateEnum { I, X, Y, Z, H, S, T, CNOT, Unknown };
 GateEnum hashGate(CustomOp op)
 {
     return llvm::StringSwitch<GateEnum>(op.getGateName())
-        .Cases("Identity", "I", GateEnum::I)
-        .Cases("PauliX", "X", GateEnum::X)
-        .Cases("PauliY", "Y", GateEnum::Y)
-        .Cases("PauliZ", "Z", GateEnum::Z)
-        .Cases("H", "Hadamard", GateEnum::H)
+        .Cases({"Identity", "I"}, GateEnum::I)
```
paul0403, line 72: I see that they moving towards initialization lists for all their containers, and I like it!

PR #2419, opened 2026-01-20
Emit global phase operations in the `to-ppr` pass

sengthai:
**Context:**

When converting quantum gates to Pauli Product Rotations (PPR), the transformation changes the global phase of the circuit. Here we emit global phases when lowering to PPR.

This PR adds global phase compensation to the `to-ppr` compiler pass by emitting `quantum.gphase` operations during gate conversion:

- **H gate**: -π/2
- **S gate**: -π/4
- **T gate**: -π/8
- **X, Y, Z gates**: -π/2
- **CNOT gate**: +π/4

[[sc-98156]]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -43,6 +43,11 @@
   into normal PPR and PPMs with SCF dialect to support runtime execution.
   [(#2390)](https://github.com/PennyLaneAI/catalyst/pull/2390)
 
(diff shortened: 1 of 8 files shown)


lazypanda10117 (commented):
Looks good to me. It would be good ensure the gphase operation exists in all the relevant  test in MLIR. We should also check that the gphase values emitted in the IR match our code.

paul0403 (approved):

lazypanda10117 (approved):
Thanks @sengthai!

PR #2423, opened 2026-01-22
Improve `--mlir-timing`

kipawaa:
**Context:**
The `--mlir-timing` flag arbitrarily limits output resolution, in such a way that running passes on small files often yields a number of passes running in `0.000`s. This requires running on larger files for any sort of performance considerations to be valuable, which wastes time and resources.

This PR simply subclasses the usual `OutputStrategy` with `HighResolutionOutputStrategy` to increase the resolution of the output. This allows the (typically) microsecond-nanosecond precision data that is collected by the tool (platform/hardware dependent) to be reported to the user.

**Benefits:**
Increased visibility of pass performance and relative performance.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A

[[sc-108647]]


mlir/lib/Driver/CompilerDriver.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright 2023 Xanadu Quantum Technologies Inc.
+// Copyright 2026 Xanadu Quantum Technologies Inc.
```
sengthai: I think we should keep the first year as well.
```suggestion
// Copyright 2023-2026 Xanadu Quantum Technologies Inc.
```

kipawaa: Good catch, thanks!

PR #2424, opened 2026-01-22
Add a new pass to lower QEC initialization ops

sengthai:
**Context:**
- Add a new MLIR compiler pass `lower-qec-init-ops` that lowers QEC initialization operations to Quantum dialect operations
- `qec.prepare` is lowered to `quantum.customOp` with the appropriate state vector
- `qec.fabricate` is lowered to `quantum.alloc_qb` + `quantum.customOp`

[[sc-109079]]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -72,7 +72,7 @@
   to Quantum dialect operations. This pass converts `qec.prepare` to `quantum.set_state` and
   `qec.fabricate` to `quantum.alloc_qb` + `quantum.set_state`, enabling runtime execution of
   QEC state preparation operations.
-  [(#XXXX)](https://github.com/PennyLaneAI/catalyst/pull/XXXX)
+  [(#2424)](https://github.com/PennyLaneAI/catalyst/pull/2424)
 
 <h3>Documentation 📝</h3>
 
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -68,6 +68,12 @@
 * The upstream MLIR `Test` dialect is now available via the `catalyst` command line tool.
   [(#2417)](https://github.com/PennyLaneAI/catalyst/pull/2417)
 
+* A new compiler pass `lower-qec-init-ops` has been added to lower QEC initialization operations
+  to Quantum dialect operations. This pass converts `qec.prepare` to `quantum.set_state` and
+  `qec.fabricate` to `quantum.alloc_qb` + `quantum.set_state`, enabling runtime execution of
+  QEC state preparation operations.
+  [(#XXXX)](https://github.com/PennyLaneAI/catalyst/pull/XXXX)
+
 <h3>Documentation 📝</h3>
 
 * Updated the Unified Compiler Cookbook to be compatible with the latest versions of PennyLane and Catalyst.
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -69,8 +69,8 @@
   [(#2417)](https://github.com/PennyLaneAI/catalyst/pull/2417)
 
 * A new compiler pass `lower-qec-init-ops` has been added to lower QEC initialization operations
-  to Quantum dialect operations. This pass converts `qec.prepare` to `quantum.set_state` and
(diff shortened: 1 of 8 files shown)


mlir/lib/QEC/Transforms/LowerQECInitOps.cpp
```diff
@@ -0,0 +1,124 @@
...
+        size_t numQubits = op.getOutQubits().size();
+        for (size_t i = 0; i < numQubits; ++i) {
+            Value qubit;
+            if constexpr (std::is_same_v<OpType, PrepareStateOp>) {
+                // PrepareStateOp: use existing input qubits
+                qubit = op.getInQubits()[i];
+            }
+            else {
+                // FabricateOp: allocate a new qubit
+                auto allocOp = rewriter.create<AllocQubitOp>(loc);
```
lazypanda10117, line 98: I think this code works, but there are some implicit assumptions being made here. 

For instance, here we assume the `fabricate` op only returns 1 qubit (i.e, the magic state), even though its definition allows for returning multiple qubits. Let's say in the future, we can do something like `qec.fabricate blah` where `blah` is a two qubit resource state. The logic here would be trying to set the state for each individual qubit, when we should set the state for both qubits simultaneously. Such usage of `setState` for multiple qubits simultaneously  is supported in the simulator, see https://github.com/PennyLaneAI/pennylane-lightning/blob/master/pennylane_lightning/core/simulators/lightning_qubit/StateVectorLQubit.hpp#L943
I don't think the current approach takes into account where we may want to have multi-qubit resource states. Fundamentally, I don't think this is something we should be applying "per-qubit".

The same for the `qec.prepare` op.

Also, I think we can simply the logic to something like the following and remove the outer for loop.
```c++
 if constexpr (std::is_same_v<OpType, PrepareStateOp>) {
      qubits = op.getInQubits();
}
else if constexpr (std::is_same_v<OpType, FabricateOp>) {
    qubits = rewriter.create<AllocOp>; # to allocate the number of qubits required by the corresponding fabricate op.
}
Value stateVector = createStateVectorTensor(loc, rewriter, initState);
auto setStateOp = rewriter.create<SetStateOp>(loc, qubit.getType(), stateVector,  qubits);
```

dime10: Regarding whether set state can work, see the resets here:
https://github.com/PennyLaneAI/pennylane-lightning/blob/13959918434a06dd4392132c22563765e93f040f/pennylane_lightning/core/simulators/lightning_qubit/StateVectorLQubit.hpp#L933
https://github.com/PennyLaneAI/pennylane-lightning/blob/13959918434a06dd4392132c22563765e93f040f/pennylane_lightning/core/simulators/lightning_qubit/StateVectorLQubit.hpp#L979

sengthai: ohh I see. Thanks for sharing this @dime10.

sengthai: Then the only way to solve it is to lower to alloca + gates, you suggested.

doc/releases/changelog-dev.md
```diff
@@ -68,6 +68,12 @@
 * The upstream MLIR `Test` dialect is now available via the `catalyst` command line tool.
   [(#2417)](https://github.com/PennyLaneAI/catalyst/pull/2417)
 
+* A new compiler pass `lower-qec-init-ops` has been added to lower QEC initialization operations
```
dime10, line 75: This entry is ok as a standalone one for the pass, but ultimately there should a single entry about making PBC circuits executable that groups all the PRs together (not sure if there have been other PRs about this yet).

dime10 (approved):
Looks good 👍

lazypanda10117 (approved):
Looks good. Thanks!

doc/releases/changelog-dev.md
sengthai: Good idea! We haven't had a main entry about PBC execution yet.

sengthai: Since many dosctrings need to be updated in different places. I will include this one in another PR.

PR #2433, opened 2026-01-26
Fixing GlobalPhase issue in Gridsynth when using PPR basis

lazypanda10117:
**Context:** The current implementation of the gridsynth algorithm allows for two modes of output: one in the usual HST basis, and another in the PPR basis. When trying to compare the execution results between these two modes, we found that they sometimes differ by a global phase. This PR aims to address that and adds relevant test cases for gridsynth in the PPR basis.

[[sc-98156]]

### DIFF
--- a/runtime/lib/RSDecompRuntime/RSDecomp.cpp
+++ b/runtime/lib/RSDecompRuntime/RSDecomp.cpp
@@ -210,7 +210,7 @@ std::pair<bool, double> try_append_pair_expansion(std::vector<PPRGateType> &out,
 }
 
 /**
- * @brief Convert single gate for HST_to_PPR
+ * @brief Convert single gate for HST_to_PPR and return the global phase update.
  */
 double append_single_gate_expansion(std::vector<PPRGateType> &out, GateType gate)
 {
--- a/runtime/lib/RSDecompRuntime/RSDecomp.cpp
+++ b/runtime/lib/RSDecompRuntime/RSDecomp.cpp
@@ -161,9 +161,9 @@ std::pair<std::vector<PPRGateType>, double> eval_ross_algorithm_ppr(double angle
 
     auto [gates, phase] = compute_clifford_T_decomposition(angle, epsilon);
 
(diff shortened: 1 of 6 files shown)


sengthai (commented):
The implementation is good! When CI is green, ready to approve! 💯

runtime/lib/RSDecompRuntime/RSDecomp.cpp
```diff
@@ -161,91 +161,93 @@ std::pair<std::vector<PPRGateType>, double> eval_ross_algorithm_ppr(double angle
 
     auto [gates, phase] = compute_clifford_T_decomposition(angle, epsilon);
 
-    std::vector<PPRGateType> ppr_gates = HST_to_PPR(gates);
+    auto [ppr_gates, ppr_phase_update] = HST_to_PPR(gates);
```
paul0403, line 164: Not blocking, just saying that I feel like we should only use `auto` when in standard situations (e.g. loops) or when the type is immediately obvious from RHS (e.g. casts, templated calls, ...)

But feel free to ignore

paul0403 (approved):

sengthai (approved):
Nice! 👍

runtime/lib/RSDecompRuntime/RSDecomp.cpp
lazypanda10117: I used `auto` since the type is going to look something like `std::pair<std::vector<PPRGateType>, double>`, which is a little hard to read. Also, `compute_clifford_T_decomposition` also uses `auto` so I thought it is probably fine to do `auto` here as well.

paul0403: Yeah it's fine, that's why I said non blocking haha

dime10: Generally I agree but you need auto for structured binding unfortunately (which can add clarity)

PR #2434, opened 2026-01-26
Split Quantum dialect into multiple files

mudit2812:
**Context:**
The Quantum dialect is a monolith that contains a LOT of operations. This PR splits the dialect up into multiple files such that the file structure assists developers in finding what they need when reading the Quantum dialect's source code.

* Change `quantum` to a folder
  * Add `attributes.py`
  * Add `operations` folder
    * Add `gate_ops.py`
    * Add `measurement_ops.py`
    * Add `observable_ops.py`
    * Add `qubit_ops.py`
    * Add `miscellaneous_ops.py`
* Add `GateOp`, `UnitaryGateOp`, `TerminalMeasurementOp`, and `ObservableOp` base classes which allows us to group classes based on their intended purpose.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-96822]

### DIFF
--- a/frontend/catalyst/python_interface/dialects/quantum/operations/gate_ops.py
+++ b/frontend/catalyst/python_interface/dialects/quantum/operations/gate_ops.py
@@ -63,16 +63,6 @@ class GateOp(IRDLOperation):
 class UnitaryGateOp(GateOp):
     """Base class for operations representing unitary gates."""
 
-    irdl_options = (AttrSizedOperandSegments(as_property=True),)
-
-    adjoint = opt_prop_def(UnitAttr)
-
-    in_ctrl_qubits = var_operand_def(QubitType)
-
-    in_ctrl_values = var_operand_def(i1)
-
-    out_ctrl_qubits = var_result_def(QubitType)
-
 
 ###########################################
 ############## Unitary Gates ##############
@@ -94,7 +84,10 @@ class CustomOp(UnitaryGateOp):
         `:` type($out_qubits) (`ctrls` type($out_ctrl_qubits)^ )?
     """
 
-    irdl_options = (AttrSizedResultSegments(as_property=True),)
+    irdl_options = (
+        AttrSizedOperandSegments(as_property=True),
+        AttrSizedResultSegments(as_property=True),
+    )
 
     params = var_operand_def(Float64Type())
 
@@ -102,7 +95,15 @@ class CustomOp(UnitaryGateOp):
 
     gate_name = prop_def(StringAttr)
 
+    adjoint = opt_prop_def(UnitAttr)
+
+    in_ctrl_qubits = var_operand_def(QubitType)
+
+    in_ctrl_values = var_operand_def(i1)
+
     out_qubits = var_result_def(QubitType)
+
+    out_ctrl_qubits = var_result_def(QubitType)
 
     traits = traits_def(NoMemoryEffect())
 
@@ -168,9 +169,17 @@ class GlobalPhaseOp(UnitaryGateOp):
         `:` (`ctrls` type($out_ctrl_qubits)^ )?
     """
 
-    irdl_options = (ParsePropInAttrDict(),)
+    irdl_options = (AttrSizedOperandSegments(as_property=True), ParsePropInAttrDict())
 
     params = operand_def(Float64Type())
+
+    adjoint = opt_prop_def(UnitAttr)
+
+    in_ctrl_qubits = var_operand_def(QubitType)
+
+    in_ctrl_values = var_operand_def(i1)
+
+    out_ctrl_qubits = var_result_def(QubitType)
 
     def __init__(
         self,
@@ -220,13 +229,24 @@ class MultiRZOp(UnitaryGateOp):
         `:` type($out_qubits) (`ctrls` type($out_ctrl_qubits)^ )?
     """
 
-    irdl_options = (AttrSizedResultSegments(as_property=True),)
+    irdl_options = (
+        AttrSizedOperandSegments(as_property=True),
+        AttrSizedResultSegments(as_property=True),
+    )
 
     theta = operand_def(Float64Type())
 
     in_qubits = var_operand_def(QubitType)
 
+    adjoint = opt_prop_def(UnitAttr)
+
+    in_ctrl_qubits = var_operand_def(QubitType)
+
+    in_ctrl_values = var_operand_def(i1)
+
     out_qubits = var_result_def(QubitType)
+
+    out_ctrl_qubits = var_result_def(QubitType)
 
     traits = traits_def(NoMemoryEffect())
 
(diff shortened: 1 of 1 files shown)


frontend/catalyst/python_interface/dialects/quantum/operations/measurement_ops.py
```diff
@@ -0,0 +1,221 @@
...
+
+from ..attributes import ObservableSSAValue, ObservableType, QubitSSAValue, QubitType
+
+##############################################
+################ Base classes ################
+##############################################
+
+
+class TerminalMeasurementOp(IRDLOperation):
+    """Base class for operations representing terminal measurement processes."""
```
mudit2812, line 44: All operations in this file with the exception of `MeasureOp` inherit this base class

frontend/catalyst/python_interface/dialects/quantum/operations/observable_ops.py
```diff
@@ -0,0 +1,132 @@
...
+
+##############################################
+################ Base classes ################
+##############################################
+
+
+class ObservableOp(IRDLOperation):
+    """Base class for operations representing observables."""
+
+    obs = result_def(ObservableType)
```
mudit2812, line 43: This base class is used for all the operations in this file. I removed the `obs` result declaration from all the operations and put it here because all operations define it the same way.

frontend/catalyst/python_interface/dialects/quantum/operations/gate_ops.py
```diff
@@ -0,0 +1,532 @@
...
+
+    irdl_options = (AttrSizedOperandSegments(as_property=True),)
+
+    adjoint = opt_prop_def(UnitAttr)
+
+    in_ctrl_qubits = var_operand_def(QubitType)
+
+    in_ctrl_values = var_operand_def(i1)
+
+    out_ctrl_qubits = var_result_def(QubitType)
```
mudit2812: These are new, and all operations in this file inherit one of these base classes. This will help us in the future to do stuff like `isinstance(custom_op, UnitaryGateOp)`. `GateOp` is for gate-like operations, and `UnitaryGateOp` is specifically for unitary gates. The former is only used for state prep ops. Since all unitary gates share `adjoint`, `in_ctrl_qubits`, `in_ctrl_values`, and `out_ctrl_qubits`, I put them in this base class rather than duplicating them for each operation. I can't do the same for `in_qubits` and `out_qubits` because `GlobalPhaseOp` has neither of those fields.

frontend/catalyst/python_interface/dialects/quantum/operations/measurement_ops.py
mudit2812: Also, note that all the terminal measurements in this file have a field defined as
```python
obs = operand_def(ObservableType)
```
I have not moved this to the `TerminalMeasurementOp` base class because it causes failures. The `@irdl_op_definition` decorator iterates over the `__mro__` of an `IRDLOperation`, collecting fields (operand, result, property, declarations, irdl_options, etc.) from the main class, and then its base classes. So, putting operands and results in the base class means that they get collected _after_ the fields in the main class.

But, for terminal measurements, `obs` is the _first_ operand being defined. Thus, putting `obs` in the base class causes it to get collected _after_ all other operands when it should've been collected first. This is the same reason we're able to put `in_ctrl_qubits` and `in_ctrl_values` in `UnitaryGateOp` - they're the last operands to get collected either way.

frontend/catalyst/python_interface/dialects/quantum/operations/gate_ops.py
andrijapau: Beauty

mehrdad2m: While I really love to have the base classes here, I am a bit hesitant of defining the gates/irdl_options here as derived gates are appending them with more and that imo makes the code less readable cause it require you to look at two places to get a hold of what a gate's parameters are. Maybe it's because I am used to the MLIR tablegen version of the dialects as this typically doesn't happen in MLIR. My personal preference is to use base classes either as pure types with no implementation detail or as a complete implementation that doesn't get much modifications in derived operations other than customization of names, verifiers, etc, but I would be curious to know xdsl user's preferences cause ATEOTD, they will be using the dialect.
Also this again showcases how annoying the current design of `GlobalPhaseOp` is 😆

frontend/catalyst/python_interface/dialects/quantum/operations/measurement_ops.py
```diff
@@ -0,0 +1,221 @@
...
+from catalyst.python_interface.xdsl_extras import MemRefConstraint, TensorConstraint
+
+from ..attributes import ObservableSSAValue, ObservableType, QubitSSAValue, QubitType
+
+##############################################
+################ Base classes ################
+##############################################
+
+
+class TerminalMeasurementOp(IRDLOperation):
```
mehrdad2m, line 43: Nice, I think it's worth unifying these classes with the MLIR impl as well cause in future, these could be used in passes that would be ported back to C++. not blocking for now, but maybe an issue page to track them would be nice.

frontend/catalyst/python_interface/dialects/quantum/operations/gate_ops.py
mudit2812: I see your point, and I think that there will definitely be less contention about having the fields be defined in the operations themselves rather than the base class. I'll remove the change and re-request review.

frontend/catalyst/python_interface/dialects/quantum/operations/measurement_ops.py
mudit2812: The MLIR impls are what motivated me to make these base classes in the first place. However, I don't think these necessary have to be unified with the MLIR impls, because it might not have any relation to user experience. The reason why I named this class `TerminalMeasurementOp` instead of the MLIR `MeasurementOp` is precisely because of this, because `MeasurementOp` sounds way too similar to `MeasureOp`

mehrdad2m: Oh, I missed the fact that we have a `Measurement_Op` in MLIR. I just searched for `TerminalMeasurementOp` and didn't find any. Yeah you are right it is indeed very confusing. Especially that `measureOp` is not a `MeasurementOp` 🥲

frontend/catalyst/python_interface/dialects/quantum/operations/gate_ops.py
mudit2812: [91bfa7b](https://github.com/PennyLaneAI/catalyst/pull/2434/commits/91bfa7baad972a665285f9070ba607436d9d0496)

mehrdad2m (approved):
Thanks!

PR #2437, opened 2026-01-27
Split non commuting pass is added to Catalyst in MLIR

rniczh:
**Context:**

Adds the split-non-commuting pass, which splits quantum functions that measure non-commuting observables into multiple execution groups (In this inital implemenation, we treat each expval as different group). Each group runs in a separate call from the main function.

**Steps:**
1. We first run the `split-to-single-terms` pass to decompose Hamiltonian expval values into single-term observables before this pass runs.
2. Then we do the transformation to construct groups for each expval.

**Before:**
```mlir
func.func @circuit() -> (f64, f64, f64) attributes {qnode} {
  %c100 = arith.constant 100 : i64
  quantum.device shots(%c100) ["", "", ""]
  %0 = quantum.alloc(3) : !quantum.reg
  ...
  %ev0 = quantum.expval %obs_X {group = 0} : f64
  %ev1 = quantum.expval %obs_Y {group = 1} : f64
  %ev2 = quantum.expval %obs_Z {group = 2} : f64
  return %ev0, %ev1, %ev2 : f64, f64, f64
}
```

**After:**
```mlir
func.func @circuit() -> (f64, f64, f64) {
  %r0 = call @circuit.group.0() : () -> f64
  %r1 = call @circuit.group.1() : () -> f64
  %r2 = call @circuit.group.2() : () -> f64
  return %r0, %r1, %r2 : f64, f64, f64
}

func.func private @circuit.group.0() -> f64 attributes {qnode} {
  %c100 = arith.constant 100 : i64
  %c3 = arith.constant 3 : i64
  %divided = arith.divsi %c100, %c3 : i64
  quantum.device shots(%divided) ["", "", ""]
  ...
}

// ... group 1 ...
// ... group 2 ...
```

Shots are distributed among groups using integer division (rounded down); e.g., 100 shots with 3 groups yields 33 shots per group

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-93667]


mlir/lib/Quantum/Transforms/split_non_commuting.cpp
```diff
@@ -0,0 +1,373 @@
...
+    /// Remove measurement operations (and their observable / intermediate chains) that do not
+    /// belong to the target group.
+    static void removeGroup(func::FuncOp groupFunc, int targetGroup)
+    {
+        Operation *returnOp = groupFunc.front().getTerminator();
+
+        // Identify return values that belong to other groups.
+        llvm::DenseSet<Value> returnValuesToRemove;
+        for (Value operand : returnOp->getOperands()) {
+            int groupId = findGroupForReturnValue(operand);
```
rniczh: this function has been executed before, don't need to perform it twice

frontend/catalyst/python_interface/transforms/quantum/__init__.py
```diff
@@ -24,7 +24,6 @@
 )
 from .merge_rotations import MergeRotationsPass, merge_rotations_pass
 from .parity_synth import ParitySynthPass, parity_synth_pass
-from .split_non_commuting import SplitNonCommutingPass, split_non_commuting_pass
```
rniczh, line 27: conflict with xdsl's split non commuting

mlir/lib/Catalyst/Transforms/ApplyTransformSequencePass.cpp
```diff
@@ -124,16 +124,12 @@ LogicalResult applyTransformsWithSubpassTracking(Operation *payload,
...
 
-            if (result.isDefiniteFailure()) {
-                // hook after pass failed
+            // Reports the error whether definite or silenceable
+            if (failed(result.checkAndReport())) {
                 passInstrumentor->runAfterPassFailed(subPass.get(), payload);
                 return failure();
             }
 
-            if (result.isSilenceableFailure()) {
```
rniczh, line 133: this cause the `failure` from registered pass be ignored

frontend/catalyst/python_interface/transforms/quantum/__init__.py
rniczh: Confirmed in offline, remove the xdsl's split non commuting.

mlir/lib/Catalyst/Transforms/ApplyTransformSequencePass.cpp
rniczh: ci passed

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
rniczh: Done https://github.com/PennyLaneAI/catalyst/pull/2437/commits/107f7eb0a0145923b6e6802ede4ae41b8a757a60

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
```diff
@@ -0,0 +1,441 @@
...
+
+        return groupId;
+    }
+
+    /// Trace back from a return value through the SSA graph to find which
+    /// group's measurement produced it. Returns the group id, or -1 if not found.
+    static int findGroupForReturnValue(Value returnValue)
+    {
+        SmallVector<Operation *, 4> worklist;
+        llvm::SmallPtrSet<Operation *, 4> visited;
```
sengthai, line 96: 4 operations mean {I, X, Y, Z} Pauli?

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
```diff
@@ -0,0 +1,441 @@
...
+        llvm::DenseSet<Value> returnValuesToRemove;
+        for (auto [position, operand] : llvm::enumerate(returnOp->getOperands())) {
+            int groupId = returnValueGroupIds[position];
+            bool keep = (groupId == targetGroup);
+            if (!keep) {
+                returnValuesToRemove.insert(operand);
+            }
+        }
+
+        std::deque<Operation *> removeOps;
```
sengthai, line 217: 👍

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
```diff
@@ -0,0 +1,441 @@
...
+                                     ArrayRef<int> returnValueGroupIds, ModuleOp moduleOp)
+    {
+        // clone the entire function
+        func::FuncOp groupFunc = funcOp.clone();
+        std::string groupName = funcOp.getSymName().str() + ".group." + std::to_string(groupIdx);
+        groupFunc.setSymName(groupName);
+        groupFunc.setPrivate();
+
+        SymbolTable modSymTable(moduleOp);
+        modSymTable.insert(groupFunc);
```
sengthai: How did `goupName` resolve the possibly duplicated name? 🤔

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
```diff
@@ -0,0 +1,441 @@
...
+            Operation *op = worklist.pop_back_val();
+            if (!op || !visited.insert(op).second) {
+                continue;
+            }
+
+            if (isSupportedMeasOp(op)) {
+                assert(op->hasAttrOfType<IntegerAttr>("group") &&
+                       "measurement op must have group attribute");
+                return static_cast<int>(op->getAttrOfType<IntegerAttr>("group").getInt());
+                continue;
```
sengthai: `continue` is unreachable dead code
```suggestion
```

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
```diff
@@ -0,0 +1,441 @@
...
+
+    /// Analyze which return-value positions belong to each group.
+    /// Returns a pair of (groupPositions, returnValueGroupIds):
+    /// - groupPositions: group_id -> list of positions in the return operand list
+    /// - returnValueGroupIds: list of group_ids for each return value
+    ///
+    /// Return values that don't trace back to any measurement (e.g. constants produced by
+    /// `split-to-single-terms` for Identity observables) are assigned to group 0 so that group 0's
+    /// function computes and returns them.
+    static std::optional<std::pair<llvm::DenseMap<int, SmallVector<int>>, SmallVector<int>>>
```
sengthai: it seems to me that the implementation never returns nullopt, so then I think we can just use `std::pair<...>`

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
```diff
@@ -0,0 +1,441 @@
...
+            int numGroups = calculateNumGroups(funcOp);
+
+            // If there is only one group, no need to split the execution.
+            if (numGroups <= 1) {
+                continue;
+            }
+
+            // Analyze return value positions for each group
+            auto analysisResult = analyzeGroupReturnPositions(funcOp, numGroups);
+            if (!analysisResult) {
```
sengthai: because of this: https://github.com/PennyLaneAI/catalyst/pull/2437/changes#r2795507744
`(!analysisResult)` can never be true

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
```diff
@@ -0,0 +1,441 @@
...
+        funcOp.walk([&](ExpvalOp expvalOp) {
+            auto namedObsOp = expvalOp.getObs().getDefiningOp<NamedObsOp>();
+            if (namedObsOp && namedObsOp.getType() == NamedObservable::Identity) {
+                // Identity expval is always 1.0
+                OpBuilder builder(expvalOp);
+                Value one = arith::ConstantOp::create(builder, expvalOp.getLoc(),
+                                                      builder.getF64FloatAttr(1.0));
+                expvalOp.replaceAllUsesWith(one);
+                expvalOp.erase();
+                namedObsOp.erase();
```
sengthai: I think erasing operations during a `walk` is generally unsafe. 🤔

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
```diff
@@ -0,0 +1,441 @@
...
+        if (!shots) {
+            return;
+        }
+
+        // Compute the divided shots and pass it to the device op
+        OpBuilder builder(deviceOp);
+        Location loc = deviceOp.getLoc();
+        Value numGroupsVal = arith::ConstantOp::create(
+            builder, loc, builder.getI64IntegerAttr(static_cast<int64_t>(numGroups)));
+        Value dividedShots = arith::DivSIOp::create(builder, loc, shots, numGroupsVal);
```
sengthai: We may lost shot during division. example 100 shots / 3 groups = 33 per group = 99 total (1 shot lost). I think it's acceptable, but it should at least be documented. Or, we want to distribute that 1 shot to the entire group.

mlir/include/Quantum/Transforms/Passes.td
```diff
@@ -66,6 +66,10 @@ def SplitMultipleTapesPass : Pass<"split-multiple-tapes"> {
     let summary = "Given a qnode containing multiple tapes, split each tape into its own function.";
 }
 
+def SplitNonCommutingPass : Pass<"split-non-commuting"> {
+    let summary = "Split quantum functions non-commuting observables into multiple executions.";
```
sengthai, line 70: don't we need to add `let dependentDialects = ["arith::ArithDialect"]` in here, since it may generate those dialects?

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
sengthai: A new `SymbolTable` is built for each call to this function. Would it be more efficient to build it once and pass it in?

mlir/lib/Catalyst/Transforms/ApplyTransformSequencePass.cpp
```diff
@@ -124,16 +124,12 @@ LogicalResult applyTransformsWithSubpassTracking(Operation *payload,
 
             DiagnosedSilenceableFailure result = state.applyTransform(transformInterface);
 
-            if (result.isDefiniteFailure()) {
-                // hook after pass failed
+            // Reports the error whether definite or silenceable
+            if (failed(result.checkAndReport())) {
                 passInstrumentor->runAfterPassFailed(subPass.get(), payload);
                 return failure();
             }
```
sengthai, line 131: Is that a bug fix? It deserves its own PR for clarity, though I understand it was needed to make this pass error-free.

sengthai (commented):
Well done, well written! I have a few comments below.

frontend/test/lit/test_split_non_commuting.py
dime10, line 1: Per the reasoning described [here](https://github.com/PennyLaneAI/catalyst/pull/2441#discussion_r2771230597), please do not test functionality from MLIR passes with frontend lit tests. Such tests should be covered by MLIR lit tests. Pytest integration tests can provide additional security the pass works in an end-to-end scenario, but unless there is something specific happening in the frontend or lowering mechanism that needs testing, there shouldn't be a reason to use frontend lit tests.

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
rniczh: No, it just used for reserve the space for worklist at the begin. So ideally, tracking from the return value to the measurement op may only contains two ops, for example:
```
meas = measurment ...
tensor_result = from_element meas
return tesnor_result
```

But I don't want to be too aggressive, so I kept 4 ops space here. But well, it just a number for inital space for walking the ops.

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
rniczh: Thanks, I updated in this commit: https://github.com/PennyLaneAI/catalyst/pull/2437/commits/29fe86e96195ee777a3009796ae16bfacad5bbb1

just sepearte the conversion here to another loop. So collect all the expval-identity pairs and conver them to 1.0 in another loop.

mlir/include/Quantum/Transforms/Passes.td
rniczh: Done. https://github.com/PennyLaneAI/catalyst/pull/2437/commits/29fe86e96195ee777a3009796ae16bfacad5bbb1

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
rniczh: I moved it out of this function, so just need to intialize once 
https://github.com/PennyLaneAI/catalyst/pull/2437/commits/29fe86e96195ee777a3009796ae16bfacad5bbb1

mlir/lib/Catalyst/Transforms/ApplyTransformSequencePass.cpp
rniczh: I'm not sure actaully, it seems to me that it's a bug, for the case that return failure from the pass of transform regsitered seqequence will be ignored. But one of the description of this story mentioned that it needs to raise an error if there are unsupported operations.

frontend/test/lit/test_split_non_commuting.py
rniczh: Thanks! Then I think the lit is not needed here. Just removed it in this commit:
https://github.com/PennyLaneAI/catalyst/pull/2437/commits/29fe86e96195ee777a3009796ae16bfacad5bbb1

And I also removed the frontend lit test of `split-to-single-terms` as well.
https://github.com/PennyLaneAI/catalyst/pull/2437/commits/15f18898e349d6801e4c5ff1f599466be6304159

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
```diff
@@ -0,0 +1,442 @@
...
+                arith::ConstantOp::create(builder, expvalOp.getLoc(), builder.getF64FloatAttr(1.0));
+            expvalOp.replaceAllUsesWith(one);
+            expvalOp.erase();
+            namedObsOp.erase();
+        }
+    }
+
+    void runOnOperation() override
+    {
+        ModuleOp moduleOp = cast<ModuleOp>(getOperation());
```
dime10: This cast is problematic because it leads to crashes (failed assertions if enabled, I imagine worse if not) as soon as the pass is run on something that's not a ModuleOp. `cast` should never be used unless you have already checked that it is indeed the type you expect.

There is a few options to improve this:
- Make your pass into a filtered operation pass. This way the pass manager guarantees that your pass is run only on the op kind you expect (for example the ModuleOp).
- Gracefully handle different op kinds. For example:
  - a FuncOp: check if it is a qnode, if yes do thing, else exit
  - a ModuleOp: look for FuncOps inside, then process as such (you already follow this approach further down)
  - an unknown op type: exit

The question is which approach would be the best for our "QNode" passes? 🤔 I'm trying to come up with guidelines for this actually (to help the FTQC team), so this pass is a good example. 
In this case, since we are inserting new functions, so I think only a module-level pass makes sense really because modifying operations outside the one a pass is called on is illegal. So either the filtered op pass or the generic one which exists on non-modules should work here.

mlir/include/Quantum/Transforms/Passes.td
```diff
@@ -66,6 +66,12 @@ def SplitMultipleTapesPass : Pass<"split-multiple-tapes"> {
     let summary = "Given a qnode containing multiple tapes, split each tape into its own function.";
 }
 
+def SplitNonCommutingPass : Pass<"split-non-commuting"> {
```
rniczh: add "mlir::ModuleOp" https://github.com/PennyLaneAI/catalyst/pull/2437#discussion_r2796181956

rniczh:
```cpp
```

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
rniczh: Thanks for pointing that out! Yeah, since I need to insert functions into the module op, I think the module pass is the best fit here. I will update this pass to the module pass. Just recorded it in the following comment. https://github.com/PennyLaneAI/catalyst/pull/2437#discussion_r2796246283

> The question is which approach would be the best for our "QNode" passes?

I prefer the more restricted option, filtered operation pass, so module pass here. This way, we don't need to perform any cast, and `getOperation()` is guaranteed to return a `ModuleOp`.

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
rniczh: Done. https://github.com/PennyLaneAI/catalyst/pull/2437/commits/29fe86e96195ee777a3009796ae16bfacad5bbb1

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
rniczh: Done. https://github.com/PennyLaneAI/catalyst/pull/2437/commits/29fe86e96195ee777a3009796ae16bfacad5bbb1

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
rniczh: Just confirmed with @lillian542, we can do the rounded down here. I updated the description in the changelog and this PR as well, it should be more clear now. https://github.com/PennyLaneAI/catalyst/pull/2437/commits/9133c9f5bbee19591775466502caaf14146ce327

mlir/include/Quantum/Transforms/Passes.td
rniczh: https://github.com/PennyLaneAI/catalyst/pull/2437/commits/dbe6a9830d1a23395b7dae5533c4900182142982

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
rniczh: I updated it in this commit: https://github.com/PennyLaneAI/catalyst/pull/2437/commits/dbe6a9830d1a23395b7dae5533c4900182142982

mlir/lib/Catalyst/Transforms/ApplyTransformSequencePass.cpp
sengthai: > But one of the description of this story mentioned that it needs to raise an error if there are unsupported operations.

for unspported operations, i think because of missing defining dependentDialects in SplitNonCommutingPass in pass.td ?

Otherwise, if that's a bug, I believe this should have its own PR. So, it is convenient for us to track. Also, this change might impact other passes because ' ApplyTransformSequencePass` is used every time in Catalyst's pipeline.

rniczh: No, i mean, since this PR is not going to support all the measurement ops, so if there are some unsupported operations, it should raise an error, but if this pass is run by the `ApplyTransformSequencePass`, it cannot be captured since it just ignore the failure

rniczh: Hmm...It seems to me that the issue is not such trivial, so name sequence pass is designed to be like this, 

In `catalyst/mlir/llvm-project/mlir/lib/Dialect/Transform/IR/TransformOps.cpp`:
You will see it returns a silence error. That's mean if this pass is run by trasnform sequence pass, and you wouldn't get any error even you return a failure.
```
// Run the pass or pass pipeline on the current target operation.
if (failed(pm.run(target))) {
  auto diag = emitSilenceableError() << "pass pipeline failed";
  diag.attachNote(target->getLoc()) << "target op";
  return diag;
}
```

rniczh: Just discussed with @paul0403, we can still capture those slience error, no need to 100% follows what mlir did

sengthai: how we still capture those silence error? 🤔

sengthai (approved):
Thanks for working on this one, @rniczh! Well done! 🥳

mlir/lib/Quantum/Transforms/split_to_single_terms.cpp
```diff
@@ -50,6 +50,15 @@ struct SplitToSingleTermsPass : public impl::SplitToSingleTermsPassBase<SplitToS
         tensor::FromElementsOp fromElementsWrapper;
     };
 
+    /// Check if an operation is a supported measurement operation.
+    static bool isSupportedMeasOp(Operation *op)
+    {
+        if (auto measOp = dyn_cast<MeasurementProcess>(op)) {
+            return isa<ExpvalOp>(op);
+        }
+        return false;
```
dime10: Not a big deal, but I guess the double cast is redundant?
```suggestion
        return isa<ExpvalOp>(op);
```

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
```diff
@@ -0,0 +1,455 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Split non-commuting pass: splits quantum functions that measure non-commuting
+// observables into multiple executions, with one group per observable.
+//
+// Precondition: split-to-single-terms pass has been run first to decompose Hamiltonian expvals.
```
dime10: I was going to ask what would happen if the precondition is not fulfilled, but I guess this comment might be outdated since the pass always runs split-to-single-terms?

dime10 (approved):
Thanks @rniczh 🚀

mlir/lib/Quantum/Transforms/split_to_single_terms.cpp
rniczh: Yes, thanks!

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
rniczh: Oh, yeah, exactly. And if the `split-to-single-terms` is not performed first, the behaviour of this pass would not really make sense, even it still works. Since the goal of  the`split-non-commuting` pass is to split the non-commuting observables to multiple execution, but for hamiltonian expval values `exp(....hamiltonian ops...)`, it  just create **one** group, because the current grouping strategy is just dispatch each `exp(...)` to a single execution. So it requires the `split-to-single-terms` to be executed first.

rniczh: Removed this description: https://github.com/PennyLaneAI/catalyst/pull/2437/commits/925861b07b4ac2d070851f0d4fd6e156b604039e

doc/releases/changelog-dev.md
```diff
@@ -409,6 +409,77 @@
...
+
+  Consider the following example:
+  ```python
+  import pennylane as qml
+  from catalyst import qjit
+  from catalyst.passes import apply_pass
+
+  @qjit
+  @apply_pass("split-non-commuting")
+  @qml.qnode(qml.device("lightning.qubit", wires=3))
```
lillian542:
```suggestion

  @qjit
  @qml.transform(pass_name="split-non-commuting")
  @qml.qnode(qml.device("lightning.qubit", wires=3))
```

doc/releases/changelog-dev.md
```diff
@@ -409,6 +409,77 @@
...
+  def circuit():
+      # Hamiltonian H = Z(0) + 2 * X(1) + 3 * Identity
+      return qml.expval(qml.Z(0) + 2 * qml.X(1) + 3 * qml.Identity(2))
```
lillian542:
```suggestion
      # Hamiltonian H = Z(0) + 2 * X(0) + 3 * Identity
      return qml.expval(qml.Z(0) + 2 * qml.X(0) + 3 * qml.Identity(2))
```
I think the purpose of the transform will be clearer if the expval includes non-commuting measurements that need to be split

frontend/test/pytest/test_split_non_commuting.py
```diff
@@ -0,0 +1,263 @@
...
+        return group0(), group1(), group2(), group3()
+
+    def post_processing():
+        term1, term2, term3, term4 = circ2()
+        # Compute weighted sum using the post-processing function
+        return post_process_fn(term1, term2, term3), term4
+
+    # Validate that the pass was applied
+    assert "hamiltonian" in circ1.mlir
+    assert "hamiltonian" not in circ1.mlir_opt
```
lillian542:
```suggestion
    assert "Hamiltonian" not in circ1.mlir_opt
```
This is why this sanity check didn't catch that the pass wasn't being used with the `apply_pass` decorator - the `mlir_opt` includes `Hamiltonian`, not `hamiltonian` 🙈 Same in the other tests below.

Can we update this in the `split_to_single_terms` tests too?

mlir/test/Quantum/SplitNonCommutingTest.mlir
```diff
@@ -0,0 +1,142 @@
...
+    %obs_x1 = quantum.namedobs %q1[PauliX] : !quantum.obs
+    %obs_y2 = quantum.namedobs %q2[PauliY] : !quantum.obs
+    %identity = quantum.namedobs %q2[Identity] : !quantum.obs
+    %expval_z0 = quantum.expval %obs_z0 : f64
+    %expval_x1 = quantum.expval %obs_x1 : f64
+    %expval_y2 = quantum.expval %obs_y2 : f64
+    %expval_identity = quantum.expval %identity : f64
+    quantum.dealloc %reg : !quantum.reg
+    quantum.device_release
+    return %expval_z0, %expval_x1, %expval_y2, %expval_identity : f64, f64, f64, f64
```
lillian542, line 140: Can we add a test in this format that verifies the behaviour when the transformed MLIR contains a Hamiltonian (akin to the first test in the pytest file)?

frontend/test/pytest/test_split_non_commuting.py
```diff
@@ -0,0 +1,263 @@
...
+        qp.CNOT(wires=[0, 1])
+        return (
+            qp.expval(qp.Z(wires=0)),
+            qp.expval(qp.Y(wires=1)),
+            qp.expval(qp.X(wires=0)),
+        )
+
+    res_ref = circuit_ref()
+    assert res == res_ref
+
```
lillian542, line 261: Can we add a test for the error message raised when a non-expval measurement process is included in the return?

lillian542 (commented):
Thanks @rniczh! Looks good, everything seems to behave as expected :) I've made a couple of minor suggestions to the docs and tests, but after that no further comments from me!

frontend/test/pytest/test_split_non_commuting.py
rniczh: Thanks! I updated in this commit: https://github.com/PennyLaneAI/catalyst/pull/2437/commits/72211f391588492f376904a300ea6978f5b2b31e

rniczh: It seems like that `hamiltonian` is the correct one, instead of `Hamiltonian`. The issue of `apply_pass` didn't work is due to the program capture. But the tests here didn't use program capture at all. I changed it back in the commit: https://github.com/PennyLaneAI/catalyst/pull/2437/commits/ff19cd7c971b8d69f19b9048f383307c82f6e573.

mlir/test/Quantum/SplitNonCommutingTest.mlir
rniczh: Sure, I added the test here: https://github.com/PennyLaneAI/catalyst/pull/2437/commits/ead8b928c4f8bd84f98aed152aa304227f214bc3

frontend/test/pytest/test_split_non_commuting.py
rniczh: Sure, I added the test here: https://github.com/PennyLaneAI/catalyst/pull/2437/commits/ff19cd7c971b8d69f19b9048f383307c82f6e573

frontend/test/pytest/test_split_non_commuting.py
lillian542: But the idea with the sanity check is that it should be something that _fails_ if the pass wasn't applied. The pass should remove any hamiltonians from the circuit, and we want to confirm that that has happened. But `assert "hamiltonian" not in circ1.mlir_opt` always passes, even if there are still Hamiltonians in the circuit, because `mlir_opt` is in LLVM, and it capitalizes Hamiltonian. 

You can check this manually. This will pass without any assert error, even though there is clearly still a Hamiltonian in the optimized circuit:

```
dev = qp.device("lightning.qubit", wires=3)

@qp.qjit
@qp.qnode(dev)
def circ1():
    return qp.expval(qp.X(0)+qp.Y(0))


assert "hamiltonian" in circ1.mlir
assert "hamiltonian" not in circ1.mlir_opt
```

However, `assert "Hamiltonian" not in circ1.mlir_opt` _will_ raise an assert error on this circuit that hasn't removed the Hamiltonian. So that's the assert that would confirm that the pass has genuinely been applied.

lillian542 (approved):
Thanks @rniczh, this looks great 🚀

frontend/test/pytest/test_split_non_commuting.py
rniczh: Clearly understand now! it's runtime function `__catalyst__qis__HamiltonianObs` instead of the one in quantum dialect. Fixed in this commit: https://github.com/PennyLaneAI/catalyst/pull/2437/commits/a8ca231c7c63bde25b3a0b0d5b40908199f796ad

PR #2441, opened 2026-01-27
Split to single terms is added to Catalyst in MLIR

rniczh:
**Context:**

This PR introduces a new MLIR transformation pass `--split-to-single-terms` that splits Hamiltonian expectation value measurements into individual single-term measurements with classical post-processing.

Before:
```mlir
func @circ1(%arg0) -> (tensor<f64>, tensor<f64>) {qnode} {  
    // ... quantum ops ...  
    %H = quantum.hamiltonian(%coeffs) %obs1, %obs2 : !quantum.obs  
    %result1 = quantum.expval %H : f64   // H = c_0 * (Z x X) + c_1 * Y
    %result2 = quantum.expval %Z : f64  
    return %result, %result2}
```

After:
```mlir
func @circ1.quantum() -> (tensor<f64>, tensor<f64>, tensor<f64>) {qnode} {  
    // ... quantum ops ...  
    %expval0 = quantum.expval %obs1 : f64  
    // Individual leaf expvals  
    %expval1 = quantum.expval %obs2 : f64  
    %other = quantum.expval %Z : f64  
    return %expval0, %expval1, %other
}
func @circ1(%arg0) -> (tensor<f64>, tensor<f64>) {  
    // ... setup ...
    %call:3 = call @circ1.quantum(%arg0)  
    // Extract coefficients and compute weighted sum  
    %result = c0 * %call#0 + c1 * %call#1 
    return %result, %call#2  
}
```

TODO: 
~- [ ] Intergate to cataylst pipeline~

Now, we just use the `apply_pass` to enable this pass from frontend

e.g.
```python
@qjit
@apply_pass("split-to-single-terms")
@qml.qnode(dev)
def circ1():
    return qml.expval(2 * (qml.Z(0) @ qml.X(1)) + 3 * qml.Y(2)), qml.expval(qml.Z(1))

print(circ1())
```

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-107543]


lillian542 (commented):
Thanks @rniczh! I haven't read through it carefully yet, but based on examining the IR before/after and the execution results for various examples, things here make sense to me, and I'd say it does what we expect it to do and covers the cases we want to cover. A few thoughts:

1. Kind of a detail, but I notice that the args are only used in `func.func public @circ_0` and not in `func.func public @circ.quantum_0` - they're the coefficients for post-processing - so its a little confusing that they are passed into the `quantum` function. But maybe trying to decide which things to pass along instead of just passing everything through is playing with fire :sweat_smile: 

2. Also wondering if @dime10 has thoughts here regarding the `qnode` attribute with the quantum execution function vs the post-processing function - both the `circ_0` and the `circ.quantum_0` have the `qnode` attribute now. We can always come back and adjust if needed, so not a blocking concern.

3. This can be looked into separately, it's out of scope for this task, but this pass doesn't work with the `outline_state_evolution_pass`. I'll add a story to track ensuring compatibility with `outline_state_evolution_pass` as a task for later (at least for `split_non_commuting`).

rniczh:
Thanks @lillian542 !
1. Yes, I agree, and the reason of why those args still be passed to the quantum function is beacuse it was cloned from the orignial function, and now I think we can check unused args and remove them. I fixed it in this commit https://github.com/PennyLaneAI/catalyst/pull/2441/commits/a80bfaa300cfa691b9f20074c3e07f00a9fd90bf

rniczh:
@lillian542 

3. For the issue you mentioned, that was because I didn't reset the result attribute when I updating the function type. So when it switch to xdsl, they will verify the IR and raise an error for this.

```cpp
// Update res_attrs to match the new number of return values
unsigned numResults = quantumFuncType.getNumResults();
SmallVector<DictionaryAttr> newResAttrs(numResults,
                                        DictionaryAttr::get(quantumFunc.getContext()));
function_interface_impl::setAllResultAttrDicts(quantumFunc, newResAttrs);
```

I fixed it in the commit https://github.com/PennyLaneAI/catalyst/pull/2441/commits/6d8579355ffada4a72ecc2e66e63191eca748101 and passed the following test
```python
import pennylane as qp
from catalyst.passes import apply_pass
from catalyst.python_interface.transforms import outline_state_evolution_pass

qp.capture.disable()


@qp.qjit(keep_intermediate=True)
@apply_pass("split-to-single-terms")
@outline_state_evolution_pass
@qp.set_shots(1000)
def circ_stst():
    qp.RX(1.23, 0)
    qp.RY(0.45, 1)
    return (
        qp.expval(2 * qp.Y(0) + qp.X(1)),
        qp.expval(0.7 * qp.Y(0) + 3 * qp.X(1)),
        qp.counts(wires=[0, 1]),
    )


print(circ_stst())
```

doc/releases/changelog-dev.md
```diff
@@ -130,6 +130,53 @@
   QEC state preparation operations.
   [(#2424)](https://github.com/PennyLaneAI/catalyst/pull/2424)
 
+* A new compiler pass `split-to-single-terms` has been added to optimize QNode functions containing
+  Hamiltonian expectation values by spliting them into individual leaf observable expvals.
```
lillian542: I think something like this makes more sense - its not so much an optimization as something that we sometimes need to do because the backend doesn't support the more optimal version 😅 

```suggestion
* A new compiler pass `split-to-single-terms` has been added for QNode functions containing
  Hamiltonian expectation values. It facilitates execution on devices that don't natively support expectation values of sums of observables by splitting them into individual leaf observable expvals.
```

frontend/test/pytest/test_split_to_single_terms.py
```diff
@@ -0,0 +1,134 @@
...
+    dev = qml.device("lightning.qubit", wires=3)
+    hamiltonian_obs, post_process_fn = hamiltonian
+
+    # Circuit with Hamiltonian observable
+    # Expected: split into individual terms with coefficients
+    @qjit
+    @apply_pass("split-to-single-terms")
+    @qml.qnode(dev)
+    def circ1():
+        return qml.expval(hamiltonian_obs), qml.expval(qml.Z(1))
```
lillian542, line 55: For a lot of the observables in the Hamiltonians above, the output for `expval(obs)` is just 0, which makes this test less convincing in terms of correctly mapping the coefficients to the various terms. 

Maybe we could add a `qml.Rot` on some random angles on each wire, so that the results are more specific?

frontend/test/pytest/test_split_to_single_terms.py
```diff
@@ -0,0 +1,134 @@
...
+    """
+    Test split-to-single-terms with tensor product observables.
+    """
+    dev = qml.device("lightning.qubit", wires=3)
+
+    @qjit
+    @apply_pass("split-to-single-terms")
+    @qml.qnode(dev)
+    def circ1():
+        return qml.expval(2 * (qml.Z(0) @ qml.X(1)) + 3 * qml.Y(2)), qml.expval(qml.Z(1))
```
lillian542, line 100: Similarly, `Z(0)@X(1)` and `Y(2)` are both 0 here, so this isn't really testing the correct mapping of the coefficients.

frontend/test/pytest/test_split_to_single_terms.py
```diff
@@ -0,0 +1,134 @@
...
+    Identity observables are removed from the quantum circuit since their
+    expectation value is always 1, and their coefficient is added in post-processing.
+    """
+    dev = qml.device("lightning.qubit", wires=3)
+
+    @qjit
+    @apply_pass("split-to-single-terms")
+    @qml.qnode(dev)
+    def circ1():
+        return qml.expval(qml.Z(0) + 2 * qml.X(1) + 0.7 * qml.Identity(2))
```
lillian542, line 139: Its less relevant to the purpose of the test, since this one is really about how the `Identity` is handled, but `2 * X(1)` here is also just `2*0` without some rotation around Y on wire 1.

mlir/test/Quantum/SplitToSingleTermsTest.mlir
```diff
@@ -0,0 +1,229 @@
...
+// Input circuit has:
+//   - Hamiltonian H = 1.0 * (Z x X) + 1.0 * (arg0 * Y)
+//   - expval(H) returns weighted sum
+//   - expval(Z) on qubit 1 (non-Hamiltonian)
+//
+// After transformation:
+//   - circ.quantum: returns individual expvals [<Z x X>, <Y>, <Z>]
+//   - circ: calls circ.quantum, computes weighted sum, returns [<H>, <Z>]
+
+module @circ {
```
lillian542: Could we update this test (or add additional tests) so that it would have caught the two issues we encountered earlier? I.e. adding a `CHECK NOT` making sure there is no `quantum.hamiltonian` present in the transformed circuit, and adding a couple of gates in the circuit before the MPs and making sure that they are present in `circ.quantum` but not in `circ`. 

In fact, would it work to check that there are no `quantum` anythings inside `circ`?

frontend/test/pytest/test_split_to_single_terms.py
```diff
@@ -0,0 +1,134 @@
...
+"""
+
+import numpy as np
+import pennylane as qml
+import pytest
+
+from catalyst import qjit
+from catalyst.passes import apply_pass
+
+
```
lillian542, line 26: I know this should *technically* be tested elsewhere because its related to the transform infrastructure and not be the responsibility of these tests, but as a sanity check, can we throw in a 
```
assert "hamiltonian" in circ1.mlir
assert "hamiltonian" not in circ1.mlir_opt
```
to validate that the pass _did_ change the executed program? 

Otherwise the test technically can't distinguish between "the pass does what is expected" and "the pass wasn't applied".

frontend/test/pytest/test_split_to_single_terms.py
```diff
@@ -0,0 +1,134 @@
...
+    "hamiltonian",
+    [
+        [qml.Z(0) + qml.X(1) + 2 * qml.Y(2), lambda term1, term2, term3: term1 + term2 + 2 * term3],
+        [
+            3 * qml.Z(0) + qml.X(1) + 2 * qml.Y(2),
+            lambda term1, term2, term3: 3 * term1 + term2 + 2 * term3,
+        ],
+    ],
+)
+def test_split_to_single_terms_integration(hamiltonian):
```
lillian542, line 38: Can we use [the fixture added here](https://github.com/PennyLaneAI/catalyst/pull/2438) to have this run with both `capture` and the old frontend?

For the test for Identity, it will need the `capture_todo` label

lillian542 (commented):
I have a few comments/suggestions for the tests :)

doc/releases/changelog-dev.md
rniczh: Oh, yes, shouldn't be the `optimize`.

frontend/test/pytest/test_split_to_single_terms.py
rniczh: Done: https://github.com/PennyLaneAI/catalyst/pull/2441/commits/059c821fdeb38390995a13ecf5f3a9edc7f28a5f For identiy, I didn't set any `capture_todo` label, but seems like it will mark as xfail as expected?

frontend/test/pytest/test_split_to_single_terms.py
rniczh: Sure, I added those asserations in this commit https://github.com/PennyLaneAI/catalyst/pull/2441/commits/059c821fdeb38390995a13ecf5f3a9edc7f28a5f

frontend/test/pytest/test_split_to_single_terms.py
rniczh: Thanks for pointing out. I updated the test in this commit: https://github.com/PennyLaneAI/catalyst/pull/2441/commits/2183d4ac2fa05e91335cc6cd7f7ca580d63a217f

frontend/test/pytest/test_split_to_single_terms.py
rniczh: Done. Same as https://github.com/PennyLaneAI/catalyst/pull/2441#discussion_r2765649961

frontend/test/pytest/test_split_to_single_terms.py
rniczh: Done. Same as https://github.com/PennyLaneAI/catalyst/pull/2441#discussion_r2765649961

mlir/test/Quantum/SplitToSingleTermsTest.mlir
rniczh: Sure, no problem https://github.com/PennyLaneAI/catalyst/pull/2441/commits/059c821fdeb38390995a13ecf5f3a9edc7f28a5f

frontend/test/pytest/test_split_to_single_terms.py
lillian542: Thanks, what you did is what I meant, label is probably not a good word for it 😅

lillian542 (approved):
Thanks @rniczh! I'm not experienced enough with C++ to provide a meaningful review for the implementation in the `.cpp` file, but in terms of the requirements for behaviour/functionality, this looks great!

doc/releases/changelog-dev.md
```diff
@@ -153,6 +153,53 @@
   QEC state preparation operations.
   [(#2424)](https://github.com/PennyLaneAI/catalyst/pull/2424)
 
+* A new compiler pass `split-to-single-terms` has been added for QNode functions containing
+  Hamiltonian expectation values. It facilitates execution on devices that don't natively support expectation values of sums of observables by splitting them into individual leaf observable expvals.
+  [(#2441)](https://github.com/PennyLaneAI/catalyst/pull/2441)
+
```
joeycarter, line 211:
```suggestion

Consider the following example:
```

(or similar)

doc/releases/changelog-dev.md
```diff
@@ -153,6 +153,53 @@
...
+  ```mlir
+  func @circ1(%arg0) -> (tensor<f64>, tensor<f64>) {qnode} {
+      // ... quantum ops ...
+      %H = quantum.hamiltonian(%coeffs) %obs1, %obs2 : !quantum.obs
+      %result1 = quantum.expval %H : f64   // H = c_0 * (Z x X) + c_1 * Y
+      %result2 = quantum.expval %Z : f64
+      return %result, %result2}
+  ```
+
+  After:
```
joeycarter:
```suggestion
  **After:**
```

doc/releases/changelog-dev.md
```diff
@@ -153,6 +153,53 @@
...
+      %other = quantum.expval %Z : f64
+      return %expval0, %expval1, %other
+  }
+  func @circ1(%arg0) -> (tensor<f64>, tensor<f64>) {
+      // ... setup ...
+      %call:3 = call @circ1.quantum()
+      // Extract coefficients and compute weighted sum
+      %result = c0 * %call#0 + c1 * %call#1
+      return %result, %call#2
+  }
```
joeycarter: And same here.

doc/releases/changelog-dev.md
```diff
@@ -153,6 +153,53 @@
...
+  The pass transforms the function by splitting the Hamiltonian into individual observables:
+
+  **Before:**
+  ```mlir
+  func @circ1(%arg0) -> (tensor<f64>, tensor<f64>) {qnode} {
+      // ... quantum ops ...
+      %H = quantum.hamiltonian(%coeffs) %obs1, %obs2 : !quantum.obs
+      %result1 = quantum.expval %H : f64   // H = c_0 * (Z x X) + c_1 * Y
+      %result2 = quantum.expval %Z : f64
+      return %result, %result2}
```
joeycarter: This example doesn't look right, there are two expvals returned here, but only one in the circuit above. This is what I get:

```mlir
func.func public @circuit(%arg0: tensor<1xi64>) -> tensor<f64> attributes {qnode} {
  %cst = stablehlo.constant dense<1.000000e+00> : tensor<f64>
  %0 = quantum.alloc( 3) : !quantum.reg
  %2 = quantum.namedobs %1[ PauliZ] : !quantum.obs
  %3 = quantum.extract %0[ 1] : !quantum.reg -> !quantum.bit
  %4 = quantum.namedobs %3[ PauliX] : !quantum.obs
  %5 = quantum.extract %0[ 2] : !quantum.reg -> !quantum.bit
  %6 = quantum.namedobs %5[ PauliY] : !quantum.obs
  %7 = stablehlo.convert %arg0 : (tensor<1xi64>) -> tensor<1xf64>
  %8 = quantum.hamiltonian(%7 : tensor<1xf64>) %6 : !quantum.obs
  %9 = stablehlo.broadcast_in_dim %cst, dims = [] : (tensor<f64>) -> tensor<3xf64>
  %10 = quantum.hamiltonian(%9 : tensor<3xf64>) %2, %4, %8 : !quantum.obs
  %11 = quantum.expval %10 : f64
  %from_elements = tensor.from_elements %11 : tensor<f64>
  quantum.device_release
}
```

Alternatively, is this the MLIR you wanted to illustrate, but the Python example above is not correct?

mlir/test/Quantum/SplitToSingleTermsTest.mlir
```diff
@@ -0,0 +1,293 @@
+// Copyright 2026 Xanadu Quantum Technologies Inc.
```
joeycarter, line 1: Sorry to be a pain, but the tests in these files feel more appropriate as frontend lit tests rather than as MLIR unit tests. Would it be possible to strip out all the ops that aren't necessary for the split-to-single-terms pass? For instance, I don't think we need any gate ops here, among many others. It would be easier to reason about what the pass is doing as a unit test that way. I assume you generated the input MLIR from a PennyLane program, so you could also put those into a frontend lit test and copy over the `CHECK`s you have here.

mlir/lib/Quantum/Transforms/split_to_single_terms.cpp
```diff
@@ -0,0 +1,548 @@
...
+    bool isIdentityObservable(Value obs)
+    {
+        if (auto namedObsOp = obs.getDefiningOp<NamedObsOp>()) {
+            return namedObsOp.getType() == NamedObservable::Identity;
+        }
+        return false;
+    }
+
+    /// Recursively collect all leaf observables from a potentially nested Hamiltonian
+    void collectLeafObservables(Value obs, SmallVectorImpl<Value> &leafObs)
```
joeycarter, line 63: Curious why you chose `SmallVectorImpl` here rather than `SmallVector`?

joeycarter (changes requested):
Thanks @rniczh! This is fantastic work 🤩 

I only caught one issue in the example given in the changelog, and a nitpick on the MLIR FileCheck tests, but otherwise it looks great.

doc/releases/changelog-dev.md
rniczh: Thanks! This example is wrong. What I actuall want to express is this, and I adjusted the  changelog in this commit: https://github.com/PennyLaneAI/catalyst/pull/2441/commits/aa6335b45a7d13efc3506b3281eb0c17d77f9982

```python
@qjit
@apply_pass("split-to-single-terms")
def circuit():
    # Hamiltonian H = Z(0) @ X(1) + 2*Y(2)
    return qml.expval(qml.Z(0) @ qml.X(1) + 2 * qml.Y(2))
```

And yes, It should only contain one result in the original circuit

doc/releases/changelog-dev.md
rniczh: Same as https://github.com/PennyLaneAI/catalyst/pull/2441/changes#r2770489843. Fixed in the commit https://github.com/PennyLaneAI/catalyst/pull/2441/commits/aa6335b45a7d13efc3506b3281eb0c17d77f9982

mlir/lib/Quantum/Transforms/split_to_single_terms.cpp
rniczh: The main reason to use `SmallVectorImpl` is flexibility. If a parameter uses `SmallVector`, you must specify the size N.

According to the LLVM documentation: https://llvm.org/docs/ProgrammersManual.html#llvm-adt-smallvector-h
> It’s rarely appropriate to use SmallVector<T, N> as a parameter type. If an API only reads from the vector, it should use ArrayRef. Even if an API updates the vector the “small size” is unlikely to be relevant; such an API should use the SmallVectorImpl<T> class, which is the “vector header” (and methods) without the elements allocated after it. Note that SmallVector<T, N> inherits from SmallVectorImpl<T> so the conversion is implicit and costs nothing.

So in this case where I need to modify `leafObs`, I use `SmallVectorImpl`.

joeycarter: Got it, good to know. Thanks!

mlir/test/Quantum/SplitToSingleTermsTest.mlir
lillian542: I actually requested they be added, but I'm probably also more used to frontend lit tests. But the motivation was making sure the ops ended up in the quantum function instead of the post-processing function, a previous iteration of this accidentally had them both places.

rniczh: recap: https://github.com/PennyLaneAI/catalyst/pull/2441#discussion_r2764762696

> so you could also put those into a frontend lit test and copy over the CHECKs you have here.

Just added the frontend lit test to cover the CHECK as well. https://github.com/PennyLaneAI/catalyst/pull/2441/commits/eae5f5470dd979547e1686701dd8c3c445a94864

lillian542: Can you elaborate on how we separate the responsibility of the frontend lit tests vs MLIR unit tests?

dime10: This is actually a good observation and something important to keep in mind. As a reminder, MLIR filecheck tests should
- rarely use auto-generated input: this usually means more noise and makes the tests harder to read and maintain
  - instead, tests should focus on the effects of the transform being tested, with a minimal amount of IR needed to sensibly do so
- absolutely NOT check against autogenerated output (especially from the transform being tested): this would defeat the purpose of testing

It's not much but there are some official guidelines here: https://mlir.llvm.org/getting_started/TestingGuide/#filecheck-best-practices

rniczh: @dime10 Thanks for explanation! Originally I wanted to include dependencies in the test, but that made it too complex (and more noise) as you mentioned. Now I've simplified this part of the test to mainly just verify that the operations appear. https://github.com/PennyLaneAI/catalyst/pull/2441/commits/94fe8dd85ba25f2292ca9ff0f96c279fc1cf9b91

And I think gate ops are necessary in the test from this pass's perspective. Previously we missed removing gate ops from entry function, creating redundancy. Now that gate removal in the entry function is part of the pass's functionality, so we need to cover it in mlir lit tests?

frontend/test/lit/test_split_to_single_terms.py
```diff
@@ -0,0 +1,175 @@
...
+    # CHECK: return {{.*}}, {{.*}}, %[[ONE]]
+
+    # CHECK: func.func public @circuit{{.*}}
+    # CHECK-NOT: quantum.custom.*
+    # CHECK-NOT: quantum.namedobs.*
+    # CHECK-NOT: quantum.expval.*
+    # CHECK: %[[CALL:.*]]:3 = call @circuit.quantum{{.*}}
+    # CHECK: stablehlo.multiply
+    # CHECK: stablehlo.multiply
+    # CHECK: stablehlo.multiply
```
sengthai, line 165: Just curious, are these `multiply` from `return qml.expval(qml.Z(0) + 2 * qml.X(1) + 0.7 * qml.Identity(2))` ? It seems to me that only two `multiply` ops are needed.

mlir/test/Quantum/SplitToSingleTermsTest.mlir
sengthai: >Can you elaborate on how we separate the responsibility of the frontend lit tests vs MLIR unit tests?

@lillian542  That's a good question, and I often found myself scratching my head when writing a test. 😅

Both tests intentionally verify the correctness of MLIR output from passes. In MLIR, the test ensures that the pass works as expected on a specific task or answers the question "_Does my pass output correctly when given the MLIR IR?_". The frontend lit tests answer the question "_Does my pass still output correctly when I run it from Pennylane?_".

mlir/include/Quantum/Transforms/Passes.td
```diff
@@ -66,6 +66,10 @@ def SplitMultipleTapesPass : Pass<"split-multiple-tapes"> {
     let summary = "Given a qnode containing multiple tapes, split each tape into its own function.";
 }
 
+def SplitToSingleTermsPass : Pass<"split-to-single-terms"> {
+    let summary = "Split quantum functions into single-term observables.";
```
sengthai, line 70: Is this pass working with another measurement process besides observable expvals?

mlir/test/Quantum/SplitToSingleTermsTest.mlir
joeycarter: > And I think gate ops are necessary in the test from this pass's perspective. Previously we missed removing gate ops from entry function, creating redundancy. Now that gate removal in the entry function is part of the pass's functionality, so we need to cover it in mlir lit tests?

@rniczh Ah, so you want to test that the pass moves the gates into the new function that's created. In that case I think it's fine.

joeycarter: I think we can still strip out quite a few ops from the input MLIR in these tests, with the goal of only having the minimal number of ops to test the effects of the pass. For example:

* All the `stablehlo.constant` / `tensor.extract` ops could be condensed (e.g. to a single `stablehlo.constant` that returns a scalar rather than a tensor), some other `stablehlo` ops could be removed entirely.
* The attributes alongside the qnode attribute (i.e. `diff_method`, `llvm.linkage`) can be removed.
* The upper level modules and functions, e.g. the one with `catalyst.launch_kernel`, the `@setup` / `@teardown` functions. I don't think the pass touches these at all, probably we only need a single module with the `qnode` function as the input MLIR.

mlir/lib/Quantum/Transforms/split_to_single_terms.cpp
```diff
@@ -0,0 +1,548 @@
...
+
+        // Find all qnode functions with Hamiltonian expvals
+        SmallVector<func::FuncOp> funcsToProcess;
+        moduleOp->walk([&](func::FuncOp funcOp) {
+            if (!funcOp->hasAttr("qnode")) {
+                return;
+            }
+
+            bool hasHamiltonian = false;
+            funcOp.walk([&](ExpvalOp expvalOp) {
```
sengthai, line 505: Since ExpValOp normally happens in the end of circuit. I would recommend to use`WalkOrder::PostOrder` and `ReverseIterator`. So, it will traverse from bottom to top.
```suggestion
            funcOp.walk<WalkOrder::PostOrder, ReverseIterator>([&](ExpvalOp expvalOp) {
```

mlir/lib/Quantum/Transforms/split_to_single_terms.cpp
```diff
@@ -0,0 +1,548 @@
...
+            if (hasHamiltonian) {
+                funcsToProcess.push_back(funcOp);
+            }
+        });
+
+        for (func::FuncOp origFunc : funcsToProcess) {
+            Location loc = origFunc.getLoc();
+            OpBuilder moduleBuilder(origFunc);
+
+            // Clone origFunc -> quantumFunc (<circuit_name>.quantum)
```
sengthai: Minor naming.
```suggestion
            // Clone origFunc -> origFunc.quantum (<circuit_name>.quantum)
```

mlir/lib/Quantum/Transforms/split_to_single_terms.cpp
```diff
@@ -0,0 +1,548 @@
...
+            // quantumFunc will contain the quantum operations and return individual expvals
```
sengthai:
```suggestion
            // origFunc.quantum will contain the quantum operations and return individual expvals
```

mlir/lib/Quantum/Transforms/split_to_single_terms.cpp
```diff
@@ -0,0 +1,548 @@
...
+            // origFunc will be the entry point that calls quantumFunc and does post-processing
+            IRMapping cloneMapping;
+            std::string quantumFuncName = origFunc.getName().str() + ".quantum";
```
sengthai, line 526: And I wonder what if the user has the same function name as `<circuit_name>.quantum`? What will happen? 🤔
I know that it is impossible to generate from current python, just what if MLIR is generated from somewhere else with `.quantum`

sengthai (commented):
Looks good so far! Before diving deep to main function `rewriteQuantumFunc`, I have a few questions here.

frontend/test/lit/test_split_to_single_terms.py
rniczh: Since `1.0`, `2.0`, `0.7` are passed to the function instead of the constant number in the begining (when converting from frontend to MLIR). So here we can only see the args.

mlir/include/Quantum/Transforms/Passes.td
rniczh: No, it's not expected to split Hamiltonians inside any measurement process that isn't `expval`.

mlir/lib/Quantum/Transforms/split_to_single_terms.cpp
rniczh: I think this doesn't apply here because you still need to traverse all operations. Even if you encounter it early, you can't be sure there aren't more operations later. `PostOrder` traversal would be appropriate when you genuinely need to process from back to front, or as you mentioned, but only when there's a single occurrence that allows early return.

mlir/lib/Quantum/Transforms/split_to_single_terms.cpp
rniczh: Good catch! Insert to the symbol table can handle this automatically. Fixed: https://github.com/PennyLaneAI/catalyst/pull/2441/commits/39f0cac9364bbd26b73f998c969f8f77260fc6ac

mlir/test/Quantum/SplitToSingleTermsTest.mlir
dime10: > **Both tests intentionally verify the correctness of MLIR output from passes.** In MLIR, the test ensures that the pass works as expected on a specific task or answers the question "Does my pass output correctly when given the MLIR IR?". The frontend lit tests answer the question "Does my pass still output correctly when I run it from Pennylane?".

This is a difficult question, but I think the highlighted statement is not quite right. In order the clearly divide responsibilities of the different test suites, the frontend tests should only ever test the frontend components of a pass, if there even are any (e.g. that the pass primitive correctly appears in the IR). They should not test the action (or output) of a pass written in MLIR. We also might use them to test that the lowering mechanism to MLIR is working correctly.
On the other hand, the action of MLIR passes should be (unit) tested in the MLIR component of Catalyst.

rniczh: @joeycarter Thanks for the feedbacks! I refactored the test in this commit https://github.com/PennyLaneAI/catalyst/pull/2441/commits/369398154c995abf28d68fb01ff59ead27f9ac66 It's way more clear now

joeycarter: Looks much better, thanks! :rocket: 

I'll resolve this thread, but feel free to continue discussing.

joeycarter (approved):
All good on my end! 🚀

sengthai (approved):
Excellent! Thanks @rniczh

PR #2442, opened 2026-01-27
Small bug fix in dynamic allocation

paul0403:
**Context:**
There was a small bug in dynamic allocation with control flow where I remap the outer qregs to the qreg args of the control flow region. I was checking whether the entire map was empty instead of checking whether each qreg was needed.

Check each qreg.

**Benefits:**
"Global wires" can be used together with dynamically allocated wires inside a control flow.


sengthai (approved):
Thanks for the quick fix! @paul0403

rniczh (approved):
Nice 💯

PR #2444, opened 2026-01-27
[LLVM UPDATE] Patch MLIR Python generics

mehrdad2m:
**Context:**
Summary
MLIR has updated [Value type](https://github.com/llvm/llvm-project/commit/31536e6e9a54f54b26f97ce227d294d29b2b845f#diff-acfa9a5ea4439eae3f3e5a57ccbb8e72944842f6ae7816624c1d75e2908f3e62) to be generic in the type of the value it holds and now the generated dialect bindings emit `Value[Type] / OpResult[Type]` annotations. 

This would crash Catalyst when importing the generated python files because JAX's LLVM version is older and doesn’t expose `__class_getitem__` which causes an error complaining about indexing.

patched `mlir/llvm-project/mlir/tools/mlir-tblgen/OpPythonBindingGen.cpp` to postpone evaluation of type hints to avoid the crash while import. Thanks @rniczh for the idea!

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


paul0403 (commented):
Personally I'm ok with this if we have to do this, we already patch a bunch of stuff due to jax... but I do wonder whether there's a better way 🤔 

Because we never emit `Value` or `OpResult` raw in any IR right? So this is purely just for python imports missing some class methoeds....

paul0403 (approved):

mehrdad2m:
@rniczh idea worked like a charm. Tagged for another round of review. Thanks!

paul0403 (approved):

rniczh (approved):
Thanks @mehrdad2m 🎉

dime10:
definitely hurts to see more patches being added 💔 

Is there an known end-of-life date for this patch? If so please add it somewhere in the code.

Issue #1912, opened 2025-07-17
Compiler crash while computing circuit gradient (qjit and vmap) with lightning.qubit and lightning.kokkos


dime10:
A forum user is [reporting](https://discuss.pennylane.ai/t/compiler-crash-while-computing-circuit-gradient-qjit-and-vmap-with-lightning-qubit-and-lightning-kokkos/8773) a crash when using jax.grad on top of a qjit program with vmap.

While there are known Enzyme issues with complex programs like those using `vmap`, this one crashes at the MLIR layer so is likely a different problem.

The report is reproduced below:

---

I have an issue computing the ‘jit-ed’ and ‘vmap-ed’ gradient of a circuit. It seems that the compiler is crashing for some reason I do not understand. A minimal example and some further information are written below. Does anybody have an idea how to resolve this issue?

```py
import catalyst
import pennylane as qml
import jax.numpy as jnp

dev = qml.device("lightning.qubit", wires=2)
    
@qml.qnode(dev, interface='jax')
def circuit(x):
    qml.RX(x, wires=0)
    return qml.expval(qml.PauliZ(0))

batch_circuit = catalyst.vmap(circuit)
compiled_circuit = catalyst.qjit(batch_circuit)
grad_compiled = catalyst.grad(compiled_circuit)  # note this dispatches to jax.grad, which inserts catalyst.jvp within the compiled program

batch_input = jnp.array([0.1, 0.2, 0.3])

print(compiled_circuit(batch_input))
print(grad_compiled(batch_input))
```

<details open>
<summary>Traceback</summary>
<br>

2025-07-17 09:07:09.765846: W external/xla/xla/service/gpu/nvptx_compiler.cc:760] The NVIDIA driver’s CUDA version is 12.4 which is older than the ptxas CUDA version (12.9.86). Because the driver is older than the ptxas version, XLA is disabling parallel compilation, which may slow down compilation. You should update your NVIDIA driver or use the NVIDIA-provided CUDA forward compatibility packages.

[...]

PR #332, opened 2023-10-25
Fix compilation of Jacobian with for loops

rmoyard:
### Before submitting

Please complete the following checklist when submitting a PR:

- [ ] All new functions and code must be clearly commented and documented.

- [ ] Ensure that code is properly formatted by running `make format`.
      The latest version of black and `clang-format-14` are used in CI/CD to check formatting.

- [ ] All new features must include a unit test.
      Integration and frontend tests should be added to [`frontend/test`](../frontend/test/),
      Quantum dialect and MLIR tests should be added to [`mlir/test`](../mlir/test/), and
      Runtime tests should be added to [`runtime/tests`](../runtime/tests/).

When all the above are checked, delete everything above the dashed
line and fill in the pull request template.

------------------------------------------------------------------------------------------------------------

**Context:**

**Description of the Change:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

PR #2445, opened 2026-01-27
[LLVM update] Fix CI segfaults in the xfailed grad+vmap tests

mehrdad2m:
**Context:**
There has been a old issue with using grad + vmap captured by https://github.com/PennyLaneAI/catalyst/issues/1912 that seems to be partially resolved as we updated LLVM + Enzyme. This most likely related to the update from enzyme. As a result of this three of the tests previously marked  as xfail now have changed status. One passes and the other two segfault (yikes!). Turns out that the example discussed in https://github.com/PennyLaneAI/catalyst/issues/1912 which is the floowing:
```python
@vmap
def circuit(x):
    qml.RX(x, wires=0)

@qjit
@grad
def loss(x):
    return jnp.sum(circuit(x))

print(loss(jnp.array([0.1, 0.2, 0.3])))
```
now runs with no compilation error. However a case like below segfaults at runtime:

```python 
@vmap
def circuit(x, data):
    qml.RX(x, wires=0)
    qml.RX(data, wires=0)

@qjit
@grad
def loss(x, data):
    return jnp.sum(circuit(x, data))

print(loss(jnp.array([0.1, 0.2, 0.3]), jnp.array([0.1, 0.2, 0.3])))
```
the main difference is the addition of `data` as the second argument.

Remove the xfail for the passed test and move the other two xfails inside the test to avoid their execution and hence the segmentation fault.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -19,6 +19,7 @@
   [(#2415)](https://github.com/PennyLaneAI/catalyst/pull/2415)
   [(#2416)](https://github.com/PennyLaneAI/catalyst/pull/2416)
   [(#2444)](https://github.com/PennyLaneAI/catalyst/pull/2444)
+  [(#2445)](https://github.com/PennyLaneAI/catalyst/pull/2445)
 
   - The StableHLO version has been updated to
   [v1.13.7](https://github.com/openxla/stablehlo/tree/v1.13.7).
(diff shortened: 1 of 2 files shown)


paul0403 (approved):

dime10:
Looking at the tests in question, it's weird that the error there could have gone away by updating Enzyme, because they reference PR #332, which fixes a bug in our gradient transform wrt some instruction placement. At the time we decided not to merge 332 because we figured a compiler error was better than a crash (that arose once we fixed the compiler error). If the error is no longer raised anyway, I think it would be worth reviving 332 to see if the issue described there has somehow gone away or else merge it.

kipawaa (approved):

PR #2446, opened 2026-01-27
Add support for measurements and observables to `null.qubit` runtime resource tracking

jzaia18:
**Context:**
Resource tracking does not currently take into account measurements or observables. This leads to confusing behaviour with `qml.specs` where measurements are generally available except when using `level="device"`.

Tracks observables and measurements in the `ResourceTracker` object.

**Benefits:**
Enables `qml.specs` to retrieve measurement data from runtime resource tracking

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-109207]


runtime/lib/backend/null_qubit/ResourceTracker.hpp
```diff
@@ -18,6 +18,8 @@
 #include <unordered_map>
 #include <vector>
 
+#include <iostream>
+
```
kipawaa:
```suggestion
```
I don't think this include is necessary

runtime/tests/Test_ResourceTracker.cpp
kipawaa, line 1: Codecov is complaining about some uncovered lines, and though some seem redundant, I think it would be good to cover them for completeness

kipawaa, line 1: Since this is for `qml.specs` integration, it would be good to include an integration test as well to ensure that the resource tracking correctly passes up to specs.

kipawaa (changes requested):
Looks good! Just a bit more testing to make codecov happy and I can approve 👍

runtime/lib/backend/null_qubit/ResourceTracker.hpp
jzaia18: Oops, forgot to remove that from testing :sweat_smile:

jzaia18:
```suggestion
```

runtime/tests/Test_ResourceTracker.cpp
jzaia18: See https://github.com/PennyLaneAI/catalyst/pull/2448

runtime/lib/backend/null_qubit/ResourceTracker.hpp
kipawaa: Ah, I see you're a man of culture as well 😉

runtime/tests/Test_ResourceTracker.cpp
kipawaa: Ah, I missed that, sorry!

kipawaa (approved):
Looks great! 💯

runtime/lib/backend/null_qubit/ResourceTracker.hpp
```diff
@@ -338,6 +365,94 @@ struct ResourceTracker final {
...
+     * @brief Registers an observable with a given id
+     *
+     * @param obs_id The observable identifier (e.g., Identity, PauliZ, Hadamard)
+     * @return ObsIdType An identifier for the created observable
+     */
+    auto Observable(ObsId obs_id) -> ObsIdType
+    {
+        ObsIdType id = static_cast<ObsIdType>(observable_names_.size() + 1);
+        std::string name = "Observable";
+        switch (obs_id) {
```
paul0403, line 378: Should we have an default case in this switch to catch errors?

runtime/lib/backend/null_qubit/ResourceTracker.hpp
```diff
@@ -138,6 +143,28 @@ struct ResourceTracker final {
...
+     * @param meas_name Optional specific measurement name to count. If empty, returns total count
+     * of all measurements
+     *
+     * @return The count of measurements matching the specified name, or total measurement count if
+     * no name provided
+     */
+    auto GetNumMeasurements(const std::string &meas_name = "") -> std::size_t
+    {
+        if (meas_name != "") {
+            return measurements_[meas_name];
```
paul0403, line 158: Here too, maybe we can raise a more meaningful error message rather than depending on whatever cpp says when the map doesn't have the key, when other parts of the code ask for an invalid measurement. Not that a user would have the means to supply anything here directly, but just for our own maintainability.

runtime/tests/Test_NullQubit.cpp
```diff
@@ -970,6 +976,61 @@ TEST_CASE("Test NullQubit device resource tracking integration", "[NullQubit]")
...
+        size_t _offset = 0U;
+        size_t _sizes[2] = {shots, num_qubits};
+        size_t _strides[2] = {1, 1};
+
+        DataView<double, 2> sample_view(_data_aligned, _offset, _sizes, _strides);
+        sim->Sample(sample_view);
+        sim->PartialSample(sample_view, {Qs[0], Qs[2]});
+    }
+
+    // sim->Counts();
```
paul0403: Is this line supposed to be removed?

paul0403 (commented):
Looks good, really nice PR! Just a few places regarding potential error cases 💯

runtime/tests/Test_NullQubit.cpp
jzaia18: Good catch

runtime/lib/backend/null_qubit/ResourceTracker.hpp
jzaia18: It returns 0 as written, which I think should be the expected behaviour. If you query for `expval(PauliZ)` and we haven't seen any, it should return 0

runtime/lib/backend/null_qubit/ResourceTracker.hpp
jzaia18: As written it would just use the literal string "Observable" as the default value. I did this to prevent errors in the future in case we create new ObsIdTypes. If it would be preferable to instead raise an error I can do this as well

paul0403 (approved):

PR #2447, opened 2026-01-27
Fix cli hang

### LINKED ISSUE #2411: `catalyst` CLI tool hangs without input file
# Issue
When calling `catalyst --list-passes` I noticed that the command hangs rather than perform the desired action. This happens as well when called directly with `catalyst`. When provided an input file (`catalyst --list-passes test.mlir`), the command runs but the file is optimized and no passes are listed. This makes the pass list inaccessible via the `catalyst` utility.

# Expected behaviour
I would expect behaviour similar to `quantum-opt`, which when called directly listens to `stdin` for mlir to process, or an error to be raised.

# MRE
`./mlir/build/bin/catalyst` from project root.

kipawaa:
**Context:**
the `catalyst` cli tool was silently listening to stdin when not provided an input file, regardless of CLI arguments except for help.

This adds a message before listening to stdin, and moves the listening to after parsing other flags like `--list-passes` and `--show-dialects`.

**Benefits:**
`catalyst` works as advertised!

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
Closes #2411


mehrdad2m (approved):
LGTM

lazypanda10117 (approved):

mlir/lib/Driver/CompilerDriver.cpp
```diff
@@ -560,6 +560,7 @@ LogicalResult runEnzymePasses(const CompilerOptions &options,
 std::string readInputFile(const std::string &filename)
 {
     if (filename == "-") {
```
mehrdad2m, line 562:
```suggestion
    if (filename == "-" && llvm::errs().is_displayed()) {
```
I am wondering if we could do something like this? If it works, it would be a more systematic fix.

mehrdad2m: inspired by https://github.com/llvm/llvm-project/blob/113f01aa82d055410f22a9d03b3468fa68600589/clang-tools-extra/clangd/tool/ClangdMain.cpp#L869-L874

kipawaa: Great find! I like this much better 💯 added in [fb8d9d2](https://github.com/PennyLaneAI/catalyst/pull/2447/commits/fb8d9d2e6705bb4b3f22695f7d294bcdf3ef1c66) and removed the hack in mlir-graph

mehrdad2m (approved):

PR #2453, opened 2026-01-28
Split compiler driver into separate components

mlxd:
**Context:** This PR divides the compiler driver into multiple components, allowing for better isolation of the corresponding interactions between them. This is a precursor to updating the target generation layer for multiple output formats.

* Isolated the CompilerDriver implementation to use multiple headers/translation units
* Added explicit namespace scoping to make clear delineations for source of truth (fixes issues with clangd)
* Added more precise timing control of the instrumentation functionality using `std::chrono`  defined ratios (ns, us, ms all supported).
* Defined a `main` driver method that ties all the surrounded functionality into a single location.

**Benefits:**
* Makes isolation of the target definition and generation calls easier
* Static analysis helps to identify the paths to and through MLIR, LLVM, and our implementations
* Access to higher-precision timer divisions for the instrumentation layer via templated arguments

**Possible Drawbacks:**
* Potential change in compilation times due to the restructured driver (LTO may be warranted) --- UPDATE: LTO did not improve times. 

**Related GitHub Issues:**


mlir/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ project(Catalyst LANGUAGES CXX C)
 set(CMAKE_BUILD_WITH_INSTALL_NAME_DIR ON)
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
-set(CMAKE_CXX_STANDARD 20 CACHE STRING "C++ standard to conform to")
+set(CMAKE_CXX_STANDARD 23 CACHE STRING "C++ standard to conform to")
```
mlxd:
```suggestion
set(CMAKE_CXX_STANDARD 20 CACHE STRING "C++ standard to conform to")
```

No need, with the LLVM backported functionality.

mlxd:
Note: LTO caused a perf regression with clang, and failed to compile using G++. Disabling it for now, as the multi-file change did not adversely affect compile times.

mlir/Makefile
```diff
@@ -75,7 +75,7 @@ llvm:
 	cmake -G Ninja -S llvm-project/llvm -B $(LLVM_BUILD_DIR) \
 		-DCMAKE_BUILD_TYPE=$(BUILD_TYPE) \
 		-DLLVM_BUILD_EXAMPLES=OFF \
-		-DLLVM_TARGETS_TO_BUILD="host" \
+		-DLLVM_TARGETS_TO_BUILD="host;X86;AArch64;AMDGPU;NVPTX;WebAssembly" \
```
mlxd: To be removed before merging

.clang-tidy
```diff
@@ -0,0 +1,61 @@
+---
```
mlxd: Useful for IDE. Can be removed before merging.

.clangd
```diff
@@ -0,0 +1,43 @@
+CompileFlags:
```
mlxd: Useful for IDE. Can be removed before merging.

mlir/include/Driver/CatalystLLVMTarget.h
```diff
@@ -20,7 +20,7 @@
 #include "mlir/IR/DialectRegistry.h"
 #include "mlir/Support/LogicalResult.h"
 
-#include "CompilerDriver.h"
+#include "CompilerDriver.hpp"
```
dime10: Not sure how you want to handle this, but just letting you know the convention we use here is from LLVM where all the headers are `.h`. We rarely use `.hpp` except in few headers that colocated with source files.

mlir/include/Driver/CompilerDriver.hpp
```diff
@@ -0,0 +1,354 @@
+// Copyright 2023 Xanadu Quantum Technologies Inc.
```
dime10, line 1: Not sure if we still follow this convention?
```suggestion
// Copyright 2023-2026 Xanadu Quantum Technologies Inc.
```

dime10: Thank you for adding all the docstrings btw!

mlir/include/Driver/CompilerDriver.h
```diff
@@ -0,0 +1,354 @@
...
+                                      mlir::DialectRegistry &registry);
+
+/**
+ * @brief Entry point to the MLIR portion of the compiler using the command-line interface.
+ *
+ * @param argc `main` function argc
+ * @param argv `main` function argv
+ * @return int Return code of execution.
+ */
+int QuantumDriverMainFromCL(int argc, char **argv);
```
dime10: conventional final newline missing (recommend setting this by default in your editor formatting options)
```suggestion
int QuantumDriverMainFromCL(int argc, char **argv);

```

mlir/include/Driver/PassInstrumentation.h
```diff
@@ -0,0 +1,28 @@
...
+    void dumpIRAfterPass(mlir::Pass *pass, mlir::Operation *op);
+
+    const driver::CompilerOptions &options;
+    driver::CompilerOutput &output;
+    catalyst::utils::Timer<> &timer;
+    // Store fingerprints before each pass to detect changes
+    llvm::DenseMap<mlir::Pass *, std::optional<mlir::OperationFingerPrint>> beforePassFingerprints;
+};
+
+} // namespace catalyst
```
dime10:
```suggestion
} // namespace catalyst

```

mlir/include/Driver/PassRegistry.hpp
dime10, line 1: This header is empty I think?

mlir/tools/catalyst-cli/CMakeLists.txt
```diff
@@ -55,4 +55,4 @@ target_link_libraries(catalyst-cli PRIVATE ${LIBS})
 llvm_update_compile_flags(catalyst-cli)
 mlir_check_all_link_libraries(catalyst-cli)
 export_executable_symbols_for_plugins(catalyst-cli)
-set_target_properties(catalyst-cli PROPERTIES OUTPUT_NAME catalyst)
+set_target_properties(catalyst-cli PROPERTIES OUTPUT_NAME catalyst)
```
dime10:
```suggestion
set_target_properties(catalyst-cli PROPERTIES OUTPUT_NAME catalyst)

```

mlir/lib/Driver/PassInstrumentation.cpp
```diff
@@ -0,0 +1,105 @@
...
+        s << *op;
+    }
+    std::string fileName = pipelineName.str();
+    if (auto funcOp = dyn_cast<mlir::func::FuncOp>(op)) {
+        fileName += std::string("_") + funcOp.getName().str();
+    }
+    dumpToFile(this->options, this->output.nextPassDumpFilename(fileName), tmp);
+}
+
+} // namespace catalyst
```
dime10:
```suggestion
} // namespace catalyst

```

mlir/lib/Driver/CMakeLists.txt
```diff
@@ -53,11 +53,13 @@ set(LIBS
...
 add_mlir_library(CatalystCompilerDriver
+    Main.cpp
     CompilerDriver.cpp
     CatalystLLVMTarget.cpp
+    PassInstrumentation.cpp
     Pipelines.cpp
+    LineUtils.cpp
     HighResolutionOutputStrategy.cpp
-
     LINK_LIBS PRIVATE
```
dime10, line 66:
```suggestion

    LINK_LIBS PRIVATE
```

mlir/lib/Driver/Main.cpp
```diff
@@ -0,0 +1,393 @@
+#include "Catalyst/Transforms/BufferizableOpInterfaceImpl.h"
```
dime10: Missing copyright header

mlir/lib/Driver/Main.cpp
```diff
@@ -0,0 +1,393 @@
...
+#include "stablehlo/integrations/c/StablehloPasses.h"
+
+#include "RegisterAllPasses.h"
+
+#include "Driver/CatalystLLVMTarget.h"
+#include "Driver/CompilerDriver.hpp"
+#include "Driver/HighResolutionOutputStrategy.h"
+#include "Driver/LineUtils.hpp"
+#include "Driver/Timer.hpp"
+#include "stablehlo/transforms/optimization/Passes.h"
```
dime10: Headers aren't properly collected (like with like), sorted (stdlib, 3rd party, local), and inconsistently use `<>` vs `""` (I think we only use `<>` for stdlib, so best to follow that).

mlir/include/Catalyst/Transforms/BufferizableOpInterfaceImpl.h
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #pragma once
+#include "mlir/IR/DialectRegistry.h"
```
maliasadi, line 16: why is this header for registring/extending dialects needed here?

mlir/include/Driver/PassRegistry.hpp
maliasadi, line 1: This file can be removed

mlir/lib/Driver/Main.cpp
```diff
@@ -0,0 +1,393 @@
...
+
+namespace {
+using namespace catalyst;
+using namespace catalyst::driver;
+} // namespace
```
dime10: Hadn't see this form before so I looked it up. Apparently the anonymous namespace has [no effect around a using directive](https://stackoverflow.com/a/10417103), so I'd recommend removing for consistency/simplicity
```suggestion
using namespace catalyst;
using namespace catalyst::driver;
```

mlir/lib/Driver/Main.cpp
```diff
@@ -0,0 +1,393 @@
...
+
+    if (mlir::failed(result)) {
+        llvm::errs() << "Compilation failed:\n" << output->diagnosticMessages << "\n";
+        return llvm::to_underlying(ErrorCode::Failure);
+    }
+
+    if (Verbose)
+        llvm::outs() << "Compilation successful:\n" << output->diagnosticMessages << "\n";
+    return llvm::to_underlying(ErrorCode::Success);
+}
```
dime10:
```suggestion
}

```

mlir/include/Driver/CompilerDriver.h
```diff
@@ -0,0 +1,354 @@
...
+
+#include "Driver/Pipelines.h"
+#include "Driver/Timer.hpp"
+
+namespace catalyst::driver {
+
+/// Verbosity level
+// TODO: Adjust the number of levels according to our needs. MLIR seems to print few really
+// low-level messages, we might want to hide these.
+enum class Verbosity { Silent = 0, Urgent = 1, Debug = 2, All = 3 };
```
maliasadi, line 47: Could you add docstrings for `Verbosity` similar to the other enum classes?

```suggestion
enum class Verbosity : int { Silent = 0, Urgent = 1, Debug = 2, All = 3 };
```

mlir/include/Driver/CompilerDriver.h
```diff
@@ -0,0 +1,354 @@
...
+
+/**
+ * @brief Controls the stage for dumping the IR.
+ *
+ */
+enum SaveTemps { None, AfterPipeline, AfterPassChanged, AfterPass };
```
maliasadi, line 53:
```suggestion
enum class SaveTemps : int { None = 0, AfterPipeline, AfterPassChanged, AfterPass };
```

mlir/include/Driver/CompilerDriver.h
```diff
@@ -0,0 +1,354 @@
...
+
+/**
+ * @brief Defines the different functional stages of the driver.
+ *
+ */
+enum Action { OPT, Translate, LLC, All };
```
maliasadi, line 59:
```suggestion
enum class Action : int { OPT = 0, Translate, LLC, All };
```

mlir/include/Driver/CompilerDriver.h
```diff
@@ -0,0 +1,354 @@
...
+
+/**
+ * @brief Defines the format of the input data.
+ *
+ */
+enum InputType { MLIR, LLVMIR, OTHER };
```
maliasadi, line 65:
```suggestion
enum class InputType : int { MLIR = 0, LLVMIR, OTHER };
```

mlir/include/Driver/CompilerDriver.h
```diff
@@ -0,0 +1,354 @@
...
+ * @param clHasManualPipeline
+ * @param moduleOp
+ * @return llvm::LogicalResult
+ */
+llvm::LogicalResult runPipeline(mlir::PassManager &pm, const CompilerOptions &options,
+                                CompilerOutput &output, Pipeline &pipeline,
+                                bool clHasManualPipeline, mlir::ModuleOp moduleOp);
+
+/**
+ * @brief
```
maliasadi: TODO

mlir/include/Driver/CompilerDriver.h
```diff
@@ -0,0 +1,354 @@
...
+ * @param options Compiler configuration options.
+ * @param pipeline
+ * @param clHasManualPipeline
+ * @return llvm::LogicalResult
+ */
+llvm::LogicalResult configurePipeline(mlir::PassManager &pm, const CompilerOptions &options,
+                                      Pipeline &pipeline, bool clHasManualPipeline);
+
+/**
+ * @brief
```
maliasadi: TODO

mlir/include/Driver/CompilerDriver.h
```diff
@@ -0,0 +1,354 @@
...
+ * @param timer
+ * @param timing
+ * @return llvm::LogicalResult
+ */
+llvm::LogicalResult preparePassManager(mlir::PassManager &pm, const CompilerOptions &options,
+                                       CompilerOutput &output, catalyst::utils::Timer<> &timer,
+                                       mlir::TimingScope &timing);
+
+/**
+ * @brief
```
maliasadi: Here and in other places...

mlir/include/Driver/CompilerDriver.h
maliasadi:
```suggestion
int QuantumDriverMainFromCL(int argc, char **argv);

```

mlir/include/Driver/HighResolutionOutputStrategy.h
```diff
@@ -17,17 +17,15 @@
 #include "mlir/Support/Timing.h"
 #include "llvm/Support/raw_ostream.h"
 
-using namespace mlir;
```
maliasadi, line 20: nice! no namespace-wiide directives in headers 👍

mlir/lib/Driver/CompilerDriver.cpp
```diff
@@ -83,129 +80,14 @@
...
 
 #include "Enzyme.h"
-#include "Timer.hpp"
+
+namespace {
 
 using namespace mlir;
 using namespace catalyst;
 using namespace catalyst::driver;
 namespace cl = llvm::cl;
```
dime10, line 95:
```suggestion

using namespace mlir;
using namespace catalyst;
using namespace catalyst::driver;
namespace cl = llvm::cl;

namespace {
```

mlir/include/Driver/Timer.hpp
maliasadi: could this benefit from the [MLIR Timer](https://mlir.llvm.org/doxygen/classmlir_1_1Timer.html) class. If we can remove this custom impl or simplify it.

dime10 (approved):
Thank you for the nice refactor! Just some cleanup

mlir/lib/Driver/Main.cpp
maliasadi, line 1: So, `CompilerDriver.cpp` is now decomposed into this file (entry points) + `CompilerDriver.cpp` with pass functions, `PassInstrumentation.cpp` with IR dumping functions, and `LineUtils.cpp` with diagnostics/instrumentation boilerplate.

.clang-tidy
maliasadi: Why shouldn't we keep this? It's useful but it seems a bit redundant with the one in `./runtime`.

mlir/include/Driver/CompilerDriver.h
maliasadi, line 1: Could we rearrange the functions in this file to better show th compilation stages:

1. Entrypoints (`QuantumDriverMain`, `QuantumDriverMainFromCL`) and Helper functions
2. Parsing (MLIR, LLVM, parsePipeline, readInputFile and erifyInputType, ...)
3. Registration (register*)
4. Pass Management
5. Lowering

mlir/include/Driver/Timer.h
```diff
@@ -13,50 +13,89 @@
 // limitations under the License.
 #pragma once
 
+#include <bits/types/struct_timeval.h>
```
maliasadi: Could we remove this? It's an internal glibc header and non-portable. https://stackoverflow.com/questions/43255796/redefinition-of-struct-timeval-in-several-headers.

mlir/lib/Driver/PassInstrumentation.cpp
maliasadi:
```suggestion
} // namespace catalyst

```

mlir/lib/Driver/PassInstrumentation.cpp
```diff
@@ -0,0 +1,105 @@
+#include "mlir/Dialect/Func/IR/FuncOps.h"
```
maliasadi, line 15: Missing copyright. Could you double-check headers of all the new files?

mlir/include/Driver/Timer.h
```diff
@@ -13,50 +13,89 @@
...
     bool debug_timer;
 
     // Manage the call order of `start` and `stop` methods
     bool running;
 
-    // Start and stop time points using steady_clock
-    std::chrono::time_point<std::chrono::steady_clock> start_wall_time_;
-    std::chrono::time_point<std::chrono::steady_clock> stop_wall_time_;
+    // Start and stop time points using high_resolution_clock
+    std::chrono::time_point<std::chrono::high_resolution_clock> start_wall_time_;
```
maliasadi, line 97: On modern Linux/macOS machines `high_resolution_clock == steady_clock`; this change is just for clarity, right?

mlir/lib/Driver/CompilerDriver.cpp
```diff
@@ -455,13 +237,14 @@ LogicalResult runCoroLLVMPasses(const CompilerOptions &options,
     return success();
 }
 
-LogicalResult runO2LLVMPasses(const CompilerOptions &options,
-                              std::shared_ptr<llvm::Module> llvmModule, CompilerOutput &output)
+llvm::LogicalResult catalyst::driver::runO2LLVMPasses(const CompilerOptions &options,
```
maliasadi, line 250: The implementation of this function has similarities with `runEnzymePasses` and `runCoroLLVMPasses`. Could we use a templated `runXPasses` to avoid the identical boilerplate in these functions?

mlir/lib/Driver/Main.cpp
```diff
@@ -0,0 +1,393 @@
...
+#include "Driver/LineUtils.hpp"
+#include "Driver/Timer.hpp"
+#include "stablehlo/transforms/optimization/Passes.h"
+
+namespace {
+using namespace catalyst;
+using namespace catalyst::driver;
+} // namespace
+
+llvm::LogicalResult QuantumDriverMain(const CompilerOptions &options, CompilerOutput &output,
```
maliasadi, line 57: What do you think about introducing a compilation context struct (similar to the context manager in the runtime.. I had this idea in the ADR I worked on in Q3) to avoid one monolithic func handling everything?

mlir/lib/Driver/CompilerDriver.cpp
```diff
@@ -577,9 +361,11 @@ std::string readInputFile(const std::string &filename)
     return buffer.str();
 }
 
-LogicalResult preparePassManager(PassManager &pm, const CompilerOptions &options,
-                                 CompilerOutput &output, catalyst::utils::Timer &timer,
-                                 TimingScope &timing)
+llvm::LogicalResult catalyst::driver::preparePassManager(PassManager &pm,
```
maliasadi: Could we add `[[nodiscard]]` to all these functions here? 

```suggestion
[[nodiscard]] llvm::LogicalResult catalyst::driver::preparePassManager(PassManager &pm,
```

mlir/lib/Driver/CompilerDriver.cpp
```diff
@@ -594,8 +380,10 @@ LogicalResult preparePassManager(PassManager &pm, const CompilerOptions &options
     return success();
 }
 
-LogicalResult configurePipeline(PassManager &pm, const CompilerOptions &options, Pipeline &pipeline,
-                                bool clHasManualPipeline)
+llvm::LogicalResult catalyst::driver::configurePipeline(PassManager &pm,
```
maliasadi:
```suggestion
[[nodiscard]] llvm::LogicalResult catalyst::driver::configurePipeline(PassManager &pm,
```

mlir/lib/Driver/CompilerDriver.cpp
```diff
@@ -639,14 +429,15 @@ LogicalResult runPipeline(PassManager &pm, const CompilerOptions &options, Compi
     return success();
 }
 
-LogicalResult runLowering(const CompilerOptions &options, MLIRContext *ctx, ModuleOp moduleOp,
-                          CompilerOutput &output, TimingScope &timing)
+llvm::LogicalResult catalyst::driver::runLowering(const CompilerOptions &options, MLIRContext *ctx,
```
maliasadi:
```suggestion
[[nodiscard]] llvm::LogicalResult catalyst::driver::runLowering(const CompilerOptions &options, MLIRContext *ctx,
```

mlir/lib/Driver/PassInstrumentation.cpp
```diff
@@ -0,0 +1,105 @@
...
+                                                     : op->getParentOfType<mlir::ModuleOp>();
+        s << mod;
+    }
+    else {
+        s << *op;
+    }
+    std::string fileName = pipelineName.str();
+    if (auto funcOp = dyn_cast<mlir::func::FuncOp>(op)) {
+        fileName += std::string("_") + funcOp.getName().str();
+    }
```
maliasadi:
```suggestion
    std::string fileName;
    llvm::raw_string_ostream os(fileName);
    os << pipelineName;
    if (auto funcOp = dyn_cast<mlir::func::FuncOp>(op)) {
        os << '_' << funcOp.getName();
    }
```

maliasadi (commented):
Nice work @mlxd! ⛱️

.clang-tidy
mlxd: The runtime one only applies to the runtime. This applies everywhere. Happy to remove it for now and revisit later.

mlir/include/Driver/CatalystLLVMTarget.h
mlxd: Just getting back to this. I think it makes more sense to define files with templated contents with .hpp, but since LLVM convention is .h for everything will relabel to the preferred.

mlir/include/Driver/CompilerDriver.h
mlxd: Will add doscstring. We don't need to state as int.

mlir/include/Driver/CompilerDriver.h
mlxd: Not needed yet. Will ignore.

mlir/include/Driver/CompilerDriver.h
mlxd: as above

mlir/include/Driver/CompilerDriver.h
mlxd: As above

mlir/include/Driver/Timer.h
mlxd: Good call -- will replace with `#include <sys/time.h>`
If it isn't working on MacOS, I'll add comp defs.

mlir/include/Driver/Timer.h
mlxd: It is system dependent, and not always guaranteed to be steady_clock.

mlir/lib/Driver/CompilerDriver.cpp
mlxd: Potentially in a future follow up.

mlir/lib/Driver/PassInstrumentation.cpp
mlxd: Good catch

mlir/include/Catalyst/Transforms/BufferizableOpInterfaceImpl.h
mlxd, line 16: The downstream use of the .o required linkage against the impl from that file.

mlir/include/Driver/CompilerDriver.h
mlxd: We can revisit this in future and do so then.

mlir/lib/Driver/Main.cpp
mlxd: A little confused as to why the formatter isn't configured to allow this automatically.
I'll rearrange the imports, but would be good if we can have automated ordering with a custom https://clang.llvm.org/docs/ClangFormatStyleOptions.html#includecategories --- I think we may have this in the dev guidelines already.

mlxd: Revised the imports. Tried an automated strategy and it got most of the way, so should be able to fix this with a little effort.

mlir/include/Driver/CatalystLLVMTarget.h
mlxd: Relabelled all files to .h

mlir/include/Driver/CompilerDriver.hpp
mlxd: >Not sure if we still follow this convention?

Convention is to not update the dates and use the original entries.

mlir/include/Driver/CompilerDriver.h
mlxd: Resolved.

mlxd: Also, this should be part of the formatter config if it is not already.

mlir/include/Driver/PassRegistry.hpp
mlxd: Ooops. Removed it.

mlir/lib/Driver/Main.cpp
mlxd: Let's revisit this later.

mlir/include/Driver/Timer.h
```diff
@@ -14,49 +14,88 @@
...
+template <IsRatio Duration = std::milli> class CPUTimeInstance {
+  public:
+    CPUTimeInstance() : t{} {}
+
+    /**
+     * @brief Return an fixed integer type casting of time values at the chosen precision.
+     *
+     * @return int64_t
+     */
+    int64_t getTime()
```
maliasadi: Casting to double (line 66) but returning `int64_t`:

```suggestion
    double getTime()
```

mlir/include/Driver/Timer.h
```diff
@@ -14,49 +14,88 @@
...
+    {
+        if (clock_gettime(CLOCK_PROCESS_CPUTIME_ID, &t) == 0) {
+            return static_cast<double>(t.tv_sec * Duration::den + t.tv_nsec);
+        }
+        return 0;
```
maliasadi:
```suggestion
        return 0.0;
```

mlir/lib/Driver/Pipelines.cpp
```diff
@@ -232,3 +232,19 @@ std::vector<Pipeline> getDefaultPipeline()
 
 } // namespace driver
 } // namespace catalyst
+
+namespace llvm {
+
+inline raw_ostream &operator<<(raw_ostream &oss, const catalyst::driver::Pipeline &p)
```
maliasadi: We can remove `inline` as it's now in the cpp file and has no effect.

```suggestion
raw_ostream &operator<<(raw_ostream &oss, const catalyst::driver::Pipeline &p)
```

mlir/lib/Driver/LineUtils.cpp
```diff
@@ -0,0 +1,175 @@
...
+{
+    const auto num_lines = std::count(opStrBuf.cbegin(), opStrBuf.cend(), '\n');
+    if (!name.empty()) {
+        std::cerr << "[DIAGNOSTICS] After " << std::setw(25) << std::left << name;
+    }
+    std::cerr << "\t" << std::fixed << "programsize: " << num_lines << std::fixed << " lines\n";
+}
+
+/**
+ * @brief
```
maliasadi: Could you update docstring of this function?

mlir/lib/Driver/PassInstrumentation.cpp
```diff
@@ -0,0 +1,121 @@
...
+CatalystPassInstrumentation::CatalystPassInstrumentation(const driver::CompilerOptions &options,
+                                                         driver::CompilerOutput &output,
+                                                         catalyst::utils::Timer<> &timer)
+    : options(options), output(output), timer(timer)
+{
+}
+
+void CatalystPassInstrumentation::runBeforePass(mlir::Pass *pass, mlir::Operation *operation)
+{
+    if (this->options.verbosity >= driver::Verbosity::Debug && !this->timer.is_active()) {
```
maliasadi, line 37: Could we remove these redundant  `this->` for the sake of consistency?

maliasadi (approved):
Thanks @mlxd! Happy to approve with just a few more comments, mostly minor! Could you also update changelog before merging the PR?

mlir/lib/Driver/Pipelines.cpp
mlxd: Having CT and WAE here would be useful to remind me of these as stuff is moved around. Cheers.

mlir/lib/Driver/PassInstrumentation.cpp
mlxd: Given the prevalence of these across the codebase, I'd rather we focus on fixing that in a follow-up later, since these appear everywhere.

PR #2455, opened 2026-01-29
Add hierarchical qubit types to xDSL Quantum dialect

mudit2812:
Replicate the changes to the MLIR Quantum dialect made in https://github.com/PennyLaneAI/catalyst/pull/2369 to the xDSL dialect.

* Add `QubitLevel` and `QubitRole` attributes
* Add `level` and `role` parameters to `QubitType`
* Add `level` attribute to `QuregType`

[sc-107540]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -327,7 +327,7 @@
   [(#2419)](https://github.com/PennyLaneAI/catalyst/pull/2419)
 
 * New qubit-type specializations have been added to Catalyst's MLIR and xDSL type system. These new qubit
-  types include `!quantum.bit<logical>`, `!quantum.bit<pbc>` and `!quantum.bit<physical>`. The
+  types include `!quantum.bit<logical>`, `!quantum.bit<qec>` and `!quantum.bit<physical>`. The
   original `!quantum.bit` type continues to be supported and used as the default qubit type.
   [(#2369)](https://github.com/PennyLaneAI/catalyst/pull/2369)
   [(#2455)](https://github.com/PennyLaneAI/catalyst/pull/2455)


frontend/catalyst/python_interface/dialects/quantum/attributes.py
```diff
@@ -45,40 +90,254 @@ class QubitType(ParametrizedAttribute, TypeAttribute):
...
+            delimiter=parser.Delimiter.ANGLE, parse=parser.parse_str_literal
+        )
+        optional_params = optional_params or []
+
+        final_params = []
+        match len(optional_params):
+            case 0:
+                final_params = [
+                    StringAttr(QubitLevel.Abstract.value),
+                    StringAttr(QubitRole.Null.value),
```
joeycarter, line 158: Just want to confirm, since this is a little different in MLIR. When you have a `StringAttr` in xDSL, is it printed with or without quotation marks? In other words, want we want to have is this, for example:

```mlir
!quantum.bit<physical, data>
```

and not this:

```mlir
!quantum.bit<"physical", "data">
```

frontend/catalyst/python_interface/dialects/quantum/attributes.py
```diff
@@ -45,40 +90,254 @@ class QubitType(ParametrizedAttribute, TypeAttribute):
...
+            case 2:
+                if optional_params[0] not in QubitLevel.__members__.values():
+                    raise ParseError(f"Invalid value for 'QubitType.level': {optional_params[0]}.")
+                if optional_params[1] not in QubitRole.__members__.values():
+                    raise ParseError(f"Invalid value for 'QubitType.role': {optional_params[1]}.")
+                final_params = optional_params
+
+            case _:
+                raise ParseError(
+                    f"Expected 2 or less parameters for 'QubitType', got {optional_params}."
```
joeycarter: [Grammar]

```suggestion
                    f"Expected 2 or fewer parameters for 'QubitType', got {optional_params}."
```

Could also say, "Expected at most two parameters for..."

frontend/catalyst/python_interface/dialects/quantum/attributes.py
```diff
@@ -45,40 +90,254 @@ class QubitType(ParametrizedAttribute, TypeAttribute):
...
+    @classmethod
+    def parse_parameters(cls, parser: AttrParser) -> list[Attribute]:
+        optional_params = parser.parse_optional_comma_separated_list(
+            delimiter=parser.Delimiter.ANGLE, parse=parser.parse_str_literal
+        )
+
+        optional_params = optional_params or []
+        if len(optional_params) > 1:
+            raise ParseError(
+                f"Expected 1 or less parameters for 'QuregType', got {optional_params}."
```
joeycarter: As above.

frontend/catalyst/python_interface/dialects/quantum/attributes.py
mudit2812: Since I implemented custom printers and parsers, I had a bit more control. Generally, string attrs print with the quotes, but with the custom printing this is what we get:
```pycon
>>> from catalyst.python_interface.dialects import quantum
>>> print(quantum.QubitType(level="logical", role="xcheck"))
!quantum.bit<logical, xcheck>
```

For what it's worth, the first implementation of the custom printers did print the double quotation marks, but that didn't cause any issues with conversion to and from MLIR in my local testing.

mudit2812: The xDSL printer has a `print_string` method which doesn't print the quotes, and also a `print_string_literal` which does print the quotes.

joeycarter: Nice, thanks!

frontend/catalyst/python_interface/dialects/quantum/attributes.py
```diff
@@ -14,22 +14,68 @@
...
 ################################################################
+# pylint: disable=no-member
+
+
+class QubitLevel(StrEnum):
+    """Qubit levels enum in the hierarchical qubit representation."""
+
+    Abstract = "abstract"
+    Logical = "logical"
+    PBC = "pbc"
```
joeycarter: Why is this `pbc`? It should be `qec`.

joeycarter: Aha, it was changed in https://github.com/PennyLaneAI/catalyst/pull/2482. We'll have to change this back (cc @sengthai)

joeycarter: Fixed in #2499 (thanks @sengthai!)

mudit2812: Fixed

doc/releases/changelog-dev.md
```diff
@@ -326,18 +326,19 @@
   global phase correctness.
   [(#2419)](https://github.com/PennyLaneAI/catalyst/pull/2419)
 
-* New qubit-type specializations have been added to Catalyst's MLIR type system. These new qubit
+* New qubit-type specializations have been added to Catalyst's MLIR and xDSL type system. These new qubit
   types include `!quantum.bit<logical>`, `!quantum.bit<pbc>` and `!quantum.bit<physical>`. The
```
joeycarter: Looks like we missed the changelog when we reverted the qec -> qec change. Mind fixing this too while we're here?

```suggestion
  types include `!quantum.bit<logical>`, `!quantum.bit<qec>` and `!quantum.bit<physical>`. The
```

frontend/catalyst/python_interface/xdsl_extras/traits.py
```diff
@@ -232,3 +232,28 @@ def verify(self, op: Operation) -> None:
                 f"all of {{{names_str}}} must have the same {self.summary}: got "
                 f"{self.summary}s {results_str}"
             )
+
+
+@dataclass(frozen=True)
+class AllTypesMatch(OpTrait):
```
mehrdad2m, line 238: I think this can be a useful trait to be upstreamed. I can see some uses in stableHLO too but we haven't added those operations yet.

frontend/catalyst/python_interface/xdsl_extras/traits.py
```diff
@@ -232,3 +232,28 @@ def verify(self, op: Operation) -> None:
...
+
+    def verify(self, op: Operation) -> None:
+        """Verify that the operation attributes all share the same types."""
+        for attr in self.attr_names:
+            value = getattr(op, attr, None)
+            if not isinstance(value, (SSAValue, SSAValues)):
+                return
+
+        AllMatchSameOperatorTrait(
+            attr_names=self.attr_names, operator=self.same_ssa_types_operator, summary=self.summary
```
mehrdad2m, line 258: I don't think we would need a variable `summary` here since we are specializing `AllMatchSameOperatorTrait` for `type`. This would be consistent with [upstream implementation](https://github.com/llvm/llvm-project/blob/8f264586d7521b0e305ca7bb78825aa3382ffef7/mlir/include/mlir/IR/OpBase.td#L556-L557). 
```suggestion
            attr_names=self.attr_names, operator=self.same_ssa_types_operator, summary="type"
```

mehrdad2m: Note that `AllMatchSameOperatorTrait` already handles messages like `Qubit ins and outs must have the same size and types` see https://github.com/PennyLaneAI/catalyst/blob/72c9d182a774f1856f0527fda2e5e6ebe40031a7/frontend/catalyst/python_interface/xdsl_extras/traits.py#L228-L234
which basically expects summary to be a type. Maybe that naming is misleading. So cases like:
```
AllTypesMatch(
    ("in_qubits", "out_qubits"),
    "Qubit ins and outs must have the same size and types",
)
```
shouldn't produce a good error message.

mudit2812:
> Looks good, thanks @mudit2812!
> 
> Any thoughts why CI is failing?

Not sure. I can take a look next week

frontend/catalyst/python_interface/dialects/quantum/attributes.py
```diff
@@ -45,40 +91,298 @@ class QubitType(ParametrizedAttribute, TypeAttribute):
...
+            for s in level_constr:
+                if isinstance(s, str):
+                    assert s in QubitLevel.__members__.values()
+                    attr_list.append(StringAttr(s))
+                elif isinstance(s, StringAttr):
+                    assert s.value in QubitLevel.__members__.values()
+                    attr_list.append(s)
+                else:
+                    raise ValueError(f"Invalid value for 'QubitType.level' constraint: {s}.")
+            self.level_constr = AnyOf(attr_list)
```
mehrdad2m, line 289: This is a suggestion and non-blocking, but I feel like this code can be refactored into a function to make the `__init__` implementation a bit simpler. Especially that the same logic is repeated for `QubitRole`

frontend/catalyst/python_interface/dialects/quantum/attributes.py
```diff
@@ -45,40 +91,298 @@ class QubitType(ParametrizedAttribute, TypeAttribute):
...
+        level = self.level.data
+        role = self.role.data
+
+        if level not in QubitLevel:
+            raise VerifyException(
+                f"Invalid value {level} for 'QubitType.level'. Allowed values "
+                f"are {list(QubitLevel.__members__.values())}."
+            )
+
+        if role not in QubitRole:
```
mehrdad2m, line 119:
```suggestion
        if level not in QubitLevel.__members__.values():
            raise VerifyException(
                f"Invalid value {level} for 'QubitType.level'. Allowed values "
                f"are {list(QubitLevel.__members__.values())}."
            )

        if role not in QubitRole.__members__.values():
```
I think this should get rid of the CI errors. This is only a problem with python 3.11. see https://docs.python.org/3/library/enum.html#enum.EnumType.__contains__

mehrdad2m (commented):
Thanks @mudit2812  🚀 Happy to approve after comments are addressed and CI is green

PR #2456, opened 2026-01-29
Raise an error if shots=0 when using measurements_from_samples pass

lillian542:
**Context:**
The `measurements_from_samples` pass returns `nan` on `lightning` if there are no shots provided, and raises a cryptic RuntimeError on `null.qubit`. In neither case is it obvious that the issue is related to shots or to the pass. This  has happened several times to various people, and takes a while to debug.

After we confirm that `shots` is an integer, we raise an error if it's 0.

**Benefits:**
Looking at the error traceback makes it immediately obvious what happened, so you don't spend 30+ minutes trying to debug things, just to realize you forgot to `set_shots` 🤦‍♀️

[sc-100896]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -20,6 +20,10 @@
 * `qml.vjp` can now be used with Catalyst and program capture.
   [(#2279)](https://github.com/PennyLaneAI/catalyst/pull/2279)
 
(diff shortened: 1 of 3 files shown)


paul0403 (approved):
Good idea!

joeycarter (approved):
💯

PR #2457, opened 2026-01-29
[Feature] Add `capture` argument to `@qjit` for local program capture control

JerryChen97:
## Summary

This PR introduces a `capture` keyword argument to the `@qjit` decorator, enabling per-function control over PennyLane's program capture frontend without affecting the global `qml.capture.enabled()` state.

## Motivation

PennyLane's new program capture system (`qml.capture`) is the future of quantum program representation, but enabling it globally via `qml.capture.enable()` switches **all** of PennyLane to an unmaintained experimental execution pathway. This creates a dilemma for Catalyst users who want to test the new capture-based compilation frontend:

- **Without this PR**: Users must call `qml.capture.enable()` globally, which affects all PennyLane code in their application, potentially breaking non-Catalyst workflows
- **With this PR**: Users can opt-in to capture on a per-`@qjit` basis, keeping the rest of their code on the stable legacy pathway

## API Design

The `capture` parameter accepts three values:

```python
@qjit(capture="global")  # Default: defer to qml.capture.enabled(); equivalently, one can think of this as a `None` value
@qjit(capture=True)      # Force capture ON for this function
@qjit(capture=False)     # Force capture OFF for this function
```

#### 1. **Why `"global"` as the default (not `True` or `False`)?**

The default behavior should preserve backwards compatibility. Users who haven't called `qml.capture.enable()` shouldn't suddenly get the experimental capture pathway. By defaulting to `"global"`, we:
- Maintain existing behavior for all current Catalyst code
- Allow gradual migration as the capture system matures
- Respect the user's explicit global choice when `qml.capture.enable()` is called

#### 2. **Why tri-state (`"global"` / `True` / `False`) instead of just boolean?**

A boolean with a default of `False` would force users to always specify `capture=True` when using capture, even if they've enabled it globally. The `"global"` option provides ergonomic flexibility:
- Legacy code: `@qjit` → uses old frontend (current behavior)
- Global adopters: `qml.capture.enable()` + `@qjit` → all functions use capture
- Selective adopters: `@qjit(capture=True)` → single function uses capture without global enable

#### 3. **Why `capture` instead of `use_capture` or `enable_capture`?**

Shorter is better for frequently-used APIs. The parameter name mirrors PennyLane's `qml.capture` module name, making the connection obvious. The tri-state nature (`"global"` / `True` / `False`) makes it clear it's a mode selector, not just a boolean flag.

### Scope Enforcement Pattern

The implementation uses a **"scope enforcement"** pattern rather than **"state management"**:

```python
# Snapshot, enforce, restore from snapshot
with ensure_capture_mode(target_mode):
    compile()  # Mode is guaranteed for this scope
# Automatically restored, even if compilation crashes
```

### Key Components

1. **`ensure_capture_mode()` context manager** ([contexts.py](frontend/catalyst/tracing/contexts.py))
   - Takes a boolean: the desired capture state for the scope
   - Snapshots the pre-existing state
   - Enforces the target state (only toggles if necessary)
   - **Always restores from the snapshot** (not from re-checking current state)
   - Protects against nested toggles drifting out of sync

2. **`_get_effective_capture_mode()` method** ([jit.py](frontend/catalyst/jit.py))
   - Resolves the effective boolean mode from the tri-state `capture` option
   - `"global"` → returns current `qml.capture.enabled()` value
   - `True` / `False` → returns that explicit boolean
   - Called once at key decision points (init, compilation)

3. **`capture` field in `CompileOptions`** ([pipelines.py](frontend/catalyst/pipelines.py))
   - Added to the dataclass with default `"global"`
   - Validated in `__post_init__` (only allows `"global"`, `True`, `False`)
   - Becomes part of the QJIT object's "identity card"
   - Prevents cache collisions between capture modes

### Critical: Cache Safety

Since `capture` fundamentally changes the compilation pipeline (different JAXPR lowering, different MLIR dialects), it **must** be part of `CompileOptions`. This ensures:
- Two functions with identical code but different `capture` settings are treated as distinct compilations
- No risk of accidentally running `capture=True` MLIR with `capture=False` expectations (or vice versa)

## Testing

Comprehensive test suite in [test_capture_kwarg.py](frontend/test/pytest/test_capture_kwarg.py):

1. **Basic functionality**: Validate the tri-state works as documented
2. **Context manager correctness**: Ensure `ensure_capture_mode()` is re-entrant and exception-safe
3. **Integration tests**: Verify quantum circuits compile and run correctly in both modes
4. **Cache identity tests**: Confirm different capture modes produce distinct `CompileOptions` objects
5. **AOT preservation tests**: Ensure the flag persists through ahead-of-time compilation

### Selective testing of capture without global side effects:

```python
import pennylane as qml
from catalyst import qjit

# Global state remains on legacy pathway
assert not qml.capture.enabled()

@qjit(capture=True)
def test_new_frontend(theta):
    qml.RX(theta, wires=0)

# Compiles with capture, but doesn't affect global state
result = test_new_frontend(0.5)
assert not qml.capture.enabled()  # Still False!
```

### Mixing legacy and capture in the same codebase:

```python
qml.capture.enable()  # Some code needs capture globally

@qjit(capture=False)  # But this function needs the old frontend
@qml.qnode(dev)
def legacy_circuit(x):
    # Uses old QNode patching, even though capture is globally enabled

@qjit  # capture="global" by default, so uses capture
@qml.qnode(dev)
def modern_circuit(x):
    # Uses new capture-based frontend
```

## Related Issues

- Internal ticket: [sc-109341]
- Addresses the need for Catalyst to safely experiment with capture without breaking user code


frontend/catalyst/tracing/contexts.py
```diff
@@ -120,6 +121,56 @@ def am_inside_accelerate():
         return AccelerateContext._am_inside_accelerate > 0
 
 
+@contextmanager
```
albi3ro: Given that this is not a tracing context and only used in `jit.py`, could we just leave it as a private detail of `jit.py`?

albi3ro (commented):
Just one comment about moving the location of the context manager, but other than that looks good 👍

frontend/catalyst/pipelines.py
```diff
@@ -137,6 +142,7 @@ class CompileOptions:
     circuit_transform_pipeline: Optional[dict[str, dict[str, str]]] = None
     pass_plugins: Optional[Set[Path]] = None
     dialect_plugins: Optional[Set[Path]] = None
+    capture: Optional[Union[str, bool]] = "global"
```
albi3ro:
```suggestion
    capture: bool | Literal['global'] = "global"
```
`None` is actually not a valid option. This is how we want to type hint this now.

frontend/catalyst/tracing/contexts.py
JerryChen97: Now it became a private helper in `jit.py`! @albi3ro

frontend/test/pytest/test_capture_kwarg.py
```diff
@@ -0,0 +1,374 @@
...
+        """Test that capture=True uses the capture pathway without global enable."""
+        qml.capture.disable()  # Make sure global capture is off
+        assert not qml.capture.enabled()
+
+        dev = qml.device(backend, wires=2)
+
+        @qjit(capture=True)
+        @qml.qnode(dev)
+        def circuit(theta):
+            qml.RX(theta, wires=0)
```
albi3ro, line 163:
```suggestion
            assert qml.capture.enabled()
            qml.RX(theta, wires=0)
```

frontend/test/pytest/test_capture_kwarg.py
```diff
@@ -0,0 +1,374 @@
...
+    def test_capture_false_uses_old_pathway(self, backend):
+        """Test that capture=False uses old pathway even if global is on."""
+        qml.capture.enable()
+        try:
+            dev = qml.device(backend, wires=2)
+
+            @qjit(capture=False)
+            @qml.qnode(dev)
+            def circuit(theta):
+                qml.RX(theta, wires=0)
```
albi3ro, line 183:
```suggestion
                assert not qml.capture.enabled()
                qml.RX(theta, wires=0)
```

frontend/test/pytest/test_capture_kwarg.py
```diff
@@ -0,0 +1,374 @@
...
+        """Test that capture='global' respects the global setting."""
+        dev = qml.device(backend, wires=2)
+
+        # With global capture off
+        qml.capture.disable()
+
+        @qjit(capture="global")
+        @qml.qnode(dev)
+        def circuit_old_pathway(theta):
+            qml.RX(theta, wires=0)
```
albi3ro, line 203:
```suggestion
            assert not qml.capture.enabled()
            qml.RX(theta, wires=0)
```

frontend/test/pytest/test_capture_kwarg.py
```diff
@@ -0,0 +1,374 @@
...
+        result = circuit_old_pathway(0.5)
+        assert jnp.allclose(result, jnp.cos(0.5))
+
+    def test_capture_true_classical_only(self):
+        """Test that capture=True works for classical-only functions."""
+        qml.capture.disable()
+
+        @qjit(capture=True)
+        def classical_fn(x, y):
+            return x**2 + y**2
```
albi3ro, line 216:
```suggestion
            assert qml.capture.enabled()
            return x**2 + y**2
```

frontend/test/pytest/test_capture_kwarg.py
```diff
@@ -0,0 +1,374 @@
...
+
+    def test_multiple_qjits_different_capture_modes(self, backend):
+        """Test multiple QJIT functions with different capture modes."""
+        dev = qml.device(backend, wires=1)
+        qml.capture.disable()
+
+        @qjit(capture=True)
+        @qml.qnode(dev)
+        def circuit_capture(theta):
+            qml.RY(theta, wires=0)
```
albi3ro, line 230:
```suggestion
            assert qml.capture.enabled()
            qml.RY(theta, wires=0)
```

frontend/test/pytest/test_capture_kwarg.py
```diff
@@ -0,0 +1,374 @@
...
+            return qml.expval(qml.PauliZ(0))
+
+        @qjit(capture=False)
+        @qml.qnode(dev)
+        def circuit_old(theta):
+            qml.RY(theta, wires=0)
```
albi3ro, line 237:
```suggestion
            assert not qml.capture.enabled()
            qml.RY(theta, wires=0)
```

frontend/catalyst/jit.py
```diff
@@ -69,6 +72,49 @@
 jax.config.update("jax_enable_x64", True)
 
 
+@contextmanager
+def _ensure_capture_mode(enable_capture: bool):
```
dime10: Private helpers at the bottom please, a lot of Catalyst modules have been structured to show the most relevant code first, rather than having scroll down for it: API functions --> implementation methods and classes --> private helpers

frontend/catalyst/jit.py
```diff
@@ -91,6 +137,7 @@ def qjit(
     circuit_transform_pipeline=None,
     pass_plugins=None,
     dialect_plugins=None,
+    capture="global",
```
dime10, line 97: This doesn't have too much impact, but any reason not to use `None`? Just seems more standard and avoids having 2 different data types for the flag

dime10 (approved):

frontend/catalyst/jit.py
JerryChen97: I personally prefer `None` as well actually; might need @albi3ro 's input on the motivation behind the original design

frontend/catalyst/jit.py
JerryChen97: https://github.com/PennyLaneAI/catalyst/pull/2457/commits/62cdf0e6a02d701b26c7b915ff7069023a2ad7fb

frontend/catalyst/jit.py
albi3ro: We would still be having two different data types with `None`. I choose `"global"` since it was a lot more clear that we are using the global toggle.

albi3ro (approved):
Tests do actually seem a bit overkill, but at the same time I feel like I can't complain about that.

But thanks 👍 Looks good to me :) Glad to have this.

PR #2458, opened 2026-01-29
Implement one shot mcm as an mlir pass

paul0403:
**Context:**
The current way catalyst handles the one shot mcm method is to perform a tape-level transform at the python level.
This is problematic in two ways:
1. It is very messy and has many limitations
2. It is not compatible with capture

Implement one shot mcm as an mlir pass.

For the various MPs, here are their one-shot logic:
- [x] State: not supported, since it is only meaningful in analytical mode
- [x] Var: only supported on MCMs. The 1-shot var is always zero, so we need to calculate var from samples, but we do not have the ability to sample a named observable yet in catalyst
- [x] Expval and probs: add up results from the single shot executions and divide by the number of shots
- [x] Sample: Concatenate individual shot's samples together
- [x] Counts: Add the counts from each shot

When the returned MP is on MCMs, the one-shot kernel needs some massaging as well:
- [x] Expval: expval on a single-shot mcm is just that mcm itself
- [x] Probs: each shot returns either [1,0] or [0,1], depending on the mcm result
- [x] Sample: each shot returns 0 or 1, i.e. the mcm itself
- [x] Coutns: each shot returns [0,1] as eigvals, and [1,0] or [0,1] as the mcm result. However, counts on mcm in plxpr path is currently blocked by
```
NotImplementedError: CountsMP returns a dictionary, which is not compatible with capture
```
Resolved in https://github.com/PennyLaneAI/pennylane/pull/9022

**Benefits:**
No longer need to do the one shot transform in legacy frontend, so one-shot will be capture compatible.
As a standalone mlir pass, we can also simplify its logic significantly, since it no longer has to deal with the jax-generated army of unnecessary stablehlo ops from python.

**Possible Drawbacks:**
Since this is a new implementation, we might lose some feature from what is currently supported in legacy frontend's one-shot implementation. However, I'm not super concerned here, as the legacy frontend's one-shot implementation already feels quite random and clunky in what features it supports.

**Related GitHub Issues:**
[sc-109620]


mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
```diff
@@ -0,0 +1,326 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#define DEBUG_TYPE "one-shot-mcm"
+
+#include <algorithm>
```
paul0403: no need

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
```diff
@@ -0,0 +1,326 @@
...
+    void handleSampleOneShot(IRRewriter &builder, func::FuncOp qfunc, Value shots, Operation *mod)
+    {
+        // Total sample is the sample of each shot concatenated together.
+        // We inialize an empty tensor, then insert sample results of each shot
+        // Sample's return shape is tensor<shots X num_qubits>
+
+        OpBuilder::InsertionGuard guard(builder);
+        Location loc = mod->getLoc();
+
+        // Type f64Type = builder.getF64Type();
```
paul0403: no need

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
```diff
@@ -0,0 +1,381 @@
...
+            return WalkResult::interrupt();
+        });
+
+        if (illegalMP) {
+            return signalPassFailure();
+        }
+
+        size_t numReturns = isCounts ? 2 : 1;
+        assert(qfunc.getResultTypes().size() == numReturns &&
+               "Multiple terminal MPs not yet supported in one shot transform");
```
paul0403: I feel like multiple MPs in the return shouldn't be too hard to support, let me figure out something here

mlir/lib/Catalyst/Transforms/ApplyTransformSequencePass.cpp
```diff
@@ -19,6 +19,10 @@
 
 #include <cassert>
 
+#include "mlir/Dialect/Index/IR/IndexDialect.h"
```
rniczh: Is it required to update the files under `mlir/lib/Catalyst`?

mlir/lib/Catalyst/Transforms/CMakeLists.txt
```diff
@@ -30,6 +30,7 @@ get_property(conversion_libs GLOBAL PROPERTY MLIR_CONVERSION_LIBS)
 set(LIBS
     ${dialect_libs}
     ${conversion_libs}
+    ExternalStablehloLib
```
rniczh: Same as https://github.com/PennyLaneAI/catalyst/pull/2458/changes#r2755905009

mlir/include/Catalyst/Transforms/Passes.td
```diff
@@ -132,6 +132,13 @@ def MemrefToLLVMWithTBAAPass : Pass<"memref-to-llvm-tbaa", "mlir::ModuleOp"> {
 
 def ApplyTransformSequencePass : Pass<"apply-transform-sequence"> {
     let summary = "Apply the passes scheduled with the transform dialect.";
+
+    let dependentDialects = [
```
rniczh: Same as https://github.com/PennyLaneAI/catalyst/pull/2458/changes#r2755905009

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
```diff
@@ -0,0 +1,381 @@
...
+                if (!backwardSlice.contains(&*op)) {
+                    eraseWorklist.push_back(&*op);
+                }
+            }
+        }
+    }
+
+    for (Operation *op : eraseWorklist) {
+        op->erase();
+    }
```
rniczh: I think you can just `walk` the ops and erase them in the reverse order?
```cpp
SmallVector<Operation *> eraseWorklist;
qfunc.walk([&](Operation *op) {
    if (op != qfunc && !backwardSlice.contains(op)) {
        eraseWorklist.push_back(op);
    }
});

for (Operation *op : llvm::reverse(eraseWorklist)) {
    op->erase();
}
```

rniczh (commented):
Thanks @paul0403 🥳  This is a big milestone, we can finally get rid of the one-shot logic in the frontend :) Just some minor comments from my side

mlir/lib/Catalyst/Transforms/ApplyTransformSequencePass.cpp
paul0403: These are needed because the `--apply-transform-sequence` pass is the entry point to apply user-specified quantum passes (i.e. those having their own decorator, or using the `apply_pass` decorator from frontend). We never had quantum passes that creates ops, so it was never an issue.

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
paul0403: I actually chose a triple loop instead of a walk here, i.e. this was not an oversight!

Because the goal of this step is to erase all ops in the func except the shots `Value`, I don't actually need to recursively walk into the ops. For example, imagine the quantum kernel has a for loop itself. My erasure worklist here only needs to see the for op, not the ops inside the for op!

So I just loop over the top level and save some walk time.

mlir/lib/Catalyst/Transforms/ApplyTransformSequencePass.cpp
paul0403: Tested, it seems like they are unnecessary :thinking: Removed them in https://github.com/PennyLaneAI/catalyst/pull/2458/commits/06c089972225b1751710099b5c5df9b41940f48d, let's wait for CI

frontend/test/lit/test_mid_circuit_measurement.py
```diff
@@ -95,3 +103,176 @@ def circ():
...
+    """
+    Test that the mlir implementation of --one-shot-mcm pass can be used from frontend with expval
+    on a mid circuit measurement
+    """
+
+    # CHECK: transform.apply_registered_pass "one-shot-mcm"
+    qml.Hadamard(wires=0)
+    m = qml.measure(0)
+    return qml.expval(m)
+
```
paul0403: @albi3ro This is the status I'll leave you in after this PR, does it look like you would have all you need? 

i.e. on a circuit with mcm, we can use `@qml.transform(pass_name="one-shot-mcm")`, calling the new underlying mlir one shot pass (also implemented in this PR) and executing the resulting new mlir

albi3ro: Looks great :)

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
```diff
@@ -0,0 +1,434 @@
...
+    void editKernelMCMExpval(IRRewriter &builder, func::FuncOp qfunc, quantum::MCMObsOp mcmobs)
+    {
+        // If the kernel returns expval on a mcm,
+        // the single-shot expval of a mcm is just the mcm boolean result itself
+        // So we just cast it to the correct type and return it.
+
+        OpBuilder::InsertionGuard guard(builder);
+        Location loc = qfunc->getLoc();
+
+        // Erase all users of the mcm obs: they new return does not need them.
```
joeycarter: Typo?

```suggestion
        // Erase all users of the mcm obs: the new return does not need them.
```

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
```diff
@@ -0,0 +1,434 @@
...
+
+        // shots Value is I64, need to turn into tensor<f64> then broadcast for division
+        auto int2floatCastOp = builder.create<arith::SIToFPOp>(loc, f64Type, shots);
+        auto shotsFromElementsOp = builder.create<tensor::FromElementsOp>(
+            loc, RankedTensorType::get({}, f64Type), int2floatCastOp.getResult());
+
+        // TODO: no need to broadcast if it's expval
+        auto broadcastedShots = builder.create<stablehlo::BroadcastInDimOp>(
+            loc, expvalOrProbsType, shotsFromElementsOp.getResult(),
+            builder.getDenseI64ArrayAttr({}));
```
joeycarter: Were you planning to address this TODO in this PR?

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
```diff
@@ -0,0 +1,434 @@
...
+                }
+            }
+        }
+        return oneShotSampleShape;
+    }
+
+    void handleSampleOneShot(IRRewriter &builder, func::FuncOp qfunc, Value shots, Operation *mod)
+    {
+        // Total sample is the sample of each shot concatenated together.
+        // We inialize an empty tensor, then insert sample results of each shot
```
joeycarter:
```suggestion
        // We initialize an empty tensor, then insert sample results of each shot
```

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
```diff
@@ -0,0 +1,434 @@
...
+        // Create the for loop.
+        // Each loop iteration adds to the total counts.
+        // counts return type is <blah x i64>, for both eigens and counts
+        builder.setInsertionPointToEnd(&qfunc.getBody().front());
+        auto countsSum = builder.create<stablehlo::ConstantOp>(
+            loc, countsType,
+            DenseElementsAttr::get(countsType, builder.getIntegerAttr(builder.getI64Type(), 0)));
+
+        // We also need to yield the eigens from the kernel call inside the loop body
+        // However, this requries the loop to have an iteration argument for the eigens tensor
```
joeycarter:
```suggestion
        // However, this requires the loop to have an iteration argument for the eigens tensor
```

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
```diff
@@ -0,0 +1,434 @@
...
+
+        if (illegalMP) {
+            return signalPassFailure();
+        }
+
+        size_t numReturns = isCounts ? 2 : 1;
+        assert(qfunc.getResultTypes().size() == numReturns &&
+               "Multiple terminal MPs not yet supported in one shot transform");
+
+        // Get the shots SSA value. It is an oeprand to the device init op.
```
joeycarter:
```suggestion
        // Get the shots SSA value. It is an operand to the device init op.
```

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
```diff
@@ -0,0 +1,434 @@
...
+                mp = _mp;
+            }
+            else if (isa<quantum::CountsOp>(_mp)) {
+                isCounts = true;
+                qfunc = _mp->getParentOfType<func::FuncOp>();
+                mp = _mp;
+            }
+            else if (isa<quantum::VarianceOp>(_mp)) {
+                _mp->emitOpError(
+                    "VarianceOp is not currently supported for conversion to one-shot MCM.");
```
joeycarter: Hopefully this doesn't come back to bite us one day! I think in principle it shouldn't be too difficult to implement; it should be the same as expval, except instead of computing the mean it would compute the variance.

joeycarter (approved):
Looks great, thanks @paul0403! This is a complex transform so I hope I haven't missed anything, but the logic of the pass makes sense to me, at least. Two points:

* Don't forget the changelog 😉 
* Is this going to cause issues with other passes, since it's moving the quantum function into its own kernel funcOp? e.g. if we adopt the "better containers for quantum programs" in the MLIR 2.0 proposal that moves the state evolution part into its own function; the ongoing work with split-non-commuting might also be relevant in this respect.

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
rniczh: Thanks for explanation! Like what we discussed in offline if you want to try `WalkResult::skip()`, but I'm aslo okay with the current implemantion, you can resolve this comment at anytime you want :)

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
paul0403: It would be a completely different approach, I'm not actually doing any measurement from samples transform here, e.g. for expval and probs I'm just adding up the one shot expvals and probs and divide.

This wouldn't work for var because the one-shot var would just be zero (there's only one data point), so that would require an actual measurements from sample transform, which was not implemented in this PR.

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
paul0403: Thanks for catching all of these! Does vscode have spell check :sweat_smile: ?

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
paul0403: yes

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
joeycarter: It does :wink:

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
joeycarter: Ah you're right, please ignore :sweat_smile:

dime10: There is a limitation regarding the support of variance. The simple strategy for this transform is to maintain the existing MP, but computed on a single shot now, and then average it later. This works well for expval for example:
- `expval(o)` on 1000 shots ---> `expval(o)` on 1 shot, averaged 1000 times

However, with `var(o)` this strategy doesn't work because you cannot compute a "local" variance on 1 shot. What would be required is to turn the `var(o)` MP into a `sample(o)` MP, and then compute the variance from the samples. The only issue is that we don't support sampling observables in Catalyst, otherwise we would have already implemented this in the existing transform.

We can however support var on MCMs, and I don't think we should shy away from adding it in this PR. Doing so means we'll already have the variance computation ready to go if we do add support for sampling observables down the road (really quite a gap in Catalyst at the moment).

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
dime10: It does have a spell check plugin. Surely sublime does too? 😄

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
dime10: Yeah this is definitely important! The structure of main pass function is a little odd as well, maybe something like 1. walk all qnodes 2. for each qnode, look at MPs, 3. do x, y, z etc. would be cleaner?

paul0403: Yes, I was going to revisit the overall structure after I finish the individual components!

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
paul0403: Yes, var on MCMs is now possible, since we don't need to care about observable (i.e. when it's var of paulixyz), and it's just the raw statistical var of a bunch of booleans.

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
paul0403: image
It does, but I'm starting to remember why I turned it off.....

paul0403: Wait I found the sublime setting to spellcheck comments only

paul0403: It's a beautiful world

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2458/commits/4e3e99232879d2fb1d436cedcf9309b25482962c I refactored my logic quite a bit to support multiple MPs in the return, haven't added tests yet but all should be good now

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
paul0403: Done as part of the big refactor in https://github.com/PennyLaneAI/catalyst/commit/4e3e99232879d2fb1d436cedcf9309b25482962c

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2458/commits/c6fca72c2d9771affe9be2fb4d2fa806bddf61de test for multiple MPs

frontend/catalyst/from_plxpr/qfunc_interpreter.py
```diff
@@ -48,6 +48,7 @@
     gphase_p,
     hamiltonian_p,
     hermitian_p,
+    mcmobs_p,
```
paul0403, line 51: changelog

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
paul0403: @dime10 You know what I just realized? Variance of a single MCM is actually super easy! Because the samples of a single MCM are just a set of 0s and 1s. There's this neat little stat formula 
```
Variance(blah) = Expval(blah^2) - [Expval(blah)]^2
```

but for both 1 and 0, their square is just themselves, so
```
Variance(a set of MCMs) = expval(a set of MCMs) - expval(a set of MCMs)^2
```

e.g. {0,0,1,1}. Mean is 2/4=0.5. Variance is
(a) Actual computation: ((0-0.5)^2 + (0-0.5)^2 + (1-0.5)^2 + (1-0.5)^2)/4 = 0.25
(b) With the trick: 0.5-0.5^2 = 0.25

So we don't even need to do anything fancy with stablehlo! We can literally just generate the same one shot kernel as expval on MCMs, and after the expval MCM's postprocessing after the for loop is done (at which point we have the total expval across all shots), we just do the formula, which is a few simple scalar operations!

frontend/test/pytest/test_one_shot_mcm.py
```diff
@@ -0,0 +1,285 @@
...
+        return qml.expval(qml.X(0))
+
+    res = circuit()
+    assert res.dtype == "float64"
+    assert res.shape == ()
+    assert np.allclose(res, 1.0, atol=0.01, rtol=0.01)
+
+
+@pytest.mark.usefixtures("use_capture")
+def test_mlir_one_shot_pass_expval_mcm(backend):
```
paul0403: Note to self: confirm all these tests' results with default.qubit

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
dime10: True, nice find!

paul0403: https://github.com/PennyLaneAI/catalyst/pull/2458/commits/378f19ab405ecb16130931b679c51fb3a8569650

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
```diff
@@ -0,0 +1,816 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#define DEBUG_TYPE "one-shot-mcm"
+
+#include <functional>
```
paul0403: no need

doc/releases/changelog-dev.md
```diff
@@ -2,6 +2,50 @@
 
 <h3>New features since last release</h3>
 
+* A new MLIR transformation pass `--one-shot-mcm` is available.
```
dime10: Actually, I would like to propose that the pass be renamed. 'dynamic one-shot' is the "official" name we gave the technique (e.g. in the [simulation techniques](
https://docs.pennylane.ai/en/stable/introduction/dynamic_quantum_circuits.html#simulation-techniques) docs). But also, single-shot executions are useful for more than just MCMs, but to capture any sort of non-determinism correctly (e.g. gate noise, classical randomness).

Framing it as an execution method then, rather than a technique to "implement" mcms, is much more appropriate and helpful imo.

frontend/catalyst/jax_primitives.py
```diff
@@ -1832,6 +1834,21 @@ def _named_obs_lowering(jax_ctx: mlir.LoweringRuleContext, qubit: ir.Value, kind
...
+#
+@mcmobs_p.def_abstract_eval
+def _mcmobs_abstract_eval(*mcms):
+    return AbstractObs(len(mcms), mcmobs_p)
+
+
+def _mcm_obs_lowering(jax_ctx: mlir.LoweringRuleContext, *mcms: list[ir.Value]):
+    ctx = jax_ctx.module_context.context
+    result_type = ir.OpaqueType.get("quantum", "obs", ctx)
+    extracted = [extract_scalar(mcm, "mcmobs") for mcm in mcms]
```
dime10, line 1839: Minor optimization would be to grab the input value from the from_elements op, but hopefully the from_elements + extract can be canonicalized away anyway?

mlir/include/Quantum/IR/QuantumOps.td
```diff
@@ -867,6 +867,38 @@ def NamedObsOp : Observable_Op<"namedobs"> {
     }];
 }
 
+def MCMObsOp : Observable_Op<"mcmobs"> {
+    let summary = "Define a pseudo-observable of mid-circuit measurements for use in terminal measurements";
```
dime10: I would strongly encourage everyone to differentiate between "terminal measurements", something that's widely understood in the circuit picture as quantum measurements at the end of a circuit (essentially mcms that are the last operation on a qubit), and "measurement processes". The former are elementary operations in a circuit that are applied to a quantum state during a single execution, whereas the latter is a concept in PennyLane that converts a whole circuit into a classical function, usually involving repeated execution of the circuit and post-processing of aggregated results.
```suggestion
    let summary = "Define a pseudo-observable of mid-circuit measurements for use in measurement processes";
```

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
```diff
@@ -0,0 +1,815 @@
...
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/Index/IR/IndexDialect.h"
+#include "mlir/Dialect/Index/IR/IndexOps.h"
+#include "mlir/Dialect/SCF/IR/SCF.h"
+#include "mlir/Dialect/Tensor/IR/Tensor.h"
+#include "mlir/IR/BuiltinAttributes.h"
+#include "mlir/IR/BuiltinOps.h"
+#include "mlir/Pass/Pass.h"
+#include "stablehlo/dialect/StablehloOps.h"
+#include "llvm/ADT/SmallSet.h"
```
dime10: minor: sorting is off here

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
```diff
@@ -0,0 +1,815 @@
...
+        Operation *mod = getOperation();
+        Location loc = mod->getLoc();
+        IRRewriter builder(mod->getContext());
+
+        // Collect all qnode functions.
+        // We find qnode functions by identifying the parent function ops of MPs
+        SetVector<func::FuncOp> qnodeFuncs;
+        mod->walk([&](MeasurementProcess _mp) {
+            qnodeFuncs.insert(_mp->getParentOfType<func::FuncOp>());
+        });
```
dime10: We should have probably have a simpler method of getting QNodes than this, but this is probably the most reliable atm 😅

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
```diff
@@ -0,0 +1,815 @@
...
+
+    // For whatever reason, upstream mlir decided to not include the op itself into its backward
+    // slice
+    backwardSlice.insert(shots.getDefiningOp());
+
+    SmallVector<Operation *> eraseWorklist;
+    for (auto &region : qnodeFunc->getRegions()) {
+        for (auto &block : region.getBlocks()) {
+            for (auto op = block.rbegin(); op != block.rend(); ++op) {
+                if (!backwardSlice.contains(&*op)) {
```
dime10: What does `&*` do?

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
```diff
@@ -0,0 +1,815 @@
...
+            for (auto op = block.rbegin(); op != block.rend(); ++op) {
+                if (!backwardSlice.contains(&*op)) {
+                    eraseWorklist.push_back(&*op);
+                }
+            }
+        }
+    }
+
+    for (Operation *op : eraseWorklist) {
+        op->erase();
```
dime10: Normally, all IR mutation has to go through the rewriter, what was your reasoning to side-step it here?

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
```diff
@@ -0,0 +1,815 @@
...
+                    eraseWorklist.push_back(&*op);
```
dime10, line 133: Can't we erase right here (saves us an iteration)? Or is there an advantage to the worklist?

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
```diff
@@ -0,0 +1,815 @@
...
+        // llvm::enumerate() marks the iterators with const
+        auto mp = const_cast<catalyst::quantum::MeasurementProcess &>(_mp);
+
+        if (isa<quantum::StateOp>(mp)) {
+            mp->emitOpError("StateOp is not compatible with shot-based execution, and has no "
+                            "valid conversion to one-shot MCM.");
+            return signalPassFailure();
+        }
+
+        else if (isa<quantum::VarianceOp>(mp)) {
```
dime10: minor: since you're always casting afterwards you could use dyn_cast

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
```diff
@@ -0,0 +1,815 @@
...
+        }
+
+        else if (isa<quantum::SampleOp>(mp)) {
+            prepareForLoopSampleArgs(builder, oneShotKernel, cast<quantum::SampleOp>(mp), i,
+                                     loopIterArgs, loopIterArgsMPKinds);
+        }
+
+        else if (isa<quantum::CountsOp>(mp)) {
+            quantum::CountsOp countsOp = cast<quantum::CountsOp>(mp);
+            if (handledCountsOps.contains(countsOp)) {
```
dime10: What makes counts different that it needs this extra bookkeeping?

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
```diff
@@ -0,0 +1,815 @@
...
+#define GEN_PASS_DEF_ONESHOTMCMPASS
+#include "Quantum/Transforms/Passes.h.inc"
+
+struct OneShotMCMPass : public impl::OneShotMCMPassBase<OneShotMCMPass> {
+    using impl::OneShotMCMPassBase<OneShotMCMPass>::OneShotMCMPassBase;
+
+    // Misc helpers
+    std::optional<MeasurementProcess> getMPFromValue(Value);
+
+    // Methods to update the one shot kernel
```
dime10: Why are some of the helpers functions and others methods? Are you using something from the pass instance? (searched for `this` but it doesn't appear to be used)

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
```diff
@@ -0,0 +1,815 @@
...
+                                            SmallVector<std::string> &loopIterArgsMPKinds)
+{
+    OpBuilder::InsertionGuard guard(builder);
+
+    Operation *retOp = oneShotKernel.getBody().back().getTerminator();
+    assert(isa<func::ReturnOp>(retOp) && "Expected a qnode function to return values");
+
+    SmallVector<MeasurementProcess> qnodeMPs;
+    for (Value returnValue : retOp->getOperands()) {
+        qnodeMPs.push_back(*getMPFromValue(returnValue));
```
dime10: This returns an optional right, what happens if the value is absent?

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
```diff
@@ -0,0 +1,815 @@
...
+    // Since the kernel was a clone of the original qnodeFunc, their arguments are the same!
+    auto kernalCallOp = func::CallOp::create(
+        builder, loc, oneShotKernel.getFunctionType().getResults(), oneShotKernel.getSymName(),
+        qnodeFunc.getBody().front().getArguments());
+
+    SmallVector<Value> loopYields;
+
+    for (auto [i, mpKind] : llvm::enumerate(loopIterArgsMPKinds)) {
+        if (mpKind == "expval" || mpKind == "variance") {
+            // Add the expval from each iteration
```
dime10: might be good to say what you're doing with the variance

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
```diff
@@ -0,0 +1,815 @@
...
+        else if (mpKind == "sample") {
+            // Insert the one shot kernel's sample into the full sample result tensor
+            ArrayRef<int64_t> oneShotSampleShape =
+                cast<ShapedType>(oneShotKernel.getFunctionType().getResults()[i]).getShape();
+            auto zero = builder.getIndexAttr(0);
+            auto one = builder.getIndexAttr(1);
+            SmallVector<OpFoldResult> offsets = {forOp.getInductionVar(), zero};
+            SmallVector<OpFoldResult> sizes = {builder.getIndexAttr(oneShotSampleShape[0]),
+                                               builder.getIndexAttr(oneShotSampleShape[1])};
+            SmallVector<OpFoldResult> strides = {one, one};
```
dime10: Why `OpFoldResult`?

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
```diff
@@ -0,0 +1,815 @@
...
+            // Add the probs from each iteration
+            auto addOp = stablehlo::AddOp::create(builder, loc, kernalCallOp.getResult(i),
+                                                  forOp.getRegionIterArg(i));
+            loopYields.push_back(addOp.getResult());
+        }
+
+        else if (mpKind == "sample") {
+            // Insert the one shot kernel's sample into the full sample result tensor
+            ArrayRef<int64_t> oneShotSampleShape =
+                cast<ShapedType>(oneShotKernel.getFunctionType().getResults()[i]).getShape();
```
dime10: I'm guessing this is to handle various single-shot sample sizes due to wires? Does it work if the single-shot sample has a dynamic shape?

dime10 (commented):
Looks great! Some things I noticed:

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
paul0403: You can't erase things in a container while iterating through the same container. This is a common pitfall in llvm no?

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
paul0403: It has two results, the eigens and the counts. Both would be the returned from the function, so the same counts op would be identified as the source MP op twice. The second time around, we shouldn't redo the processing.

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
paul0403: I think it will work because under the hood dynamic dimensions are just the `ShapedType::kdynamic`, which is `INT_MIN` I think? But good point, I hadn't really thought of dynamic shapes here. Let me test it tomorrow.

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
paul0403: I will confirm tomorrow, but this seems to be the age-old problem of clang-format wanting to put the llvm headers at the end of the list... (you remember how we though `l` coming before `m` means llvm headers would be put at the top?)

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
paul0403: This gets a raw pointer to the element the iterator is pointing to.

This is needed because the elements in these containers that these iterators point to are actually `Operaion`, not `Operation *`!
https://github.com/llvm/llvm-project/blob/7952896611fb7e5e4dade24c9fc5583c7dc42363/mlir/include/mlir/IR/Block.h#L146-L156

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
paul0403: https://github.com/llvm/llvm-project/blob/7952896611fb7e5e4dade24c9fc5583c7dc42363/mlir/include/mlir/Dialect/Tensor/IR/TensorOps.td#L924-L929

Short (and boring) answer: that's what the builder of `tensor.insert_slice` op wants 

Long answer: `tensor.insert_slice` op can take in both static attributes and dynamic values for these ancillary parameters. Here, the op we want to create is 
```mlir
// say num_qubits = 2, total shots = 1000
%inserted = tensor.insert_slice %one_shot_sample into %full_samples[%loop_index, 0] [1, 2] [1, 1] : tensor<1x2xi64> into tensor<1000x2xi64>
```

i.e. inserting a size 1x2 tensor into position (i, 0) in the full sample result tensor, with stride 1 in both dimensions (i.e. no special striding).

So, we have both static and dynamic pieces of data for these parameters. So what should we do in the MLIR CAPI to create that op? The op signature is 
```
  let arguments = (ins
    AnyRankedTensor:$source,
    AnyRankedTensor:$dest,
    Variadic<Index>:$offsets,
    Variadic<Index>:$sizes,
    Variadic<Index>:$strides,
    DenseI64ArrayAttr:$static_offsets,
    DenseI64ArrayAttr:$static_sizes,
    DenseI64ArrayAttr:$static_strides
  );
```
So in principle we could painstakingly create all 6 fields and pass it into the default builder for the op. But that doesn't feel very nice, especially considering that the "offset" field consists of both a dynamic (`%loop_index`) and a static (`0`) piece, so it's not even immediately clear how we even pass it in! I mean, do we put the `%loop_index` value into the `Variadic<Index>:$offsets` field, and `0` attribute into the ` DenseI64ArrayAttr:$static_offsets` field? This isn't even mentioning the need to carefully match all the operand segment sizes. This feels quite error-prone and messes with my brain a bit.

They provided that convenience builder for exactly this reason.[ The `OpFoldResult` type is a "union" between an `Attribute` and a `Value`](https://github.com/llvm/llvm-project/blob/7952896611fb7e5e4dade24c9fc5583c7dc42363/mlir/include/mlir/IR/OpDefinition.h#L272), so we can just initialize a container of them with either an `Attribute` or a `Value`, and it will figure out the rest for us!

(Secret third answer, and the actual answer: I found out how to create a `tensor.insert_slice` op on super grep and copied it)

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
paul0403: So this comment just made me realize that I never implemented returning classical values from the qfunc in one shot mode! 

Well, let me do that after addressing the other minor comments first!

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
paul0403: Just confirmed, `clang-format` keeps putting `llvm` down at the end :cry:

paul0403: wait no, if we do 
```
include llvm
include mlir
```
then `llvm` will get kicked to the end, but if we leave a space
```
include llvm

inclue mlir
```
then it will stay...

anyway, I left a space https://github.com/PennyLaneAI/catalyst/pull/2458/commits/60bc6ec1e5e694d05cce4794f844d5f69dd5e9f9

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
paul0403: I didn't bother here because this is not a pattern rewrite, but a flat pass with no patterns, so there's no "IR state" that the rewriter needs to repeatedly track. But you're right, we should still go through the builder. https://github.com/PennyLaneAI/catalyst/pull/2458/commits/4b8b2199e6ae867c774193a21c6ad31a4b0bd9b2

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
paul0403: Yes, I kinda don't want to rely on the `qnode` func attribute because I don't want these assumptions from the frontend, especially with one of the IR improvements being updates to the quantum program's structure.

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
dime10: Ah makes sense, such a weird thing to see cause you'd think the two operators are inverses of each other 😅

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
dime10: Oh good point, that rule only applies to rewrite patterns

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
dime10: gotcha! would be a good comment

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
dime10: Haha yeah, I was wondering if those had been tested. I didn't say anything because I thought we are going to enforce MPs for QNodes in the future anyway, but maybe it's better to be robust here.

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
dime10: Well the thing is if you have kdynamic you would normally need an SSA value to go with it.

~Separately, it would be good to name your variables here `static_offsets` etc., just like in [the op](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorinsert_slice-tensorinsertsliceop), to make it clear which type of operand you are providing.~ Scrap that, now that I know you are using a mixed version^^

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
dime10: Interesting, I didn't know about this. That builder is indeed very convenient! 
Just to answer your question regarding the construction with separated static and dynamic arguments, it is well-defined (and not as complex as you make it out to be, just more verbose):
- the static array has length equal to the total number of args, where they are dynamic you provide a kdynamic sentinel value, otherwise the constant value directly
- the dynamic array is only as long as the number of dynamic args, the order corresponds to the order the sentinels appear in the static array

The reason I commented was because I was trying to figure out what was going on regarding the args, whether you were using static or dynamic ones, etc., but it looked very confusing to me, also with the OpFoldResult type which seemed out of place ("what does folding have to do with anything here?" 😅).
But now it makes sense, and I guess they use OpFoldResult precisely because it allows both attributes and values to be mixed. Thanks for the explanation :)

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
dime10: Hmm yes, generally you do have to be careful when erasing elements in a container that is being iterated. I thought that's what your comment was about though:
> // We need triple loop to explicitly iterate in reverse order, due to erasure.

that you were explicitly iterating in reverse so that the erasure of elements wouldn't conflict with the iteration. 

I was trying to find out definitively whether this style of erasure is ok or not, but I don't know for certain (maybe it would require looking into the containers MLIR is using). The context where I do know operation erasure is explicitly allowed during traversal is a [post-order walk](https://mlir.llvm.org/doxygen/namespacemlir_1_1detail.html#a152b851682ac1cad31c84dc1a6fc38cb):
> A callback on a block or operation is allowed to erase that block
 or operation only if the walk is in post-order. See non-void method for
 pre-order erasure.

paul0403: > // We need triple loop to explicitly iterate in reverse order, due to erasure.

This comment was about the fact that, since the purpose of this function is to clear the body of a func op, and we cannot just call `op.clear()` because we still need to shots SSA value to be kept, we have to explicitly loop and erase. But then, if we loop in the normal order, erasing earilier ops before later ops will run into the "erasing ops while still have uses" error.

This is a separate thing from the other thing, which is erasing ops in a block while iterating through the block. This will usually cause a pass runtime error saying something like iterator reached the end.

I didn't actually know that post-order walk allows it! I guess it kinda make sense since the iterator would never "run out of space" in the container if we're deleting from the end!

But anyway, I didn't do a walk here, because the purpose of this function is just to clear the body. I only need to erase the top level ops. For example, if the qfunc has a for loop, I only need to erase the for loop, and there's no need to walk into the for loop and erase things inside recursively. Hongsheng pointed out that there is `WalkResult::skip`, which combined with your suggestion here of post-order walking can achieve the same effect, but I don't think there's really anything wrong with the reverse manual iteration and the erasure worklist.

dime10: Agreed, I'm not suggesting you use a walk here (like you say there is no reason to descend into nested regions). I brought it up in case it could help us answer the original question (more for the knowledge at this point since the code is also fine as is). For instance, a walk surely iterates over a block right? If that is the case, and they are using the same kind of iterators as you are here, then deleting while iterating would actually be fine. I'm assuming a walk even uses a forward iterator rather than a reverse one.

In any case, happy to keep the worklist. A minor simplification would be to reverse iterate the worklist rather than the blocks, if that is the reason for the triple loop nest :)

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
paul0403: It was mostly arbitray :sweat: I don't actually know the best organization, does it make sense to pull all helpers out of the pass class? None of them rely on anything regarding the particular instance of the pass, i.e. none of them communicate any sort of shared data between them on the same pass application instance.

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
rniczh: Just sharing the way they (llvm) erase the op while walking the iterator:
`catalyst/mlir/llvm-project/llvm/lib/CodeGen/GlobalISel/InstructionSelect.cpp`

```cpp
for (auto MII = MBB.rbegin(), End = MBB.rend(); MII != End;) {
  MachineInstr &MI = *MII;
  ++MII;

  if (MI.getOpcode() != TargetOpcode::COPY)
    continue;
  Register SrcReg = MI.getOperand(1).getReg();
  Register DstReg = MI.getOperand(0).getReg();
  if (SrcReg.isVirtual() && DstReg.isVirtual()) {
    auto SrcRC = MRI.getRegClass(SrcReg);
    auto DstRC = MRI.getRegClass(DstReg);
    if (SrcRC == DstRC) {
      MRI.replaceRegWith(DstReg, SrcReg);
      MI.eraseFromParent();
    }
  }
}
```

Or in the forward iteration: `catalyst/mlir/llvm-project/llvm/lib/CodeGen/PeepholeOptimizer.cpp`
```cpp
for (MachineBasicBlock::iterator MII = MBB.begin(), MIE = MBB.end();
     MII != MIE;) {
  MachineInstr *MI = &*MII;
  // We may be erasing MI below, increment MII now.
  ++MII;
```

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
dime10: Yeah in that case I would go with either one or the other consistently :)
Currently, it also makes reading the file a bit awkward, with some helpers above and some below, and the entry point somewhere in the middle.

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2458/commits/9f27dfb32393082669baa3c913bf1e98aa13ed33

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2458/commits/9f27dfb32393082669baa3c913bf1e98aa13ed33

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
paul0403: > A minor simplification would be to reverse iterate the worklist rather than the blocks

https://github.com/PennyLaneAI/catalyst/pull/2458/commits/9f27dfb32393082669baa3c913bf1e98aa13ed33

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2458/commits/9f27dfb32393082669baa3c913bf1e98aa13ed33

mlir/lib/Quantum/Transforms/one_shot_mcm.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2458/commits/6b2a90af0bab3faecc06122ae140f3f53bd2106b I moved all helpers to the anonymous namespace to the top, since they don't really care about the pass instance, and they don't need to exist outside this file for the rest of the `quantum` namespace.

doc/releases/changelog-dev.md
paul0403: renamed https://github.com/PennyLaneAI/catalyst/pull/2458/commits/309d5a13461355b166663b8e53222afc7dabb46e

mlir/include/Quantum/IR/QuantumOps.td
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2458/commits/3685be523572bfbd9afef357e64c543c2068453d

frontend/catalyst/jax_primitives.py
paul0403: I don't know why [we generate a from_elements op in our own measure op's lowering](https://github.com/PennyLaneAI/catalyst/blob/d9f8cd2f35eefd19d34130c3e09e030c16da3df7/frontend/catalyst/jax_primitives.py#L1674). Is it just because stablehlo needs tensors?

paul0403: In any case, canonicalization does know to deal with them
image

dime10: >Is it just because stablehlo needs tensors?

Precisely.

> canonicalization does know to deal with them

Awesome!

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
paul0403: Well, we didn't even implement qubit-invariant compilation in the plxpr path, so the single-shot sample will not have a dynamic shape for now: its shape will always be `<1xstatic_num_qubits>`. Although, I think in principle we can very easily support dynamic number of wires, because it's just another operand on the sample op. We will immediately have this functionality once the frontend is ready. Let me try.

paul0403: https://github.com/PennyLaneAI/catalyst/pull/2458/commits/2420153f22ec53eceef7c1e5f4a6e298378be01f

Dynamic number of wires on samples. 

That was something... but now that I have the infra setup, I can probably do dynamic number of wires on counts and probs too.

Although we won't be able to test this end-to-end due to restrictions in capture fronted, we will be ready when the frontend becomes available.

paul0403: https://github.com/PennyLaneAI/catalyst/pull/2458/commits/433192b92289931462273ac1f27dac5a09a2f684 dynamic number of qubits on probs

paul0403: https://github.com/PennyLaneAI/catalyst/pull/2458/commits/037d7c3293c41522f4f5487bd744d12cf80f58a5 dynamic number of qubits on counts

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2458/commits/9116d896f7e5e9ec2bd9982e582cbf824fc38447 added an assert

doc/releases/changelog-dev.md
```diff
@@ -2,6 +2,50 @@
...
+  them one shot at a time, sampling a dynamic execution path for each shot. The `--dynamic-one-shot`
+  pass first transforms the circuit so that each circuit execution only contains a singular shot,
+  then performs the appropriate classical statistical postprocessing across the execution results
+  from all shots.
+  [(#2458)](https://github.com/PennyLaneAI/catalyst/pull/2458)
+
+  With this new MLIR pass, one shot execution mode is now available when capture is enabled.
+
+  ```python
+  qml.capture.enable()
```
dime10: I believe we've moved to the qjit kwarg for this, should make sure no new PRs with the old paradigm.

dime10:
The lit test segfaults on my machine, so I ran the sanitizer:
> ==63432==ERROR: AddressSanitizer: heap-use-after-free

We really should start mandatory asan runs on all C++ PRs 😬

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
```diff
@@ -0,0 +1,991 @@
...
+        // Collect all qnode functions.
+        // We find qnode functions by identifying the parent function ops of MPs
+        SetVector<func::FuncOp> qnodeFuncs;
+        mod->walk([&](MeasurementProcess _mp) {
+            qnodeFuncs.insert(_mp->getParentOfType<func::FuncOp>());
+        });
+
+        // For each qnode function, find the returned MPs
+        // Then handle the one shot logic for each MP type
+        for (auto qnodeFunc : qnodeFuncs) {
```
dime10, line 1078: The now classical function calling the new qnode function should have its `qnode` attribute removed :)

doc/releases/changelog-dev.md
paul0403: Oh, I updated for the tests but forgot the changelog :melting_face:

paul0403:
Re: segfault

It happens for the case of counts op on MCMs.
I am remembering the processed counts ops in a set, so we don't double process it from the eigens result and the counts result. But when there's counts on MCM, the one-shot kernel's counts op is actually erased (since the MCMs themselves are the results). So that set remembering processed counts ops now contains a bad pointer!

In the loop over all MPs of the qfunc's returned values, I change to just skip the loop iteration after the "eigens" values entirely, and eliminated the need for that "visited counts" set  https://github.com/PennyLaneAI/catalyst/pull/2458/commits/6a72b6dfd340a13b484c331ac687fb548f7d2a0f

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
```diff
@@ -0,0 +1,992 @@
...
+
+    // CountsOp has two results, the eigens and the counts.
+    // Both would be returned from the quantum function, so the same counts op would be
+    // identified as the source MP op twice.
+    // The second time around, we shouldn't redo the processing.
+    size_t skip_counts_iterator = -1;
+    for (auto [i, _mp] : llvm::enumerate(qnodeMPs)) {
+        if (i == skip_counts_iterator) {
+            continue;
+        }
```
paul0403, line 823: I don't like this: this, along with the 
```C
        else if (auto counts = dyn_cast<quantum::CountsOp>(mp)) {
            prepareForLoopCountsArgs(builder, oneShotKernel, counts, i, loopIterArgs,
                                     loopIterArgsMPKinds, cloneMapper);
            skip_counts_iterator = i + 1;
        }
``` 
assumes that the eigen `Value` and counts `Value` are always returned from the qfunc next to each other. 

While this assumption is true, due to how lowering is done in the frontend, again I don't want any implicit assumptions from the frontend, since these kind of assumptions invalidate all potential user/non-user passes happening between the frontend and the current pass. For example, there's nothing stopping a user pass from reordering their qfunc's return `Value`s' order, for whatever reason.

This was why I opted for a "visited counts ops" set to begin with, instead of just skipping over the value after eigens explicitly. 

This now is just a proof of concept to validate that this is indeed the source of the segfault. I'll come up with a better organization tomorrow.

doc/releases/changelog-dev.md
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2458/commits/78ecb50c0d813f2ecc545571ba0b70f035009a71

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2458/commits/b01e6f33ca9c41e719993f6a381d4b002f0ac419

paul0403:
> For now if shots=0 a simple exit would be sensible

https://github.com/PennyLaneAI/catalyst/pull/2458/commits/2ee8fd8b1f45841c566270b835f185a70cbe2e92

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2458/commits/a3c0ff789e0e3626d9f3eb5691c36e8dc5d95c5f

frontend/test/pytest/test_dynamic_one_shot.py
dime10, line 1: Just a question regarding the existing tests for one shot. Are we testing new functionality in this file, or expanding missing coverage for existing functionality?
I'm also wondering where are the existing tests are located, should they not live together?

frontend/catalyst/from_plxpr/qfunc_interpreter.py
```diff
@@ -267,28 +272,34 @@ def interpret_measurement(self, measurement):
...
-            raise NotImplementedError("Measurements of mcms are not yet supported.")
-
-        if measurement.obs:
-            obs = self._obs(measurement.obs)
+
+        prim = measurement_map[type(measurement)]
+        mcm_mv = measurement.mv
+        mcm_obs = measurement.obs
+        mp_is_on_mcm = (mcm_mv is not None) or (
+            mcm_obs is not None and not isinstance(mcm_obs, qml.operation.Operator)
```
rniczh, line 280: Is it because the mcm is the only qml.operator that can be as the observables?

dime10 (approved):
Thanks for the nice work on this @paul0403 !

frontend/test/pytest/test_dynamic_one_shot.py
paul0403: The existing ones live in `test_midcircuit_measurements.py`, but I wanted to separate the new pass and the old tape transform. 

But I guess that's also possible with just adding a new class in that file. I'll move it there.



> Are we testing new functionality in this file, or expanding missing coverage for existing functionality?

I would argue from a development perspective, this is completely new code and hence a new feature, but from a user side probably this is just an improvement to an existing feature, e.g. now we can do one shot with capture, measurement_from_samples, etc.

paul0403: https://github.com/PennyLaneAI/catalyst/pull/2458/commits/e4fa4655b2aa21cecddf655e26c7cee73fc3b4f2

frontend/test/lit/test_mid_circuit_measurement.py
```diff
@@ -95,3 +95,47 @@ def circ():
 
 
 print(test_one_shot_with_passes.mlir)
+
```
rniczh, line 98: There is a discussion about the meaning of frontend lit test:
https://github.com/PennyLaneAI/catalyst/pull/2441#discussion_r2771230597

@dime10 
> but unless there is something specific happening in the frontend or lowering mechanism that needs testing

So just wondering, for this case, is it worth to create a mlir test here?

rniczh (approved):
Thanks @paul0403 , just some questions

frontend/test/lit/test_mid_circuit_measurement.py
paul0403: Yes, there is new logic in the plxpr conversion for the new mcm obs op. Notice that I'm only testing the generation of the mcm obs ops, not the one-shot execution results! Those I use pytests.

frontend/catalyst/from_plxpr/qfunc_interpreter.py
paul0403: I don't really know, but this was the old condition to catch MPs on MCMs so I just trusted it 😅 probably comes from Christina

PR #2459, opened 2026-02-02
Fix Global Phase on CNOT in to-ppr conversion

sengthai:
**Context:**
Fix Global Phase on CNOT in to-ppr conversion


lazypanda10117 (approved):
Thanks!

paul0403 (approved):

PR #2460, opened 2026-02-02
Run PPR/PPM lowering passes in the main pipeline

sengthai:
**Context:**
The PPR/PPM lowering passes (`lower-qec-init-ops`, `unroll-conditional-ppr-ppm`) are now run
  as part of the main quantum compilation pipeline. When using `to-ppr` and `ppr-to-ppm` transforms,
  these passes are applied automatically during compilation; we no longer need to stack them
  explicitly.

[[sc-109665]]


frontend/catalyst/pipelines.py
```diff
@@ -226,6 +226,9 @@ def get_quantum_compilation_stage(_options: CompileOptions) -> List[str]:
         # keep inlining modules targeting the Catalyst runtime.
         # But qnodes targeting other backends may choose to lower
         # this into something else.
+        # TODO: Remove these once PBC has its own pipeline
+        "lower-qec-init-ops",
+        "unroll-conditional-ppr-ppm",
```
sengthai: These two passes are required before lowering to CAPI. However, to ensure it runs before lowering is tricky, as they don't have their own pipelines and add before SCF lowering. The quick solution is to place them here. Feel free to suggest any approach!

sengthai: As discussed with David, registering passes here is acceptable. However, once we have PBC pipeline, we will revisit this later.

frontend/catalyst/pipelines.py
```diff
@@ -229,6 +229,9 @@ def get_quantum_compilation_stage(_options: CompileOptions) -> List[str]:
         "inline-nested-module",
         "lower-mitigation",
         "adjoint-lowering",
+        # TODO: We can remove 2 passes below once PBC has its own pipeline.
+        "lower-qec-init-ops",
```
lazypanda10117, line 248: I am not sure if we want to `lower-qec-init-ops` in the default pipeline? Especially in the vISA implementation, I think we explicitly exclude this pass.

sengthai: I thought it shouldn't be in the default pipeline, either. But since its behavior supports runtime and ensures it works, as with `lower-mitigation`, it is fine to be here. when the time we support vISA lowering, we still use this pass for running PPR/PPM.

dime10: >Especially in the vISA implementation, I think we explicitly exclude this pass.

That pathway won't be using the default compile to executable pipeline anyway right

lazypanda10117: Right i guess that's true. But i am still not sure if adding to `quantum-compilation-stage` is a good idea. This is only needed for execution in the PBC pipeline.

dime10: It's needed whenever the user applies PBC passes right? In a way having these lowerings reintegrates the PBC dialect into the main pipeline, rather than it being separate. For the user this is relatively seamless, because they don't have to worry about different compilation pipelines.

Do you have a different workflow you would like to suggest?

lazypanda10117: Hmm, i guess this is true. I don't really have an alternative suggestion other than a PBC specific pipeline. I can approve this for now.

lazypanda10117 (approved):
Looks good to me. Thanks Sengthai!

frontend/test/lit/test_pauli_rot_and_measure.py
```diff
@@ -404,7 +404,14 @@ def circuit():
         qml.PauliRot(np.pi / 4, "Y", wires=0)
         qml.pauli_measure("X", wires=0)
 
-    # CHECK: qec.select.ppm
+    # CHECK: qec.ppm ["X", "Y"]
+    # CHECK: qec.ppm ["Y", "Z"]
+    # CHECK: scf.if
+    # CHECK: qec.ppm ["Y"]
+    # CHECK: }
```
paul0403:
```suggestion
```

There's usually no need to check these sort of lines. Same for the ones below.

frontend/test/lit/test_pauli_rot_and_measure.py
```diff
@@ -404,7 +404,14 @@ def circuit():
         qml.PauliRot(np.pi / 4, "Y", wires=0)
         qml.pauli_measure("X", wires=0)
 
-    # CHECK: qec.select.ppm
+    # CHECK: qec.ppm ["X", "Y"]
+    # CHECK: qec.ppm ["Y", "Z"]
```
paul0403, line 408: Just for my understanding, this is the effect of `--unroll-conditional-ppr-ppm` right? And now that we run that pass in the default pipeline, its effect is now visible in `mlir_opt`?

frontend/test/pytest/test_peephole_optimizations.py
```diff
@@ -27,7 +27,6 @@
     disentangle_swap,
     merge_ppr_ppm,
     merge_rotations,
-    ppm_compilation,
```
paul0403, line 30: Why do we no longer do `ppm_compilation`? Sorry I'm not super familiar with the PBC passes.

paul0403: Oh, is this the port from the legacy catalyst frontend pass decorator to `qml.transform(pass_name="ppm-compilation")`?

mlir/lib/QEC/Transforms/ConversionPatterns.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+} // namespace
+
+namespace catalyst {
+namespace qec {
+
+void populateConversionPatterns(LLVMTypeConverter &typeConverter, RewritePatternSet &patterns)
+{
+    // Pauli-based Computational Operations
+    patterns.add<PPRotationBasedPattern<PauliRotOp>>(typeConverter, patterns.getContext());
```
paul0403, line 197: Question: this `PauliRotOp` is in the `qec` namespace, but the old pattern file (the one that got moved to this new one) seems to be adding conversion patterns to the one in the `quantum` namespace. Does this mean the PPR op in the quantum dialect has no lowering again?

frontend/test/lit/test_pauli_rot_and_measure.py
sengthai: Yes, `--unroll-conditional-ppr-ppm` did that. And it affect in `mlir_opt`.

frontend/test/pytest/test_peephole_optimizations.py
sengthai: Yes, we use `qml.transform(pass_name="ppm-compilation")` instead.

mlir/lib/QEC/Transforms/ConversionPatterns.cpp
sengthai: > this PauliRotOp is in the qec namespace

Actually, the `PauliRotOp` is not in `qec` namespace, but it seems those lowering are identical with PPR ops, I encapsulated `PauliRotOp` lowering in here.

> Does this mean the PPR op in the quantum dialect has no lowering again?

Good question! 
Yes, PPR ops won't be lowered again. In fact, Some PPR ops are lowered to quantum ops (e.g `qec.fabricate`). So then `convert-quantum-to-llvm` will take care of quantum ops.

frontend/test/lit/test_pauli_rot_and_measure.py
```diff
@@ -404,7 +404,13 @@ def circuit():
...
+    # CHECK: scf.if
+    # CHECK: qec.ppm ["Y"]
+    # CHECK: else
+    # CHECK: qec.ppm ["X"]
+    # CHECK: }
```
paul0403, line 413: @sengthai these others too!

mlir/lib/QEC/Transforms/ConversionPatterns.cpp
paul0403: So what happens now with the `quantum.pauli_rot` ops coming from the frontend?

sengthai: Ohh one thing to know that that PPR Op and pauliOp will lower to `__catalyst__qis__PauliRot`, so it will go to that path which lowers to `__catalyst__qis__PauliRot` from frontend.

frontend/test/lit/test_pauli_rot_and_measure.py
sengthai: I think this one is needed to show to ensure no operations in that else-branch.

mlir/lib/QEC/Transforms/ConversionPatterns.cpp
paul0403: Sorry what do you mean? It lowers to a runtime CAPI stub from the frontend directly? 😱 That doesn't sound right.

sengthai: From Jaxpr, it compiles down to `quantum.pauli_rot` first.

mlir/lib/QEC/Transforms/ConversionPatterns.cpp
```diff
@@ -0,0 +1,208 @@
...
+        args.push_back(
+            LLVM::ConstantOp::create(rewriter, loc, rewriter.getI64IntegerAttr(numQubits)));
+        args.insert(args.end(), adaptor.getInQubits().begin(), adaptor.getInQubits().end());
+
+        LLVM::CallOp::create(rewriter, loc, fnDecl, args);
+
+        // Replace the op with the input qubits
+        SmallVector<Value> values;
+        values.insert(values.end(), adaptor.getInQubits().begin(), adaptor.getInQubits().end());
+        rewriter.replaceOp(op, values);
```
rniczh: I think you could just replace op with `adaptor.getInQubits()`. Don't need to copy the inQubits to vector first.

```suggestion
        rewriter.replaceOp(op, adaptor.getInQubits());
```

mlir/lib/QEC/Transforms/ConversionPatterns.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+        // Build the arguments: pauliStr, theta, modifiers, numQubits, qubits...
+        int64_t numQubits = adaptor.getInQubits().size();
+        SmallVector<Value> args;
+        args.push_back(pauliWordPtr);
+        args.push_back(thetaValue);
+        args.push_back(LLVM::ZeroOp::create(rewriter, loc, ptrType));
+        args.push_back(
+            LLVM::ConstantOp::create(rewriter, loc, rewriter.getI64IntegerAttr(numQubits)));
+        args.insert(args.end(), adaptor.getInQubits().begin(), adaptor.getInQubits().end());
```
rniczh: just append?

```suggestion
        args.append(adaptor.getInQubits().begin(), adaptor.getInQubits().end());
```

mlir/lib/QEC/Transforms/ConversionPatterns.cpp
```diff
@@ -0,0 +1,208 @@
...
+        LLVM::LLVMFuncOp fnDecl = catalyst::ensureFunctionDeclaration<LLVM::LLVMFuncOp>(
+            rewriter, op, qirName, qirSignature);
+
+        // Build the arguments
+        int64_t numQubits = adaptor.getInQubits().size();
+        SmallVector<Value> args;
+        args.push_back(pauliWordPtr);
+        args.push_back(
+            LLVM::ConstantOp::create(rewriter, loc, rewriter.getI64IntegerAttr(numQubits)));
+        args.insert(args.end(), adaptor.getInQubits().begin(), adaptor.getInQubits().end());
```
rniczh: same as https://github.com/PennyLaneAI/catalyst/pull/2460/changes#r2795212246

```suggestion
        args.append(adaptor.getInQubits().begin(), adaptor.getInQubits().end());
```

mlir/lib/QEC/Transforms/ConversionPatterns.cpp
```diff
@@ -0,0 +1,208 @@
...
+        if (static_cast<int16_t>(op.getRotationSign()) == -1) {
+            Value one = LLVM::ConstantOp::create(rewriter, loc, rewriter.getI1Type(),
+                                                 rewriter.getBoolAttr(true));
+            mres = LLVM::XOrOp::create(rewriter, loc, mres, one);
+        }
+
+        // Replace the op with the measurement result and the input qubits
+        SmallVector<Value> values;
+        values.push_back(mres);
+        values.insert(values.end(), adaptor.getInQubits().begin(), adaptor.getInQubits().end());
```
rniczh:
```suggestion
        values.append(adaptor.getInQubits().begin(), adaptor.getInQubits().end());
```

rniczh (commented):
Just some fly-by comments

mlir/lib/QEC/Transforms/ConversionPatterns.cpp
paul0403: Discussed offline, I understand the structure now.

paul0403 (approved):
Thanks!

PR #2464, opened 2026-02-04
fix: update `mlir_specs` for new `marker` implementation

andrijapau:
Changes required after https://github.com/PennyLaneAI/pennylane/pull/9007 is merged.

Essentially, now that markers work with Catalyst, we need it to pick up the marker's name if we place a marker after an MLIR pass.

[sc-110163]


frontend/catalyst/python_interface/inspection/specs.py
jzaia18: Maybe it would be easier to handle the marker names here directly? By overriding the pass name that is used in the cache

frontend/catalyst/python_interface/inspection/specs.py
```diff
@@ -109,7 +117,19 @@ def _specs_callback(previous_pass, module, next_pass, pass_level=0):
...
+            # NOTE: Ensures that markers on the same level are joined
+            # correctly as a string delimited by commas.
+            (
+                ", ".join(m if not isinstance(m, str) else [m])
+                if (m := level_to_markers.get(lvl))
+                else f"{cache[lvl][1]} (MLIR-{lvl})"
+            ): cache[lvl][0]
+            for lvl in level
+            if lvl in cache
+        }
```
jzaia18: I think this return statement is a bit complex and would be better split into several parts. See above comment, I think we can handle getting the marker names within the callback directly

jzaia18 (commented):
Looking good! Will you need to make similar changes for the graph visualization?

frontend/catalyst/python_interface/inspection/specs.py
andrijapau: Yeah that's a good idea. Done here: https://github.com/PennyLaneAI/catalyst/pull/2464/commits/88c7971dc91f9a2baf45d6161930b91b42732aa3

frontend/catalyst/python_interface/inspection/specs.py
andrijapau: See other [comment](https://github.com/PennyLaneAI/catalyst/pull/2464#discussion_r2794427863) for discussion.

frontend/catalyst/python_interface/inspection/specs.py
```diff
@@ -83,7 +91,12 @@ def _specs_callback(previous_pass, module, next_pass, pass_level=0):
...
-        pass_name = pass_instance.name if hasattr(pass_instance, "name") else pass_instance
+        pass_name = pass_instance
+        if m := level_to_markers.get(pass_level):
+            pass_name = ", ".join(m if not isinstance(m, str) else [m])
+        elif hasattr(pass_instance, "name"):
+            pass_name = pass_instance.name
+
         cache[pass_level] = (
             result,
             pass_name if pass_level else "Before MLIR Passes",
```
andrijapau:
```suggestion
            pass_name if pass_level else "Before MLIR Passes (MLIR-0)",
```

frontend/catalyst/python_interface/inspection/specs.py
```diff
@@ -83,7 +91,12 @@ def _specs_callback(previous_pass, module, next_pass, pass_level=0):
         pass_instance = previous_pass if previous_pass else next_pass
         result = specs_collect(module)
 
-        pass_name = pass_instance.name if hasattr(pass_instance, "name") else pass_instance
+        pass_name = pass_instance
+        if m := level_to_markers.get(pass_level):
+            pass_name = ", ".join(m if not isinstance(m, str) else [m])
+        elif hasattr(pass_instance, "name"):
+            pass_name = pass_instance.name
```
andrijapau:
```suggestion
            pass_name = pass_instance.name + f"(MLIR {pass_level})"
```

frontend/catalyst/python_interface/inspection/specs.py
andrijapau:
```suggestion
        return {f"{cache[lvl][1]}": cache[lvl][0] for lvl in level if lvl in cache}
```

frontend/catalyst/python_interface/inspection/specs.py
```diff
@@ -49,6 +54,9 @@ def mlir_specs(
           specified level
     """
 
+    if level_to_markers is None:
+        level_to_markers = defaultdict(tuple)
```
JerryChen97:
```suggestion
    level_to_markers = defaultdict(tuple) if level_to_markers is None
```
wish this makes codefac happy...

frontend/catalyst/python_interface/inspection/specs.py
```diff
@@ -101,15 +114,15 @@ def _specs_callback(previous_pass, module, next_pass, pass_level=0):
         pass
 
     if level == "all":
-        return {f"{cache[lvl][1]} (MLIR-{lvl})": cache[lvl][0] for lvl in sorted(cache.keys())}
+        return {f"{cache[lvl][1]}": cache[lvl][0] for lvl in sorted(cache.keys())}
```
jzaia18:
```suggestion
        return {cache_item[1]: cache_item[0] for cache_item in cache.values()}
```

I think this should just work, now that we don't need the level itself in the output string here. I'm also not confident that the sort was doing anything, the cache should always be filled in the correct order

frontend/catalyst/python_interface/inspection/specs.py
```diff
@@ -101,15 +114,15 @@ def _specs_callback(previous_pass, module, next_pass, pass_level=0):
...
 
     if isinstance(level, (tuple, list)):
         if any(lvl not in cache for lvl in level):
             missing = [str(lvl) for lvl in level if lvl not in cache]
             raise ValueError(
                 f"Requested specs levels {', '.join(missing)} not found in MLIR pass list."
             )
-        return {f"{cache[lvl][1]} (MLIR-{lvl})": cache[lvl][0] for lvl in level if lvl in cache}
+        return {f"{cache[lvl][1]}": cache[lvl][0] for lvl in level if lvl in cache}
```
jzaia18, line 117: Same suggestion as above

frontend/test/pytest/test_specs.py
jzaia18, line 1: Can you add a test where we explicitly have a tape transform AND also an MLIR pass, and use markers on both passes?

frontend/catalyst/python_interface/inspection/specs.py
andrijapau: Hmm. Good point - let me give it a shot!

frontend/catalyst/python_interface/inspection/specs.py
andrijapau: I think this one is different because we're iterating over `level` rather than the cache?

frontend/test/pytest/test_specs.py
andrijapau: Good idea!

frontend/catalyst/python_interface/inspection/specs.py
andrijapau: I think would need an else condition too 😓

frontend/test/pytest/test_specs.py
andrijapau: Although, how do I do that? :)

andrijapau: Done here: https://github.com/PennyLaneAI/catalyst/pull/2464/commits/315fa791a9ea618ae6b6a82e15458abce8bbe965

JerryChen97 (approved):

.github/workflows/check-catalyst.yaml
```diff
@@ -476,6 +476,10 @@ jobs:
         # Install graphviz for testing the mlir-op-graph integration
         sudo apt-get install -y graphviz
         make frontend
+        
+    - name: Install PennyLane branch
+      run: |
+        pip install --no-deps --force git+https://github.com/PennyLaneAI/pennylane@fix-marker
```
jzaia18: Make sure to set these back to normal before merging

frontend/catalyst/python_interface/inspection/specs.py
```diff
@@ -83,10 +91,14 @@ def _specs_callback(previous_pass, module, next_pass, pass_level=0):
         pass_instance = previous_pass if previous_pass else next_pass
         result = specs_collect(module)
 
-        pass_name = pass_instance.name if hasattr(pass_instance, "name") else pass_instance
+        pass_name = str(pass_instance) + f" (MLIR-{pass_level})"
+        # Always prioritize marker label is it exists
```
jzaia18:
```suggestion
        # Always prioritize marker label if it exists
```

frontend/test/pytest/test_specs.py
```diff
@@ -280,14 +280,100 @@ def test_basic_passes_multi_level(self, simple_circuit):
             check_specs_header_same(actual, single_level_specs, skip_level=True)
             check_specs_resources_same(res, single_level_specs["resources"])
 
+    def test_marker_with_tape_and_mlir_transforms(self, simple_circuit):
```
jzaia18: Can you also add a test for this same circuit (with 2 tape transforms, 2 MLIR passes, and markers) with `level="all"`? Both to make sure its counting the levels correctly, and to make sure it displays with the markers.

Sorry to keep nitpicking this one so much, I just want to make sure we thoroughly cover all cases (especially since its easy to break this with future updates.

jzaia18 (commented):
Last few comments, looking good!

frontend/test/pytest/test_specs.py
andrijapau: No worries at all! Let me put that together

andrijapau: Done here: https://github.com/PennyLaneAI/catalyst/pull/2464/commits/dfa5130eddf059c76bc0ec126dd229537b64940d

I also took the liberty of making a test class for marker integration since the number of tests are starting to grow. :)

jzaia18 (approved):
Awesome, looking good!

.github/workflows/check-catalyst.yaml
andrijapau: Done!

PR #2465, opened 2026-02-05
fix: Update tests to new queuing behaviour

dwierichs:
**Context:**
In PennyLane PR [#8131](https://github.com/PennyLaneAI/pennylane/pull/8131), we changed queuing behaviour of some operators, in particular in operator math.
This breaks a small number of tests in the Catalyst frontend.

Update the tests to either expect the changed behaviour or to avoid the new queuing behaviour where it is unwanted (e.g. construct a static observable outside of a QNode, which can be considered the better practice imo).

Note that there is also a follow-up bugfix PR in PL itself:

**Benefits:**
Bug fix/compatibility.

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -54,6 +54,10 @@
 
 <h3>Bug fixes 🐛</h3>
 
+* Fix a bug caused by a change in queuing behaviour in PennyLane in 
+  [PL #8131](https://github.com/PennyLaneAI/pennylane/pull/8131)
+  [(#2408)](https://github.com/PennyLaneAI/catalyst/pull/2408)
+
 * Fix a bug with the xDSL `ParitySynth` pass that caused failure when the QNode being transformed
   contained operations with regions.
   [(#2408)](https://github.com/PennyLaneAI/catalyst/pull/2408)
@@ -177,4 +181,5 @@ Mudit Pandey,
 Andrija Paurevic,
 David D.W. Ren,
 Paul Haochen Wang,
+David Wierichs,
 Jake Zaia.
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -54,10 +54,6 @@
 
 <h3>Bug fixes 🐛</h3>
 
-* Fix a bug caused by a change in queuing behaviour in PennyLane in 
-  [PL #8131](https://github.com/PennyLaneAI/pennylane/pull/8131)
-  [(#2408)](https://github.com/PennyLaneAI/catalyst/pull/2408)
-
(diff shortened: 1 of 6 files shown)


mlir/llvm-project
dime10, line 1: I'm guessing you haven't updated to after our llvm upgrade, not a big deal, these three files just need to be reverted

doc/releases/changelog-dev.md
```diff
@@ -54,6 +54,10 @@
 
 <h3>Bug fixes 🐛</h3>
 
+* Fix a bug caused by a change in queuing behaviour in PennyLane in 
+  [PL #8131](https://github.com/PennyLaneAI/pennylane/pull/8131)
+  [(#2408)](https://github.com/PennyLaneAI/catalyst/pull/2408)
+
```
dime10: No code changes in the Catalyst package right (only tests)? In that case we don't put a changelog entry for the release notes
```suggestion
```

frontend/test/pytest/test_measurements_results.py
```diff
@@ -426,29 +426,26 @@ def expval(x: float):
...
-            obs_matrix = np.array(
-                [
-                    [0.5, 1.0j, 0.0, -3j],
-                    [-1.0j, -1.1, 0.0, -0.1],
-                    [0.0, 0.0, -0.9, 12.0],
-                    [3j, -0.1, 12.0, 0.0],
-                ]
-            )
-
-            obs = qml.Hermitian(obs_matrix, wires=[0, 1])
```
dime10, line 445: In this test the constructed observable is used, so this should be a legitimate usage pattern right (if I understand your slack explanation correctly)?

dwierichs: Ah yes, this is perfectly valid. But leaving this as is makes this PR blocked by [this PL PR](https://github.com/PennyLaneAI/pennylane/pull/9029) 🙃

mlir/llvm-project
dwierichs: ✅

frontend/test/pytest/test_measurements_results.py
dime10: Ah fair enough. My only concern is that this test seems to have added value in its current form since it covers the issue in question? Simplifying it with your change means if something of the sort breaks again we won't know, if I'm interpreting things correctly?

dwierichs: I think this is something that should be caught by PL tests, not by Catalyst tests. I added corresponding tests for `LinearCombination` [here](https://github.com/PennyLaneAI/pennylane/pull/9029) :)

JerryChen97 (approved):

albi3ro (approved):
Thanks!

PR #2467, opened 2026-02-06
Fix bug in PCPhase bind

comp-phys-marc:
**Context:** We are trying to see if we can run certain circuits with capture, qjit and decomposition. At the moment, this is limited by an error arising from `jax_primitives` due to a workaround which has thus far been left in place in the bind function of `PCPhase`.

**Description of the Change:** We remove the workaround and match the signature of the `PCPhase` to what is expected in `jax_primitives`.

**Benefits:** We are closer to using `PCPhase` in circuits with capture, qjit and decomps.

**Possible Drawbacks:** Not sure yet how this change will propagate.

**Related ShortCut Stories:** [sc-101673] [sc-101776]


JerryChen97 (changes requested):
Might need sync the submodule so that we dont touch enzyme etc via this pr🤔

doc/releases/changelog-dev.md
```diff
@@ -106,6 +106,10 @@
 
 <h3>Bug fixes 🐛</h3>
 
+* Fix a bug in the bind call function for `PCPhase` where the signature did not match what was 
+  expected in `jax_primitives`.
```
JerryChen97:
```suggestion
  expected in `jax_primitives`. `ctrl_qubits` was missing from positional arguments in previous signature.
```
Let's be more specific about this fix!

frontend/catalyst/from_plxpr/qfunc_interpreter.py
```diff
@@ -355,11 +355,13 @@ def _pcphase_bind_call(*invals, op, qubits_len, params_len, ctrl_len, adjoint, h
     params_len += 1
 
     ctrl_inputs = invals[qubits_len + 2 :]
+    ctrl_wires = invals[qubits_len + 1 : -len(ctrl_inputs)]
```
JerryChen97, line 357: Why `ctrl_inputs` starts from `qubits_len + 2` while `ctrl_wires` starts from `qubits_len + 1` til `-len(ctrl_inputs)`? Considering that, after `ctrl_inputs` is defined, `-len(ctrl_inputs)` is exactly `qubits_len + 2`, I'll say this implies that `ctrl_wires` is literally the single index `qubits_len + 1`?
Anyway this relationship looks weird to me; haven't checked further details, but it feels like a bug...

JerryChen97 (commented):
Also mind add a minimal test somewhere? Not sure if this is easy to do; curious why Catalyst's tests never caught this obvious bug...

frontend/catalyst/from_plxpr/qfunc_interpreter.py
comp-phys-marc: After the qubits_len we take an `angle`. After that comes the `ctrl_wires` and finally the `ctrl_inputs`. It indeed seems to imply that the `ctrl_wires` are exactly one index, and in the case I have seen it is in fact true. This bind function is therefore written assuming that there is only one control wire, prior to this PR.

comp-phys-marc: @maliasadi can you please confirm whether there is a good reason to expect a single control here?

maliasadi: Yes, Looking at `jax_primitives.py` (line 1330-1331)

``` python
ctrl_qubits = qubits_or_params[qubits_len + params_len : qubits_len + params_len + ctrl_len]
ctrl_values = qubits_or_params[qubits_len + params_len + ctrl_len :]
```

With `params_len += 1` (now 2 after adding `dim`), the bind call `[*wires, angle, dim, *ctrl_wires, *ctrl_inputs]` will be correctly parsed.

frontend/catalyst/from_plxpr/qfunc_interpreter.py
```diff
@@ -355,11 +355,13 @@ def _pcphase_bind_call(*invals, op, qubits_len, params_len, ctrl_len, adjoint, h
...
 
     return qinst_p.bind(
         *wires,
         angle,
         dim,
+        *ctrl_wires,
```
maliasadi, line 363: Could you add tests in `frontend/tests/pytest/from_plxpr`?

comp-phys-marc: I don't see any other tests there for `PCPhase` or `Controlled`. Where should I be adding a test and what should it look like?

paul0403: You just want a circuit execution test containing a pc phase right? You can add it to `test_capture_integration.py`, there are a bunch of circuit tests there.

paul0403: Remember to check the non trivial cases, e.g. PC phase in a control, adjoint, loop, etc, although these shouldn't need their own changes other than the basic pc phase bind update here.

comp-phys-marc: Thanks, I added `PCPhase` to the basic circuit test as well as the `Controlled` tests since my change is about the control wires being handled properly.

paul0403 (approved):

JerryChen97 (approved):
Nice!

frontend/test/pytest/from_plxpr/test_capture_integration.py
paul0403, line 114: So, the `backend` argument here is a string determining the device. It will be one of `"lightning.qubit"` and `"lightning.kokkos"`. We can xfail when on kokkos.

But, since we don't want to xfail the entire original test, let's make 3 new tests, with the circuit just containing a pcphase gate 🙂

comp-phys-marc: Would 2 tests make sense, one for PCPhase and another with the original circuit?

frontend/test/pytest/from_plxpr/test_capture_integration.py
```diff
@@ -348,6 +350,37 @@ def circuit(theta):
...
+        capture_result = captured_circuit(theta)
+
+        qml.capture.disable()
+
+        # Capture disabled
+
+        @qml.qnode(device)
+        def circuit(theta):
+            qml.ctrl(qml.PCPhase, control=[1], control_values=[False])(theta, 2, wires=[0])
+            return qml.state()
```
paul0403:
```suggestion
        @qml.qnode(device)
        def circuit(theta):
            qml.ctrl(qml.PCPhase, control=[1], control_values=[False])(theta, 2, wires=[0])
            return qml.state()
         
        capture_result = qjit(capture=True)(theta)
```
we can just do this right?

comp-phys-marc: Sure

frontend/test/pytest/from_plxpr/test_capture_integration.py
```diff
@@ -348,6 +350,33 @@ def circuit(theta):
...
 
+    @pytest.mark.parametrize("theta", (jnp.pi, 0.1, 0.0))
+    def test_ctrl_pcphase(self, backend, theta):
+        """Test the integration for a PCPhase circuit with control."""
+        if backend == "lightning.kokkos":
+            pytest.xfail(reason="PCPhase not yet implemented on Kokkos.")
+
+        device = qml.device(backend, wires=3)
+
+        # Capture enabled
```
paul0403, line 361: I guess let's add a `qml.capture.enable()` here just to stay in line with the rest of this test file. We can deal with the `qjit(capture=True)` migration later.

frontend/test/pytest/from_plxpr/test_capture_integration.py
```diff
@@ -348,6 +350,31 @@ def circuit(theta):
 
         assert jnp.allclose(capture_result, circuit(theta))
 
+    @pytest.mark.parametrize("theta", (jnp.pi, 0.1, 0.0))
+    def test_ctrl_pcphase(self, backend, theta):
+        """Test the integration for a PCPhase circuit with control."""
+        if backend == "lightning.kokkos":
+            pytest.xfail(reason="PCPhase not yet implemented on Kokkos.")
```
paul0403:
```suggestion
            pytest.xfail(reason="Controlled PCPhase not yet implemented on Kokkos.")
```

paul0403 (approved):

PR #2468, opened 2026-02-06
Patch JAX's implicitly added return outvar for dynamic shapes

paul0403:
**Context:**
There was a mysterious failure when all of the following conditions are met: 
- use capture
- apply a pass
- returns `qml.sample()` from a qnode
- use dynamic number of shots

When these conditions are met, JAX leaks an argument from the inner jaxpr of the pennylane transform primitive out into the returned outvar of the top level jaxpr, as the dynamically-shaped shot dimension for the samples.

See https://xanaduhq.slack.com/archives/C06CNPQLK2T/p1770325357111409   

```python
qml.capture.enable()

@qjit
def workflow(shots:int):
    @qml.transform(pass_name="cancel-inverses")
    @qml.qnode(dev, shots=shots)
    def aloha():
        qml.Hadamard(wires=0)
        return qml.sample()

    return aloha()

x = workflow(5)
print(x)
```

```
File "/home/paul.wang/catalyst_new/catalyst/frontend/catalyst/jit.py", line 735, in capture
    return trace_from_pennylane(
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/paul.wang/catalyst_new/catalyst/frontend/catalyst/from_plxpr/from_plxpr.py", line 536, in trace_from_pennylane
    jaxpr = from_plxpr(plxpr)(*plxpr.in_avals)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/paul.wang/catalyst_new/catalyst/frontend/catalyst/from_plxpr/from_plxpr.py", line 204, in wrapped_fn
    return jax.make_jaxpr(original_fn)(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/paul.wang/catalyst_new/catalyst/cat_env/lib/python3.12/site-packages/pennylane/capture/base_interpreter.py", line 389, in eval
    outval = self.read(var)
             ^^^^^^^^^^^^^^
  File "/home/paul.wang/catalyst_new/catalyst/cat_env/lib/python3.12/site-packages/pennylane/capture/base_interpreter.py", line 267, in read
    return var.val if isinstance(var, jax.extend.core.Literal) else self._env[var]
                                                                    ~~~~~~~~~^^^^^
KeyError: Var(id=134741294431552):int64[]
```

In jax/_src/interpreters/partial_eval.py/_add_implicit_outputs, the vars for the dynamic dimensions of a returned jaxpr value are added to the jaxpr outvars.
Starting from v0.7.1, the dynamic dimension is actually no longer necessary to return, judging from the test suite we have (including those that deal with dynamic shapes).
We just patch this method to not add any implicit outputs.

**Benefits:**
Things work.

**Possible Drawbacks:**
Might trainwreck dynamic shapes in the future, but even JAX itself kinda stopped caring about dynamic shapes.

### DIFF
--- a/frontend/catalyst/jax_extras/patches.py
+++ b/frontend/catalyst/jax_extras/patches.py
@@ -22,7 +22,15 @@ from functools import partial
 import jax
 import jax._src.interpreters.partial_eval as pe
 from jax._src import config, core, source_info_util
-from jax._src.core import JaxprEqnContext, abstractify, standard_vma_rule
+from jax._src.core import (
+    JaxprEqnContext,
+    abstractify,
+    standard_vma_rule,
+    InDBIdx,
+    OutDBIdx,
+    Var,
+    DShapedArray,
+)
 from jax._src.interpreters import mlir
 from jax._src.interpreters.partial_eval import (
     DynamicJaxprTracer,
@@ -56,6 +64,7 @@ __all__ = (
     "patched_dyn_shape_staging_rule",
     "patched_pjit_staging_rule",
     "patched_multi_broadcast_in_dim",
+    "patched_add_implicit_outputs",
 )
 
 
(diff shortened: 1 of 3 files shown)


paul0403:
That being said, at the same time this issue smells oddly similar to this other dynamic shape issue in JAX (just JAX, no PL whatsoever) from a year ago, where also the extra implicit outvar caused some issues. https://github.com/jax-ml/jax/issues/27877

paul0403:
I don't know why, but after updating to pl version `0.45.0.dev24` this bug went away on its own.

PR #2470, opened 2026-02-09
Graph Decomp + qjit with work wires

kipawaa:
**Context:**
Decomposition rules are captured and compiled, but parameters like `num_work_wires` aren't accepted with qjit.

Adds the `num_work_wires` keyword to the decomposition interpreter in Catalyst, allowing work wire specification to be passed to the usual graph decomposition system.
Also adds the `op_type` parameter to the `@decomposition_rule` decorator, allowing decomposition rules to be manually targeted for decomposition testing.

**Benefits:**
Improves feature parity of graph decomposition vs graph decompositions + qjit.

**Possible Drawbacks:**
Multiple UIs now exist for decomposition.

**Related GitHub Issues:**
N/A

[sc-110999]


frontend/catalyst/jax_primitives.py
```diff
@@ -428,23 +427,89 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
-def decomposition_rule(func=None, *, is_qreg=True, num_params=0, pauli_word=None):
+def decomposition_rule(func=None, *, is_qreg=True, num_params=0, pauli_word=None, op_type=None):
```
maliasadi, line 431: You added an `op_type` parameter to the `@decomposition_rule` decorator which is to set as a `target_gate` attribute on the function. This shouldn't be required when capturing and compiling these rules using [this function](https://github.com/PennyLaneAI/catalyst/blob/bed8068e59db28d98352f5df6c4767480a0cf56f/frontend/catalyst/from_plxpr/decompose.py#L255). `_create_decomposition_rule` sets `target_gate` in the decomposition system with capture and graph.

maliasadi: I don't have any problems with keeping this as it seems useful for testing! But let's get back to this when we integrate/test this work with `qml.transform.decompose`

frontend/test/lit/test_decomposition.py
maliasadi, line 1: Nice! This shows that we can compile decomposition rules with `qp.allocate` down to MLIR and apply `decompose-lowering` 👍 

Could you test the support with `qml.transform.decompose`? You can see `qp.allocate` mlir capture tests  [here](https://github.com/PennyLaneAI/catalyst/blob/bed8068e59db28d98352f5df6c4767480a0cf56f/frontend/test/lit/test_decomposition.py#L1228) where you can add more tests and also check `mlir_opt`.

maliasadi (changes requested):

frontend/test/lit/test_decomposition.py
kipawaa: added a simple example in [41e571c](https://github.com/PennyLaneAI/catalyst/pull/2470/commits/41e571c5de074a011b851feb7e9aa6d0990b696f), expanded with `num_work_wires` in [bb8466f](https://github.com/PennyLaneAI/catalyst/pull/2470/commits/bb8466f2ca3bbd2665f5d1f17370e2ffd64196e4) and [8394436](https://github.com/PennyLaneAI/catalyst/pull/2470/commits/8394436360abda30a12348babf791f6c104c7361)

frontend/catalyst/jax_primitives.py
kipawaa: I wasn't sure initially how we wanted to trigger this, and I noticed that in some contexts the attribute wasn't set (running without graph). I wasn't sure if we wanted to allow lowering of these rules in those contexts, and it made checking simpler. Let me know if you want it removed though, I realize now that with graph enabled by default this will be unnecessary!

doc/releases/changelog-dev.md
```diff
@@ -94,6 +94,10 @@
   `shots` aren't set. Instead, an informative error message is raised.
   [(#2456)](https://github.com/PennyLaneAI/catalyst/pull/2456)
 
+* Added the `op_type` parameter to the `@decomposition_rule` decorator, allowing the `decompose-lowering` pass to
+  register and apply decomposition rules in MLIR.
+  [(#2470)](https://github.com/PennyLaneAI/catalyst/pull/2470)
```
astralcai: Can we update this as well?

frontend/test/lit/test_decomposition.py
```diff
@@ -1312,3 +1312,293 @@ def f(i):  # pylint: disable=unused-argument
...
+
+    qml.capture.enable()
+
+    @decomposition_rule(is_qreg=True, op_type="PauliZ")
+    def my_decomp(wires):
+        with qml.allocate(2, restored=False) as work_wires:
+            qml.X(wires[0])
+            qml.X(wires[1])
+            qml.H(work_wires[0])
+            qml.H(work_wires[1])
```
astralcai, line 1331: It's nice that it works almost out of the box like that.

doc/releases/changelog-dev.md
kipawaa: Fixed in [97c09cc](https://github.com/PennyLaneAI/catalyst/pull/2470/commits/97c09cc8eaf915bff820f6bdb1729fb3b2cd5b9b). Thanks!

frontend/test/lit/test_decomposition.py
```diff
@@ -1312,3 +1312,329 @@ def f(i):  # pylint: disable=unused-argument
...
+        qml.cond(angle > 1.2, true_func, false_func)()
+
+    @qml.qjit
+    @partial(
+        qml.transforms.decompose,
+        gate_set={qml.CNOT, qml.H, qml.X, "Conditional", "MidMeasure"},
+        fixed_decomps={qml.CRX: my_decomp},
+        num_work_wires=3,
+    )
+    @qml.qnode(qml.device("lightning.qubit", wires=2))
```
astralcai: How would you have three work wires when the device only has two wires in total?

kipawaa: Am I wrong to assume that `qp.allocate` allocates a new register with new wires? That was my impression

frontend/catalyst/jax_primitives.py
```diff
@@ -428,23 +428,89 @@ def wrapper(*args, **kwargs):
...
     """
     Denotes the creation of a quantum definition in the intermediate representation.
+
+    Args:
+        func (Callable): the subroutine to apply in place of the replaced gate.
+        is_qreg (bool): ???
+        num_params (int): ???
+        pauli_word (???): ???
```
kipawaa, line 439: I could use help filling in these values. I haven't found documentation for them before, and while I have an intuitive understanding, I'd like to have something more concrete before writing it down here 😅

frontend/test/lit/test_decomposition.py
astralcai: `qp.allocate` does allocate a new register with new wires, but the question is where the wires come from. In PL, device wires is interpreted as the pool of all available wires, and then the circuit may act on a subset of those wires, the rest of the wires (device wires minus circuit wires) is the pool of available work wires. When `qp.allocate` happens, we pick a wire (any wire) from that pool of unused device wires.

astralcai: Essentially, if this isn't allowed:
```python
import pennylane as qml

def circuit():
    qml.X(0)
    qml.Y(1)
    qml.Z(2)
    return qml.expval(qml.Z(0))
```
```pycon
>>> circuit()
WireError: Cannot run circuit(s) on lightning.qubit as they contain wires not found on the device: {2}
```
then this shouldn't be either:
```python
def circuit():
    qml.X(0)
    qml.Y(1)
    with qml.allocate(1) as aux_wires:
        qml.Z(aux_wires[0])
    return qml.expval(qml.Z(0))
```

kipawaa: Makes sense, thanks for the example! I've updated the device wires for a number of tests in [d563da5](https://github.com/PennyLaneAI/catalyst/pull/2470/commits/d563da5247b61166a32ef8a29fff189e24b0fd8b) and [39a10eb](https://github.com/PennyLaneAI/catalyst/pull/2470/commits/39a10eb9a11f58c6234e354b5179d6777258a205).

dime10: Yes, but this is a known difference between Catalyst and PennyLane. Dynamically allocated wires go beyond what is initially allocated by a device.

astralcai (commented):
I'm assuming that execution is out of scope here?

astralcai:
> > I'm assuming that execution is out of scope here?
> 
> @astralcai Yes, execution is out of scope here, though hopefully it can be added soon

In that case everything LGTM here, although I'm not sure I should be a blocking reviewer here since I go on vacation after tomorrow.

frontend/test/lit/test_decomposition.py
kipawaa: Thanks @dime10 for clarifying. @astralcai should I reset the number of device wires? Otherwise I'll resolve this

maliasadi (approved):
Thanks @kipawaa! Happy to approve

astralcai (approved):

frontend/catalyst/jax_primitives.py
kipawaa: Since work on graph decomp is still ongoing, and this documentation isn't user-facing, I'll just resolve this for now so we can move forward with other development.

PR #2471, opened 2026-02-09
Update xDSL Universe to use callables to return dialects and passes

mudit2812:
**Context:**
xDSL's `Universe` class expects the dialect and transform dictionaries used to initialize it to contain the following signature:
```python
dict[str, Callable[[], Dialect | ModulePass]]
```
I.e., the values are supposed to be callables with no arguments that return dialects or passes. The original implementation was incorrect, as the values were just the dialects or passes themselves. This PR fixes that.

* Update the implementation of `CATALYST_XDSL_UNIVERSE` to set up the dialect and transform dictionaries correctly
* Add a test that invokes `xdsl-opt` to verify that the universe is integrated with xDSL correctly.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-111182]


andrijapau (approved):
LGTM! Nice catch.

mehrdad2m (approved):

PR #2472, opened 2026-02-10
Graph decomposition for device and stopping_condition

lazypanda10117:
### Context:
Supporting device-specific gateset decomposition in the program capture flow. This is enabled by extending the `handle_qnode` function.  As stated in Ali's comment, the new `handle_qnode` supports the following cases:

1. User defined stopping_condition in qml.decompose: Use PLxPR decompose with user stopping condition!
2. No user decomposition (default): Load device capabilities --> device-specific gate set --> Use graph and decompose-lowering. Fallback to PLxPR when graph fails.
3. User specified gate_set : Graph-based decomposition on user gate set (the current support). Fallback to PLxPR when graph fails.

[sc-110277]
[sc-110276]
[sc-107559]

### DIFF
--- a/frontend/catalyst/from_plxpr/from_plxpr.py
+++ b/frontend/catalyst/from_plxpr/from_plxpr.py
@@ -294,17 +294,14 @@ def handle_qnode(
     )
     graph_succeeded = False
 
-    if (
-        stopping_condition := self.decompose_tkwargs.get("stopping_condition")
-        and self.requires_decompose_lowering
-    ):
+    stopping_condition = self.decompose_tkwargs.get("stopping_condition")
+    if stopping_condition is not None and self.requires_decompose_lowering:
         # print("Case 1: Using plxpr decompose for templates.")
         closed_jaxpr = _apply_compiler_decompose_to_plxpr(
             inner_jaxpr=qfunc_jaxpr,
             consts=consts,
             ncargs=non_const_args,
             tgateset=list(self.decompose_tkwargs.get("gate_set", [])),
-            stopping_condition=stopping_condition,
         )
 
         # Use the plxpr decompose transform and ignore graph decomposition
@@ -324,7 +321,6 @@ def handle_qnode(
             consts=consts,
             ncargs=non_const_args,
             tgateset=list(self.decompose_tkwargs.get("gate_set", [])),
-            stopping_condition=stopping_condition,
         )
 
         stopping_condition = None
@@ -705,7 +701,7 @@ def _apply_compiler_decompose_to_plxpr(
     if kwargs is None:
         kwargs = {}
 
-    if stopping_condition:
+    if stopping_condition is not None:
         kwargs["stopping_condition"] = stopping_condition
 
     final_jaxpr = qml.transforms.decompose.plxpr_transform(inner_jaxpr, consts, (), kwargs, *ncargs)
--- a/frontend/test/pytest/test_peephole_optimizations.py
+++ b/frontend/test/pytest/test_peephole_optimizations.py
@@ -538,7 +538,6 @@ def test_merge_rotation_arbitrary_angle_ppr():
     assert ir_opt.count('pbc.ppr.arbitrary ["Z", "Y"]') == 1
 
 
-@pytest.mark.xfail(reason="PPM execution with ppr-to-ppm pass is not fully supported yet.")
 @pytest.mark.usefixtures("use_capture")
 def test_clifford_to_ppm():
 


frontend/catalyst/device/capabilities_utils.py
```diff
@@ -0,0 +1,90 @@
+# Copyright 2025 Xanadu Quantum Technologies Inc.
```
maliasadi:
```suggestion
# Copyright 2026 Xanadu Quantum Technologies Inc.
```

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -281,20 +283,65 @@ def handle_qnode(
     consts = args[shots_len : n_consts + shots_len]
     non_const_args = args[shots_len + n_consts :]
 
-    closed_jaxpr = (
```
maliasadi, line 284: The new `handle_qnode` supports the following cases:

1. User defined `stopping_condition` in `qml.decompose`: Use PLxPR decompose with user stopping condition!
2. No user decomposition (default): Load device capabilities --> device-specific gate set --> Use graph and `decompose-lowering`
3. User specified `gate_set` :  Graph-based decomposition on user gate set (the current support)

cc: @astralcai @isaacdevlugt if you guys have any comments or suggestions

frontend/catalyst/from_plxpr/decompose_utils.py
```diff
@@ -0,0 +1,87 @@
...
+    device,
+    execution_config,
+    shots_len=None,
+) -> DeviceCapabilities:
+    """Load and filter device capabilities for decomposition/compilation."""
+    capabilities = load_device_capabilities(device)
+
+    if execution_config is None:
+        execution_config = qml.devices.ExecutionConfig()
+        execution_config = device.setup_execution_config
```
maliasadi: Fix the dead assignment here

frontend/catalyst/device/decomposition.py
```diff
@@ -228,9 +228,9 @@ def catalyst_acceptance(
             return match
 
     elif is_supported(op, capabilities):
-        return op.name
+        return True
```
maliasadi: `str` --> `bool`! You need to update the old system (w/o capture)

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -313,8 +360,11 @@ def handle_qnode(
                 consts=closed_jaxpr.consts,
                 ncargs=non_const_args,
                 tkwargs=self.decompose_tkwargs,
+                stopping_condition=stopping_condition,
             )
 
+    # print("Closed Jaxpr", closed_jaxpr)
```
maliasadi:
```suggestion
```

frontend/catalyst/from_plxpr/decompose_utils.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def catalyst_acceptance(op, capabilities, diff_method):
+    """Check if an operation is supported by the device."""
+    return _catalyst_acceptance(op, capabilities, diff_method)
+
+
+def calculate_diff_method(device, closed_jaxpr):
+    """Calculate the diff method for the device."""
+    return _calculate_diff_method(device, closed_jaxpr)
```
maliasadi, line 42: is this correct? 🤔 

We don't have any logic to deal with "diff" ops in the decomposition system

frontend/catalyst/device/capabilities_utils.py
maliasadi, line 1: I don't see the benefits of this file tbh where we still need to call `decomposition.py` methods in `from_plxpr`.

frontend/catalyst/from_plxpr/decompose_utils.py
maliasadi, line 1: Could we merge the util methods with the one in ./catalyst/device or add these methods to `from_plxpr/decompose.py` ?

maliasadi (changes requested):
Thanks @lazypanda10117! The main remaining **TODO** is to report/fix the edge cases:

1. `Counts` Not Implemented with Capture <- this issue is a program capture limitation and doesn't have anything to do with decomposition

1. Almost all xDSL / PPR parse errors are related to producing a different IR as of applying decomposition! We can fix most of them by updating the MLIR output to be compatible with the checks.
 
1. HybridOp decomp assertion <-- we should be able to fix all these issues by ensureing that HybridOps are in the stopping _condition OR excluded from decomposition.
 
1. Ops missing from device gateset <- Lightning/null devices support more gates than Catalyst can represent! You need to take the intersection of these ops and decompose the remaining ones in PLxPR.
    - `KeyError: 'StatePrep'`
    - `KeyError: 'BasisState'`
    - `KeyError: 'QubitUnitary'`
    - `KeyError: 'QFT'`

1. `_OperatorNode Can't Be Captured Without Wire Count` <- the graph decomposition produces `_OperatorNode` objects that require a wire count to be captured/lowered into MLIR, but the wire count isn't available at graph construction time. cc: @astralcai 

1. In specs/pass counting/circuit drawing tests: the current pipeline adds extra `decompose-lowering` pass to the compiled module. We need to update the logic (and maybe tests) to avoid this.

1. there are some other errors such as `DID NOT RAISE/WARN` or `Cannot resolve wire for SliceOp` or `TracerBoolConversionError`; let's get back to them after fixing the above as some of them are tied with those edge cases.

frontend/catalyst/device/decomposition.py
lazypanda10117: oops, my bad, i wasn't planning to push this change

frontend/catalyst/from_plxpr/decompose_utils.py
lazypanda10117: Yes we can, I just think this is cleaner since this is the subset of functions (with slight modifications) that are needed from device capabilities.

frontend/catalyst/device/capabilities_utils.py
lazypanda10117: I wanted to pick out the functions that are relevant to capabilities since those are the only "useful" functions in `from_plxpr`. I don't want `from_plxpr` to have explicit dependence on files in the `device/` folder.

frontend/catalyst/from_plxpr/decompose_utils.py
lazypanda10117: This is used in catalyst acceptance. I am just porting over the function.

frontend/catalyst/device/qjit_device.py
```diff
@@ -636,12 +597,5 @@ def check_device_wires(wires):
...
-    Checks if a device capabilities requires shots.
-
-    A device requires shots if all of its MPs are finite shots only.
-    """
-    return all(
-        ExecutionCondition.FINITE_SHOTS_ONLY in MP_conditions
-        for _, MP_conditions in capabilities.measurement_processes.items()
-    )
+    """Checks if a device capabilities requires shots."""
+    return _requires_shots_impl(capabilities)
```
maliasadi: we don't really need to split impl from defs in these helper functions.

frontend/catalyst/device/capabilities_utils.py
maliasadi: Could you revert the changes in the `device/` subdir? We don't really need any of these changes in `device/`. This is also not helping us to avoid explicit dep on files in this subdir.

lazypanda10117: Done!

frontend/catalyst/device/decomposition.py
lazypanda10117: Reverted.

frontend/catalyst/from_plxpr/from_plxpr.py
astralcai: If a user applies a `@qml.decompose` on top of a qnode, it would be applied in addition to the decomposition to the device gate set right? That is the current behaviour at least. Whether the user has a `@qml.decompose` or not does not change the fact that a decomposition to the device gate set is always going to happen. Is that not the case here?

lazypanda10117: >If a user applies a @qml.decompose on top of a qnode, it would be applied in addition to the decomposition to the device gate set right?

My understanding is that when a user specifies a transform, we will only use that user-specified gateset. We only do device-level decomposition when no such user-defined transform is provided.

PR #2474, opened 2026-02-10
Test audit/story 1 control flow

JerryChen97:
**Context:**
Test Audit: to reveal those remaining incompatibility of capture + qjit.

Scope/Files to audit:
 - [x] test_loops.py

The following have already been covered by another fixture (a global one at the file level). Since our goal here is to collect the information about what's missing for capture + catalyst environment, instead of a thorough auditing, it's safe to skip them.
 - [x] test_conditionals.py
 - [x] test_quantum_control.py
 - [x] test_switch.py (do we have qml.switch? no not yet)

Apply `capture_mode` fixture to those tests who:
1. not already scoped by e.g. `disable_capture`
2. indeed called `qjit`

Adjustments:
1. some tests used to return `measure()` directly. Since they were only designed to test loops, we test same thing with `expval` now.
2. `TestWhileLoopOperatorAccess` and `TestForLoopOperatorAccess` were not actually changed. They were adjusted only to keep consistent with the convenient `while_loop` `for_loop` auto dispatching in the test file.
3. Testing nested loops: 
> one test without the qnode to test the classical stuff, and then a second to test gates inside a qnode

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-111254]


frontend/test/pytest/test_loops.py
```diff
@@ -19,12 +19,38 @@
...
+    if qml.capture.enabled():
+        return qml.while_loop(*args, **kwargs)
+    return catalyst_while_loop(*args, **kwargs)
+
+
+def measure(*args, **kwargs):
+    """Use catalyst.measure always - qml.measure not yet compatible with qjit(capture=True)."""
+    if qml.capture.enabled():
+        return qml.measure(*args, **kwargs)
+    return catalyst_measure(*args, **kwargs)
```
JerryChen97: @albi3ro Do you think this hack is a good idea here? I was thinking this way when some day we finally get rid of all the `catalyst_**` we can easily get rid of these as well; but also feel unsafe since this is somewhat against our initial idea of toggling things locally for each test, which however involves modifying everyone to have a new inline dispatch inside testing method

frontend/test/pytest/test_loops.py
```diff
@@ -19,12 +19,38 @@
...
+    """Select qml.while_loop when capture is enabled, else catalyst.while_loop."""
+    if qml.capture.enabled():
+        return qml.while_loop(*args, **kwargs)
+    return catalyst_while_loop(*args, **kwargs)
+
+
+def measure(*args, **kwargs):
+    """Use catalyst.measure always - qml.measure not yet compatible with qjit(capture=True)."""
+    if qml.capture.enabled():
+        return qml.measure(*args, **kwargs)
```
andrijapau: I think `qml.measure` correctly dispatches to catalyst's measure when needed 🤔

JerryChen97: https://github.com/PennyLaneAI/catalyst/pull/2474/commits/9db66b6676a32b7352b281eb10a0f88efd1712c1

frontend/test/pytest/test_loops.py
JerryChen97: I see quite somewhere else we have been using the same trick! I'll just take it for granted lol

PR #2478, opened 2026-02-11
Add libmlir_apfloat_wrappers to Catalyst wheels

maliasadi:
**Context:**
:heavy_check_mark: : checked [the suite](https://github.com/PennyLaneAI/pennylane-benchmarks/actions/runs/21913644497/job/63284238238#step:25:131).

After upgrading to a newer version of LLVM #2415, Catalyst `make all wheels` and Linux wheels fail due to a missing so library:

```
self = <CDLL '/tmp/circuit_qpe62labp8b/circuit_qpe.so', handle 0 at 0x7fbac5827690>
name = '/tmp/circuit_qpe62labp8b/circuit_qpe.so', mode = 0, handle = None
use_errno = False, use_last_error = False, winmode = None

    def __init__(self, name, mode=DEFAULT_MODE, handle=None,
                 use_errno=False,
                 use_last_error=False,
                 winmode=None):
        self._name = name
        flags = self._func_flags_
        if use_errno:
            flags |= _FUNCFLAG_USE_ERRNO
        if use_last_error:
            flags |= _FUNCFLAG_USE_LASTERROR
        if _sys.platform.startswith("aix"):
            """When the name contains ".a(" and ends with ")",
               e.g., "libFOO.a(libFOO.so)" - this is taken to be an
               archive(member) syntax for dlopen(), and the mode is adjusted.
               Otherwise, name is presented to dlopen() as a file argument.
            """
            if name and name.endswith(")") and ".a(" in name:
                mode |= ( _os.RTLD_MEMBER | _os.RTLD_NOW )
        if _os.name == "nt":
            if winmode is not None:
                mode = winmode
            else:
                import nt
                mode = nt._LOAD_LIBRARY_SEARCH_DEFAULT_DIRS
                if '/' in name or '\\' in name:
                    self._name = nt._getfullpathname(self._name)
                    mode |= nt._LOAD_LIBRARY_SEARCH_DLL_LOAD_DIR
    
        class _FuncPtr(_CFuncPtr):
            _flags_ = flags
            _restype_ = self._func_restype_
        self._FuncPtr = _FuncPtr
    
        if handle is None:
>           self._handle = _dlopen(self._name, mode)
                           ^^^^^^^^^^^^^^^^^^^^^^^^^
E           OSError: libmlir_apfloat_wrappers.so.22.0git: cannot open shared object file: No such file or directory

```

**Benefits:**
- Fixes Catalyst 0.45-dev wheels 

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-111530]

### DIFF
--- a/frontend/test/pytest/conftest.py
+++ b/frontend/test/pytest/conftest.py
@@ -84,6 +84,33 @@ def use_both_frontend(request):
         yield
 
 
+@pytest.fixture(params=[True, False], ids=["capture=True", "capture=False"])
+def capture_mode(request):
+    """Parametrize tests to run with capture=True and capture=False.
+
+    This fixture returns a boolean that should be passed to @qjit(capture=...).
+    Unlike use_both_frontend, this does NOT toggle the global capture state,
+    allowing more isolated and explicit testing.
+
+    Usage:
+        def test_example(backend, capture_mode):
+            @qjit(capture=capture_mode)
+            @qml.qnode(qml.device(backend, wires=1))
+            def circuit():
+                ...
+
+    Markers:
+        @pytest.mark.old_frontend - Skip when capture_mode=True
+        @pytest.mark.capture_todo - xfail when capture_mode=True
+    """
+    if request.param:  # capture=True
+        if "old_frontend" in request.keywords:
+            pytest.skip("Test is specific to the old frontend and should not run with capture.")
+        if "capture_todo" in request.keywords:
+            pytest.xfail("Not expected to work yet with program capture.")
+    return request.param
+
+
 def pytest_collection_modifyitems(items, config):  # pylint: disable=unused-argument
     """Modify collected items as needed."""
     xdsl_tests_skipped = "not xdsl" in config.getoption("markexpr")


Makefile
```diff
@@ -221,6 +221,7 @@ wheel:
 	cp $(OQC_BUILD_DIR)/oqc_python_module.so $(MK_DIR)/frontend/catalyst/lib
 	cp $(OQC_BUILD_DIR)/backend/*.toml $(MK_DIR)/frontend/catalyst/lib/backend
 	cp $(COPY_FLAGS) $(LLVM_BUILD_DIR)/lib/libmlir_float16_utils.* $(MK_DIR)/frontend/catalyst/lib
+	cp $(COPY_FLAGS) $(LLVM_BUILD_DIR)/lib/libmlir_apfloat_wrappers.* $(MK_DIR)/frontend/catalyst/lib || true # optional
```
mehrdad2m, line 224: Is this a new library that they are shipping or it is just being used now? Why is it optional?

paul0403: It's a new library, so if we don't mark it optional then benchmark runs that build catalyst from source before the llvm update will fail, I guess that's the reason?

maliasadi: This is optional as it's not required on macOS wheels 
https://github.com/PennyLaneAI/catalyst/actions/runs/21912127908/job/63269468520?pr=2478

PR #2479, opened 2026-02-11
Add Catalyst resource tracker analysis pass

sengthai:
**Context:**
Currently, Pennylane resource tracker (`qml.specs`) counts operations across tapes, MLIR (xDSL), and runtime (in nullqubit device). However, it cannot pass the result to another pass in MLIR, nor can it specify the individual pass to be tracked.

This analysis pass will explicitly count operations in MLIR (C++) and can input the results into other passes following qml.specs.  

Add a MLIR resource tracker with three components: `ResourceResult` (mirror Python Impl.), `ResourceAnalysis`, and `ResourceTrackerPass`

**Benefits:**
- Other MLIR passes can query cached resource counts via `getAnalysis<ResourceAnalysis>()`.
- Integrates with MLIR's `-mlir-pass-statistics` for pipeline-level resource inspection.

**Usage:**
  ```bash
  quantum-opt --resource-tracker='output-json=true' input.mlir
  quantum-opt --resource-tracker -mlir-pass-statistics input.mlir
  ```

**Stories**
- [x] Implement Resource Tracker on analysis pass [[sc-110867]]
- [x] Handle control flow logic in Resource Tracker [[sc-110868]]


mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -0,0 +1,426 @@
...
+//===----------------------------------------------------------------------===//
+
+static int64_t getIntFromConstant(arith::ConstantOp op)
+{
+    assert(isa<IntegerAttr>(op.getValue()));
+    return cast<IntegerAttr>(op.getValue()).getValue().getSExtValue();
+}
+
+/// Count iterations of a scf.for loop. Returns -1 if dynamic bounds.
+static int64_t countForLoopIterations(scf::ForOp forOp)
```
rniczh: Since you are operating on the `scf:ForOp`, there are some useful APIs in mlir can help you to calculate the static iteartions.

If you look into the mlir passes and see how mlir optimize the trivial loop like with 0 iteartion or 1 iteartion. You will find a rewrite pattern: `SimplifyTrivialLoops`. It's using `op.getStaticTripCount();` (where op is `scf::ForOp`) to get the number of (static) iteartions for the current loop. It returns `-1` if it's dynamic, otherwise 0 for 0 iteration, 1 for 1 iteartion, ...etc.

Here, in your implemenation, if you want to count the number of loop iteartions, this should be the best fit here. Of course, another choice is to use our implemenation in `SCFUtil.cpp`, there is a `int64_t countStaticForloopIterations(Operation *op)` function can let you walk through the loops.

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -0,0 +1,426 @@
...
+            analyzeRegion(region, result, /*adjointMode=*/false);
+        }
+        funcResults[funcOp.getName()] = std::move(result);
+
+        if (funcOp->hasAttr("qnode")) {
+            entryFunc = funcOp.getName();
+        }
+    });
+
+    entryFuncName = entryFunc.str();
```
rniczh, line 183: what if there are multiple qnode functions in the module? for example, `split-non-commuting` will duplicate the qnode function for non-commuting observable.

rniczh (commented):
Thanks @sengthai :)

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: Thanks @rniczh! Good to know about `SimplifyTrivialLoops`! it means the `countStaticForloopIterations` has the same functionality with `op.getStaticTripCount()`?

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: Good point! here we assume the main/entry func has qnode attribute. So, how do we know which func among qnode function is the entry func?

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
rniczh: Not exactly the same, the one in upstream is way more complete than what we have implemented on our end. It can match a value from a constant foldable operation, and handle symbolic bounds. So from my perspective, I will suggest to use the upstream one. But it's depend on you.

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
rniczh: No XD  so in `jit.py`, it assumes the first function is entry point
```cpp
# WARNING: assumption is that the first function is the entry point to the compiled program.
entry_point_func = self.mlir_module.body.operations[0]
PPrestype = entry_point_func.type.results
```

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: Great API! i'm gonna use `getStaticTripCount`. I think we should add `SimplifyTrivialLoops` in our pipeline.

sengthai: Done! 
[be4909f](https://github.com/PennyLaneAI/catalyst/pull/2479/commits/be4909fb76df2f1c5e7ac876b080eab4fbbc27cf)

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: I've just applied a simple solution. since the `op->walk` traverses in order, so we just take the first one! 
[5950888](https://github.com/PennyLaneAI/catalyst/pull/2479/commits/5950888d0130b5f20efd0e8e49ab9c1872cb2300)

rniczh: But probably it doesn't has qnode attribute on the entry function. This is the qnode definition: https://github.com/PennyLaneAI/catalyst/pull/2441#issuecomment-3843147065

sengthai: But what does Pennylane's program look like where the main function has no MPs?

sengthai: 🤔

rniczh: So it call another process and get back the result from them (those are marked as qnode)

rniczh: Probably no, but in mlir we could transform it into the strucat that the entry function calls the qnode function (with mp) to get the result back from them and do the post processing. That's how we did in `split-to-single-terms` for ftqc.

doc/releases/changelog-dev.md
```diff
@@ -61,6 +61,16 @@
   # [ARTIQ] Generated ELF: /path/to/circuit.elf
   ```
 
+* Added a scalable MLIR resource tracker analysis pass (`resource-tracker`) that counts quantum
```
rniczh, line 141: Statements such as for loops and if conditions must be enclosed within curly brackets
e.g.
```cpp
if (...) {
    ...
}
```
instead of 
```cpp
if (...)
   ...
```

mlir/include/Catalyst/Analysis/ResourceAnalysis.h
```diff
@@ -0,0 +1,78 @@
...
+
+    // get the entry function name (the function marked with "qnode")
+    // returns empty string if no entry function was found
+    llvm::StringRef getEntryFunc() const { return entryFuncName; }
+
+    // MLIR analysis invalidation hook.
+    bool isInvalidated(const mlir::AnalysisManager::PreservedAnalyses &pa)
+    {
+        return !pa.isPreserved<ResourceAnalysis>();
+    }
```
rniczh: I think this is redundant, in `catalyst/mlir/llvm-project/mlir/include/mlir/Pass/AnalysisManager.h`, It's already the default behaviour as what you did here.
```cpp
template <typename AnalysisT>
bool isInvalidated(AnalysisT &analysis, const PreservedAnalyses &pa) {
  if constexpr (llvm::is_detected<has_is_invalidated, AnalysisT>::value)
    /// Implementation of 'isInvalidated' if the analysis provides a definition.
    return analysis.isInvalidated(pa);
  else
    /// Default implementation of 'isInvalidated'.
    return !pa.isPreserved<AnalysisT>();
}
```

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: Now we just take the first function of the program as the entry point.
[900253f](https://github.com/PennyLaneAI/catalyst/pull/2479/commits/900253f72d8e1cbf9638a4341716cca3f2845f0e)

doc/releases/changelog-dev.md
sengthai: [900253f](https://github.com/PennyLaneAI/catalyst/pull/2479/commits/900253f72d8e1cbf9638a4341716cca3f2845f0e)

mlir/include/Catalyst/Analysis/ResourceAnalysis.h
sengthai: good point! i removed that.
[1a2a483](https://github.com/PennyLaneAI/catalyst/pull/2479/commits/1a2a483d8e0d577d610514d5a06284d30a5561d7)

rniczh (approved):
🥳

mlir/include/Catalyst/Analysis/ResourceAnalysis.h
```diff
@@ -0,0 +1,72 @@
...
+
+#pragma once
+
+#include "llvm/ADT/StringMap.h"
+
+#include "PBC/IR/PBCOps.h"
+#include "mlir/Dialect/SCF/IR/SCF.h"
+#include "mlir/IR/Operation.h"
+
+#include "Catalyst/Analysis/ResourceResult.h"
```
paul0403, line 22: Note on include order: cpp first, then upstream, then us


```suggestion
#include "llvm/ADT/StringMap.h"

#include "mlir/Dialect/SCF/IR/SCF.h"
#include "mlir/IR/Operation.h"

#include "Catalyst/Analysis/ResourceResult.h"
#include "PBC/IR/PBCOps.h"
```

mlir/include/Catalyst/Analysis/ResourceResult.h
```diff
@@ -0,0 +1,61 @@
...
+
+#pragma once
+
+#include "llvm/ADT/DenseMap.h"
+#include "llvm/ADT/StringMap.h"
+
+namespace catalyst {
+
+// method for merging two ResourceResult values
+enum class MergeMethod { Sum, Max, Min };
```
paul0403: A bit scared about this somewhat generic name in the big `catalyst` namespace. Wanna rename it to `ResourceTrackerMergeMethod`? I'm also open to other names.

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
paul0403: Yes the one we have only works if the loop bounds are trivially static. We don't do things like SCCP in our pipeline. Although folding is also an option.... but if there's an upstream one, we should definitely use that!

mlir/lib/Catalyst/Analysis/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+    ResourceResult.cpp
+    ResourceAnalysis.cpp
+)
+
+get_property(dialect_libs GLOBAL PROPERTY MLIR_DIALECT_LIBS)
+set(LIBS
+    ${dialect_libs}
+    MLIRQuantum
+    MLIRPBC
+    MLIRMBQC
```
paul0403, line 13: I guess this means every new quantum-ish dialect (e.g. the reference semantics dialect) needs to manually opt into the resource tracking framework here. We unfortunately do not have a base interface for all quantum-ish ops across dialects. I don't even know if that's possible, since op interfaces in mlir seem to be only defined within a dialect...

Well, anyway, no action needed here.

mlir/lib/Catalyst/Analysis/ResourceResult.cpp
```diff
@@ -0,0 +1,158 @@
...
+{
+    switch (method) {
+    case MergeMethod::Max:
+        return std::max(a, b);
+    case MergeMethod::Min:
+        return std::min(a, b);
+    case MergeMethod::Sum:
+        return a + b;
+    }
+    return a + b; // unreachable, silences warning
```
paul0403:
```suggestion
    return -1; // unreachable, silences warning
```

Might as well return a meaningful error code if it's not supposed to be reached. 

Also, if you add a `default` to the `switch`, is the complaint still there?

mlir/lib/Catalyst/Analysis/ResourceResult.cpp
```diff
@@ -0,0 +1,158 @@
...
+void ResourceResult::multiplyByScalar(int64_t scalar)
+{
+    for (auto &opEntry : operations) {
+        for (auto &sizeEntry : opEntry.getValue()) {
+            sizeEntry.second *= scalar;
+        }
+    }
+
+    for (auto &entry : measurements) {
+        entry.getValue() *= scalar;
```
paul0403, line 89: I don't see this `getValue()` method on [`StringMap`'s documentation](https://llvm.org/doxygen/classllvm_1_1StringMap.html) 🤔 is this an internal method?

paul0403: Same with `getKey()` in the function above

mlir/lib/Catalyst/Analysis/ResourceResult.cpp
```diff
@@ -0,0 +1,158 @@
...
+        return std::max(a, b);
+    case MergeMethod::Min:
+        return std::min(a, b);
+    case MergeMethod::Sum:
+        return a + b;
+    }
+    return a + b; // unreachable, silences warning
+}
+
+void ResourceResult::mergeWith(const ResourceResult &other, MergeMethod method)
```
paul0403, line 42: Just wanna say I love this organization, it feels super clean 🤩

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -0,0 +1,414 @@
...
+//===----------------------------------------------------------------------===//
+// Operation categorization helpers
+//===----------------------------------------------------------------------===//
+
+/// Get the name to use for a quantum gate op.
+static std::string getGateOpName(Operation *op, bool isAdjoint)
+{
+    std::string name =
+        llvm::TypeSwitch<Operation *, std::string>(op)
+            .Case<quantum::CustomOp>([](auto customOp) { return customOp.getGateName().str(); })
```
paul0403, line 69: What about the other ops in the quantum dialect that don't go to custom op? e.g. pcphase, qubit unitary, ...

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -0,0 +1,414 @@
...
+    return llvm::TypeSwitch<Operation *, int>(op)
+        .Case<quantum::CustomOp>([](auto customOp) {
+            return static_cast<int>(customOp.getInQubits().size() +
+                                    customOp.getInCtrlQubits().size());
+        })
+        .Case<quantum::PauliRotOp>(
+            [](auto pauliRotOp) { return static_cast<int>(pauliRotOp.getInQubits().size()); })
+        .Case<quantum::MultiRZOp>(
+            [](auto multiRZOp) { return static_cast<int>(multiRZOp.getInQubits().size()); })
+        .Default(0);
```
paul0403: Same here, e.g. qubit unitary op would go to the default case here, and get counted as 0 qubits (which is wrong)!

I think we can use the [gate interface](https://github.com/PennyLaneAI/catalyst/blob/cce42944c26b7f6adb6afaa27d31cfa0d9887107/mlir/include/Quantum/IR/QuantumInterfaces.td#L107) here! Something like 
```C
.Case<quantum::QuantumGate>(
            [](auto gate) { return static_cast<int>(gate.getQubitOperands().size()); })
```

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -0,0 +1,414 @@
...
+    }
+}
+
+void ResourceAnalysis::collectOperation(Operation *op, ResourceResult &result, bool isAdjoint)
+{
+    // Quantum gates
+    if (isa<quantum::CustomOp, quantum::PauliRotOp, quantum::GlobalPhaseOp, quantum::MultiRZOp,
+            quantum::QubitUnitaryOp, quantum::SetStateOp, quantum::SetBasisStateOp>(op)) {
+        std::string name = getGateOpName(op, isAdjoint);
+        int nQubits = getGateQubitCount(op);
```
paul0403, line 302: e.g. here, qubit unitary would return 0

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -0,0 +1,414 @@
...
+/// Get the name for a PBC operation.
+static std::string getPBCOpName(Operation *op)
+{
+    return llvm::TypeSwitch<Operation *, std::string>(op)
+        .Case<pbc::PPRotationOp>([](auto pprOp) -> std::string {
+            int16_t rk = pprOp.getRotationKindAttr().getValue().getSExtValue();
+            if (rk == 0) {
+                return "PPR-identity";
+            }
+            return "PPR-pi/" + std::to_string(std::abs(rk));
```
paul0403, line 109: Should we distinguish between the X,Y,Z PPRs?

paul0403: Since you said this just mirrors the frontend's implementation in xdsl and runtime's implementation on null qubit, if they don't do it then don't worry about it here. I think [runtime passes the same "PauliRot" name to null qubit resource tracking](https://github.com/PennyLaneAI/catalyst/blob/cce42944c26b7f6adb6afaa27d31cfa0d9887107/runtime/lib/capi/RuntimeCAPI.cpp#L880) regardless of whether it's X, Y or Z.

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -0,0 +1,414 @@
...
+        .Case<quantum::ProbsOp>([](auto) { return "probs"; })
+        .Case<quantum::StateOp>([](auto) { return "state"; })
+        .Default([](Operation *o) { return o->getName().getStringRef().str(); });
+}
+
+//===----------------------------------------------------------------------===//
+// ResourceAnalysis implementation
+//===----------------------------------------------------------------------===//
+
+ResourceAnalysis::ResourceAnalysis(Operation *op)
```
paul0403, line 146: Having the "entry point function" of the file in the middle of the file is a bit hard to read 😅 Do you want to move it to the end of the file?

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -0,0 +1,414 @@
...
+{
+    LLVM_DEBUG(dbgs() << "ResourceAnalysis: analyzing operation " << op->getName() << "\n");
+
+    StringRef entryFunc;
+    op->walk([&](func::FuncOp funcOp) {
```
paul0403: Given that you are assuming the entry point is the first non-declaration function, and furthermore assuming it is the first one visited in the walk, it would probably be better to specify the walk order to be pre-order.

Otherwise, you can actually get the first occurance of an op of a certain type in another op like this:
```C
auto firstFuncOp = *op.getOps<func::FuncOp>().begin();
```

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -0,0 +1,414 @@
...
+        if (entryFunc.empty()) {
+            entryFunc = funcOp.getName();
+        }
+    });
+
+    entryFuncName = entryFunc.str();
+
+    if (!entryFunc.empty()) {
+        resolveFunctionCalls(entryFunc);
+    }
```
paul0403: What is the case when there's no entry func? 😅

paul0403 (commented):
Thanks @sengthai !

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
paul0403: Although, I guess the `getOps` way can't distinguish between function declarations and definitions.

mlir/lib/Catalyst/Analysis/ResourceResult.cpp
rniczh: It's the entry of the `StringMap` : `StringMapEntry`, so it supports `getKey()` or `getValue()`

paul0403: Ah yes! (I can see it from vscode now :face_holding_back_tears: )

mlir/include/Catalyst/Analysis/ResourceResult.h
sengthai: Yeah make sense! As for this `MergeMethod`, which is only used in `ResourceResult`, I will move it into `ResourceResult`  instead. 
[2db6284](https://github.com/PennyLaneAI/catalyst/pull/2479/commits/2db628406a4feb2d198d3f95c5bb9358e61131df)

mlir/lib/Catalyst/Analysis/CMakeLists.txt
sengthai: You're right that this is a bit of a maintenance burden. It would be nice to define the interface for quantum-related ops.

mlir/lib/Catalyst/Analysis/ResourceResult.cpp
sengthai: If we add `default`, the last `return` is still unreachable.

mlir/lib/Catalyst/Analysis/ResourceResult.cpp
sengthai: Thanks, @paul0403 ! this was inspired by Jake's qml.spec.

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: Good catch! 
[fc1b7ca](https://github.com/PennyLaneAI/catalyst/pull/2479/commits/fc1b7ca6196dda8c4c0cf2c15fbbbd17f24dfeb4)

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: Nice! I think we can use `QuantumOperation` which covers setState, setBasicSate, and etc. 
[deaf9c9](https://github.com/PennyLaneAI/catalyst/pull/2479/commits/deaf9c9d43d894256cb1aa7cd2f304e676f33143)

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: Fixed! [fc1b7ca](https://github.com/PennyLaneAI/catalyst/pull/2479/commits/fc1b7ca6196dda8c4c0cf2c15fbbbd17f24dfeb4)

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: > Should we distinguish between the X,Y,Z PPRs?

It is okay to ignore the Pauli string --- there is no difference between those Pauli strings. Its angle would be more important.

> Since you said this just mirrors the frontend's implementation in xdsl and runtime's implementation on null qubit, if they don't do it then don't worry about it here. 

Yes, they don't do it.

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: haha fair! [e7c6034](https://github.com/PennyLaneAI/catalyst/pull/2479/commits/e7c6034de5fdd088048a8052f6787f861c0767f0)

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: > it would probably be better to specify the walk order to be pre-order.

the default order of `op->walk` is post-order, however, it would be clearer to explicitly with pre-order. 🙌

sengthai: [58576ed](https://github.com/PennyLaneAI/catalyst/pull/2479/commits/58576ed4cf796f8e661bc29dd064859f66f232db)

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: haha good point! This can't happen in practice. I'll replace the else with an assertion.[58576ed](https://github.com/PennyLaneAI/catalyst/pull/2479/commits/58576ed4cf796f8e661bc29dd064859f66f232db)

paul0403 (approved):
Fantastic 🥳

PR #2481, opened 2026-02-12
Adding entry points for desired functions

isaacdevlugt:
**Context:** This PR will add entry points in Catalyst for functionality that is implemented in Catalyst *only*, making it so that the user can access these functions from PennyLane instead.

**Description of the Change:** Adds entry points to several functions

**Benefits:** Scalable docs

**Possible Drawbacks:**

**Related:** https://github.com/PennyLaneAI/pennylane/pull/9020 

[sc-113789](https://app.shortcut.com/xanaduai/story/113789/entry-points-in-catalyst)


setup.py
```diff
@@ -149,8 +149,14 @@ def parse_dep_versions():
         "cuda_quantum.ops = catalyst.api_extensions",
         "cuda_quantum.qjit = catalyst.third_party.cuda:cudaqjit",
     ],
-    "xdsl.universe": [
-        "catalyst-xdsl-universe = catalyst.python_interface.xdsl_universe:CATALYST_XDSL_UNIVERSE"
+    "catalyst.passes": [
```
isaacdevlugt: Whoops! 


```suggestion
    "xdsl.universe": [
        "catalyst-xdsl-universe = catalyst.python_interface.xdsl_universe:CATALYST_XDSL_UNIVERSE"
    ],
    "catalyst.passes": [
```

isaacdevlugt:
Sorry I should have opened this as a draft 😬. Whoops!

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -1398,44 +1493,87 @@ def circuit():
...
+      PPM-w3: 1
+      PPM-w1: 1
+      PPR-pi/2-w1: 1
+      PPR-pi/2-w2: 1
+      PPR-Phi-w1: 1
+    <BLANKLINE>
+    Measurements:
+        expval(PauliZ): 1
+
+<<<<<<< HEAD
```
andrijapau: lol merge artifact

setup.py
```diff
@@ -152,6 +152,18 @@ def parse_dep_versions():
     "xdsl.universe": [
         "catalyst-xdsl-universe = catalyst.python_interface.xdsl_universe:CATALYST_XDSL_UNIVERSE"
     ],
+    "catalyst.passes": [
```
andrijapau: Doesn't this have to be `catalyst.compilation_passes` aka match whatever is [here](https://github.com/PennyLaneAI/pennylane/pull/9020/changes#diff-ad5fb81dc63349c19267e307d3e74a843b45c6084b13f24c44aed7b0eb4ef2dc)?

setup.py
```diff
@@ -152,6 +152,18 @@ def parse_dep_versions():
...
+        "to_ppr = catalyst.passes:to_ppr",
+        "merge_ppr_ppm = catalyst.passes:merge_ppr_ppm",
+        "commute_ppr = catalyst.passes:commute_ppr",
+        "ppm_compilation = catalyst.passes:ppm_compilation",
+        "ppr_to_ppm = catalyst.passes:ppr_to_ppm",
+        "reduce_t_depth = catalyst.passes:reduce_t_depth",
+        "decompose_arbitrary_ppr = catalyst.passes:decompose_arbitrary_ppr",
+    ],
+    "pennylane.drawer": [
+        "draw_graph = catalyst:draw_graph",
```
isaacdevlugt, line 167: Images on PennyLane for `draw_graph` aren't showing. Not sure why 🤔

frontend/catalyst/passes/builtin_passes.py
```suggestion
```

frontend/catalyst/python_interface/inspection/draw.py
```diff
@@ -230,7 +230,7 @@ def circuit():
     >>> fig, ax = catalyst.draw_graph(circuit, level=0)()
     >>> fig.savefig('path_to_file.png', dpi=300, bbox_inches="tight")
 
-    .. figure:: ../../../doc/_static/catalyst-draw-graph-level0-example.png
+    .. figure:: /_static/catalyst-draw-graph-level0-example.png
```
isaacdevlugt, line 232: For some reason this paths doesn't translate to PennyLane's render of these docs 🤔

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -921,71 +943,116 @@ def ppr_to_ppm(qnode=None, *, decompose_method="pauli-corrected", avoid_y_measur
 
     .. note::
 
-        For improved integration with the PennyLane frontend, including inspectability with
-        :func:`pennylane.specs`, please use :func:`pennylane.transforms.ppr_to_ppm`.
+        This transform requires decorating the workflow with :func:`@qml.qjit <pennylane.qjit>`. In
+        addition, the circuits generated by this pass are currently not executable on any
```
dime10: Didn't we add support for this?

isaacdevlugt: Ah yes sorry! This is copy-pasta from PL-stable 😅

isaacdevlugt: I'll push an update shortly :)

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -635,182 +635,195 @@ def to_ppr(qnode):
 
     .. note::
 
-        For improved integration with the PennyLane frontend, including inspectability with
-        :func:`pennylane.specs`, please use :func:`pennylane.transforms.to_ppr`.
+        This transform requires decorating the workflow with :func:`@qml.qjit <pennylane.qjit>`. In
```
dime10: It's a bit weird to have this warning in the catalyst docs (of course they are using catalyst), and also since it's pointing to qml.qjit which could be interpreted as "instead of catalyst.qjit", but I imagine this is to keep the docs the same across the board?

setup.py
```diff
@@ -152,6 +152,18 @@ def parse_dep_versions():
     "xdsl.universe": [
         "catalyst-xdsl-universe = catalyst.python_interface.xdsl_universe:CATALYST_XDSL_UNIVERSE"
     ],
+    "pennylane.transforms": [
```
dime10, line 155: I'm assuming this allows PL to dynamically set the docstrings on those mirror functions? 

One thing I noticed with dynamically setting attributes like this, is that IDE tools can have trouble picking it up. For instance, with Microsoft's Pylance (VS Code, Cursor, etc), the tooltip for `qml.qnode` is significantly less helpful, both for the signature (no kwargs breakdown) and the docstring (no meaningful description):

Image

We can see the reason is most likely the dynamic setting:
https://github.com/PennyLaneAI/pennylane/blob/7e9d97b9b589cd3a7091aa8f9b120a5c060146bf/pennylane/workflow/qnode.py#L883-L889

Unfortunately I'm noticing this more and more when using PL/Catalyst as a user (e.g. `qml.qjit` doesn't show the full signature because of the [dynamic dispatching](https://github.com/PennyLaneAI/pennylane/blob/7e9d97b9b589cd3a7091aa8f9b120a5c060146bf/pennylane/compiler/qjit_api.py#L295-L297)/[lack of fixed kwargs](https://github.com/PennyLaneAI/pennylane/blob/7e9d97b9b589cd3a7091aa8f9b120a5c060146bf/pennylane/compiler/qjit_api.py#L21), `catalyst.qjit` doesn't show a proper tooltip either because of this [logging decorator](https://github.com/PennyLaneAI/catalyst/blob/7f94484694454a33101df42d5fa71a79b735b0f3/frontend/catalyst/jit.py#L76), "go to definition" often has trouble with PennyLane modules, etc.), which meaningfully diminishes the programming experience.

I don't know what this means for this PR, but I think if tool integration isn't something on our radar yet, I believe it's worth looking into because of the outsized impact it has on the PennyLane user experience.

dime10 (approved):

setup.py
isaacdevlugt: Really appreciate this comment! Had no idea this was happening. Personally, I always go to the source for docs and don't use the Pylance tool in VSCode. 

It's a bit of whack-a-mole here lol. We're making these changes to have better consistency in documentation, both with the user in mind and for the developer; having one place where documentation lives for compilation passes is obviously a better experience in most cases. 

But, as you've pointed out, it seems there is a downside to doing this! I'm not sure if we can find a remedy immediately, but let's chat offline about how we can address this going forward.

dime10: No worries, giving how this might be a wider issue I think we it's okay to merge this PR for now :) 

It's not just for docs btw, it affects the autosuggestion as your typing as well (whether it's function kwarg names or function names from a module, etc).

frontend/catalyst/passes/builtin_passes.py
isaacdevlugt: > but I imagine this is to keep the docs the same across the board?

You got it! Though, the ``qjit`` point is a good one. I suppose we could just do `~.qjit`? Would docs render in such a way that PL and Catalyst docs would point to `pennylane.qjit` or `catalyst.qjit`, respectively?

dime10: Hmm great question, we would need to try it out 😅

isaacdevlugt: But, another point is that we want PL to be the frontend... so having this doesn't bother me that much

josh146: I'd be interested to know if `~.qjit` would work for both library docs, I suspect it would!

isaacdevlugt:
```suggestion
        This transform requires decorating the workflow with :func:`@qjit <~.qjit>`. In
```

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -141,192 +141,114 @@ def circuit(x: float):
...
-    QNode during :func:`~.qjit` compilation.
+    """A peephole optimization for replacing ``CNOT`` gates with single-qubit gates.
+
+    .. note::
+
+        This transform requires decorating the workflow with :func:`pennylane.qjit`.
 
     Args:
-        fn (QNode): the QNode to apply the disentangle CNOT compiler pass to
+        fn (QNode): the QNode to apply the pass to.
```
dime10: Nitpick, but as fas as I know our guidance is either full sentence syntax (capitalization + punctuation), or sentence fragments without either.

isaacdevlugt:
```suggestion
        fn (QNode): the QNode to apply the pass to
```

😄

frontend/catalyst/passes/pass_api.py
```diff
@@ -295,19 +295,6 @@ def __init__(
         super().__init__(name, *options, **valued_options)
 
 
-_PASS_NAME_OVERRIDES = {
```
andrijapau, line 298: Thanks 😄

andrijapau (approved):
Thanks for walking me through the changes and demoing the feature @isaacdevlugt 🎉 

Looks good to me. :letsgo:

PR #2483, opened 2026-02-13
Provide some guidelines for how to write QNode passes 

dime10:
[sc-107497]


joeycarter (approved):
Looks good! 💯 

I have no notes for now, but may want to revisit next time I write a qnode pass. Thanks @dime10!

frontend/catalyst/python_interface/transforms/qnode-transform-guide.md
```diff
@@ -0,0 +1,162 @@
...
+    schedule for the QNode scope (anonymous module with the `transform.with_named_sequence`
+    attribute), as well as one or more functions. The `launch_kernel` instruction defines which
+    function is the starting for the execution of a quantum kernel.
+  - NOTE: The quantum kernel abstraction presented here only exists until the `inline-nested-module`
+          pass is run. However, most quantum transformations (including anything scheduled from the
+          frontend) happen *before* the quantum kernel is lowered to a simpler representation, so
+          this should be of little impact to most quantum transform developers.
+          (see Catalyst's [default pipeline](https://github.com/PennyLaneAI/catalyst/blob/v0.14.0/frontend/catalyst/pipelines.py#L229))
+- Within a quantum kernel, individual quantum executions are represented by functions that
+  carry the special `qnode` attribute. We will refer to these as *qnode functions* (name TBD):
```
dime10, line 74: The naming might be a bit confusing (and as usual is historical). We didn't have the quantum kernel abstraction initially, so QNodes were directly translated into functions with the qnode attribute. Now we have the quantum kernel, which the PL QNode is translated to (see [here](https://github.com/PennyLaneAI/catalyst/blob/3bed914ab2e3bdad42c9a2e639466ad33e749d97/frontend/catalyst/from_plxpr/from_plxpr.py#L345) for example). Yet we still have individual functions within the kernel representing separate quantum executions, identified by the qnode attribute. 
Happy to consider changes though, especially once we tackle the "better containers for quantum programs".

frontend/catalyst/python_interface/transforms/qnode-transform-guide.md
```diff
@@ -0,0 +1,162 @@
...
+  - This applies to xDSL as well as MLIR passes, since both are scheduled on the quantum kernel.
+- A pass is resposible for traversing the module it is scheduled on and identify all qnode functions
+  present within (identifiable via the `qnode` attribute).
+- Unless otherwise required, each qnode function should be processed once by the pass.
+- Unless otherwise required, each qnode function should be processed independently.
+- A pass should exit early without error if it cannot find the operations or patterns it is
+  designed to target. Execution should only be aborted where IR invariants or Catalyst program
+  model expectations are clearly violated.
+- Passes that wish to replace an existing qnode function with multiple quantum executions
+  or with additional classical processing can do so by:
```
dime10, line 111: Let me know if this is clear enough. We could add IR example if additional clarity is needed :)

mudit2812 (commented):
> Let me know if this is clear enough. We could add IR example if additional clarity is needed :)

We already have a google doc with an IR example, so it would be great to move it here if this is going to be a centralized guidance doc. Let me see if I can find the doc and send you a link on Slack.

frontend/catalyst/python_interface/transforms/qnode-transform-guide.md
lillian542: It's clear to me, but we've already discussed so I might not be the best test-case. Maybe we could add a link to the new MLIR implementation of `split_non_commuting` as an example of multiple quantum executions, and `split_to_single_terms` as an example of a single quantum execution that adds post processing?

dime10: @mudit2812 added the example from the doc here [5a96123](https://github.com/PennyLaneAI/catalyst/pull/2483/commits/5a9612343f3bbf17bca54d5c0d3136ee6b34b7c5)
@lillian542 links added here [b6bb749](https://github.com/PennyLaneAI/catalyst/pull/2483/commits/b6bb74989405cc0748515e10700b1b4565377db6)

frontend/catalyst/python_interface/transforms/qnode-transform-guide.md
paul0403: I think the best definition of a "qnode" is indeed what you outlined in #2497 , that it contains a terminal MP. Terminal MPs translate quantum states on the device to classical outputs back to the user, and I think that is a good boundary to draw along the "quantum blackbox".

Although it is true that (what you mentioned in the verification PR)

> Catalyst still uses MP-free QNodes as a feature

but I think we still can keep this definition in mind, since we're moving away from this technicality (by demanding classical outputs to be wrapped in `qml.sample()`, though I don't actually know when that is happening).

frontend/catalyst/python_interface/transforms/qnode-transform-guide.md
```diff
@@ -0,0 +1,281 @@
...
+- Within a quantum kernel, individual quantum executions are represented by functions that
+  carry the special `qnode` attribute. We will refer to these as *qnode functions* (name TBD):
+  - A precise definition is difficult to provide and may not currently apply to all transformations
+    or stages in Catalyst, however, for the purpose of this guide a qnode function will contain:
+    - device [initialization](https://github.com/PennyLaneAI/catalyst/blob/v0.14.0/frontend/catalyst/python_interface/dialects/quantum.py#L389)
+      and [release](https://github.com/PennyLaneAI/catalyst/blob/v0.14.0/frontend/catalyst/python_interface/dialects/quantum.py#L412)
+      instructions
+    - one or more [measurement processes](https://github.com/PennyLaneAI/catalyst/blob/v0.14.0/frontend/catalyst/python_interface/dialects/quantum.py#L1058)
+
+    Conversely, such operations should not appear in any other kind of function. Quantum state
```
paul0403, line 82: What about, say, two MPs, and the user function does some classical post processing? e.g. split non commuting

```
func.func X_execution(){
   device_init
   Hadamard
   return expval(X)(wires=0)
}

func.func Y_execution(){
   device_init
   Hadamard
   return expval(Y)(wires=0)
}

func.func workflow(){
    return X_execution() + Y_execution()
}
```

Would there be two qnodes in this case?

And if the answer is yes, what's the pros and cons? 
What I can think of now is:
- Some passes might use the qnode attribute to mean "entry point". The validity of this interpretation is up for debate (I don't think this is a valid use of the attribute), but with your definition this use will no longer be valid.
- There are now 2 qnode functions, whereas in python frontend the user just has 1 qnode function; but probably this is not an issue for any pass writers, which is what we care about here.

But I think there's far more pros:
- The "main" function in this case is just classical postprocessing, and of course the quantum passes should ignore it. 
- Each qnode function is its own execution on the device (in this particular instance, it's because X and Y cannot be simultaneously measured, so 2 device executions is needed no matter what), and requires its own device instance. Having this kind of device handling explicitly marked in the IR does seem beneficial, although I can't think of a concrete use case yet. Maybe shots budgeting? 
- Most importantly, each qnode function is now its own independent unit. Quantum passes can just look for the "qnode" attribute, and ignore anything outside the qnode func ops.

paul0403: Ah, I jumped the gun. I see your other definitions below. I think we're on the same page! 💯

frontend/catalyst/python_interface/transforms/qnode-transform-guide.md
```diff
@@ -0,0 +1,281 @@
...
+### Guidelines
+
+During the design of quantum compilation passes targeting Catalyst programs with the above
+structure, the following principles should be applied to ensure composability with other passes
+and smooth integration into the compilation infrastracture.
+
+- Passes targeting PennyLane QNodes must be written to run on `ModuleOp`s.
+  - The reason is that PennyLane QNodes are translated to quantum kernel modules in the IR
+    (see assumptions section).
+  - This applies to xDSL as well as MLIR passes, since both are scheduled on the quantum kernel.
```
paul0403, line 102: Up to you, but I think it's also good to mention the MLIR rule that passes must not alter anything outside the scope of their target operations. If passes are targeting only the func op with the qnode attribute, as opposed to its surrounding quantum kernel module, these passes will not be able to (safely) create any other functions in the quantum kernel module. This invalidates a whole class of transforms that involves some sort of outling, e.g. dynamic one shot, split non commuting, ...

paul0403: (Although for local transformations this is fine, like cancel inverses)

frontend/catalyst/python_interface/transforms/qnode-transform-guide.md
dime10: Yes, because one QNode in PL can actually mean multiple executions, we define "qnode function" (as in a function with the "qnode" attribute) here to mean 1 execution. The quantum kernel on the other hand maintains the scope of the original PL "QNode". The naming is a bit unfortunate as mentioned in my other comment, but I'm happy to rename the attribute if we think of something better.

frontend/catalyst/python_interface/transforms/qnode-transform-guide.md
dime10: > If passes are targeting only the func op with the qnode attribute, as opposed to its surrounding quantum kernel module, these passes will not be able to (safely) create any other functions in the quantum kernel module.

Passes mustn't target a func op per the guidelines in this doc.

frontend/catalyst/python_interface/transforms/qnode-transform-guide.md
dime10: Yeah agreed that definition is reasonable even with the MP-free QNodes since that feature will be removed in the near future.
The other concern though I encountered looking at existing passes is how the device init/release and qureg allocation/deallocation instructions should be considered in the definition? 
- Obviously you can't have an MP without a device, so it seems like it should be core to the definition as well. Is it allowed to appear outside the function with the MP though?
  - I would think that each device instruction is core to a particular execution since it sets the shot number for example.
  - On the other hand, the device information (besides things like shots) is generally identical across a PL QNode and could thus be scoped to a quantum kernel.
- For allocations, there is a bit more flexibility since they should be able to appear in subroutines etc. But an MP needs to act on _some_ qubits, so does the allocate instruction need to be in the definition too? If so, can it appear outside the function with the MP?

Examples of passes that might break out of the norm of device + alloc + mp all in a single function:
- adjoint diff method: The device_release and register dealloc calls are moved outside of the "qnode" scope, since the lowering generates tape toggling and the device gradient call in the classical context.
- zne: The MPs are placed in a subroutine and called from the "main" qnode function which carries the device and allocation instructions. The deallocation is placed together with the MP subroutine.

I don't think we have to answer these questions right now though, this is the sort of thing I want to iron out and formalize with the better containers for quantum programs. At the moment the guide just works with something that's practical for most scenarios in the current design.

paul0403: I think these feel like peculiarities with those two specific passes, rather than the definition being unsound!

paul0403 (approved):

frontend/catalyst/python_interface/transforms/qnode-transform-guide.md
dime10: Interestingly, there is the prior definition of a "tape" in the IR for the split tape pass like this:
> // A tape is the operations between a "quantum.device" and a "quantum.device_release", inclusive.

Conceptually the tape is similar to what we're defining as qnode functions as here

paul0403: Yes, I remember struggling to define what was the scope of a tape when I was working on the split tapes. I think eventually we settled on the device init and release since that's usually the first and last quantum op surrounding a block of "semantically together" quantum ops (which now we call "qnode" in this doc)

PR #2487, opened 2026-02-18
The builtin passes and `apply_pass` use `qml.transform(pass_name="name")`

lillian542:
**Context:**
If you use `apply_pass` with program capture enabled, everything runs but it does not apply the pass. The same problem is true for any builtin pass that is accessed via `from catalyst.passes import pass_name`. This has created some situations that are difficult to debug. 

We switch the `apply_pass` function to use the more general `qml.transform(pass_name="name-of-pass")` UI instead of `PassPipelineWrapper`, which isn't compatible with program capture. We also update the builtin passes to use the new more general UI.

Note that there are still two functions in the `pass_api`, `pipeline` and `apply_pass_plugin`, that do return a `PassPipelineWrapper`, so it is still currently possible to encounter one in specific cases.

**Benefits:**
Less confusing behaviour when switching to program capture.

**Downsides**

- The `PassPipelineWrapper` raised a clear error message if passed a non-Qnode function, and these errors are no longer raised when we switch to `qml.transform`.
- The `pipeline` and `apply_pass_plugin` will still return PassPipelineWrapper, which creates an issue if these are used alongside (specifically if they are used before) builtin passes or `qml.transform(pass_name...)`

[sc-112140][sc-104261]


dime10:
> @dime10 Now that we have `@qml.transform(pass_name="...")`, which works with both with and without capture, @albi3ro suggested that we can deprecate catalyst legacy frontend's own `@apply_pass`, which currently silently does nothing when capture is enabled (instead of explicitly telling the user the pass will not be applied).
> 
> Any objections?

Sure, just check with @isaacdevlugt

frontend/catalyst/passes/pass_api.py
albi3ro:
```suggestion
    return qml.transform(pass_name=pass_name, *flags, **valued_options)
```

isaacdevlugt:
> > @dime10 Now that we have `@qml.transform(pass_name="...")`, which works with both with and without capture, @albi3ro suggested that we can deprecate catalyst legacy frontend's own `@apply_pass`, which currently silently does nothing when capture is enabled (instead of explicitly telling the user the pass will not be applied).
> > Any objections?
> 
> Sure, just check with @isaacdevlugt

@dime10 @albi3ro no issues from my side :). We have something better now!

frontend/catalyst/passes/pass_api.py
```diff
@@ -202,11 +202,7 @@ def qnode():
             def module():
                 return qnode()
     """
-
-    def decorator(qnode):
-        return PassPipelineWrapper(qnode, pass_name, *flags, **valued_options)
-
-    return decorator
+    return qml.transform(pass_name=pass_name, *flags, **valued_options)
```
albi3ro:
```suggestion
    return lambda obj: qml.transform(pass_name=pass_name)(obj, *flags, **valued_options)
```
more or less

paul0403:
Note that "frontend tests" CI fails but "frontend tests: kokkos" CI succeeds because frontend lit tests are only run in the "frontend tests" job, if you were wondering.

PR #2490, opened 2026-02-19
Add lowering of conditional pbc.ppm in unroll-conditional-ppr-ppm pass

### LINKED ISSUE #2475: Support unroll conditional PPM
Currently, we are able to unroll `SelectPPMeasurementOp` and conditional `PPRotationOp` (see [here](https://github.com/PennyLaneAI/catalyst/blob/483c6cbd5e6459310e984c66259165be1293bb60/mlir/lib/QEC/Transforms/UnrollConditionalPPRPPM.cpp)). However, conditional `PPMeasurmentOp` has not been supported yet.

**Goal:**
Here, we need to lower the conditional `PPMeasurementOp` as well.

The code that produce conditional `PPMeasurementOp`.
```python
import pennylane as qp

dev = qp.device("null.qubit", wires=2)

pipe = [("pipe", ["quantum-compilation-stage"])]

qp.capture.enable()
@qp.qjit(pipelines=pipe, target="mlir")
@qp.transform(pass_name="ppr-to-ppm")(decompose_method="clifford-corrected", avoid_y_measure=True)
@qp.transform(pass_name="to-ppr")
@qp.qnode(dev)
def cir_inject_magic_state():
    qp.T(0)
    qp.CNOT([0, 1])

print(cir_inject_magic_state.mlir_opt)
```

sengthai:
**Context:**
The `unroll-conditional-ppr-ppm` pass already lowers `pbc.select.ppm` and conditional
`pbc.ppr` (with `cond(...)`) into `scf.if` blocks for runtime execution. However,
conditional `pbc.ppm` operations were not handled, which meant any `pbc.ppm ... cond(%c)`
would survive the pass and fail during later lowering stages.

**Related GitHub Issues:**
close #2475 
[[sc-111303]]

### DIFF
--- a/mlir/lib/PBC/Transforms/ConversionPatterns.cpp
+++ b/mlir/lib/PBC/Transforms/ConversionPatterns.cpp
@@ -137,7 +137,7 @@ struct PPMeasurementOpPattern : public OpConversionPattern<PPMeasurementOp> {
                                   ConversionPatternRewriter &rewriter) const override
     {
         if (op.getCondition()) {
-            return op.emitOpError("PPMeasurementOp with condition is not supported.");
+            return op.emitOpError("Conversion of PPMeasurementOp with condition to LLVM is not supported. Expected op to be unrolled with --unroll-conditional-ppr-ppm.");
         }
 
         Location loc = op.getLoc();
--- a/mlir/lib/PBC/Transforms/UnrollConditionalPPRPPM.cpp
+++ b/mlir/lib/PBC/Transforms/UnrollConditionalPPRPPM.cpp
@@ -163,23 +163,12 @@ struct LowerCondPPM : public OpRewritePattern<PPMeasurementOp> {
         IntegerAttr rotationSign = op.getRotationSignAttr();
         ValueRange inQubits = op.getInQubits();
 
-        SmallVector<mlir::Type> resultTypes;
-        resultTypes.push_back(rewriter.getI1Type());
-        for (auto qubit : inQubits) {
-            resultTypes.push_back(qubit.getType());
-        }
-
-        auto ifOp = scf::IfOp::create(rewriter, loc, resultTypes, condition, true);
+        auto ifOp = scf::IfOp::create(rewriter, loc, op->getResultTypes(), condition, true);
         {
             OpBuilder::InsertionGuard guard(rewriter);
             rewriter.setInsertionPointToStart(&ifOp.getThenRegion().front());
-            auto ppm =
-                PPMeasurementOp::create(rewriter, loc, rewriter.getI1Type(), TypeRange(inQubits),
-                                        pauliProduct, rotationSign, inQubits);
-            SmallVector<mlir::Value> yieldValues;
-            yieldValues.push_back(ppm.getMres());
-            yieldValues.append(ppm.getOutQubits().begin(), ppm.getOutQubits().end());
-            scf::YieldOp::create(rewriter, loc, yieldValues);
+            auto ppm = rewriter.clone(op);
+            scf::YieldOp::create(rewriter, loc, ppm->getResults());
         }
 
         {


doc/releases/changelog-dev.md
```diff
@@ -312,6 +312,7 @@
 * A new compiler pass `unroll-conditional-ppr-ppm` for lowering conditional PPR and PPMs
   into normal PPR and PPMs with SCF dialect to support runtime execution.
   [(#2390)](https://github.com/PennyLaneAI/catalyst/pull/2390)
+  [(#XXXX)](https://github.com/PennyLaneAI/catalyst/pull/XXXX)
```
rniczh:
```suggestion
  [(#2490)](https://github.com/PennyLaneAI/catalyst/pull/2490)
```

rniczh (approved):
💯

mlir/lib/PBC/Transforms/ConversionPatterns.cpp
```diff
@@ -136,6 +136,10 @@ struct PPMeasurementOpPattern : public OpConversionPattern<PPMeasurementOp> {
     LogicalResult matchAndRewrite(PPMeasurementOp op, PPMeasurementOpAdaptor adaptor,
                                   ConversionPatternRewriter &rewriter) const override
     {
+        if (op.getCondition()) {
+            return op.emitOpError("PPMeasurementOp with condition is not supported.");
```
paul0403: Maybe we can indicate an action for the user here, since it is relatively straightforward?

```suggestion
            return op.emitOpError("Conversion of PPMeasurementOp with condition to LLVM is not supported. Expected op to be unrolled with --unroll-conditional-ppr-ppm.");
```

(This will hit clang format line-too-long, but I'll leave you to do the editing)

mlir/lib/PBC/Transforms/UnrollConditionalPPRPPM.cpp
```diff
@@ -134,6 +135,68 @@ struct LowerCondPPR : public OpRewritePattern<PPRotationOp> {
...
+        IntegerAttr rotationSign = op.getRotationSignAttr();
+        ValueRange inQubits = op.getInQubits();
+
+        SmallVector<mlir::Type> resultTypes;
+        resultTypes.push_back(rewriter.getI1Type());
+        for (auto qubit : inQubits) {
+            resultTypes.push_back(qubit.getType());
+        }
+
+        auto ifOp = scf::IfOp::create(rewriter, loc, resultTypes, condition, true);
```
paul0403:
```suggestion
        auto ifOp = scf::IfOp::create(rewriter, loc, op->getResultTypes(), condition, true);
```
no?

mlir/lib/PBC/Transforms/UnrollConditionalPPRPPM.cpp
```diff
@@ -134,6 +135,68 @@ struct LowerCondPPR : public OpRewritePattern<PPRotationOp> {
...
+        {
+            OpBuilder::InsertionGuard guard(rewriter);
+            rewriter.setInsertionPointToStart(&ifOp.getThenRegion().front());
+            auto ppm =
+                PPMeasurementOp::create(rewriter, loc, rewriter.getI1Type(), TypeRange(inQubits),
+                                        pauliProduct, rotationSign, inQubits);
```
paul0403:
```suggestion
            auto ppm = rewriter.clone(op);
```

mlir/lib/PBC/Transforms/UnrollConditionalPPRPPM.cpp
```diff
@@ -134,6 +135,68 @@ struct LowerCondPPR : public OpRewritePattern<PPRotationOp> {
...
+            SmallVector<mlir::Value> yieldValues;
+            yieldValues.push_back(ppm.getMres());
+            yieldValues.append(ppm.getOutQubits().begin(), ppm.getOutQubits().end());
+            scf::YieldOp::create(rewriter, loc, yieldValues);
```
paul0403:
```suggestion
            scf::YieldOp::create(rewriter, loc, ppm->getResults());
```

mlir/lib/PBC/Transforms/UnrollConditionalPPRPPM.cpp
```diff
@@ -134,6 +135,68 @@ struct LowerCondPPR : public OpRewritePattern<PPRotationOp> {
...
+        }
+
+        {
+            OpBuilder::InsertionGuard guard(rewriter);
+            rewriter.setInsertionPointToStart(&ifOp.getElseRegion().front());
+            auto falseCst = arith::ConstantOp::create(rewriter, loc, rewriter.getBoolAttr(false));
```
paul0403, line 177: In my understanding, a conditioned PPM op means "perform the PPM if the condition is true, do nothing if the condition is false".

In the false branch, the PPM never happened, so it follows that the input qubits to the original PPM gets yielded from the if op, since no gates acted on them; but does it really make sense for the measurement result to take on false? Since there were no measurement actions to begin with, and keep in mind that "false" doesn't stand for the measurement never happened, it stands for one of the possible measurement results!

paul0403: In my opinion, it would make more sense to return a [poison value](https://mlir.llvm.org/docs/Dialects/LLVM/#llvmmlirpoison-llvmpoisonop) here, since if execution went through the false branch, the "measurement result placeholder SSA value" from the PPM should never be used, since it doesn't correspond to any real actions. 
```C
LLVM::PoisonOp::create(rewriter, loc, rewriter.getI1Type());
```

```mlir
%0 = llvm.mlir.poison : i1
```

See https://mlir.llvm.org/OpenMeetings/2022-11-03-Poison-Semantics-For-MLIR.pdf, the poison value was introduced to combat illegal values, which is precisely what we have here: a value not corresponding to any concrete runtime action, only existing to satisfy the yield structure of its enclosing region, and should never be used in any semantically meaningful way. 

Note that a poison value does not break the program when it is created (by the poison op), only when it is used: precisely what we want here. See https://llvm.org/docs/LangRef.html#poison-values for the semantics of poison values in LLVM. Notably, 

> This means that immediate undefined behavior occurs if a poison value is used as **an instruction operand** that has any values that trigger undefined behavior.

In other words, the poison boolean value must not be used is any semantically meaningful way. For example, the "measurement result" must not be the boolean conditional for some other scf.if ops, since the "measurement" never happened in the false branch.

dime10: That's a good observation! ~I don't know if we have poison values in MLIR? I suppose from the LLVM dialect.~ Ah you already linked to it 👍 
Another option would be to assert that the measurement result has no uses and not return it from the if statement. I don't know how we use the conditional PPM, but you would only be able to use the outcome in a branch that's also conditioned on the PPM condition I guess 🤔 

But the poison value does seem like the most flexible solution :)

paul0403: > Another option would be to assert that the measurement result has no uses and not return it from the if statement.

I thought about this, but I suppose it's perfectly fine for the true branch's measurement result to have uses, so the if op itself does need to yield the boolean.

paul0403: > I don't know if we have poison values in MLIR? I suppose from the LLVM dialect.

We do, it was one of the LLVM updates that I did. https://github.com/llvm/llvm-project/pull/125629

lazypanda10117: I like the poison value idea. I checked the actual ppm_to_ppr implementation, and I think not having the if op itself yield a boolean would also work. In the PBC pipeline, the only place we use it is to decompose conditional `Pi/4` rotations into PPMs.

Sengthai and I actually chatted about this issue before and decided not to handle the case of conditional PPM, but I guess there's a need to do so now.

doc/releases/changelog-dev.md
isaacdevlugt: @sengthai I might actually add a bug-fix entry as well as including this PR in this changelog entry 👀

mlir/lib/PBC/Transforms/UnrollConditionalPPRPPM.cpp
sengthai: Thanks everyone for this discussion! 

@lazypanda10117 and I also discussed the usability of conditional PPM. Now I'm thinking of stopping support for conditional PPM and using `scf:ifOp` instead. First, this conditional PPM only appears when lowering conditional PPR(pi/4) into PPM, so using `scf:ifOp` would be better. Second, we are also aware that it doesn't make sense to make a conditional PPM from the get go. Therefore, this PR should be closed and replaced with another PR to remove the conditional PPM from the dialect.

doc/releases/changelog-dev.md
sengthai: Thanks @isaacdevlugt!

PR #2496, opened 2026-02-19
Put device preprocessing transforms in `transform.named_sequence`

mudit2812:
**Context:**
* This PR adds support for device preprocessing with program capture.

* Add a new `catalyst.from_plxpr.device_utils` submodule, which contains a function for generating a device preprocessing pipeline based on device capabilities and the execution config.
  * For transforms that don't have a `pass_name`, the transform name is converted into kebab case and then processed as follows:
    * Some transforms don't have a `pass_name` but have xDSL implementations. So, the kebab-ed transform name is compared against the dictionary of registered xDSL transforms, and if an xDSL transform is available, that is used.
    * Otherwise, a dummy xDSL transform is created, registered, and used. This transform has a `dummy_transform` boolean argument that can be seen in the printed MLIR to indicate to devs that the pass is a dummy transform.
    * If any dummy transforms are added to the pipeline, a warning is raised to let the user know that some dummy transforms are present.
* Append the transforms returned by this function to the qnode pass_pipeline
* Add a `skip_preprocess` argument to `qjit`, `trace_from_pennylane`, `from_plxpr`, and `WorkflowInterpreter`, which can be used to skip appending device preprocessing transforms to the pass pipeline.
* Add a `_preprocess_warn=True` argument to `from_plxpr` and `WorkflowInterpreter` which can be used to toggle the dummy transform warnings off. This argument is private and not exposed by `qjit`. I added it as a testing convenience.
* I also found a bug with `catalyst.python_interface.pass_api.TransformInterpreterPass`, where it was not interpreting pass options correctly and was preventing me from adding the `dummy_transform` argument to the dummy xDSL transforms, which is also fixed in this PR.
* The preprocessing function also does MCM related validation and preprocessing.

**Benefits:**
* Device preprocessing available with program capture.

**Possible Drawbacks:**
* The dummy transforms warning will be raised for every single QJIT compilation when program capture is used unless `skip_preprocess=True`, which is not the default. This might get annoying, especially for developers who will probably see a barrage of warnings when running tests.

**Related GitHub Issues:**

[sc-111493] [sc-112073]

### DIFF
--- a/frontend/test/pytest/from_plxpr/test_preprocessing.py
+++ b/frontend/test/pytest/from_plxpr/test_preprocessing.py
@@ -44,15 +44,19 @@ class CapabilitiesDevice(NullQubit):
 
     @property
     def capabilities(self) -> DeviceCapabilities:
+        """Capabilities."""
         return self._capabilities
 
     @capabilities.setter
     def capabilities(self, obj: DeviceCapabilities):
+        """Capabilities setter."""
         self._capabilities = obj
         self._capabilities.qjit_compatible = True
 
     @property
     def qjit_capabilities(self):
+        """qjit_capabilities alias to capabilities. Used to override Catalyst behaviour
+        where it prioritizes the TOML file over device.capabilities."""
         return self._capabilities
 
 


frontend/catalyst/device/qjit_device.py
```diff
@@ -397,6 +398,29 @@ def preprocess(
             unitary_support=getattr(self.original_device, "_to_matrix_ops", None),
         )
 
+        # MCM preprocessing. We use the **original** execution_config rather than the
+        # config returned by original_device.preprocess, so that the MCMConfig doesn't
+        # get resolved by the original device.
+        mcm_config = execution_config.mcm_config
+        if mcm_config.mcm_method is not None:
```
andrijapau: Why do we need this outer `if`? Doesn't the `elif` below cover it?

andrijapau: I feel like this is all you need
```python
if method == MCM_METHOD.ONE_SHOT:
    ...
elif method is not None:
    supported = {...}
    if method not in supported:
        raise CompileError(...)
```

frontend/catalyst/device/qjit_device.py
```diff
@@ -397,6 +398,29 @@ def preprocess(
             unitary_support=getattr(self.original_device, "_to_matrix_ops", None),
         )
 
+        # MCM preprocessing. We use the **original** execution_config rather than the
+        # config returned by original_device.preprocess, so that the MCMConfig doesn't
+        # get resolved by the original device.
+        mcm_config = execution_config.mcm_config
```
andrijapau: Can we do all execution config processing before the pipeline processing?

andrijapau: Or better yet use `setup_execution_config` and `preprocess_transforms` if possible

frontend/catalyst/python_interface/pass_api/transform_interpreter.py
```diff
@@ -92,7 +92,9 @@ def run_apply_registered_pass_op(
         # ---- xDSL path ----
         if pass_name in self.passes:
             pass_class = self.passes[pass_name]()
-            options = get_pyval_from_xdsl_attr(op.options)
+            options = {
+                k.replace("-", "_"): v for k, v in get_pyval_from_xdsl_attr(op.options).items()
+            }
```
mudit2812, line 97: This was needed because the pass options dictionary had all options in kebab case

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -193,7 +200,13 @@ def f(x):
...
     """
 
-    original_fn = partial(WorkflowInterpreter().eval, plxpr.jaxpr, plxpr.consts)
+    original_fn = partial(
+        WorkflowInterpreter(
+            skip_preprocess=skip_preprocess, _preprocess_warn=_preprocess_warn
+        ).eval,
+        plxpr.jaxpr,
+        plxpr.consts,
+    )
```
albi3ro, line 209:
```suggestion
    interpreter = WorkflowInterpreter(
        skip_preprocess=skip_preprocess, _preprocess_warn=_preprocess_warn
    )
    original_fn = partial(interpreter.eval,plxpr.jaxpr,plxpr.consts)
```
just breaking a long line up into two steps.

frontend/catalyst/from_plxpr/device_utils.py
JerryChen97, line 1: Is there any specific requirement so that we need to use `BoundTransform` and construct lists of them instead of directly using `CompilePipeline` here?

frontend/catalyst/device/qjit_device.py
mudit2812: Resolved.

frontend/catalyst/from_plxpr/device_utils.py
mudit2812: The `quantum_kernel_p` primitive takes a tuple as its `pipeline` input. We can change it to `CompilePipeline`, but this late in the compilation pipeline it's quite useless because the abstractions in `CompilePipeline` are no longer useful

JerryChen97: I see. This makes sense to me!

frontend/catalyst/python_interface/pass_api/transform_interpreter.py
JerryChen97: Instead of patching the usecase, would it be easier to directly adjust the implementation of `get_pyval_from_xdsl_attr`? I saw only three uses, two of which are for its own recursive definition, so actually this line is the only place this helper gets used

frontend/catalyst/from_plxpr/device_utils.py
```diff
@@ -0,0 +1,305 @@
...
+
+    return BoundTransform(transform, args, kwargs)
+
+
+@lru_cache
+def _get_xdsl_transform(original_transform: Transform | BoundTransform) -> tuple[Transform, bool]:
+    """Create an xDSL transform to insert into the compile pipeline. A boolean indicating
+    whether the transform is a dummy transform is also returned."""
+    # pylint: disable=import-outside-toplevel
+    from catalyst.python_interface.pass_api import compiler_transform, is_xdsl_pass
```
JerryChen97, line 284: Is there risk of circular deps here

frontend/catalyst/python_interface/pass_api/transform_interpreter.py
mudit2812: The purpose of `get_pyval_from_xdsl_attr` is far more general than what we are using it for here. There is no reason why every single string should replace underscores with hyphens.

PR #2497, opened 2026-02-19
Add verification for `qnode` attribute functions (now `quantum.node`)

dime10:
_Note I don't think this PR is a must, but while working on the [transform post-processing epic](https://app.shortcut.com/xanaduai/epic/107240) one of the ideas was to better formalize some of the attributes we use and might have implicit assumptions for. The verification introduced here is not very strong, but it is something. There are several older passes in Catalyst which currently don't follow the proposed notion of qnode functions (e.g. they might [not propagate the attribute](https://github.com/PennyLaneAI/catalyst/blob/main/mlir/lib/Gradient/Transforms/GradMethods/ParameterShift.cpp) in new functions they create, or they might also have [unusual structures](https://github.com/PennyLaneAI/catalyst/blob/main/mlir/lib/Mitigation/Transforms/MitigationMethods/Zne.cpp) like having MPs in a function but device init/quantum allocs in a different one)._

In this PR, the existing `qnode` attribute is renamed to `quantum.node` to indicate the dialect origin. Additionally, verification logic is introduced to formalize the notion of a "qnode" function (see also #2483), which is characterized by the presence of measurement process operations. These operations represent an implicit transition from quantum state manipulation to execution logic and classical post-processing. While quantum gate operations can appear in functions that are not qnodes, measurement processes should not as they have this "non-local" effect. The attribute helps identify functions with quantum execution semantics during transformations.

The new attribute and verification logic is currently implemented during the "quantum kernel" phase only, i.e. before `inline-nested-modules` is called. All functions carrying the attribute ~must have at least one measurement process~ (I have removed this verification for now since Catalyst still uses MP-free QNodes as a feature), and all measurement processes appearing within a nested module (quantum kernel) must be contained in a func op with that attribute.

To limit the scope of this PR the new attribute is replaced with the previous `qnode` attribute after lowering the quantum kernel module.

A bug in the xDSL split-non-commuting pass is also fixed (required to pass the new verification). https://github.com/PennyLaneAI/catalyst/pull/2497/commits/9f75c0ec87abd647f540739ba8ab647622e5a833

[sc-107492]


mlir/lib/Quantum/IR/QuantumOps.cpp
```diff
@@ -325,6 +322,25 @@ LogicalResult QubitUnitaryOp::verify()
...
 // ----- measurements
 
+static LogicalResult verifyInQNodeFunction(Operation *op)
+{
+    auto parentModule = op->getParentOfType<ModuleOp>();
+    auto rootModule = parentModule->getParentOfType<ModuleOp>();
+    bool inQuantumKernel = parentModule && rootModule;
+    if (!inQuantumKernel) {
+        return success(); // strict verification only for quantum kernels
+    }
```
paul0403, line 336: Just for my understanding, this is so that after `--inline-nested-modules` you don't want the verifier to run anymore, and this is because 

> To limit the scope of this PR the new attribute is replaced with the previous qnode attribute after lowering the quantum kernel module.

?

mlir/lib/Quantum/Transforms/split_to_single_terms.cpp
```diff
@@ -485,7 +485,7 @@ struct SplitToSingleTermsPass : public impl::SplitToSingleTermsPassBase<SplitToS
         removeDeadOpsBeforeOp(origFunc, callOp.getOperation(), /*reserveDeviceOps=*/false);
 
         // Remove qnode attribute
-        origFunc->removeAttr("qnode");
+        origFunc->removeAttr("quantum.node");
```
paul0403, line 491: Note that there's also a line like this in the new `--dynamic-one-shot`, probably got merged after you created this branch

https://github.com/PennyLaneAI/catalyst/blob/d5ee2b4be9b073a28f3cdca5dd26ad7fc03f8c98/mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp#L1079

mlir/test/Catalyst/Async.mlir
```diff
@@ -146,37 +146,34 @@ module @workflow {
 
 // Test to make sure that async is placed before uses even in the presence of control flow.
 
-module @bar {
```
paul0403, line 149: Why was this test impacted? Was this what fueled the "parent module of parent module" logic in the new verifier for the MPs 😅 ?

mlir/test/Quantum/VerifierTest.mlir
```diff
@@ -473,3 +473,36 @@ func.func @state_good(%q0 : !quantum.bit, %q1 : !quantum.bit, %c : i64, %in_stat
...
+    %obs = quantum.namedobs %q[PauliZ] : !quantum.obs
+    // expected-error@+1 {{requires parent function to carry 'quantum.node' attribute}}
+    %exp = quantum.expval %obs : f64
+    return
+}
+}
+
+// -----
+
+// COM: e.xpected-error @below {{attribute 'quantum.node' requires at least one measurement process operation in the function body}}
```
paul0403, line 508:
```suggestion
// COM: expected-error @below {{attribute 'quantum.node' requires at least one measurement process operation in the function body}}
```

typo?

mlir/test/Quantum/SplitMultipleTapesTest.mlir
```diff
@@ -22,17 +22,29 @@
...
+    %qreg = quantum.alloc( 1) : !quantum.reg
+    %obs = quantum.compbasis qreg %qreg : !quantum.obs
+    %sample = quantum.sample %obs : tensor<1x1xf64>
+    quantum.dealloc %qreg : !quantum.reg
     quantum.device_release
     %result = stablehlo.subtract %tape0_out, %tape1_out : tensor<f64>
+    %qreg_out = quantum.alloc( 1) : !quantum.reg
+    %obs_out = quantum.compbasis qreg %qreg_out : !quantum.obs
+    %sample_out = quantum.sample %obs_out : tensor<1x1xf64>
+    quantum.dealloc %qreg_out : !quantum.reg
```
paul0403: This is not necessary to add right now right? Since you removed the verification that qnode functions must have a MP.

Also, let's say even if that restriction is in place. Why does the classical post processing after the two tape functions (which are themselves qnodes) needs a MP too? 🤔 Is it just because after create the tape func ops, the qnode attribute on the original function is still there?

paul0403: Well, I'm not the biggest fan of the split tapes pass, but I'm not super worried about any hacks to need to do to it (if there's truly no other options), since it's exclusively a thing in legacy frontend only.

mlir/test/Quantum/VerifierTest.mlir
JerryChen97: This proves that the pr is not AI-generated (

dime10: That was a way to prevent a test failure from the absence of the error, but then I thought COM would be clearer and not mistaken for a typo (just forgot to remove the first fix) 😅

mlir/lib/Quantum/IR/QuantumOps.cpp
dime10: This is because there are passes further down that won't respect the model. So I limited the scope of the verifier to the quantum kernel abstraction, rather than overhauling everything in one big PR. 

Keeping the existing qnode attribute is a result of the above, not the other way around.

mlir/lib/Quantum/Transforms/split_to_single_terms.cpp
dime10: yeah makes sense

mlir/test/Catalyst/Async.mlir
dime10: Yeah unfortunately tests that nest modules would trigger the verifier. There isn't really another reliable method that I could think of though since we don't have a separate op for the quantum kernel :(

mlir/test/Quantum/SplitMultipleTapesTest.mlir
dime10: Omg you're right, that's a mistake, since there is no active device!

dime10:
```suggestion
```

mlir/test/Quantum/VerifierTest.mlir
dime10: Turns out the COM does not prevent the diagnostic check from expecting the error, seems like the dot is required after all

mlir/test/Quantum/SplitMultipleTapesTest.mlir
```diff
@@ -68,30 +76,54 @@ module @circuit_twotapes_module {
...
```
paul0403: Remember to remove the quantum ops in these "post processing" blocks for the other split tape tests as well!

dime10: [15a864f](https://github.com/PennyLaneAI/catalyst/pull/2497/commits/15a864f952388da16d84372343d6ba499a0454da)

mlir/lib/Quantum/Transforms/split_to_single_terms.cpp
dime10: I updated those but I also found that the pass wasn't attaching the attribute to the newly created function: [fe6396d](https://github.com/PennyLaneAI/catalyst/pull/2497/commits/fe6396d470b3ba8a54fc6d235cfc12f4e3e4b62e)

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
```diff
@@ -1076,7 +1077,7 @@ struct DynamicOneShotPass : public impl::DynamicOneShotPassBase<DynamicOneShotPa
         // For each qnode function, find the returned MPs
         // Then handle the one shot logic for each MP type
         for (auto qnodeFunc : qnodeFuncs) {
-            qnodeFunc->removeAttr("qnode");
+            qnodeFunc->removeAttr("quantum.node");
```
paul0403, line 1082: While we are here, we might as well change the detection logic above to just check for the attribute, now that the pass guide is officially merged in.

https://github.com/PennyLaneAI/catalyst/pull/2458#discussion_r2790225394

mlir/lib/Quantum/Transforms/split_to_single_terms.cpp
paul0403: Nice catch!

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
dime10: Do you not want to make the pass a module pass btw (like the guide suggests)? The variable is already called `mod` but there is no guarantee of this right now. Due to the walk the detection code is fairly flexible, but to remove it it would be easier to assume we are actually looking at a module.

paul0403: I think it needs to be a module pass right? Since it is creating functions in the quantum kernel module, outside the original qnode function.

Is it not a module pass now?

dime10: I'll convert it. Right now it's an op-agnistic pass, so it can be scheduled on any op type.

dime10: [4a3cc84](https://github.com/PennyLaneAI/catalyst/pull/2497/commits/4a3cc84cbf520dfedaec4a15cf1bea507e4d798e)

paul0403 (approved):

mlir/lib/Quantum/IR/QuantumDialect.cpp
```diff
@@ -114,6 +114,34 @@ void QuantumDialect::initialize()
...
+    if (attrName != "quantum.node") {
+        return success();
+    }
+
+    if (!isa<func::FuncOp>(op)) {
+        return op->emitOpError() << "attribute '" << attrName << "' is only valid on 'func.func'";
+    }
+
+    if (!isa<UnitAttr>(namedAttr.getValue())) {
+        return op->emitOpError() << "attribute '" << attrName << "' must be a unit attribute";
```
mehrdad2m, line 130: Can we add a test for this?

mlir/lib/Quantum/IR/QuantumOps.cpp
```diff
@@ -325,6 +326,25 @@ LogicalResult QubitUnitaryOp::verify()
...
+
+    auto parentFunc = op->getParentOfType<func::FuncOp>();
+    if (!parentFunc) {
+        return op->emitOpError("must be nested inside a 'func.func' operation");
```
mehrdad2m, line 340: Same here

mlir/test/Quantum/StaticAllocaTest.mlir
```diff
@@ -15,143 +15,119 @@
...
+  cf.br ^bb1
+^bb1:
+  %q1, %q2 = quantum.unitary(%arg0 : memref<2x2xcomplex<f64>>) %arg1 ctrls(%arg2) ctrlvals(%arg3) : !quantum.bit ctrls !quantum.bit
+  return
 }
 
 // -----
 
 // CHECK-LABEL: @static_alloca_hermitian
-module @static_alloca_hermitian {
```
mehrdad2m, line 58: Is this for being consistent with the other tests? Cause I don't think this would trigger the verifier right?

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
```diff
@@ -1062,21 +1063,23 @@ struct DynamicOneShotPass : public impl::DynamicOneShotPassBase<DynamicOneShotPa
...
+            if (func->hasAttrOfType<UnitAttr>("quantum.node")) {
+                qnodeFuncs.push_back(func);
+            }
+        }
 
         // For each qnode function, find the returned MPs
         // Then handle the one shot logic for each MP type
         for (auto qnodeFunc : qnodeFuncs) {
-            qnodeFunc->removeAttr("qnode");
+            qnodeFunc->removeAttr("quantum.node");
```
mehrdad2m, line 1082: I am not very familiar with the implementation of the pass but I am wondering if removing the attribute here would cause the new verifier to fail on the transformed module?

mehrdad2m (approved):
Thanks @dime10, I just have some minor comments.

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
paul0403: This is fine, with one-shot, the entry point function (which was a qnode) will have its role replaced, essentially becoming the "main" function, which is completely classical. The qnode function is cloned into another function.

paul0403: [example](https://github.com/PennyLaneAI/catalyst/blob/ab780961fafbaee58c2bd6a20a39df72b48240ce/mlir/test/Quantum/DynamicOneShotTest.mlir#L18)

paul0403: But this does remind me, @dime10 is the attribute added to the cloned quantum function? The one called `one_shot_kernel`. I don't remember off the top of my head and seems like you didn't need to add the attribute to these kernels in the updated tests :thinking:

mehrdad2m: yeah but as you mentioned, I think it should be tranferred to the cloned function that has the quantum ops. In the example you shared: 
```
// CHECK: func.func public @test_expval.quantum.one_shot_kernel(%arg0: f64) -> f64
// CHECK:  [[one:%.+]] = arith.constant 1 : i64
// CHECK:  quantum.device shots([[one]]) ["", "", ""]
// CHECK:  quantum.expval
```
I think verifier would fail on `quantum.expval`

mlir/lib/Quantum/IR/QuantumDialect.cpp
dime10: Sure thing, thanks for catching that

mlir/test/Quantum/StaticAllocaTest.mlir
dime10: Yeah this test was probably just for consistency since I was changing basically the whole file

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
dime10: Let me double check whether the attribute is transferred, you'd think it would be otherwise we'd have verifier failures, but I'll confirm

paul0403: I'm pretty sure it will be transferred, the one shot kernel is created by calling `clone()` on the original qnode function, and the attributes are moved over. But thanks for double checking!

It didn't show on the example I shared because it was before this PR no? So the original qnode function also didn't have the attribute :sweat_smile:

dime10: You can see the attribute is being set in the one shot kernel (the new qnode with 1 shot), which I actually fixed earlier in [this commit](https://github.com/PennyLaneAI/catalyst/pull/2497/commits/fe6396d470b3ba8a54fc6d235cfc12f4e3e4b62e).

Let me know if this answers your question :)

mlir/lib/Quantum/IR/QuantumOps.cpp
dime10: https://github.com/PennyLaneAI/catalyst/pull/2497/commits/f4de94150312829b40fd30f908ea8e88f5bcef8d

mlir/lib/Quantum/IR/QuantumDialect.cpp
dime10: https://github.com/PennyLaneAI/catalyst/pull/2497/commits/f4de94150312829b40fd30f908ea8e88f5bcef8d

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
mehrdad2m: nice, thanks for checking. I think I got confused by the lack of them in the dynamic-one-shot test which was probably made before this.

mehrdad2m (approved):

mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
dime10: Oh I see, I've added the attribute to the DOS checks, thanks :) https://github.com/PennyLaneAI/catalyst/pull/2497/commits/452e3fdec1e6b3861a433ef2fa1452d03ab3885a

PR #2505, opened 2026-02-23
update scipy-openblas32 version

mehrdad2m:
**Context:**
As part of dependancy updates, bump the version of scipy-openblas32

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -209,7 +209,7 @@
   * The Enzyme version has been updated to
   [v0.0.238](https://github.com/EnzymeAD/Enzyme/releases/tag/v0.0.238).
 
-* The version of scipy-openblas32 used by Catalyst has been updated to v0.3.31.
+* The version of `scipy-openblas32` used by Catalyst has been updated to the latest version (v0.3.31).
   [(#2505)](https://github.com/PennyLaneAI/catalyst/pull/2505)
 
 * When an integer argnums is provided to `catalyst.vjp`, a singleton dimension is now squeezed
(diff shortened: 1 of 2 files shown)


doc/releases/changelog-dev.md
```diff
@@ -209,6 +209,9 @@
   * The Enzyme version has been updated to
   [v0.0.238](https://github.com/EnzymeAD/Enzyme/releases/tag/v0.0.238).
 
+* The version of scipy-openblas32 used by Catalyst has been updated to v0.3.31.
```
```suggestion
* The version of `scipy-openblas32` used by Catalyst has been updated to the latest version (v0.3.31).
```

setup.py
```diff
@@ -128,7 +128,7 @@ def parse_dep_versions():
     f"jax=={jax_version}",
     f"jaxlib=={jax_version}",
     "numpy>2.0.0",
-    "scipy-openblas32>=0.3.26",  # symbol and library name
+    "scipy-openblas32>=0.3.31",  # symbol and library name
```
maliasadi, line 131: do we need to update docs?

maliasadi (approved):

setup.py
mehrdad2m: I didn't find anywhere in the docs that this package was referenced.

dime10, line 131: Why bump this at all if it is a minimum version? Unless we have a new incompatibility, by otherwise it seems unnecessary

mehrdad2m: No specific reason on the Catalyst side. This was a request by the lightning team as they are updating their dependancy and this is for the sake of consistency. Maybe @maliasadi has more context on why we want this bump. cc @josephleekl

maliasadi: We don't really need this bump as long as you are not pinning to a specific version on CIs. 

For the context, we are hard-pinning to an older version of this lib in Lightning. @josephbowles is updating the pinned version to ensure that we're building/testing with the latest versions and the version is compatible across L-C

josephleekl: actually lightning also doesn't hard pin it, only `scipy-openblas32>=0.3.26`. So I guess there's no action for both repos? @maliasadi

PR #2508, opened 2026-02-23
Support lowering from MCM to OpenAPL's MeasurePulse

rniczh:
**Context:**

Add support for **mid-circuit measurement (MCM)** on the OQD backend via OpenAPL.  `qml.measure` is lowered to OpenAPL **MeasurePulse** entries and `ion.readout` (didn't has way to lower to the artiq for now) so that fluorescence detection can be executed by the hardware at runtime.

**Measurement Support:**

- **`measurement_duration`** is set to 1e-4 seconds for fluorescence detection
- **`[[detection_beam]]`**: Add section defining detection beam(s) for measurement in `gate.toml`. Defined by OQD team.

We added a new op `ion.measure_pulse` for representing a fluorescence measurement pulse and a `ion.readout` operations for read the classical result from the result of measure pulse. `ion.measure_pulse` will be used inside `ion.parallelprotocol` like regular pulses. And added the `__catalyst__oqd__measure_pulse`, executing it to convert the operation to OpenAPL at runtime. 

Now `Measure(0)` in test for OQD is expected to return a non-null placeholder result (`false`) instead of throwing.

**Detuning in new testbench:**

The algorithm for the MS gate was changed: the old code computed per-qubit detunings using phonon data, while the new code reads pre-computed beam values directly from the TOML calibration database. So no need to read the phonon data anymore. 

e.g.
Pulse | Old code | New code
-- | -- | --
Pulse2 (qubit0 blue sideband) | base + phonon_qubit0_ComX | beams2[baseIdx+2].detuning
Pulse5 (qubit1 blue sideband) | base + phonon_qubit1_ComX | beams2[baseIdx+2].detuning

**Memory Issues:**

Fix memory issue, the Gate to Pulse transform during the OQD pipeline cost a huge amount memory due to the SCFToControlFlow pass. We fixed it by folding the angle constants, If the angle is a compile-time constant, fold the normalisation and return a single `arith.constant`. This avoids emitting one `scf.if` block per gate, which causes `O(N_gates)` basic-block explosion in SCFToControlFlow pass and the downstream LLVM passes.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-112277]


paul0403 (approved):

mlir/include/Ion/Transforms/oqd_database_types.hpp
```diff
@@ -37,6 +37,25 @@ struct Beam {
...
+// Detection beam for fluorescence measurement. Extends Beam with transition.
+// transition_index maps to LevelTransition:
+//   0=downstate->estate
+//   1=downstate->estate2
+//   2=upstate->estate
+//   3=upstate->estate2
+struct DetectionBeam {
+    double rabi, detuning;
+    std::vector<int64_t> polarization, wavevector;
+    int transition_index;
```
mehrdad2m: This seems to be very similar to the existing Beam except for the `transition_index` I am wondering if we can avoid the duplication of code somehow. Maybe we can extend Beam using `std::optional<int>` ?

mlir/include/Ion/IR/IonOps.td
```diff
@@ -200,6 +200,52 @@ def PulseOp : Ion_Op<"pulse"> {
     }];
 }
 
+def MeasurePulseOp : Ion_Op<"measure_pulse"> {
```
mehrdad2m: it seems to me that this operation is exactly the same as PulseOp except for the name. In tat case, we should be able to avoid the duplication by adding a base op for both to inherit from.

mlir/lib/Ion/Transforms/GatesToPulsesPatterns.cpp
```diff
@@ -603,18 +502,143 @@ struct GatesToPulsesRewritePattern : public mlir::OpConversionPattern<CustomOp>
...
+        : mlir::OpRewritePattern<MeasureOp>(ctx)
+    {
+        detectionBeams = dataManager.getDetectionBeamParams();
+        measurementDuration = dataManager.getMeasurementDuration();
+    }
+
+    LogicalResult matchAndRewrite(MeasureOp op, PatternRewriter &rewriter) const override
+    {
+        if (detectionBeams.empty()) {
+            return failure();
```
mehrdad2m, line 575: can we add an `op.emitError()` here?

mehrdad2m (approved):
🚀

mlir/include/Ion/Transforms/oqd_database_types.hpp
rniczh: Thanks for the comments 👍 ! I refactored the code to reuse the beam.  https://github.com/PennyLaneAI/catalyst/pull/2508/commits/7d4d8ce82302c48a58c75036253d327908ce5594

mlir/include/Ion/IR/IonOps.td
rniczh: Thanks! I created a BasePulseOp in this comment to let both PulseOp and MeasurePulseOp to inherit from.
https://github.com/PennyLaneAI/catalyst/pull/2508/commits/7d4d8ce82302c48a58c75036253d327908ce5594

mlir/lib/Ion/Transforms/GatesToPulsesPatterns.cpp
rniczh: Sure, updated here https://github.com/PennyLaneAI/catalyst/pull/2508/commits/7d4d8ce82302c48a58c75036253d327908ce5594

mehrdad2m (approved):

PR #2511, opened 2026-02-23
Remove condition operand from PPMeasurementOp and lower conditional PPR decomposition via scf.if

### LINKED ISSUE #2475: Support unroll conditional PPM

sengthai:
**Context:**
The conditional PPM performs measurement if the condition is satisfied. This PPM returns whether 0 or 1 as `i1` type. However, if the condition is `false`, the PPM still returns 0, which can easily confuse whether it is a measurement result. Simply, it is illegal state (a measurement result without a measurement). Removing it makes the illegal state unrepresentable.

One solution that is discussed in [here](https://github.com/PennyLaneAI/catalyst/pull/2490#discussion_r2831158981) is to use a **poison value**, which solves the measurement-result-without-a-measurement, yet still produces an SSA measurement result that escapes into the wider IR. Poison makes the "don't use this" contract explicit, but the problematic SSA value still exists and flows through the program. 

Why did it exist in the first place?
It exists only because `DecomposeCliffordPPR` chose to propagate the condition forward rather than wrapping the decomposition in control flow, and this choice forces every downstream pass to either handle or explicitly reject conditioned PPMs. 

To sum up, removing conditional PPM makes the illegal state unrepresentable, which is the best solution here.

- Removed `condition` operand from `PPMeasurementOp`
- Updated the `decompose-clifford-ppr` pass so that conditional pi/4 PPR decompositions are now emitted inside an `scf.if ` region instead of using conditional PPM. 
- Cleaned up related builders, assembly format, Python/XDSL dialect definition, tests, and documentation.

**Related GitHub Issues:**
close https://github.com/PennyLaneAI/catalyst/issues/2475
close https://github.com/PennyLaneAI/catalyst/pull/2490
close [[sc-111303]]


mlir/test/PBC/PPRToPPM.mlir
```diff
@@ -93,9 +93,14 @@ func.func @test_ppr_to_ppm_with_condition(%q0 : !quantum.bit, %q1 : !quantum.bit
...
-    // CHECK: {{.*}} = pbc.ppr ["X", "Y", "Z"](2) {{.*}} cond([[q_1]]) : !quantum.bit
+    // CHECK: [[if:%.+]]:3 = scf.if [[m0]]
+    // CHECK:   [[q_0:%.+]] = quantum.alloc_qb : !quantum.bit
+    // CHECK:   [[m1:%.+]], [[o1:%.+]]:4 = pbc.ppm ["X", "Y", "Z", "Y"](-1)
+    // CHECK:   [[m2:%.+]], {{.*}} = pbc.ppm ["X"] [[o1]]#3
+    // CHECK:   [[q_1:%.+]] = arith.xori [[m1]], [[m2]] : i1
+    // CHECK:   {{.*}} = pbc.ppr ["X", "Y", "Z"](2) {{.*}} cond([[q_1]])
+    // CHECK:   scf.yield
+    // CHECK: } else {
+    // CHECK:   scf.yield
```
paul0403: Can you check the yielded results here too?

Note that in general, checking for names like `%arg0`, `%arg1`, etc in the lit test is fine, since these arguments are guaranteed names, unlike op results.

doc/releases/changelog-dev.md
```diff
@@ -270,6 +262,12 @@
 
 <h3>Internal changes ⚙️</h3>
 
+* Removed the `condition` operand from `pbc.ppm` (Pauli Product Measurement) operations.
+  Conditional PPR decompositions in the `decompose-clifford-ppr` pass now emit the
+  measurement logic inside an `scf.if` region rather than propagating the condition
+  to inner PPM ops.
+  [(#XXXX)](https://github.com/PennyLaneAI/catalyst/pull/XXXX)
```
paul0403: Remember to update the PR number

paul0403 (approved):
Nice, just a small comment 💯

mlir/test/PBC/PPRToPPM.mlir
sengthai: Thanks Paul! I improved the test here.

[e2465c6](https://github.com/PennyLaneAI/catalyst/pull/2511/commits/e2465c623730eeda3a12ccbbe7f35615f7dc572b)

doc/releases/changelog-dev.md
sengthai: Done in here!

[e2465c6](https://github.com/PennyLaneAI/catalyst/pull/2511/commits/e2465c623730eeda3a12ccbbe7f35615f7dc572b)

lazypanda10117 (approved):
Looks good, thanks!

PR #2512, opened 2026-02-24
Add the QEC Logical dialect (MLIR and xDSL)

joeycarter:
**Context:** This is the first in a series of PRs to build a prototype QEC compilation pipeline in Catalyst. The initial prototype will consider the following layers of abstraction, from top to bottom: _abstract_ (`quantum` dialect), _QEC logical_ (`qecl` dialect; this PR), and _QEC physical_ (`qecp` dialect; upcoming PR).

**Description of the Change:** This PR adds the MLIR QEC logical dialect, called `qecl`. This is an initial PR that adds all of the dialect boilerplate and the two fundamental types of the QEC logical program representation: `!qecl.codeblock` and  `!qecl.hyperreg`. Operations within the `qecl` dialect that act on these types will be added in an upcoming PR. A mirror of this MLIR dialect has also been added as an xDSL dialect in the Python interface to Catalyst.

> [!NOTE]

[sc-111760]


mlir/include/QecLogical/IR/QecLogicalDialect.h
```diff
@@ -0,0 +1,25 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include "mlir/IR/Dialect.h"
+
+#include "QecLogical/IR/QecLogicalTypes.h"
```
sengthai, line 19: I'm unsure if this is a naming convention in Catalyst, but it appears that all acronyms are written in uppercase on folder names, classes, or structs, such as `MBQC`, `PBC`, or `RTIO`. Since `QEC` is also an acronym, I recommend using uppercase for it as well, like `QECLogical` and `QECPhysical`.

joeycarter: For single-acronym names, I agree that `QEC`, `MBQC`, etc. are fine, but for long names containing a mix of acronyms and non-acronyms, I find the style `ABCCamelCase` hard on the eyes: where does the acronym end and the next word begin? In the absence of an established convention I would prefer to keep this as `QecLogical`.

sengthai: ohh my bad, i've just learned these keywords (QEC, MBQC.. ) are `Initialisms`. And i noticed in MLIR dialect folder that most Initialisms + non-initialisms are used `ABCCamelCase` style (`GPUTransformOps.cpp`, `MPIOps.cpp`, `MBQCInterfaces.td` ). And you are right about the acronym and shortening type, they use `CamelCase` like (`ArmNeon`, `ArmSVE`, `PtrDialect.cpp`...). So in here, i think we still should use `ABCCamelCase` style. 🧐

frontend/test/pytest/python_interface/dialects/test_qecl_dialect.py
```diff
@@ -0,0 +1,88 @@
...
+        expected_name is not None
+    ), f"Unexpected attribute {attr_class_name} found in QecLogical dialect"
+    assert attr.name == expected_name
+
+
+@pytest.mark.parametrize(
+    "pretty_print", [pytest.param(True, id="pretty_print"), pytest.param(False, id="generic_print")]
+)
+def test_assembly_format(run_filecheck, pretty_print):
+    """Test the assembly format of the mbqc ops."""
```
lillian542: so these ops look the same with or without `pretty_print`?

lillian542 (approved):
Thanks @joeycarter, looks good to me! 

I have no opinion on the CamelCase discussion, happy to go with whatever you two settle on there.

frontend/test/pytest/python_interface/dialects/test_qecl_dialect.py
joeycarter: Not quite, their semantics are the same with or without `pretty_print`, but the way they're represented in the IR text is different. With pretty print, we use the op's custom assembly format, e.g.

```mlir
%block0 = qecl.extract_block %hyperreg[0] : !qecl.hyperreg<3 x 1> -> !qecl.codeblock<1>
```

Without pretty print, we use the op's generic assembly format, e.g.

```mlir
%block0 = "qecl.extract_block"(%hyperreg) <{idx_attr = 0 : index}> : (!qecl.hyperreg<3 x 1>) -> !qecl.codeblock<1>
```

These ops both mean the same thing in the IR, but the generic format is more verbose so we generally only use it for testing (like we're doing here), or for other niche applications.

dime10: One thing that's cool about the generic format is that it's isomorphic to the MLIR data structures in memory, so can be processed by any MLIR tool whether it was compiled with knowledge of the Catalyst dialects or not.

The pretty format is generally easier on the eyes for humans though.

joeycarter: :point_up:

multiphaseCFD: very minor, i believe it's for codeblock ops instead of mbqc

multiphaseCFD (commented):
Thanks @joeycarter. Just one comment. Happy to approve.

frontend/test/pytest/python_interface/dialects/test_qecl_dialect.py
joeycarter: Agh, sorry, this mistake got copied around a couple of times! Fixed: 342896448

mlir/include/QecLogical/IR/QecLogicalDialect.h
joeycarter: Hey @sengthai, I'm going to resolve this thread. I understand that initialisms are generally capitalized in LLVM for CamelCase names, but I know of no written guidelines stating that this is a strict requirement. This dialect is experimental, and in the interest of speed (and also for the readability reasons I mentioned above) I'm going to keep the naming conventions as they are.

multiphaseCFD (approved):

PR #2514, opened 2026-02-24
Target decomposition rules via `decomposition_rule` attribute

kipawaa:
**Context:**
Currently, graph decomposition rules are targeted by finding the `target_gate` attribute. This name is very generic, and may want to be re-used by other systems in the future. 

The `decomposition_rule` attribute is set to `True` (python) and `true` (mlir). There is no change to the `target_gate` attribute (it is still present on decomposition rules and is used exactly the same as before).

**Benefits:**
- Explicit decomposition rule detection.
- Removes the potential for detecting other systems functions with the `target_gate` attribute.
- Unambiguous attribute names

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A

[sc-112459]

### DIFF
--- a/frontend/catalyst/from_plxpr/decompose.py
+++ b/frontend/catalyst/from_plxpr/decompose.py
@@ -345,6 +345,7 @@ def _create_decomposition_rule(
     # Set custom attributes for the decomposition rule
     # These attributes are used in the MLIR decomposition pass
     # to identify the target gate and the number of wires
+    setattr(func_cp, "decomposition_rule", True)
     setattr(func_cp, "target_gate", op_name)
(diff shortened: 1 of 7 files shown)


maliasadi (commented):
Thanks @kipawaa! Giving it more thought...

- `target_gate` is only used by decomposition rules now and there is no actual collision. Where do you see these rules being used elsewhere? My feeling here is that we introduce this now and YAGNI!
- What makes a decomposition rule special except it targets a gate (`target_gate`)?

maliasadi:
re: PR  https://github.com/PennyLaneAI/catalyst/pull/2531

PR #2518, opened 2026-02-25
fix: improve testing of `setup_options` in our unified transforms API

andrijapau:
**Context:**

I was investigating https://app.shortcut.com/xanaduai/story/104262/polish-up-handling-of-options-and-valued-options?vc_group_by=day&ct_workflow=all&cf_workflow=500000005 and noticed we can improve our testing a bit here.

Adds more coverage for different common set-up options for transforms.

**Benefits:**

Better testing.

[sc-112601]

### DIFF
--- a/frontend/test/pytest/test_transform_pass_name.py
+++ b/frontend/test/pytest/test_transform_pass_name.py
@@ -16,10 +16,10 @@
 from functools import partial
 from typing import Type
 
-from catalyst.compiler import CompileError
-
 import pennylane as qml
 import pytest
+
+from catalyst.compiler import CompileError
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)


frontend/test/pytest/test_transform_pass_name.py
```diff
@@ -12,29 +12,77 @@
...
+    @partial(my_pass, **options)
+    @qml.qnode(qml.device(backend, wires=1))
+    def captured_circuit():
+        return qml.expval(qml.PauliZ(0))
+
+    expected_error = CompileError if options["option"] is None else TypeError
+    expected_msg = (
+        r"Cannot convert Python type <class 'NoneType'> to an MLIR attribute"
+        if options["option"] is None
+        else "unhashable type"
```
andrijapau, line 82: Not sure if we want to improve upon the error messaging here but that's what we currently get!

JerryChen97 (approved):
Thanks for fledging it! It's a good start

frontend/test/pytest/test_transform_pass_name.py
albi3ro: Maybe create a follow up?

albi3ro (approved):

frontend/test/pytest/test_transform_pass_name.py
andrijapau: Done [sc-112600]

andrijapau: Marked it as a P2

PR #2520, opened 2026-02-26
Make `pbc.prepare` create its own qubits

dime10:
To improve the lowering of the PBC dialect, the `prepare` op will follow the same format as the fabricate op and allocate its own qubits rather than act on existing qubits. The "reset" mode of this operation was so far unused, that is it had always been applied immediately after an `alloc_qb` operation.


sengthai (approved):
Thanks @dime10! 💯

lazypanda10117 (approved):
Thanks @dime10!

PR #2524, opened 2026-02-26
Decouple `shots` and remove `ctx` from `QJITDevice.preprocess()`

JerryChen97:
## Summary
Refactor `QJITDevice.preprocess()` to align with the standard PennyLane `Device.preprocess(execution_config)` signature. This removes non-standard parameters (`ctx` and `shots`), ensuring full compatibility with the PennyLane plugin API.

## Key Changes
- **API Standardization**: Updated `QJITDevice.preprocess` to match the standard PennyLane signature by removing explicit `ctx` and `shots` arguments.
- **Shot Management**: Catalyst-specific shots are now passed via `execution_config.device_options` using a `catalyst_` prefix. The method strips these keys before forwarding to the underlying device to prevent validation errors.
- **Context Cleanup**: Removed the `ctx` parameter from the entire call stack, including `catalyst_decompose()` and `_decompose_nested_tapes()`.

## Technical Justification: Removing `ctx`
A thorough audit of the tracing stack confirmed that `ctx` was redundant:
1. **Constant Value**: `EvaluationContext._create_tracing_context()` yields `None`. State is managed via the class-level `_mode_stack`, not instance variables.
2. **Zero Dereferences**: While `ctx` was threaded through `jax_tracer.py`, `qjit_device.py`, and `decomposition.py`, no callsite ever accessed an attribute or method on it.
3. **Class-level Logic**: Relevant tracing operations (e.g., `frame_tracing_context`) are class methods that rely on `region.trace` rather than the `ctx` instance.

## Affected Files
- **`qjit_device.py`**: Updated `preprocess` signature and implemented `device_options` stripping.
- **`decomposition.py`**: Removed `ctx` from `catalyst_decompose` and `_decompose_nested_tapes`.
- **`jax_tracer.py`**: Cleaned up `device_options` encoding and removed `ctx` from quantum operation tracing.

**Ticket:** [sc-112654]

### DIFF
--- a/frontend/catalyst/device/qjit_device.py
+++ b/frontend/catalyst/device/qjit_device.py
@@ -383,12 +383,7 @@ class QJITDevice(qml.devices.Device):
 
         # Strip catalyst-specific keys before forwarding to the original device,
         # which may validate device_options and reject unknown keys.
-        clean_options = {
-            k: v
-            for k, v in execution_config.device_options.items()
-            if not k.startswith("catalyst_")
-        }
-        clean_config = replace(execution_config, device_options=clean_options)
+        clean_config = _strip_catalyst_options(execution_config)
 
         _, config = self.original_device.preprocess(clean_config)
 
@@ -663,3 +658,8 @@ def _requires_shots(capabilities):
         ExecutionCondition.FINITE_SHOTS_ONLY in MP_conditions
         for _, MP_conditions in capabilities.measurement_processes.items()
     )
+
+
+def _strip_catalyst_options(config):
+    clean = {k: v for k, v in config.device_options.items() if not k.startswith("catalyst_")}
+    return replace(config, device_options=clean)
--- a/frontend/catalyst/device/qjit_device.py
+++ b/frontend/catalyst/device/qjit_device.py
@@ -374,6 +374,7 @@ class QJITDevice(qml.devices.Device):
         which are created based on what is both compatible with Catalyst and what is supported the
         backend according to the backend TOML file).
         """
+        # pylint: disable=unused-argument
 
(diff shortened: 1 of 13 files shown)


frontend/catalyst/device/qjit_device.py
```diff
@@ -371,7 +377,21 @@ def preprocess(
 
         if execution_config is None:
             execution_config = qml.devices.ExecutionConfig()
-        _, config = self.original_device.preprocess(execution_config)
+
+        # Extract catalyst-specific options from device_options
+        ctx = execution_config.device_options.get("catalyst_ctx")
```
dime10: I thought ctx was superfluous now, or was that a different section of code?

dime10 (approved):
No objections to this :)

frontend/catalyst/device/qjit_device.py
JerryChen97: No idea yet lol, I can defeinitely also take a look into who actually used this ctx

dime10: There were a few discussions on slack recently pointing out that it's useless 😅

JerryChen97: @dime10 I removed it from preprocess as well as the classical tracer, and noone failed :sweat_smile: didn't touch the quantum part cause it doesn't look immediately deletable to me

dime10: Nice, yeah I didn't mean everywhere but mainly around the preprocess and device. If nothing failed should be all good :)

maliasadi (approved):
LGTM!

PR #2528, opened 2026-02-27
Unified Compiler: Fix a bug where the initial pass triggers the callback twice

jzaia18:
**Context:**
Currently, the unified compiler actually triggers the passed callback to `Compiler.run` twice for the pass at level 0. This does not affect the correctness of any code, but results in performance degradation whereby the same pass may be processed by a callback function multiple times.

Adds a clause to the line which triggers the initial pass an extra time to ensure it is not extraneously called.

**Benefits:**
Features which make use of the unified compiler callback function (such as `specs` or `draw_graph`) will no longer be triggered twice on the initial pass, improving overall performance.

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-112748]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -232,6 +232,9 @@
 
 <h3>Bug fixes 🐛</h3>
 
+* Fixed a bug where the unified compiler would trigger a passed callback function 1 extra time for the initial pass level.
+  [(#2528)](https://github.com/PennyLaneAI/catalyst/pull/2528)
+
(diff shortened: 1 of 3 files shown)


mudit2812 (commented):
Thanks @jzaia18 . Can you add testing to [this file](https://github.com/PennyLaneAI/catalyst/blob/e16c714e411387b69321e97b2106886fd6276862/frontend/test/pytest/python_interface/pass_api/test_transform_interpreter.py#L284) as well?

mudit2812 (approved):

mehrdad2m (approved):

PR #2535, opened 2026-03-02
Add basic ops to the QEC logical dialect

joeycarter:

**Description of the Change:** This PRs adds basic operations to the QEC logical dialect, `qecl`, including:

* `qecl.alloc()`: allocate a logical hyper-register containing a sequence of logical codeblocks
* `qecl.dealloc`: deallocate a logical hyper-register
* `qecl.extract_block`: extract a logical codeblock from a hyper-register
* `qecl.insert_block`: insert a logical codeblock into a hyper-register

These ops closely follow the definition and implementation of their cousin `quantum`-dialect ops, except instead of acting on register of qubits, they act on hyper-registers of logical codeblocks. Like in the `quantum` dialect, these ops follow value semantics.

The equivalent operations have also been added to the xDSL python interface to Catalyst.

Verify, canonicalize and fold methods have also been added to the MLIR op definitions; they generally follow the same logic as the equivalent methods in the `quantum` dialect.

> [!NOTE]

[sc-112668]


frontend/catalyst/python_interface/dialects/qecl.py
```diff
@@ -87,10 +108,214 @@ def parse_parameters(cls, parser: Parser) -> list[IntegerAttr]:
...
+        """Verify the constraint and add resolved values to the ConstraintContext."""
+        constraint_context.set_attr_variable("hyper_reg_type", attr)
+
+    # pylint: disable=unused-argument
+    def can_infer(self, var_constraint_names: Set[str]) -> bool:
+        """Check if there is enough information to infer the attribute given the constraint
+        variables that are already set.
+        """
+        # Assume we can always infer
+        return True
```
multiphaseCFD: I checked this method in the base class and by default, this method returns False. Want to understand if there is any impacts if we use the default

frontend/catalyst/python_interface/dialects/qecl.py
```diff
@@ -87,10 +107,175 @@ def parse_parameters(cls, parser: Parser) -> list[IntegerAttr]:
...
+
+        if isinstance(idx, IntegerAttr):
+            operands = (hyper_reg, None)
+            properties = {"idx_attr": idx}
+        else:
+            operands = (hyper_reg, idx)
+            properties = {}
+
+        hyper_reg_type = get_logical_hyper_reg_type(hyper_reg)
+        result_type = LogicalCodeblockType(k=hyper_reg_type.k)
```
multiphaseCFD, line 218: So the `k` is passed from the `Hyper_Reg` to the `LogicalCodeBlock`? Make sense for the `qecl` dialect as the QEC code is supported to be known at this stage.

frontend/catalyst/python_interface/dialects/qecl.py
```diff
@@ -87,10 +107,175 @@ def parse_parameters(cls, parser: Parser) -> list[IntegerAttr]:
...
+
+
+@irdl_op_definition
+class ExtractCodeblockOp(IRDLOperation):
+    """Extract a logical codeblock value from a hyper-register."""
+
+    name = "qecl.extract_block"
+
+    assembly_format = """
+            $hyper_reg `[` ($idx^):($idx_attr)? `]` attr-dict `:` type($hyper_reg) `->` type($codeblock)
```
multiphaseCFD, line 191: why do we add `^` here?

frontend/catalyst/python_interface/dialects/qecl.py
```diff
@@ -87,10 +107,175 @@ def parse_parameters(cls, parser: Parser) -> list[IntegerAttr]:
...
+    else:
+        arg_type = arg.type
+        assert isinstance(
+            arg_type, expected_type
+        ), f"Expected value '{arg}' to have type '{expected_type.name}'"
+
+    return arg_type
+
+
+def get_logical_hyper_reg_type(
```
multiphaseCFD, line 145: Any consideration to call this wrapper instead of call the `_get_type_from_ssa_value_or_operation` directly?

multiphaseCFD (commented):
Nice work Joey! Looks good to me!

frontend/catalyst/python_interface/dialects/qecl.py
joeycarter: It looks like you were reviewing an old commit, this class is now removed. :slightly_smiling_face:

frontend/catalyst/python_interface/dialects/qecl.py
joeycarter: Maybe "passed" from the hyper-register to the codeblock is not the best word, but yes, they must agree. The QEC code does not necessarily need to be known at the QEC logical layer, but the value of *k* must.

joeycarter: Ah, but in the context of the op constructor, yes this value is passed to define the logical codeblock type. :slightly_smiling_face:

frontend/catalyst/python_interface/dialects/qecl.py
joeycarter: This is the syntax for optional elements in the assembly format; putting the `^` symbol marks that element as the "anchor" of the optional group. See here for more info: https://mlir.llvm.org/docs/DefiningDialects/Operations/#optional-groups

frontend/catalyst/python_interface/dialects/qecl.py
joeycarter: You mean, why create this wrapper function when we can just call `_get_type_from_ssa_value_or_operation()` directly? Mainly out of convenience and that it makes the op constructions more concise. You'll see in the upcoming PRs that I also add wrappers for codeblock types and qubits types in the QEC physical dialect. Separating them out by type makes it more expressive, in my opinion.

mlir/lib/QecLogical/IR/QecLogicalOps.cpp
```diff
@@ -26,3 +30,227 @@ using namespace catalyst::qecl;
...
+}
+
+/**
+ * @brief Canonicalize insert-codeblock op.
+ *
+ * Removes sequential extract-insert op pairs acting on the same index, e.g. handles the pattern:
+ *
+ *   %r0 = ... : !qecl.hyperreg<3 x 1>
+ *   %b0 = extract_block %r0[0]
+ *   %r1 = insert_block %r1[0], %b0
```
lillian542: shouldn't this be `insert_block %r0[0], %b0`?

lillian542 (approved):
Thanks Joey, LGTM!

mlir/lib/QecLogical/IR/QecLogicalOps.cpp
joeycarter: Yes, thank you! Fixed: 648342b44

multiphaseCFD (approved):
Thanks Joey! Nice work

PR #2537, opened 2026-03-02
Add basic ops to the QEC physical dialect

joeycarter:

**Description of the Change:** This PRs adds basic operations to the QEC physical dialect, `qecp`, including:

* `qecp.alloc()`: allocate a physical hyper-register containing a sequence of physical codeblocks
* `qecp.dealloc`: deallocate a physical hyper-register
* `qecp.alloc_aux`: allocate a single auxiliary physical qubit
* `qecp.dealloc_aux`: deallocate a single auxiliary physical qubit
* `qecp.extract_block`: extract a physical codeblock from a hyper-register
* `qecp.insert_block`: insert a physical codeblock into a hyper-register
* `qecp.extract`: extract a physical qubit from a codeblock
* `qecp.insert`: insert a physical qubit into a codeblock

These ops closely follow the definition and implementation of their cousin `quantum`-dialect ops, except instead of acting on register of qubits, they act on hyper-registers of physical codeblocks or on codeblocks containing physical qubits. Like in the `quantum` dialect, these ops follow value semantics.

The equivalent operations have also been added to the xDSL python interface to Catalyst.

Verify, canonicalize and fold methods have also been added to the MLIR op definitions; they generally follow the same logic as the equivalent methods in the `quantum` dialect.

> [!NOTE]

[sc-112669]


frontend/catalyst/python_interface/dialects/qecp.py
```diff
@@ -151,10 +172,321 @@ def parse_parameters(cls, parser: Parser) -> list[IntegerAttr]:
...
+        super().__init__(
+            operands=operands,
+            result_types=(result_type,),
+            properties=properties,
+        )
+
+
+@irdl_op_definition
+class InsertQubitOp(IRDLOperation):
+    """Update the physical qubit value of a codeblock."""
```
multiphaseCFD: only a data physical qubit insertion ops is allowed with `InsertQubitOp`, right? If yes, we might need to update the docstr here.

frontend/catalyst/python_interface/dialects/qecp.py
```diff
@@ -151,10 +172,321 @@ def parse_parameters(cls, parser: Parser) -> list[IntegerAttr]:
...
+        in_hyper_reg_type = get_physical_hyper_reg_type(in_hyper_reg)
+
+        super().__init__(
+            operands=operands, properties=properties, result_types=(in_hyper_reg_type,)
+        )
+
+
+@irdl_op_definition
+class ExtractQubitOp(IRDLOperation):
+    """Extract a physical qubit value from a codeblock."""
```
multiphaseCFD: Same as below

frontend/catalyst/python_interface/dialects/qecp.py
```diff
@@ -151,10 +172,321 @@ def parse_parameters(cls, parser: Parser) -> list[IntegerAttr]:
...
+            raise ValueError(
+                f"{self.name} op expected a qubit with role '{str(QecPhysicalQubitRole.Aux)}', "
+                f"but got '{qubit_type.role.data}'"
+            )
+
+        super().__init__(operands=(qubit,))
+
+
+@irdl_op_definition
+class ExtractCodeblockOp(IRDLOperation):
```
multiphaseCFD, line 305: Do we want to add ops that `alloc` and `dealloc` `codeblock` here? Good for now but it would be required for `PBC`.

frontend/catalyst/python_interface/dialects/qecp.py
```diff
@@ -151,10 +172,321 @@ def parse_parameters(cls, parser: Parser) -> list[IntegerAttr]:
...
+
+        super().__init__(
+            operands=operands,
+            result_types=(result_type,),
+            properties=properties,
+        )
+
+
+@irdl_op_definition
+class InsertCodeblockOp(IRDLOperation):
```
multiphaseCFD, line 348: Would it be a good practice to abstract an base insertop class and all other insertops can inherit from? Same thing for the extractop

mlir/lib/QecPhysical/IR/QecPhysicalOps.cpp
```diff
@@ -26,3 +29,431 @@ using namespace catalyst::qecp;
...
+        return success();
+    }
+
+    return failure();
+}
+
+/**
+ * @brief Canonicalize physical hyper-register deallocation op.
+ *
+ * Erase alloc/dealloc op pairs if allocated hyper-register is immediately deallocated.
```
multiphaseCFD, line 208: Could we update the docstr? It looks to me the code below erase the alloc/dealloc op pairs if an allocated hyper-register is only consumed by a dealloc op.

mlir/lib/QecPhysical/IR/QecPhysicalOps.cpp
```diff
@@ -26,3 +29,431 @@ using namespace catalyst::qecp;
...
+        return success();
+    }
+
+    return failure();
+}
+
+/**
+ * @brief Canonicalize aux qubit deallocation op.
+ *
+ * Erase alloc/dealloc op pairs if allocated aux qubit is immediately deallocated.
```
multiphaseCFD, line 242: same thing

frontend/catalyst/python_interface/dialects/qecp.py
joeycarter: Good point, thanks. Updated here: 9ddb0926e

frontend/catalyst/python_interface/dialects/qecp.py
joeycarter: Ditto: 9ddb0926e

mlir/lib/QecPhysical/IR/QecPhysicalOps.cpp
```diff
@@ -26,3 +29,431 @@ using namespace catalyst::qecp;
...
+        }
+    }
+
+    return failure();
+}
+
+/**
+ * @brief Canonicalize extract-codeblock op.
+ *
+ * Removes sequential insert-extract op pairs acting on the same index, e.g. handles the pattern:
```
multiphaseCFD, line 262: Is it possible that there is a case that: there are a few ops between a `insert_block`/`extract_block` pair?

mlir/lib/QecPhysical/IR/QecPhysicalOps.cpp
joeycarter: Not sure I follow, I believe the docstring matches the behaviour of the function. What would you propose modifying/adding?

frontend/catalyst/python_interface/dialects/qecp.py
joeycarter: We won't need these for the initial prototype, so let's not worry about that just yet. We can revisit this point when we start thinking about a PBC -> QEC conversion.

frontend/catalyst/python_interface/dialects/qecp.py
joeycarter: I'm on the fence about that. My take is that we should get the prototype compilation pipeline working first. When we start thinking about productionizing it, then we can refactor and make the dialect ops more modular. Is that all right with you?

mlir/lib/QecPhysical/IR/QecPhysicalOps.cpp
```diff
@@ -26,3 +29,431 @@ using namespace catalyst::qecp;
...
+ */
+template <typename IndexingOp> LogicalResult foldConstantIndexingOp(IndexingOp op, Attribute idx)
+{
+    bool hasNoIdxAttr = !op.getIdxAttr().has_value();
+    bool isConstantIdx = isa_and_nonnull<IntegerAttr>(idx);
+    if (hasNoIdxAttr && isConstantIdx) {
+        auto constantIdx = cast<IntegerAttr>(idx);
+        op.setIdxAttr(constantIdx.getValue());
+
+        // Remove the dynamic Value
```
multiphaseCFD, line 406: dynamic Value or dynamic value?

mlir/lib/QecPhysical/IR/QecPhysicalOps.cpp
joeycarter: Yes that's certainly possible, but that wouldn't trigger the canonicalization.

I should add, these `canonicalize` methods are essentially the same as what's currently defined in the `quantum` dialect.

multiphaseCFD (commented):
Thanks @joey! Nice work!

mlir/lib/QecPhysical/IR/QecPhysicalOps.cpp
joeycarter: Either or—`Value` is the class name in MLIR.

multiphaseCFD (approved):
LGTM! Thanks @joeycarter !

frontend/test/pytest/python_interface/dialects/test_qecp_dialect.py
```diff
@@ -93,6 +230,97 @@ def test_assembly_format(run_filecheck, pretty_print):
 
     // CHECK: [[hyperreg:%.+]] = "test.op"() : () -> !qecp.hyperreg<3 x 1 x 7>
     %hyperreg = "test.op"() : () -> !qecp.hyperreg<3 x 1 x 7>
+
+    // CHECK: [[hreg0:%.+]] = qecp.alloc() : !qecp.hyperreg<3 x 1 x 7>
```
lillian542, line 234: I don't know if this is allowed or difficult or a good idea, but I think to me this would be easier to read and keep track of which number is which if it printed as 

```
!qecp.hyperreg<3><1 x 7>
```
or similar to indicate a hyperreg with width 3, containing <1x7> codeblocks.

lillian542: I'm sure I'll get used to this format very quickly though, and it matches the logical dialect more closely. I think 
```
!qecl.hyperreg<3><1>
``` 
would look odd, so it would be hard to make them match if this were changed.

lillian542: It also strikes me as a bit odd because we can extract codeblocks using indices based on the first number, and from there we extract qubits based on the last number, but the middle number is never actually an index. This is not blocking and I don't have a suggestion for how else we might do it (I'm not really sure what the conventions and constraints on this kind of thing are), but it feels a little misleading to call it `3x1x7` when it behaves like it has a shape of `3x7`.

mlir/include/QecPhysical/IR/QecPhysicalDialect.td
```diff
@@ -47,7 +47,15 @@ def QecPhysical_Dialect : Dialect {
 
         #### Example
 
-        [TODO: Fill in example once the basic ops are defined (future PR)]
+        ```mlir
+        %0 = qecp.alloc() : !qecp.hyperreg<3 x 1 x 7>
+        %1 = qecp.extract_block %0[0] : !qecp.hyperreg<3 x 1 x 7> -> !qecp.codeblock<1 x 7>
+        %2 = qecp.extract %1[0] : !qecp.codeblock<1 x 7> -> !qecp.qubit<data>
```
lillian542: Could we extract the qubit at some other index instead? To make it unambiguous that we are extracting from the size-7 object, and not the size-1 object.

mlir/test/QecPhysical/CanonicalizationTest.mlir
```diff
@@ -0,0 +1,329 @@
...
+// CHECK-LABEL: test_alloc_aux_dce
+func.func @test_alloc_aux_dce() {
+    // CHECK-NOT: qecp.alloc_aux
+    %q = qecp.alloc_aux : !qecp.qubit<aux>
+    return
+}
+
+// -----
+
+// CHECK-LABEL: test_alloc_aux_no_cse
```
lillian542, line 69: I know `dce` stands for "dead code elimination", but what is `cse` an abbreviation for?

lillian542 (approved):
My only real comment is that I think the `hyperreg<3 x 1 x 7>` format for printing is a bit unintuitive to me, but I don't know how much freedom we have to change that, and I'll get used to it quickly. Not a blocking concern.

frontend/test/pytest/python_interface/dialects/test_qecp_dialect.py
joeycarter: It's an understandable concern. It's true that someone reading this might interpret it to mean that the `x` implies multiplication to get the total number of qubits represented by the type (which is true for logical hyper-registers and codeblocks, but not the case for physical hyper-registers and codeblocks where we include the value of _k_).

The type's assembly format can be more verbose and still represent the same information to make it more human-readable. For example,

```mlir
!qecl.hyperreg<3, k=1>
!qecl.codeblock<k=1>

!qecp.hyperreg<3, k=1, n=7>
!qecp.codeblock<k=1, n=7>
```

Or even something like:

```mlir
!qecl.hyperreg<3 x codeblock<k=1>>
!qecl.codeblock<k=1>

!qecp.hyperreg<3 x codeblock<k=1, n=7>>
!qecp.codeblock<k=1, n=7>
```

or more concisely,

```mlir
!qecl.hyperreg<3 x codeblock<1>>
!qecl.codeblock<1>

!qecp.hyperreg<3 x codeblock<1, 7>>
!qecp.codeblock<1, 7>
```

As far as using these types go, I see them as all the same, and my personal preference is to keep the type assembly format as concise as possible to reduce typing in tests and line lengths in IR output, especially since these types are going to appear in copious amounts in the IR for large programs.

But, that being said we should come to some consensus on the assembly format now before we start building everything else on top of these types.

mlir/test/QecPhysical/CanonicalizationTest.mlir
joeycarter: **C**ommon **S**ub-expression **E**limination: https://mlir.llvm.org/docs/Passes/#-cse

In short, if you have multiple, identical ops that have no memory side-effects, CSE will eliminate all but one of these ops, and reorganize things such that the ops that use the results of the eliminated op(s) now use the result of the op that is kept.

In this test, since we have two allocation ops which _do_ have memory side-effects, we want to ensure that CSE does not eliminate one of them.

mlir/include/QecPhysical/IR/QecPhysicalDialect.td
joeycarter: Done: 934ec9ebd

frontend/test/pytest/python_interface/dialects/test_qecp_dialect.py
joeycarter: Discussed offline. We will continue with this assembly format and revisit if necessary.

PR #2538, opened 2026-03-03
[xDSL] add non-commuting check for diagonalize_measurements pass

multiphaseCFD:
### Before submitting

**Context:**
[sc-93668]
This PR add non-commuting check for the `diagonalize_measurements` pass. An error will raise if the circuit (moduleop) is not qwc.  

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


doc/releases/changelog-dev.md
```diff
@@ -150,6 +150,7 @@
 
 * The `diagonalize-final-measurements` xDSL pass now accepts the optional keyword arguments ``to_eigvals`` and ``supported_base_obs``.
```
joeycarter: Do you also want to add that it now raises an error if the program contains measurements of non-commuting observables?

frontend/catalyst/python_interface/transforms/quantum/diagonalize_measurements.py
```diff
@@ -76,6 +74,59 @@ def _diagonalize(obs: NamedObsOp, supported_base_obs) -> bool:
...
 
+class NonCommutingObservableValidator:
+    """
+    Validates if all quantum observables in a builtin.ModuleOp object act on
+    disjoint qubit sets.
+    """
+
+    def __init__(self, ops: builtin.ModuleOp):
+        self.visited_qubits = set()
+        self.visited_qreg = False
```
joeycarter: I take it that this implementation assumes there is only ever one register in the program. Are we comfortable with this assumption? Does this assumption introduce another known limitation?

frontend/catalyst/python_interface/transforms/quantum/diagonalize_measurements.py
```diff
@@ -76,6 +74,59 @@ def _diagonalize(obs: NamedObsOp, supported_base_obs) -> bool:
     )  # pragma: no cover
 
 
+class NonCommutingObservableValidator:
```
joeycarter, line 80: I'm worried that this class name is a misnomer, since it doesn't actually check the commutation relations of the observables, only that no qubit/register is acted upon by multiple observable ops. If I have two observables `I` and `X` and acting on wire 0, for instance, they commute and can be measured simultaneously, but as is this implementation would raise an error if I'm not mistaken.

joeycarter (commented):
Thanks @multiphaseCFD, I had a couple of questions (see below).

The main one is, is the purpose of this PR to raise an error if there are non-commuting observables acting on the same wire(s), or if there are _any_ observables acting on the same wire(s)? It looks like you have the latter, which is not what the PR title is suggesting, so I wanted to confirm.

joeycarter:
> > Thanks @multiphaseCFD, I had a couple of questions (see below).
> > The main one is, is the purpose of this PR to raise an error if there are non-commuting observables acting on the same wire(s), or if there are _any_ observables acting on the same wire(s)? It looks like you have the latter, which is not what the PR title is suggesting, so I wanted to confirm.
> 
> Thanks @joeycarter! yes, it checks the [`wires` grouping strategy](https://docs.pennylane.ai/en/stable/code/api/pennylane.transforms.split_non_commuting.html).

Got it, thanks. In that case can you add a comment somewhere to that effect? Another thing you can do is to have the `NonCommutingObservableValidator` constructor take a `grouping_strategy` argument that for now is restricted to be `wires`; that should make it clearer to anyone else reading this what its intent is.

frontend/catalyst/python_interface/transforms/quantum/diagonalize_measurements.py
multiphaseCFD: Done

frontend/catalyst/python_interface/transforms/quantum/diagonalize_measurements.py
```diff
@@ -73,9 +75,88 @@ def _diagonalize(obs: NamedObsOp, supported_base_obs) -> bool:
...
+                        raise RuntimeError(self._error_msg)
+            else:
+                # Check overlapping qubits for conflicts (more than 1 non-Identity)
+                for q in self.overlapped_qubits:
+                    obs = self.obs_on_qubits[q]
+                    if len(obs - {"Identity"}) > 1:
+                        raise RuntimeError(self._error_msg)
+        except RuntimeError:
+            # Fallback to stricter check if QWC logic fails
+            self._run_non_overlapping_validation()
```
joeycarter: Minor, but since any `RuntimeError` that is raised is immediately caught, it would be better I think (and probably slightly more efficient) to use a bool flag, like `success`, which is set to `False` instead of raising a `RuntimeError`, and then replace the `except` block with

```python
if not success:
    self._run_non_overlapping_validation()
```

frontend/catalyst/python_interface/transforms/quantum/diagonalize_measurements.py
```diff
@@ -73,9 +75,88 @@ def _diagonalize(obs: NamedObsOp, supported_base_obs) -> bool:
...
+
+    def _run_non_overlapping_validation(self):
+        """Strictest check: no two observables can share a qubit."""
+        if self.overlapped_qubits or (self.visited_qreg and self.obs_on_qubits):
+            raise RuntimeError(self._error_msg)
```
joeycarter: And here, since this `RuntimeError` would be visible to the user, it would be better to raise a `pennylane.exceptions.CompileError`.

frontend/catalyst/python_interface/transforms/quantum/diagonalize_measurements.py
```diff
@@ -73,9 +75,88 @@ def _diagonalize(obs: NamedObsOp, supported_base_obs) -> bool:
...
 
 
+class NonCommutingObservableValidator:
+    """
+    Validates if all quantum observables of an operation commute.
+    """
+
+    _error_msg = "Only observables are not qwc. Please apply the `split-non-commuting` pass first."
+
+    def __init__(self, op, supported_base_obs: tuple[str]):
```
joeycarter: Minor, since `supported_base_obs` can be any sequence or set (I think), maybe this type hint would be more appropriate:


```suggestion
    def __init__(self, op, supported_base_obs: Collection[str]):
```

(you'll need `from collections.abc import Collection` at the top if so).

frontend/catalyst/python_interface/transforms/quantum/diagonalize_measurements.py
joeycarter, line 283: Just noticed this. Instead of writing out the supported base observables explicitly, it would be more future-proof if we insert the string from the set directly, e.g.

```python
msg = f"... subset of ({", ".join(_obs_allowed_diagonalization)}) passed as..."
```

frontend/catalyst/python_interface/transforms/quantum/diagonalize_measurements.py
```diff
@@ -73,9 +75,88 @@ def _diagonalize(obs: NamedObsOp, supported_base_obs) -> bool:
...
+        else:
+            self._run_non_overlapping_validation()
+
+    def _analyze_module(self):
+        """Single walk through the IR to populate qubit and observable maps."""
+        for op_ in self.op.walk():
+            if isinstance(op_, NamedObsOp):
+                self._register_qubits([op_.qubit], op_.type.data.value)
+                if op_.type.data.value == "Hadamard":
+                    self.is_qwc_compatible = False
```
joeycarter: It looks like this is saying whenever there is a Hadamard observable, any other observables acting on the same qubit is not qwc, but Identity is also a valid observable. H and I commute, so they should be simultaneously observable. In other words, this should work:

```python

qml.capture.enable()

@qml.qjit
@diagonalize_final_measurements_pass
@qml.qnode(dev, shots=1)
def circuit():
    return qml.expval(qml.H(0)), qml.expval(qml.I(0))
```

But it gives

```
RuntimeError: Only observables are not qwc. Please apply the `split-non-commuting` pass first.
```

This is a corner case for sure, but I don't think it should result in an error. Similarly for multiple Hadamard observables on the same qubit.

frontend/catalyst/python_interface/transforms/quantum/diagonalize_measurements.py
```diff
@@ -73,9 +75,88 @@ def _diagonalize(obs: NamedObsOp, supported_base_obs) -> bool:
     )  # pragma: no cover
 
 
+class NonCommutingObservableValidator:
+    """
+    Validates if all quantum observables of an operation commute.
+    """
+
+    _error_msg = "Only observables are not qwc. Please apply the `split-non-commuting` pass first."
```
joeycarter: [Suggestion] Probably best to avoid acronyms in error messages:


```suggestion
    _error_msg = "Observables are not qubit-wise commuting. Please apply the `split-non-commuting` pass first."
```

frontend/catalyst/python_interface/transforms/quantum/diagonalize_measurements.py
```diff
@@ -84,68 +165,74 @@ def __init__(self, supported_base_obs: set[str]):
...
+
+                        gate = CustomOp(in_qubits=qubit, gate_name=name, params=param_ssa_values)
+                    else:
+                        gate = CustomOp(in_qubits=qubit, gate_name=name)
+
+                    rewriter.insert_op(gate, insert_point)
+
+                    qubit = gate.out_qubits[0]
+
+                # we need to replace the initial qubit use everwhere EXCEPT the use that is now the
```
joeycarter:
```suggestion
                # we need to replace the initial qubit use everywhere EXCEPT the use that is now the
```

joeycarter (changes requested):
Thanks @multiphaseCFD! 🚀 

I think we're in good shape here. I caught one corner case that should be addressed before merging, but otherwise I mostly had only minor suggestions and editorial things.

frontend/catalyst/python_interface/transforms/quantum/diagonalize_measurements.py
multiphaseCFD: Nice catch! I thought I removed this logic. Removed

frontend/catalyst/python_interface/transforms/quantum/diagonalize_measurements.py
multiphaseCFD: Not really for the current use case, since the `supported_base_obs` is required to be hashable and only `tuple` is accepted for now. I think it's good use this type hint for now.

joeycarter: Why does it have to be hashable? Also you create a new `NonCommutingObservableValidator` instance using a `set` here:

https://github.com/PennyLaneAI/catalyst/blob/285feb13719707aeb66ebcd46715cf443d0aa450/frontend/catalyst/python_interface/transforms/quantum/diagonalize_measurements.py#L281

where `self.supported_base_obs` is a `set`. In fact, this is how I caught it initially because my Python type checker was complaining about an invalid argument type given the type hint `tuple[str]`.

Anyway, it's a minor point :slightly_smiling_face:

multiphaseCFD: Make sense! Let me update it. Thanks @joeycarter !

frontend/catalyst/python_interface/transforms/quantum/diagonalize_measurements.py
multiphaseCFD: Good point! But this error msg is different from the tape transform. Let's keep the error message aligned with the frontend for now.

doc/releases/changelog-dev.md
```diff
@@ -178,6 +178,9 @@
   circuit()
   ```
 
+* The `diagonalize-final-measurements` xDSL pass now supports the observable commutativity check. A new `NonCommutingObservableValidator` is added to check the commutativity of each circuit (a `func.func` with a `quantum.node` attribute) in the IR. The `qubit-wise commutativity`(QWC) check would be applied if there are only `Pauli` observables in the circuit. Otherwise, the more strict non-overlapping observable check is applied.
+  [(#2538)](https://github.com/PennyLaneAI/catalyst/pull/2538)
```
joeycarter: [Suggestion] I think this reads more clearly, but please feel free to reword if I have misrepresented anything about this change.

```suggestion
* The `diagonalize-final-measurements` xDSL pass now includes an observable-commutativity check and
  raises an error if non-commuting terms are encountered. The check is applied to each `qnode` in
  the IR (that is, a `func.func` op with a `quantum.node` attribute). If the measurement contains
  only Pauli or Hadamard observables, the *qubit-wise commutativity* (QWC) check is applied.
  Otherwise, the more strict *non-overlapping observable* check is applied.
  [(#2538)](https://github.com/PennyLaneAI/catalyst/pull/2538)
```

frontend/catalyst/python_interface/transforms/quantum/diagonalize_measurements.py
```diff
@@ -73,9 +77,89 @@ def _diagonalize(obs: NamedObsOp, supported_base_obs) -> bool:
...
+    def _run_non_overlapping_validation(self):
+        """Strictest check: no two observables can share a qubit."""
+        if self.overlapped_qubits or (self.visited_qreg and self.obs_on_qubits):
+            raise CompileError(self._error_msg)
+
+
 class DiagonalizeFinalMeasurementsPattern(
     pattern_rewriter.RewritePattern
-):  # pylint: disable=too-few-public-methods
+):  # pylint: disable=too-few-public-methods, trailing-whitespace, cell-var-from-loop
```
joeycarter: Can we remove the `trailing-whitespace` pragma here?

```suggestion
):  # pylint: disable=too-few-public-methods, cell-var-from-loop
```

joeycarter: I should also double check, are you sure it's safe to disable `cell-var-from-loop` warnings here? If so, could we move it to be directly above the line

```python
op.qubit.replace_by_if(qubit, lambda use: use in uses_to_change)
```

to avoid disabling it for the entire function?

joeycarter (approved):
Thanks @multiphaseCFD! Looks great! 💯 

I only had a few minor suggestions, so happy to provisionally approve.

frontend/catalyst/python_interface/transforms/quantum/diagonalize_measurements.py
multiphaseCFD: Done!

multiphaseCFD: Done

frontend/catalyst/python_interface/transforms/quantum/diagonalize_measurements.py
```diff
@@ -84,68 +168,75 @@ def __init__(self, supported_base_obs: set[str]):
...
+
+                qubit = op.qubit
+
+                insert_point = InsertPoint.before(op)
+
+                commute_obs = [
+                    use.operation
+                    for use in op.qubit.uses
+                    if isinstance(use.operation, (NamedObsOp)) and use.operation is not op
+                ]
```
lillian542, line 194: This is defined far from where it's used, and took me a moment to interpret - could we add a comment above here?

frontend/catalyst/python_interface/transforms/quantum/diagonalize_measurements.py
```diff
@@ -73,6 +77,86 @@ def _diagonalize(obs: NamedObsOp, supported_base_obs) -> bool:
...
+        self.visited_qreg = False
+        self.is_qwc_compatible = True
+
+        # 1. Single pass to gather all required IR metadata
+        self._analyze_module()
+
+        # 2. Determine validation strategy
+        diagonalize_all = self.supported_base_obs.issubset(_default_supported_obs)
+
+        if self.is_qwc_compatible and diagonalize_all:
```
lillian542: How is whether or not we `diagonalize_all` related to whether or not we can use the QWC check?

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_diagonalize_measurements.py
```diff
@@ -426,25 +432,21 @@ def circuit():
...
-            func.func @test_func() {
+            func.func @test_func() attributes {quantum.node} {
                 %0 = "test.op"() : () -> !quantum.bit
                 // CHECK: quantum.custom "Hadamard"()
                 // CHECK-NEXT: quantum.namedobs [[q:%.+]][PauliZ]
                 %1 = quantum.namedobs %0[PauliX] : !quantum.obs
                 %2 = quantum.var %1 : f64
-                // CHECK: quantum.custom "Hadamard"()
-                // CHECK-NEXT: quantum.namedobs [[q:%.+]][PauliZ]
+                // CHECK: quantum.namedobs [[q:%.+]][PauliZ]
```
lillian542, line 442: Can we add a `CHECK-NOT: quantum.custom "Hadamard"` here, to confirm it wasn't diagonalized twice? And then lets update the name and docstring of this test to reflect its new behaviour.

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_diagonalize_measurements.py
```diff
@@ -605,12 +607,8 @@ def circuit(x):
             qml.RX(x, 0)
             return qml.expval(qml.Y(0)), qml.var(qml.Y(0))
 
-        with pytest.raises(
-            RuntimeError, match="the circuit contains multiple observables with the same wire"
-        ):
-            _ = circuit(1.23)
+        _ = circuit(1.23)
```
lillian542: I think we usually avoid `_ = circuit(...)` in the integration tests unless we're specifically testing that _an error is raised_, so there's no need to examine the output of executing the circuit. 

Since this isn't testing an error is raised anymore, could we update it to check for correct results? The expval should be `-sin(phi)` and the variance `1-sin(phi)**2` (with `phi=1.23`) in this case.

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_diagonalize_measurements.py
```diff
@@ -843,9 +788,9 @@ def expected_res(x, y):
 
         assert np.allclose(expected_res(phi, theta), circuit_compiled(phi, theta))
 
-    def test_overlapping_observables_raises_error(self):
+    def test_overlapping_commute_observables_no_error_raises(self):
```
lillian542: Isn't this identical to the test in line 597?

lillian542 (commented):
Thanks @multiphaseCFD, looks good! I just left a couple of minor comments and questions

frontend/catalyst/python_interface/transforms/quantum/diagonalize_measurements.py
multiphaseCFD: I was tring to use the same logic as implemented in the pennylane [here](https://github.com/PennyLaneAI/pennylane/blob/494cd943c4ce675f8c392a9b8fb98626d1b78c02/pennylane/transforms/diagonalize_measurements.py#L172-L175).

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_diagonalize_measurements.py
multiphaseCFD: Good catch! Removed

frontend/catalyst/python_interface/transforms/quantum/diagonalize_measurements.py
lillian542: Ah, I see, thanks! That's related to an implementation detail in the python version regarding diagonalizing using a function from the `pauli` module (`qml.pauli.diagonalize_qwc_pauli_words`) vs doing it based on `op.diagonalizing_gates`. The `diagonalize_qwc_pauli_words` is more efficient, but can't handle only diagonalizing a subset of the observables. 

It's not related to validation, and it's not relevant here - we always use `op.diagonalizing_gates` in this implementation.

frontend/catalyst/python_interface/transforms/quantum/diagonalize_measurements.py
multiphaseCFD: add more explainations

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_diagonalize_measurements.py
multiphaseCFD: Done

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_diagonalize_measurements.py
```diff
@@ -426,25 +432,22 @@ def circuit():
         """
```
lillian542:
```suggestion
def test_overlapping_commuting_observables(self, run_filecheck):
"""Test the case where multiple overlapping (commuting) observables exist in
       the same circuit (diagonalization is only performed once).

       @qml.qjit(target="mlir")
       @qml.qnode(qml.device("lightning.qubit", wires=1))
       def circuit():
           return qml.var(qml.X(0)), qml.var(qml.X(0))
       """
```

lillian542 (approved):
Just one suggestion to update a test docstring now that the behaviour has changed. Looks good to me!

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_diagonalize_measurements.py
```diff
@@ -423,28 +429,33 @@ def test_overlapping_observables_raises_error(self, run_filecheck):
...
-        """
+def test_overlapping_commuting_observables(self, run_filecheck):
+"""Test the case where multiple overlapping (commuting) observables exist in
+       the same circuit (diagonalization is only performed once).
+
+       @qml.qjit(target="mlir")
+       @qml.qnode(qml.device("lightning.qubit", wires=1))
+       def circuit():
+           return qml.var(qml.X(0)), qml.var(qml.X(0))
+       """
```
lillian542: Aaah, it looks like I messed something up with indentation and which lines to replace when adding a suggestion here, sorry 😅

multiphaseCFD: np! Done!

PR #2539, opened 2026-03-03
Add RegisterDecompRuleResourcePass to compute and register resources of the decomposition rules at MLIR

maliasadi:
**Context:**
The `resource-tracking` pass computes comprehensive resource metrics, including gate counts,
measurements, qubit allocations, device name, as well as structured classical ops.
For operations within the `Quantum`, `PBC`, and `MBQC` dialects,
It leverages `ResourceAnalysis` to retrieve cached results and integrates directly
with the MLIR Pass Statistics mechanism for reporting.

This pass uses `ResourceTrackerPass` to compute resource metrics of functions marked with the `target_gate` attribute, effectively filtering for decomposition rules.

Usage:
```
quantum-opt input.mlir -register-decomp-rule-resource
```

**Benefits:**
- Use these resources in the graph-based decomposition for the AOT/JIT compiled rules

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-112233]


mlir/include/Catalyst/Transforms/Passes.td
sengthai, line 1: Thanks @maliasadi for implementing this. Here, we added options to update the attribute functions specifically for graph decomposition, which change the MLIR. However, the purpose of ResourceTrackerPass here is only to observe and track the operations.  And our initial goal for ResourceTracker is to enable other passes to use and build on it. To make it more modular, I would suggest adding a new pass to inject the attributes into the functions. So then, the pass is easier to maintain while keeping the purpose clear, and keeping it modular.

maliasadi: Ah that was my original implementation, but I ended up removing the new pass and added new formatting options  to `resource-tracking`. Here, I've added two options to the pass: `update-attr` and `decomp-attr`. I assumed there will be some use cases for `update-attr` as an option. Would you like me to add both to a new pass or just the `decomp-attr`?

sengthai: > Ah that was my original implementation, but I ended up removing the new pass and added new formatting options to resource-tracking.

Ohh my bad, I thought you were building a pass for tracking. I just realized that you want to utilize the result from resource-tracking, then inject them as attributes.

> Would you like me to add both to a new pass or just the decomp-attr?

I think for a new pass you don't need both of these options. It would be great to move to the implementation of `buildResourceDict`  in a new pass. And, you can call `getAnalysis<ResourceAnalysis>()` to get the result, so then you can inject gate info into the attribute.

maliasadi: Updated the pass as we discussed offline!

doc/releases/changelog-dev.md
```diff
@@ -142,12 +142,22 @@
...
   quantum-opt --resource-tracker='output-json=true' input.mlir
   quantum-opt --resource-tracker -mlir-pass-statistics input.mlir
   ```
 
+* Added a pass to compute resource metrics of functions marked with the `target_gate` attribute,
+  effectively filtering for decomposition rules.
+  [(#2539)](https://github.com/PennyLaneAI/catalyst/pull/2539)
+
+  ```bash
+  quantum-opt input.mlir -register-decomp-rule-resource
```
kipawaa, line 157: I think that expanding this to a simple example that shows some of the attributes would be helpful, since it's not documented anywhere else.

mlir/lib/Catalyst/Transforms/RegisterDecompRuleResourcePass.cpp
```diff
@@ -0,0 +1,144 @@
...
+    void runOnOperation() final
+    {
+        auto &analysis = getAnalysis<ResourceAnalysis>();
+        const auto &results = analysis.getResults();
+        auto module = llvm::cast<mlir::ModuleOp>(getOperation());
+
+        MLIRContext *ctx = &getContext();
+        OpBuilder builder(ctx);
+
+        for (auto func : module.getOps<mlir::func::FuncOp>()) {
```
kipawaa, line 51: To my understanding, this will only apply to top level functions (i.e. in the main file module). Is it relevant to us to expand this to a traversal of the module, or to sub-functions or anything else?
Ex.
```mlir
builtin.module {
  func.func public @outer_func() attributes {target_gate = "Hadamard"} { // this will be annotated
    func.func public @inner_func() attributes {target_gate = "Hadamard"} { // this will not be annotated 
      ...
    }
  }
  builtin.module {
    func.func public @submodule_func() attributes {target_gate = "RX"} { // this also won't be annotated
      ...
    }
  }
}
```
I know we typically guarantee that decomposition rules will be top-level (i.e. in the file module), but I want to verify that this is intended. It also might be worth documenting this, since it may not be immediately obvious to a reader/user.

kipawaa (commented):
Thanks Ali, this looks great! Just a quick question and changelog comment

mlir/lib/Catalyst/Transforms/RegisterDecompRuleResourcePass.cpp
maliasadi: I don't know of any decomp rules with an outer-inner func recursive definition. While decomposition is a recursive algorithm, the rules themselves are standalone quantum funcs. I'll add a code comment to clarify the scope of this pass in the code. Addressed [bfbb87e](https://github.com/PennyLaneAI/catalyst/pull/2539/commits/bfbb87e3f1f2f96b63387c48891e37721f121d32)

doc/releases/changelog-dev.md
maliasadi: Fixed [bfbb87e](https://github.com/PennyLaneAI/catalyst/pull/2539/commits/bfbb87e3f1f2f96b63387c48891e37721f121d32)

kipawaa (approved):
Great work, thank you!

sengthai (approved):
Looks so good! Thanks @maliasadi! 💯

PR #2542, opened 2026-03-04
Remove support for using QNodes as input to inspection tools

mudit2812:
`catalyst.python_interface.draw` and `catalyst.python_interface.generate_mlir_graph` allowed using QNodes as input. Both these functions are supposed to only work with QJIT (and in fact they wrap the input QNode in QJIT before anything else). Supporting QNode inputs for MLIR-level functionality does not make sense, and is also causing issues with #2496 .

Also sneaking in a change to support getting the value of constants cast using `index_cast` in `get_constant_from_ssa`

[sc-113126] [sc-113127]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -150,9 +150,13 @@
 
 <h3>Improvements 🛠</h3>
 
+* `catalyst.python_interface.utils.get_constant_from_ssa` can now extract constant values cast using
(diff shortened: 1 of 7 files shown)


andrijapau (approved):
LGTM 🚀

 This always confused me, glad it's getting cleaned up

andrijapau (approved):

mehrdad2m (approved):

PR #2544, opened 2026-03-04
Add qecl.qec

joeycarter:

**Description of the Change:** This PRs adds the `qecl.qec` op to represent a single cycle of a quantum error-correction protocol at the logical layer.

```mlir
%hreg = qecl.alloc() : !qecl.hyperreg<3 x 1>
%b0 = qecl.extract_block %hreg[0] : !qecl.hyperreg<3 x 1> -> !qecl.codeblock<1>
%b1 = qecl.encode [zero] %b0 : !qecl.codeblock<1>
%b2 = qecl.qec %b1 : !qecl.codeblock<1>
```

> [!NOTE]

[sc-111752]


frontend/catalyst/python_interface/dialects/qecl.py
```diff
@@ -343,6 +343,33 @@ def __init__(
...
+
+    assembly_format = """
+            $in_codeblock attr-dict `:` type($in_codeblock)
+        """
+
+    def __init__(
+        self,
+        in_codeblock: LogicalCodeBlockSSAValue | Operation,
+    ):
+        operands = (in_codeblock,)
```
multiphaseCFD, line 366: An open question for discussion: I believe the current design is based on the assumption of the online correction for detected errors. If yes, what is your thoughts on how to extend it to support logical pauli frame?

multiphaseCFD (commented):
Looks good to me. Thanks @joeycarter

frontend/catalyst/python_interface/dialects/qecl.py
joeycarter: Good question. If we do decide to go for a multi-level Pauli Frame tracking approach, then we should be able to apply a Pauli frame tracking pass at the `quantum` dialect level to update the logical Pauli frame for Pauli gates, Clifford gates, etc. In that case, the logical Pauli frame tracking ops will already be in place before it reaches the QEC logical dialect.

Then for physical Pauli frame tracking, we could apply a pass after converting to the QEC physical dialect. In that case, all Pauli corrections (expressed as physical Pauli gates), would be converted to their corresponding Pauli frame ops.

I think a full multi-level Pauli frame tracking prototype is relatively advanced and not something we'll get to soon. A physical-only Pauli frame tracker seems more appropriate as a next step to our initial QEC prototype, since it should be reasonably straightforward to convert Pauli ops in the QEC protocols to Pauli frame ops, and so forth for Clifford and non-Clifford gates.

lillian542: I think the task of updating the physical Pauli frame will live inside the decoder or be more directly integrated with it rather than being part of the main program that is compiled in MLIR, for a couple of reasons:

1. I would expect the decoder needs to store the Pauli tracker info anyway, because I anticipate it will need it to decode correctly
2. I would expect we want to avoid back-and-forth where possible to cut down on communication time between separate hardware entities. The decoder determines what errors may have occurred, and we don't really need that info back until it's time to make a correction, so it doesn't make sense to me that the main program would act as a middle-man here.

Maybe as an initial prototype we want to prototype the behaviour "inside the decoder" in MLIR, or as some kind of runtime thing, or maybe we go straight for hardware integration with an FPGA running a decoder program. TBD, I think 🤷 

In terms of how that integrates with the logical pauli frame, I _think_ it can be separate from the physical frame tracker? If the physical tracker is decoding errors based on syndrome measurements, it's only impacted by correlations between the data qubits, so it seems like it doesn't need to know if we applied logical gates. But I might be missing something here.

lillian542 (approved):

frontend/catalyst/python_interface/dialects/qecl.py
joeycarter: I think there's some more discovery work to do on this topic. Let's revisit Pauli frame tracking when we get the initial QEC prototype working.

multiphaseCFD (approved):
Nice one @joeycarter !

PR #2549, opened 2026-03-05
Add the qecl.measure op

joeycarter:

**Description of the Change:** This PRs adds the `qecl.measure` op to represent a logical single-qubit projective measurement in the computational basis.

For example:

```mlir
%0 = ... : !qecl.codeblock<1>
%mres, %1 = qecl.measure %0[0] : i1, !qecl.codeblock<1>
```

> [!NOTE]

[sc-113245]


frontend/catalyst/python_interface/dialects/qecl.py
```diff
@@ -533,6 +541,55 @@ def __init__(
...
+    idx_attr = opt_prop_def(IntegerAttr.constr(type=IndexType, value=AtLeast(0)))
+
+    mres = result_def(IntegerType(1))
+
+    out_codeblock = result_def(T)
+
+    def __init__(
+        self,
+        in_codeblock: LogicalCodeBlockSSAValue | Operation,
+        idx: int | IntegerAttr | SSAValue[IndexType] | Operation,
```
multiphaseCFD, line 694: `idx` is included as an argument here, but measureOp in our quantum dialect doesn't use classical indexing. I would suggest to add an description for it

frontend/catalyst/python_interface/dialects/qecl.py
```diff
@@ -533,6 +541,55 @@ def __init__(
...
+    ):
```
multiphaseCFD, line 695: could you explain the scenario that `idx` is passed as an `operation`?

frontend/catalyst/python_interface/dialects/qecl.py
```diff
@@ -533,6 +541,55 @@ def __init__(
...
+@irdl_op_definition
+class MeasureOp(IRDLOperation):
+    """A logical single-qubit projective measurement in the computational basis."""
+
+    T: ClassVar = VarConstraint("T", anyLogicalCodeblock)
+
+    name = "qecl.measure"
+
+    assembly_format = """
+            $in_codeblock `[` ($idx^):($idx_attr)? `]` attr-dict `:` type($mres) `,` type($in_codeblock)
```
multiphaseCFD, line 678: the signature defined here is different from `quantum.measure`, which accepts `postselect`. Do we want to add it as well?

multiphaseCFD (commented):
Thansks Joey! Just a couple of Qs

frontend/catalyst/python_interface/dialects/qecl.py
joeycarter: This is the index in the codeblock, same as for the Clifford gate ops. I've added documentation here: db181c0e8

frontend/catalyst/python_interface/dialects/qecl.py
joeycarter: This is a common pattern in xDSL op constructors for the operation's input operands. In other words, you can construct the op by passing in specific SSA values or by passing in an operation itself, where the input operand is taken as the result of the other operation.

For example, if we want `idx` to be the SSA value from an `arith.constant` operation, we can either pass in the SSA value that `arith.constant` returns or the `arith.constant` op object itself; they're equivalent.

frontend/catalyst/python_interface/dialects/qecl.py
joeycarter: I don't think we need `postselect` here. For HW, I think it's more realistic to deallocate the measured qubit and allocate a new one as needed, rather than magically putting it into the desired state.

lillian542: Maybe worth adding a comment somewhere? If we eventually end up moving this out of "experimental" mode, we may want to consider either adding support for postselect (since people are used to it), or catching it and raising a useful error message if the `measure` op that is being lowered to the logical QEC dialect has a `postselect` mode specified.

But I agree that for our purposes we can just ignore `postselect`, so no need to spend time on it now.

frontend/catalyst/python_interface/dialects/qecl.py
multiphaseCFD: Thanks Joey! I understand we could support SSA value be passed to the operation. I was wondering if we need that? I was thinking we might need this feature when converting idx from one layer to another, right?

joeycarter: Ah, so you're saying, should we support both the case where idx is an attribute and where it is a (potentially dynamic) SSA value? i.e. do we need to support both

```mlir
%b0 = ... : !qecl.codeblock<1>
%mres, %b1 = qecl.measure %b0[0] : i1, !qecl.codeblock<1>
```

and

```mlir
%b0 = ... : !qecl.codeblock<1>
%idx = arith.constant 0 : index
%mres, %b1 = qecl.measure %b0[%idx] : i1, !qecl.codeblock<1>
```
 or potentially,


```mlir
func.func @my_func(%arg0 : index) {
  %b0 = ... : !qecl.codeblock<1>
  %mres, %b1 = qecl.measure %b0[%arg0] : i1, !qecl.codeblock<1>
}
```

I would say long-term the answer is yes, since the wires can be dynamic in the initial program. Strictly speaking, we likely won't need this functionality for the initial prototype, since we should always be dealing with static wire indices (which get lowered to static hyper-register and codeblock indices in the QEC logical layer). But it was trivial to add this support so I thought we might as well add it now :slightly_smiling_face:

multiphaseCFD: Thanks @joeycarter !

multiphaseCFD (approved):
LGTM!

lillian542 (approved):
Looks good, thanks @joeycarter!

frontend/catalyst/python_interface/dialects/qecl.py
joeycarter: Sounds good, I added a note here: 291035b9a

PR #2550, opened 2026-03-05
`ApplyTransformSequencePass` can apply multiple `transform.named_sequences`

mudit2812:
**Context:**
We want to support multiple transform sequences per QNode for device preprocessing. This is a precursor to enable that.

* Update the C++ pass to allow multiple transform sequences
* Rework the xDSL pass to allow multiple transform sequences

**Benefits:**
Enabling the creation of more inspectable compilation pipelines

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-113260]


frontend/test/pytest/python_interface/dialects/test_transform_dialect.py
```diff
@@ -126,12 +126,14 @@ def apply(self, _ctx: Context, _module: builtin.ModuleOp) -> None:
 
     program = """
         builtin.module {
-          builtin.module {
-            transform.named_sequence @__transform_main(%arg0 : !transform.op<"builtin.module">) {
-              %0 = "transform.apply_registered_pass"(%arg0) <{options = {"custom_print" = "Hello from custom option!"}, pass_name = "test-hello-world"}> : (!transform.op<"builtin.module">) -> !transform.op<"builtin.module">
-              transform.yield
+            builtin.module {
+                builtin.module {
```
dime10: Where is the third module from? 😅

mlir/lib/Catalyst/Transforms/ApplyTransformSequencePass.cpp
```diff
@@ -177,29 +177,23 @@ struct ApplyTransformSequencePass
...
             }
-        }
+        });
 
         transformer.erase();
     }
 };
 
-} // namespace catalyst
+} // namespace catalyst
```
dime10:
```suggestion
} // namespace catalyst

```

dime10 (approved):

frontend/test/pytest/python_interface/dialects/test_transform_dialect.py
mudit2812: Previously, we were testing the `TransformInterpreterPass` which was called by `ApplyTransformSequencePass`. The latter always expected the "QJIT structure": a workflow module with nested modules, each of which had its own "transformer" module. Since the `TransformInterpreterPass` no longer exists, the tests need to adhere to the 3 nested modules structure

mlir/lib/Catalyst/Transforms/ApplyTransformSequencePass.cpp
```diff
@@ -177,26 +177,20 @@ struct ApplyTransformSequencePass
...
             }
-        }
-        else {
-            if (failed(transform::applyTransforms(payload, transformer_main_sequence, {},
-                                                  transform::TransformOptions(), false))) {
-                return signalPassFailure();
+            else {
+                if (failed(transform::applyTransforms(payload, op, {},
+                                                      transform::TransformOptions(), false))) {
+                    return signalPassFailure();
```
mehrdad2m: I think this logic would allow the IR to be invalid if a pass fails and would continue walking the invalid IR. We could  interrupt the walk here as well to make sure we halt once a pass fails.

frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
```diff
@@ -43,42 +49,214 @@ def is_xdsl_pass(pass_name: str) -> bool:
...
-            for region in mod.regions:
-                for block in region.blocks:
-                    for operation in block.ops:
-                        if isinstance(operation, builtin.ModuleOp) and operation.get_attr_or_prop(
-                            "transform.with_named_sequence"
-                        ):
-                            block.erase_op(operation)  # pragma: no cover
+            transformer = None
+            for _op in mod.ops:
+                if isinstance(_op, builtin.ModuleOp):
```
mehrdad2m: why did we remove the check for  `transform.with_named_sequence`?  This would now assume that the first module is always a transform module.

frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
```diff
@@ -43,42 +49,214 @@ def is_xdsl_pass(pass_name: str) -> bool:
...
+        """Rewrite modules containing transform.named_sequences."""
+        if not isinstance(next(iter(transformer.body.ops), None), transform.NamedSequenceOp):
+            return  # pragma: no cover
+
+        payload: builtin.ModuleOp = transformer.parent_op()
+        rewriter.erase_op(transformer)
+
+        for i, ns in enumerate(transformer.ops):
+            if len(ns.body.ops) == 1:
+                if i == 0:
```
mehrdad2m: Could you help me understand this part? Is this to handle an empty transform module? in that case, why would we increase `pass_level`?

mehrdad2m (approved):
LGTM. just some minor comments.

frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
mudit2812: I will admit that this logic likely only works for 2 or less named sequences, but here is the logic. Note that below when I say "call _pre_pass_callback", I mean calling it when pass_level=0. For any other values, _pre_pass_callback does not do anything.
* If both transform sequences are empty, _pre_pass_callback is only called once (at the beginning of the first sequence).
* If the 1st sequence is empty but the 2nd one is not, _pre_pass_callback is only called with `pass_level=0` once (at the beginning of the first sequence), after which we only ever have calls to `_post_pass_callback`.
* If the 1st sequence is not empty but the 2nd one is, _pre_pass_callback is only called with `pass_level=0` once (at the beginning of the first sequence), and then when iterating over the 2nd sequence, no callbacks are called.
* If both sequences are non-empty, we also get the correct behaviour.

frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
mudit2812: I think that is a mistake. Nice catch!

frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
mudit2812: The more I think about it, I think that the logic I implemented would work for any number of named sequences

frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
mudit2812: [84f7902](https://github.com/PennyLaneAI/catalyst/pull/2550/commits/84f79027d6549b17c7ab986ea8033c22223690f0)

mlir/lib/Catalyst/Transforms/ApplyTransformSequencePass.cpp
mudit2812: I'm not sure I entirely understand. Maybe we can go over it tomorrow

mudit2812: I made this change based on what I thought you meant. Let me know what you think: [8c60e3e](https://github.com/PennyLaneAI/catalyst/pull/2550/commits/8c60e3ef4af245de54669b61b2ae4599935c89eb)

frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
mudit2812: Actually, I realized there was a lot of opportunity to clean up the logic: [fe0db26](https://github.com/PennyLaneAI/catalyst/pull/2550/commits/fe0db26c637e2c1ef5e690e0303184ca8faa8398)

mehrdad2m: Thanks for the explanation, it looks great now.

mlir/lib/Catalyst/Transforms/ApplyTransformSequencePass.cpp
mehrdad2m: yeah that's exactly what I meant.

mehrdad2m (approved):

PR #2551, opened 2026-03-05
Stop casting transform kwargs from tuple to list

lillian542:
**Context:**
Current conversion turns tuples into lists, and then doesn't turn them back. It's preventing us from passing a tuple to an xDSL transform through the PL/capture frontend.

We only modify the transform arguments to make them lists where its used for the old plxpr infrastructure, instead of also doing it for `tkwarg`s passed to the MLIR/xDSL passes.

**Benefits:**
We don't need special handling to cast things back to tuples for passes that have some `kwarg=tuple(1, 2, 3..)` argument


JerryChen97 (commented):
LGTM actually! I'll tag @rniczh to see if any further concern here

JerryChen97 (approved):

rniczh (approved):
LGTM @lillian542 👍

PR #2525, opened 2026-02-26
remove internal `PassPipelineWrapper/QNodeWrapper` in favour of PennyLane's new unified transforms API 

andrijapau:
Copies some of the changes made here: https://github.com/PennyLaneAI/catalyst/pull/2487.

The goal of this PR is to get rid of `PassPipelineWrapper` and `QNodeWrapper` and use the new unified transforms API.

Complimentary PR in PennyLane that follows up on the removal: https://github.com/PennyLaneAI/pennylane/pull/9123

[sc-112140][sc-104261]

PR #2552, opened 2026-03-06
[Base] The native-MLIR graph decomposition

maliasadi:
**Context:**
The objective is to develop a high-performance, hybrid decomposition framework that further migrates the graph-decomposition logic from Python to the native MLIR compiler. The new framework needs to adhere to four pillars:

1. We will migrate the `GraphDecomposition` build and solver logic from Python into C++ https://github.com/PennyLaneAI/catalyst/pull/2578. By co-locating the solver with the decomposition-rule rewrite MLIR pass into a `graph-decomposition` system at MLIR, we enable late-stage circuit rewrites and multi-pass optimization cycles that were previously inaccessible in the frontend-only framework.

2. To minimize runtime latency, PL’s standard and built-in decomposition rules will be precompiled for release, and lazily re-compiled if necessary on import https://github.com/PennyLaneAI/catalyst/pull/2531 https://github.com/PennyLaneAI/catalyst/pull/2539. By treating built-in rules as static compiler assets rather than dynamic JIT targets, we try to eliminate the redundant capture and compilation overhead typically incurred during every workflow execution.

3. To maintain the flexibility of PL, the system will need to still support JIT compilation of rules. While standard rules are handled by the AOT backend, custom user-defined device rules will be resolved via a JIT fallback mechanism https://github.com/PennyLaneAI/catalyst/pull/2568. This is to ensure that researcher-defined gatesets and rules remain fully supported without requiring a full compiler re-build.

4. Short-term, a "PLxPR Guard" stage will handle edge cases before MLIR lowering, acting as a fallback until the C++ system offers full support. Compiler-incompatible operators and complex high-level templates (e.g., Subroutine/Template work) will continue to be handled/decomposed at the PLxPR level; so that the backend receives a valid, lowerable MLIR representation while deferring hardware-specific passes to MLIR/C++. This step is temporary and will be revisited aligned with the updates in the Operator class and the program capture mechanism for Symbolic Ops.

**Benefits:**
- Enable "late-stage" and iterative decomposition passes at MLIR with a C++ decomposition graph
- Eliminate redundant compilation overhead of built-in decomposition rules during JIT compilation of workflows
 
**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-113535]
[sc-113537]
[sc-115445]


mlir/include/Quantum/Utils/Decomp.h
```diff
@@ -0,0 +1,28 @@
+// Copyright 2023 Xanadu Quantum Technologies Inc.
```
dime10:
```suggestion
// Copyright 2026 Xanadu Quantum Technologies Inc.
```

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -1440,3 +1444,87 @@ def circuit():
...
+        options |= {
+            "alt-decomps": {
+                to_name(op): tuple(rule.__name__ for rule in rules)
+                for op, rules in alt_decomps.items()
+            }
+        }
+
+    graph_decomposition_pass = {"graph-decomposition": options}
+
+    return qml.transform(pass_name="decompose-arbitrary-ppr", **graph_decomposition_pass)(qnode)
```
maliasadi: re: integration with frontend 
@kipawaa PPW was removed in https://github.com/PennyLaneAI/catalyst/pull/2525.

runtime/lib/DecompGraphSolver/DGTypes.hpp
```diff
@@ -0,0 +1,184 @@
...
+ * The OperatorNode struct captures the essential information about an operator, including
+ * its name, the number of wires it acts on, and the number of parameters it has.
+ * This information is crucial for the graph decomposition solver to determine how operators
+ * can be combined and decomposed to achieve the desired target gateset while optimizing for
+ * resource usage.
+ */
+struct OperatorNode {
+    std::string name;
+    std::size_t numWires{0};
+    std::size_t numParams{0};
```
astralcai: I guess this is sufficient for now since simple gates don't have any resource params, and there is special handling for `PauliRot` which takes a string as input? In the long term we potentially want this to be more general, but that's a separate discussion after the new operator abstractions ADR is finalized.

runtime/lib/DecompGraphSolver/DGTypes.hpp
```diff
@@ -0,0 +1,184 @@
...
+    bool adjoint{false};
```
astralcai, line 51: We're skipping controlled operators for now?

runtime/lib/DecompGraphSolver/DGTypes.hpp
```diff
@@ -0,0 +1,184 @@
...
+ * and number of parameters.
+ */
+struct OperatorNodeHash {
+    std::size_t operator()(const OperatorNode &node) const
+    {
+        std::size_t h1 = std::hash<std::string>{}(node.name);
+        // TODO: commented out when we capture numWires and numParams for the target gateset
+        // std::size_t h2 = std::hash<std::size_t>{}(node.numWires);
+        // std::size_t h3 = std::hash<std::size_t>{}(node.numParams);
+        std::size_t h4 = std::hash<bool>{}(node.adjoint);
```
astralcai: Wait so two operators of the same type but different number of wires (e.g., `MultiRZ(num_wires=3)` and `MultiRZ(num_wires=5)`) would have the same hash right now?

runtime/lib/DecompGraphSolver/DGTypes.hpp
maliasadi: Yes, we don't have any support for Controlled Ops from the frontend -> compiler in the decomp system. We'll revisit this as part of the SymOp decomposition support in the future.

runtime/lib/DecompGraphSolver/DGTypes.hpp
maliasadi: This comparison is mainly used for circuit gates & rules with the target gateset. The target gateset is captured and lowered as a list of strings. Previously, we had to comment out `h2` and `h3` because info regarding `numWires` and `numParams` was unavailable. Now that E2E workflow is complete, we can try to resolve this TODO in the solver.


Update:

@astralcai  The hash function now only considers the name of the operator for hashing, which means that different `OperatorNode` instances with the same name but different numWires, numParams, or adjoint values will have the same hash. We deal with ops with arbitrary wires (e.g., MultiRZ) using the `operator=` method. This is intentional to allow for wildcard matching based on the name from the frontend and compiler, but it also means that care must be taken when using `OperatorNode` instances that may have the same name but different parameters, as they will be treated as the same key in hash-based containers. The number of wires and parameters can be encoded in the name if needed to distinguish them in the hash when converting MLIR operations to OperatorNodes.

runtime/lib/DecompGraphSolver/DGTypes.hpp
maliasadi: Yes, we can extend OpNode definition and the graph to support more custom ops.

frontend/test/lit/test_decomposition.py
```diff
@@ -1690,7 +1608,220 @@ def circuit():
...
+
+    def x_to_rx(wire):
+        qml.RX(np.pi, wire)
+
+    def y_to_ry(wire):
+        qml.RY(np.pi, wire)
+
+    def h_to_rx_ry(wire):
+        qml.RX(np.pi / 2, wire)
+        qml.RY(np.pi / 2, wire)
```
astralcai, line 1654: I see that these rules don't have resources defined. How does the graph use them?

runtime/lib/DecompGraphSolver/DGTypes.hpp
astralcai: A `MultiRZ` on 3 wires and a `MultiRZ` on 5 wires should be two different nodes in the graph. How do you currently achieve that if they have the same hash?

frontend/test/lit/test_decomposition.py
kipawaa: Resources for cpp graph decomposition are added by the `register-decomp-rule-resource` pass 😄 it's been very convenient

kipawaa: That said, this test isn't looking for resources or even to apply the graph decomposition, decomp rules or otherwise, only verifying that the names given in `fixed_decomps` and `alt_decomps` are lowered to pass options

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
```diff
@@ -0,0 +1,388 @@
...
+                        llvm::StringRef(outputNode.name).drop_front(8).drop_back(1).str();
+                }
+                ruleNode.output = outputNode;
+            }
+            else {
+                continue; // skip this rule if target_gate attribute is missing
+            }
+
+            // Convert resources attribute
+            if (auto resourcesAttr = func->getAttrOfType<DictionaryAttr>("resources")) {
```
astralcai: How do we deal with the operators with non-trivial resource params, whose decomposition rules have a resource function as opposed to a static resource dictionary? Is that out of scope for now?

runtime/lib/DecompGraphSolver/DGBuilder.cpp
```diff
@@ -0,0 +1,314 @@
...
+        // Register all target gates
+        for (const auto &[op, _] : gateset.ops) {
+            registerOp(op);
+        }
+
+        // Register all rules
+        for (RuleId ruleId = 0; ruleId < rules.size(); ruleId++) {
+            const auto &rule = rules[ruleId];
+            const auto id = registerOp(rule.output);
+            const auto output_vertex = opIdToVertex[id];
```
astralcai, line 174: If I understand correctly, we seem to be registering all known rules, and by extension, all operators associated with those rules, even if most of them are going to be irrelevant to the circuit of interest? Don't we eventually end up creating a huge graph that contains every single operator and every single decomposition rule for even the simplest circuits?

runtime/lib/DecompGraphSolver/DGBuilder.cpp
```diff
@@ -0,0 +1,314 @@
...
+
+        for (const auto &rule : rules) {
+            appendRulesForOutput(rule.output);
+        }
+        for (const auto &[op, _] : fixedDecomps) {
+            appendRulesForOutput(op);
+        }
+        for (const auto &[op, _] : altDecomps) {
+            appendRulesForOutput(op);
+        }
```
astralcai, line 153: If a decomposition rule is registered for an operator in `fixed_decomps`, theoretically that would be the only rule that the graph needs to be aware of. We shouldn't need to add any other decomposition rule for that operator to the graph right?

frontend/test/lit/test_decomposition.py
maliasadi: Ditto! the pass was introduced in [a different PR](https://github.com/PennyLaneAI/catalyst/pull/2539).

frontend/test/pytest/from_plxpr/test_decompose_transform.py
```diff
@@ -28,9 +28,392 @@
...
+from catalyst.passes import graph_decomposition
+
 
 class TestGraphDecomposition:
-    """Test the new graph-based decomposition integration with from_plxpr."""
+    """Test the graph decomposition transform."""
+
+    @pytest.mark.usefixtures("use_capture")
+    def test_inordered_params_precompiled_rule(self):
+        """Test that unordered parameters in precompiled rules are handled correctly."""
```
kipawaa: What is meant by "unordered" here? I'm not sure I understand the purpose of this test.

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
maliasadi: We currently have a partial support for these rules. This will be handled by the [register-decomp-rule-resource](https://github.com/PennyLaneAI/catalyst/pull/2539) pass as, in the current support, rules will be captured+compiled with static number of wires & params; meaning the number of iterations in a structured decomposition rule is known at compile-time and will be collected correctly. For conditional branches, we rely on the default behaviour of the Analytics pass in MLIR (`qml.specs`) to collect and register resources, that counts ops in all branches if the value of the branch is unknown at compile-time. 

A generic support is out-of-scope for this work. We'll revisit this in the python callback discovery.

frontend/test/pytest/from_plxpr/test_decompose_transform.py
```diff
@@ -28,9 +28,392 @@
...
+            qml.SingleExcitationMinus(0.5, wires=[0, 1])
+            qml.DoubleExcitation(0.5, wires=[0, 1, 2, 3])
+            return qml.expval(qml.Z(0))
+
+        expected_resources = {"GlobalPhase": 6, "RX": 6, "RY": 30, "CNOT": 24, "Hadamard": 12}
+        resources = qml.specs(circuit, level="device")()["resources"].gate_types
+        assert resources == expected_resources
+
+    @pytest.mark.usefixtures("use_capture")
+    def test_alt_decomps(self):
```
kipawaa, line 93: I think it would be best to test user functions with a mock precompiled rules file, so that we can be sure no built-in rules are being used instead.

frontend/test/pytest/from_plxpr/test_decompose_transform.py
```diff
@@ -28,9 +28,392 @@
...
+            qml.PauliX(0)
+            qml.RX(0.1, wires=0)
+            return qml.expval(qml.PauliX(0))
+
+        expected_resources = {"RY": 1, "RZ": 2}
+        resources = qml.specs(circuit, level="device")()["resources"].gate_types
+        assert resources == expected_resources
+
+    @pytest.mark.usefixtures("use_capture")
+    def test_multi_graph_decomposition(self):
```
kipawaa, line 194: So great to see this running! 🥳

kipawaa (commented):
Great testing, and I especially appreciate the xfails for the future work! Just one question + request :)

runtime/lib/DecompGraphSolver/DGBuilder.cpp
maliasadi: We need to filter rules during compilation anyways, but doing this at the MLIR level for intermediate unsupported ops can be slow if there's too much back-and-forth between the compiler and the solver! To answer your question: Yes, we build the graph with all ops and rules. But it only takes a few milliseconds using the Boost Graph Library. This further allows us to isolate the solver logic and handle all unsupported ops effectively, even those intermediate ones that appear during decomposition

runtime/lib/DecompGraphSolver/DGBuilder.cpp
maliasadi: Yes, I don't build the graph with built-in or alternative rules for ops in `fixed_decomps`. You can find the logic in `materializeRules` (line 124-128).

runtime/lib/DecompGraphSolver/DGSolver.cpp
```diff
@@ -0,0 +1,202 @@
...
+            for (const auto &rule : graph.getAllRulesFor(root)) {
+                rules_error.push_back(rule.name);
+            }
+            throw Core::GraphSolverFailedError(
+                root, rules_error); // all rules failed for this root operator
+        }
+    }
+
+    for (const auto &[op, entry] : solvedMap) {
+        result.optimizedMap.emplace(op, entry);
```
astralcai: What's the difference between the `solvedMap` and the `optimizedMap`?

runtime/lib/DecompGraphSolver/DGSolver.cpp
```diff
@@ -0,0 +1,202 @@
...
+Core::ChosenDecompRule DecompositionSolver::solveOperator(const Core::OperatorNode &op)
+{
+    // Check if the operator has already been solved
+    if (const auto it = solvedMap.find(op); it != solvedMap.end()) {
+        return it->second;
+    }
+
+    if (visited.find(op) != visited.end()) {
+        return invalidRule(op); // cycle detected, return invalid rule to prevent infinite recursion
+    }
```
astralcai, line 110: This doesn't necessarily mean that there's a cycle right? Let's say your circuit looks like
```python
def circuit():
    qml.CRX(0.5, wires=[1, 2])
    qml.CNOT([2, 3])
```
Since `CRX` decomposes to a `CNOT`, after calling `solveOperator` for `CRX`, `CNOT` would also exist in `visited`, so now when you call `solveOperator` again on the `CNOT` in `solve`:
```cpp
for (const auto &root : result.solvedRoots) {
    const auto chosen_rule = solveOperator(root);
```
You would get an `invalidRule` and then error out?

runtime/lib/DecompGraphSolver/DGSolver.cpp
```diff
@@ -0,0 +1,202 @@
...
+            visited_.insert(currentNode_);         // add to visited in case of exceptions
+            solvingStack_.push_back(currentNode_); // push to stack in case of exceptions
+        }
+        ~VisitGuard()
+        {
+            visited_.erase(currentNode_);
+            if (!solvingStack_.empty()) {
+                solvingStack_.pop_back();
+            }
+        }
```
astralcai: I'm a bit confused about what's happening here. Could you explain the logic? What exactly is `visited` and `solvingStack` here? I previously thought that `visited` correspond to the operators that you've found a best decomposition rule for, and `solvingStack` contains operators that we are still exploring. If that's the case, why would we ever erase things from `visited`?

frontend/test/pytest/from_plxpr/test_decompose_transform.py
maliasadi: It's a simple decomposition check for multi-param gates. With "unordered", I referred to the 3 params in `qml.Rot`. In the past, we had a problem with successfully lowering and decomposing `Rot` with abstracted params. I renamed the test to fix the confusion.

runtime/lib/DecompGraphSolver/DGTypes.hpp
maliasadi: Note that we don't have support for `MultiRZ` in the graph-decomposition pass yet! But the support will be provided similar to the current graph-based decomposition by mingling the number of wires with the name of Ops and Rules via the python callback approach. In the graph, this will look like:

```
    const OperatorNode multiRZ3{"MultiRZ3"};
    const OperatorNode multiRZ5{"MultiRZ5"};
    const OperatorNode rz{"RZ"};

    const WeightedGateset gateset{{{rz, 1.0}}};

    // simple "incorrect" rules:  
    const std::vector<RuleNode> rules{
        {"multiRZ3_to_rz", multiRZ3, {{rz, 3}}},
        {"multiRZ5_to_rz", multiRZ5, {{rz, 5}}},
    };

    const DecompositionGraph graph({multiRZ3, multiRZ5}, gateset, rules);
    DecompositionSolver solver(graph);
    const auto result = solver.solve();
```

A proper solution will require updates in the lowering of Gateset and Resource Analysis MLIR tool to capture and lower multi-wires operators with variadic/static number of wires. In that case, we can revert the hash function to depend on the number of wires and number of params. We have a ticket in Q2 to enable multi-wires ops and we'll revisit the strategy when we have access to the decomp rules of multi-wires ops (sc-116231)

runtime/lib/DecompGraphSolver/DGSolver.cpp
maliasadi: No, it doesn't return an `invalidRule`! This is already solved while finding the first decomposition path. this returns the prev solved results (lines 120-122)

Example 1:

``` c++
    const OperatorNode CRX{"CRX", 2, 1, false};
    const OperatorNode CNOT{"CNOT", 2, 0, false};

    const std::vector<RuleNode> rules{ 
        {"crx_to_rz_ry", CRX, {
            {OperatorNode{"RZ"}, 2},
            {OperatorNode{"RY"}, 2},
            {CNOT, 2}}
        },
        {"cnot_to_rz", CNOT, {
            {OperatorNode{"RZ"}, 1}}
        },
    };

    const WeightedGateset gateset{
        {{OperatorNode{"RZ"}, 1.0}, {OperatorNode{"RY"}, 1.0}, {CNOT, 1.0}}};

    const DecompositionGraph graph({CRX, CNOT}, gateset, rules);

    DecompositionSolver solver(graph);
    const auto result = solver.solve();
    showSolution(result);
```

``` console
Decomposition Solution:
  Operator: RZ[w:-1][p:-1]
    Chosen Rule: BasisRule [basis]
    Total Cost: 1
    Basis Counts:
      - RZ[w:-1][p:-1]: 1
  Operator: RY[w:-1][p:-1]
    Chosen Rule: BasisRule [basis]
    Total Cost: 1
    Basis Counts:
      - RY[w:-1][p:-1]: 1
  Operator: CNOT[w:2][p:0]
    Chosen Rule: BasisRule [basis]
    Total Cost: 1
    Basis Counts:
      - CNOT[w:2][p:0]: 1
  Operator: CRX[w:2][p:1]
    Chosen Rule: crx_to_rz_ry
    Total Cost: 6
    Basis Counts:
      - CNOT[w:2][p:0]: 2
      - RY[w:-1][p:-1]: 2
      - RZ[w:-1][p:-1]: 2
```


Example 2: removing CNOT from the target gateset and the CNOT decomposition rule from `rules` will raise the following error:

``` console
  Decomposition rule not found for operator 'CRX[w:2][p:1]'. Tried rules:
    - crx_to_rz_ry
```

Example 3: just removing CNOT from the target gateset! The graph will reuse the prev solved rule for CNOT.

``` c++
    const WeightedGateset gateset2{
        {{OperatorNode{"RZ"}, 1.0}, {OperatorNode{"RY"}, 1.0}}};

    const DecompositionGraph graph2({CRX, CNOT}, gateset2, rules);

    DecompositionSolver solver2(graph2);
    const auto result2 = solver2.solve();
    showSolution(result2);
```

runtime/lib/DecompGraphSolver/DGSolver.cpp
maliasadi: They are semantically identical. `optimizedMap` inside `GraphResult is redundant and can be replaced. But I kept them separate for the ease of debugging the graph for devs. How? 
1. `solvedMap` cannot be removed. it's the memoization table that `solveOperator` reads/writes during recursion calls
2. `optimizedMap` can be removed as the graph is now quite stable and nevertheless It's just a copy of `solvedMap` packaged into `GraphResult` for external consumption

I'll remove `optimizedMap` in favour or `optimizedMap` :ok_hand: 

Update: [2b9a4fb](https://github.com/PennyLaneAI/catalyst/pull/2552/commits/2b9a4fba0f66e89d336ebcfe0b25a56967fd64fb)

runtime/lib/DecompGraphSolver/DGSolver.cpp
maliasadi: > operators we've found a best rule for...

They are stored and cached in `solvedMap`. That's the "memoization" cache that persists across calls and never erases rules/ops.

But why `visited` erases nodes is because `visited` tracks the "current" recursion path and not the permanently solved nodes. `VisitGuard` can be rewritten simply by `try-catch` instead! 

In the constructor: we mark the node as "I'm exploring this right now!" by inserting it into `visited` and the `solvingStack`. While in the destructor, we unmark it (pop it from `solvingStack`) when the function returns an invalid rule or throw an exception.

astralcai: What's the difference between `visited` and `solvingStack` then?

runtime/lib/DecompGraphSolver/DGSolver.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+        total_cost += child.totalCost * static_cast<double>(input.multiplicity);
+        for (const auto &[basis_op, count] : child.basisCounts) {
+            solution.basisCounts[basis_op] += count * input.multiplicity;
+        }
+    }
+
+    if (total_cost == 0.0) {
+        return invalidRule(solution.op); // invalid rule
+    }
```
astralcai: What about "null" rules? For example, the user could provide an empty decomposition rule for `GlobalPhase` (if they don't care about global phases and just want to decompose them to nothing). In the PL implementation, there is some special logic that deals with this. Is this supported here?

maliasadi: Null rules are not supported in catalyst. Could you share the unit tests in PL? We can definitely add this as a follow-up; this will be quite straightforward and don't want to make this PR complicated as it.

runtime/lib/DecompGraphSolver/DGSolver.cpp
maliasadi: It's for a special optimization that I'm doing in the graph for fast O(1) node lookup and cycle detection. They track the same set of nodes but serve different purposes:

`solvingStack` (vector) is responsible for:
- Ordered trace for diagnostics like "what's the exact sequence of ops so far?"
- It preserves insertion order for printing the decomp graph in order in case of failures or to avoid resolving pre-solved nodes 

while `visited` that is an unordered set:
- it's used for a O(1) cycle detection and unordered_set O(1) lookup for node lookups, especially useful when we want to switch over a new/different decomp path.
- Also `visited.count(node) -> boolean` check before recursing through the graph.

runtime/lib/DecompGraphSolver/DGSolver.cpp
astralcai: So [here](https://github.com/PennyLaneAI/pennylane/blob/4c31d796dcaa6f6ef96cc46b311f9171ea26102e/tests/transforms/test_decompose_transform_graph.py#L489
) is a test where `null_decomp` is explicitly assigned to `GlobalPhase` so that it is removed from the circuit. There is also the more subtle case of `Pow` decomposition rules. For example, a `Pow(Hadamard, z=2)` is decomposed to nothing because `Hadamard` is self-inverse.

astralcai (approved):
Thanks for answering all my questions! The graph solver makes sense! It actually makes more sense then the one I originally wrote in PL, where all the edges are inverted just so that I could use `djikstra_search` on it 🤷🏻‍♂️.

runtime/lib/DecompGraphSolver/DGSolver.cpp
maliasadi: Thanks! we'll add this in a follow-up PR (sc-116539)

runtime/lib/DecompGraphSolver/DGTypes.hpp
astralcai: Sounds good. We can come back to figuring out a strategy here after the abstractions ADR.

doc/releases/changelog-dev.md
```diff
@@ -291,11 +291,84 @@
...
+  ``` python
+  import pennylane as qp
+  import pennylane.numpy as np
+
+  from catalyst import qjit
+  from catalyst.jax_primitives import decomposition_rule
+  from catalyst.passes import cancel_inverses, graph_decomposition, merge_rotations
+
+
+  @decomposition_rule(op_type="PauliX")
```
dime10: Q: Are we happy with a string here and/or did we consider using the class type directly?

dime10: I see that the maps passed to the transform all use the class type.

doc/releases/changelog-dev.md
```diff
@@ -291,11 +291,84 @@
...
+      qp.PauliX(0)
+      qp.RY(y, wires=0)
+      qp.PauliY(0)
+      qp.RY(x + y, wires=0)
+
+      # register custom decomposition rules, required
+      # when using the decomposition_rule decorator
+      x_to_rx(int)
+      y_to_ry(int)
+      h_to_rx_ry(int)
```
dime10, line 357: Q: Can we not trace these automatically from the rules explicitly mentioned in the decomposition decorator?

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -1549,3 +1553,112 @@ def circuit():
...
+def graph_decomposition(
+    qnode=None,
+    *,
+    gate_set: Iterable[type | str] | dict[type | str, float],
+    fixed_decomps: dict | None = None,
+    alt_decomps: dict | None = None,
+    _builtin_rule_path: Path = BYTECODE_FILE_PATH,
+):
+    R"""
+    Specify that the ``-graph-decomposition`` MLIR compiler pass for applying optimal gate
```
dime10: I believe I mentioned this in some other PR before, but the phrasing "optimal gate decompositions" may be overpromising because the rules themselves are not guaranteed to be optimal. The only guarantee we can make is that we pick the shortest (optimal) path in the decomp rule graph.

frontend/catalyst/resources/decomposition_rules.mlirbc
dime10, line 1: I thought this file would be .gitignore-d?

mlir/include/Quantum/Transforms/Passes.td
```diff
@@ -128,6 +128,45 @@ def MergeRotationsPass : Pass<"merge-rotations"> {
...
+    /*CLI name*/"alt-decomps", 
+    /*Type*/"std::string", 
+    /*Description*/"Maps an operator to a `|` separated list of alternative decomposition rules that will be considered alongside any built-in rules for the operator.">,
+    Option<
+    /*C++ name*/"bytecodeRulesFile", 
+    /*CLI name*/"bytecode-rules", 
+    /*Type*/"std::string", 
+    /*Default*/[{ "./frontend/catalyst/resources/decomposition_rules.mlirbc" }],
+    /*Description*/"A path to a bytecode file of compiled decomposition rules.">
+    ];
```
dime10, line 158: Indentation needed, it's difficult to parse currently. Unfortunately, we don't have a formatter for tablegen files.

mlir/include/Quantum/Transforms/Passes.td
```diff
@@ -128,6 +128,45 @@ def MergeRotationsPass : Pass<"merge-rotations"> {
     let dependentDialects = ["math::MathDialect"];
 }
 
+def GraphDecompositionPass : Pass<"graph-decomposition"> {
+    let summary = "Decompose gates using an MLIR-native graph-based framework.";
+
+    let description = "Example Usage: `catalyst --tool=opt --pass-pipeline='builtin.module(graph-decomposition{gate-set=RX=1.0,RY=2.0,GlobalPhase=0.0 fixed-decomps=PauliX=x_to_rx,Hadamard=hadamard_to_rz_rx alt-decomps=PauliY=y_to_ry|y_to_rot,PauliZ=z_to_rz|z_to_rot bytecode-rules=\"./rule_library.mlirbc\"})'";
```
dime10: Improved formatting for the docs: 
```suggestion
    let description = [{
        Example Usage:
        ```
        catalyst --tool=opt --pass-pipeline='builtin.module(graph-decomposition{gate-set=RX=1.0,RY=2.0,GlobalPhase=0.0 fixed-decomps=PauliX=x_to_rx,Hadamard=hadamard_to_rz_rx alt-decomps=PauliY=y_to_ry|y_to_rot,PauliZ=z_to_rz|z_to_rot bytecode-rules=\"./rule_library.mlirbc\"})'
        ```
    }];
```

frontend/test/pytest/from_plxpr/test_decompose_transform.py
```diff
@@ -45,7 +49,387 @@ def _normalize_gate_types(gate_types):
 
 
 class TestGraphDecomposition:
-    """Test the new graph-based decomposition integration with from_plxpr."""
+    """Test the graph-decomposition built-in transform."""
+
+    @pytest.mark.usefixtures("use_capture")
```
rniczh: After offline-discussion with @kipawaa , this line is redundant

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
```diff
@@ -0,0 +1,452 @@
...
+namespace catalyst {
+namespace quantum {
+#define GEN_PASS_DEF_GRAPHDECOMPOSITIONPASS
+#include "Quantum/Transforms/Passes.h.inc"
+
+struct GraphDecompositionPass : public impl::GraphDecompositionPassBase<GraphDecompositionPass> {
+    using GraphDecompositionPassBase::GraphDecompositionPassBase;
+    void runOnOperation() final
+    {
+        // Debugging output for command-line options
```
rniczh, line 54: I think you could probably use `LLVM_DEBUG` to dump the log message? And using `quantum-opt ... -debug-only=your-pass-name` to enable it. What do you think?

```
LLVM_DEBUG(
llvm::dbgs() << ... << "\n"
)
```

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
```diff
@@ -0,0 +1,452 @@
...
+                }
+                ruleNode.output = outputNode;
+            }
+            else {
+                continue; // skip this rule if target_gate attribute is missing
+            }
+
+            // Convert resources attribute
+            if (auto resourcesAttr = func->getAttrOfType<DictionaryAttr>("resources")) {
+                DictionaryAttr operations =
```
rniczh: null check?

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
```diff
@@ -0,0 +1,452 @@
...
+                               llvm::SmallVector<mlir::OwningOpRef<mlir::func::FuncOp>> &rules)
+    {
+        if (userRuleNames.empty()) {
+            return;
+        }
+
+        PassManager pm(&getContext());
+        pm.addPass(createRegisterDecompRuleResourcePass());
+        if (failed(pm.run(module))) {
+            return signalPassFailure();
```
rniczh: The caller for this function cannot capture this failure()

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
```diff
@@ -0,0 +1,452 @@
...
+        llvm::StringRef filename, mlir::MLIRContext *context,
+        llvm::SmallVector<mlir::OwningOpRef<mlir::func::FuncOp>> &ruleRegistry)
+    {
+        mlir::ParserConfig config(context);
+        mlir::OwningOpRef<mlir::ModuleOp> moduleOp =
+            mlir::parseSourceFile<mlir::ModuleOp>(filename, config);
+
+        if (!moduleOp) {
+            return;
+        }
```
rniczh, line 224: how about `assert`?

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
```diff
@@ -0,0 +1,452 @@
...
+        });
+
+        for (auto rule : llvm::make_early_inc_range(userRules)) {
+            rule->remove();
+            rules.push_back(mlir::OwningOpRef<mlir::func::FuncOp>(rule));
+        }
+    }
+
+    void getOperators([[maybe_unused]] ModuleOp module,
+                      [[maybe_unused]] std::vector<OperatorNode> &operators)
```
rniczh:
```suggestion
    void getOperators(ModuleOp module,
                      std::vector<OperatorNode> &operators)
```

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
```diff
@@ -0,0 +1,452 @@
...
+        module.walk([&](mlir::func::FuncOp func) {
+            if (func->hasAttr("target_gate")) {
+                userRules.push_back(func);
+                if (std::find(userRuleNames.begin(), userRuleNames.end(), func.getName()) !=
+                    userRuleNames.end()) {
+                    graphRules.push_back(mlir::OwningOpRef<mlir::func::FuncOp>(func.clone()));
+                }
+            }
+            return WalkResult::skip();
+        });
```
rniczh, line 264: I think you can construct the llvm::StringSet for userRuleNames at the beging, so later in this `walk`, you don't need to do the linear search every time. roughly O(N * M) vs O(N)

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
```diff
@@ -0,0 +1,452 @@
...
+     * @param opToFixedDecompName  Parsed mapping from operator name to fixed-rule name.
+     * @param setOfRules           The full list of available decomposition rules.
+     * @return Core::FixedDecomps  Mapping from OperatorNode to its fixed RuleNode.
+     */
+    FixedDecomps buildFixedDecomps(const llvm::StringMap<std::string> &opToFixedDecompName,
+                                   const std::vector<RuleNode> &setOfRules)
+    {
+        // Index rules by name for O(1) lookup instead of scanning the vector
+        // for every fixed-decomp entry.
+        std::unordered_map<std::string, const RuleNode *> rulesByName;
```
rniczh: `StringMap`? And I found you constructed the `rulesByName` twice in this pass. One is here, and the other one is in `buildAltDecomps`. Do you want to construct it from the caller and pass it into these build decomp functions?

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
```diff
@@ -0,0 +1,452 @@
...
+
+            if (it == ruleNameToFuncOp.end() || !it->second) {
+                // skip if the rule is not found or
+                // the function op is null or
+                // it is already moved
+                continue;
+            }
+            module.push_back(it->second.release());
+        }
+    };
```
rniczh:
```suggestion
    }
```

rniczh (commented):
Thanks @kipawaa, @maliasadi. It's awesome! I leaved some comments on the graph decomposition part. 👍

mlir/test/CMakeLists.txt
```diff
@@ -43,9 +45,14 @@ if (CATALYST_GTEST_AVAILABLE)
   list(APPEND DIALECT_TESTS_DEPEND CatalystUnitTests)
 endif()
 
+add_custom_target(create_test_bytecode quantum-opt --emit-bytecode --register-decomp-rule-resource ${CMAKE_CURRENT_SOURCE_DIR}/Quantum/GraphDecomposition/test_rules.mlir -o ${BYTECODE_PATH}
```
dime10: minor formatting
```suggestion
add_custom_target(create_test_bytecode 
    quantum-opt --emit-bytecode --register-decomp-rule-resource ${CMAKE_CURRENT_SOURCE_DIR}/Quantum/GraphDecomposition/test_rules.mlir -o ${BYTECODE_PATH}
```

runtime/lib/DecompGraphSolver/CMakeLists.txt
dime10, line 1: The decomp solver does not seem to depend on the runtime, and does not seem to be used by anything in the runtime, why is it here?

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
maliasadi: Ah, I didn't know about this! I'll fix this. Thanks!

Fixed [cee388f](https://github.com/PennyLaneAI/catalyst/pull/2552/commits/cee388fbac21d729949ec54c65d255c330ae32f4)

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
maliasadi: Fixed this here and elsewhere!

runtime/lib/CMakeLists.txt
```diff
@@ -1,6 +1,52 @@
...
+if(POLICY CMP0167)
+    cmake_policy(SET CMP0167 NEW)
+endif()
+
+# This is the version tested and shipped with Ubuntu 24 apt.
+# Using older versions (e.g. 1.66 on rhel8 or 1.74 on Ubuntu 22)
+# can cause failure, or will require linking the Boost random library
+# explicitly. When using 1.83 or newer, the decompsolver target
+# will not require external linking to Boost libraries.
+set(MIN_BOOST_VERSION 1.83.0)
```
dime10, line 13: It sounds like we're all good, but let's make sure to run the wheel builds on this PR to be sure :)

dime10 (changes requested):
Thanks to everyone who worked on this, this is great work 🎉

runtime/lib/CMakeLists.txt
maliasadi: We tested the wheels with Boost and `libgraphsolver.a` updates :green_apple:  We removed the GH label for faster CI checks while applying reviewers suggestions last week and we will test the wheels again after resolving code comments

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
maliasadi: This will not occur with the current implementation of `DecompRuleResourcePass` as it always specifies "operations" in a decomposition rule resources. The "operations" attribute could be empty though! But we can add this null check in case the behaviour is changed in `DecompRuleResourcePass`.

mlir/include/Quantum/Transforms/Passes.td
```diff
@@ -128,6 +128,45 @@ def MergeRotationsPass : Pass<"merge-rotations"> {
     let dependentDialects = ["math::MathDialect"];
 }
 
+def GraphDecompositionPass : Pass<"graph-decomposition"> {
```
rniczh: Because the GraphDecompose pass always cast the op to the module op in `runOperation` at the begin, I think we can specify it as module pass here. Then you can get the module op directly from `getOperation`

```cpp
ModuleOp module = cast<ModuleOp>(getOperation());
```
vs
```cpp
```

```suggestion
def GraphDecompositionPass : Pass<"graph-decomposition", "mlir::ModuleOp"> {
```

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
maliasadi: What about this?
```suggestion
            signalPassFailure();
            return;
```

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
```diff
@@ -0,0 +1,452 @@
...
+            }
+        }
+        return success();
+    }
+
+    void loadBuiltInDecompositionRules(
+        llvm::StringRef filename, mlir::MLIRContext *context,
+        llvm::SmallVector<mlir::OwningOpRef<mlir::func::FuncOp>> &ruleRegistry)
+    {
+        mlir::ParserConfig config(context);
```
rniczh: You can get the context by using `getContext()`.

```cpp
mlir::ParserConfig config(&getContext());
```

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
maliasadi: Good catch! What about logging an error instead of a generic assertion when it's failed? something like:

``` c++
        if (!moduleOp) {
            mlir::emitError(mlir::UnknownLoc::get(context))
                << "failed to load built-in decomposition rules from '" << filename
                << "': the rules file could not be parsed";
            return;
        }

```

Fixed [cee388f](https://github.com/PennyLaneAI/catalyst/pull/2552/commits/cee388fbac21d729949ec54c65d255c330ae32f4)

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
```diff
@@ -0,0 +1,452 @@
...
+                    RuleTerm term;
+                    auto res_int = operation.getValue();
+                    if (auto intAttr = mlir::dyn_cast<IntegerAttr>(res_int)) {
+                        term.op.name = operation.getName().str();
+                        if (term.op.name.find("(") != std::string::npos) {
+                            term.op.name = term.op.name.substr(0, term.op.name.find("("));
+                        }
+                        term.multiplicity = intAttr.getInt();
+                        ruleNode.inputs.push_back(term);
+                    }
```
rniczh: I have a little confused here. So previously, to construct the `ruleNode.output` for the `Adjoint(GateName)`, it will assign the `{adjoint=true, name=GateName}` to the `ruleNode.outupt`. But here, it will take the `Adjoint` as the `term.op.name`. I guess what you expected is to get the `GateName` from `GateName(1)`, right? So for `Adjoint(GateName)(1)`, now the `Adjoint` will be captured instead of the `GateName`.

```cpp
outputNode.name = outputGateAttr.getValue().str();
if (outputNode.name.starts_with("Adjoint(") && outputNode.name.ends_with(")")) {
    outputNode.adjoint = true;
    outputNode.name =
        llvm::StringRef(outputNode.name).drop_front(8).drop_back(1).str();
}
ruleNode.output = outputNode;
```

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
rniczh: Looks great!

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
rniczh: Sure, thanks!

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
rniczh: This change is functionally identical, `signalPassFailure()` returns void, so both forms do the same thing.
`signalPassFailure()` only sets a flag on the pass that the pass infrastructure checks after `runOnOperation()` finishes. It does not stop execution.

So if we want the early return happen in the pass, you should return a bool to the caller and let them handle that error:
```cpp
LogicalResult loadUserDecompositionRules(...) {
    // ...
    if (failed(pm.run(module))) {
        module.emitError("...");
        return failure();
    }
    // ...
    return success();
}
```

So the caller can check and propagate the error:
```cpp
if (failed(loadUserDecompositionRules(...))) {
    return signalPassFailure(); 
}
```

frontend/catalyst/passes/builtin_passes.py
maliasadi: This was probably in PR #2568! I agree with your point. I'll update the description as follow:

```
    Specify that the ``-graph-decomposition`` MLIR compiler pass for applying the graph-based
    decomposition should be applied to the decorated QNode during :func:`~.qjit` compilation.

    The graph-based decomposition pass decomposes gates into a weighted target ``gate_set``
    by applying user-provided and built-in decomposition rules. The graph-based framework
    allows multiple decomposition rules to be defined for a quantum operation,
    and the graph solver will determine the optimal decomposition rules to apply,
    minimizing the overall gate count or the cost according to user-specified weights.
```

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
maliasadi: Got it! Thanks -- FIxed [3710df6](https://github.com/PennyLaneAI/catalyst/pull/2552/commits/3710df630139e2dd94c6ced94c82886bc7f0504b)

mlir/include/Quantum/Transforms/Passes.td
maliasadi: Fixed [58bbc54](https://github.com/PennyLaneAI/catalyst/pull/2552/commits/58bbc548d8da568b83ac56c52ae77a09d90f5c50)

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
maliasadi: What's the difference?

mlir/include/Quantum/Transforms/Passes.td
kipawaa: Great suggestions! Changed this to a filtered pass in [cdf1b48](https://github.com/PennyLaneAI/catalyst/pull/2552/commits/cdf1b484c0725e38f9d05e62789e0bf6e75415b5)

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
kipawaa: This is a great improvement, applied in [9fd005a](https://github.com/PennyLaneAI/catalyst/pull/2552/commits/9fd005aef08b33259bed2bf3964a6247e4180ac5)

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
kipawaa: good catch! I've constructed it in the main `runOnOperation` and passed it to the builders in [56a5159](https://github.com/PennyLaneAI/catalyst/pull/2552/commits/56a5159b912da512ba5d9ce2d5031552f4e8e112)

frontend/catalyst/resources/decomposition_rules.mlirbc
kipawaa: It was added to the ignore file, but not untracked 😅 removed in [93c22d7](https://github.com/PennyLaneAI/catalyst/pull/2552/commits/93c22d738079c816e2d5009dd021c453c3ac3ced)

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
maliasadi: I quickly implemented this to check the graph solver support for Adjoint Ops. But we don't have support for Adjoint decomposition rules from PL/frontend! We can drop this from the PR as well, focusing the logic on what's actually supported

runtime/lib/DecompGraphSolver/CMakeLists.txt
maliasadi: @dime10 The graph solver has zero runtime dep and is only consumed by the MLIR `graph_decomposition` pass. I added this under `runtime/lib/` for build system convenience (reusing the Boost fetch and `Catch2` test infra), not for any arch reason!

Alternatively, it can be moved to `mlir/lib/DecompGraphSolver/` with headers in `mlir/include/DecompGraphSolver/`. But for that I need to hoist the Graph Boost Lib to the top-level CMake. This can be done as a follow-up to keep this PR focused on the pass. Let me know if you think adding this to `mlir` subdir is overall a better call.

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
kipawaa: @maliasadi it simplifies the function signature, since we can get the context directly from within the function. I've made this change (along with similar accesses to the module operation) in [9b5a58b](https://github.com/PennyLaneAI/catalyst/pull/2552/commits/9b5a58b7763a8be67bd7e7112c7a9b44c6e467bb)

doc/releases/changelog-dev.md
maliasadi: Good question/suggestion! Fixed [6d27ef4](https://github.com/PennyLaneAI/catalyst/pull/2552/commits/6d27ef4e015fc1ad7c471821cc972291152d0ee0)

dime10: 🤩

runtime/lib/DecompGraphSolver/CMakeLists.txt
dime10: For context, the runtime and the compiler are intended to be independent systems, that could be built separately. That's why we never had a single unifying cmake file. We can see this separation being broken by the modifications to the various the build & ci scripts in this pr, which add some kind of link from the runtime to the compiler.

Given what you've said, I would say it's important to move this component fully into the compiler module. Is it possible to add the boost dependency to both the runtime and the compiler separately, or is there some kind of issue with that?

mlir/include/Quantum/Transforms/Passes.td
```diff
@@ -128,6 +128,43 @@ def MergeRotationsPass : Pass<"merge-rotations"> {
...
+        /*CLI name*/"alt-decomps",
+        /*Type*/"std::string",
+        /*Description*/"Maps an operator to a `|` separated list of alternative decomposition rules that will be considered alongside any built-in rules for the operator.">,
+        Option<
+        /*C++ name*/"bytecodeRulesFile",
+        /*CLI name*/"bytecode-rules",
+        /*Type*/"std::string",
+        /*Default*/[{ "./frontend/catalyst/resources/decomposition_rules.mlirbc" }],
+        /*Description*/"A path to a bytecode file of compiled decomposition rules.">
+    ];
```
dime10, line 158: slight polish:
```suggestion
    let options = [
        ListOption<
            /*C++ name*/"targetGateSetOption",
            /*CLI name*/"gate-set",
            /*Type*/"std::string",
            /*Description*/"The accepted gates to decompose to.">,
        ListOption<
            /*C++ name*/"fixedDecompsOption",
            /*CLI name*/"fixed-decomps",
            /*Type*/"std::string",
            /*Description*/"Maps an operator to a decomposition rule that will be applied.">,
        ListOption<
            /*C++ name*/"altDecompsOption",
            /*CLI name*/"alt-decomps",
            /*Type*/"std::string",
            /*Description*/
            "Maps an operator to a `|` separated list of alternative decomposition "
            "rules that will be considered alongside any built-in rules for the operator.">,
        Option<
            /*C++ name*/"bytecodeRulesFile",
            /*CLI name*/"bytecode-rules",
            /*Type*/"std::string",
            /*Default*/[{ "./frontend/catalyst/resources/decomposition_rules.mlirbc" }],
            /*Description*/"A path to a bytecode file of compiled decomposition rules.">
    ];
```

maliasadi: we need a formatter for TD files :smile:

runtime/lib/DecompGraphSolver/CMakeLists.txt
maliasadi: > For context, the runtime and the compiler are intended to be independent systems, that could be built separately.

+100 :)

But this is currently mostly true with preexisting exceptions:
- The runtime already depends on the `./mlir` headers at build time
- The runtmie makefile also depends on LLVM_DIR for accessing headers/impl src code

The compiler also depends on the runtime 
- Gridsynth decomposition which also doesn't depend on the runtime src exists in `./runtime`

This PR is adding a new violation by adding a link-time dependency from compiler --> runtime for `libgraphsolver.a`! I'll create a PR fixing these exceptions, in particular the compiler -> runtime dependencies for Gridsynth and Graph decompositions.

There are no technical concerns with adding the boost dependency to both the runtime and the compiler seperately. This will require fetching the Boost headers **twice** and introduce some CMake code **duplication**. But moving both Gridsynth and GraphDecomp to the compiler will fix both of these build concerns.

doc/releases/changelog-dev.md
maliasadi: I'm working on this! :loading:

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
rniczh: Sure, do you mind to add a TODO here for recording?

rniczh: And you can resolve this comment anytime you want.

rniczh (approved):
Thanks for you guys 👍

doc/releases/changelog-dev.md
dime10: Nice! It could also be a follow up btw, just thought I'd bring it up because from a usability perspective this would eventually be a pretty big issue 😅

runtime/lib/DecompGraphSolver/CMakeLists.txt
dime10: > I'll create a PR fixing these exceptions, in particular the compiler -> runtime dependencies for Gridsynth and Graph decompositions.

I don't think we need to treat "Gridsynth" and "graph decomposition" in the same way, the former is actually a runtime component (code that runs while the user program is executing), while the latter is a compiler component (code that runs while the user program is being compiled). Therefor it's natural for one of them to live in the runtime module while the other lives in the compiler module.

> This will require fetching the Boost headers twice and introduce some CMake code duplication.

I don't see why this is a problem if they both need boost. There are other libraries that both are using I'm sure (including stdlib and nanobind), and that is okay!

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
maliasadi: I fixed it [088ab36](https://github.com/PennyLaneAI/catalyst/pull/2552/commits/088ab36babd26896ed5a15a52f9ed8336bc7e4e7) and did a bit of clean up

runtime/lib/DecompGraphSolver/CMakeLists.txt
maliasadi: > Therefor it's natural for one of them to live in the runtime module while the other lives in the compiler module.

This still goes against the principle that the runtime and compiler should be independent systems capable of being built separately. our prev point :)


> I don't see why this is a problem if they both need boost. 

I think we both agree that there are no technical concerns! There are only build concerns, same as nanobind and other common deps in both the compiler and the runtime.

doc/releases/changelog-dev.md
maliasadi: Sounds good! This is a good question! 

For context, this decorator is used under the hood by `qp.decompose` -- this will be only used by power users who want to use the builtin `graph_decomposition` pass instead of the fully integration `qp.decompose`. 

After working on this for 0.5h, I'll probably create a new PR if I can properly resolve this as it requires changes to the PLxPR -> Cataxpr interpreter logic and can fix some of our problems with `qp.decompose` integration as well, for which we need reviewers from the core team.

runtime/lib/DecompGraphSolver/CMakeLists.txt
dime10: > This still goes against the principle that the runtime and compiler should be independent systems capable of being built separately. our prev point :)

How so? Gridsynth and graph decomp do not rely on each other in any way, do they?

doc/releases/changelog-dev.md
dime10: > For context, this decorator is used under the hood by qp.decompose -- this will be only used by power users who want to use the builtin graph_decomposition pass instead of the fully integration qp.decompose.

Thanks for the context! Does `graph_decomposition` offer anything `qp.decompose` doesn't? I.e., is there a reason anyone would need to use `graph_decomposition`?

runtime/lib/DecompGraphSolver/CMakeLists.txt
maliasadi: They don't depend on each other, and that's exactly the point. I was conflating two things:

1. Runtime <=> Compiler independence: Gridsynth belongs in `./runtime`, as it executes at circuit runtime. GraphDecomp belongs in `./mlir` because it executes at compile time. Each lives where it's consumed. This preserves the separation! :+1: 

2. Boost as a shared dependency: Both Gridsynth and GraphDecomp use Boost, but that's an impl detail, not a coupling! They can each fetch Boost independently in their own CMake trees.

So to be clear: I'll move GraphDecomp into `mlir/lib/DecompGraphSolver/` and leave Gridsynth where it is. There will be no cross-references between the two. I'll put this in a follow-up PR. (sc-116823)

doc/releases/changelog-dev.md
maliasadi: With PR #2698, `graph_decomposition` is now fully integrated with `qp.decompose` to perform the graph-based decomposition at MLIR to support multiple decomposition transformations. `qp.decompose` will be the user interface for decomposition that also handles the decomposition of templates and defines a fallback system for SymOps and generally when the graph fails. Us (developers) and power users may use `graph_decomposition` when they want to use the graph-decomposition pass directly without any `qp.decompose` workarounds .

We'll update the docs in that PR including "sharp bits" with decompose + capture + catalyst.

PR #2698 unifies UI for decomposition:

``` python
@qml.register_resources({qml.RX: 1})
def x_to_rx(wire: int):
    qml.RX(np.pi, wire)

@qml.register_resources({qml.RY: 1})
def y_to_ry(wire: int):
    qml.RY(np.pi, wire)

@qml.register_resources({qml.RX: 1, qml.RY: 1})
def h_to_rx_ry(wire: int):
    qml.RX(np.pi / 2, wire)
    qml.RY(np.pi / 2, wire)


@qml.qjit
@qml.decompose(gate_set={qml.Rot})
@qml.transforms.merge_rotations
@qml.decompose(
    gate_set={qml.RX, qml.RY},
    fixed_decomps={qml.PauliX: x_to_rx, qml.PauliY: y_to_ry},
    alt_decomps={qml.H: [h_to_rx_ry]},
)
def circuit(x: float, y: float):
    qml.H(0)
    qml.H(0)
    qml.RX(x, wires=0)
    qml.PauliX(0)
    qml.RY(y, wires=0)
    qml.PauliY(0)
    qml.RY(x + y, wires=0)

    return qml.state()
 ```

File: runtime/lib/DecompGraphSolver/CMakeLists.txt
Comment (dime10) [reply runtime/lib/DecompGraphSolver/CMakeLists.txt:1]: Sounds good 👍

File: doc/releases/changelog-dev.md (hunk 18 above)
Comment (dime10) [reply doc/releases/changelog-dev.md:357]: Nice :)

File: frontend/catalyst/resources/decomposition_rules.mlirbc
Comment (dime10) [reply frontend/catalyst/resources/decomposition_rules.mlirbc:1]: I don't know if my interface is bugging out but the file is still here 😅 

Edit: Oh wait, is this because of the wheels issue you brought up?^^

File: .github/workflows/build-wheel-linux-arm64.yaml
Comment (dime10) [inline .github/workflows/build-wheel-linux-arm64.yaml:1]: here

File: .github/workflows/build-wheel-linux-x86_64.yaml
Comment (dime10) [inline .github/workflows/build-wheel-linux-x86_64.yaml:1]: here

File: .github/workflows/build-wheel-macos-arm64.yaml
Comment (dime10) [inline .github/workflows/build-wheel-macos-arm64.yaml:1]: here

File: .github/workflows/check-catalyst.yaml
Comment (dime10) [inline .github/workflows/check-catalyst.yaml:1]: here

File: frontend/catalyst/pipelines.py [hunk 37]
```diff
@@ -250,6 +250,7 @@ def get_quantum_compilation_stage(_options: CompileOptions) -> List[str]:
         # TODO: We can remove this pass below once PBC has its own pipeline.
         "lower-pbc-init-ops",
         "disable-assertion" if _options.disable_assertions else None,
+        "symbol-dce",
```
dime10: Without context pass pipelines can be difficult to understand. Let's add a comment here about why symbol-dce was added :)

mlir/CMakeLists.txt
dime10, line 1: here

runtime/Makefile
dime10, line 1: here

dime10 (approved):
🚀 

I'm marking all the spots we need to reverse in the [follow-up](https://app.shortcut.com/xanaduai/story/116823/move-decompgraph-library-to-mlir) with comments, but we can close them immediately.

frontend/catalyst/pipelines.py
kipawaa: Good call, added in [bcfd35e](https://github.com/PennyLaneAI/catalyst/pull/2552/commits/bcfd35e02c06513617312843c59a9910818e7179)

frontend/catalyst/resources/decomposition_rules.mlirbc
kipawaa: It was re-added then re-removed to sort out wheels stuff, all resolved now as of [3916c0b](https://github.com/PennyLaneAI/catalyst/pull/2552/commits/3916c0bd3c5ac88b367c3a0898fea2f0afa9ce3c)

PR #2556, opened 2026-03-06
Update transform sequence lowering pipeline to support multiple pipelines

mudit2812:
**Context:**
We want to have multiple compilation stages representable in MLR as separate transform sequences. The support to interpret such transform programs is being added in #2550. This PR updates the QNode MLIR lowering pipeline to leverage this architecture to lower multiple pipelines per QNode.

* Update `quantum_kernel_p` to accept a `pipelines` argument instead of `pipeline`. This should be a tuple of pairs, where the first entry is the name of the stage, and the second entry is the pipeline for that stage.
* Update `from_plxpr` to use the above. The user transform program is now provided as `(("main", pipeline),)`
* Update `jax_primitive_utils`. The most significant change is to the workflow for lowering the transform program:
  * Added a new `transform_module_lowering` function that is responsible for lowering the transform program to an MLIR module. This function calls the `transform_named_sequence_lowering` function for each pipeline in `pipelines`
  * Update `transform_named_sequence_lowering` to take a `sym_name` argument which is used to create the symbol name for the current pipeline as `__transform_{sym_name}`

**Benefits:**
Building towards more inspectable compilation stages

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-113262]


frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -482,7 +483,7 @@ def wrapper(*args):
     # Apply the corresponding Catalyst pass counterpart
     next_eval = copy(self)
     t = qml.transform(pass_name=catalyst_pass_name)
-    bound_pass = qml.transforms.core.TransformContainer(t, args=targs, kwargs=dict(tkwargs))
+    bound_pass = qml.transforms.core.BoundTransform(t, args=targs, kwargs=dict(tkwargs))
```
mudit2812, line 486: `TransformContainer` is a deprecated alias for `BoundTransform`

paul0403 (commented):
Nice and straightforward 💯 

I was gonna say can you add a frontend lit test for the case when there's multiple pipelines, but then I realized probably there's no way to do that until the user frontend for multiple pipelines is complete? aka needs #2525 ?

paul0403 (approved):

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -359,11 +359,12 @@ def calling_convention(*args):
         gateset = [_get_operator_name(op) for op in self.decompose_tkwargs.get("gate_set", [])]
         setattr(qnode, "decompose_gatesets", [gateset])
 
+    pipelines = (("main", tuple(self._pass_pipeline)),)
```
dime10, line 362: How would another transform be injected?

frontend/catalyst/jax_primitives_utils.py
```diff
@@ -349,69 +347,73 @@ def transform_named_sequence_lowering(jax_ctx: mlir.LoweringRuleContext, pipelin
...
+            from catalyst.python_interface.pass_api import is_xdsl_pass
+
+            if is_xdsl_pass(name):
+                uses_xdsl_passes = True
+                apply_registered_pass_op.operation.attributes["catalyst.xdsl_pass"] = (
+                    ir.UnitAttr.get()
+                )
 
-            transform_yield_op = YieldOp(operands_=[])  # pylint: disable=unused-variable
+        _ = YieldOp(operands_=[])
```
dime10: Even simpler? 😅 
```suggestion
        YieldOp(operands_=[])
```

dime10 (approved):
Thanks @mudit2812 🥳

frontend/catalyst/from_plxpr/from_plxpr.py
mudit2812: In #2557 , it's done like this:
```python
pipelines = (("main", (transform1, ...)), ("device", (transform2, ...)))
```

PR #2557, opened 2026-03-06
Add device preprocessing to `from_plxpr`

mudit2812:
**Context:**
* This PR adds support for device preprocessing with program capture.

* Add a new `catalyst.from_plxpr.device_utils` submodule, which contains a function for generating a device preprocessing pipeline based on device capabilities and the execution config.
  * For transforms that don't have a `pass_name`, the transform is processed as follows:
    * a dummy MLIR transform, `empty`, is used. This transform has a `key` string argument that is used to indicate to devs which transform this pass is standing in for.
    * If any dummy transforms are added to the pipeline, a warning is raised to let the user know that some dummy transforms are present.
      * Note that this is only true for dummy transforms added conditionally. Transforms that are _always_ added which don't have native implementations do not trigger this warning.
* Create a `"device"`pass pipeline that is added to the `quantum_kernel_p` binding in `from_plxpr`.
* Add a `skip_preprocess` argument to `qjit`, `trace_from_pennylane`, `from_plxpr`, and `WorkflowInterpreter`, which can be used to skip appending device preprocessing transforms to the pass pipeline.
* Add a `_preprocess_warn=True` argument to `from_plxpr` and `WorkflowInterpreter` which can be used to toggle the dummy transform warnings off. This argument is private and not exposed by `qjit`. I added it as a testing convenience.
* The preprocessing function also does MCM related validation and preprocessing.

**Benefits:**
* Device preprocessing available with program capture.

**Possible Drawbacks:**
* Depending on the device being used (and hence, which preprocessing transforms are used), the dummy transform warning can get annoying.

**Related GitHub Issues:**

[sc-111493] [sc-112073]


frontend/catalyst/from_plxpr/device_utils.py
```diff
@@ -0,0 +1,312 @@
...
+}
+
+
+def create_device_preprocessing_pipeline(
+    device: qml.devices.Device, execution_config: ExecutionConfig, shots: int, warn: bool = True
+) -> list[BoundTransform]:
+    """Create a pipeline of device preprocessing transforms for lowering QNodes."""
+    capabilities: DeviceCapabilities = get_qjit_device_capabilities(
+        get_device_capabilities(device, shots=shots)
+    )
```
albi3ro: This is rather weird syntax, but definitely outside the scope of this PR to fix...

paul0403: The `get_device_capabilities` function is actually not used anywhere in the codebase, and was just a convenience function for tests: #2114 

The code around the device capabilities in the legacy frontend has been cleaned up a bit, see the details at https://github.com/PennyLaneAI/catalyst/pull/2089#discussion_r2421256201. Essentially, in the legacy `QJITDevice` class, 

> init loads from raw toml, preprocess filters based on shots, and that's it. It's just that the unused get_device_capabilities needs to still be maintained because of the weirdness of the tests.

So I would advice against using this function. 

The existing legacy impl follows this order:
1. In `QJITDevice.__init__`: `self.capabilities = get_qjit_device_capabilities(_load_device_capabilities(original_device))` to load the toml capabilities. The `get_qjit_device_capabilities` is a thin layer of processing, doing things like taking the intersection between the toml and the supported gates in catalyst runtime (so this should always be called)
2.  In `QJITDevice.preprocess()`: 
```python
        shots_not_provided = (shots is None) or (
            isinstance(shots, qml.measurements.shots.Shots) and shots.total_shots is None
        )
        if shots_not_provided and _requires_shots(self.capabilities):
            raise CompileError(
                textwrap.dedent(
                    f"""
                {self.original_device.name} does not support analytical simulation.
                Please supply the number of shots on the qnode.
                """
                )
            )
        capabilities = filter_device_capabilities_with_shots(
            capabilities=self.capabilities,
            shots_present=bool(shots),
            unitary_support=getattr(self.original_device, "_to_matrix_ops", None),
        )
        # ... compute necessary tape transform program from `capabilities`...
```
aka the judgement of whether shots is necessary, and the filtering of capabilities with shots, happen in `QJITDevice.preprocess()`. Note that `QJITDevice.preprocess()` is not called in the plxpr path (hence the whole point of your PR).

So I think what needs to happen here is 
```python
        raw_capabilities = get_qjit_device_capabilities(_load_device_capabilities(device))
        # ... check finite_shots_only, which you're already doing below
        capabilities = filter_device_capabilities_with_shots(
            capabilities=raw_capabilities,
            shots_present=shots_present,
            unitary_support=getattr(device, "_to_matrix_ops", None),
        )
```
(note the here `device` is already `qml.devices.Device`, so it would correspond to the `self.original_device` on the legacy `QJITDevice` class)

This way we compute the same functionality as the legacy frontend, without touching the `get_device_capabilities()` function.

frontend/catalyst/from_plxpr/device_utils.py
```diff
@@ -0,0 +1,288 @@
...
+            f"'qml.qjit(capture=True)'. Currently, only 'fill_shots' or None are supported."
+        )
+
+    if mcm_config.mcm_method == MCM_METHOD.ONE_SHOT:
+        shots_present = qml.math.is_abstract(shots) or shots != 0
+        if not shots_present:
+            raise CompileError("Cannot use mcm_method='one-shot' with analytic mode.")
+        pipeline.append(
+            _safe_create_bound_transform(
+                Transform(pass_name="dynamic-one-shot"), unsupported_transforms
```
paul0403, line 138: Did you mean to pass `warn` into these? Because I'm pretty sure without it the `unsupported_transforms` list will always be empty 😅

frontend/catalyst/from_plxpr/device_utils.py
```diff
@@ -0,0 +1,288 @@
...
+        pipeline.append(_safe_create_bound_transform(split_to_single_terms, unsupported_transforms))
+
+    if not (
+        capabilities.observables
+        and capabilities.measurement_processes.keys() - {"CountsMP", "SampleMP"}
+    ):
+        if "SampleMP" in capabilities.measurement_processes:
+            pipeline.append(
+                _safe_create_bound_transform(
+                    measurements_from_samples, unsupported_transforms, args=(device.wires,)
```
paul0403, line 183: Like here, I think `measurements_from_samples/counts` still don't have mlir/xdsl implementation. The 
```python
from catalyst.device.decomposition import (
    measurements_from_counts,
    measurements_from_samples,
)
```

is still tape-based, and here it would get silently added as empty pass right? If you don't pass your `warn` arg to the `_safe_create_bound_transform`

paul0403: Note: there is a measurement_from_samples in xdsl added by @joeycarter , but I remember it was quite primitive, e.g. it only handles measurements in Z basis? But I think it was just an implementation of what ftqc team needed, not a full general purpose measurement_from_samples.

Also Joey didn't implement measurement_from_counts in xdsl.

frontend/catalyst/from_plxpr/device_utils.py
```diff
@@ -0,0 +1,288 @@
...
+        else:
+            raise CompileError(f"{device.name} does not support observables or samples/counts.")
+
+    else:
+        supported_named_obs = set(_named_obs_dict.keys()).intersection(capabilities.observables)
+        if len(supported_named_obs) != len(_named_obs_dict):
+            supported_base_obs = [_named_obs_dict[obs] for obs in supported_named_obs]
+            pipeline.append(
+                _safe_create_bound_transform(
+                    diagonalize_measurements,
```
paul0403, line 201: Same here, I don't think `diagonalize_measurements` have a mlir implementation.

The only passes added a while back were dynamic one shot #2458 , split non commuting #2437 , and split to single terms #2441

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -131,11 +132,16 @@ def _get_device_kwargs(device) -> dict:
...
 # pylint: disable=line-too-long
-def from_plxpr(plxpr: ClosedJaxpr) -> Callable[..., Jaxpr]:
+def from_plxpr(
+    plxpr: ClosedJaxpr, skip_preprocess: bool = False, _preprocess_warn: bool = True
+) -> Callable[..., Jaxpr]:
     """Convert PennyLane variant jaxpr to Catalyst variant jaxpr.
 
     Args:
         jaxpr (ClosedJaxpr): PennyLane variant jaxpr
+        skip_preprocess (bool): Controls whether or not to skip quantum device preprocessing.
```
paul0403, line 142: Reminder to add docstring for the `_preprocess_warn` arg

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -360,6 +373,12 @@ def calling_convention(*args):
         setattr(qnode, "decompose_gatesets", [gateset])
 
     pipelines = (("main", tuple(self._pass_pipeline)),)
+    if not self._skip_preprocess:
+        device_pipeline = create_device_preprocessing_pipeline(
+            qnode.device, execution_config, shots, warn=self._preprocess_warn
+        )
+        pipelines += (("device", device_pipeline),)
```
```suggestion
        device_preprocessing_pipeline = create_device_preprocessing_pipeline(
            qnode.device, execution_config, shots, warn=self._preprocess_warn
        )
        pipelines += (("device", device_preprocessing_pipeline),)
```

paul0403: Just to be a bit more explicit : )

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -490,7 +509,14 @@ def wrapper(*args):
...
 def trace_from_pennylane(
-    fn, static_argnums, dynamic_args, abstracted_axes, sig, kwargs, debug_info=None
+    fn,
+    static_argnums,
+    dynamic_args,
+    abstracted_axes,
+    sig,
+    kwargs,
+    skip_preprocess=False,
+    debug_info=None,
```
paul0403, line 523: Did you also mean to add the `warn` argument here? Because this `trace_from_pennylane` function is the overall entry point (from the `QJIT` object)

frontend/test/pytest/test_specs.py
```diff
@@ -248,11 +248,11 @@ def test_invalid_levels(self, simple_circuit):
         ):
             qml.specs(no_passes, level=-1)()
 
-        with pytest.raises(check=ValueError, match="Requested specs levels 2"):
-            qml.specs(no_passes, level=2)()
+        with pytest.raises(check=ValueError, match="Requested specs levels 10"):
+            qml.specs(no_passes, level=10)()
```
paul0403, line 252: Why are some of the specs different?

frontend/test/pytest/python_interface/inspection/test_draw_unified_compiler.py
```diff
@@ -706,7 +706,7 @@ def circuit():
     def test_empty_passpipeline(self):
         """Tests that it works with an empty pass pipeline."""
 
-        @qml.qjit
+        @qml.qjit(skip_preprocess=True)
```
paul0403, line 709: Just curious, what was your criteria for judging which tests' preprocessing to skip?

frontend/test/pytest/test_mbqc.py
```diff
@@ -242,7 +242,7 @@ def test_explicit_rz_in_mbqc(rz_angle, mbqc_pipeline):
 
     # RZ circuit in the MBQC representation
     @qjit(pipelines=mbqc_pipeline)
-    @qml.qnode(dev, mcm_method="tree-traversal")
```
paul0403, line 245: @joeycarter is tree traversal critical for you guys?

frontend/test/pytest/from_plxpr/test_preprocessing.py
```diff
@@ -0,0 +1,757 @@
...
+    def capabilities(self, obj: DeviceCapabilities):
+        """Capabilities setter."""
+        self._capabilities = obj
+        self._capabilities.qjit_compatible = True
+
+    @property
+    def qjit_capabilities(self):
+        """qjit_capabilities alias to capabilities. Used to override Catalyst behaviour
+        where it prioritizes the TOML file over device.capabilities."""
+        return self._capabilities
```
paul0403, line 61: This is a really elegant solution to the testing problem pointed out in #2114 ! Your tests never need to call the janky `get_device_capabilities()` function!

Maybe, we can even update [the logic inside `_load_device_capabilities()`](https://github.com/PennyLaneAI/catalyst/blob/fff84065364c40a08f0730780cacff8899bdbca4/frontend/catalyst/device/qjit_device.py#L540-L543), so we don't even need this bandage method on the flexible-capability test device. But of cource not in this PR. I'll update #2114 to record this solution.

frontend/test/pytest/from_plxpr/test_preprocessing.py
```diff
@@ -0,0 +1,757 @@
...
+        assert pipelines
+        device_pipeline = pipelines[1][1]
+
+        if transform_needed:
+            assert any(t.pass_name == "split-to-single-terms" for t in device_pipeline)
+        else:
+            assert not any(t.pass_name == "split-to-single-terms" for t in device_pipeline)
+
+    @pytest.mark.parametrize(
+        "non_sample_supported,transform_needed", [(True, False), (False, True)]
```
paul0403: Small nitpick: I would prefer to do this like 
```python
has_transform = any(pass is measurement_from_samples for pass in device_pipeline)
if sample_supported:
   assert has_transform
else:
   assert not has_transform
```

Seems a bit more straightfoward to not have to deal with an extra negation 😆 also, the second "parametrization" isn't really a parametrization if it actually depends on the first one right?

But this is mostly a matter of taste. Feel free to ignore if you don't want to update.

frontend/test/pytest/from_plxpr/test_preprocessing.py
```diff
@@ -0,0 +1,757 @@
...
+        if transform_needed:
+            assert any(
+                t.pass_name == "empty" and t.kwargs["key"] == "measurements_from_counts"
+                for t in device_pipeline
+            )
+        else:
+            assert not any(
+                t.pass_name == "empty" and t.kwargs["key"] == "measurements_from_counts"
+                for t in device_pipeline
+            )
```
paul0403: But at least we should change these double `any` checks to 

```python
has_transform = any(...)
if transform_needed:
      assert has_transform
else:
      assert not has_transform
```

to save some testing time 😅

frontend/test/pytest/from_plxpr/test_preprocessing.py
```diff
@@ -0,0 +1,757 @@
...
+        cjaxpr = from_plxpr_no_warn(jaxpr, skip_preprocess=False)()
+
+        pipelines = None
+        for eqn in cjaxpr.eqns:
+            if eqn.primitive == quantum_kernel_p:
+                pipelines = eqn.params.get("pipelines", None)
+                break
+
+        assert pipelines
+        device_pipeline = pipelines[1][1]
```
paul0403: I feel like this snippet shows up often enough to be factored out into its own testing util function in this file.

frontend/test/pytest/from_plxpr/test_preprocessing.py
```diff
@@ -0,0 +1,757 @@
...
+                if eqn.params["qnode"].__name__ == "f1":
+                    f1_pipelines = eqn.params["pipelines"]
+                elif eqn.params["qnode"].__name__ == "f2":
+                    f2_pipelines = eqn.params["pipelines"]
+
+        assert f1_pipelines
+        assert f2_pipelines
+
+        assert f1_pipelines[0][1][0].pass_name == "merge-rotations"
+        assert f2_pipelines[0][1][0].pass_name == "cancel-inverses"
```
paul0403, line 629: One small thing I'm worried about as I'm reading these tests is, are we expecting the user pipeline to always come before the device pipeline? I know it's because you're just appending to the pipelines tuple in the implementation, but I guess it would be better to document this somewhere in a comment, since logically there's no inherent order between the two.

paul0403: In fact, looking at [the mlir `--apply-transform-sequence` pass](https://github.com/PennyLaneAI/catalyst/blob/7e41efeab1a4c62ce2e32d37f4e75be889a5f60d/mlir/lib/Catalyst/Transforms/ApplyTransformSequencePass.cpp#L180), it doesn't actually pick out any `transform.named_sequence` as the main entry point, but simply walks over the IR, and apply each transform sequence as they are encountered. 

In fact, since [the walker visits the ops from top to bottom](https://github.com/llvm/llvm-project/blob/9963fe7d7e562f5d86b0d5bcac78f6c4d8e696a8/mlir/lib/IR/Visitors.cpp#L21) (notice the triple nested op/region/block/op loop), this means we will actually visit the user transforms before the device preprocessing transforms in the current logic!

I guess we should make sure the device preprocessing transforms come as the absolute first transform sequence?

paul0403 (commented):
Amazing PR! This is a big step towards the device capabilities mliestone in capture! 💯 🥇

frontend/test/pytest/test_mbqc.py
albi3ro: Was it even doing anything before?

frontend/test/pytest/test_specs.py
albi3ro: There's more levels because the device levels now exist.

frontend/catalyst/from_plxpr/device_utils.py
mudit2812: @paul0403 warn is True by default

frontend/catalyst/from_plxpr/device_utils.py
mudit2812: @paul0403 warn is True by default

frontend/catalyst/from_plxpr/from_plxpr.py
mudit2812: We always want to warn when using capture+qjit. `_preprocess_warn` only exists for us to use for testing, which is why it's only present in `from_plxpr`, not in `trace_from_pennylane` or `QJIT`

frontend/test/pytest/python_interface/inspection/test_draw_unified_compiler.py
mudit2812: Here we wanted to test that no drawings are created if the transform program is empty, so `skip_preprocess` is needed to enforce that. In all other tests, I did my best to test with both `True` and `False`

frontend/test/pytest/test_mbqc.py
mudit2812: No it wasn't doing anything. I confirmed on the FTQC slack channel that it is not necessary

frontend/test/pytest/from_plxpr/test_preprocessing.py
mudit2812: Yes, that makes total sense. I'll make the change

frontend/test/pytest/from_plxpr/test_preprocessing.py
mudit2812: No, device transforms should always happen after user transforms. User transforms do not guarantee that the final program will be executable on the target backend, so we need further processing to ensure that is the case

frontend/catalyst/from_plxpr/device_utils.py
paul0403: But if nothing else calls `_safe_create_bound_transform` with `warn`, and `warn` arg defaults to true, isn't the `warn` argument on that function unnecessary? Because I think this function should always append to the `unsupported_transforms` list, regardless of warn or not: the top-level warn toggle only controls if the warnings are printed or not (via the final `if unsupported_transforms and warn: warnings.warn(...)` in the `create_device_preprocessing_pipeline()` function)

frontend/test/pytest/from_plxpr/test_preprocessing.py
paul0403: > device transforms should always happen after user transforms. User transforms do not guarantee that the final program will be executable on the target backend

Ah yes! Of course, what was I thinking :sweat_smile:

frontend/test/pytest/from_plxpr/test_preprocessing.py
```diff
@@ -0,0 +1,757 @@
...
+# limitations under the License.
+"""
+Test suite for device preprocessing with program capture.
+"""
+
+from functools import partial
+
+import jax
+import pennylane as qml
+import pytest
```
paul0403, line 22:
```suggestion
import jax
import pytest

import pennylane as qml
```

albi3ro (commented):
changelog?

frontend/catalyst/from_plxpr/device_utils.py
mudit2812: `warn=False` does get used [here](https://github.com/PennyLaneAI/catalyst/blob/f93106ea691adc08c00e8ff413993136962cb192/frontend/catalyst/from_plxpr/device_utils.py#L217). The point of that argument is that we want device preprocessing transforms that do not yet exist to raise a warning, but the transforms that are added unconditionally (`validate_measurements`, `verify_operations`) shouldn't raise this warning otherwise the user will get a warning EVERY time compilation happens

frontend/test/pytest/from_plxpr/test_preprocessing.py
mudit2812: isort will undo this change

frontend/catalyst/from_plxpr/device_utils.py
paul0403: Ah ok! Thanks for explaining!

frontend/catalyst/from_plxpr/device_utils.py
mudit2812: Thanks Paul, updated accordingly

paul0403 (approved):
🥳

albi3ro (approved):
Looks great! Thanks so much for getting this all done. 🚀

PR #2558, opened 2026-03-09
Workflow: master -> main

JerryChen97:
We started to rename `master` to `main` recently:
https://github.com/PennyLaneAI/pennylane/commit/3c2ff8afa518171011b97e380dc50d1daf78437c
https://github.com/PennyLaneAI/pennylane-lightning/commit/c12077636280d3fe099390ceec45d2bb30738c1d

As a result, the LLL workflow should also target `main` of Pennylane

### DIFF
--- a/.github/workflows/check-pl-compat.yaml
+++ b/.github/workflows/check-pl-compat.yaml
@@ -153,7 +153,7 @@ jobs:
       uses: actions/checkout@v4
       with:
         repository: PennyLaneAI/pennylane-lightning
-        ref: main
+        ref: master
         path: lightning_build
         fetch-depth: 0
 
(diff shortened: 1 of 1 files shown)


andrijapau (approved):
LGTM, do you mind just updating the description to provide context?

.github/workflows/check-pl-compat.yaml
```diff
@@ -153,7 +153,7 @@ jobs:
       uses: actions/checkout@v4
       with:
         repository: PennyLaneAI/pennylane-lightning
-        ref: master
+        ref: main
```
JerryChen97:
```suggestion
        ref: master
```
I noticed that lightning seems to have not yet had their own `main` for now. So probably just keep it as is. @runora95 @andrijapau

dime10 (approved):
Thanks Yushao!

PR #2562, opened 2026-03-10
Add a new pass to mark entry point args non-writable for bufferization

mehrdad2m:
**Context:**
qjitted functions can mutate numpy input buffers since bufferization treats entry-point arguments as writable and reuses them. This bug was exposed in `qml.math.decomposition.givens_decomposition`, where repeated calls on the same input produced corrupted results.

Added a pass to Mark entry-point arguments as non-writable before one-shot bufferization. This prevents bufferization from reusing input buffers for in-place updates. 

**Benefits:**
This preserves functional input semantics for qjit by making a copy of the input buffers before writing onto them.

**Possible Drawbacks:**
May introduce additional buffer copies during compilation, with a potential performance or memory cost in some cases.

**Related GitHub Issues:**
[sc-113518]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -175,6 +175,7 @@
       return qp.expval(qp.X(0))
 
   circuit()
+  ```
 
 * Added a pass to compute resource metrics of functions marked with the `target_gate` attribute,
   effectively filtering for decomposition rules in the MLIR-native decomposition framework.
@@ -320,8 +321,12 @@
   were updated in-place. Entry-point arguments are now treated as non-writable during bufferization,
   preserving the expected immutability of user inputs.
   [(#2562)](https://github.com/PennyLaneAI/catalyst/pull/2562)
+  
+* Fixed a bug in the `split-non-commuting` pass where dead `NamedObsOp`s were left behind after
+  erasing composite obs (`TensorOp`, `HamiltonianOp`).
+  [(#2567)](https://github.com/PennyLaneAI/catalyst/pull/2567)
 
-* Fix a bug where `draw_graph` failed at rendering measurements containing scalar products of observables. 
+* Fix a bug where `draw_graph` failed at rendering measurements containing scalar products of observables.
   [(#2545)](https://github.com/PennyLaneAI/catalyst/pull/2545)
 
 * Fixed a bug where the unified compiler would trigger a passed callback function 1 extra time for the initial pass level.
@@ -649,7 +654,7 @@
   }
   ```
 
-* A new MLIR op, `MCMObsOp`, is defined as a pseudo-observable of mid-circuit measurements for use in 
+* A new MLIR op, `MCMObsOp`, is defined as a pseudo-observable of mid-circuit measurements for use in
   measurement processes. It is also registered in xDSL.
   [(#2458)](https://github.com/PennyLaneAI/catalyst/pull/2458)
   [(#2536)](https://github.com/PennyLaneAI/catalyst/pull/2536)
--- a/frontend/catalyst/_version.py
+++ b/frontend/catalyst/_version.py
@@ -16,4 +16,4 @@
 Version number (major.minor.patch[-label])
 """
 
-__version__ = "0.15.0-dev47"
+__version__ = "0.15.0-dev48"
--- a/mlir/lib/Quantum/Transforms/split_non_commuting.cpp
+++ b/mlir/lib/Quantum/Transforms/split_non_commuting.cpp
@@ -67,12 +67,6 @@ struct SplitNonCommutingPass : public impl::SplitNonCommutingPassBase<SplitNonCo
     /// Check if an operation is a supported measurement operation
     static bool isSupportedMeasOp(Operation *op) { return isa<ExpvalOp>(op); }
 
-    /// Check if an operation is an observable operation.
-    static bool isObservableOp(Operation *op)
-    {
-        return isa<NamedObsOp, ComputationalBasisOp, HamiltonianOp, TensorOp, HermitianOp>(op);
-    }
-
     /// Calculate the number of groups: one group per observable
     static int calculateNumGroups(func::FuncOp funcOp)
     {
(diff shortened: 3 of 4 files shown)


mlir/include/Quantum/Transforms/Passes.td
```diff
@@ -50,6 +50,15 @@ def CopyGlobalMemRefPass : Pass<"cp-global-memref"> {
     ];
 }
 
+def MarkEntryPointArgsNonWritablePass : Pass<"mark-entry-point-args-non-writable", "mlir::ModuleOp"> {
```
paul0403: Probably makes more sense to put this in `Catalyst` dialect, since it's not really quantum related?

mlir/lib/Quantum/Transforms/mark_entry_point_args_non_writable.cpp
```diff
@@ -0,0 +1,56 @@
...
+#define GEN_PASS_DEF_MARKENTRYPOINTARGSNONWRITABLEPASS
+#include "Quantum/Transforms/Passes.h.inc"
+
+struct MarkEntryPointArgsNonWritablePass
+    : impl::MarkEntryPointArgsNonWritablePassBase<MarkEntryPointArgsNonWritablePass> {
+    using MarkEntryPointArgsNonWritablePassBase::MarkEntryPointArgsNonWritablePassBase;
+
+    void runOnOperation() final
+    {
+        auto module = getOperation();
```
paul0403: `module` is a keyword in cpp, can we use `mod` or `moduleOp`?

mlir/test/Quantum/EntryPointArgsNonWritable.mlir
```diff
@@ -0,0 +1,44 @@
...
+
+// ATTR-LABEL: func.func public @entry(
+// ATTR-SAME: %[[ARG0:.*]]: tensor<4xf32> {bufferization.writable = false}
+// ATTR-SAME: %[[IDX:.*]]: index
+func.func public @entry(%arg0: tensor<4xf32>, %arg1: index) -> tensor<4xf32> attributes {llvm.emit_c_interface} {
+  return %arg0 : tensor<4xf32>
+}
+
+// ATTR-LABEL: func.func private @helper(
+// ATTR-SAME: %[[ARG0:.*]]: tensor<4xf32>
```
paul0403:
```suggestion
// ATTR-SAME: %[[ARG0:.*]]: tensor<4xf32>)
```

How about adding the `)` to check that is attribute is not added?

mlir/test/Quantum/EntryPointArgsNonWritable.mlir
```diff
@@ -0,0 +1,44 @@
...
+  return %arg0 : tensor<4xf32>
+}
+
+// -----
+
+// BUFFERIZE-LABEL: func.func public @entry(
+// BUFFERIZE-SAME: %[[ARG0:.*]]: memref<4xf32>
+// BUFFERIZE: %[[ALLOC:.*]] = memref.alloc() {{.*}} : memref<4xf32>
+// BUFFERIZE: memref.copy %[[ARG0]], %[[ALLOC]] : memref<4xf32> to memref<4xf32>
+// BUFFERIZE: memref.store %{{.*}}, %[[ALLOC]]{{\[}}%{{.*}}{{\]}} : memref<4xf32>
```
paul0403:
```suggestion
// BUFFERIZE: memref.store %{{.*}}, %[[ALLOC]][{{%.*}}] : memref<4xf32>
```

square brakets don't actually need to be escaped right?

mlir/include/Quantum/Transforms/Passes.td
mehrdad2m: yeah that make sense. I'll fix it

mlir/lib/Quantum/Transforms/mark_entry_point_args_non_writable.cpp
mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/2562/commits/678d802d6c8dde7ac69907b48eaddeab6e1be942

mlir/include/Quantum/Transforms/Passes.td
mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/2562/commits/5b8142c64588fa26feb930759cc0426dcab2ad48

mlir/lib/Catalyst/Transforms/mark_entry_point_args_non_writable.cpp
```diff
@@ -0,0 +1,54 @@
...
+        auto mod = getOperation();
+        auto nonWritable = BoolAttr::get(&getContext(), false);
+
+        for (auto funcOp : mod.getOps<func::FuncOp>()) {
+            if (!funcOp->hasAttr(LLVM::LLVMDialect::getEmitCWrapperAttrName())) {
+                continue;
+            }
+
+            for (auto [index, arg] : llvm::enumerate(funcOp.getArguments())) {
+                if (!isa<TensorType, MemRefType>(arg.getType())) {
```
rniczh: Do we need to check the `MemRefType`?

paul0403: You are right, we don't need to check memrefs. Good catch!

mehrdad2m: Yeah I don't think we need it for our pipeline, but I wan't sure if there is or will be any case where we would add a `MemRefType` before the main one-shot-bufferization pass, so added it for completeness. Happy to remove it if you think this scenario would not happen.

paul0403: So bufferization itself doesn't do anything to existing memref `Value`s. In addition, this attribute is only meant to be attached to tensor args to begin with https://github.com/llvm/llvm-project/blob/230e4656170e3764db28b54d91153a1117290171/mlir/include/mlir/Dialect/Bufferization/IR/BufferizationBase.td#L45

So I think it's ok to skip the memref args

mehrdad2m:
```suggestion
                if (!isa<TensorType>(arg.getType())) {
```

frontend/test/pytest/test_buffer_args.py
```diff
@@ -58,6 +59,25 @@ def test_buffer_args(fn, params):
     assert jnp.allclose(interpreted_fn(*params), jitted_fn(*params))
 
 
+def test_qjit_does_not_mutate_numpy_input_buffers():
+    """A copied input that is functionally updated must not alias the caller buffer."""
+
+    @qjit
+    def update_with_copy(x):
+        y = qml.math.copy(x)
```
dime10, line 67: Is the copy relevant to the test? Would it produce the same behaviour if x is mutated? (without the bug fix)

PR #2563, opened 2026-03-10
Add Tanner graph type to qecp dialect

joeycarter:

**Description of the Change:** This PRs adds two abstractions relating to Tanner graphs to the `qecp` dialect:

1. Type `!qecp.tanner_graph` to represent a Tanner graph by its adjacency matrix in Compressed Sparse Column (CSC) form.
2. Op `qecp.assemble_tanner` to assemble `row_idx` and `col_ptr` arrays into a CSC adjacency matrix representing a Tanner graph.

Example:

```mlir
%row_idx = arith.constant dense<[3, 3, 4, 4, 0, 1, 1, 2]> : tensor<8xi32>
%col_ptr = arith.constant dense<[0, 1, 3, 4, 6, 8]> : tensor<6xi32>
%tgraph = qecp.assemble_tanner %row_idx, %col_ptr : tensor<8xi32>, tensor<6xi32> -> !qecp.tanner_graph<8, 6, i32>

// Yields the CSC adjacency matrix:
//    ┌           ┐
//    │ 0 0 0 1 0 │
//    │ 0 0 0 1 1 │
//    │ 0 0 0 0 1 │
//    │ 1 1 0 0 0 │
//    │ 0 1 1 0 0 │
//    └           ┘
```

The equivalent type and operation have also been added to the xDSL python interface to Catalyst.

> [!NOTE]

[sc-111756]


frontend/catalyst/python_interface/dialects/qecp.py
```diff
@@ -476,6 +537,38 @@ def __init__(
...
+class AssembleTannerGraphOp(IRDLOperation):
+    """Assemble a Tanner graph in CSC form from the given input arrays."""
+
+    name = "qecp.assemble_tanner"
+
+    assembly_format = """
+            $row_idx `,` $col_ptr attr-dict `:` type($row_idx) `,` type($col_ptr) `->` type($tanner_graph)
+        """
+
+    row_idx = operand_def(
```
multiphaseCFD, line 698: Normally, CSC matrix stores: row, col and non-zero data. I understand that we don't need to store data here since all non-zero data here are `1`s. But I believe it would be valuable to add an explanation on why non-zero elements are not stored here.

frontend/test/pytest/python_interface/dialects/test_qecp_dialect.py
```diff
@@ -214,6 +226,19 @@ def test_qecp_op_constructor_insert(self, idx):
         assert insert_op.result_types[0].k == self.k
         assert insert_op.result_types[0].n == self.n
 
+    def test_qecp_op_constructor_assemble_tanner(self):
+        """Test the constructor of the qecp.assemble_tanner op."""
+        row_idx_val = create_ssa_value(TensorType(i32, (8,)))
+        col_ptr_val = create_ssa_value(TensorType(i32, (6,)))
+        assemble_tanner_op = qecp.AssembleTannerGraphOp(
+            row_idx=row_idx_val,
+            col_ptr=col_ptr_val,
```
multiphaseCFD, line 324: Could we add a test on how to transverse a `assemble_tanner_op`?

frontend/catalyst/python_interface/dialects/qecp.py
```diff
@@ -476,6 +537,38 @@ def __init__(
...
+    col_ptr = operand_def(
+        TensorConstraint(element_type=BaseAttr(IntegerType), rank=1)
+        | (MemRefConstraint(element_type=BaseAttr(IntegerType), rank=1))
+    )
+
+    tanner_graph = result_def(TannerGraphType)
+
+    def __init__(
+        self,
+        row_idx: SSAValue | Operation,
```
multiphaseCFD, line 712: Do we want to add an attribute (X-type vs Z-type) here? It would allow us to support non self-dual code.

multiphaseCFD (commented):
Nice work Joey! A couple of Qs

frontend/catalyst/python_interface/dialects/qecp.py
joeycarter: That's already explained here:

https://github.com/PennyLaneAI/catalyst/blob/ca22a184cae1713267ad5396a97f641cbdd7518f/mlir/include/QecPhysical/IR/QecPhysicalTypes.td#L108-L148

There's a note at the top of qecp.py explaining that the MLIR definition should be taken as the source of truth, which is why I didn't copy over this long docstring :slightly_smiling_face:

frontend/test/pytest/python_interface/dialects/test_qecp_dialect.py
joeycarter: There isn't really anything to test here, at the moment. This PR is only adding the definitions of the Tanner graph type and the op that assembles this type from its component arrays; the traversal of a Tanner graph is functionality that doesn't exist yet, and won't be added until we start writing the pass that lowers the QEC logical dialect to QEC physical. We will certainly add tests then :slightly_smiling_face:

frontend/catalyst/python_interface/dialects/qecp.py
joeycarter: We could still support non-self-dual codes with this representation, we just have to add a column to the adjacency matrix for each auxiliary qubit (as well as the corresponding rows). I want to avoid assigning an attribute to the entire Tanner graph type or to the `assemble_tanner` op since in general a Tanner graph can have auxiliary qubits for both X and Z checks, even though right now we're only considering CSS codes where we split the Tanner graph into two separate values, one for X and one for Z. This is explained a bit more in the next PR #2571.

One attribute I might consider adding, although I don't think it's essential so I'm tempted to leave it off for now until we truly need it, is a way to annotate the adjacency matrix to indicate which columns correspond to data qubits, which columns correspond to X/Z-checks for CSS codes, which columns correspond to non-CSS checks, etc. I don't think this is strictly necessary since the author of the pass that adds the Tanner graph can keep track of this themselves when they define it, but it might prove useful down the line. Any thoughts?

frontend/test/pytest/python_interface/dialects/test_qecp_dialect.py
multiphaseCFD: Thanks for the clarification

multiphaseCFD (commented):
Thanks @joeycarter . One more question: what is the reason that `IntegerType` is adopted here instead of `IndexType` here? I think all components in the type and ops would be signless.

lillian542 (approved):
LGTM, thanks!

multiphaseCFD (approved):

PR #2565, opened 2026-03-10
[v0.14.1] Pin to malt verision and add gast dependency

paul0403:
**Context:**
`malt` dropped `gast` as a dependency, so we need to make the `gast` dependency explicit.
We also pin to a specific `malt` version, where `gast` has not been dropped by it yet.

This change is for a bug fix release.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -108,7 +108,7 @@ pl_version = dep_versions.get("pennylane")
 lq_version = dep_versions.get("lightning")
 
 pl_min_release = "0.44.1"
-lq_min_release = pl_min_release
+lq_min_release = "0.44"
 
 if pl_version is not None:
     pennylane_dep = f"pennylane=={pl_version}"  # use TestPyPI wheels, git is not allowed on PyPI


doc/releases/changelog-0.14.1.md
```diff
@@ -0,0 +1,13 @@
+# Release 0.14.1 (current release)
+
+<h3>Bug fixes</h3>
+
+* The dependency `diastatic-malt` is now explicitly pinned to version 2.15.2.
+  The dependency `gast` is now explicit.
```
dime10:
```suggestion
* The ``gast`` package is now an explicit dependency in Catalyst. The ``gast`` package was previously
  pulled in transitively by ``diastatic-malt``, but ``diastatic-malt==2.15.3`` dropped ``gast`` as a dependency, which caused an error when importing Catalyst.
```

setup.py
```diff
@@ -107,7 +107,7 @@ def parse_dep_versions():
 pl_version = dep_versions.get("pennylane")
 lq_version = dep_versions.get("lightning")
 
-pl_min_release = "0.44.0"
+pl_min_release = "0.44.1"
 lq_min_release = pl_min_release
```
dime10:
```suggestion
lq_min_release = "0.44"
```

PR #2567, opened 2026-03-10
Remove unused obs after split non commuting pass

rniczh:
**Context:**

Unused NamedObservables should be removed as well

Add unused obs to remove group

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


mlir/lib/Quantum/Transforms/split_non_commuting.cpp
```diff
@@ -70,7 +70,7 @@ struct SplitNonCommutingPass : public impl::SplitNonCommutingPassBase<SplitNonCo
     /// Check if an operation is an observable operation.
     static bool isObservableOp(Operation *op)
     {
-        return isa<NamedObsOp, ComputationalBasisOp, HamiltonianOp, TensorOp, HermitianOp>(op);
+        return isa<NamedObsOp, ComputationalBasisOp, HermitianOp>(op);
```
multiphaseCFD: 👍

multiphaseCFD (approved):
LGTM! Thanks @rniczh . Please don't forget to add this to the changelog

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
dime10: It would be nice to add a bit more context here. Reading this code it would difficult to understand why these operations are _not_ included in the observable criteria (and thus subsequently _included_ in the deletion), given that they are observables.

rniczh: Yes, I agree. I've refactored this further to remove `isObservableOp` entirely. Let it stop at qubit/qreg boundaries instead. 

[`aad02d8` (this commit)](https://github.com/PennyLaneAI/catalyst/pull/2567/commits/aad02d8ecf8d50af4248258967fa77e99876332c)

https://github.com/PennyLaneAI/catalyst/pull/2567/changes#diff-61fba649cccaeec8376f5643da27c30b382676e1f71b1635a04c0e15624e849dR236

lillian542 (approved):

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
dime10: Looks great!

PR #2568, opened 2026-03-10
Graph decompose frontend

kipawaa:
**Context:**

Provides the interface for the `graph-decompose` pass.

**Benefits:**
graph decomposition occurs in C++ at runtime, using precompiled builtin rules and jit compiled user rules.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**

[sc-113641]
[sc-101036]

### DIFF
--- a/frontend/test/lit/test_decomposition.py
+++ b/frontend/test/lit/test_decomposition.py
@@ -1761,3 +1761,36 @@ def test_cpp_decomp_empty_args():
 
 
 test_cpp_decomp_empty_args()
+
+
+def test_cpp_decomp_string_op_names():
+    """Test that cpp decomp args work with string op names."""
+
+    def y_to_xz(op, wires):
+        qml.RX(np.pi, wires)
+        qml.RZ(np.pi, wires)
+
+    @qjit(target="mlir")
+    # CHECK: transform.apply_registered_pass "graph-decomposition" with options = {
+    # CHECK-DAG: "fixed-decomps" = {PauliX = "{{.*}}", PauliZ = "{{.*}}"}
+    # CHECK-DAG: "alt-decomps" = {PauliY = ["{{.*}}", "y_to_xz"]}
+    # CHECK: } to {{%.+}} : (!transform.op<"builtin.module">)
+    @graph_decomposition(
+        gate_set={"RX", "RY", "RZ"},
+        fixed_decomps={
+            "X": lambda wires: qml.RX(np.pi, wires),
+            "PauliZ": lambda wires: qml.RZ(np.pi, wires),
+        },
+        alt_decomps={"PauliY": [lambda wires: qml.RY(np.pi, wires), y_to_xz]},
+    )
+    @qml.qnode(qml.device("lightning.qubit", wires=2))
+    def circuit():
+        qml.X(0)
+        qml.Y(1)
+        qml.Z(0)
+        return qml.probs()
+
+    print(circuit.mlir)
+
+
+test_cpp_decomp_string_op_names()


frontend/catalyst/passes/builtin_passes.py
```diff
@@ -17,10 +17,13 @@
 import copy
 import functools
 import json
+from pathlib import Path
+from typing import Iterable
 
 from catalyst.compiler import _options_to_cli_flags, _quantum_opt
 from catalyst.passes.pass_api import PassPipelineWrapper
 from catalyst.utils.exceptions import CompileError
+from catalyst.utils.precompile_decomposition_rules import BYTECODE_FILE_NAME, DEFAULT_RULE_DIR
```
maliasadi, line 28: could we create and import `BYTECODE_FILE_PATH` instead of `BYTECODE_FILE_NAME`, `DEFAULT_RULE_DIR`?

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -1480,3 +1483,70 @@ def circuit():
...
     return PassPipelineWrapper(qnode, "decompose-arbitrary-ppr")
+
+
+def graph_decomposition(
+    qnode=None,
+    *,
+    gate_set: Iterable,
+    fixed_decomps: dict | None = None,
+    alt_decomps: dict | None = None,
+    rule_path: Path = DEFAULT_RULE_DIR / BYTECODE_FILE_NAME,
```
maliasadi: Probably add `builtin`, `standard` or something else to the name of this optional kwarg to better convey the use case:

```suggestion
    builtin_rule_path: Path = DEFAULT_RULE_DIR / BYTECODE_FILE_NAME,
```

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -1480,3 +1483,70 @@ def circuit():
...
+        The QNode itself will not be changed or transformed by applying these decorators.
+
+        As a result, circuit inspection tools such as :func:`~.draw` will continue
+        to display the circuit as written in Python.
+
+        To instead view the optimized circuit, the MLIR must be viewed
+        after the ``"QuantumCompilationStage"`` stage via the
+        :func:`~.get_compilation_stage` function.
+
+    Args:
```
maliasadi, line 1513: Update Args

frontend/catalyst/passes/builtin_passes.py
maliasadi: This is also a dev utility and it's not supposed to be used by users.

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -1480,3 +1483,70 @@ def circuit():
...
+        fn (QNode): the QNode to apply the graph decomposition compiler pass to.
+
+    Returns:
+        ~.QNode:
+
+    **Example**
+
+    TODO add an example once the implementation is complete.
```
maliasadi: We could use the example in [the ADR](https://github.com/PennyLaneAI/adrs/pull/92)

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -1480,3 +1483,70 @@ def circuit():
...
+        return functools.partial(
+            graph_decomposition,
+            gate_set=gate_set,
+            fixed_decomps=fixed_decomps,
+            alt_decomps=alt_decomps,
+            rule_path=rule_path,
+        )
+
+    if not isinstance(gate_set, dict):
+        gate_set = {op.__name__: 1.0 for op in gate_set}
```
maliasadi: Need to update the logic to support string gates too. From PL, users can provide a mix of PL Op types and Op strings: https://github.com/PennyLaneAI/pennylane/blob/494cd943c4ce675f8c392a9b8fb98626d1b78c02/pennylane/decomposition/gate_set.py#L24

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -1480,3 +1483,70 @@ def circuit():
...
+
+    if alt_decomps:
+        options |= {
+            "alt_decomps": {
+                op.__name__: tuple(rule.__name__ for rule in rules)
+                for op, rules in alt_decomps.items()
+            }
+        }
+
+    graph_decomposition_pass = {"graph-decomposition": options}
```
maliasadi, line 1569: Thanks for incorporating `fixed_decomps` and `alt_decomps` as options in the pass! This is better than the original suggestion in the ADR 🤩

frontend/test/lit/test_decomposition.py
```diff
@@ -1674,3 +1675,77 @@ def circuit():
...
+        qml.RY(np.pi, wire)
+
+    def h_to_rx_ry(wire):
+        qml.RX(np.pi / 2)
+        qml.RY(np.pi / 2)
+
+    @qjit(target="mlir")
+    # CHECK: "graph-decomposition" with options = {
+    # CHECK-DAG: "gate-set" = {Hadamard = 1.000000e+00 : f64, RX = 1.000000e+00 : f64, RY = 1.000000e+00 : f64}
+    # CHECK-DAG: "fixed-decomps" = {PauliX = "x_to_rx", PauliY = "y_to_ry"}
```
maliasadi, line 1712: Could we sync the name from the frontend down to the IR? I can also fix this in the pass PR later but we can use underscore everywhere, e.g. `fixed_decomps` instead of `fixed-decomp`, or `alt_decomps` instead of `alt-decomps`.

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -1480,3 +1483,70 @@ def circuit():
         ...
     """
     return PassPipelineWrapper(qnode, "decompose-arbitrary-ppr")
+
+
+def graph_decomposition(
+    qnode=None,
+    *,
+    gate_set: Iterable,
```
maliasadi: valid types for gate_set for the reference: https://github.com/PennyLaneAI/pennylane/blob/494cd943c4ce675f8c392a9b8fb98626d1b78c02/pennylane/transforms/decompose.py#L378

long-term, we need compatibility with [Gateset](https://github.com/PennyLaneAI/pennylane/blob/494cd943c4ce675f8c392a9b8fb98626d1b78c02/pennylane/decomposition/gate_set.py#L24).

frontend/test/lit/test_decomposition.py
kipawaa: I think this may be an option, though we've got a pretty strong convention for dashes in the CLI options, following MLIR pass conventions, and python won't allow dashes for variable names. I'm not sure about making this type of change. Though I do love the idea of unifying the names, I think adhering to convention might take precedence here. What are your thoughts given this?

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -1480,3 +1483,70 @@ def circuit():
...
+        gate_set = {op.__name__: cost for op, cost in gate_set.items()}
+
+    options = {"gate_set": gate_set, "bytecode-rules": str(rule_path)}
+
+    if fixed_decomps:
+        options |= {
+            "fixed_decomps": {op.__name__: rule.__name__ for op, rule in fixed_decomps.items()}
+        }
+
+    if alt_decomps:
```
maliasadi, line 1561: from my branch:

```suggestion
    if alt_decomps is not None:
```

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -1480,3 +1483,70 @@ def circuit():
...
+        )
+
+    if not isinstance(gate_set, dict):
+        gate_set = {op.__name__: 1.0 for op in gate_set}
+    else:
+        gate_set = {op.__name__: cost for op, cost in gate_set.items()}
+
+    options = {"gate_set": gate_set, "bytecode-rules": str(rule_path)}
+
+    if fixed_decomps:
```
maliasadi, line 1556:
```suggestion
    if fixed_decomps is not None:
```

frontend/catalyst/passes/builtin_passes.py
kipawaa: This is more related to [this PR](https://github.com/PennyLaneAI/catalyst/pull/2531), but I'll implement it there so that we can sync it up the branches. I think it makes sense to do this after merging this branch though, to reduce the amount of syncing required.

frontend/test/lit/test_decomposition.py
maliasadi: Ok, sounds good if this is the current convention at MLIR :+1:

frontend/catalyst/passes/builtin_passes.py
kipawaa: I changed the name to `_builtin_rule_path` [here](https://github.com/PennyLaneAI/catalyst/pull/2568/commits/ffac7d7d948676d6ef79639bc948a1ce9ae34022) - I think having this parameter around for a bit longer to allow different rule caches in development and testing might be useful, but I'm open to removing it if you think it should not be exposed at all.

frontend/catalyst/passes/builtin_passes.py
kipawaa: Thanks! [91a3fd5](https://github.com/PennyLaneAI/catalyst/pull/2568/commits/91a3fd54669c4ce14c7350b5906e57b81701eb12)

frontend/catalyst/passes/builtin_passes.py
kipawaa: Good point, updated here: [91a3fd5](https://github.com/PennyLaneAI/catalyst/pull/2568/commits/91a3fd54669c4ce14c7350b5906e57b81701eb12). It might be worth adding sample IR output once we merge the pass PR, but I think this is good for now.

frontend/catalyst/passes/builtin_passes.py
kipawaa: [updated](https://github.com/PennyLaneAI/catalyst/pull/2568/commits/c488b72e77d55a8307e9ed21d72ca747a65283a7) to use the `to_name` utility from PL, so we should have near-identical support. I've also [added some tests](https://github.com/PennyLaneAI/catalyst/pull/2568/commits/c2be377369a7aa37e3bf62e4d1bb50403405d2ac) for this.

frontend/catalyst/passes/builtin_passes.py
kipawaa: I updated the types [here](https://github.com/PennyLaneAI/catalyst/pull/2568/commits/c488b72e77d55a8307e9ed21d72ca747a65283a7), not sure about supporting Gateset at the moment but it looks as though adding a `__dict__` method to gateset that we can use internally here would be quite reasonable.

frontend/catalyst/passes/builtin_passes.py
kipawaa: Is your goal to treat `{}` differently from `None`? I was intentionally looking for the same behaviour (no argument to the compiler), and I've [added tests](https://github.com/PennyLaneAI/catalyst/pull/2568/commits/82c0343921fef2963aa880304c4e72d4c8cef0c2) to reflect this

maliasadi: This is either None or a non-empty dict of ops: rules.

kipawaa: if we make the suggested change, then `graph_decomposition(..., alt_decomps = {})` will result in the transform having option `"alt-decomps" = {}`. We could add some method of ensuring that the dict is non-empty, but as of right now an empty dict is a valid input and I think is handled best by the falsy route through this if statement.

maliasadi: I used `is not None` as both `alt_decomps` and `fixed_decomps` are optional kwargs and I originally wanted to differentiate when these flags are specified in the pass even when they are empty dicts; following the workaround in the PL's python implementation. But this is actually not needed for the C++ graph. All good!

frontend/test/pytest/test_specs.py
```diff
@@ -229,6 +229,7 @@ def test_invalid_levels(self, simple_circuit):
         with pytest.raises(check=ValueError, match="Requested specs levels 2, 3"):
             qml.specs(no_passes, level=[2, 3])()
 
+    @pytest.mark.xfail(reason="pending on specs fix")
```
dime10, line 232: It's good practice to provide a pr, issue, or ticket reference so that anyone can check later whether the referenced fix has been completed or not

frontend/catalyst/resources/decomposition_rules.mlirbc
dime10, line 1: The plan is to check the file into the git tree now?

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -1480,3 +1485,87 @@ def circuit():
...
+def graph_decomposition(
+    qnode=None,
+    *,
+    gate_set: Iterable[type | str] | dict[type | str, float],
+    fixed_decomps: dict | None = None,
+    alt_decomps: dict | None = None,
+    _builtin_rule_path: Path = DEFAULT_RULE_DIR / BYTECODE_FILE_NAME,
+):
+    R"""
+    Specify that the ``-graph-decomposition`` MLIR compiler pass for applying optimal gate
```
dime10, line 1499: Optimality is a strong claim, imo we should be careful with such language. Maybe "optimal gate decompositions" could be interpreted in different ways, but I don't think our rules themselves are necessary optimal, the only thing we can guarantee is that we choose the optimal path from the given set of rules.

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -1480,3 +1485,87 @@ def circuit():
...
+    .. note::
+
+        The QNode itself will not be changed or transformed by applying these decorators.
+
+        As a result, circuit inspection tools such as :func:`~.draw` will continue
+        to display the circuit as written in Python.
+
+        To instead view the optimized circuit, the MLIR must be viewed
+        after the ``"QuantumCompilationStage"`` stage via the
+        :func:`~.get_compilation_stage` function.
```
dime10, line 1511: We have inspection tools now beyond just looking at the MLIR, including qml.specs and catalyst.draw_graph

PR #2570, opened 2026-03-11
Add missing backticks to changelog

JerryChen97:
Without this triple backtick to close the python env the entire changelog afterwards will be regarded as Python env
image


paul0403 (approved):

rniczh (approved):
Good catch!

PR #2573, opened 2026-03-11
Fix dynamic-one-shot MLIR pass to correctly update one-shot kernel's `res_attrs`

mudit2812:
**Context:**
* If a QNode has `n` measurements and `m` returns, where `n != m`, `dynamic-one-shot` fails when fed an MLIR module that was parsed from a generic syntax string. This was happening because `func.func` is printed with `res_attrs` in generic syntax, which `dynamic-one-shot` was not handling correctly. This PR fixes this behaviour.

* Update `dynamic_one_shot.cpp` to update the `res_attrs` of the one-shot kernel such that its length matches the number of returns by the kernel.
* Add a unit test for this. I confirmed that the test was failing on `main` and passes in this branch.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-113740]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -9,6 +9,7 @@
   then performs the appropriate classical statistical postprocessing across the execution results
   from all shots.
   [(#2458)](https://github.com/PennyLaneAI/catalyst/pull/2458)
+  [(#2573)](https://github.com/PennyLaneAI/catalyst/pull/2573)
 
   With this new MLIR pass, one shot execution mode is now available when capture is enabled.
 


mudit2812:
> I wonder is this attribute always there (just not displayed), and normally deduced upon creation from the function type?

@dime10 `res_attrs` is optional ([source](https://github.com/llvm/llvm-project/blob/e868e45a754e01fde74df576af2149028c55fc68/mlir/include/mlir/Dialect/Func/IR/FuncOps.td#L297)), but I guess is always rendered in the generic format

PR #2576, opened 2026-03-11
Add qecp.decode_physical_meas op

joeycarter:

**Description of the Change:** This PRs adds the `qecp.decode_physical_meas` op to represent physical-measurement decoding (the process that decodes a set of physical measurements and returns the corresponding logical measurement).

For example:

```mlir
// e.g. from the transversal measurement of a (k = 1, n = 7) physical codeblock
%0 = ... : tensor<7xi1>
%1 = qecp.decode_physical_meas %0 : tensor<7xi1> -> tensor<1xi1>
```

> [!NOTE]

[sc-111759]


frontend/catalyst/python_interface/dialects/qecp.py
```diff
@@ -602,6 +602,35 @@ def __init__(
         super().__init__(operands=operands, result_types=(err_idx_type,))
 
 
+@irdl_op_definition
+class DecodePhysicalMeasurementOp(IRDLOperation):
+    """
+    Decode physical measurement results and return the corresponding logical measurement.
```
multiphaseCFD, line 780: As the current implementation requires measurement results stored as 0 or 1 instead of eigenvalues, would be better to add a description for this

frontend/catalyst/python_interface/dialects/qecp.py
```diff
@@ -602,6 +602,35 @@ def __init__(
...
+    """
+
```
multiphaseCFD, line 788: we might also need to add `terminalMeausrementOp` and `observable`s

multiphaseCFD: as we discussed, it's not a priority for now

frontend/catalyst/python_interface/dialects/qecp.py
joeycarter: I updated the docstring to say the measurements are taken to be in the computational basis. It should be well understand that the values are booleans given that the elements types are `i1`.

cc6bba17f

lillian542 (approved):
Thanks @joeycarter! We can continue the discussions started above about what this will need to look like for different measurement procedures separately, this seems like a good place to start :)

multiphaseCFD (approved):
Thanks @joeycarter !

PR #2577, opened 2026-03-12
Support rtio.rpc op for OQD's AWG end to end pipeline

rniczh:
**Context:**

A new `rtio.rpc` operation is added to the RTIO dialect.  It represents a host RPC call triggered by the kernel, optionally carrying runtime arguments and supporting both synchronous and async modes. The op is lowered to `rpc_send` / `rpc_recv` LLVM calls (the ARTIQ RPC wire protocol). It is required by both AWG control (`program_awg`, `awg_close`) and measurement result collection (`set_dataset`, `transfer_data`).

The RTIO dialect has no representation for RPC calls. Our use cases:

| Use case | Callee | Args from kernel | Mode |
|---|---|---|---|
| AWG setup | `program_awg` | none | synchronous |
| AWG teardown | `awg_close` | none | synchronous |
| Dataset init | `set_dataset` | key (str), value (array ptr) | synchronous |
| Measurement result | `transfer_data` | key (str), index (i64), value (f64) | async |

The ARTIQ runtime exposes `rpc_send(int32_t service_id, const char *tag, ...)` and `rpc_recv(void *result)` as extern C functions linked into the ELF. 

Added `rtio.rpc` in RTIO dialect.
This op contains the following attributes:
```
- callee:      The symbolic name of the host-side callable to invoke.
- is_async:    If present the call is async.
- rpc_id:      Unique ID to determine which RPC function be executed.
 ```

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-113163]

### CODE REVIEW

File: mlir/lib/Driver/CompilerDriver.cpp [hunk 1]
```diff
@@ -1102,4 +1102,4 @@ int QuantumDriverMainFromCL(int argc, char **argv)
     if (Verbose)
         llvm::outs() << "Compilation successful:\n" << output->diagnosticMessages << "\n";
     return 0;
-}
+}
```
mehrdad2m:
```suggestion
}

```

mlir/include/RTIO/IR/RTIOOps.td
```diff
@@ -167,6 +167,53 @@ def RTIOSyncOp : RTIO_Op<"sync"> {
...
+def RTIORPCOp : RTIO_Op<"rpc"> {
+    let summary = "Invoke a host RPC function from the kernel";
+    let description = [{
+        Represents an outgoing RPC call from the kernel to the host.
+
+        Attributes:
+        - callee:    The symbolic name of the host-side callable to invoke.
+        - tag:       type-tag string encoding the return and argument types,
+                     which is compatiable with the ARTIQ RPC wire protocol.
+                     (e.g. "n:" for void no-arg, "n:si" for void(str,i32)).
```
mehrdad2m: This ARTIQ related part seems out of context. Ideally we should try to not pollute this dialect with ARTIQ details.

mlir/include/RTIO/IR/RTIOOps.td
```diff
@@ -167,6 +167,53 @@ def RTIOSyncOp : RTIO_Op<"sync"> {
...
+        - is_async:  If present the call is async.
+        - rpc_id:    Populated by the RPC-ID assignment sub-pass.
```
mehrdad2m: I think the `rpc_id` explanation is a bit too specific. We should explain what the attribute is or why it is useful. not what we do with it in the lowering.

mlir/lib/RTIO/IR/RTIOOps.cpp
```diff
@@ -21,6 +22,36 @@
 using namespace mlir;
 using namespace catalyst::rtio;
 
+//===----------------------------------------------------------------------===//
+// RPC Tag Verification
+//===----------------------------------------------------------------------===//
+
+namespace {
+/// ARTIQ RPC tag format: <return_type>:<arg_types>
```
mehrdad2m: Hmm again I am wondering if this would be better suited in a lower level place, maybe in the the lowering pass. What do you think?

mlir/lib/RTIO/IR/RTIOOps.cpp
```diff
@@ -34,5 +65,63 @@ LogicalResult RTIOSyncOp::verify()
...
+            return emitOpError("RPC with return value must be synchronous (remove 'async')");
+        }
+        if (getResults().size() != 1) {
+            return emitOpError("tag return type '")
+                   << returnPart << "' requires exactly one result";
+        }
+        char retCode = returnPart[0];
+        if (returnPart.size() != 1) {
+            return emitOpError("tag return type must be single character (i, I, f, s, O)");
+        }
```
mehrdad2m:
```suggestion
        if (returnPart.size() != 1) {
            return emitOpError("tag return type must be single character (i, I, f, s, O)");
        }
        char retCode = returnPart[0];
```
Let's move this after the size check to avoid any potential segfaults

mlir/include/RTIO/IR/RTIOOps.td
mehrdad2m: now that I think more, this could be a general enough scheme that is not ARTIQ dependant, but still you may want lower to any arbitrary tagging system, so not tying up it up to ARTIQ seams reasonable?

mlir/lib/RTIO/IR/RTIOOps.cpp
mehrdad2m: I think this is a bit nitpicky given that currently, we only use this dialect in the context of lowering to ARTIQ. So if the work needed to move it down is substantial, I think it would be fine to have it here with a TODO for now since the dialect is not used anywhere else, but keeping it independent of the ARTIQ internal details would be nice.

mlir/lib/RTIO/Transforms/RTIOEventToARTIQPatterns.cpp
```diff
@@ -157,6 +235,127 @@ struct ChannelOpLowering : public OpConversionPattern<RTIOChannelOp> {
...
+        else {
+            // 4. For sync calls, call rpc_send(id, tag_ptr, args_ptr) and rpc_recv(slot_ptr)
+            artiq.rpcSend(serviceId, tagPtr, argsArray);
+
+            // Parse tag for return type
+            StringRef tag = op.getTag();
+            size_t colon = tag.find(':');
+            StringRef returnPart = colon != StringRef::npos ? tag.take_front(colon) : "n";
+
+            int slotSize = (returnPart == "n") ? 8 : getSlotSizeForReturnCode(returnPart[0]);
```
mehrdad2m: Why do we single out "n"? couldn't we include it in the switch case within `getSlotSizeForReturnCode`?

mlir/lib/RTIO/Transforms/RTIOEventToARTIQPatterns.cpp
```diff
@@ -157,6 +235,127 @@ struct ChannelOpLowering : public OpConversionPattern<RTIOChannelOp> {
...
+            auto slotTy = LLVM::LLVMArrayType::get(i8Ty, slotSize);
```
mehrdad2m: Can you explain the logic here? I don't understand why we allocate an array of bytes instead of using a concrete resultTy?

mehrdad2m (commented):
Thanks @rniczh, Looks great. minor comments

mlir/include/RTIO/IR/RTIOOps.td
rniczh: Yeah, that's right, for ARTIQ that tag can be derived from this operand types I think

rniczh: I removed the tag https://github.com/PennyLaneAI/catalyst/pull/2577/commits/adb3b83fe78e1ec98f4c9d102fa9cc1b9960fab8. which can be derived from the operand/result types during the conversion from rtio to artiq

mlir/include/RTIO/IR/RTIOOps.td
rniczh: Thanks! I updated the explanation here. https://github.com/PennyLaneAI/catalyst/pull/2577/commits/adb3b83fe78e1ec98f4c9d102fa9cc1b9960fab8
which is the only way that host device to communicate with the RPC server (through the rpc id to determine which function should be used)

mlir/lib/RTIO/IR/RTIOOps.cpp
rniczh: Yes, https://github.com/PennyLaneAI/catalyst/pull/2577/commits/adb3b83fe78e1ec98f4c9d102fa9cc1b9960fab8. I removed the tag, it will only be constructed when converting to the artiq llvm ir

mlir/lib/RTIO/Transforms/RTIOEventToARTIQPatterns.cpp
rniczh: originally I was want to distinguish the `n` with other types (since it's void). But yeah, I think it's fine to move it into the `getSlotSzieForReturnCode` function. It doesn't need to be separated out. https://github.com/PennyLaneAI/catalyst/pull/2577/commits/d36e3b53f073f68138dbdbe12599bb922120dccf

mlir/lib/RTIO/Transforms/RTIOEventToARTIQPatterns.cpp
rniczh: This is the wire protocol of the ARTIQ, they only allow you to pass raw bytes buffer with llvm ptr, except for the the rpc id (the first arg).

mlir/include/RTIO/IR/RTIOOps.td
mehrdad2m: Thanks, looks great. Let's update the PR description as well.

mehrdad2m (approved):
🚀

mlir/include/RTIO/IR/RTIOOps.td
rniczh: Done

paul0403 (approved):

PR #2578, opened 2026-03-12
C++ Graph decomposition builder and solver with Boost

maliasadi:
**Context:**
Quantum decomposition often requires choosing a decomposition pathway from multiple possible decomposition rules/paths until all resulting operations are expressed in a provided weighted target `gateset`. To support decomposition in MLIR, we implement a standalone C++ library to model the PL's graph-decomposition module with two exported classes: `DecompositionGraph` (constructor) and `DecompositionSolver` (optimized solver). 

The graph builder constructs the graph based on a set of operators (`OperatorNode`), target gateset (`WeightedGateSet`) and  rules (`RuleNode`) powered by [the Boost Graph Library](https://www.boost.org/doc/libs/latest/libs/graph/doc/index.html). The solver can be used to find the optimal pathway through all the available rules based on the operators and gateset. 

-  `libdelightning.so` 

**Benefits:**
- Enabling the graph-decomposition in MLIR/C++
- Supporting multiple decomposition rules for operators
- Opimizing the decomposition pathway based on weighted gateset
- Reusing solved rules from the cache system. This will let us to update the existing graph/solver without necessarily creating a new one when the rules and gateset aren't changed.

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-113536]


runtime/lib/DecompGraphLib/DecompositionGraph.cpp
```diff
@@ -0,0 +1,204 @@
...
+    Core::WeightedGateset gateset;
+    std::vector<Core::RuleNode> rules;
+
+    std::unordered_map<Core::OperatorNode, OperatorId, Core::OperatorNodeHash> opToId;
+    std::vector<Core::OperatorNode> idToOp;
+    std::unordered_map<OperatorId, Vertex> opIdToVertex;
+
+    std::unordered_map<RuleId, Vertex> ruleIdToVertex;
+    std::unordered_map<Core::OperatorNode, std::vector<Core::RuleNode>, Core::OperatorNodeHash>
+        rulesForop;
```
kipawaa:
```suggestion
        opToRules;
```
I think this would be nice for consistency with the other names :)

runtime/lib/DecompGraphLib/DecompositionGraph.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+            const auto ruleVertex =
+                boost::add_vertex(GraphVertex{VertexType::Rule, RuleVertex{ruleId}}, graph);
+            // const auto ruleVertex = boost::add_vertex(GraphVertex{VertexType::Rule,
+            // RuleVertex{ruleId}}, graph);
+            ruleIdToVertex.emplace(ruleId, ruleVertex);
+            rulesForop[rule.output].push_back(rule);
+        }
+
+        for (RuleId ruleId = 0; ruleId < rules.size(); ruleId++) {
```
kipawaa: I think this loop could be merged with the previous, rather than iterating twice

runtime/lib/DecompGraphLib/DecompositionGraph.cpp
```diff
@@ -0,0 +1,204 @@
...
+        for (const auto &[op, _] : gateset.ops) {
+            registerOp(op);
+        }
+
+        // Register all rules and create vertices
+        for (RuleId ruleId = 0; ruleId < rules.size(); ruleId++) {
+            const auto &rule = rules[ruleId];
+            registerOp(rule.output);
+            for (const auto &inout : rule.inputs) {
+                registerOp(inout.op);
```
kipawaa:
```suggestion
            for (const auto &input : rule.inputs) {
                registerOp(input.op);
```

runtime/lib/DecompGraphLib/DecompositionSolver.hpp
```diff
@@ -0,0 +1,163 @@
...
+        for (const auto &rule : all_rules) {
+            auto candidate = evalRule(rule);
+            if (!best_rule.has_value() || candidate.totalCost < best_rule->totalCost) {
+                best_rule = std::move(candidate);
+            }
+        }
+
+        if (!best_rule.has_value()) {
+            RT_FAIL("No valid decomposition rule found for operator");
+        }
```
kipawaa: I believe this is redundant with the assert above that the set of rules for this op is non-empty.

runtime/lib/DecompGraphLib/DecompositionSolver.cpp
```diff
@@ -0,0 +1,101 @@
...
+        else {
+            chosen_rule = bestRule(op);
+        }
+
+        solvedMap.emplace(op, chosen_rule);
+        visited.erase(op);
+        solvingStack.pop_back();
+        return chosen_rule;
+    }
+    catch (...) {
```
kipawaa: To me it seems that this is here to catch when an op has no decomp rules in `bestRule`, which has an assert. I don't mind catching that to either try a different decomp path, or provide a helpful error/warning, but reducing the scope of this `try... catch` so that we know that was the error would be helpful.

maliasadi: I actually have a better implementation for this. I'll rewrite this part avoiding `try-catch`

maliasadi: Fixed [f476d7e](https://github.com/PennyLaneAI/catalyst/pull/2578/commits/f476d7ee6e8868e76d08dab302a03029006a7819)

doc/code/dialects/Quantum/QuantumPasses.md
dime10, line 1: Is this PR pulling in changes from Isaacs PRs? Might be worth filtering those out to keep things clean

mlir/test/Quantum/QubitTypesTest.mlir
dime10, line 1: There seem to be more than one PR mixed in here actually

maliasadi: I rebased the branch but got pulled into a meeting and didn’t have time to properly fix the conflicts :smile: 

Fixed now :)

PR #2519, opened 2026-02-25
Add the QEC Physical dialect (MLIR and xDSL)

joeycarter:
**Context:** This is the second in a series of PRs to build a prototype QEC compilation pipeline in Catalyst. The initial prototype will consider the following layers of abstraction, from top to bottom: _abstract_ (`quantum` dialect), _QEC logical_ (`qecl` dialect; added in #2512), and _QEC physical_ (`qecp` dialect; this PR).

**Description of the Change:** This PR adds the MLIR QEC physical dialect, called `qecp`. This is an initial PR that adds all of the dialect boilerplate and the three fundamental types of the QEC physical program representation: `!qecp.qubit`, `!qecp.codeblock` and  `!qecp.hyperreg`. Operations within the `qecp` dialect that act on these types will be added in an upcoming PR. A mirror of this MLIR dialect has also been added as an xDSL dialect in the Python interface to Catalyst.

> [!NOTE]
> This dialect is experimental and subject to change extensively in the future.

[sc-112450]

PR #2581, opened 2026-03-12
Revert addition of hierarchical qubits and registers

joeycarter:
**Context:** In preparation for QEC work in Catalyst, we added hierarchical qubit and register types to the `quantum` dialect in #2369 and #2431, respectively. These qubit abstraction are now being defined in the QEC logical and physical dialects (see for example #2512 and #2519) and are no longer suitable to live in the `quantum` dialect.

**Description of the Change:** Reverts #2369 and #2431.

**Benefits:** Remove unused code.

**Possible Drawbacks:** None.

[sc-113073]


multiphaseCFD (approved):
Thanks @joeycarter !

PR #2582, opened 2026-03-12
Cast result of measure_p back to int during plxpr interpreting

### LINKED ISSUE #2579: `qp.measure` dtype mismatch in `qp.cond` with capture enabled
Comparing a mid-circuit measurement result with an integer literal (`m == 1`) inside a `qp.cond` branch **with capture enabled** raises a `TypeError` due to a JAX `lax.eq` dtype mismatch between bool and int64. Users must use a float literal (`m == 1.0`) as a workaround, which is unintuitive.

Example:

``` python
import pennylane as qp

@qp.qjit(capture=True)
@qp.qnode(qp.device("null.qubit", wires=2))
def circuit():
    m = qp.measure(0)
    qp.cond(m == 1, qp.CNOT)(wires=[0, 1])
    return qp.state()

circuit()
```

```
lib/python3.12/site-packages/pennylane/capture/base_interpreter.py", line 382, in eval
    outvals = primitive.bind(*subfuns, *invals, **params)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: lax.eq requires arguments to have the same dtypes, got bool, int64. (Tip: jnp.equal is a similar function that does automatic type promotion on inputs).
```

Versions:

``` console
Name: pennylane
Version: 0.45.0.dev52
JAX version:             0.7.1
Catalyst version:        0.15.0.dev47
```

paul0403:
**Context:**
The plxpr measure primitive returns an int type, but the catalyst meaure_p primitive returns a bool type.

When interpreting the plxpr (to convert to catalyst jaxpr), the result of `measure_p.bind()` is therefore a bool type, but the plxpr being interpreted would expect the replacement equation for the plxpr measure equation to also return an int type!

This was never exposed by our tests, and we encountered it when we tried to use a mcm result as a conditional predicate. When users explicitly request
```python
m = qml.measure(0)
if m == 0:
   ...
```

The arithmetic comparisons from the input plxpr are all on int types. The plxpr for the above scenario looks like 
```
          h:i64[] = measure[postselect=None reset=False] 0:i64[]
          i:bool[] = eq h 0:i64[]
          cond[
            # ... cond body ...
          ] i
```  

The `eq` equation is a comparison between int types, as that's what the plxpr measure primitive returns. The binded catalyst measure_p returns a bool, which is not accepted by the `eq` equation (which stays unchanged and still expects an int: the plxpr conversion interpreter doesn't have any interpretation rules for classical primitives), leading to an error like 

```
TypeError: lax.eq requires arguments to have the same dtypes, got bool, int64. (Tip: jnp.equal is a similar function that does automatic type promotion on inputs).
```

When interpreting plxpr's measurement primitive, after binding the catalyst measure primitive (which returns bool), cast the result to int type.

**Benefits:**
Types match: during interpretation, the new equation (catalyst measure) and the replaced equation (plxpr measure) must return the same types. 

**Related GitHub Issues:**
closes #2579 
[sc-113905], [sc-112370]

### DIFF
--- a/frontend/catalyst/from_plxpr/qfunc_interpreter.py
+++ b/frontend/catalyst/from_plxpr/qfunc_interpreter.py
@@ -710,7 +710,11 @@ def handle_measure(self, wire, reset, postselect):
             ]
         )
         out_wire = cond_p.bind(
-            result, out_wire, out_wire, branch_jaxprs=correction, num_implicit_outputs=None
+            jnp.bool(result),
+            out_wire,
+            out_wire,
+            branch_jaxprs=correction,
+            num_implicit_outputs=None,
         )[0]
 
     in_qreg[in_qreg.global_index_to_local_index(wire)] = out_wire
--- a/frontend/catalyst/from_plxpr/qfunc_interpreter.py
+++ b/frontend/catalyst/from_plxpr/qfunc_interpreter.py
@@ -699,6 +699,7 @@ def handle_measure(self, wire, reset, postselect):
         _[0] for _ in get_in_qubit_values([wire], self.qubit_index_recorder, self.init_qreg)
     )
     result, out_wire = measure_p.bind(in_wire, postselect=postselect)
+    result = jnp.astype(result, int)
 
     if reset:
         # Constants need to be passed as input values for some reason I forgot about.
--- a/frontend/catalyst/from_plxpr/qfunc_interpreter.py
+++ b/frontend/catalyst/from_plxpr/qfunc_interpreter.py
@@ -699,7 +699,6 @@ def handle_measure(self, wire, reset, postselect):
         _[0] for _ in get_in_qubit_values([wire], self.qubit_index_recorder, self.init_qreg)
     )
     result, out_wire = measure_p.bind(in_wire, postselect=postselect)
-    result = jnp.astype(result, int)
 
     if reset:
         # Constants need to be passed as input values for some reason I forgot about.
@@ -710,7 +709,7 @@ def handle_measure(self, wire, reset, postselect):
             ]
         )
         out_wire = cond_p.bind(
-            jnp.bool(result),
+            result,
             out_wire,
             out_wire,
             branch_jaxprs=correction,
@@ -718,6 +717,7 @@ def handle_measure(self, wire, reset, postselect):
         )[0]
 
     in_qreg[in_qreg.global_index_to_local_index(wire)] = out_wire
+    result = jnp.astype(result, int)
     return result
 
 
--- a/frontend/catalyst/from_plxpr/qfunc_interpreter.py
+++ b/frontend/catalyst/from_plxpr/qfunc_interpreter.py
@@ -653,6 +653,7 @@ def handle_pauli_measure(self, *invals, pauli_word, **params):
     result, *out_qubits = outvals  # First element is the measurement result
     for in_qreg, w, new_wire in zip(in_qregs, invals, out_qubits):
         in_qreg[in_qreg.global_index_to_local_index(w)] = new_wire
+    result = jnp.astype(result, int)
     return result
 
 
--- a/frontend/catalyst/jax_primitives.py
+++ b/frontend/catalyst/jax_primitives.py
@@ -1835,7 +1835,14 @@ def _mcmobs_abstract_eval(*mcms):
 def _mcm_obs_lowering(jax_ctx: mlir.LoweringRuleContext, *mcms: list[ir.Value]):
     ctx = jax_ctx.module_context.context
     result_type = ir.OpaqueType.get("quantum", "obs", ctx)
-    extracted = [extract_scalar(mcm, "mcmobs") for mcm in mcms]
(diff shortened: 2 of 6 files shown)


JerryChen97 (approved):

frontend/catalyst/from_plxpr/qfunc_interpreter.py
```diff
@@ -699,6 +699,7 @@ def handle_measure(self, wire, reset, postselect):
         _[0] for _ in get_in_qubit_values([wire], self.qubit_index_recorder, self.init_qreg)
     )
     result, out_wire = measure_p.bind(in_wire, postselect=postselect)
+    result = jnp.astype(result, int)
```
maliasadi: We may need the same `jax.astype` cast for `handle_pauli_measure` above.

frontend/catalyst/jax_primitives.py
maliasadi, line 1: So this fix provides a round-trip conversion from bool -> int -> bool; while seems to be a good solution with minimal changes to the prim, the long-term fix should come from PL returning `bool` from the PLxPR measure primitive. What's the plan to fix this in the PLxPR pipeline?

paul0403: > the long-term fix should come from PL returning `bool` from the PLxPR measure primitive

Well, that's why I made a poll :laughing: seems like this isn't the consensus...

frontend/catalyst/from_plxpr/qfunc_interpreter.py
paul0403: Ah true, thanks for the reminder!

frontend/catalyst/jax_primitives.py
dime10: Int is generally considered safer for users in the frontend, because they may perform integer math on measurement results without considering the data type of the result.

frontend/catalyst/from_plxpr/qfunc_interpreter.py
```diff
@@ -709,7 +710,11 @@ def handle_measure(self, wire, reset, postselect):
             ]
         )
         out_wire = cond_p.bind(
-            result, out_wire, out_wire, branch_jaxprs=correction, num_implicit_outputs=None
+            jnp.bool(result),
```
dime10: If you move the astype(result, int) to below the reset, you don't have to have this extra bool cast here :)

frontend/catalyst/jax_primitives.py
```diff
@@ -1835,7 +1835,14 @@ def _mcmobs_abstract_eval(*mcms):
 def _mcm_obs_lowering(jax_ctx: mlir.LoweringRuleContext, *mcms: list[ir.Value]):
     ctx = jax_ctx.module_context.context
     result_type = ir.OpaqueType.get("quantum", "obs", ctx)
-    extracted = [extract_scalar(mcm, "mcmobs") for mcm in mcms]
+    i1type = ir.IntegerType.get_signless(1)
+
+    extracted = []
+    for mcm in mcms:
+        shape = mcm.type.shape
```
dime10: Can mcms have different shapes? :o

dime10 (approved):

frontend/catalyst/from_plxpr/qfunc_interpreter.py
paul0403: good point! https://github.com/PennyLaneAI/catalyst/pull/2582/commits/a4461c09dde18f3e10c109e54f6dec3bb6ba6c22

frontend/catalyst/jax_primitives.py
paul0403: True, no need to get shape each time :sweat-smile: https://github.com/PennyLaneAI/catalyst/pull/2582/commits/9959fe9dd955f97f640773d567ab513738149e42

frontend/catalyst/from_plxpr/qfunc_interpreter.py
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2582/commits/69466477723a7b180811c158ffe70a943052ed7e 

And I checked, the PL ppm primitive indeed also returns an int https://github.com/PennyLaneAI/pennylane/blob/c1604a4bb538d023e44144ae33d7aa136805d3a8/pennylane/ops/mid_measure/pauli_measure.py#L134

frontend/catalyst/jax_primitives.py
maliasadi: I see the point of returning integers for users, but I wonder if it makes sense to do the int -> bool cast upstream in PL when generating PLxPR?

maliasadi (approved):

frontend/catalyst/jax_primitives.py
paul0403: Wouldn't any user integer math code require the plxpr to have int type on measurement results?

I think this is also why we need the `handle_measure` to return int type, since the surrounding uses of it will expect an int. It is the `i1` users in mlir (e.g. control flow predicates, and mcm obs op input) that are the exception actually.

PR #2588, opened 2026-03-16
Remove the local stablehlo dialect from the python_interface module

mehrdad2m:
**Context:**
As the migration of the StableHLO dialect to `xdsl-jax` repo is now complete, we can remove the local implementation and point to the upstream version in the `xdsl-jax `repo. 

-  Remove the `stablehlo` folder from the `python_interface` module
- Fix the imports to get the dialect from upstream

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-101832]


mudit2812 (approved):
Looks great. This is blocked until the new version of xdsl-jax is out, right?

paul0403 (approved):

PR #2589, opened 2026-03-17
Add convert-quantum-to-qecl dialect-conversion pass

joeycarter:
**Context:** This PR is the first in a series to add a prototype compilation pipeline that lowers a PennyLane program expressed in Catalyst's `quantum` MLIR dialect to an equivalent program expressed using quantum error-correction protocols. This PR adds the first stage in this pipeline to lower a program in the `quantum` dialect into the QEC Logical (`qecl`) dialect.

**Description of the Change:** This PR adds the `convert-quantum-to-qecl` compiler pass, which performs a dialect conversion from `quantum` to `qecl`. The pass is implemented in xDSL.

**Example:**

Before:

```mlir
func.func public @circuit() -> tensor<1x1xi64> attributes {quantum.node} {
  %c1_i64 = arith.constant 1 : i64
  quantum.device shots(%c1_i64) ["", "", ""]
  %out_qubits = quantum.custom "Hadamard"() %1 : !quantum.bit
  %mres, %out_qubit = quantum.measure %out_qubits : i1, !quantum.bit
  %2 = quantum.mcmobs %mres : !quantum.obs
  %3 = quantum.sample %2 : tensor<1x1xf64>
  %4 = stablehlo.convert %3 : (tensor<1x1xf64>) -> tensor<1x1xi64>
  %5 = quantum.insert %0[ 0], %out_qubit : !quantum.reg, !quantum.bit
  quantum.device_release
  return %4 : tensor<1x1xi64>
}
```

After applying convert-quantum-to-qecl (with k = 1):

```mlir
func.func public @circuit() -> tensor<1x1xi64> attributes {quantum.node} {
  %c1_i64 = arith.constant 1 : i64
  quantum.device shots(%c1_i64) ["", "", ""]
  %1 = qecl.extract_block %0[0] : !qecl.hyperreg<1 x 1> -> !qecl.codeblock<1>
  %2 = qecl.encode[zero] %1 : !qecl.codeblock<1>
  %3 = qecl.insert_block %0[0], %2 : !qecl.hyperreg<1 x 1>, !qecl.codeblock<1>
  %4 = qecl.extract_block %3[0] : !qecl.hyperreg<1 x 1> -> !qecl.codeblock<1>
  %out_qubits = qecl.qec %4 : !qecl.codeblock<1>
  %out_qubits_1 = qecl.hadamard %out_qubits[0] : !qecl.codeblock<1>
  %out_qubits_2 = qecl.qec %out_qubits_1 : !qecl.codeblock<1>
  %mres, %5 = qecl.measure %out_qubits_2[0] : i1, !qecl.codeblock<1>
  %6 = quantum.mcmobs %mres : !quantum.obs
  %7 = quantum.sample %6 : tensor<1x1xf64>
  %8 = stablehlo.convert %7 : (tensor<1x1xf64>) -> tensor<1x1xi64>
  %9 = qecl.insert_block %3[0], %5 : !qecl.hyperreg<1 x 1>, !qecl.codeblock<1>
  qecl.dealloc %9 : !qecl.hyperreg<1 x 1>
  quantum.device_release
  func.return %8 : tensor<1x1xi64>
}
```

**Current limitations:**

* The `quantum` program must already be expressed using only Clifford gates (H, S, S†, and CNOT). This will eventually be expanded to support Clifford+T ops.
* The conversion pass does yet not support control-flow operations (`scf.for`, `scf.if`, etc.)

[sc-111763]


frontend/catalyst/python_interface/dialects/qecl.py
```diff
@@ -148,9 +148,6 @@ def parse_parameters(cls, parser: AttrParser) -> Sequence[Attribute]:
 LogicalCodeBlockSSAValue: TypeAlias = SSAValue[LogicalCodeblockType]
 LogicalHyperRegisterSSAValue: TypeAlias = SSAValue[LogicalHyperRegisterType]
 
-anyLogicalCodeblock = ContainerOf(LogicalCodeblockType)
-anyLogicalHyperRegister = ContainerOf(LogicalHyperRegisterType)
```
joeycarter, line 152: I'm sneaking this change into this PR since `anyLogicalCodeblock = ContainerOf(LogicalCodeblockType)` was incorrect given where it's being used; it should have been `base(LogicalCodeblockType)` (and similarly for `LogicalHyperRegisterType`).

frontend/catalyst/python_interface/dialects/qecl.py
```diff
@@ -652,9 +649,9 @@ def __init__(
         idx_trgt: int | IntegerAttr | SSAValue[IndexType] | Operation,
     ):
         if isinstance(idx_ctrl, int):
-            idx_ctrl = IntegerAttr(idx_ctrl, 64)
+            idx_ctrl = IntegerAttr(idx_ctrl, IndexType())
```
joeycarter, line 652: Another small mistake I'm fixing in this PR.

frontend/catalyst/python_interface/dialects/qecl.py
multiphaseCFD: not familiar with both? could you clarify why the previous implementation does not meet our need?

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
```diff
@@ -0,0 +1,589 @@
...
+    k: int
+
+    def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
+        """Apply the convert-quantum-to-qecl pass."""
+        if self.k != 1:
+            raise NotImplementedError(
+                f"The {self.name} only supports QEC codes where the number of logical qubits per "
+                f"codeblock, k, is 1, but got k = {self.k}"
+            )
+
```
multiphaseCFD, line 601: Wondering if it makes sense to add a safeguard to ensure this pass only process a circuit with supported operations?

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
```diff
@@ -0,0 +1,589 @@
...
+                ub_op,
+                step_op,
+                for_each_codeblock_op,
+            )
+
+        assert ops_to_insert, "Sequence of ops to insert is empty"
+        return ops_to_insert
+
+
+# MARK: Extract Op Pattern
```
lillian542, line 231: What do these do? IS `# MARK` something specific the way `# CHECK` is?

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
```diff
@@ -0,0 +1,589 @@
...
+        idx = _get_idx_value_or_attr_from_extract_or_insert_op(op, rewriter)
+
+        ops_to_insert = (
+            # Convert type quantum.reg -> qecl.hyperreg
+            # (to be resolved by ReconcileUnrealizedCastsPass)
+            conv_cast_op := builtin.UnrealizedConversionCastOp.get(
+                (qreg_owner_op.results[0],), (qreg_owner_op.operands[0].type,)
+            ),
+            extract_codeblock_op := qecl.ExtractCodeblockOp(
+                hyper_reg=conv_cast_op.results[0], idx=idx
```
multiphaseCFD, line 270: I guess the implementation here assumes `k=1`. If `k=2`, I guess `idx` here would be assigned with something else.

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
```diff
@@ -0,0 +1,589 @@
...
+            ),
+            qec_cycle_op := qecl.QecCycleOp(in_codeblock=extract_codeblock_op.codeblock),
+            _cast_to_qubit(qec_cycle_op.out_codeblock),
+        )
+
+        rewriter.replace_op(op, ops_to_insert)
```
multiphaseCFD, line 276: Is the `replace_op` sufficient? Do we need to replace all uses of `op.results[0]` with the `cast_op` here?

multiphaseCFD (commented):
Thanks Joey! Nice work, a quick first round of review.

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
```diff
@@ -0,0 +1,589 @@
...
+  %6 = quantum.mcmobs %mres : !quantum.obs
+  %7 = quantum.sample %6 : tensor<1x1xf64>
+  %8 = stablehlo.convert %7 : (tensor<1x1xf64>) -> tensor<1x1xi64>
+  %9 = qecl.insert_block %3[0], %5 : !qecl.hyperreg<1 x 1>, !qecl.codeblock<1>
+  qecl.dealloc %9 : !qecl.hyperreg<1 x 1>
+  quantum.device_release
+  func.return %8 : tensor<1x1xi64>
+}
+```
+"""
```
lillian542, line 78: Can we add a "known limitations" section in here as well, like we did with the other recent passes? I've found it very helpful when returning to things after a few iterations and needing to remember where we left off.

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
```diff
@@ -0,0 +1,589 @@
...
+                        ),
+                        ctrl_conv_cast_op := _cast_to_qubit(ctrl_qecl_cycle_op.out_codeblock),
+                        trgt_conv_cast_op := _cast_to_qubit(trgt_qecl_cycle_op.out_codeblock),
+                    )
+                    new_results = (ctrl_conv_cast_op.results[0], trgt_conv_cast_op.results[0])
+
+            case _:
+                raise NotImplementedError(
+                    f"Conversion of op '{op.name}' only supports gates 'Hadamard', 'S' and 'CNOT', "
+                    f"but got {gate_name}"
```
lillian542: Since they are in the dialect, can we also include X, Y, Z and Identity?

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
```diff
@@ -0,0 +1,589 @@
...
+        """Rewrite pattern for `quantum.dealloc` ops."""
+
+        qreg_owner_op = op.qreg.owner
+        if not _is_type_convertible(qreg_owner_op, qecl.LogicalHyperRegisterType):
+            _raise_failed_to_convert_op_compile_error(op)
+
+        ops_to_insert = (
+            conv_cast_op := builtin.UnrealizedConversionCastOp.get(
+                (qreg_owner_op.results[0],), (qreg_owner_op.operands[0].type,)
+            ),
```
lillian542, line 337: Why are `_cast_to_qubit` and `_cast_to_qureg` their own helper functions, while `_cast_to_codeblock` and `_cast_to_hyperreg` are not?

lillian542 (commented):
Thanks @joeycarter! Seems easy to follow and made it a lot clearer to me how lowering to a new dialect works in practice. I left a few comments, but nothing major :)

frontend/catalyst/python_interface/dialects/qecl.py
joeycarter: When used as a `VarConstraint`, `ContainerOf[X]` expands to

```python
VarConstraint[X | VectorType[X] | TensorType[X]]
```

which is not what we want here (we only want a single codeblock or hyper-register).

Using `VarConstraint("T", base(X))` rather than `VarConstraint("T", X)` is less important but really handy when using a Python type checker (like the xDSL repo is doing).

Writing

```python
T: ClassVar = VarConstraint("T", LogicalCodeblockType)
```

means that `T` has the generic type `VarConstraint[Attribute]`, while using `base()`:

```python
T: ClassVar = VarConstraint("T", base(LogicalCodeblockType))
```

means that `T` has the *specific* type `VarConstraint[LogicalCodeblockType]`. The `base()` function is essentially a fancy [`typing.cast`](https://docs.python.org/3/library/typing.html#typing.cast) wrapper.

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
joeycarter: I'm not sure that's necessary here. MLIR has a concept of "legal" and "illegal" operations when performing a dialect conversion (see [here](https://mlir.llvm.org/docs/DialectConversion/#modes-of-conversion)), but I don't believe xDSL has an equivalent concept. Also, if this pass hits an op it doesn't support, that will likely lead to dangling `unrealized_conversion_cast` ops that `ReconcileUnrealizedCastsPass` won't be able to resolve. That will raise a warning during compilation, so these ops won't go unnoticed.

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
joeycarter: That's correct. I've added a note explaining this assumption in 0cd0395b5

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
joeycarter: Yes it's sufficient; `rewriter.replace_op()` calls `replace_all_uses_with()` to replace the old result with the last result in the list of new ops.

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
joeycarter: Ah, this is a little vscode trick that makes the comments larger in the minimap, I can remove them if they're distracting :sweat_smile:

More info here: https://www.stefanjudis.com/blog/vs-code-minimap-section-headers/

lillian542: Oh, cool! I'm fine with leaving them there :)

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
joeycarter: Good point, done: bf3c9eca2

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
joeycarter: Yes, thank you! This was an oversight on my part. Added here: dd9b10769

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
lillian542: This is more curiosity about your thought process than anything else, if there's not a quick and simple answer feel free to ignore this one 😅

joeycarter: The main reason was because we can cast directly to `!quantum.reg` and `!quantum.bit` since they aren't parameterized, so the helper function can take in a single argument. Whereas, casting to hyper-registers and codeblocks requires either passing in the individual parameters to construct the type or a reference to that type from another SSA value, hence any helper function would just be a direct wrapper around `builtin.UnrealizedConversionCastOp.get()` so it felt unnecessary :slightly_smiling_face:

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
multiphaseCFD: Wasn't aware of that before. Thanks

lillian542 (approved):
LGTM, thanks Joey!

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
multiphaseCFD: Thanks for the clarification and the reference. Sounds good to me

multiphaseCFD (approved):
🥇

PR #2590, opened 2026-03-17
Allow quantum adjoint op to take in arbitrary arguments

paul0403:
**Context:**
Currently, the `quantum.adjoint` operation can only take in a single register value.
We generalize it to take in an arbitrary list of register and qubit values. 

This is to prepare for the migration of the frontend to reference semantics, when we remove the constraint that scopes must take in the global qreg. 

`quantum.adjoint` operation can take in `Variadic<AnyTypeOf<[QubitType, QuregType]>>` instead of just a single `QuregType`.

`qref.adjoint` operation takes in no operands anymore, as it only needs to capture all references by clousure.

Some logic in adjoint lowering pass is updated correspondingly.

I also noticed a small bug in the adjoint lowering pass, where classical Values from within the adjoint region are not replaced by their clone outside the region when creating the new call to the adjointed func op. I fixed this too.

**Benefits:**
Prepare for future reference semantics work on them.

[sc-111989]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -232,7 +232,7 @@
 
 <h3>Improvements 🛠</h3>
 
-* The `quantum.adjoint` and `qref.adjoint` operations can now take in multiple quantum values, allowing
+* The `quantum.adjoint` operation can now take in multiple quantum values, allowing
   both qubits and registers, as opposed to constraining the operand to be a single quantum register.
   [(#2590)](https://github.com/PennyLaneAI/catalyst/pull/2590)
 
@@ -448,6 +448,7 @@
 * A new dialect `QRef` was created. This dialect is very similar to the existing `Quantum` dialect,
   but it is in reference semantics, whereas the existing `Quantum` dialect is in value semantics.
   [(#2320)](https://github.com/PennyLaneAI/catalyst/pull/2320)
+  [(#2590)](https://github.com/PennyLaneAI/catalyst/pull/2590)
 
   Unlike qubit (or qreg) SSA values in the `Quantum` dialect, a qubit (or qreg) reference SSA value
   in the `QRef` dialect is allowed to be used multiple times. The operands of gates and observables
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -232,6 +232,10 @@
 
 <h3>Improvements 🛠</h3>
 
+* The `quantum.adjoint` and `qref.adjoint` operations can now take in multiple quantum values, allowing
+  both qubits and registers, as opposed to constraining the operand to be a single quantum register.
+  [(#2590)](https://github.com/PennyLaneAI/catalyst/pull/2590)
+
 * Catalyst with program capture now supports device preprocessing. Currently, preprocessing transforms
   that do not have native MLIR or xDSL implementations will be replaced with empty transforms.
   [(#2557)](https://github.com/PennyLaneAI/catalyst/pull/2557)
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -341,6 +341,9 @@
 
 <h3>Bug fixes 🐛</h3>
 
+* Quantum adjoint can now be used on subroutines with classical arguments.
+  [(#2590)](https://github.com/PennyLaneAI/catalyst/pull/2590)
+
 * Fixed a bug where input array arguments could be mutated during execution when copied inputs
   were updated in-place. Entry-point arguments are now treated as non-writable during bufferization,
   preserving the expected immutability of user inputs.
--- a/frontend/catalyst/jax_primitives.py
+++ b/frontend/catalyst/jax_primitives.py
@@ -2765,7 +2765,7 @@ def _adjoint_lowering(
     ), f"Expected a single result of quantum.register type, got: {output_types}"
 
     # Build an adjoint operation with a single-block region.
-    op = AdjointOp(output_types[0], qargs[0])
+    op = AdjointOp(output_types, qargs)
     adjoint_block = op.regions[0].blocks.append(*[mlir.aval_to_ir_types(a)[0] for a in aqargs])
     with ir.InsertionPoint(adjoint_block):
         source_info_util.extend_name_stack("adjoint")
--- a/frontend/catalyst/python_interface/dialects/quantum/operations/miscellaneous_ops.py
+++ b/frontend/catalyst/python_interface/dialects/quantum/operations/miscellaneous_ops.py
@@ -28,7 +28,6 @@ from xdsl.irdl import (
     ParsePropInAttrDict,
     irdl_op_definition,
     lazy_traits_def,
-    operand_def,
     opt_operand_def,
     opt_prop_def,
     prop_def,
--- a/frontend/catalyst/python_interface/dialects/quantum/operations/miscellaneous_ops.py
+++ b/frontend/catalyst/python_interface/dialects/quantum/operations/miscellaneous_ops.py
@@ -16,6 +16,7 @@ This file contains the definition of miscellaneous operations in the
 Quantum dialect.
 """
 from collections.abc import Sequence
+from typing import ClassVar
 
 from xdsl.dialects.builtin import (
     StringAttr,
@@ -24,8 +25,10 @@ from xdsl.dialects.builtin import (
 )
 from xdsl.ir import Block, Operation, Region
 from xdsl.irdl import (
+    AnyOf,
     IRDLOperation,
     ParsePropInAttrDict,
+    VarConstraint,
     irdl_op_definition,
     lazy_traits_def,
     opt_operand_def,
@@ -53,15 +56,17 @@ from ..attributes import QubitSSAValue, QubitType, QuregSSAValue, QuregType
 class AdjointOp(IRDLOperation):
     """Calculate the adjoint of the enclosed operations"""
 
+    T: ClassVar = VarConstraint("T", AnyOf([QubitType, QuregType]))
+
     name = "quantum.adjoint"
 
     assembly_format = """
-        `(` $args `)` attr-dict `:` type($args) `->` type($outs) $region
+        `(` $args `)` attr-dict `:` type($outs) $region
     """
 
-    args = var_operand_def(QuregType | QubitType)
+    args = var_operand_def(T)
 
-    outs = var_result_def(QuregType | QubitType)
+    outs = var_result_def(T)
 
     region = region_def("single_block")
 
@@ -74,19 +79,6 @@ class AdjointOp(IRDLOperation):
     ):
         result_types = tuple(arg.type for arg in args)
         super().__init__(operands=(args,), result_types=(result_types,), regions=(region,))
-
-
-@irdl_op_definition
-class YieldOp(IRDLOperation):
-    """Return results from quantum program regions"""
-
-    name = "quantum.yield"
-
-    assembly_format = "attr-dict ($retvals^ `:` type($retvals))?"
-
-    retvals = var_operand_def(QuregType | QubitType)
-
-    traits = traits_def(HasParent(AdjointOp), IsTerminator(), Pure(), ReturnLike())
 
 
 @irdl_op_definition
@@ -150,3 +142,16 @@ class NumQubitsOp(IRDLOperation):
     """
 
     num_qubits = result_def(i64)
+
+
+@irdl_op_definition
+class YieldOp(IRDLOperation):
+    """Return results from quantum program regions"""
+
+    name = "quantum.yield"
+
+    assembly_format = "attr-dict ($retvals^ `:` type($retvals))?"
+
+    retvals = var_operand_def(QuregType | QubitType)
+
+    traits = traits_def(HasParent(AdjointOp), IsTerminator(), Pure(), ReturnLike())
--- a/frontend/catalyst/python_interface/dialects/quantum/operations/miscellaneous_ops.py
+++ b/frontend/catalyst/python_interface/dialects/quantum/operations/miscellaneous_ops.py
@@ -36,6 +36,7 @@ from xdsl.irdl import (
     result_def,
     traits_def,
     var_operand_def,
+    var_result_def,
 )
 from xdsl.traits import (
     HasParent,
@@ -46,7 +47,7 @@ from xdsl.traits import (
     SingleBlockImplicitTerminator,
 )
 
-from ..attributes import QuregSSAValue, QuregType
+from ..attributes import QubitSSAValue, QubitType, QuregSSAValue, QuregType
 
 
 @irdl_op_definition
@@ -56,12 +57,12 @@ class AdjointOp(IRDLOperation):
     name = "quantum.adjoint"
 
     assembly_format = """
-        `(` $qreg `)` attr-dict `:` type(operands) $region
+        `(` $args `)` attr-dict `:` type($args) `->` type($outs) $region
     """
 
-    qreg = operand_def(QuregType)
+    args = var_operand_def(QuregType | QubitType)
 
-    out_qreg = result_def(QuregType)
+    outs = var_result_def(QuregType | QubitType)
 
     region = region_def("single_block")
 
@@ -69,10 +70,24 @@ class AdjointOp(IRDLOperation):
 
     def __init__(
         self,
-        qreg: QuregSSAValue | Operation,
+        args: Sequence[QuregSSAValue | QubitSSAValue] | Operation,
         region: Region | Sequence[Operation] | Sequence[Block],
     ):
-        super().__init__(operands=(qreg,), result_types=(QuregType(),), regions=(region,))
+        result_types = tuple(arg.type for arg in args)
+        super().__init__(operands=(args,), result_types=(result_types,), regions=(region,))
+
+
+@irdl_op_definition
+class YieldOp(IRDLOperation):
+    """Return results from quantum program regions"""
+
+    name = "quantum.yield"
+
+    assembly_format = "attr-dict ($retvals^ `:` type($retvals))?"
+
+    retvals = var_operand_def(QuregType | QubitType)
+
+    traits = traits_def(HasParent(AdjointOp), IsTerminator(), Pure(), ReturnLike())
 
 
 @irdl_op_definition
@@ -136,16 +151,3 @@ class NumQubitsOp(IRDLOperation):
     """
 
     num_qubits = result_def(i64)
-
-
-@irdl_op_definition
-class YieldOp(IRDLOperation):
-    """Return results from quantum program regions"""
-
-    name = "quantum.yield"
-
-    assembly_format = "attr-dict ($retvals^ `:` type($retvals))?"
-
-    retvals = var_operand_def(QuregType)
-
-    traits = traits_def(HasParent(AdjointOp), IsTerminator(), Pure(), ReturnLike())
--- a/frontend/test/pytest/test_adjoint.py
+++ b/frontend/test/pytest/test_adjoint.py
@@ -590,6 +590,25 @@ class TestCatalyst:
         observed = qjit(circuit)()
         assert_allclose(expected, observed)
 
+    @pytest.mark.usefixtures("use_both_frontend")
+    def test_adjoint_subroutine_with_classical_args(self, backend):
+        """Test an adjoint on a subroutine, with classical arguments"""
+
+        @qml.templates.Subroutine
+        def f(x, wires):
+            qml.IsingXX(x, wires)
+
+        dev = qml.device(backend, wires=4)
+
+        @qml.qnode(dev)
+        def circuit():
+            qml.adjoint(f)(0.5, (0, 1))
+            return qml.probs(wires=0)
+
+        expected = circuit()
+        observed = qjit(circuit)()
+        assert_allclose(expected, observed)
+
     def test_adjoint_outside_qjit(self, backend):
         """Test that the hybrid adjoint can be used from outside qjit & qnode."""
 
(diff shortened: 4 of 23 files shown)


mlir/lib/Quantum/Transforms/AdjointPatterns.cpp
```diff
@@ -374,21 +384,40 @@ class AdjointGenerator {
...
+        }
+
+        size_t current_reversed_result = 0;
+        for (size_t i = 0; i < args.size(); i++) {
+            if (!isa<QuregType, QubitType>(args[i].getType())) {
+                continue;
+            }
+            args[i] = reversedResults[current_reversed_result];
+            current_reversed_result++;
+        }
```
paul0403: Use std::queue

mlir/lib/Quantum/Transforms/AdjointPatterns.cpp
```diff
@@ -374,21 +384,40 @@ class AdjointGenerator {
...
-        // Map the initial quantum register with the adjoint result
-        remappedValues.map(getQuantumReg(initQreg).value(), adjointCallOp.getResult(0));
+        size_t i = 0;
+        for (auto callOperand : callOp.getArgOperands()) {
+            if (!isa<QuregType, QubitType>(callOperand.getType())) {
+                continue;
+            }
+            remappedValues.map(callOperand, adjointCallOp.getResult(i));
+            i++;
+        }
```
paul0403: Use std::queue

frontend/catalyst/jax_primitives.py
```diff
@@ -2765,7 +2765,7 @@ def _adjoint_lowering(
     ), f"Expected a single result of quantum.register type, got: {output_types}"
 
     # Build an adjoint operation with a single-block region.
-    op = AdjointOp(output_types[0], qargs[0])
```
dime10, line 2797: Don't we need to update the assertions above accordingly?

mlir/include/QRef/IR/QRefInterfaces.td
```diff
@@ -17,34 +17,6 @@
 
 include "mlir/IR/OpBase.td"
 
-def QuantumRegion : OpInterface<"QuantumRegion"> {
```
dime10, line 20: We could update this interface to represent any region op with strictly equivalent quantum input/outputs, but only if it's useful. Maybe we tackle this once we actually have multiple such operations though (e.g. the control region), so fine to remove for now.

mlir/include/Quantum/IR/QuantumOps.td
```diff
@@ -742,21 +742,21 @@ def QubitUnitaryOp : UnitaryGate_Op<"unitary", [ParametrizedGate, NoMemoryEffect
 class Region_Op<string mnemonic, list<Trait> traits = []> :
         Quantum_Op<mnemonic, traits # [NoMemoryEffect]>;
 
-def AdjointOp : Region_Op<"adjoint", [QuantumRegion, SingleBlockImplicitTerminator<"YieldOp">]> {
+def AdjointOp : Region_Op<"adjoint", [SingleBlockImplicitTerminator<"YieldOp">]> {
```
dime10: Instead of using custom verifier logic, it would be nice to enforce the relationship between inputs and outputs declaratively. I know you said you've tried it and it changed some of the python bindings' signatures, but I don't think that should deter us, especially since Joey was already using `AllTypesMatch<["in_qubits", "out_qubits"]>` in [his PR](https://github.com/PennyLaneAI/catalyst/pull/2369) 🤔

mlir/lib/QRef/IR/QRefOps.cpp
```diff
@@ -210,6 +210,19 @@ LogicalResult AdjointOp::verify()
...
+        return emitOpError("Adjoint op number of operands must be the same as the number of "
+                           "arguments on its block");
+    }
+
+    for (auto [operand, bbArg] : llvm::zip_equal(this->getArgs(), b.getArguments())) {
+        if (operand.getType() != bbArg.getType()) {
+            return emitOpError(
+                "Adjoint op operand types must be the same as the argument types on its block");
+        }
+    }
```
dime10: Huh I wouldn't have thought to check this, but I guess it's not guaranteed? 😅  (in func it would be with the special syntax)

frontend/test/pytest/test_adjoint.py
```diff
@@ -590,6 +590,25 @@ def circuit():
         observed = qjit(circuit)()
         assert_allclose(expected, observed)
 
+    @pytest.mark.usefixtures("use_both_frontend")
```
paul0403, line 593: @albi3ro is this still the best way to say "this pytest runs on both frontends"? Or should I parametrize on `qjit(capture=...)`?

And what's the scope of @JerryChen97 's test suite maintainence work? Will this sort of thing be deal with?

mlir/lib/Quantum/Transforms/AdjointPatterns.cpp
```diff
@@ -12,10 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <mlir/IR/Block.h>
+#include <mlir/IR/Value.h>
```
dime10: Wrong location for these headers 😅

mlir/lib/Quantum/Utils/QuantumSplitting.cpp
```diff
@@ -224,19 +224,14 @@ void AugmentedCircuitGenerator::generate(Region &region, OpBuilder &builder)
...
+                return isa<QuregType, QubitType>(value);
+            });
 
             // Classical call operations are cloned for the backward pass
             if (!quantum) {
                 builder.clone(op, oldToCloned);
             }
-
-            assert(!(quantum && multiReturns) && "Adjoint does not support functions with multiple "
-                                                 "returns that contain a quantum register.");
```
dime10, line 239: How is that functionality handled now? If we were asserting this case away before, presumably it's because it wasn't implemented :o

mlir/test/Gradient/AdjointTest.mlir
```diff
@@ -70,7 +74,9 @@ func.func @gradCallScalarTensor(%arg0: f64) -> tensor<2x3xf64> {
 // Check tensor to scalar
 func.func private @funcTensorScalar(%arg0: tensor<3xf64>) -> f64 attributes {qnode, diff_method = "adjoint"} {
     %0 = quantum.alloc(1) : !quantum.reg
-    %1 = quantum.adjoint(%0) : !quantum.reg {}  // prevent folding of dealloc into alloc
+    %1 = quantum.adjoint(%0) : !quantum.reg -> !quantum.reg {
+    ^bb0(%arg1: !quantum.reg):
```
dime10, line 78: This is because of the stricter verification now? I suppose we could modify the verifier to only check block args if a block exists, but then I'm also wondering why one would want an empty adjoint op in the first place 😅

mlir/test/Mitigation/ZneFoldingGlobalTest.mlir
```diff
@@ -23,7 +23,7 @@
     // CHECK:      [[qReg:%.+]] = call @simpleCircuit.quantumAlloc([[nQubits]]) : (i64) -> !quantum.reg
     // CHECK:      [[outQregFor:%.+]]  = scf.for %arg2 = [[c0]] to %arg1 step [[c1]] iter_args([[inQreg:%.+]] = [[qReg]]) -> (!quantum.reg) {
         // CHECK:      [[outQreg1:%.+]] = func.call @simpleCircuit.withoutMeasurements(%arg0, [[inQreg]]) : (tensor<3xf64>, !quantum.reg) -> !quantum.reg
-        // CHECK:      [[outQreg2:%.+]] = quantum.adjoint([[outQreg1]]) : !quantum.reg {
+        // CHECK:      [[outQreg2:%.+]] = quantum.adjoint([[outQreg1]]) : !quantum.reg -> !quantum.reg {
```
dime10: We can omit duplicated types because of the guaranteed relationship between inputs and outputs, similar to how we omit duplicates types in quantum gates.

mlir/test/QRef/UnitTests.mlir
```diff
@@ -240,6 +240,19 @@ func.func @test_adjoint_op(%r: !qref.reg<2>)
 
 // -----
 
+func.func @test_adjoint_multiple_args(%r: !qref.reg<2>, %q: !qref.bit)
+{
+    qref.adjoint(%r, %q) : !qref.reg<2>, !qref.bit {
+    ^bb0(%arg0: !qref.reg<2>, %arg1: !qref.bit):
```
dime10: Is it useful for the reference adjoint to carry inputs? Since it's not isolated from above we could omit them, but maybe you found it useful to have them.

dime10 (approved):
Looks great, thanks Paul! No changes needed really, just have some questions and minor comments :)

mlir/lib/Quantum/Utils/QuantumSplitting.cpp
paul0403: The new logic does not distinguish between qregs and qubits at all! I added some tests with multiple qubit values, but I can add one with qreg+qbit

dime10: Isn't this about call ops with quantum args in them?

mlir/test/Gradient/AdjointTest.mlir
paul0403: Yes, this is due to the stuff I added in the verifier.

For 

> I'm also wondering why one would want an empty adjoint op in the first place

in this particular test, this is just a user of the qreg value, so that the alloc and dealloc are not immediately next to each other and get folded away.

mlir/lib/Quantum/Utils/QuantumSplitting.cpp
paul0403: Yeah I'm talking about call ops too!

paul0403: So my new logic can handle subroutines with any combination of classical, qubit and qreg args/results, and the corresponding call from inside the adjoint region.

mlir/test/Quantum/AdjointTest.mlir
```diff
@@ -228,6 +267,58 @@ func.func private @workflow_adjoint(%arg0: f64) -> tensor<4xcomplex<f64>> attrib
...
+  %0 = quantum.alloc( 2) : !quantum.reg
+  %1 = quantum.extract %0[ 0] : !quantum.reg -> !quantum.bit
+  %2 = quantum.extract %0[ 1] : !quantum.reg -> !quantum.bit
+  %3:2 = quantum.adjoint(%1, %2) : !quantum.bit, !quantum.bit -> !quantum.bit, !quantum.bit {
+    ^bb0(%arg1: !quantum.bit, %arg2: !quantum.bit):
+      %4 = quantum.custom "PauliX"() %arg1 : !quantum.bit
+      %5 = quantum.custom "RX"(%arg0) %4 : !quantum.bit
+      %6 = quantum.custom "PauliZ"() %arg2 : !quantum.bit
+      %cst = arith.constant 4.000000e-01 : f64
+      %7:2 = func.call @circuit(%cst, %5, %6): (f64, !quantum.bit, !quantum.bit) -> (!quantum.bit, !quantum.bit)
```
paul0403, line 308: @dime10 here this new test has a call with 2 qubit args/resutls

frontend/test/pytest/test_adjoint.py
JerryChen97: Yep `qjit(capture=...)` is recommended!

JerryChen97: The difference in convention has not caused much trouble so far; therefore we haven't prioritized these changes yet.
Test audit work has been mostly focusing gathering the info and data on what more incompatibilities are out there in the wild.

JerryChen97: But if you do find it bit annoying to see the fixture `use_both_frontend` everywhere I can also create a separate story to do it. This should be quick fix

doc/releases/changelog-dev.md
```diff
@@ -337,6 +341,9 @@
 
 <h3>Bug fixes 🐛</h3>
 
+* Quantum adjoint can now be used on subroutines with classical arguments.
```
mudit2812:
```suggestion
* :func:`~pennylane.adjoint` can now be used on subroutines with classical arguments.
```

frontend/catalyst/python_interface/dialects/quantum/operations/miscellaneous_ops.py
```diff
@@ -56,23 +56,37 @@ class AdjointOp(IRDLOperation):
...
+class YieldOp(IRDLOperation):
+    """Return results from quantum program regions"""
+
+    name = "quantum.yield"
+
+    assembly_format = "attr-dict ($retvals^ `:` type($retvals))?"
+
+    retvals = var_operand_def(QuregType | QubitType)
+
+    traits = traits_def(HasParent(AdjointOp), IsTerminator(), Pure(), ReturnLike())
```
mudit2812: The reason why this was lower in the file was because all the operations are defined alphabetically. I don't mind breaking that "soft" restriction, but also prefer to stay consistent

mudit2812 (commented):
The xDSL changes look great to me @paul0403 . Since there are also C++ changes, I'd rather not be a blocking reviewer, so I'll leave approvals to the compilation team

mlir/lib/Quantum/Transforms/AdjointPatterns.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2590/changes/4d9d1ff3f9045ade01672ebdf768be58469582bb 

vscode auto-added them :sweat_smile:

frontend/test/pytest/test_adjoint.py
paul0403: No it's fine, I don't think this fixture is annoying : ) I was just wondering about the plans for the test audit. Thanks!

paul0403: For now I'll use the `use_both_frontend` fixture to keep it consistent with the rest of the file.

mlir/lib/Quantum/Utils/QuantumSplitting.cpp
dime10: Nice!

mlir/test/Gradient/AdjointTest.mlir
dime10: I feel like the adjoint should fold itself if empty 😅

mlir/lib/QRef/IR/QRefOps.cpp
paul0403: Yes, I was surprised as well, that's why I added this verifier.

mlir/include/Quantum/IR/QuantumOps.td
paul0403: It's not about the `AllTypesMatch` trait, it's about the assembly format. It has to see both the `type($args)` and the `type($results)`, even with the `AllTypesMatch` trait:
```
error: type of result #0, named 'results', is not buildable and a buildable type cannot be inferred
        `(` $args `)` attr-dict `:` type(operands) $region
        ^
/home/paul.wang/catalyst_new/catalyst/mlir/include/Quantum/IR/QuantumOps.td:745:5: note: in custom assembly format for this operation
def AdjointOp : Region_Op<"adjoint", [SingleBlockImplicitTerminator<"YieldOp">, AllTypesMatch<["args", "results"]>]> {
    ^
note: suggest adding a type constraint to the operation or adding a 'type($results)' directive to the custom assembly format
        `(` $args `)` attr-dict `:` type(operands) $region

```

It works for Joey because the types of `$out_qubits` are in custom op's assembly format. 

The only way to not demand the `type($results)` on the assembly format is to add the `SameOperandsAndResultType` trait: https://github.com/llvm/llvm-project/blob/a70f82fab606df7824fa0a8129d0083a86ac40b6/mlir/include/mlir/Interfaces/InferTypeOpInterface.td#L440

But this requires including the `InferTypeOpInterface.td` file. Looking at the file's header:
```
// This file contains a set of interfaces that can be used to define information
// related to type inference.
//
// This interface is also used by ODS to create builders for operations that
// do not require result type to be specified. Including this interface in
// dialect op definitions is sufficient to result in such builders being
// automatically generated for trivially buildable result types.
```

I checked the cpp builders, the default ones are still there. But in the python bindinds the op class signatures got changed: for ops whose outputs are trivially knowable (e.g. alloc op's output is always a single qreg), the python binding no longer takes in the output type.

paul0403: Ok, through the tried-and-true method of trying random things until something works, this following combination works:
```
def AdjointOp : Region_Op<"adjoint", [...AllTypesMatch<["args", "results"]>...]> {
    ...
    let assemblyFormat = [{
        `(` $args `)` attr-dict `:` type($results) $region
    }];
```

I'll update and remove the duplicated types I added to the tests :sweat_smile:

mlir/test/QRef/UnitTests.mlir
paul0403: Oh yes you are right! I was mostly thinking about the conversion and how annoying the `scf.if` op was, when it didn't carry any operands.

But then I realized the annoyance was because in the output value semantic IR the `if` op cannot take operands, not because in the input! For example, for `scf.for` op, I was perfectly happy to make the reference `for` to have no operands and just see things via closure, and add the quantum operands to the value `for`. 

So as long as the value `adjoint` can receive operands, the reference `adjoint` might not even need it, since it is a region! This also simplifies its verifiers.

mlir/include/Quantum/IR/QuantumOps.td
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2590/commits/2025aea69ff9804effdd69038cb39416492e6b33

mlir/test/Mitigation/ZneFoldingGlobalTest.mlir
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2590/commits/2025aea69ff9804effdd69038cb39416492e6b33

mlir/include/Quantum/IR/QuantumOps.td
dime10: I'm glad you figured out how to simplify the assembly :) But just to clarify, this comment wasn't about that, but about enforcing the relationship between the inputs/outputs _declaritively_ rather than through C++ logic. The trait does that, regardless of what's going on with the assembly format.

(It's the [comment further down](https://github.com/PennyLaneAI/catalyst/pull/2590#discussion_r2955503759) that was about the assembly format)

mlir/test/Gradient/AdjointTest.mlir
dime10: Also thoughts on this?

> I suppose we could modify the verifier to only check block args if a block exists

paul0403: The block is guaranteed to exist since the region on the adjoint op is 
```
let regions = (region SizedRegion<1>:$region);
```

But I don't remember if `SizedRegion<num_blocks>` is for building the op or parsing the op (or both), let me try your suggestion

dime10: You make a good point! So it's just that the block is not printed if it has no ops and no args?

mlir/include/Quantum/IR/QuantumOps.td
paul0403: I removed the verifier manually checking the args and result types now that I'm using `AllTypesMatch` :partying_face:

mlir/test/Gradient/AdjointTest.mlir
paul0403: I would suspect so, because if you remember, the old test was just 
```
    %1 = quantum.adjoint(%0) : !quantum.reg {}  // prevent folding of dealloc into alloc
    quantum.dealloc %1 : !quantum.reg
    return %arg0 : f64
```

i.e. no block to parse, but my new verifier was checking
```C++
    Block &b = this->getRegion().front();
    if (b.getNumArguments() != this->getArgs().size()) {
        return emitOpError("Adjoint op number of operands must be the same as the number of "
                           "arguments on its block");
    }
```

I added the block explicitly to these tests since this new verifier was failing, but it was an actual op error emitted, not a segfault! Which means the `Block &b` actually found a real `Block` in memory, making me believe that the block exists, just not printed if it's empty.

But I'll also try your suggestion, since I'm also not the biggest fan of needing to explicitly write out the empty block just for a verifier (I'm currently dealing with the xdsl failure in CI)

dime10: Well I think you're right, there will always be a block so the suggestion won't work.

So the things I would actually change is not use adjoint in this test, and canonicalize empty adjoints 😅 But that's out of scope :)

rniczh (approved):
💯

frontend/catalyst/python_interface/dialects/quantum/operations/miscellaneous_ops.py
paul0403: Didn't know about the alphabetical order, I moved it back down!

frontend/catalyst/jax_primitives.py
paul0403: I didn't update it because I didn't want to touch anything in the frontend. As of this PR, the jaxpr generated by the frontend still follows the assertion here.

mlir/test/QRef/UnitTests.mlir
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2590/commits/0096de13ee444283be991885f7d2c741e5e020fd

frontend/catalyst/jax_primitives.py
paul0403: I'll update these when I migrate the frontend to reference semantics, alongside all other ops that have this assumption (the control flow ones and subroutine).

PR #2591, opened 2026-03-18
Bump xdsl and xdsl-jax versions in prepration for removing the local stableHLO dialect

mehrdad2m:
**Context:**
We need to bump the xdsl and xdsl-jax version to be able to point to the upstream stableHLO and remove the local implementation in https://github.com/PennyLaneAI/catalyst/pull/2588

- Bump the versions
- xdsl's greedyPatternRewrite now runs dead-code elimination (DEC) to remove unused code. This causes quantum bits that are not deallocated to be removed and fails some tests. Added `quanutm.dealloc` to the end of the test IRs to keep the `quantum.custom` operation live and avoid elimination

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-101832]


setup.py
```diff
@@ -130,8 +130,8 @@ def parse_dep_versions():
     "numpy>2.0.0",
     "scipy-openblas32>=0.3.26",  # symbol and library name
     "diastatic-malt==2.15.3",
-    "xdsl==0.56.1",
-    "xdsl-jax==0.2.0",
+    "xdsl==0.57.3",
+    "xdsl-jax==0.3.0",
```
mehrdad2m:
```suggestion
    "xdsl-jax==0.4.0",
```
Update once the new version is available

Makefile
```diff
@@ -122,6 +123,8 @@ frontend:
 	# versions of a package with the same version tag (e.g. 0.38-dev0).
 	$(PYTHON) -m pip uninstall -y pennylane
 	$(PYTHON) -m pip install -e . --extra-index-url https://test.pypi.org/simple $(PIP_VERBOSE_FLAG)
+	# TODO: Remove once xdsl-jax is released.
+	$(PYTHON) -m pip install --no-deps --force-reinstall $(XDSL_JAX_URL) $(PIP_VERBOSE_FLAG)
```
mehrdad2m:
```suggestion
```
This should be removed once the new xdsl-jax version is released

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_merge_rotations.py
```diff
@@ -209,6 +224,7 @@ def test_adjoint_property(self, first_adj, second_adj, sign, run_filecheck):
                 // CHECK-NOT: "quantum.custom"
                 %1 = quantum.custom "RX"(%arg0) %0 {adj_string_0}: !quantum.bit
                 %2 = quantum.custom "RX"(%arg1) %1 {adj_string_1}: !quantum.bit
+                quantum.dealloc_qb %2 : !quantum.bit
```
mudit2812, line 227: Seems like a good change, but it seems unrelated to the content of the PR

paul0403: The version bump covered an unstream xdsl change that runs dead code elimination.

[Value semantics quantum.custom ops are pure](https://github.com/PennyLaneAI/catalyst/blob/802260577fa7a2fcc85b47204c70535e7bc06b04/frontend/catalyst/python_interface/dialects/quantum/operations/gate_ops.py#L107), so if they have no users they'll get removed, and the the parent gates will get recursively removed up the gate chain. The alloc op will also get removed if there's no users, due to its [explicit canonicalization pattern we have](https://github.com/PennyLaneAI/catalyst/blob/802260577fa7a2fcc85b47204c70535e7bc06b04/mlir/lib/Quantum/IR/QuantumOps.cpp#L104) (not sure if this canonicalization is in the xdsl alloc op).

The dealloc op is the concrete user of the entire chain, so that they don't get removed. You can see that [it has no explicit memory effect traits](https://github.com/PennyLaneAI/catalyst/blob/802260577fa7a2fcc85b47204c70535e7bc06b04/frontend/catalyst/python_interface/dialects/quantum/operations/qubit_ops.py#L92), which means it has the full default set of memory effects.

paul0403: Long story short, this is needed for the version bump in this PR.

paul0403 (approved):

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_merge_rotations.py
mudit2812: Oh, interesting. Thanks for explaining :)

mudit2812 (approved):

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_merge_rotations.py
mehrdad2m: Thanks @paul0403 for the explanation. For reference, this is the [upstream PR](https://github.com/xdslproject/xdsl/issues/5485) in xdsl that broke the test upon updating the xdsl version

setup.py
```diff
@@ -130,8 +130,8 @@ def parse_dep_versions():
     "numpy>2.0.0",
     "scipy-openblas32>=0.3.26",  # symbol and library name
     "diastatic-malt==2.15.3",
-    "xdsl==0.56.1",
-    "xdsl-jax==0.2.0",
+    "xdsl==0.58.0",
```
mudit2812: There was another release last night

```suggestion
    "xdsl==0.59.0",
```

mehrdad2m: yeah, there should be another xdsl-jax corresponding to that. I'll bump it when that's out as well.

mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/2591/commits/2cba55e2ffe0d66674938d2d941975456060441c

mudit2812 (approved):

PR #2592, opened 2026-03-18
Enable grouping MLIR passes to reduce the number of roundtrips with the xDSL pipeline

mudit2812:
**Context:**
Currently, when an MLIR transform is present in the pipeline with the unified compiler, we print the IR, invoke `quantum-opt` and parse the output back to xDSL. This happens for every single MLIR pass in the pipeline. This creates a lot of avoidable overhead if there are a lot of consecutive MLIR passes. We want to collect consecutive MLIR passes into a single pipeline that can be given to `quantum-opt`.

* Make `callback` argument of `ApplyTransformSequencePattern` required. This rewrite pattern should only be called when callbacks are being used.
* Add a `ApplyTransformSequenceNoCallbackPattern`, which creates sub-schedules containing consecutive xDSL/MLIR passes that are applied in one go.
  * This pattern is only used when there are no callbacks. We need to do this because callbacks **have** to be called after pass, so we can't schedule multiple passes together.

**Benefits:**
Much better performance when there are multiple consecutive MLIR passes

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-114436]

### DIFF
--- a/frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
+++ b/frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
@@ -208,7 +208,6 @@ class ApplyTransformSequenceNoCallbackPattern(RewritePattern):
     ):
         self.ctx = ctx
         self.passes = passes
-        self.pass_level = 0
 
     @op_type_rewrite_pattern
     def match_and_rewrite(self, transformer: builtin.ModuleOp, rewriter: PatternRewriter):
--- a/frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
+++ b/frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
@@ -259,7 +259,7 @@ class ApplyTransformSequenceNoCallbackPattern(RewritePattern):
     def _apply_xdsl_pipeline(
         self, payload: builtin.ModuleOp, pass_ops: list[transform.ApplyRegisteredPassOp]
     ) -> None:
-        """Interpret a sequence of ``ApplyRegisteredPassOp``\ s corresponding to xDSL passes."""
+        r"""Interpret a sequence of ``ApplyRegisteredPassOp``\ s corresponding to xDSL passes."""
         pipeline = []
         for op in pass_ops:
             pass_class = self.passes[op.pass_name.data]()
@@ -277,7 +277,7 @@ class ApplyTransformSequenceNoCallbackPattern(RewritePattern):
         pass_ops: list[transform.ApplyRegisteredPassOp],
         rewriter: PatternRewriter,
     ) -> builtin.ModuleOp:
-        """Interpret a sequence of ``ApplyRegisteredPassOp``\ s corresponding to MLIR passes."""
+        r"""Interpret a sequence of ``ApplyRegisteredPassOp``\ s corresponding to MLIR passes."""
         buffer = StringIO()
         Printer(stream=buffer, print_generic_format=True).print_op(payload)
 
--- a/frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
+++ b/frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
@@ -17,6 +17,7 @@ from collections.abc import Sequence
 from dataclasses import dataclass, field
 from functools import partial
 from io import StringIO
+from itertools import groupby
 from typing import Any, Callable
 
 from xdsl.context import Context
@@ -213,7 +214,8 @@ class ApplyTransformSequenceNoCallbackPattern(RewritePattern):
     def match_and_rewrite(self, transformer: builtin.ModuleOp, rewriter: PatternRewriter):
         """Rewrite modules containing transform.named_sequences."""
         payload: builtin.ModuleOp = transformer.parent_op()
-        rewriter.erase_op(transformer)
+        # Detach the transformer from the payload module without destroying it
+        transformer.detach()
 
         pass_ops = []
         for ns in transformer.ops:
@@ -225,35 +227,16 @@ class ApplyTransformSequenceNoCallbackPattern(RewritePattern):
         if len(pass_ops) == 0:
             return
 
-        schedule = self._cluster_passes(pass_ops)
-        for subschedule in schedule:
-            if subschedule[0].pass_name.data in self.passes:
-                self._apply_xdsl_pipeline(payload, subschedule)
+        # self.passes maps pass names to xDSL passes. If the pass name is in self.passes, we know
+        # that it is an xDSL pass. Otherwise, we assume that the name corresponds to an MLIR pass.
+        pass_groups = groupby(pass_ops, key=lambda op: is_xdsl_pass(op.pass_name.data))
+
+        for is_xdsl_group, group in pass_groups:
+            group = tuple(group)
+            if is_xdsl_group:
+                self._apply_xdsl_pipeline(payload, group)
             else:
-                payload = self._apply_mlir_pipeline(payload, subschedule, rewriter)
-
-    def _cluster_passes(
-        self, pass_ops: list[transform.ApplyRegisteredPassOp]
-    ) -> list[list[transform.ApplyRegisteredPassOp]]:
-        """Create a schedule that clusters consecutive xDSL/MLIR passes together."""
-        if not pass_ops:  # pragma: no cover
-            return []
-
-        schedule = []
-        cur_schedule = [pass_ops[0]]
-        cur_state = pass_ops[0].pass_name.data in self.passes
-
-        for op in pass_ops[1:]:
-            new_state = op.pass_name.data in self.passes
-            if new_state == cur_state:
-                cur_schedule.append(op)
-            else:
-                schedule.append(cur_schedule)
-                cur_schedule = [op]
-                cur_state = new_state
-
-        schedule.append(cur_schedule)
-        return schedule
+                payload = self._apply_mlir_pipeline(payload, group, rewriter)
 
     def _apply_xdsl_pipeline(
         self, payload: builtin.ModuleOp, pass_ops: list[transform.ApplyRegisteredPassOp]
--- a/frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
+++ b/frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
@@ -237,7 +237,7 @@ class ApplyTransformSequenceNoCallbackPattern(RewritePattern):
         self, pass_ops: list[transform.ApplyRegisteredPassOp]
     ) -> list[list[transform.ApplyRegisteredPassOp]]:
         """Create a schedule that clusters consecutive xDSL/MLIR passes together."""
-        if not pass_ops:
+        if not pass_ops:  # pragma: no cover
             return []
 
         schedule = []
--- a/frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
+++ b/frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
@@ -227,9 +227,7 @@ class ApplyTransformSequenceNoCallbackPattern(RewritePattern):
         if len(pass_ops) == 0:
             return
 
-        # self.passes maps pass names to xDSL passes. If the pass name is in self.passes, we know
-        # that it is an xDSL pass. Otherwise, we assume that the name corresponds to an MLIR pass.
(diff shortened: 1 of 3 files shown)


frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
```diff
@@ -193,7 +196,100 @@ def interpret_apply_registered_pass_op(
...
+
+        pipeline = _create_mlir_cli_schedule(pass_ops)
+        modified = _quantum_opt(*pipeline, "-mlir-print-op-generic", stdin=buffer.getvalue())
+
+        data = Parser(self.ctx, modified).parse_module()
+        rewriter.replace_op(payload, data)
+        return data
+
+
+def _create_mlir_cli_schedule(pass_ops: Sequence[transform.ApplyRegisteredPassOp]) -> list[str]:
```
mudit2812, line 283: This function didn't need to be updated because I implemented it so that we could create a schedule for multiple passes. We just didn't need to do it in the past.

frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
```diff
@@ -193,7 +196,100 @@ def interpret_apply_registered_pass_op(
         return data
 
 
-def _create_mlir_schedule(pass_ops: Sequence[transform.ApplyRegisteredPassOp]) -> list[str]:
+class ApplyTransformSequenceNoCallbackPattern(RewritePattern):
```
dime10, line 203: It feels like there could be a bit of duplication, but it's not terrible

frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
```diff
@@ -193,7 +196,100 @@ def interpret_apply_registered_pass_op(
...
+        self.pass_level = 0
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, transformer: builtin.ModuleOp, rewriter: PatternRewriter):
+        """Rewrite modules containing transform.named_sequences."""
+        payload: builtin.ModuleOp = transformer.parent_op()
+        rewriter.erase_op(transformer)
+
+        pass_ops = []
+        for ns in transformer.ops:
```
dime10: Is transformer not gone at this point 😳

frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
```diff
@@ -193,7 +196,100 @@ def interpret_apply_registered_pass_op(
...
+            assert isinstance(ns, transform.NamedSequenceOp)
+            pass_ops.extend(
+                op for op in ns.body.ops if isinstance(op, transform.ApplyRegisteredPassOp)
```
dime10, line 229: What other kind of ops might we expect?

frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
```diff
@@ -193,7 +196,100 @@ def interpret_apply_registered_pass_op(
...
+            )
+
+        if len(pass_ops) == 0:
+            return
+
+        schedule = self._cluster_passes(pass_ops)
+        for subschedule in schedule:
+            if subschedule[0].pass_name.data in self.passes:
```
dime10: This condition is (edit:) *not immediately obvious to me, what are you checking?

dime10 (commented):
I don't quite get the clustering logic yet, but this is really nice to have 💯

frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
```diff
@@ -193,7 +196,99 @@ def interpret_apply_registered_pass_op(
...
+                self._apply_xdsl_pipeline(payload, subschedule)
+            else:
+                payload = self._apply_mlir_pipeline(payload, subschedule, rewriter)
+
+    def _cluster_passes(
```
mehrdad2m: suggestion: I found that we have this utility in Python that seems to do the exact same thing as this function. 

```python
itertools.groupby(pass_ops, key=lambda op: op.pass_name.data in self.passes)
```

frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
mudit2812: It's no longer in the payload module, but it still exists.

frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
mudit2812: `transform.yield`

frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
mudit2812: `self.passes` is a registry of _xDSL_ passes. If the pass name is not in `self.passes`, then we assume that the schedule contains MLIR passes

frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
mehrdad2m: I agree that this feels a bit weird. It probably works because the operation is not truly destructed when calling the `erase_op` in xdsl and only the use-def edges are modified. This pattern would definitely not work if written in MLIR 😆

frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
mudit2812: I'll add a dev comment to make this clearer.

mehrdad2m (approved):

frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
mudit2812: This is **almost** the same logic in the MLIR pass as well. Find the transformer module, erase it from the qnode module, and then interpret the transformer. The difference is that in the MLIR pass, a clone of the transformer is saved, not the original transformer.

Although, I don't find this to be particularly weird. The python object wouldn't be destroyed until we exit the function or assign a different value to `transformer`, and it is already detached from the qnode module, so I don't really see why this is odd.

frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
mudit2812: Lol nice. I'll use this

frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
mudit2812: I tried to minimize it as best as possible, but yes, there is definitely still a bit of duplication.

frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
mudit2812: Oh, I see what you meant @mehrdad2m . The MLIR `Operation::erase()` literally also destroys the object

mudit2812: Perhaps it'll be slightly easier to understand if I use `op.detach()` instead? That would more robustly indicate that the operation is removed from its parent block but not destroyed.

mehrdad2m: I see, yeah I guess it is fine in Python, just coming from MLIR it would feel weird because you would have to explicitly do the clone

dime10: > The difference is that in the MLIR pass, a clone of the transformer is saved, not the original transformer.

That is a huge difference, because erase actually does delete the object in MLIR. So the api in xDSL doesn't really do what it says on the tin, because of Python's garbage collector. Conceptually, it does feel like a use after free.

There is a separate detach function in MLIR (not sure what the name is) to remove it from the parent, but not delete it. Maybe that would be more appropriate? 
While this works in xDSL, given that xDSL -> MLIR porting is supposed to be major use case of the Python interface, I don't know if we want to encourage this pattern since copying it as is to C++ would definitely crash the compiler.

Edit: Your new messages weren't loaded yet, but yeah that's perfect 👍

frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
dime10: Thanks, that will help :)

mehrdad2m (approved):

frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
```diff
@@ -193,7 +197,85 @@ def interpret_apply_registered_pass_op(
...
+    ):
+        self.ctx = ctx
+        self.passes = passes
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, transformer: builtin.ModuleOp, rewriter: PatternRewriter):
+        """Rewrite modules containing transform.named_sequences."""
+        payload: builtin.ModuleOp = transformer.parent_op()
+        # Detach the transformer from the payload module without destroying it
+        transformer.detach()
```
dime10: Thanks for updating this. I got curious and wanted to know what the difference between xDSL `erase` and `detach` were, and whether `detach` would be enough for Python GC to delete it later.

Definitions:
```py
### Rewriter ###

    def erase_op(op: Operation, safe_erase: bool = True):
        """
        Erase an operation.
        Check that the operation has no uses, and has a parent.
        If safe_erase is True, check that the operation has no uses.
        Otherwise, replace its uses with ErasedSSAValue.
        """
        if (block := op.parent) is not None:
            block.erase_op(op, safe_erase=safe_erase)
        else:
            op.erase(safe_erase=safe_erase)

### Block ###

    def erase_op(self, op: Operation, safe_erase: bool = True) -> None:
        """
        Erase an operation from the block.
        If safe_erase is True, check that the operation has no uses.
        """
        op = self.detach_op(op)
        op.erase(safe_erase=safe_erase)

    def detach_op(self, op: Operation) -> Operation:
        """
        Detach an operation from the block.
        Returns the detached operation.
        """
        if op.parent is not self:
            raise ValueError("Cannot detach operation from a different block.")
        op.parent = None

        prev_op = op.prev_op
        next_op = op.next_op

        if prev_op is not None:
            # detach op from linked list
            prev_op._next_op = next_op  # pyright: ignore[reportPrivateUsage]
            # detach linked list from op
            op._prev_op = None  # pyright: ignore[reportPrivateUsage]
        else:
            # reattach linked list if op is first op this block
            assert self._first_op is op
            self._first_op = next_op

        if next_op is not None:
            # detach op from linked list
            next_op._prev_op = prev_op  # pyright: ignore[reportPrivateUsage]
            # detach linked list from op
            op._next_op = None  # pyright: ignore[reportPrivateUsage]
        else:
            # reattach linked list if op is last op in this block
            assert self._last_op is op
            self._last_op = prev_op

        return op

### Operation ###

    def erase(self, safe_erase: bool = True, drop_references: bool = True) -> None:
        """
        Erase the operation, and remove all its references to other operations.
        If safe_erase is specified, check that the operation results are not used.
        """
        assert self.parent is None, (
            "Operation with parents should first be detached " + "before erasure."
        )
        if drop_references:
            self.drop_all_references()
        for result in self.results:
            result.erase(safe_erase=safe_erase)

    def detach(self):
        """Detach the operation from its parent block."""
        if self.parent is None:
            raise ValueError("Cannot detach a toplevel operation.")
        self.parent.detach_op(self)
```

So it looks like a "rewriter erase" is a "block detach" (removing the op from the linked list) + an "op erase". The op erase actually appears do some work in order to guarantee that no other references point to it, as well as erasing the results (with safe erase this is just an error check). 

Worth including an erase call at the end or is that overkill?

dime10 (approved):

frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py
mudit2812: I think detach is safe enough for us, but for the sake of being safe, I think makes sense to use `rewriter.erase_op(op)` instead of `op.detach()`. I'll add a dev comment to drive home the fact that the operation will be _erased from the block_ without destroying the object

mudit2812: [41a6bbb](https://github.com/PennyLaneAI/catalyst/pull/2592/commits/41a6bbb6c0535d1d79ee0eb6e93f7ce0fd13102d)

PR #2593, opened 2026-03-18
Fix type hints in xDSL compiler_transform

joeycarter:
**Context:** The `module_pass` variable in the `compiler_transform` module had the type hint `ModulePass`, suggesting it should be an instance of this class, but really it should be `type[ModulePass]` to indicate that it should be a *type* that inherits from `ModulePass`.

**Description of the Change:** Change type hint `ModulePass` -> `type[ModulePass]`, where appropriate.

**Benefits:** Better type hints, fewer Python type-checker warnings in developers' IDEs.

**Possible Drawbacks:** None, type hints have no effect at runtime for most practical purposes.


mudit2812 (approved):
Thanks @joeycarter !

mehrdad2m (approved):

PR #2596, opened 2026-03-19
Fix catalyst `--emit-bytecode` flag

### LINKED ISSUE #2594: Catalyst CLI `--emit-bytecode` emits text
The `catalyst` CLI utility with `tool=opt` does not correctly emit bytecode when called with the `--emit-bytecode` flag. The source of this issue is likely that the CLI argument `--emit-bytecode` is not correctly registered to the utility (it accepts the argument but does not apply any changes based on the value). The corresponding logic in `mlir-opt` can be found [here](https://github.com/llvm/llvm-project/blob/8f264586d7521b0e305ca7bb78825aa3382ffef7/mlir/lib/Tools/mlir-opt/MlirOptMain.cpp#L602-L611).

MRE:
```shell
catalyst --tool=opt --emit-bytecode
module {}
```
expected output (generated by `quantum-opt --emit-bytecode`):
```
MLIR22.0.0git
P7builtinmodule<stdin>
```
actual output:
```mlir
module {
}
```

kipawaa:
**Context:**
The `catalyst` CLI tool has some dangling options (i.e. accepts the flag but does not change behaviour). One of these is the `--emit-bytecode` option.

Correctly registers and applied bytecode logic when the `--emit-bytecode` flag is present.

**Benefits:**
The `catalyst` CLI tool can emit bytecode, and responds correctly to the flag.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
closes #2594

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -340,6 +340,9 @@
 
 <h3>Bug fixes 🐛</h3>
 
+* Fixed a bug where the `catalyst` CLI tool would emit text when called with `--emit-bytecode`.
+  [(#2596)](https://github.com/PennyLaneAI/catalyst/pull/2596)
+
 * Fixed a bug where input array arguments could be mutated during execution when copied inputs
   were updated in-place. Entry-point arguments are now treated as non-writable during bufferization,
   preserving the expected immutability of user inputs.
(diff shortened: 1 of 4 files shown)


dime10 (approved):
Thanks @kipawaa !

paul0403 (approved):

PR #2597, opened 2026-03-19
Fix quantum kernel detection

dime10:
PR #2497 Introduced a verifier for the `quantum.node` attribute, and also expanded verification for measurement processes. The issue is that detection logic is not very robust because we don't have a dedicated quantum kernel operation. The verification is intended to only run on operations within a quantum kernel.

The detection logic is updated to detect the presence of the transform program in the quantum kernel, making it more robust / narrow. This also allows us to revert the test changes from the previous PR (actual changes in this PR are very small, see commits).

#### Details

The verifier is intended to catch missing quantum.node attributes within quantum kernels (the abstraction frontend-scheduled passes run on), as per [the guide](https://github.com/PennyLaneAI/catalyst/blob/main/frontend/catalyst/python_interface/transforms/qnode-transform-guide.md). For instance, the following should raise an error:
```mlir
module @root {
  func.func public @jit_main_function() { ... }
  module @kernel_module {
    module attributes {transform.with_name_sequence} { ... }
    func.func @circuit() {  // missing quantum.node attribute
      ...
      quantum.expval %obs : f64
    }
  }
}
```

The issue prompting this fix is that the verifier was erroneously triggering when sending and intermediate output from the compiler back into the opt tool. For instance, while the following IR would not trigger a verification error while already loaded in memory (only one module):
```mlir
module {
  func.func @circuit() {
    ...
    quantum.expval %obs : f64
  }
}
```
it will trigger when parsing it back with our command line tools.

From testing I discovered that round-tripping a piece of IR with opt will run the verifier twice, and one of those times an additional module is wrapped around the program (for what reason I don't know yet). So the above program will be verified twice with:
1) `quantum.expval` having two parent modules
2) --------"-------- having one parent module

Note that this phenomenon is not the same as opt inserting a module into the IR when none is present in the input, e.g.
```mlir
quantum.expval %obs : f64
```
turning into
```
module {
  quantum.expval %obs : f64
}
```
This behaviour can be turned off with `--no-implicit-module` and is orthogonal to behaviour described above.

[sc-107492]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -289,6 +289,16 @@
   `shots` aren't set. Instead, an informative error message is raised.
   [(#2456)](https://github.com/PennyLaneAI/catalyst/pull/2456)
 
+* The quantum kernel abstraction in Catalyst's IR (a nested module operation with its own transform
+  schedule and entry point and subroutine functions representing a PennyLane QNode) has been
+  documented and equipped with additional verification. Transformation passes scheduled from the
+  frontend must ensure, and can rely on, the presence of the `quantum.node` attribute to indicate
+  which functions in the module represent a separate quantum execution (with device initialization,
+  shots configuration, and set of measurement processes).
+  [(#2483)](https://github.com/PennyLaneAI/catalyst/pull/2483)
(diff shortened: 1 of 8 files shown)


rniczh (approved):
looks good to me 👍

mehrdad2m (approved):

PR #2598, opened 2026-03-19
Faster and more robust MLIR generation for inspection functionality

dime10:
The `get_mlir_module` function is used by a few Python interface functions (e.g. drawing, specs) to fetch the starting IR for their process. In particular, when the program doesn't already have an MLIR representation (e.g. from AOT compilation or a previous execution), the code will run full jit compilation on the QJIT object.

This is unnecessary, because we are only interested in the starting module. We can save a good amount of time by not compiling down all the way to LLVM and binary (which can be costly for large programs). Additionally, if we wish to inspect programs that cannot be compiled to that level, this would also fail.

[sc-114933]

### DIFF
--- a/frontend/catalyst/python_interface/inspection/xdsl_conversion.py
+++ b/frontend/catalyst/python_interface/inspection/xdsl_conversion.py
@@ -18,7 +18,6 @@ from __future__ import annotations
 import contextlib
 import inspect
 from collections.abc import Callable
-from copy import deepcopy
 from itertools import compress
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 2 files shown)


frontend/catalyst/python_interface/inspection/xdsl_conversion.py
```suggestion
```

mudit2812 (approved):
Beautiful. This will probably also fix the issue we noticed with not being able to lower to LLVM IR when the MLIR is in generic format, since we won't do a full compilation

mehrdad2m (approved):

frontend/catalyst/python_interface/inspection/xdsl_conversion.py
dime10: Github is bugging out and won't accept this suggestion lol I'll do it manually

dime10: https://github.com/PennyLaneAI/catalyst/pull/2598/commits/2aec1df8d801718021b9710db7413cd1bfbcda11

PR #2599, opened 2026-03-19
Allow easy-ish MLIR compilation from Python without llvm or binary generation

dime10:
The compilation options to do this are already available (`lower_to_llvm`, as well as the [recently added](https://github.com/PennyLaneAI/catalyst/commit/64b8d27de5106c4727b7e3bafa92c71dfec8a336) `link` option), however there is an assertion that still prevents these flags from working with the `QJIT.compile` stage.

Note that these flags are *not* exposed to the `qjit` decorator, so must be provided upon manual `QJIT` construction or set on the instantiated object.

[sc-115268]


rniczh (approved):

paul0403 (approved):

mehrdad2m (approved):

frontend/test/pytest/test_compiler.py
```diff
@@ -270,6 +270,32 @@ def f():
...
+        """Test that QJIT compilation succeeds without LLVM lowering (or linking)."""
+
+        def f(x):
+            return x + 1.0
+
+        test_pipes = [("test_pipe", ["canonicalize"])]
+        options = CompileOptions(lower_to_llvm=False, link=False, pipelines=test_pipes)
+        compiled = QJIT(f, options)
+
+        compiled.workspace = compiled._get_workspace()
```
dime10:
```suggestion
        compiled.workspace = compiled._get_workspace()  # pylint: disable=protected-access
```

frontend/test/pytest/test_compiler.py
```diff
@@ -270,6 +270,32 @@ def f():
...
+        compiled.workspace = compiled._get_workspace()  # pylint: disable=protected-access
+        compiled.jaxpr, *_ = compiled.capture((0.5,))
+        compiled.mlir_module = compiled.generate_ir()
+        compiled.compile()
+
+        assert compiled.jaxpr is not None
+        assert compiled.mlir is not None
+        assert compiled.mlir_opt is not None
+
+        assert compiled.llvm_ir is None
```
dime10: I just realized that setting `keep_intermediate=True` will fail this assertion. It appears that the keep intermediate mechanism will bypass the other compile options and produce llvm_ir anyway (possibly also object code), which defeats the point of the options.

@mehrdad2m Do you know why this could be?

frontend/test/pytest/test_compiler.py
```diff
@@ -270,6 +270,32 @@ def f():
...
+            return x + 1.0
+
+        test_pipes = [("test_pipe", ["canonicalize"])]
+        options = CompileOptions(lower_to_llvm=False, link=False, pipelines=test_pipes)
+        compiled = QJIT(f, options)
+
+        compiled.workspace = compiled._get_workspace()  # pylint: disable=protected-access
+        compiled.jaxpr, *_ = compiled.capture((0.5,))
+        compiled.mlir_module = compiled.generate_ir()
+        compiled.compile()
```
mehrdad2m, line 293:
```suggestion
        compiled.compiled_function, compiled.llvm_ir = compiled.compile()
```
I think we should explicitly set these, otherwise nothing get assigned to them internally.

frontend/test/pytest/test_compiler.py
mehrdad2m: hmm I don't see any object file created when testing locally. Do you have a test you can share where you see this behaviour? What happens for me is that instead of an LLVM module, the MLIR module is printed as `llvm_ir`. This is because naturally `catalyst --tool opt input -o output.mlir` prints the transformed MLIR to the output file even if it is not lowering to llvm IR. Then in `run_from_ir`, we open the output file: https://github.com/PennyLaneAI/catalyst/blob/ae7f53f7910cd1ea4cf90f9b0e78d6e45ee6ba63/frontend/catalyst/compiler.py#L479-L483
and return the out_IR which is assigned to `llvm_ir` at the call site. 
So within the function `out_IR` is just an IR that could be either an LLVM IR or MLIR IR, but the caller only expects the LLVM IR.  
Now if we don't care about the out_IR to be general we can easily fix this by checking against `lower_to_llvm` so something like
```python
        if self.options.lower_to_llvm and os.path.exists(output_ir_name):
            with open(output_ir_name, "r", encoding="utf-8") as f:
                out_IR = f.read()
        else:
            out_IR = None
```
instead

frontend/test/pytest/test_compiler.py
dime10: That's fine right, these are `None` anyway since the point of the options is not to link (i.e. no shared lib and thus no compiled_function object), and not to generate llvm (so no llvm_ir object).

dime10: In this test my goal is to run the compiler but without going lower than the MLIR representation (e.g. it's possible that some of the passes or cmd options generate an output).

frontend/test/pytest/test_compiler.py
dime10: Thanks, let me look into this again maybe it was something else I saw

frontend/test/pytest/test_compiler.py
mehrdad2m: yeah but to test weather we generate a `llvm_ir` object or not, you need to check the output of the `compile()` function otherwise this assertion `assert compiled.llvm_ir is None` would be meaningless and will always pass no matter what the options are.

dime10: Yeah that's a good point. I mean it does test that those properties aren't otherwise written to (without assuming that only `jit_compile` and `aot_compile` write to them), but it doesn't test what I ultimately want to test. Ideally, the test would check that the compiler ran but without the translate to llvm or link step being hit. Right now:
- it doesn't actually test that the compiler ran
  - I was originally going to test this with the keep_intermediate option, when I ran into the issue in my other comment.
  - If that's not a problem anymore I could write such a test, but maybe there is a better way to check this (do you have any ideas?).
  - If we had a proper mlir_out field this would easier to check (well, assuming we are writing to it somewhere), but the `.mlir_opt` property is independent from the main compilation functions.
- testing whether the lower stages ran is very weak like you point out
  - This is a bit harder to test, checking the output of `compile()` would be a good start, but maybe there are better ways? 🤔

The other thing to consider is whether we want to use `.compile()` for this use case. It could still be a nice way to trigger the compiler, but I just realized `.mlir_opt` also does the same thing while returning the output IR. It's just it's a property and I don't really like that it has an action every-time you access it, that's not how a property should behave.

mehrdad2m: yeah I don't want to encourage testing using `.mlir_opt` either. Hopefully we can find a better and to remove it in future. 
So the problem is that we want to test how the C++ compiler ran from python and we have very limited ways to interact with it: 
1. To observe the output result which you mentioned. 
2. To observe the input command line and the flags passed to it and reason about what the C++ tool does given that input. This is not ideal imo.
3. To observe the compiler's verbosity. One form of this is the `keep_intermediate` dump files. Another is the `verbose=True` option, which already prints `Compilation successful` to confirm the compiler ran. We could maybe extend this to also report which stages completed (e.g. "opt done", "translate done", "llc done"), which would give us a direct way to assert that specific stages did or didn't run. I personally prefer the last option (verbose=True) since it would not mess with the file system. What do you think?

dime10: Interesting idea for 3, I like it!

dime10: I've implemented your suggestion in the last two commits, let me know what you think :)

frontend/test/pytest/test_compiler.py
dime10: I couldn't reproduce either, closing

frontend/test/pytest/test_compiler.py
mehrdad2m: Nice! Looks great!

mehrdad2m (approved):

PR #2600, opened 2026-03-19
feat: enable basic visualization of `gridsynth` with `draw_graph`

andrijapau:
**Context:**

This PR accomplishes a few things,
- Adds support for `IndexSwitchOp`
- Adds support for `CallOp`
- Improves `FuncOp` handling

**Benefits:**

This would allow us to visualize `gridsynth` in a simple form like,

```python
import pennylane as qml
from catalyst import *

null_dev = qml.device("null.qubit", wires=1)

@qjit(capture=True)
@qml.transforms.gridsynth
@qml.qnode(null_dev)
def circuit():
    qml.RZ(0.5, wires=0)
    return qml.probs()

print(draw_graph(circuit)())
```

image

**Possible Drawbacks:**

Some users might want to see an expand all switch cases. This is not possible at the moment, but it could be introduced later through some visualization configuration that is passed to `draw_graph`.

**Related GitHub Issues:** 

Fixes https://github.com/PennyLaneAI/catalyst/issues/2608
Fixes https://github.com/PennyLaneAI/catalyst/issues/2607
Fixes #2609


frontend/test/pytest/python_interface/inspection/test_construct_circuit_dag.py
andrijapau:
```suggestion
from xdsl.dialects import test
```

frontend/catalyst/python_interface/inspection/construct_circuit_dag.py
```diff
@@ -539,14 +560,14 @@ def _if_op(self, operation: scf.IfOp):
         # blocking
         if (
             before_dyn_node_uid == current_dyn_node_uid
-            and sum(1 for s in self._cluster_uid_stack if "conditional" in s) == 1
+            and sum(1 for s in self._cluster_uid_stack if self._is_branching_cluster(s)) == 1
```
andrijapau:
```suggestion
            and sum(1 for s in self._cluster_uid_stack if "conditional" in s) == 1
```

PR #2603, opened 2026-03-19
add dynamic shapes support to for_loop with capture

albi3ro:
**Context:**

Makes `qml.for_loop` work with dynamic shapes and program capture.

Adds a lowering rule for plxpr's for loop. Translates to the same primitive in `from_plxpr` instead of translate to the legacy frontend version.

**Benefits:**

Abstract shape support! Also, we will soon be able to get rid of the registrations entirely!

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-114569]


JerryChen97 (commented):
Which test(s) covered the case we would like to fix in https://github.com/PennyLaneAI/pennylane/pull/9299?

albi3ro:
> Which test(s) covered the case we would like to fix in [PennyLaneAI/pennylane#9299](https://github.com/PennyLaneAI/pennylane/pull/9299)?



> Which test(s) covered the case we would like to fix in [PennyLaneAI/pennylane#9299](https://github.com/PennyLaneAI/pennylane/pull/9299)?

There's a whole bunch all over the place, but I did just happen to fix one instance of it in `test_from_plxpr.py`, so that should show up in the git diff.

frontend/catalyst/from_plxpr/control_flow.py
```diff
@@ -25,6 +25,7 @@
 from pennylane.capture.primitives import cond_prim as plxpr_cond_prim
 from pennylane.capture.primitives import for_loop_prim as plxpr_for_loop_prim
 from pennylane.capture.primitives import while_loop_prim as plxpr_while_loop_prim
+from pennylane.math import is_abstract
```
albi3ro:
```suggestion
```

frontend/catalyst/from_plxpr/control_flow.py
```diff
@@ -288,15 +290,15 @@ def handle_for_loop(
...
     )
 
+    qregs = [*[dyn_qreg.get() for dyn_qreg in dynalloced_qregs], self.init_qreg.get()]
     start_plus_args_plus_qreg = [
+        *abstract_shapes,
         start,
         *args,
         *[dyn_qreg.get() for dyn_qreg in dynalloced_qregs],
-        self.init_qreg.get(),
+        qregs[-1],
```
paul0403: What happened here to the qregs? 😅

frontend/catalyst/jax_primitives.py
```diff
@@ -2681,6 +2682,64 @@ def _for_loop_lowering(
     return for_op_scf.results
 
 
+# pylint: disable=too-many-arguments
+def _pl_for_loop_lowering(
```
paul0403, line 2686: 🥳

frontend/test/pytest/test_jax_dynamic_api.py
```diff
@@ -400,317 +400,367 @@ def fun(sz, idx, val):
         assert_array_and_dtype_equal(result, expected)
 
 
-def test_quantum_tracing_1():
-    """Test that catalyst tensor primitive is compatible with quantum tracing mode"""
+class TestForLoopDynamicShapes:
```
paul0403, line 403: The git diff looks a bit butchered, but essentially you just added the `TestForLoopDynamicShapes` tests and everything moved down (with some occasional changes, which confused the diff renderer) right?

paul0403 (approved):
Amazing! 🥳 

And, after #2694 , we can get rid of the registrations altogether (whose only purpose is to handle the reference-to-value semantics conversion during the plxpr-to-catxpr interpreter), and just do nothing to the pl loop primitives!

frontend/test/pytest/test_jax_dynamic_api.py
albi3ro: Yep.  I grouped all the for loop stuff into a class and left everything else below it.

frontend/catalyst/from_plxpr/control_flow.py
albi3ro: stuck everything in `qregs` then didn't actually use the `qregs`... 🤦 . Tidied up now.

frontend/test/lit/test_decomposition.py
JerryChen97, line 1: Is this also related to for loops?

albi3ro: This was a failure in L/L/L.  I couldn't figure out how to fix it, and I got the go ahead to just delete it.

JerryChen97 (approved):
Thanks a lot!!! Can't wait to see `for_loop` has full DS support!

frontend/test/lit/test_decomposition.py
paul0403: This is essentially the version bump changing some decomp rules.

PR #2604, opened 2026-03-19
Add global phase merging pass

dime10:
Useful for reducing number of instructions. Within a given scope, control-free global phase ops can always be merged into a single one.

Also cleans up assembly format inconsistencies between the quantum and qref dialect, as well as renaming the confusing `params` name to `angle` for (single) float param of the operation.

--- 

Another useful pass would be removing (control-free) global phases entirely, since some workflows may not care about them.

An improvement on this pass would merge them across certain region kinds as well.

[sc-114931]


mlir/lib/Quantum/Transforms/merge_global_phase.cpp
```diff
@@ -0,0 +1,73 @@
...
+                    auto phases = block.getOps<GlobalPhaseOp>();
+                    auto simplePhases = llvm::make_filter_range(phases, [](GlobalPhaseOp phaseOp) {
+                        return phaseOp.getInCtrlQubits().empty();
+                    });
+                    if (simplePhases.empty() ||
+                        std::next(simplePhases.begin()) == simplePhases.end()) {
+                        continue;
+                    }
+
+                    GlobalPhaseOp firstPhase = *simplePhases.begin();
```
lazypanda10117: I think this can be problematic to take the first global phase op because the later global phase ops may depend on values that are defined after this op?

For example, consider this scenario:

```mlir
func.func @merge_simple_global_phases(%arg0: f64, %arg1: f64) {
    quantum.gphase(%arg0)
    %c1 = arith.constant 2.0 : f64
    %c2 = arith.mulf %arg1, %c1 : f64
    quantum.gphase(%c2)
    return
}
```

Right now, this would give the following error:

```sh
tmp/merge_test.mlir:7:5: error: operand #1 does not dominate this use
    quantum.gphase(%c2)
    ^
tmp/merge_test.mlir:7:5: note: see current operation: %0 = "arith.addf"(%arg0, %2) <{fastmath = #arith.fastmath<none>}> : (f64, f64) -> f64
tmp/merge_test.mlir:6:11: note: operand defined here (op in the same block)
    %c2 = arith.mulf %arg1, %c1 : f64
          ^
```

dime10: Great catch, thank you! Taking the last one should be safe right?

lazypanda10117: yep, taking the last one should be good!

dime10: https://github.com/PennyLaneAI/catalyst/pull/2604/commits/446f3c3a20758640a4775c30c114bf030cadf9e4

lazypanda10117 (approved):

mlir/test/Quantum/MergeGlobalPhaseTest.mlir
```diff
@@ -0,0 +1,119 @@
...
+// CHECK-LABEL: func.func @keep_controlled_global_phases(
+func.func @keep_controlled_global_phases(%arg0: f64, %arg1: f64, %arg2: f64, %q: !quantum.bit) {
+    // CHECK-SAME: [[A:%.+]]: f64, [[B:%.+]]: f64, [[C:%.+]]: f64, [[Q:%.+]]: !quantum.bit
+
+    // CHECK-NOT: quantum.gphase
+    // CHECK: quantum.gphase([[B]]) ctrls([[Q]])
+    // CHECK-NOT: quantum.gphase
+    // CHECK: [[SUM:%.+]] = arith.addf [[C]], [[A]]
+    // CHECK: quantum.gphase([[SUM]])
+    // CHECK-NOT: quantu.gphase
```
mehrdad2m:
```suggestion
    // CHECK-NOT: quantum.gphase
```
I just realized how dangerous typos can be in `CHECK-NOT`s lol

mlir/lib/Quantum/Transforms/merge_global_phase.cpp
```diff
@@ -0,0 +1,73 @@
...
+    {
+        ModuleOp mod = getOperation();
+        OpBuilder builder(mod->getContext());
+
+        mod.walk([&](Operation *op) {
+            for (Region &reg : op->getRegions()) {
+                for (Block &block : reg.getBlocks()) {
+                    auto phases = block.getOps<GlobalPhaseOp>();
+                    auto simplePhases = llvm::make_filter_range(phases, [](GlobalPhaseOp phaseOp) {
+                        return phaseOp.getInCtrlQubits().empty();
```
mehrdad2m: How about adjoints? I think we should either exclude them here or implement proper logic to support those.

mlir/lib/Quantum/Transforms/merge_global_phase.cpp
```diff
@@ -0,0 +1,73 @@
...
+    using impl::CombineGlobalPhasesPassBase<CombineGlobalPhasesPass>::CombineGlobalPhasesPassBase;
+
+    void runOnOperation() final
+    {
+        ModuleOp mod = getOperation();
+        OpBuilder builder(mod->getContext());
+
+        mod.walk([&](Operation *op) {
+            for (Region &reg : op->getRegions()) {
+                for (Block &block : reg.getBlocks()) {
```
mehrdad2m:
```suggestion
        mod.walk([&](Block *block) {
```
I am wondering if we could do this instead? 🤔

mlir/test/Quantum/MergeGlobalPhaseTest.mlir
dime10: lol, thanks!

mlir/lib/Quantum/Transforms/merge_global_phase.cpp
dime10: Let me look into it, that might be a nice simplification!

mlir/lib/Quantum/Transforms/merge_global_phase.cpp
dime10: Good catch! I think it's also missing in the xdsl version

mlir/lib/Quantum/Transforms/merge_global_phase.cpp
dime10: https://github.com/PennyLaneAI/catalyst/pull/2604/commits/445252f3f996b5d7d100bb33728c48b7427fec30

mlir/test/Quantum/MergeGlobalPhaseTest.mlir
dime10: https://github.com/PennyLaneAI/catalyst/pull/2604/commits/a2cdf69eaa1e63c3d90447f05407a9c1d0b04f86

mlir/lib/Quantum/Transforms/merge_global_phase.cpp
dime10: https://github.com/PennyLaneAI/catalyst/pull/2604/commits/5c1c019a7a203a08f5955b80e028a48edef14da0, let me know what you think

mehrdad2m: Looks great 😍 Thanks for adding this. Can we add a similar test for the xdsl version as well?

dime10: Yeah you're right 🙈 To be honest I'd rather just remove it 😅, because why support two versions of the same thing? @mudit2812 Do you have a strong reason for wanting to keep it, given the additional maintenance overhead surfaced here?

We haven't really talked about our porting strategy for xDSL passes if they've been prototyped there, but I was assuming once in MLIR we would delete the xDSL version. Maybe worth agreeing on a strategy?

mehrdad2m: yeah make sense. I think you would want to keep something like this only if there is further prototyping on the basis of the existing pass that needs to be done in future, otherwise, no point in keeping it. Maybe one strategy could be to remove the pass after consulting the original writers?

mudit2812: The xDSL pass was added because the FTQC squad needed it, so we can definitely remove it. We do have a couple of other duplicate passes as well, but those were just for prototyping (cancel inverses, merge rotations)

dime10: Alright I'll check in with them if they're happy using the MLIR going forward, thanks!

dime10: https://github.com/PennyLaneAI/catalyst/pull/2604/commits/daefaf55429ff7739914a61b04cb7640cd01e756

mehrdad2m (approved):

PR #2605, opened 2026-03-20
Refactor post-processing in measurements_from_samples

lillian542:
**Context:**
One of several PRs for [[sc-107541](https://app.shortcut.com/xanaduai/story/107541)]. This one addresses refactoring to follow the new guidelines in the [qnode transform guide
](https://github.com/PennyLaneAI/catalyst/blob/main/frontend/catalyst/python_interface/transforms/qnode-transform-guide.md) regarding transforms that add post-processing

We refactor to follow the new guidelines in the new [qnode transform guide](https://github.com/PennyLaneAI/catalyst/blob/main/frontend/catalyst/python_interface/transforms/qnode-transform-guide.md). The main difference is we create a separate function for each quantum.node that calls the quantum node and applies post-processing on the results, i.e. we go from including post-processing inside the qnode:

```
  func.func public @circuit() -> tensor<f64> attributes {quanutm.node} {
    %cst = arith.constant dense<0> : tensor<i64>
    %c1000_i64 = arith.constant 1000 : i64
    quantum.device shots(%c1000_i64)
    %0 = op.test —> !quantum.bit
    %1 = quantum.compbasis %0 : !quantum.obs
    %2 = quantum.sample %1 : tensor<1000x1xf64>
    %3 = call @expval_from_samples.tensor.1000x1xf64(%2, %cst) : (tensor<1000x1xf64>, tensor<i64>) -> tensor<f64>
    return %3 : tensor<f64>
```
to calling the qnode (which returns samples), and performing post-processing on the results:

```
  func.func public @circuit() -> tensor<f64> {
    %0 = call @circuit.from_samples() : () -> tensor<1000x1xf64>
    %1 = call @expval_from_samples.tensor.1000x1xf64(%0) : (tensor<1000x1xf64>) -> tensor<f64>
    return %1 : tensor<f64>
  }
  func.func public @circuit.from_samples() -> tensor<1000x1xf64> attributes {quantum.node} {
    %c1000_i64 = arith.constant 1000 : i64
    quantum.device shots(%c1000_i64) 
    %0 = op.test —> !quantum.bit
    %1 = quantum.compbasis %0 : !quantum.obs
    %2 = quantum.sample %1 : tensor<1000x1xf64>
    return %2 : tensor<1000x1xf64>
  }
```

Key changes are:

 - We add a new Pattern, `AddPostProcessingPattern`, that inserts a classical function that calls the quantum_node for each in the module. This should be generic for re-use with other 1:1 passes that add post-processing (but won't work for passes that create multiple quantum.nodes meant to be recombined in post-processing) 
 - We update the `match_and_rewrite` functions to look for `FuncOp`s instead of measurement processes, and apply the pattern to each quantum.node
 - We add a method on the main `MeasurementsFromSamples` pattern that gets the module the `quantum.node` is nested in, walks through it, and finds the CallOp for the `quantum.node`, so that it is available as in insertion point for post-processing. 
 - We add an `update_returns` function that updates the operations and types returned in the outer function and the qnode to reflect the conversion to samples an the addition of a post-processing step
 - We no longer assume that the shots are the same for all `quantum.node`s - this is technically true currently, because weighted shot distribution for `split-non-commuting` isn't added yet, but its not a safe assumption to make.

**Benefits:**
More future-facing technical implementation (follows new convention, won't break when shot distribution is added). Will be easier to maintain and improves readability of the IR (especially when there are multiple measurements output). Maintains cleaner separation between hybrid/quantum execution and purely classical post-processing.

**Possible Drawbacks:**
Not sure if/how these changes will impact performance. We previously "matched" directly with the MPs in the pattern rewriter, and now we match with all FuncOps, check if they are `quantum.nodes`, and then walk through that to find the MPs. I'm not sure whether this will impact performance meaningfully. 

Note that this will need to be moved from xDSL to MLIR eventually, so in terms of efficiency, "good enough for now" is fine.

**Related GitHub Issues:**
None

[sc-117793]


frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -624,12 +812,12 @@ def _postprocessing_expval(samples, column):
     Returns:
         jax.core.ShapedArray: The expectation value for each requested column.
     """
-    return jnp.mean(1.0 - 2.0 * samples[:, column], axis=0)
+    return jnp.mean(1.0 - 2.0 * samples[:, 0], axis=0)
```
lillian542, line 707: We just hardcode this to 0 for now because it makes things easier - this will be changed to a more general function that can handle multi-wire observables based on eigenvalues in the next PR

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -81,6 +85,54 @@ def apply(self, _ctx: context.Context, op: builtin.ModuleOp) -> None:
...
+        func_op.sym_name = builtin.StringAttr(qnode_name)
+        rewriter.notify_op_modified(func_op)
+
+        # create the outer (postprocessing) fn with the original node name
+        func_type = func_op.function_type
+        outer_fn = func.FuncOp(name=outer_fn_name, function_type=func_type, visibility="public")
+        rewriter.insert_op(outer_fn, InsertPoint.before(func_op))
+
+        # call the renamed quantum_node inside the new outer FuncOp
+        call_args = outer_fn.body.block.args
```
multiphaseCFD: Could we assign `call_args` as `func_type.inputs.data`?

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -81,6 +85,54 @@ def apply(self, _ctx: context.Context, op: builtin.ModuleOp) -> None:
 measurements_from_samples_pass = compiler_transform(MeasurementsFromSamplesPass)
 
 
+class AddPostProcessingPattern(RewritePattern):
+    """RewritePattern for making each Qnode intro a classical function that
```
```suggestion
    """RewritePattern for making each Qnode into a classical function that
```

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -81,6 +85,54 @@ def apply(self, _ctx: context.Context, op: builtin.ModuleOp) -> None:
...
+
+    def __init__(self, pass_str: str):
+        super().__init__()
+        self._pass_str = pass_str
+
+    def match_and_rewrite(
+        self, func_op: func.FuncOp, rewriter: pattern_rewriter.PatternRewriter, /
+    ):
+        """Transform a quantum node (qnode) to separate it into a quantum node
+        and an outer function that calls the quantum node. Subsequent patterns can the apply
```
```suggestion
        and an outer function that calls the quantum node. Subsequent patterns can then apply
```

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -378,71 +494,86 @@ class ExpvalAndVarPattern(MeasurementsFromSamplesPattern):
...
+            for op in self.qnode.body.walk()
+            if isinstance(op, (quantum.VarianceOp, quantum.ExpvalOp))
+        ]
+
+        for matched_op in measurement_processes:
+
+            observable_op = self.get_observable_op(matched_op)
+            in_qubit = observable_op.operands[0]
+            compbasis_op = self.insert_compbasis_op(in_qubit, observable_op, rewriter)
+            sample_op = self.insert_sample_op(compbasis_op, self._shots, 1, rewriter)
```
multiphaseCFD, line 459: Any reason to hardcode the `n_qubits` arg as `1`? Would it be the len of in_qubits?

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_measurements_from_samples_integration.py
```diff
@@ -90,6 +187,36 @@ def circuit():
...
+        dev = qml.device("lightning.qubit", wires=2)
+
+        @qml.qjit
+        @qml.transform(pass_name="measurements-from-samples")
+        @qml.transform(pass_name="split-non-commuting")
+        @qml.qnode(dev, shots=shots)
+        def circuit():
+            initial_ops[0](wires=0)
+            initial_ops[1](wires=1)
+            return qml.expval(2 * qml.Z(wires=0))
```
multiphaseCFD, line 217: Would `split-non-commuting` change the IR?

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -90,19 +142,44 @@ class MeasurementsFromSamplesPattern(RewritePattern):
...
-        assert isinstance(
-            shots, int
-        ), f"Expected `shots` to be an integer value but got {type(shots).__name__}"
-        if shots == 0:
-            raise ValueError("The measurements_from_samples pass requires non-zero shots")
-        self._shots = shots
+
+        self._shots = None
+        self.qnode: func.FuncOp = None
+        self.call_op: func.CallOp = None
```
joeycarter: Minor point, but if we're initializing to `None`, then by definition the type can also be `None`

```suggestion
        self.qnode: func.FuncOp | None = None
        self.call_op: func.CallOp | None = None
```

To be on the safe side, we could add (for example)

```python
assert self.qnode is not None
```

before using these member variables in any `MeasurementsFromSamplesPattern` method (but not necessary after assignment in any of the derived classes).

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -368,6 +445,45 @@ def _walk_back_to_alloc_op(
...
+        Args:
+            mp_index (int): The index of the measurement process to be updated. The
+                relevant function returns will be updated at this index.
+            sample_op (quantum.SampleOp): The operation whose results the Qnode should return.
+            postprocessing_func_call_op (func.CallOp): The postprocessing CallOp whose results
+                the outer function should return.
+            rewriter (PatternRewriter): The xDSL pattern rewriter.
+        """
+        # update the qnode to return the result of the SampleOp directly
+        return_op = self.qnode.get_return_op()
```
joeycarter, line 405: Here, for example:

```suggestion
        assert self.qnode is not None
        return_op = self.qnode.get_return_op()
```

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -90,19 +142,44 @@ class MeasurementsFromSamplesPattern(RewritePattern):
...
+        """Get the first ancestral builtin.ModuleOp op of a given func.func op."""
+        while (op := op.parent_op()) and not isinstance(op, builtin.ModuleOp):
+            pass  # pragma: no cover
+        if op is None:
+            raise RuntimeError(  # pragma: no cover
+                "The given qnode func is not nested within a builtin.module. Please ensure the "
+                "qnode func is defined in a builtin.module."
+            )
+
+        return op
```
joeycarter: [Optional] I find this a little easier to read (and also makes my type-checker happy). I've also swapped out the `RuntimeError` for a `CompileError`.


```suggestion
    def _get_parent_module(self, op: func.FuncOp) -> builtin.ModuleOp:
        """Get the first ancestral builtin.ModuleOp op of a given func.func op."""
        _op: Operation | None = op
        while _op := _op.parent_op():
            if isinstance(_op, builtin.ModuleOp):
                break
            if _op is None:
                raise CompileError("...")

        assert isinstance(_op, builtin.ModuleOp)
        return _op
```

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -90,19 +142,44 @@ class MeasurementsFromSamplesPattern(RewritePattern):
...
+
+    def _get_call_op(self, qnode):
```
```suggestion
    def _get_call_op(self, qnode: func.FuncOp):
```

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -90,19 +142,44 @@ class MeasurementsFromSamplesPattern(RewritePattern):
...
+        """Get the CallOp in another module function that calls this quantum_node. Postprocessing
+        will be called to act on the output of that CallOp"""
+
+        module = self._get_parent_module(qnode)
+        qnode_name = qnode.sym_name.data
+        all_call_ops = [op for op in module.body.walk() if isinstance(op, func.CallOp)]
+        qnode_call_op = [op for op in all_call_ops if qnode_name in op.callee.string_value()]
+        if not len(qnode_call_op) == 1:
+            raise CompileError(
+                f"Expected only one call_op for {qnode_name}, but recieved {qnode_call_op}"
```
```suggestion
                f"Expected only one call_op for {qnode_name}, but received {qnode_call_op}"
```

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -368,6 +445,45 @@ def _walk_back_to_alloc_op(
...
+            mp_index (int): The index of the measurement process to be updated. The
+                relevant function returns will be updated at this index.
+            sample_op (quantum.SampleOp): The operation whose results the Qnode should return.
+            postprocessing_func_call_op (func.CallOp): The postprocessing CallOp whose results
+                the outer function should return.
+            rewriter (PatternRewriter): The xDSL pattern rewriter.
+        """
+        # update the qnode to return the result of the SampleOp directly
+        return_op = self.qnode.get_return_op()
+        return_op.operands[mp_index] = sample_op.results[0]
```
joeycarter, line 407: [Optional] Nice to have since as a sanity check since `get_return_op()` can return `None`.

```suggestion
        assert return_op is not None, "QNode has no return op"
        return_op.operands[mp_index] = sample_op.results[0]
```

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -368,6 +445,45 @@ def _walk_back_to_alloc_op(
 
         return n_qubits
 
+    def update_returns(self, mp_index, sample_op, postprocessing_func_call_op, rewriter):
```
joeycarter: Mind adding type hints to the function arguments?

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -378,71 +494,86 @@ class ExpvalAndVarPattern(MeasurementsFromSamplesPattern):
...
+            # from its uses get the index this result is returned at
+            tensor_op = list(matched_op.results[0].uses)[0].operation
+            assert isinstance(
+                tensor_op, tensor.FromElementsOp
+            ), f"Expected to find a tensor.from_elements op, but got {type(tensor_op).__name__}"
+            mp_index = list(tensor_op.results[0].uses)[0].index
+
+            # Insert the call to the post-processing function
+            postprocessing_func_call_op = func.CallOp(
+                callee=builtin.FlatSymbolRefAttr(postprocessing_func_op.sym_name),
```
joeycarter: I just realize that instead of `FlatSymbolRefAttr`, we should use `SymbolRefAttr`:

```suggestion
                callee=builtin.SymbolRefAttr(postprocessing_func_op.sym_name),
```

`FlatSymbolRefAttr` is meant as a type hint, whereas `SymbolRefAttr` is the constructable type.

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -454,51 +585,81 @@ class ProbsPattern(MeasurementsFromSamplesPattern):
...
+                    jax.core.ShapedArray([self._shots, n_qubits], float)
+                )
+
+                postprocessing_func_op = self.get_postprocessing_funcs_from_module_and_insert(
+                    postprocessing_module, probs_op, postprocessing_func_name
+                )
+
+            # Insert the call to the post-processing function
+            postprocessing_func_call_op = func.CallOp(
+                callee=builtin.FlatSymbolRefAttr(postprocessing_func_op.sym_name),
```
joeycarter: Same here:

```suggestion
                callee=builtin.SymbolRefAttr(postprocessing_func_op.sym_name),
```

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -454,51 +585,81 @@ class ProbsPattern(MeasurementsFromSamplesPattern):
...
-            callee=builtin.FlatSymbolRefAttr(postprocessing_func_op.sym_name),
-            arguments=[sample_op.results[0]],
-            return_types=[builtin.TensorType(builtin.Float64Type(), shape=(2**n_qubits,))],
-        )
+            # get the index the probs MP result is returned at
+            result_index = list(probs_op.results[0].uses)[0].index
+
+            # Insert the call to the post-processing function
+            postprocessing_func_call_op = func.CallOp(
+                callee=builtin.FlatSymbolRefAttr(postprocessing_func_op.sym_name),
```
joeycarter: And here:

```suggestion
                callee=builtin.SymbolRefAttr(postprocessing_func_op.sym_name),
```

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -454,51 +585,81 @@ class ProbsPattern(MeasurementsFromSamplesPattern):
...
+            rewriter.insert_op(
+                postprocessing_func_call_op, insertion_point=InsertPoint.after(self.call_op)
+            )
+
+            # update the returns of the QNode (to return raw samples) and the
+            # outer function (to return post-processed values)
+            self.update_returns(result_index, sample_op, postprocessing_func_call_op, rewriter)
 
-        rewriter.replace_matched_op(postprocessing_func_call_op)
+            # delete now ununsed probs_op
```
```suggestion
            # delete now unused probs_op
```

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -378,71 +494,86 @@ class ExpvalAndVarPattern(MeasurementsFromSamplesPattern):
...
-        ), f"Expected to replace a tensor.from_elements op, but got {type(op_to_replace).__name__}"
-        rewriter.replace_op(op_to_replace, postprocessing_func_call_op)
+            # update the returns of the QNode (to return the raw samples) and the
+            # outer function (to return the post-processed values)
+            self.update_returns(mp_index, sample_op, postprocessing_func_call_op, rewriter)
 
-        # Finally, erase the expval/var op and its associated observable op
-        rewriter.erase_op(matched_op)
-        rewriter.erase_op(observable_op)
+            # delete now ununsed obs --> mp --> tensor chain
```
```suggestion
            # delete now unused obs --> mp --> tensor chain
```

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -454,51 +585,81 @@ class ProbsPattern(MeasurementsFromSamplesPattern):
...
-        ), "Unable to determine number of qubits from quantum.compbasis op"
+        probs_ops = [op for op in self.qnode.body.walk() if isinstance(op, quantum.ProbsOp)]
 
-        sample_op = self.insert_sample_op(compbasis_op, self._shots, n_qubits, rewriter)
+        for probs_op in probs_ops:
 
-        # Insert the post-processing function into current module or
-        # get handle to it if already inserted
-        postprocessing_func_name = f"probs_from_samples.tensor.{self._shots}x{n_qubits}xf64"
+            compbasis_op = probs_op.operands[0].owner
```
joeycarter, line 543: Good sanity check to have since the owner of an operand can also be a block argument:

```suggestion
            compbasis_op = probs_op.operands[0].owner
            assert isinstance(compbasis_op, quantum.ComputationalBasisOp)
```

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -537,33 +698,60 @@ def match_and_rewrite(self, state_op: quantum.StateOp, rewriter: PatternRewriter
...
+        CompileError: If the operator expected to contain shots values does not have `properties`.
+            This is the immediate issue that is observed when shots are dynamic.
+        TypeError: if the extracted shots are not an int
+        ValueError: if the extracted shots are zero
+
+    """
+
+    shots = _get_static_shots_value_from_device_op(quantum_node)
+    if not isinstance(shots, int):
+        raise TypeError(f"Expected `shots` to be an integer value but got {type(shots).__name__}")
```
joeycarter: The type hint on `_get_static_shots_value_from_device_op()` says that it always returns an `int`, so maybe this exception would be better as an assert (i.e. a developer assumption has been violated rather than a user error).

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_measurements_from_samples_integration.py
```diff
@@ -65,6 +65,103 @@ def circuit(x):
...
+                qml.RX(theta, 0)
+                return qml.counts()
+
+    @pytest.mark.parametrize("mp", (qml.expval, qml.var))
+    def test_overlapping_tensor(self, mp):
+        """Check that an error is raised if the circuit returns a tensor with overlapping wires."""
+
+        # Note: This error is raised by the diagonalize pass that measurements_from_samples
+        # calls, not by measurements_from_samples directly. However, the logic in this pass
+        # relies on the validation being performed, so its tested here. If this test ever breaks
```
joeycarter:
```suggestion
        # relies on the validation being performed, so it's tested here. If this test ever breaks
```

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_measurements_from_samples_lit.py
```diff
@@ -53,8 +53,12 @@ def test_1_wire_expval(self, run_filecheck):
         program = """
         builtin.module @module_circuit {
             // CHECK-LABEL: circuit
-            func.func public @circuit() -> (tensor<f64>) {
-                %0 = "stablehlo.constant"() <{value = dense<1> : tensor<i64>}> : () -> tensor<i64>
+            // CHECK: [[samples:%.+]] = func.call @circuit.from_samples() : () -> tensor<5x1xf64>
+            // CHECK: [[res:%.+]] = func.call @expval_from_samples.tensor.5x1xf64([[samples]]) :
+            // CHECK-SAME: (tensor<5x1xf64>) -> tensor<f64>
+            // CHECK: func.return [[res]] : tensor<f64>
+            func.func public @circuit() -> (tensor<f64>) attributes {quantum.node}  {
```
joeycarter, line 61: Not strictly necessary, but probably a good check to have to ensure that we still have a `quantum.node` func op with the correct name, and the following checked ops are contained within it.

```suggestion
            // CHECK: func.return [[res]] : tensor<f64>
            // CHECK: func.func public @circuit.from_samples{{.*}} attributes {quantum.node}
            func.func public @circuit() -> (tensor<f64>) attributes {quantum.node}  {
```

If you want to put this check in here we should add it to all the other lit tests as well.

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_measurements_from_samples_lit.py
```diff
@@ -53,8 +53,12 @@ def test_1_wire_expval(self, run_filecheck):
         program = """
         builtin.module @module_circuit {
             // CHECK-LABEL: circuit
-            func.func public @circuit() -> (tensor<f64>) {
-                %0 = "stablehlo.constant"() <{value = dense<1> : tensor<i64>}> : () -> tensor<i64>
+            // CHECK: [[samples:%.+]] = func.call @circuit.from_samples() : () -> tensor<5x1xf64>
```
joeycarter: Not a big problem, but we've already defined the `samples` variable below on L72.

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_measurements_from_samples_lit.py
```diff
@@ -299,46 +307,47 @@ def test_1_wire_counts(self, run_filecheck):
...
         program = """
         builtin.module @module_circuit {
             // CHECK-LABEL: circuit
-            func.func public @circuit() -> (tensor<f64>) {
-                %0 = "stablehlo.constant"() <{value = dense<1> : tensor<i64>}> : () -> tensor<i64>
+            // CHECK: [[samples0:%.+]], [[samples1:%.+]] = func.call @circuit.from_samples() : () -> (tensor<5x1xf64>, tensor<5x1xf64>)
+            // CHECK: [[res1:%.+]] = func.call @expval_from_samples.tensor.5x1xf64([[samples1]]) :
+            // CHECK-SAME: (tensor<5x1xf64>) -> tensor<f64>
+            // CHECK: [[res0:%.+]] = func.call @expval_from_samples.tensor.5x1xf64([[samples0]]) :
+            // CHECK-SAME: (tensor<5x1xf64>) -> tensor<f64>
```
joeycarter, line 251: This isn't really an issue, but I noticed that the post-processing func ops are inserted in reverse order, i.e. it post-processes `samples1` and then `samples0`. Any concerns with this?

joeycarter (commented):
Thanks @lillian542, nice work! 🚀 

The new design of the pass looks great, I only had a few minor points and suggestions.

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
mudit2812:
```suggestion
    """RewritePattern for making each QNode into a classical function that
```

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_measurements_from_samples_lit.py
lillian542: Yeah... I don't love it, I didn't think about it until I was writing the lit tests and ran into it, but when we process the MPs in order and insert every post processing function after the CallOp, they end up getting switched. 

I do think it would improve readability if they came in order. I'll try and see if just reversing the order of the list of collected MPs fixes this as simply as I think it would, or if there are complications.

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -378,71 +494,86 @@ class ExpvalAndVarPattern(MeasurementsFromSamplesPattern):
...
+        measurement_processes = [
+            op
+            for op in self.qnode.body.walk()
+            if isinstance(op, (quantum.VarianceOp, quantum.ExpvalOp))
+        ]
+
+        for matched_op in measurement_processes:
+
+            observable_op = self.get_observable_op(matched_op)
+            in_qubit = observable_op.operands[0]
```
mudit2812, line 457: Are we assuming that only single-qubit observables are supported? What about `quantum.tensor`, `quantum.hamiltonian`?

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -378,71 +494,86 @@ class ExpvalAndVarPattern(MeasurementsFromSamplesPattern):
...
-            arguments=[sample_op.results[0], column_index_op],
-            return_types=[builtin.TensorType(builtin.Float64Type(), shape=())],
-        )
+            # get the tensor_op the original MP result is passed to
+            # from its uses get the index this result is returned at
+            tensor_op = list(matched_op.results[0].uses)[0].operation
+            assert isinstance(
+                tensor_op, tensor.FromElementsOp
+            ), f"Expected to find a tensor.from_elements op, but got {type(tensor_op).__name__}"
+            mp_index = list(tensor_op.results[0].uses)[0].index
```
mudit2812: I would prefer to use a different name for this variable instead of `tensor_op`. I initially thought that it was referring to a `quantum.tensor` and got pretty confused. Perhaps `from_elts_op`?

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -454,51 +585,81 @@ class ProbsPattern(MeasurementsFromSamplesPattern):
...
+            compbasis_op = probs_op.operands[0].owner
 
-        postprocessing_func_op = self.get_postprocessing_func_op_from_block_by_name(
-            probs_op.parent_op().parent, postprocessing_func_name
-        )
+            n_qubits = None
+            if compbasis_op.qreg is not None:
+                n_qubits = self.get_n_qubits_from_qreg(compbasis_op.qreg)
+
+            elif not compbasis_op.qubits == ():
```
mudit2812:
```suggestion
            elif compbasis_op.qubits != ():
```

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
mudit2812: No dynamic shots support?

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
mudit2812: Do you mind adding a comment to `ExpvalAndVarPattern`'s doctring (or a dev comment) indicating that only single-wire observables are currently supported?

multiphaseCFD (commented):
Thanks Lillian. A few thoughts

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
lillian542: Currently, yes. This PR just refactors the initial prototype to follow the new post-processing guidelines, and the initial prototype only works for single-qubit observables. 

The next PR will add `tensor` support. 

We will require `split_non_commuting` to separate hamiltonians into terms before this pass with capture enabled, so `hamiltonian` support will not be added, but a clear error message will.

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
lillian542: No, and none is planned at this time

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
lillian542: Good point, I'll rename it!

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
lillian542: It's just what we did initially, since we were hardcoding it to a single wire in the first implementation. It's copy-pasted from main. 

It's updated to `n_qubits := len(in_qubits)` in the next PR, where we make it possible to have `n_qubits != 1`.

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_measurements_from_samples_integration.py
lillian542: Good question, and yes. We aren't really relying on the grouping/splitting of observables based on commutation here, but we are relying on the steps that `split-non-commuting` goes through to make that possible: it makes everything lists of observables and post-processing, without any sums or coefficients in the observable definitions.

Anything with a coefficient becomes a `quantum.hamiltonian` in the IR, even if its a scalar product rather than a sum, so the initial IR for the above includes:

```
%2 = quantum.namedobs %1[ PauliZ] : !quantum.obs
%3 = stablehlo.convert %arg0 : (tensor<1xi64>) -> tensor<1xf64>
%4 = quantum.hamiltonian(%3 : tensor<1xf64>) %2 : !quantum.obs
%5 = quantum.expval %4 : f64
```
where %arg0 is the 2.

Even when there are no non-commuting observables to split into separate executions, applying `split-non-commuting` still splits all hamiltonians up into observables and post-processing. So this will become a `quantum.node` function that includes

```
%2 = quantum.namedobs %1[ PauliZ] : !quantum.obs
%3 = quantum.expval %2 {group = 0 : i64} : f64
```

that is called in a post-processing function that multiplies the results of the QNode and the coefficient:

```
%2 = stablehlo.multiply %from_elements, %1 : tensor<f64>
```

This we can handle with `measurements-from-samples`, whereas the `quantum.hamiltonian` is something we haven't added support for.

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_measurements_from_samples_lit.py
lillian542:
```suggestion
```

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
lillian542: I'm having second thoughts about these existing at all. There was a previous version of this where they were used more often, but looking at it now that things are cleaned up a bit, they are actually only used in `match_and_rewrite` (where they are assigned) and in this method.

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_measurements_from_samples_lit.py
lillian542: I'm going to wait and try to resolve this in the final PR - it will make more sense there, since we're combining the measurement patterns. Here, even if we change the order, there will be weird ordering if there are multiple MPs returned.

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_measurements_from_samples_lit.py
lillian542: added!

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
lillian542: I'm going to add these asserts for now, and consider whether refactoring to remove the attributes entirely makes sense in the final PR.

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_measurements_from_samples_lit.py
lillian542: updated to `raw_samples` since they will be post-processed

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -368,81 +445,148 @@ def _walk_back_to_alloc_op(
...
     Args:
         shots (int): The number of shots (e.g. as retrieved from the DeviceInitOp).
     """
 
     @pattern_rewriter.op_type_rewrite_pattern
-    def match_and_rewrite(
-        self, matched_op: quantum.ExpvalOp | quantum.VarianceOp, rewriter: PatternRewriter, /
-    ):
+    def match_and_rewrite(self, func_op: func.FuncOp, rewriter: PatternRewriter, /):
         """Match and rewrite for quantum.ExpvalOp."""
```
multiphaseCFD: Match and rewrite for quantum.ExpvalOp and quantum.VarianceOp.

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -368,81 +445,148 @@ def _walk_back_to_alloc_op(
...
-            postprocessing_module = postprocessing_jit_func(
-                jax.core.ShapedArray([self._shots, 1], float), 0
+        if "quantum.node" not in func_op.attributes:
+            return
+
+        self._shots = get_shots(func_op)
+        self.qnode = func_op
+        self.call_op = self._get_call_op(func_op)
+
+        measurement_processes = [
```
multiphaseCFD, line 448: Not a blocker for now. Looks like we need to transverse the same qnode func twice if there are a mixed of `quantum.ExpvalOp` and `quantum.ProbOp` in the func.func, right? We might come back to this once we port this pass to the MLIR.

multiphaseCFD (approved):
LGTM! Thanks @lillian542 .

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
lillian542: This is fixed in the next PR, which should be ready for review soon :)

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
lillian542: Had to put this back to how it was because it caused test failures, though I don't understand why 😞

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -81,6 +85,54 @@ def apply(self, _ctx: context.Context, op: builtin.ModuleOp) -> None:
 measurements_from_samples_pass = compiler_transform(MeasurementsFromSamplesPass)
 
 
+class AddPostProcessingPattern(RewritePattern):
```
mudit2812: This pattern doesn't seem to be specific to measurements_from_samples. Could you move it to its own file in the `transforms` folder?

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
multiphaseCFD: I believe the failures come from the mismatch between function_type of the circuit (qnode) function and post-processing function_type

mudit2812: `func_type` is an attribute that just tells us the type of all the function inputs and outputs, it doesn't actually contain the actual arguments/results. `outer_fn.body.block.args` are the actual SSA values for the arguments. There shouldn't be any mismatch between the function type of the circuit function and the outer function

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
mudit2812: Also, I found the name of this pattern to be a bit confusing. It's not adding postprocessing, rather just structuring the QNode so that we _can_ add postprocessing later

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -90,19 +143,44 @@ class MeasurementsFromSamplesPattern(RewritePattern):
...
         """Abstract method for measurements-from-samples match-and-rewrite patterns."""
 
+    def _get_parent_module(self, op: func.FuncOp) -> builtin.ModuleOp:
+        """Get the first ancestral builtin.ModuleOp op of a given func.func op."""
+        _op: ir.Operation | None = op
+        while _op := _op.parent_op():
+            if isinstance(_op, builtin.ModuleOp):
+                break
+            if _op is None:
+                raise CompileError("...")
```
mudit2812:
```suggestion
                raise CompileError(f"{op} does not belong to a module.")
```

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -368,81 +446,148 @@ def _walk_back_to_alloc_op(
...
+        mp_index: int,
+        sample_op: quantum.SampleOp,
+        postprocessing_func_call_op: func.CallOp,
+        rewriter: PatternRewriter,
+    ):
+        """Update the return structure so that the qnode returns the output of sample
+        instead of the original output, and the outer function returns the output of
+        post-processing instead of directly returning the output of calling the qnode.
+
+        Also update result shapes in the CallOp for the Qnode call to reflect this change,
```
mudit2812:
```suggestion
        Also update result shapes in the CallOp for the QNode call to reflect this change,
```

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -368,81 +446,148 @@ def _walk_back_to_alloc_op(
...
+        since this is not updated when modifying the ReturnOps.
+
+        This function updates the self.qnode and self.call_op attributes on the Pattern.
+
+        Args:
+            mp_index (int): The index of the measurement process to be updated. The
+                relevant function returns will be updated at this index.
+            sample_op (quantum.SampleOp): The operation whose results the Qnode should return.
+            postprocessing_func_call_op (func.CallOp): The postprocessing CallOp whose results
+                the outer function should return.
```
mudit2812: The argument name seems like this is a call to the postprocessing function. Maybe `qnode_call_op` might be better?

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
mudit2812: You can basically replace this whole thing with `catalyst.python_interface.utils.get_constant_from_ssa`. If the shots are dynamic, it will return `None`, otherwise a Python int.

mudit2812 (commented):
This looks great Lillian. A few final comments. I also wanted to ask, is there any plan to support dynamic shots? Seems like a pretty important feature if we want to link this pass to the pennylane transform

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
lillian542: Good point, yes, I'll make it its own file and think about a better name 😅

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
lillian542: It is a call to the post-processing function. The `self.call_op` is the op that calls the QNode. 

I.e. at the point where we call this, in the "main" (now classical) function, we have:

```
func.func public @circ_0
   %qnode_res = call @circ.from_samples
   %processed_res = call @postprocessing_fn %qnode_res
   return %qnode_res
```

and we want to update it so that the QNode returns raw samples, update the return to give us `%processed_res`, which we get from the `postprocessing_func_call_op`.

I'll make the docstring a little more detailed and try to make it a bit clearer what's happening here. Its kind of convoluted.

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
lillian542: Thanks! I made a tiny adjustment to `get_constant_from_ssa` to support one test-case that failed when I swapped these out, but I think it's uncontroversial :)

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
mudit2812: Ah, I misunderstood. Thanks for clarifying. Resolving this comment.

frontend/catalyst/python_interface/utils.py
```diff
@@ -57,7 +57,7 @@ def get_constant_from_ssa(value: SSAValue) -> Number | None:
     if not isinstance(value.type, ShapedType):
         owner = value.owner
 
-        if isinstance(owner, arithConstantOp):
+        if isinstance(owner, (arithConstantOp, hloConstantOp)):
             const_attr = owner.value
             return const_attr.value.data
```
mudit2812: The branch below already handles `stablehlo.constant`. Did it not work properly?

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
lillian542: I've renamed it as well as I could (`WrapQNodePattern`), but I'm very open to suggestions for other names

frontend/catalyst/python_interface/utils.py
lillian542: Without this `test_1_wire_expval_shots_from_stablehlo_constop` in the lit tests failed. The test uses a StableHLO constant, but not inside a `tensorExtractOp`. But maybe this is an issue with the lit-test being written incorrectly? Are StableHLO constants always in `tensorExtractOp`s?

mudit2812: `stablehlo.constant` always returns a tensor according to its definition, so if we want a scalar constant from stablehlo, it will always follow the pattern of `stablehlo.constant` + `tensor.extract`. So, my recommendation is to update the failing test rather than this function.

lillian542: Honestly the test is maybe superfluous at this point, since we aren't using functionality implemented here to get the shots anymore, we're using the utility function. I'll revert this and change the tests.

lillian542:
> I also wanted to ask, is there any plan to support dynamic shots? Seems like a pretty important feature if we want to link this pass to the pennylane transform

When we were scoping it early, my impression was we are alright with figuring out dynamic shots later, but this would definitely be a good thing to revisit before we link the tape transform.

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -454,51 +526,82 @@ class ProbsPattern(MeasurementsFromSamplesPattern):
...
-        postprocessing_func_name = f"probs_from_samples.tensor.{self._shots}x{n_qubits}xf64"
+            compbasis_op = probs_op.operands[0].owner
+            assert isinstance(compbasis_op, quantum.ComputationalBasisOp)
 
-        postprocessing_func_op = self.get_postprocessing_func_op_from_block_by_name(
-            probs_op.parent_op().parent, postprocessing_func_name
-        )
+            n_qubits = None
+            if compbasis_op.qreg is not None:
+                n_qubits = self.get_n_qubits_from_qreg(compbasis_op.qreg)
```
multiphaseCFD, line 548: Would it be better add instance variable `self._num_qubits = None`? With that, we could just assign value to it only once.

mudit2812 (commented):
Looks great. My only blocking comment is about removing the change to `get_constant_from_ssa` and updating any affected tests accordingly.

frontend/catalyst/python_interface/utils.py
joeycarter: This is my bad, I hadn't realized that `stablehlo.constant` always returns a tensor. I was tripped up because it's valid to write

```mlir
%0 = "stablehlo.constant"() <{value = 4 : i64}> : () -> i64
```

but this is a generic MLIR op and not a valid `stablehlo.constant` op. Please go ahead and revert this change+test!

lillian542: Updated!

mudit2812: For reference, [here](https://github.com/openxla/stablehlo/blob/a93085ced0919a8b0071f54b431a8ed8c9c12c75/stablehlo/dialect/StablehloOps.td#L75-L107) is the TableGen definition in the stablehlo repo, and this is the [spec](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#constant)

mudit2812 (approved):
Nice. Customary 🚀

PR #2610, opened 2026-03-20
xdsl adjoint op can take in qreg and qubit

paul0403:
**Context:**
Update xdsl `quantum.adjoint` operation's type constraints, so it can accept `Variadic<AnyTypeOf<Qubit, Qureg>>`, and understand that the operands and results have the exact same type.

**Benefits:**
Agreement with mlir `quantum.adjoint1 op.

This is a follow-up to #2590


mudit2812 (approved):
🎉

dime10 (approved):

PR #2616, opened 2026-03-20
The measurements-from-samples pass raises a clear error when shots are dynamic

lillian542:
**Context:**
One of several PRs for [sc-107541]. This one addresses having a clear error message in the case of dynamic shots.

Currently, a cryptic error is raised for a circuit with a dynamic number of shots if `measurments-from-samples` pass is applied:

```
@qml.qjit
def workflow(a, shots):

    @measurements_from_samples_pass
    @qml.set_shots(shots)
    def circuit(x):
        qml.RX(x, 0)
        return qml.expval(qml.Z(0))
    
    return circuit(a)
```

```
>>> workflow(1.2, 100)
... 
AttributeError: 'Block' object has no attribute 'properties'
```

We catch this and add a more useful error message.

**Benefits:**
Less confusing.

**Possible Drawbacks:**
I'm not sure how to directly check if the shots are dynamic, so I've based it on what concretely goes wrong at this point, which seems like not the most robust choice.

**Related GitHub Issues:**
None

[sc-117790]


multiphaseCFD (approved):
LGTM @lillian542 . Don't forget to add an entry to the changelog

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -583,6 +590,11 @@ def _get_static_shots_value_from_first_device_op(module: builtin.ModuleOp) -> in
         return shots_int_values[0]
 
     if isinstance(shots_extract_op, arith.ConstantOp):
+        if not hasattr(shots_extract_op, "properties"):
+            raise CompileError(
+                "Cannot get number of shots. Note that using a dynamic number of shots is not "
+                "supported with measurements-from-samples."
+            )
```
joeycarter: We shouldn't need this check here, since we've already established that `shots_extract_op` is an `arith.ConstantOp` in this branch, so it should always have a `properties` field with a `"value"` attribute.

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -551,6 +551,8 @@ def _get_static_shots_value_from_first_device_op(module: builtin.ModuleOp) -> in
 
     Raises:
         CompileError: If `module` does not contain a quantum.DeviceInitOp.
+        CompileError: If the operator expected to contain shots values does not have `properties`.
+            This is the immediate issue that is observed when shots are dynamic.
```
```suggestion
        CompileError: If the operator expected to contain shots values does not have `properties`.
            This is the immediate issue that is observed when shots are dynamic, for instance as a
            result of the shots SSA value originating from a block argument rather than from an
            operation, such as an `arith.constant` op.
```

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -583,6 +590,11 @@ def _get_static_shots_value_from_first_device_op(module: builtin.ModuleOp) -> in
         return shots_int_values[0]
 
     if isinstance(shots_extract_op, arith.ConstantOp):
```
joeycarter: [Optional] This isn't necessary, but if it's any easy thing to add, we could also support the case where the constant op is a `stablehlo.constant` op. For example, we should be able to pass this test case:

```python
def test_stablehlo_constant(run_filecheck):
    program = """
        // CHECK-LABEL: circuit
        func.func public @circuit() -> (tensor<f64>) {
            %0 = "stablehlo.constant"() <{value = 1 : i64}> : () -> i64
            quantum.device shots(%0) ["", "", ""]
            return
        }
    }
    """

```

similar to what we have for `test_1_wire_expval_shots_from_arith_constantop`.

joeycarter (changes requested):
Thanks @lillian542! Just a few suggestions, happy to approve once resolved.

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
lillian542: Ah, right - thanks!

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
lillian542: I'm having trouble identifying how to find documentation for or use the StableHLO dialect with xDSL.

joeycarter: Ah, the StableHLO dialect was upstreamed in #2591, you can import the package and get the `constant` op with

```python
from xdsl_jax.dialects import stablehlo

stablehlo.ConstantOp
```

Not urgent to get this in in this PR, though, if you want to merge this now—we can come back to it.

joeycarter (approved):
💯

PR #2617, opened 2026-03-20
Add diagonalization in `measurements-from-samples`

lillian542:
**Context:**
One of several PRs for [[sc-107541](https://app.shortcut.com/xanaduai/story/107541)]. This one addresses diagonalizing in `measurements-from-samples` so the `diagonalize-measurements` pass doesn't have to be applied separately.

This is for feature-parity with tapes, where that is the current behaviour of the `measurements_from_samples` transform.

We apply the `DiagonalizeFinalMeasurementsPass` on the module before beginning the pattern-rewrite for `measurements-from-samples`

**Benefits:**
It does the same thing as the corresponding tape transform

**Possible Drawbacks:**
Less freedom to choose your compilation pipeline - the diagonalize pass is applied even though it's not present in the list of applied passes in the initial IR.

**Related GitHub Issues:**
None

[sc-117791]


multiphaseCFD (commented):
LGTM! Could you add a test that applies both `measurements-from-samples` and `diagonalize_measurements` passes to ensure everything still works as expected?

multiphaseCFD (approved):
Nice, thanks @lillian542 ! LGTM

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_measurements_from_samples.py
```diff
@@ -886,6 +846,24 @@ def circuit():
         res = circuit()
         assert res == 1.0
 
+    @pytest.mark.usefixtures("use_capture")
+    def test_integrate_with_diagonalize(self):
```
joeycarter, line 945: Could you also add a lit test that checks the IR where diagonalize is invoked? This should do (feel free to adapt as needed):

```python
def test_expval_from_sample_with_diagonalize(run_filecheck):
    program = """
        // CHECK-LABEL: circuit
        func.func public @circuit() -> (tensor<f64>) attributes {quantum.node} {
            %0 = "stablehlo.constant"() <{value = dense<1> : tensor<i64>}> : () -> tensor<i64>
            %1 = tensor.extract %0[] : tensor<i64>
            quantum.device shots(%1) ["", "", ""]

            // CHECK: [[qreg:%.+]] = quantum.alloc
            %2 = quantum.alloc(1) : !quantum.reg

            // CHECK: [[q0:%.+]] = quantum.extract [[qreg]][0]
            %3 = quantum.extract %2[0] : !quantum.reg -> !quantum.bit

            // CHECK: [[q1:%.+]] = quantum.custom "Hadamard"() [[q0]]
            // CHECK: [[obs:%.+]] = quantum.compbasis qubits [[q1]]
            %4 = quantum.namedobs %3[PauliX] : !quantum.obs

            // CHECK: [[samples:%.+]] = quantum.sample [[obs]] : tensor<1x1xf64>
            // CHECK: [[res:%.+]] = func.call @expval_from_samples.tensor.1x1xf64
            // CHECK-NOT: quantum.expval
            %5 = quantum.expval %4 : f64
            %6 = tensor.from_elements %5 : tensor<f64>

            // CHECK: return [[res]] : tensor<f64>
            func.return %6 : tensor<f64>
        }
        // CHECK-LABEL: func.func public @expval_from_samples.tensor.1x1xf64
    }
    """

```

joeycarter (changes requested):
Looks good! I only had one small request to add a lit test—happy to approve once resolved 🚀

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_measurements_from_samples.py
lillian542: Good point, thanks! Added :)

joeycarter: Looks good, thanks!

joeycarter (approved):
💯

PR #2619, opened 2026-03-23
hard-code decomp ops

kipawaa:
**Context:**
Ops that have precompiled decomposition rules are currently listed by name and captured from PennyLane. This can be simplified and made less error-prone by using a set of PennyLane ops.

Changes the set of op names to a set of ops, and removes the logic for gathering the ops from PennyLane.

**Benefits:**
Reduced complexity of precompilation. As a side-effect, this should avoid this [flaky test failure](https://github.com/PennyLaneAI/catalyst/actions/runs/23391445367/job/68046660192).

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A


dime10 (approved):
much better, thanks

paul0403 (approved):

PR #2620, opened 2026-03-23
fix: add support for `pad_with` and `normalize` kwargs on `StatePrep`

andrijapau:
Before this PR, the `normalize` and `pad_with` kwargs wouldn't do anything when `capture` is enabled.

1. `pad_with` doesn't properly create the correct state with the correct shape and errors out
```python
@qml.qjit(capture=True)
@qml.qnode(qml.device('lightning.qubit', wires=2))
def circuit(state=None):
    qml.StatePrep(state, wires=range(2), pad_with=0.)
    return qml.expval(qml.Z(0)), qml.state()

>>> res, state = circuit(np.array([15, 15]))
>>> print(res, state)
RuntimeError: [/Users/runner/work/pennylane-lightning/pennylane-lightning/pennylane_lightning/core/simulators/lightning_qubit/StateVectorLQubit.hpp][Line:995][Method:setStateVector]: Error in PennyLane Lightning: Inconsistent state and wires dimensions.
```

2. `normalize` doesn't actually normalize the state
```python
@qml.qjit(capture=True)
@qml.qnode(qml.device('lightning.qubit', wires=2))
def circuit(state=None):
    qml.StatePrep(state, wires=range(2), normalize=True)
    return qml.expval(qml.Z(0)), qml.state()

>>> res, state = circuit(np.array([15, 15, 15, 15]))
>>> print(res, state)
0.0 [15.+0.j 15.+0.j 15.+0.j 15.+0.j]
```

With this PR we add the "preprocessing" during interpretation of the `plxpr` using tracing friendly operations.

[sc-114579]


albi3ro (commented):
changelog?

andrijapau:
@albi3ro Hmm, not sure why CC is complaining. Let me look into it.

andrijapau:
> changelog?

Done!

frontend/catalyst/from_plxpr/qfunc_interpreter.py
```diff
@@ -673,6 +673,20 @@ def handle_state_prep(self, *invals, n_wires, **kwargs):
...
     wires_inval = invals[1:]
 
+    normalize = kwargs.get("normalize", False)
+    pad_with = kwargs.get("pad_with")
+    if pad_with is not None:
+        normalize = True
+        n_states = state_inval.shape[-1]
+        dim = 2**n_wires
+        if n_states < dim:
+            pad_with = jnp.array(pad_with, dtype=state_inval.dtype)
```
albi3ro, line 683: I have no idea how this is not getting hit... I have confirmed locally that it gets hit by the test case 🤔

albi3ro (approved):
So something seems to be wrong with code cov, but I'm going to approve anyway.

albi3ro (changes requested):
Oh, I see... You need a test where you provide `path_with` but don't need it:

```
@qml.qjit(capture=True)
def c():
    qml.StatePrep(np.array([1.0, 0.0]), wires=0, pad_with=0)
    return qml.state()
```

andrijapau:
> Oh, I see... You need a test where you provide `path_with` but don't need it:
> 
> ```
> @qml.qjit(capture=True)
> def c():
>     qml.StatePrep(np.array([1.0, 0.0]), wires=0, pad_with=0)
>     return qml.state()
> ```

Good catch - added the test 🤞🏼

frontend/catalyst/from_plxpr/qfunc_interpreter.py
andrijapau: resolved

albi3ro (approved):

frontend/test/pytest/test_state_prep.py
```diff
@@ -41,6 +41,51 @@ def generate_random_state(n=1, seed=None):
...
+@pytest.mark.parametrize(
+    "state",
+    [
+        np.array([1, 1]),
+        np.array([np.sqrt(1 / 2), np.sqrt(1 / 2)]),
+        np.array([np.sqrt(1 / 2), np.sqrt(1 / 2), 0, 0]),
+    ],
+    ids=["unnormalized_input", "normalized_input", "correct_input"],
+)
+def test_state_prep_pad_with_kwarg(state, capture_mode):
```
paul0403, line 53: Don't you need to use the `use_both_frontend` fixture?

andrijapau: `@qjit(capture=False)` is the old frontend, no?

paul0403: No I mean the fixture on the test itself

paul0403: Sorry, I'm not super familiar with this new `capture_mode` fixture, what does it do?

I see it's being called with `qml.qjit(capture_mode)`, but what's the value sent into it?

andrijapau: `capture_mode` fixture is relatively new - it essentially passes `True` or `False` to the test. See [fixture](https://github.com/PennyLaneAI/catalyst/blob/fdef8320b71ba1a00eb7c58099088e7e8f1d8c08/frontend/test/pytest/conftest.py#L87) for more details.

andrijapau: We don't want to use `use_both_frontend` as that toggles the global state with `qml.capture.enable` which we are moving away from.

paul0403: Ah, so it uses both frontends just by adding the `capture_mode` argument? Without the need for the `pytest.fixture("use_both_frontend")` decorator?

andrijapau: Yep

paul0403 (approved):

PR #2623, opened 2026-03-25
Test audit/story 4 gradients v2

JerryChen97:
Gradients tests used to have a [PR](https://github.com/PennyLaneAI/catalyst/pull/2507), but it was causing too many conflicts very complicated to resolve. Hence remake

Some changes:
Considering the potential conflicts, I think we better not touching the exiting `use_both_frontend` for now. Just apply the `capture_mode` to those who haven't been tested with both.


frontend/test/pytest/test_adjoint.py
andrijapau, line 1: Why is adjoint in the gradient story? 🤔

JerryChen97: Ooops you are right adjoint does not belong here

JerryChen97: Removed

PR #2628, opened 2026-03-27
fix: use `.. code-block:: bash` for installation bash script rather than `console`

andrijapau:
**Context:**

`.. code-block:: console` should be used when you are demonstrating a session (input/output),
```console
$ echo "hi"
hi
```

Changes it to use `.. code-block:: bash` which is more reflective of what the installation bash script is.

**Benefits:** Prettier and more accurate.

**Possible Drawbacks:** None

### DIFF
--- a/doc/dev/installation.rst
+++ b/doc/dev/installation.rst
@@ -476,7 +476,7 @@ Essential Steps
 
 To activate the development environment, open a terminal and issue the following commands:
 
-.. code-block:: console
+.. code-block:: bash
 
   # Clone the Catalyst repository without submodules, as they are not needed for frontend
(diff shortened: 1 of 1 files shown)


kipawaa (approved):

joeycarter (changes requested):
Looks much nicer! ❤️ 

While we're here, there's also the "Essential Steps" section further down that should use `bash` syntax highlighting: https://xanaduai-pennylane--2628.com.readthedocs.build/projects/catalyst/en/2628/dev/installation.html#essential-steps

joeycarter (approved):
💯

PR #2633, opened 2026-03-27
update unit tests of `diagonalize_measurements` pass

multiphaseCFD:
### Before submitting

**Context:**

[sc-93668]

Unit tests for the `diagonalize_measurements` pass use `use_both_frontend` fixture instead of only `use_capture` as discussed.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
https://github.com/PennyLaneAI/catalyst/issues/2631


joeycarter (commented):
Thanks, @multiphaseCFD. There might be a more MLIR-friendly solution to this problem, which is to use [CSE](https://mlir.llvm.org/docs/Passes/#-cse) (**C**ommon **S**ub-expression **E**limination). Take the example you gave in the description; we start with

```mlir
module {
  func.func public @circuit() -> (tensor<f64>, tensor<f64>) attributes {quantum.node} {
    %2 = quantum.namedobs %1[ PauliX] : !quantum.obs
    %3 = quantum.expval %2 : f64
    %4 = quantum.extract %0[ 0] : !quantum.reg -> !quantum.bit
    %5 = quantum.namedobs %4[ PauliZ] : !quantum.obs
    %6 = quantum.var %5 : f64
    %from_elements_0 = tensor.from_elements %6 : tensor<f64>
    quantum.device_release
    return %from_elements, %from_elements_0 : tensor<f64>, tensor<f64>
  }
}
```

which has two identical `quantum.extract` ops. After running CSE, e.g. with

```console
$ quantum-opt -cse circuit.mlir
```

we get

```mlir
module {
  func.func public @circuit() -> (tensor<f64>, tensor<f64>) attributes {quantum.node} {
    %2 = quantum.namedobs %1[ PauliX] : !quantum.obs
    %3 = quantum.expval %2 : f64
    %4 = quantum.namedobs %1[ PauliZ] : !quantum.obs
    %5 = quantum.var %4 : f64
    %from_elements_0 = tensor.from_elements %5 : tensor<f64>
    quantum.device_release
    return %from_elements, %from_elements_0 : tensor<f64>, tensor<f64>
  }
}
```

The second `extract` op has been removed and its uses replaced with the results of the first `extract`.

There are a few options here, e.g. run CSE from within the `diagonalize-final-measurements` pass (it's available in xDSL as the [`CommonSubexpressionElimination`](https://github.com/xdslproject/xdsl/blob/9054fe8496b62ef51a75003e307520f48948318e/xdsl/transforms/common_subexpression_elimination.py#L261) pass), or add CSE to the list of `transform.apply_registered_pass` ops before applying `diagonalize-final-measurements`, or possibly add CSE to the default pipeline. We should consult with the Compilation team.

frontend/catalyst/python_interface/transforms/quantum/diagonalize_measurements.py
```diff
@@ -287,6 +288,9 @@ def __init__(self, **options):
 
     def apply(self, _ctx: context.Context, op: builtin.ModuleOp) -> None:
         """Apply the diagonalize final measurements pass."""
+
+        CommonSubexpressionElimination().apply(_ctx, op)
```
joeycarter: Could you add a quick comment here explaining why CSE is needed here? Adding the link to this PR should be enough, and explain that it's a workaround that can be removed once we start using reference semantics to get from PLxPR to the initial MLIR representation (work that is tracked under this epic: https://app.shortcut.com/xanaduai/epic/113514)

joeycarter (approved):
Looks good, thanks @multiphaseCFD! 🚀 

Just had one request to add a comment explaining why CSE is needed, but happy to provisionally approve.

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_diagonalize_measurements.py
```diff
@@ -881,6 +881,18 @@ def circuit(x):
...
+        """A CompileError is raised for non-commuting measurements without gates applied."""
+
+        @diagonalize_final_measurements_pass
+        @qml.set_shots(10)
+        @qml.qnode(device)
+        def circuit():
+            return qml.expval(qml.X(0)), qml.expval(qml.Z(0))
+
+        with pytest.raises(CompileError, match=_non_commuting_err_msg):
+            qml.qjit(circuit)()
```
paul0403: A non blocking comment: does the "double extract" IR fail execution? I think it doesn't actually fail execution, right? Because although double extract is malformed IR, due to the specifics of the runtime, it would still execute.

multiphaseCFD: no,  the "double extract" IR would not fail. The issue is current implementation fails to detect non-commuting for the "double extract" IR

paul0403: Yes, my point is then this added test doesn't really test the thing that was failing that this PR fixes right?

paul0403: How are xdsl passes tested? Are there any non-execution based lit tests for just checking the effect of a pass? @mudit2812

multiphaseCFD: good point. just updated the test. the update test just does compilation and  would fail

paul0403: No, I meant the double extract IR doesn't actually crash anything, so pytests like this wouldn't catch any failures. The pass doesn't crash but produces wrong IR right?

Or did it crash? What was the actual error?

paul0403: But my point is, since this test passes both before and after this PR, it doesn't actually test the fix added here.

multiphaseCFD: really? my local test shows that this test fails before this PR. did you test locally and tests pass with main branch?

paul0403: > did you test locally and tests pass with main branch

yes

mudit2812: @paul0403 there is lit testing [like this](https://github.com/PennyLaneAI/catalyst/blob/cb600b3fa550add2e9b62488e52db855ebedc67b/frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_diagonalize_measurements.py#L88) as well.

multiphaseCFD: did you enable program capture?

paul0403: I have capture turned on. Both with and without the gate, I see the 

```
pennylane.exceptions.CompileError: Observables are not qubit-wise commuting. Please apply the `split-non-commuting` pass first.
```

error.

multiphaseCFD: what if capture is not manually turned on?

paul0403: Without capture, if the gate is there I see the error, if the gate is not there I don't see the error.

paul0403: This actually ties back to what I was telling @joeycarter: in the capture pipeline the frontend doesn't actually generate the double extract starting mlir, so this is not a bug there!

multiphaseCFD: moved the test out of the class using the `use_capture` fixture @paul0403

paul0403: Since the ftqc passes all basically require the capture frontend, I don't think we should change the contents or behavior of a pass (in this case, adding cse) to accomadate a frontend that the pass isn't even supposed to be used with.

In other words, since there were no double extracts in the mlir produced from the capture frontend to begin with, there isn't actually a bug

paul0403: So me (and @joeycarter ) suggest just checking that capture is turned on in the pass (which is what the pass is supposed to work with anyway).

joeycarter: :point_up: Agreed

multiphaseCFD: > Since the ftqc passes all basically require the capture frontend, I don't think we should change the contents or behavior of a pass (in this case, adding cse) to accomadate a frontend that the pass isn't even supposed to be used with.

just want to confirm with @lillian542 , do we want the pass still work as we expected if program capture is off? The context is this pass will be public-facing and would not only be used by the FTQC workflows. CC @joeycarter @paul0403

lillian542: This isn't an FTQC pass (that's why it's in `quantum` instead of `mbqc`). The `diagonalize_measurements` pass is a pretty normal user-pass, and this is used to make sure we are at feature parity with the tape pass in terms of raising informative errors when the program is incompatible with the pass.

Maybe I've misunderstood something here about how the unified compiler infrastructure makes decisions, but my understanding is that:

```
@qml.qjit
@qml.diagonalize_measurements
def circ():
   ... whatever
   return qml.expval(qml.X(0)), qml.expval(qml.Y(0))
```

will switch from applying the tape transform to using the xDSL version of this pass automatically when [this PR](https://github.com/PennyLaneAI/pennylane/pull/9100) is merged. So it will also be used with the tape-based Catalyst frontend. Is that incorrect?

lillian542: It's just being implemented by the FTQC team because we were the first to be blocked by it not being available in the capture pipeline.

paul0403:
Also I just realized if this is only a bug in the non capture path then it will not get resolved by reference semantics migration, since that only happens for the capture path.

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_diagonalize_measurements.py
lillian542: Concretely, without this change, on [this branch](https://github.com/PennyLaneAI/pennylane/pull/9100), the following correctly raises an error with capture enabled, but silently returns incorrect results with capture disabled:

```
qml.capture.disable()

@qml.qjit
@qml.transforms.diagonalize_measurements
@qml.qnode(qml.device("lightning.qubit", wires=2), shots=1000)
def circuit_test():
    return qml.expval(qml.Z(0) + qml.Y(0))

circuit_test()
```

We should actually be running all the tests for `diagonalize-final-measurements` with the `use_both_frontend` fixture, I think.

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_diagonalize_measurements.py
```diff
@@ -861,7 +858,7 @@ def device(self):
 
     @pytest.mark.parametrize("add_compbasis_meas", ["false", "wires", "qreg"])
     @pytest.mark.parametrize("obs", NON_COMMUTE_SINGLE_OBS_LIST)
-    @pytest.mark.parametrize("measurements", [qml.expval, qml.var, qml.sample])
+    @pytest.mark.parametrize("measurements", [qml.expval, qml.var])
```
multiphaseCFD, line 861: removed `qml.sample` here since sampling `obs` is not supported by `catalyst`

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_diagonalize_measurements.py
```diff
@@ -882,8 +879,21 @@ def circuit(x):
...
+        @qml.qnode(device)
+        def circuit():
+            return qml.expval(qml.X(0)), qml.var(qml.Z(0))
+
+        with pytest.raises(CompileError, match=_non_commuting_err_msg):
+            qml.qjit(circuit)
+
     @pytest.mark.parametrize("obs", NON_COMMUTE_MULTI_OBS_LIST)
-    @pytest.mark.parametrize("m", [(qml.expval, qml.var), (qml.expval, qml.sample)])
+    @pytest.mark.parametrize("m", [(qml.expval, qml.var)])
```
multiphaseCFD, line 896: Same as above

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_diagonalize_measurements.py
multiphaseCFD: only `qml.expval` and `qml.var` support `obs`. see more [here](https://github.com/PennyLaneAI/catalyst/blob/94f5efcd15497add5ba362652eee94a3306d3cae/frontend/catalyst/device/verification.py#L300-L305)

lillian542 (approved):
LGTM, thanks @multiphaseCFD

PR #2639, opened 2026-03-30
Support conditional pauli rotations and multiplexed pauli measurements in the runtime

dime10:
Simplifies and speeds up the lowering by removing the need to turn these operations from the PBC dialect into their SCF equivalents. This keeps the IR size smaller.

[sc-115269]


lazypanda10117 (approved):
Thanks David!

mlir/lib/PBC/Transforms/ConversionPatterns.cpp
```diff
@@ -72,24 +72,29 @@ template <typename T> struct PPRotationBasedPattern : public OpConversionPattern
...
+                                         IntegerType::get(ctx, 1), IntegerType::get(ctx, 64)},
+                                        /*isVarArg=*/true);
 
         LLVM::LLVMFuncOp fnDecl = catalyst::ensureFunctionDeclaration<LLVM::LLVMFuncOp>(
             rewriter, op, qirName, qirSignature);
 
         // Get the rotation angle based on the op type
         // Since the qml.PauliRot(phi) == PPR(phi/2), this rotation_kind is multiplied by 2.
         Value thetaValue;
+        Value cond;
```
sengthai, line 91:
```suggestion
        Value cond = LLVM::ConstantOp::create(rewriter, loc, rewriter.getBoolAttr(1));
```

i think we can remove `else` branches below by setting `code=1` as default.

sengthai (approved):
Nice! Thanks @dime10. 👍

mlir/lib/PBC/Transforms/ConversionPatterns.cpp
dime10: Discussed: Since this would create a ConstantOp no matter if it was needed or not, I won't apply the suggestion.

PR #2641, opened 2026-03-31
Fix double extracts when multiple named obs/hermitian obs are present

paul0403:
**Context:**
There's a bug where multiple `quantum.extract` operations from the same index were being created when there are multiple named observables or Hermitian observables on that same wire index, when capture is not enabled.

This blocks some ftqc passes, namely when trying to judge whether two non-commuting named/Hermitian observables are on the same wire or not.

This is also just incorrect program: the second extract is extracting from a "hole".

After binding the extract op for the named obs / Hermitian obs, insert it to the register promise object's cache (if it's not already there), so the second extract call can find it in the cache.

Note that this is the same strategy for gates.

**Benefits:**
Correct program.

**Related GitHub Issues:**
[sc-115297]


frontend/catalyst/jax_tracer.py
```diff
@@ -970,6 +970,10 @@ def trace_observables(
     elif isinstance(obs, KNOWN_NAMED_OBS):
         qubits = qrp.extract(wires, allow_reuse=True)
         obs_tracers = namedobs_p.bind(qubits[0], kind=type(obs).__name__)
+        # When there are multiple named obs, they could be on the same wire
+        # Must delay the insert to after all named obs are traced, to avoid double extracts
+        if all(w not in qrp.cache for w in wires):
```
dime10: Could you elaborate on the conditional? Why not insert regardless?

paul0403: Looking at the extract and insert methods:
```python
    def extract(self, wires: List[Any], allow_reuse=False) -> List[DynamicJaxprTracer]:
        ...
        for w in wires:
            if w in qrp.cache:
                qubit = qrp.cache[w]
                assert (
                    qubit is not None
                ), f"Attempting to extract wire {w} from register {qrp.base} for the second time"
                qubits.append(qubit)
                if not allow_reuse:
                    qrp.cache[w] = None
            else:
                qubits.append(qextract_p.bind(qrp.base, w))
        return qubits

    @debug_logger
    def insert(self, wires, qubits) -> None:
        qrp = self
        assert len(wires) == len(qubits), f"len(wires)({len(wires)}) != len(qubits)({len(qubits)})"
        for w, qubit in zip(wires, qubits):
            assert (w not in qrp.cache) or (
                qrp.cache[w] is None
            ), f"Attempting to insert an already-inserted wire {w} into {qrp.base}"
            qrp.cache[w] = qubit
```


The extract here for named obs has `allow_reuse=True`, which means the cached qubit value will not get overwritten to `None` in the second extract (and beyond). This breaks the assertion in the insert.

paul0403: This is kinda a super edge case, where the named obs op itself needs to write to the cache. This is because the edge case circuit is one without any gates, so when the named obs primitive's bind happens, the cache is still empty.

joeycarter (approved):
Looks good to me, thanks @paul0403! 💯

frontend/catalyst/jax_tracer.py
dime10: Somehow there is something that feels fishy to me, but I don't quite follow the code logic. Do you think there is a design flaw with the flag? 

What is the point of cached values not being set to None?

paul0403: For gates, there's always a call to `qrp.extract()` before the `qinst_p.bind()` , and a call to `qrp.insert()` after, unconditionally.
Let's follow the old logic, aka without my added (conditioned) insert after the named obs op. 

So if there are gates, e.g. X(0), Z(0), expval(X(0)), expval(Z(0)), the cache looks like this:
init: empty cache

Extract call before the X(0) sees empty cache, so it adds a new extract op.
quantum.custom "X" is binded, with input being the new extract
Insert call after the X(0) adds X_out_qubit into the cache. Cache is {0: X_out_qubit}.

Extract call before the Z(0) sees cache {0: X_out_qubit}. It finds 0 in the cache and reads X_out_qubit. On gates, `allow_reuse` is false, so the cache is overwritten with None. Cache when exiting the extract call is {0: None}.
quantum.custom "Z" is binded, with input X_out_qubit
Insert call after the Z(0) adds Z_out_qubit into the cache. Cache is {0: Z_out_qubit}.

Extract call before expval(X(0)) sees cache {0: Z_out_qubit}. It finds 0 in the cache and reads Z_out_qubit. On named obs, `allow_reuse` is true, so it doesn't overwrite the cache with None. Cache stays the same.
quantum.namedobs "X(0)" is binded, with input Z_out_qubit

Extract call before expval(Z(0)) sees cache {0: Z_out_qubit}. It finds 0 in the cache and reads Z_out_qubit. On named obs, `allow_reuse` is true, so it doesn't overwrite the cache with None. Cache stays the same.
quantum.namedobs "Z(0)" is binded, with input Z_out_qubit

--------------

So this is all good. The problem occurs when there are no gates, i.e. just expval(X(0)), expval(Z(0))

init: empty cache

Extract call before expval(X(0)) sees empty cache. So it adds a new extract op. 
quantum.namedobs "X(0)" is binded, with input being the new extract

Extract call before expval(Z(0)) sees empty cache. So it adds a new extract op. 
quantum.namedobs "Z(0)" is binded, with input being the new extract

This is the source of the original double extract problem.

-------------

Now, imagine we add an insert call after each named obs, but unconditionally

init: empty cache

Extract call before expval(X(0)) sees empty cache. So it adds a new extract op. 
quantum.namedobs "X(0)" is binded, with input being the new extract
Insert call after expval(X(0)) adds the newly extracted qubit to the cache. Cache is now {0: extract_0}

Extract call before expval(Z(0)) sees cache {0: extract_0}.  It finds 0 in the cache and reads extract_0. On named obs, `allow_reuse` is true, so it doesn't overwrite the cache with None. Cache stays the same.
quantum.namedobs "Z(0)" is binded, with input extract_0
Insert call after expval(Z(0)) hits the assertion failure `assert (w not in qrp.cache) or (qrp.cache[w] is None)`, since the cache was not overwritten with `None`!

dime10: That's a very detailed and clear explanation, thank you! :) 

Does that not mean the insert should be on exactly those wires that aren't yet in the cache, rather than an all or nothing scenario ?

paul0403: Good point! https://github.com/PennyLaneAI/catalyst/pull/2641/commits/b8a25cdae6b75c03c7334322888c57bda259db28

frontend/catalyst/jax_tracer.py
```diff
@@ -970,10 +971,17 @@ def trace_observables(
     elif isinstance(obs, KNOWN_NAMED_OBS):
         qubits = qrp.extract(wires, allow_reuse=True)
         obs_tracers = namedobs_p.bind(qubits[0], kind=type(obs).__name__)
+        # When there are multiple named obs, they could be on the same wire
+        # Must delay the insert to after all named obs are traced, to avoid double extracts
```
dime10: I think this comment confused me as well
```suggestion
        # qrp.insert ensures the extracted wires are added to cache, however this call is only
        # allowed on wires that aren't already in the cache.
```

dime10 (approved):

PR #2646, opened 2026-04-01
Allow identity named observable to take in multiple wires 

paul0403:
**Context:**
As a follow up to #2641, the logic is updated to only use the index-0 wire in named obs handling, since the MLIR named obs op cannot take in a variadic range of qubits.

**Benefits:**
Correctness when a named obs takes in multiple wires, e.g. `qml.Identity(wires=[0,1,2])`

**Related GitHub issues:**
closes #2645 
[sc-115409]

### DIFF
--- a/frontend/catalyst/jax_tracer.py
+++ b/frontend/catalyst/jax_tracer.py
@@ -976,7 +976,6 @@ def trace_observables(
         # won't actually lead to wrong execution results, since the only exception is the identity
         # But of course we should fix this at some time.
         # TODO: make the NamedObs op take in multiple qubit values
-        # https://github.com/PennyLaneAI/catalyst/issues/2645
 
         qubits = qrp.extract([wires[0]], allow_reuse=True)
         obs_tracers = namedobs_p.bind(qubits[0], kind=type(obs).__name__)
--- a/frontend/catalyst/jax_tracer.py
+++ b/frontend/catalyst/jax_tracer.py
@@ -974,8 +974,9 @@ def trace_observables(
         # When there are multiple named obs, they could be on the same wire
         # qrp.insert ensures the extracted wires are added to cache, however this call is only
         # allowed on wires that aren't already in the cache.
-        if wires[0] not in qrp.cache:
-            qrp.insert(wires, qubits)
+        for w, q in zip(wires, qubits):
+            if w not in qrp.cache:
+                qrp.insert([w], [q])
     elif isinstance(obs, qml.Hermitian):
         # TODO: remove once fixed upstream: https://github.com/PennyLaneAI/pennylane/issues/4263
         qubits = qrp.extract(wires, allow_reuse=True)
--- a/frontend/catalyst/jax_tracer.py
+++ b/frontend/catalyst/jax_tracer.py
@@ -969,19 +969,27 @@ def trace_observables(
             qubits = qrp.extract(wires, allow_reuse=True)
             obs_tracers = compbasis_p.bind(*qubits)
     elif isinstance(obs, KNOWN_NAMED_OBS):
-        qubits = qrp.extract(wires, allow_reuse=True)
+        # The MLIR NamedObs operation only takes in a single qubit value, instead of a variadic
+        # range.
+        # This is fine for most named observables, but qml.Identity can take in multiple wires
+        # While the current logic here assumes there is only one wire on the observable, this
+        # won't actually lead to wrong execution results, since the only exception is the identity
+        # But of course we should fix this at some time.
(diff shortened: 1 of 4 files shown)


mehrdad2m (approved):

frontend/catalyst/jax_tracer.py
```diff
@@ -974,8 +974,9 @@ def trace_observables(
         # When there are multiple named obs, they could be on the same wire
         # qrp.insert ensures the extracted wires are added to cache, however this call is only
         # allowed on wires that aren't already in the cache.
-        if wires[0] not in qrp.cache:
-            qrp.insert(wires, qubits)
+        for w, q in zip(wires, qubits):
```
andrijapau: Do you expect these iterables to be the same length?

frontend/test/pytest/test_measurements_results.py
```diff
@@ -251,6 +251,23 @@ def expval1(x: float):
         observed = expval1(np.pi)
         assert np.isclose(observed, expected)
 
+    def test_named_identity(self, backend):
```
JerryChen97, line 254:
```suggestion

    @pytest.mark.capture_todo
    def test_named_identity(self, backend, capture_mode):
```

frontend/test/pytest/test_measurements_results.py
```diff
@@ -251,6 +251,23 @@ def expval1(x: float):
...
+        """
+        Test expval for identity named observable on multiple wires.
+
+        Note that the identity observable on multiple wires doesn't yet work in capture:
+        https://github.com/PennyLaneAI/catalyst/issues/2645
+        """
+
+        @qjit(capture=False)
```
JerryChen97:
```suggestion
        @qjit(capture=capture_mode)
```

frontend/test/pytest/test_measurements_results.py
```diff
@@ -251,6 +251,23 @@ def expval1(x: float):
...
+        @qml.qnode(qml.device(backend, wires=3))
+        def expval():
+            return qml.expval(qml.Identity(wires=[1, 2])), qml.expval(qml.Identity(wires=[0, 1, 2]))
```
paul0403, line 262: Just a note to reviewers: this is a hand-picked example. In the old logic, the second identity's `wires[0]` is `0`, which is not added to the cache by the first one, and the insert call will insert [0,1,2], causing cache clash with the [1,2] from the first identity.

frontend/catalyst/jax_tracer.py
paul0403: Yes. 

This is the same code as the hermitian op downstairs, btw

frontend/test/pytest/test_measurements_results.py
paul0403: done

frontend/catalyst/jax_tracer.py
dime10: Just to confirm, we extract all wires but only bind one? I think this is a logic bug. The code probably just didn't consider that a named observable could have more than one wire, but didn't explicitly assert this, so we never discovered this inconsistency (good showcase of why asserting assumptions is great).

dime10: If we mean to _intentionally_ disregard anything but the first wire in such observables, a comment would certainly be appropriate, on top of only extracting the wires that we use.

andrijapau: In that case we should add `strict=True` to be safe

paul0403: This is because the named obs op in mlir cannot take in multiple wires. This has been a bug since forever. 

I agree this is incorrect, but the only instance where this occurs would be on a named obs with multiple wires, which only is identity, so technically there's no wrong execution results; and named obs ops don't even contribute to data flow inside the IR except a read at the end of a program, so it's not the most severe of bugs. I opened an issue to track it: #2645

For now, this blocks a demo, so I'm just adding this patch to get things restored to where they were (despite them being wrong). The proper solution is of course allowing the mlir named obs op to take in multiple qubits, but I think I've already exceeded my alloted "things from the side" time :sweat_smile:

paul0403: done

paul0403: Added a comment, and now only using the index 0

dime10: > Added a comment, and now only using the index 0

Good enough for me 👍

rniczh (approved):
LGTM 👍

PR #2651, opened 2026-04-06
Add dynamic shapes support for while loop

albi3ro:
**Context:**

Adds support for dynamic shapes with the while loop with program capture. 

Directly lowers the plxpr while loop instead of translating it to the legacy frontend version first.

**Description of the Change:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-114570]

PR #2655, opened 2026-04-07
add abstracted_axes support

albi3ro:
**Context:**

Adds support to be able to specify `abstracted_axes` with `capture=True`.

Note that `for_loop`, `while_loop`, `cond`, etc.  don't yet support dynamic shapes, but that should happen in #2603 and #2651 

**Benefits:**

`abstracted_axes` support.

**Possible Drawbacks:**

Maybe I shouldn't have tidyied up the jax patches in this PR.  

Also, I really don't have a good explanation for why it has to be this way.

And making a breaking change to `trace_from_pennylane`.

**Related GitHub Issues:**

[sc-114572]


frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -210,14 +224,8 @@ def f(x):
     )
     original_fn = partial(interpreter.eval, plxpr.jaxpr, plxpr.consts)
 
-    # pylint: disable=import-outside-toplevel
-    from jax._src.interpreters.partial_eval import DynamicJaxprTrace
-
     def wrapped_fn(*args, **kwargs):
```
albi3ro, line 216: This now occurs within the full jax patches in `trace_from_pennylane`.

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -527,11 +535,10 @@ def wrapper(*args):
 # pylint: disable=too-many-positional-arguments
 def trace_from_pennylane(
     fn,
+    args,
```
albi3ro, line 555: I now need the *actual* numeric arguments.  `abstracted_axes` doesn't work with `ShapedArray` objects like are present in `sig`, and I need to have the static arguments present as well, do `dynamic_args` also didn't work.

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -563,58 +567,49 @@ def trace_from_pennylane(
...
-        (lax, "_dyn_shape_staging_rule", patched_dyn_shape_staging_rule),
-        (
-            jax._src.pjit,  # pylint: disable=protected-access
-            "pjit_staging_rule",
-            patched_pjit_staging_rule,
-        ),
-        (DictPatchWrapper(pe.custom_staging_rules, jit_p), "value", patched_pjit_staging_rule),
-        (pe, "get_aval", get_aval2),
-    ):
+    ), Patcher(*get_jax_patches()):
```
albi3ro, line 602: Not actually needed for this PR, but this function was getting rather complicated and hard to read. So similar to what Yushao did in `qml.capture.get_jax_patches`, I just extracted them all to their own function in `catalyst.jax_extras.patches`.

frontend/test/pytest/from_plxpr/test_preprocessing.py
```diff
@@ -629,6 +629,8 @@ def workflow():
         assert f2_pipelines[0][1][0].pass_name == "cancel-inverses"
 
         if skip_preprocess:
+            for w in recwarn:
+                print(w)
```
albi3ro:
```suggestion
```

frontend/test/lit/test_jax_dynamic_api.py
albi3ro, line 1: Most of the lines changed here are just grouping the tests into classes for better organization.

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -524,14 +539,23 @@ def wrapper(*args):
     return next_eval.eval(inner_jaxpr, consts, *non_const_args)
 
 
+def _extract_abstract_shapes(flat_inputs):
+    abstract_shapes = []
+    for a in flat_inputs:
+        for s in a.shape:
+            # need to us "is" for comparing tracers
+            if not isinstance(s, int) and not any(s is a for a in abstract_shapes):
```
mudit2812:
```suggestion
            if not isinstance(s, int) and not any(s is _a for _a in abstract_shapes):
```

mudit2812: Since `a` is already used as the loop variable for the outermost loop

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -540,19 +564,16 @@ def trace_from_pennylane(
 
     Args:
         fn(Callable): the user function to be traced
+        args (tuple): the positional arguments to the user functions
+        kwargs(Dict[str, Any]): keyword argumemts to the function.
```
mudit2812:
```suggestion
        kwargs(Dict[str, Any]): keyword arguments to the function.
```

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -563,58 +584,54 @@ def trace_from_pennylane(
...
-            "debug_info": debug_info,
         }
 
-        args = sig
+        # we want to have the same tracers as inputs to plxpr capture and from_plxpr
+        # translation, as this tells jax which inputs match which dynamic shapes
+        # if we have concrete inputs to both, jax will get confused.
+        # instead of passing in abstracted_axes, we pass in arguments with the
+        # dynamic shapes in the right place matching the correct inputs.
+        # really confusing, but this solution mostly seems to work
```
mudit2812, line 614: > mostly seems to work

Are there any situations you have encountered that aren't working?

mudit2812 (approved):

frontend/catalyst/from_plxpr/from_plxpr.py
albi3ro: 🤦

frontend/catalyst/from_plxpr/from_plxpr.py
albi3ro: Not yet. My annoyance is basically around how this isn't necessarily an obvious and intuitive solution.

JerryChen97 (approved):

PR #2656, opened 2026-04-07
Add TensorOp support for `measurements-from-samples`

lillian542:
**Context:**

The implementation of `measurements-from-samples` pass can't yet handle Expval and Var of multi-wire observables.

We add support for `TensorOp`. Support for `HamiltonianOp` is expected to be based on the splitting of Sums in `split_non_commuting`. Specifically we:

- Expand the `validate_observable_op` method to also validate `TensorOp`, and to raise a useful error message for `HamiltonianOp`
- Add a `get_observable_op_qubits` method, and start assuming a list of qubits and using `len(qubits)` instead of `1` as the number of qubits on the observable.
- We handle the case where multiple post-processing functions generate helper functions with matching names by adding an index to the end of post-processing function names. This allows us to return mixed MPs that need different post-processing functions (that each rely on their own versions of functions like `_where`).
- We add a `create_postprocessing_obs` function that gets the eigenvalues of the NamedObs or TensorOp and generates the relevant post-processing function for each MP containing an observable (only Expval and Var). This replaces `_postprocessing_var` and `_postprocessing_expval`

Additionally, we 
- reorganize so we have a single Pattern and only iterate through each qnode once (instead of once per MP type), 
- swap the order that we process the MPs in so that the post-processing function order matches the MP order, and 
- update the tests to use expected analytic results, for more rigorous validation (previous tests only used cases where the output was expected to be 0 or 1)

**Benefits:**
- we can support TensorOps
- we can support QNodes that return mixed MPs
- we only iterate through each qubit once (more efficient)
- our tests are more rigorous

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A

[sc-107541] [sc-117786]


frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_measurements_from_samples_integration.py
```diff
@@ -210,34 +252,107 @@ def test_expval_sprod_with_split_non_commuting(self, shots, initial_ops, expecte
...
+            # CHECK-NOT: quantum.expval
+            # CHECK: quantum.sample
+            return qml.expval(coeff*qml.Z(wires=0) + qml.Y(1))
+        
+        expected_res = coeff * np.cos(phi1) - np.sin(phi2)
+        assert np.isclose(expected_res, circuit()), "Sanity check failed, is expected_res correct?"
+
+        pipeline = qml.CompilePipeline(qml.transform(pass_name="split-non-commuting"), qml.transform(pass_name="measurements-from-samples"))
+        
+        # ToDo: becomes 2 executions of 5000 shots each, can go back to 5000 once #2657 merges
```
lillian542: Remember to do this when https://github.com/PennyLaneAI/catalyst/pull/2657 merges

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -19,9 +19,6 @@
 -----------------
 
   * The compilation pass assumes a static number of shots.
-  * Usage patterns that are not yet supported with program capture are also not supported in the
-    compilation pass. For example, operator arithmetic is not currently supported, such as
-    qml.expval(qml.Y(0) @ qml.X(1)).
```
lillian542, line 24: add note about HamiltonianOp limitation

lillian542:
The remaining failing test should be fixed by this PR: https://github.com/PennyLaneAI/catalyst/pull/2693

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -119,60 +166,91 @@ def get_observable_op(cls, op: quantum.ExpvalOp | quantum.VarianceOp) -> quantum
...
+        if isinstance(op, quantum.NamedObsOp):
+            if op.type.data != "PauliZ":
+                raise NotImplementedError(
+                    "Expected all observables to be diagonalized before application of rewrite"
+                    f"pattern, but received '{op.type.data}'"
+                )
+
+        elif isinstance(op, quantum.TensorOp):
+            for obs in op.operands:
+                if obs.owner.type.data != "PauliZ":
```
joeycarter, line 203: This is a minor point, but if `obs` is an `SSAValue`, then its owner can be either another operation (it looks like we're assuming it's a `quantum.NamedObsOp` given that we're accessing its `type` attribute), or it can be a `Block` if `obs` is a block argument (in which case, it has no `type` attribute, which would result in an error).

I don't expect this to be an issue in most cases, since we don't tend to write subroutines that take an observable as an argument, but maybe we could add a check like

```python
if not isinstance(obs.owner, quantum.NamedObsOp):
    raise CompileError("...")
```

to catch any unexpected IR that we aren't supporting for now.

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -119,60 +166,91 @@ def get_observable_op(cls, op: quantum.ExpvalOp | quantum.VarianceOp) -> quantum
...
+                    raise NotImplementedError(
+                        "Expected all observables to be diagonalized before application of"
+                        f"rewrite pattern, but received '{op.type.data}'"
+                    )
```
joeycarter, line 207: Is this what you meant? `op` is the `quantum.TensorOp` here, which does not define a `type` attribute, so this would result in an error as is.

```suggestion
                    raise NotImplementedError(
                        "Expected all observables to be diagonalized before application of"
                        f"rewrite pattern, but received '{obs.owner.type.data}'"
                    )
```

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -119,60 +166,91 @@ def get_observable_op(cls, op: quantum.ExpvalOp | quantum.VarianceOp) -> quantum
...
+    @staticmethod
+    def get_observable_op_qubits(op: quantum.NamedObsOp | quantum.TensorOp):
+        """Get a list of all qubits in the observable"""
+
+        assert isinstance(
+            op, quantum.NamedObsOp | quantum.TensorOp
+        ), f"expected quantum.NamedObsOp or quantum.TensorOp but received {type(op)}"
+
+        if isinstance(op, quantum.TensorOp):
+            return [obs.owner.operands[0] for obs in op.operands]
```
joeycarter, line 224: Similar point here accessing `obs.owner`, which can be either an `Operation` or a `Block`.

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -119,60 +166,91 @@ def get_observable_op(cls, op: quantum.ExpvalOp | quantum.VarianceOp) -> quantum
...
+
+        return op.operands
+
     @staticmethod
     def insert_compbasis_op(
-        in_qubit: ir.SSAValue, ref_op: ir.Operation, rewriter: PatternRewriter
+        in_qubits: ir.SSAValue, ref_op: ir.Operation, rewriter: PatternRewriter
```
joeycarter: Since `in_qubits` is a `Sequence` of `SSAValue`s

```suggestion
        in_qubits: Sequence[ir.SSAValue], ref_op: ir.Operation, rewriter: PatternRewriter
```

You'll also need `from collections.abc import Sequence`

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -119,60 +166,91 @@ def get_observable_op(cls, op: quantum.ExpvalOp | quantum.VarianceOp) -> quantum
...
     ) -> quantum.ComputationalBasisOp:
         """Create and insert a computational-basis op (quantum.ComputationalBasisOp).
 
-        The computation-basis op uses `in_qubit` as its input operand. It is inserted *before* the
+        The computation-basis op uses `in_qubits` as its input operand. It is inserted *before* the
         given reference operation, `ref_op`, using the supplied `rewriter`.
 
         Args:
-            in_qubit (SSAValue): The SSA value used as input to the computational-basis op.
+            in_qubits (SSAValue): A list of SSA value used as input to the computational-basis op.
```
```suggestion
            in_qubits (Sequence[SSAValue]): A sequence of SSA value used as input to the computational-basis op.
```

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -119,60 +166,91 @@ def get_observable_op(cls, op: quantum.ExpvalOp | quantum.VarianceOp) -> quantum
...
             raise NotImplementedError(
-                f"Observable '{op.type.data}' used as input to measurement operation is not "
-                f"supported for the measurements_from_samples transform; currently only the "
-                f"PauliZ observable is permitted"
+                "Supported observable types for measurements-from-samples are quantum.NamedObsOp "
+                f"and quantum.TensorOp, but received {type(op).__name__}"
             )
 
+    @staticmethod
+    def get_observable_op_qubits(op: quantum.NamedObsOp | quantum.TensorOp):
```
```suggestion
    def get_observable_op_qubits(op: quantum.NamedObsOp | quantum.TensorOp) -> Sequence[ir.SSAValue]:
```

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -288,40 +375,20 @@ def get_postprocessing_funcs_from_module_and_insert(
             prev_op = postprocessing_func_op
             for _op in islice(postprocessing_module.body.ops, 1, None):
                 helper_op = _op.clone()
+                # if the helper_op calls any functions in the module, also rellabel those callees
```
```suggestion
                # if the helper_op calls any functions in the module, also relabel those callees
```

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -423,221 +490,150 @@ def update_returns(
...
-        shots (int): The number of shots (e.g. as retrieved from the DeviceInitOp).
-    """
+        # delete now unused obs --> mp --> tensor chain
+        rewriter.erase_op(from_elements_op)
+        rewriter.erase_op(mp_op)
+        if isinstance(observable_op, quantum.TensorOp):
+            inner_obs = [op.owner for op in observable_op.operands]
+            rewriter.erase_op(observable_op)
+            for o in inner_obs:
+                rewriter.erase_op(o)
```
joeycarter, line 575: Similar point here accessing op.owner, which can be either an Operation or a Block. `rewriter.erase_op` only accepts an Operation, not a Block.

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_measurements_from_samples_lit.py
```diff
@@ -234,21 +233,55 @@ def test_1_wire_counts(self, run_filecheck):
         """
 
         pipeline = (MeasurementsFromSamplesPass(),)
-        run_filecheck(program, pipeline)
+        with pytest.raises(NotImplementedError, match="operations are not supported"):
+            run_filecheck(program, pipeline)
```
joeycarter, line 237: Curious why you changed this test from `xfail` to `with pytest.raises`? Presumably we would want to support `counts` at some point in the future?

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_measurements_from_samples_lit.py
```diff
@@ -234,21 +233,55 @@ def test_1_wire_counts(self, run_filecheck):
...
+                %4 = quantum.state %3 : tensor<4xcomplex<f64>>
+
+                func.return %4 : tensor<4xcomplex<f64>>
+            }
+        }
+        """
+
+        pipeline = (MeasurementsFromSamplesPass(),)
+        with pytest.raises(NotImplementedError, match="operations are not supported"):
+            run_filecheck(program, pipeline)
```
joeycarter, line 269: It makes sense to use `with pytest.raises` here since we're not anticipating to support the `state` measurement process. (Maybe we should change the error from `NotImplementedError` to `CompileError` to make that clear?)

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_measurements_from_samples_lit.py
```diff
@@ -234,21 +233,55 @@ def test_1_wire_counts(self, run_filecheck):
...
 
     def test_2_wire_expval(self, run_filecheck):
         """Test the measurements-from-samples pass on a 2-wire circuit terminating with an expval(Z)
         measurement on each wire. Includes a non-diagonalized operation (pass inserts diagonalizing gate).
+        measurement on each wire. Includes a non-diagonalized operation (pass inserts diagonalizing gate).
```
joeycarter: Looks like this line was duplicated 🙂 

```suggestion
```

joeycarter (commented):
Thanks @lillian542! Most of my points are fairly minor, the general structure of the change looks great. 🚀

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -52,13 +51,32 @@
...
 from catalyst.python_interface.utils import get_constant_from_ssa
 
+MEASUREMENT_PROCESS_TYPES = (
+    quantum.CountsOp,
+    quantum.ExpvalOp,
+    quantum.ProbsOp,
+    quantum.SampleOp,
+    quantum.StateOp,
+    quantum.VarianceOp,
+)
```
mudit2812, line 62: Since we have better base classes now for the quantum dialect, you can just import `catalyst.python_interface.dialects.quantum.TerminalMeasurementOp`

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -103,14 +111,53 @@ def __init__(self):
...
+                pass
+            elif isinstance(mp_op, quantum.CountsOp):
+                # Currently ``qml.counts()`` is unsupported due to differences in return
+                # type/shape in PennyLane and Catalyst. It may be supported at a later time.
+                # It is included for completeness and to notify users that it is unsupported.
+                raise NotImplementedError("qml.counts() operations are not supported.")
+            elif isinstance(mp_op, quantum.StateOp):
+                # It is not possible to recover a quantum state from samples; this is included
+                # for completeness and to notify users that ``state`` mps are not supported
+                raise NotImplementedError("qml.state() operations are not supported.")
```
mudit2812, line 154: For readability, I suggest using a match-case here

mudit2812 (commented):
Looks great to me @lillian542 . The only blockers from me are Joey's comments.

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
```diff
@@ -19,24 +19,23 @@
...
+  * HamiltonianOps (and therefore Sums and SProds) are not supported directly. Applying
+    `split-non-commuting` before this pass enables circuits with Sum/SProd observables.
   * qml.counts() is not supported since the return type/shape is different in PennyLane and
     Catalyst. See
     https://docs.pennylane.ai/projects/catalyst/en/stable/dev/quick_start.html#measurements
     for more information.
 """
 
-from abc import abstractmethod
 from dataclasses import dataclass
```
lillian542, line 31:
```suggestion
from collections.abc import Sequence
from dataclasses import dataclass
```

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
lillian542: That is what I mean, thanks 😅

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
lillian542: I can add a check. Can you give me an example of how you would end up with this IR, where a block argument can be an input for a `TensorOp`? I thought [this meant](https://github.com/PennyLaneAI/catalyst/blob/0e424cfea4a879dd520a9b068b81d9326240b723/mlir/include/Quantum/IR/QuantumOps.td#L937) the input has to be an observable, but maybe if I just tell the IR the output of my subroutine is a `!quantum.obs`, that requirement is filled?

Also, would this be a valid IR? From the guidelines in the [Qnode Transform Guide](https://github.com/PennyLaneAI/catalyst/blob/main/frontend/catalyst/python_interface/transforms/qnode-transform-guide.md), we have:

> Within a quantum kernel, individual quantum executions are represented by functions that carry the special qnode attribute. We will refer to these as qnode functions (name TBD):
A precise definition is difficult to provide and may not currently apply to all transformations or stages in Catalyst, however, for the purpose of this guide a qnode function will contain:
>  - device [initialization](https://github.com/PennyLaneAI/catalyst/blob/v0.14.0/frontend/catalyst/python_interface/dialects/quantum.py#L389) and [release](https://github.com/PennyLaneAI/catalyst/blob/v0.14.0/frontend/catalyst/python_interface/dialects/quantum.py#L412) instructions
> - one or more [measurement processes](https://github.com/PennyLaneAI/catalyst/blob/v0.14.0/frontend/catalyst/python_interface/dialects/quantum.py#L1058) 
>
> Conversely, such operations should not appear in any other kind of function. Quantum state evolution [operations](https://github.com/PennyLaneAI/catalyst/blob/v0.14.0/frontend/catalyst/python_interface/dialects/quantum.py#L273) may appear in qnode or non-qnode functions (e.g. as quantum subroutines).

It's not explicit here, and maybe that's worth raising elsewhere, but I would classify all `Observable` ops as part of specifying the measurement, and I would say they don't belong in subroutines. They aren't state evolutions.

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
lillian542: Updated here: fd529e21abffe0e96cdf9c2ba1c7ce416696c4e8

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
lillian542: Updated: fd529e21abffe0e96cdf9c2ba1c7ce416696c4e8

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
lillian542: Added: fd529e21abffe0e96cdf9c2ba1c7ce416696c4e8

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
lillian542: Added: fd529e21abffe0e96cdf9c2ba1c7ce416696c4e8

frontend/catalyst/python_interface/transforms/quantum/measurements_from_samples.py
lillian542: Added a check here! fd529e21abffe0e96cdf9c2ba1c7ce416696c4e8

lillian542: But I am wondering if considering this as a possibility for a check makes sense in the context of [this thread](https://xanaduhq.slack.com/archives/C030UB33QH3/p1776444369715419)

joeycarter: Thanks! This suggestion is mainly a sanity check and is mostly to do with the fact that Python is dynamically typed more than anything. I don't expect any code generated by Catalyst to trigger this error, but given that `op.owner` is implicitly being assumed to be an op rather than a block, when in principle it can be either, putting a check here will give a clearer error message than if there were no check.

For reference, I spotted this with a Python type checker. It saw that `obs.owner` has type `Operation | Block`, and without checking its type explicitly, calling `obs.owner.type` gives a warning "Object of type `Operation | Block` has no attribute `type`"

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_measurements_from_samples_lit.py
lillian542: I was thinking of it in terms of "I want to test what error message is raised" rather than "will this potentially be implemented the future". 

With `xfail` it can appear to "fail as expected" for an _unexpected_ error that's encountered before the error we want, as long as the type of error is the same. With `pytest.raises` I can match the error message. 

But I think this is a habit that is much more relevant when testing with full workflows in Python, and is unlikely to matter in lit tests, I can put it back.

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_measurements_from_samples_lit.py
lillian542: Sure! I guess this is unreachable code anyway outside of lit testing, since there's no way for the frontend to generate an IR that returns `state` and has `shots!=0`, and this pass would already have raised an error if `shots` were 0. I'd say that's another point in favour of `CompileError` 😅

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_measurements_from_samples_lit.py
joeycarter: Got it. I have no strong opinion, I usually use `with pytest.raises` to check for errors as a result of invalid user input (and to check that the _correct_ error is raised), and `xfail` for tests that should pass but we know that they fail because of some known issue (such as it not being implement! :wink:).

I'll leave it to you.

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_measurements_from_samples_lit.py
joeycarter: Sounds good, thanks :slightly_smiling_face:

lillian542: Updated here: https://github.com/PennyLaneAI/catalyst/pull/2656/commits/3c962760e774dece665124b74a663b4841d4ac29

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_measurements_from_samples_lit.py
lillian542: I put it back to xfail :)

joeycarter (approved):
Looks great, nice work! 💯

multiphaseCFD (approved):
Great work @lillian542 !

mudit2812 (approved):

lillian542:
[sc-117786]

PR #2657, opened 2026-04-07
Add wires strategy to split-non-commuting pass

rniczh:
**Context:**

Currently, every observable gets its own execution after spit-non-commuting pass. It works, but inefficient for circuits that has measurements on non-overlap wires.

We would like to add the grouping_strategy="wires" to group up all observables on different wires together, e.g. for a circuit returning some combination of observables X(0), Y(1) and Y(0), we would get two groups, and two `quantum.node`s.

Add the grouping strategy to the pass option and support both wires grouping mechanism and deduplicated for split-non-commuting pass.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-109333]


doc/releases/changelog-dev.md
```diff
@@ -756,8 +756,12 @@
 * A new compiler pass `split-non-commuting` has been added for QNode functions that measure
   non-commuting observables. It facilitates execution on devices that don't natively support
   measuring multiple non-commuting observables simultaneously by splitting them into separate
-  circuit executions, one group per observable for now.
+  circuit executions. The pass supports a `grouping_strategy` option: the default (`None`) assigns
+  each observable to its own group, while `"wires"` groups observables on non-overlapping wires into
+  the same execution, reducing the total number of generated circuits. Duplicate observables are
```
lillian542, line 768: Can we default to `"wires"`? It's pretty much always better, and it's closer to the tape-transform behaviour (tape transform prefers "qwc" where possible, but the next highest preference is "wires").

frontend/test/pytest/test_split_non_commuting.py
```diff
@@ -291,5 +291,167 @@ def circ():
...
+        """Z(0), X(1), Y(2) on separate wires -> 1 group"""
+        dev = qp.device("lightning.qubit", wires=3)
+
+        @qjit
+        @qp.transform(pass_name="split-non-commuting")(grouping_strategy="wires")
+        @qp.qnode(dev)
+        def circ():
+            qp.RX(0.3, wires=0)
+            qp.RY(0.5, wires=1)
+            qp.RZ(0.7, wires=2)
```
lillian542: Let's make this gate RX, an RZ rotation on the ground state doesn't do anything, so `expval(Y(2))` is currently 0. Tests that return 0 aren't as reliable for catching problems.

```suggestion
            qp.RX(0.7, wires=2)
```

frontend/test/pytest/test_split_non_commuting.py
```diff
@@ -291,5 +291,167 @@ def circ():
...
+            return qp.expval(qp.Z(0)), qp.expval(qp.X(1)), qp.expval(qp.Y(2))
+
+        @qjit
+        @qp.qnode(dev)
+        def circ_ref():
+            qp.RX(0.3, wires=0)
+            qp.RY(0.5, wires=1)
+            qp.RZ(0.7, wires=2)
+            return qp.expval(qp.Z(0)), qp.expval(qp.X(1)), qp.expval(qp.Y(2))
```
lillian542, line 304: Entirely optional, but to reduce the code duplication in tests, other places we do things like: 

```
dev = qp.device("lightning.qubit", wires=3)

snc_pass = qp.transform(pass_name="split-non-commuting")(grouping_strategy="wires")

@qp.qnode(dev)
def circ():
    qp.RX(0.3, wires=0)
    qp.RY(0.5, wires=1)
    qp.RX(0.7, wires=2)
    return qp.expval(qp.Z(0)), qp.expval(qp.X(1)), qp.expval(qp.Y(2))

circ_split = qp.qjit(snc_pass(circ))
circ_ref = qp.qjit(circ)

assert np.allclose(circ_split(), circ_ref())
```

Then you can just write each circuit once :)

frontend/test/pytest/test_split_non_commuting.py
```diff
@@ -291,5 +291,167 @@ def circ():
...
+    def test_overlapping_two_groups(self):
+        """Z(0), X(1), Y(1): Z(0), X(1) in group 0, Y(1) in group 1."""
+        dev = qp.device("lightning.qubit", wires=2)
+
+        @qjit
+        @qp.transform(pass_name="split-non-commuting")(grouping_strategy="wires")
+        @qp.qnode(dev)
+        def circ():
+            qp.RX(0.4, wires=0)
+            qp.RY(0.6, wires=1)
```
lillian542, line 317: In this circuit we can add either an additional `RX` or `RZ` on wire 1 so that `expval(Y(1))` isn't zero

frontend/test/pytest/test_split_non_commuting.py
```diff
@@ -291,5 +291,167 @@ def circ():
...
+    @pytest.mark.usefixtures("use_both_frontend")
+    def test_tensor_overlap(self):
+        """Z(0)@Z(1) overlaps X(0) on wire 0 -> 2 groups."""
+        dev = qp.device("lightning.qubit", wires=2)
+
+        @qjit
+        @qp.transform(pass_name="split-non-commuting")(grouping_strategy="wires")
+        @qp.qnode(dev)
+        def circ():
+            qp.RX(0.3, wires=0)
```
lillian542: If we make this RY, then `expval(X(0))` will be non-zero

frontend/test/pytest/test_split_non_commuting.py
```diff
@@ -291,5 +291,167 @@ def circ():
...
+            qp.RY(0.5, wires=1)
+            qp.RZ(0.7, wires=2)
+            return qp.expval(qp.Z(0) + qp.X(1) + 2 * qp.Y(2))
+
+        @qjit
+        @qp.qnode(dev)
+        def circ_ref():
+            qp.RX(0.3, wires=0)
+            qp.RY(0.5, wires=1)
+            qp.RZ(0.7, wires=2)
```
lillian542: Let's make this RX - currently `expval(Y(2))` is 0. 

```suggestion
            qp.RX(0.7, wires=2)
```

Particularly significant in this test because bugs with the coefficient tracking in post-processing wouldn't show up in this integration test - the output is independent of the coefficient of Y(2) right now.

frontend/test/pytest/test_split_non_commuting.py
```diff
@@ -291,5 +291,167 @@ def circ():
...
+        def circ():
+            qp.RX(0.4, wires=0)
+            qp.RY(0.6, wires=1)
+            return qp.expval(0.5 * qp.Z(0) + 3 * qp.X(1) + qp.Y(1))
+
+        @qjit
+        @qp.qnode(dev)
+        def circ_ref():
+            qp.RX(0.4, wires=0)
+            qp.RY(0.6, wires=1)
```
lillian542: Let's add some `qp.RZ(phi, wires=1)` op here after RY, so that the Y(1) term contributes to the outcome (currently 0)

frontend/test/pytest/test_split_non_commuting.py
```diff
@@ -291,5 +291,167 @@ def circ():
...
+            return qp.expval(0.5 * qp.Z(0) + 3 * qp.X(1) + qp.Y(1))
+
+        assert np.allclose(circ(), circ_ref())
+
+    @pytest.mark.usefixtures("use_both_frontend")
+    def test_duplicate_observables(self):
+        """Z(0) appears twice, X(1) once -> deduplicated, single group."""
```
lillian542:
```suggestion
        """Z(0) appears twice -> deduplicated, single group."""
```

frontend/test/pytest/test_split_non_commuting.py
```diff
@@ -291,5 +291,167 @@ def circ():
...
+        dev = qp.device("lightning.qubit", wires=2)
+
+        @qjit
+        @qp.transform(pass_name="split-non-commuting")(grouping_strategy="wires")
+        @qp.qnode(dev)
+        def circ():
+            qp.RX(0.3, wires=0)
+            qp.RY(0.5, wires=1)
```
lillian542:
```suggestion
```

frontend/test/pytest/test_split_non_commuting.py
```diff
@@ -291,5 +291,167 @@ def circ():
...
+            return qp.expval(2 * qp.Z(0) + qp.Z(0))
+
+        @qjit
+        @qp.qnode(dev)
+        def circ_ref():
+            qp.RX(0.3, wires=0)
+            qp.RY(0.5, wires=1)
```
lillian542:
```suggestion
```

lillian542 (approved):
Thanks @rniczh! I left a few minor suggestions on the tests, but this looks great 🚀

doc/releases/changelog-dev.md
lillian542: On second thought never mind, I think this is out of scope. We will want to do it eventually, but don't worry about this for now.

frontend/test/pytest/test_split_non_commuting.py
rniczh: Thanks! I updated all the tests at once in this commit: https://github.com/PennyLaneAI/catalyst/pull/2657/commits/64b3776fa96d24d8d3cb84e478b20240d5702d00

frontend/test/pytest/test_split_non_commuting.py
rniczh: I added `qp.RZ(0.25, wires=1)` here. https://github.com/PennyLaneAI/catalyst/pull/2657/commits/64b3776fa96d24d8d3cb84e478b20240d5702d00

frontend/test/pytest/test_split_non_commuting.py
rniczh: Done. https://github.com/PennyLaneAI/catalyst/pull/2657/commits/64b3776fa96d24d8d3cb84e478b20240d5702d00

frontend/test/pytest/test_split_non_commuting.py
rniczh: Done https://github.com/PennyLaneAI/catalyst/pull/2657/commits/64b3776fa96d24d8d3cb84e478b20240d5702d00

frontend/test/pytest/test_split_non_commuting.py
rniczh: Done. `qp.RZ(0.25, wires=1)` is added. https://github.com/PennyLaneAI/catalyst/pull/2657/commits/64b3776fa96d24d8d3cb84e478b20240d5702d00

frontend/test/pytest/test_split_non_commuting.py
rniczh: Done. https://github.com/PennyLaneAI/catalyst/pull/2657/commits/64b3776fa96d24d8d3cb84e478b20240d5702d00

frontend/test/pytest/test_split_non_commuting.py
rniczh: Done. https://github.com/PennyLaneAI/catalyst/pull/2657/commits/64b3776fa96d24d8d3cb84e478b20240d5702d00

frontend/test/pytest/test_split_non_commuting.py
rniczh: Thanks for the suggestion! it's much cleaner this way. I've updated the tests in this commit: https://github.com/PennyLaneAI/catalyst/pull/2657/commits/64b3776fa96d24d8d3cb84e478b20240d5702d00

frontend/test/pytest/test_split_non_commuting.py
```diff
@@ -291,5 +291,119 @@ def circ():
...
+        @qp.qnode(dev)
+        def circ():
+            qp.RX(0.3, wires=0)
+            qp.RY(0.5, wires=1)
+            qp.RX(0.7, wires=2)
+            return qp.expval(qp.Z(0)), qp.expval(qp.X(1)), qp.expval(qp.Y(2))
+
+        circ_split = qjit(self.snc_pass(circ))
+        circ_ref = qjit(circ)
+        assert np.allclose(circ_split(), circ_ref())
```
JerryChen97:
```suggestion
    def test_non_overlapping_single_group(self, capture_mode):
        """Z(0), X(1), Y(2) on separate wires -> 1 group"""
        dev = qp.device("lightning.qubit", wires=3)

        @qp.qnode(dev)
        def circ():
            qp.RX(0.3, wires=0)
            qp.RY(0.5, wires=1)
            qp.RX(0.7, wires=2)
            return qp.expval(qp.Z(0)), qp.expval(qp.X(1)), qp.expval(qp.Y(2))

        circ_split = qjit(self.snc_pass(circ), capture=capture_mode)
        circ_ref = qjit(circ, capture=capture_mode)
        assert np.allclose(circ_split(), circ_ref())
```

frontend/test/pytest/test_split_non_commuting.py
JerryChen97, line 1: Mind using the explicitly parametrized `capture_mode` instead? :laughing:

rniczh: Okay, this is my first time to know that 😆, thanks for the suggestion, I updated all the places in this file to use capture option instead. https://github.com/PennyLaneAI/catalyst/pull/2657/commits/b5c97d7d4f0c72351486dc6d8f43703080acd56e

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
```diff
@@ -23,14 +23,22 @@
...
 //   }
+//
+// With grouping_strategy="wires", observables on non-overlapping wires share a group:
+//   func.func @circuit() -> (f64, f64, f64) {
+//     %r0:2 = call @circuit.group.0()  // ev0, ev1 (different wires)
+//     %r1   = call @circuit.group.1()  // ev2 (overlaps with ev0)
+//     return %r0#0, %r0#1, %r1
+//   }
 
+#include <llvm/ADT/SetOperations.h>
```
paul0403: Wrong include position? (cc @kipawaa 😆 )

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
```diff
@@ -67,24 +74,153 @@ struct SplitNonCommutingPass : public impl::SplitNonCommutingPassBase<SplitNonCo
...
+    /// Collect the qubit SSA values an observable acts on.
+    /// For NamedObsOp: the single qubit operand.
+    /// For TensorOp: the union of qubits from all sub-observables.
+    static llvm::DenseSet<Value> getObservableQubits(Value obs)
     {
-        int groupId = 0;
+        llvm::DenseSet<Value> qubits;
+        Operation *defOp = obs.getDefiningOp();
+        if (!defOp) {
+            return qubits;
```
paul0403, line 84: Would it be possible to pass in `qubits` as a refernce and "return via the argument"? Because `DenseSet` is a owning container, having it initialized locally in the helper function and returning it feels a bit dangerous to me 🤔

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
```diff
@@ -67,24 +74,153 @@ struct SplitNonCommutingPass : public impl::SplitNonCommutingPassBase<SplitNonCo
...
+            }
+            return qubits;
+        }
+
+        return qubits;
+    }
+
+    /// Two observables are equivalent if they have the same op kind, same attributes,
+    /// and their qubit operands are the same SSA values.
+    static bool observablesEqual(Value lhs, Value rhs)
```
paul0403, line 104: Why not just use mlir's existing op equivalence infra? 

https://mlir.llvm.org/doxygen/structmlir_1_1OperationEquivalence.html

Is there something specific that you need here?

paul0403: I believe when you set the flag as `None`, it is strict "operands must be the same SSA value" type of op equivalence.

mlir/test/Quantum/SplitNonCommutingWiresTest.mlir
```diff
@@ -0,0 +1,356 @@
...
+// CHECK: %[[CALL1:.*]] = call @circ_overlap.group.1
+// CHECK: return %[[CALL0]]#0, %[[CALL0]]#1, %[[CALL1]]
+
+// CHECK-LABEL: func.func private @circ_overlap.group.0
+// CHECK-SAME: () -> (f64, f64) attributes {quantum.node}
+// CHECK: %[[SHOTS0:.*]] = arith.constant 50
+// CHECK: quantum.device shots(%[[SHOTS0]])
+// CHECK: quantum.alloc
+// CHECK: %[[OBS_Z0:.*]] = quantum.namedobs %{{.*}}[ PauliZ]
+// CHECK: %[[OBS_X1:.*]] = quantum.namedobs %{{.*}}[ PauliX]
```
paul0403: Can you check the extract op's index as well in the test?

mlir/test/Quantum/SplitNonCommutingWiresTest.mlir
```diff
@@ -0,0 +1,356 @@
...
+// CHECK-SAME: () -> (f64, f64, f64)
+// CHECK-NOT: quantum.node
+// CHECK: %[[CALL0:.*]]:2 = call @circ_dup.group.0
+// CHECK: return %[[CALL0]]#0, %[[CALL0]]#1, %[[CALL0]]#0
+
+// CHECK-LABEL: func.func private @circ_dup.group.0
+// CHECK-SAME: () -> (f64, f64) attributes {quantum.node}
+// CHECK: quantum.alloc
+// CHECK: %[[OBS_Z0:.*]] = quantum.namedobs %{{.*}}[ PauliZ]
+// CHECK: %[[OBS_X1:.*]] = quantum.namedobs %{{.*}}[ PauliX]
```
paul0403: Same here

paul0403: (and other similar places in the tests)

paul0403 (commented):
Really nice! This is a big optimization 🥳 

Just 2 small comments regarding the mlir api

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
rniczh: Return `DenseSet` should be more safety and efficient? Without side-effect and supporting NRVO optimization as well. So this is completely safe, because the object is returned by value, not by reference. And I believe this pattern is pretty common in practice in LLVM and MLIR, for example:

mlir/llvm-project/mlir/lib/Dialect/SCF/Transforms/BufferizableOpInterfaceImpl.cpp:
```cpp
/// Helper function for loop bufferization. Return the indices of all values
/// that have a tensor type.
static DenseSet<int64_t> getTensorIndices(ValueRange values) {
  DenseSet<int64_t> result;
  for (const auto &it : llvm::enumerate(values))
    if (isa<TensorType>(it.value().getType()))
      result.insert(it.index());
  return result;
}
```

And even it's not listed in our guideline, but could be as a reference here. Usually, if it's possible, prefer return output values to output parameters.
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#rf-out

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
rniczh: Good catch! Yes, this api definitely fit here, I will update it, thanks for the suggestion!

rniczh:
```cpp
return OperationEquivalence::isEquivalentTo(lhsDef, rhsDef,
[](Value l, Value r) -> LogicalResult { return success(observablesEqual(l, r)); },
nullptr, ... some flags ...);
```

rniczh: I guess we need to set the flags to `IgnoreLocations`, since we only care about the structure.

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
rniczh: Interesting 🤔 Actually, I already removed things related to this before opened the PR. I will just remove it.

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
rniczh: Done. https://github.com/PennyLaneAI/catalyst/pull/2657/commits/b6373bc9e42225babf05032396775375a8b8810f

mlir/lib/Quantum/Transforms/split_non_commuting.cpp
rniczh: Done. https://github.com/PennyLaneAI/catalyst/pull/2657/commits/b6373bc9e42225babf05032396775375a8b8810f

mlir/test/Quantum/SplitNonCommutingWiresTest.mlir
rniczh: Sure thing. https://github.com/PennyLaneAI/catalyst/pull/2657/commits/c1aad5593fd0b3774cde46d22fad7e7a8e382949

mlir/test/Quantum/SplitNonCommutingWiresTest.mlir
rniczh: https://github.com/PennyLaneAI/catalyst/pull/2657/commits/c1aad5593fd0b3774cde46d22fad7e7a8e382949

paul0403 (approved):

PR #2658, opened 2026-04-07
Private decomp rules

kipawaa:
**Context:**
Decomposition rules in the IR are not intended for outside use, but are marked public. This prevents optimizations like symbol-dce from removing them, and exposes unnecessary interfaces.

**Description of the Change:**
Changes the visibility of `DecompositionRules` to `private`.

**Benefits:**
Decomposition functions have appropriate visibility, and can be cleaned up by passes such as `symbol-dce`.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
[sc-115881]

PR #2660, opened 2026-04-07
Remove public option from lower_callable

maliasadi:
**Context:**
Following up on #2658 

- Remove public option from lower_callable

[sc-115881]


kipawaa (approved):
Good catch, thanks!

paul0403 (approved):

PR #2665, opened 2026-04-08
Add NoiseOp to the qecl layer

multiphaseCFD:
### Before submitting

**Context:**

[sc-116013]

This PR introduces the `qecl.noise` operation to the QEC logical layer. This operation accepts a logical codeblock as input and returns a corresponding logical-layer codeblock.

Note that: this op is not intended to inject logical noise or errors but functions as a placeholder, enabling a simpler injection of noises into the QEC physical layer.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/frontend/catalyst/python_interface/dialects/qecl.py
+++ b/frontend/catalyst/python_interface/dialects/qecl.py
@@ -354,6 +354,33 @@ class EncodeOp(IRDLOperation):
 
 
 @irdl_op_definition
+class NoiseOp(IRDLOperation):
+    """An operation indicates where to inject noise."""
+
+    T: ClassVar = VarConstraint("T", anyLogicalCodeblock)
+
+    name = "qecl.noise"
+
+    in_codeblock = operand_def(T)
+
+    out_codeblock = result_def(T)
+
+    assembly_format = """
+            $in_codeblock attr-dict `:` type($in_codeblock)
+        """
+
+    def __init__(
+        self,
+        in_codeblock: LogicalCodeBlockSSAValue | Operation,
+    ):
+        operands = (in_codeblock,)
+
+        in_codeblock_type = get_logical_codeblock_type(in_codeblock)
+
+        super().__init__(operands=operands, result_types=(in_codeblock_type,))
+
+
+@irdl_op_definition
 class QecCycleOp(IRDLOperation):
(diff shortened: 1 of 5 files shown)


frontend/catalyst/python_interface/dialects/qecl.py
```diff
@@ -353,6 +353,33 @@ def __init__(
         )
 
 
+@irdl_op_definition
+class NoiseOp(IRDLOperation):
+    """An operation indicates where to inject noise."""
```
lillian542: Super minor, but something like this will "match" the format of the other descriptions more closely 😅 

```suggestion
    """Inject noise in a logical codeblock"""
```

lillian542: Actually for clarity maybe we should say "Inject physical noise on elements of a logical codeblock"? Just to be clear somewhere that `qecl.noise` isn't noise at the logical level?

mlir/include/QecLogical/IR/QecLogicalOps.td
```diff
@@ -124,6 +124,22 @@ def EncodeOp : QecLogical_Op<"encode", [AllTypesMatch<["in_codeblock", "out_code
     }];
 }
 
+def NoiseOp : QecLogical_Op<"noise", [AllTypesMatch<["in_codeblock", "out_codeblock"]>]> {
+    let summary = "An operation indicates where to inject noise.";
```
lillian542: Same comments as a above, all the other descriptions are actions ("Encode...", "Update...", "Extract...") rather than objects ("An operation..."), and it might be clearer if we specifically note that the noise itself occurs at the physical layer, even if we are choosing an injection point at the logical level.

lillian542 (approved):
I have one minor (and very optional) suggestion about the docstrings, see what you think. Looks good to me!

mudit2812 (approved):
Thanks @multiphaseCFD . The changes look good to me.

frontend/catalyst/python_interface/dialects/qecl.py
multiphaseCFD: Thanks Lillian. I prefer "Inject physical noise on elements of a logical codeblock". Updated

PR #2667, opened 2026-04-08
Support adjoint of operators who decompose with control flow

### LINKED ISSUE #2669: Templates with control flow decompositions cannot be adjointed
```py
import pennylane as qp

class MyOperator(qp.operation.Operation):

    def decomposition(self):
        with qp.queuing.AnnotatedQueue() as q:
            _my_decomposition(self.wires)

        if qp.QueuingManager.recording():
            for op in q.queue:
                qp.apply(op)

        return q.queue

def _my_decomposition(wires, **_):

    qp.H(wires[0])

    @qp.for_loop(1, len(wires))
    def _loop(i):
        array_wires = qp.math.array(wires, like=i)
        qp.CNOT(wires=[array_wires[i - 1], array_wires[i]])

    _loop()

@qp.qjit
@qp.qnode(qp.device("lightning.qubit", wires=4))
def circuit():
    MyOperator([0, 1, 2, 3])
    qp.adjoint(MyOperator)([0, 1, 2, 3])

    return qp.state()
```

```
catalyst.utils.exceptions.CompileError: Cannot compile PennyLane inverse of the hybrid op <class 'catalyst.api_extensions.control_flow.ForLoop'>.
```

dime10:
Currently, something like the following fails with an error:

```py
import pennylane as qp

class MyOperator(qp.operation.Operation):

    def decomposition(self):
        with qp.queuing.AnnotatedQueue() as q:
            _my_decomposition(self.wires)

        if qp.QueuingManager.recording():
            for op in q.queue:
                qp.apply(op)

        return q.queue

def _my_decomposition(wires, **_):

    qp.H(wires[0])

    @qp.for_loop(1, len(wires))
    def _loop(i):
        array_wires = qp.math.array(wires, like=i)
        qp.CNOT(wires=[array_wires[i - 1], array_wires[i]])

    _loop()

@qp.qjit
@qp.qnode(qp.device("lightning.qubit", wires=4))
def circuit():
    MyOperator([0, 1, 2, 3])
    qp.adjoint(MyOperator)([0, 1, 2, 3])

    return qp.state()
```

```
catalyst.utils.exceptions.CompileError: Cannot compile PennyLane inverse of the hybrid op <class 'catalyst.api_extensions.control_flow.ForLoop'>.
```

The reason is that the adjoint is applied directly to an operator, generating a symbolic `Adjoint()` op instance, whereas applying adjoint to functions generates `HybridAdjoint()` instances with a separate body region. While the latter has support for computing the adjoint of control flow operations in the compiler, the former (in Python PennyLane) does not.
This hasn't been a problem so far because `ForLoop` instances are not manipulated directly in user workflows, and so adjoint with control flow would have always generated the function version. With control flow appearing as part of decompositions however it's possible for `Adjoint(ForLoop)` instances to be generated now, for example when an adjointed operator like a template needs to be decomposed.

The fix consists of automatically "decomposing" `Adjoint(ForLoop)` into `HybridAdjoint([ForLoop])`. The same is done for while loops and conditionals. However, switch statements are not yet included because the adjoint algorithm in catalyst does not have support for it (since it was added later).

closes #2669 
[sc-116053]
[sc-116179]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -474,6 +474,11 @@
   tools like `xdsl-opt` to work with Catalyst's custom Python dialects.
   [(#2471)](https://github.com/PennyLaneAI/catalyst/pull/2471)
 
+* Fix symbolic adjoint support for control flow operation. This means operators who are the target
+  of `qp.adjoint` but require decomposition can have decompositions with control flow in them,
+  which would previously raise an error. Adjoint on functions is unaffected.
+  [(#2667)](https://github.com/PennyLaneAI/catalyst/pull/2667)
+
 * Fix a bug with the xDSL `ParitySynth` pass that caused failure when the QNode being transformed
   contained operations with regions.
   [(#2408)](https://github.com/PennyLaneAI/catalyst/pull/2408)
--- a/frontend/catalyst/api_extensions/control_flow.py
+++ b/frontend/catalyst/api_extensions/control_flow.py
@@ -1572,6 +1572,9 @@ class Cond(HybridOp):
     has_adjoint = True
 
     def adjoint(self):
+        """Produce an adjoint version of this operator. Here, we simply regenerate a HybridAdjoint
+        version of the operation, which is generally supported by Catalyst."""
+
         return qml.adjoint(lambda: qml.apply(self) and None)()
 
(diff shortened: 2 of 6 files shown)


paul0403 (approved):
Nice!

kipawaa (approved):
Looks great, thanks!

PR #2668, opened 2026-04-09
Improve clang format

kipawaa:
**Context:**
There are currently no enforced rules for C++ header formatting, which results in manual formatting and unnecessarily separated blocks of headers, as well as unorganized categories.

This defines header groups in `.clang-format` to automatically sort and organize C++ headers throughout the repo. This sorts into 3 primary categories:

1. system headers (ex. `<vector>`, `<stdio.h>` etc)
2. dependencies (ex. `llvm/*`, `mlir/*`, `stablehlo/*`, `Enzyme.h` etc.)
3. local headers (ex. `Catalyst/*`, `Quantum/*` etc)
4. Private headers (ex. `ParameterShift.h`, in general includes without paths)
5. catch-all, as a way to indicate unsorted headers. This should always be empty!

An example of how these rules apply to _all_ headers found in Catalyst, view the `.headers.h` file added to this PR.

Some questions about sorting remain:
1. Do we want to keep the main header (i.e. the header matching the current implementation file) uniquely sorted to the top? This seems to be a consistent standard across LLVM and Google.
3. Do we consider things like `OQC` and `OQD` dependencies, or local headers?
4. Should the local headers category be a catch-all, or should we explicitly list our local header categories (similar to dependencies) and have fourth catch-all category that we expect to be empty, and can alert us to "unsorted" headers.

**Benefits:**
Consistent and automated header organization.

**Possible Drawbacks:**
Potential CI issues if not using auto-formatters. This should be mitigated by configuring an autoformatter (recommended `clang-format`) to use the project `.clang-format` file.

**Related GitHub Issues:**
N/A


mlir/lib/Ion/Transforms/ConversionPatterns.cpp
```diff
@@ -12,11 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <nlohmann/json.hpp>
-
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Dialect/LLVMIR/LLVMDialect.h"
 #include "mlir/IR/IRMapping.h"
+#include <nlohmann/json.hpp>
```
mlxd: Just wondering if we should handle `<>` separately from `" "`. And whether this should be in the former or the latter category

mlir/lib/Ion/IR/IonDialect.cpp
```diff
@@ -12,10 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "mlir/IR/DialectImplementation.h" // needed for generated type parser
+#include "Ion/IR/IonDialect.h"
```
mlxd, line 15: Should this be here?

mlir/lib/Gradient/Utils/GradientShape.cpp
```diff
@@ -12,10 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "mlir/Dialect/Tensor/IR/Tensor.h"
-
 #include "Gradient/Utils/GradientShape.h"
 
+#include "mlir/Dialect/Tensor/IR/Tensor.h"
```
mlxd, line 17: Should this be here?

mlir/lib/Gradient/Transforms/PostprocessingPatterns.cpp
```diff
@@ -13,8 +13,8 @@
 // limitations under the License.
 
 #include "iostream"
```
mlxd: Aside: Do we need this? If so, it should be <>

mlir/lib/Gradient/Transforms/BufferizableOpInterfaceImpl.cpp
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "Gradient/Transforms/BufferizableOpInterfaceImpl.h"
```
mlxd, line 15: Seems too high?

mlir/lib/Gradient/Transforms/annotate_invalid_gradient_functions.cpp
```diff
@@ -12,19 +12,20 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "Quantum/IR/QuantumOps.h"
+#include "Gradient/Transforms/annotate_invalid_gradient_functions.h"
```
mlxd, line 15: Does this go here, or below the mlir headers?

mlir/lib/Gradient/Transforms/GradMethods/PS_QuantumGradient.cpp
```diff
@@ -27,6 +25,7 @@
 
 #include "Catalyst/Utils/StaticAllocas.h"
 #include "Gradient/Utils/GradientShape.h"
+#include "ParameterShift.hpp"
```
mlxd: Should this be isolated to same-dir below the others?

mlir/lib/Gradient/Transforms/GradMethods/JVPVJPPatterns.cpp
```diff
@@ -14,13 +14,14 @@
 
 #define DEBUG_TYPE "jvpvjp"
 
+#include "JVPVJPPatterns.hpp"
```
mlxd, line 17: Should be at the bottom, no?

mlir/lib/Gradient/Transforms/GradMethods/HybridGradient.cpp
```diff
@@ -12,18 +12,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "HybridGradient.hpp"
```
mlxd, line 15: Does this belong here?

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "Adjoint.hpp"
```
mlxd, line 15: Just wondering if this should be on top or bottom?

mlxd (commented):
Thanks @kipawaa 
Didn't get a chance to look at it all, but seeing a pattern for a few of them. Can you heck the height of the single-include headers that are moved to the top?

mlir/lib/Catalyst/IR/CatalystDialect.cpp
```diff
@@ -12,15 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "llvm/ADT/TypeSwitch.h" // needed for generated type parser
+#include "Catalyst/IR/CatalystDialect.h"
```
dime10, line 15: Is this a miscategorization if it's taken away from the other Catalyst/* imports?

mlir/lib/Ion/Transforms/ConversionPatterns.cpp
kipawaa: This is an external file (from https://github.com/nlohmann/json) so I sorted this into dependencies, but we could re-asses this type of one-off includes or some other type of categorization.

mlir/lib/Ion/IR/IonDialect.cpp
kipawaa: This comment applies to a few of your questions, so I clarified some details in the PR description. Essentially, both LLVM and Google (and `clang-format` by default) sort the header file corresponding to the current implementation file to the top. I believe this has some actual importance, but I'm not too familiar with it (I'll do some more reading).

mlir/lib/Gradient/Utils/GradientShape.cpp
kipawaa: as above

mlir/lib/Gradient/Transforms/BufferizableOpInterfaceImpl.cpp
kipawaa: as above

mlir/lib/Gradient/Transforms/annotate_invalid_gradient_functions.cpp
kipawaa: as above

mlir/lib/Gradient/Transforms/GradMethods/JVPVJPPatterns.cpp
kipawaa: as above

mlir/lib/Gradient/Transforms/GradMethods/Adjoint.cpp
kipawaa: as above

mlir/lib/Gradient/Transforms/GradMethods/HybridGradient.cpp
kipawaa: as above

mlir/lib/Ion/Transforms/ConversionPatterns.cpp
dime10: Great question, I think we generally only use `<>` for stdlib libraries, and maybe "system" libraries that can be considered installed? 
I don't know how we want to classify the json lib specifically, or what rule we want to establish exactly, but it would be great if the formatter also enforced `<>` vs `""` :)

kipawaa: With respect to `<>` vs `""`, I've considered them identical for now but we could separate them. I think it would be reasonable to provide some standardization for when each is used (i.e. `<>` only for system/STL or something). Let me know if you have ideas on this. I would like to keep the general categories close, but we could have all includes with `<>` sorted above/below the includes with `""` within a given category or something like this?

dime10 (commented):
Such a quality of life improvement, thank you! 💯

mlir/lib/Ion/Transforms/ConversionPatterns.cpp
dime10: > <> only for system/STL or something

I'm happy with this, is there a way to enforce it (rather than use it as a classification indicator)?

mlir/lib/Ion/IR/IonDialect.cpp
dime10: From the llvm standards

> The [Main Module Header](https://llvm.org/docs/CodingStandards.html#main-module-header) file applies to .cpp files which implement an interface defined by a .h file. This #include should always be included first regardless of where it lives on the file system. **By including a header file first in the .cpp files that implement the interfaces, we ensure that the header does not have any hidden dependencies which are not explicitly #included in the header, but should be.** It is also a form of documentation in the .cpp file to indicate where the interfaces it implements are defined.

Now they do reverse the order of the categories though so it's not as jarring

dime10: Google on the other hand has our current order I believe: 
> Related header, C system headers, C++ standard library headers, other libraries' headers, your project's headers.

They list the same benefit (regarding current module header first, before the remainder of the headers): 
> With the preferred ordering, **if the related header dir2/foo2.h omits any necessary includes, the build of dir/foo.cc or dir/foo_test.cc will break**. Thus, this rule ensures that build breaks show up first for the people working on these files, not for innocent people in other packages.

mlir/lib/Ion/Transforms/ConversionPatterns.cpp
dime10: Google recommends the following:
> Headers should only be included using an angle-bracketed path if the library requires you to do so. In particular, the following headers require angle brackets:
> ```
> C and C++ standard library headers (e.g., <stdlib.h> and <string>).
> POSIX, Linux, and Windows system headers (e.g., <unistd.h> and <windows.h>).
> In rare cases, third_party libraries (e.g., <Python.h>).
> ```

mlir/lib/Ion/IR/IonDialect.cpp
kipawaa: Exactly, both Google and LLVM recommend the main/related header first, regardless of the order of other headers. I agree that it looks out of order at first, but with the listed benefits I think it makes sense to adhere to these standards.

mlir/lib/Catalyst/IR/CatalystDialect.cpp
kipawaa: This is discussed in greater detail in [this comment](https://github.com/PennyLaneAI/catalyst/pull/2668#discussion_r3058356423).

mlir/lib/Ion/Transforms/ConversionPatterns.cpp
kipawaa: I don't think it's possible to enforce this with clang-format, but we might be able to sort unmatched headers to a separate category and use a different program to verify that this category is empty? This would let us know which headers are not caught by the existing rules.

dime10: Yeah I'm leaning towards having a last category anyway for the unsorted headers so that it forces us to keep the formatter up to date :)

It's a shame that it's all regex based and can't work with things like filesystem structure, but oh well 🤷

mlir/lib/Gradient/Transforms/GradMethods/PS_QuantumGradient.cpp
dime10: That could be a nice fourth category to distinguish the private headers from the public ones

paul0403 (approved):
Nice!

mlir/lib/Ion/Transforms/ConversionPatterns.cpp
mlxd: Yep, this approach sounds good to me too. Thanks!

mlir/lib/Ion/IR/IonDialect.cpp
mlxd: If we are adopting this, can you update the internal guidelines then to match, since this contrasts a little with what is written

mlir/lib/Ion/Transforms/ConversionPatterns.cpp
kipawaa: As of [5eda3d6](https://github.com/PennyLaneAI/catalyst/pull/2668/commits/5eda3d67c2f9a103c6cf9f68da00ba3d2453d63e) I've enforced `<>` delimiters for sys/std and `""` delimiters for everything else, with an extra category to handle headers that don't match the other patterns. I've also added a (temporary) `.headers.h` in the root directory as a quick overview, which has all of the includes used in the repo sorted with the latest categories. Let me know what you think!

mlir/lib/Gradient/Transforms/GradMethods/PS_QuantumGradient.cpp
kipawaa: I've added a category for same-dir headers in [3c93256](https://github.com/PennyLaneAI/catalyst/pull/2668/commits/3c9325604250e15c1c73b8f8c5937c1bbb2c8187).

mlir/lib/Ion/Transforms/ConversionPatterns.cpp
dime10: Awesome, that makes it really easy! The bottom category should be empty then except for the inc headers I'm guessing? (how are you treating those btw, I'm assuming they're ignored or something since they usually need to appear in a particular position) 

From there we have:
```
#include "iostream"
#include <Exception.hpp>
#include <NullQubit.hpp>
#include <QuantumDevice.hpp>
#include <boost/multiprecision/cpp_dec_float.hpp>
#include <boost/multiprecision/cpp_int.hpp>
#include <boost/multiprecision/miller_rabin.hpp>
#include <boost/random.hpp>
#include <catch2/catch_session.hpp>
#include <catch2/catch_template_test_macros.hpp>
#include <catch2/generators/catch_generators_all.hpp>
#include <catch2/matchers/catch_matchers_floating_point.hpp>
#include <catch2/matchers/catch_matchers_range_equals.hpp>
#include <catch2/matchers/catch_matchers_vector.hpp>
#include <dlfcn.h>
#include <llvm/Target/TargetMachine.h>
#include <mlir/Dialect/Arith/IR/Arith.h>
#include <mlir/Dialect/Arith/IR/Arith.h> // for arith::XOrIOp and arith::ConstantOp
#include <mlir/Dialect/Arith/IR/Arith.h>
#include <mlir/Dialect/SCF/IR/SCF.h>
#include <mlir/Dialect/SCF/Transforms/Transforms.h>
#include <mlir/IR/Builders.h>
#include <mlir/IR/DialectRegistry.h>
#include <mlir/IR/Location.h>
#include <mlir/IR/PatternMatch.h>
#include <mlir/IR/Value.h>
#include <mlir/Support/LogicalResult.h>
#include <nanobind/eval.h>
#include <nanobind/nanobind.h>
#include <nanobind/stl/string.h>
#include <nlohmann/json.hpp>
#include <pybind11/embed.h>
#include <pybind11/eval.h>
#include <pybind11/pybind11.h>
#include <pybind11/stl.h>
#include <stim/mem/simd_word.h>
#include <stim/stabilizers/flex_pauli_string.h>
#include <stim/stabilizers/pauli_string.h>
#include <toml++/toml.hpp>
```
Most just need a switch to `""` or `<>`, except maybe `<dlfcn.h>`, isn't that a system header?

dime10: Also if you wouldn't mind renaming these to `.h` since that's the convention for all our public mlir includes (`mlir/include/`):
```
./Ion/Transforms/oqd_database_managers.hpp
./Ion/Transforms/oqd_database_utils.hpp
./Ion/Transforms/oqd_database_types.hpp
```

kipawaa: Right now `.inc` headers are in the catch-all category, since they typically need special treatment. The current formatting won't move any of those so long as they are separated by code from the rest of the headers. I'll go about fixing delimiters, add dlfcn to the sys/std category, and fix transitive import issues that are raised by re-ordering the headers (this is why CI is failing) soon!

dime10: Awesome, thanks River 🎉

mlir/lib/Gradient/Transforms/PostprocessingPatterns.cpp
kipawaa: Fixed! Good catch :)

mlir/lib/Ion/Transforms/ConversionPatterns.cpp
kipawaa: The latest updates reflect all of these changes and CI is passing 😄 if all seems well, we can remove the example file and merge!

mlir/include/Ion/Transforms/oqd_database_managers.h
```diff
@@ -18,10 +18,8 @@
 #include <cassert>
 #include <set>
 
-#include <toml++/toml.hpp>
```
dime10, line 21: I think one guideline i usually see is not to rely on transitive dependencies, so if this header uses the toml library (which it does), it should include it right?

mlir/include/Quantum/Transforms/BufferizableOpInterfaceImpl.h
```diff
@@ -14,7 +14,7 @@
 
 #pragma once
 
-using namespace mlir;
+#include "mlir/IR/DialectRegistry.h"
```
dime10, line 17: thanks! worth pointing out using directives should never appear in headers 😅

mlir/include/RegisterAllPasses.h
```diff
@@ -14,6 +14,8 @@
 
 #pragma once
 
+#include "hlo-extensions/Transforms/Passes.h"
```
dime10: So maybe counter-intuitively, the extensions are (local) project code (even though they are about hlo), whereas the stablehlo project itself is third-party.
But maybe thematically we'd want it grouped with stablehlo imports (when there are any)? I'm not sure  🤔

mlir/include/CAPI/Dialects.h
```diff
@@ -14,7 +14,7 @@
 
 #pragma once
 
-#include "mlir/CAPI/Registration.h"
+#include "mlir-c/IR.h"
```
dime10, line 17: what prompted this change?

.headers.h
```diff
@@ -0,0 +1,574 @@
+#include <algorithm>
+#include <array>
+#include <bitset>
+#include <cassert>
+#include <chrono>
+#include <cmath>
+#include <complex.h>
+#include <complex>
```
dime10, line 8: Are both versions needed I wonder?

.headers.h
```diff
@@ -0,0 +1,574 @@
...
+#include <tuple>
+#include <type_traits>
+#include <unordered_map>
+#include <unordered_set>
+#include <utility>
+#include <variant>
+#include <vector>
+
+#include "Enzyme.h"
+#include "absl/base/dynamic_annotations.h"
```
dime10, line 58: apparently we don't use this, ideally we'd delete that code altogether (out of scope though)

dime10 (approved):
What a blessing to finally have this automated 🤩 

It seems you simplified some of the includes, I'm curious if you used some tool for this or were superfluous headers manually removed / swapped out?

.headers.h
kipawaa: `complex.h` is actually only included in a frontend python file and docs 😅 I just `grep`ped the entire repo for `#include` to generate this file, so it grabbed a couple of non-cpp statements as well.

.headers.h
kipawaa: It seems that `absl` is used by the frontend jax kernels, not sure I've run into those directly though so I'm not sure the significance or feasibility of removing it.

mlir/include/CAPI/Dialects.h
kipawaa: Clangd was complaining about `CAPI/Registration` being an unused import (i.e. it's only used transitively). I switched to this and it was satisfied. AFAIK `mlir-c` is public and as stable as the rest of MLIR, so I hope this is alright? I was going through and fixing simple header changes while I was already in there.

mlir/include/Ion/Transforms/oqd_database_managers.h
kipawaa: Woops! I was trying to simplify the headers as I went through and my LSP flagged that one as not used directly but it's definitely needed here. Thanks for catching that! Re-added in [921da36](https://github.com/PennyLaneAI/catalyst/pull/2668/commits/921da367032909e7c381c3e2580bb076d04dbaab)

mlir/include/RegisterAllPasses.h
kipawaa: Ah that's true. I think I'll re-sort them to local, since I think it helps to see what we have control over grouped together. Changed and sorted in [8b7dccd](https://github.com/PennyLaneAI/catalyst/pull/2668/commits/8b7dccd1f19260dd57c1ce141fc674b2db0985b0) and updated example file in [22f08ff](https://github.com/PennyLaneAI/catalyst/pull/2668/commits/22f08ff819befe3ee97dc4f0bf60324bc8812412)

kipawaa:
> I'm curious if you used some tool for this or were superfluous headers manually removed / swapped out

@dime10 I was using the `clangd` language server to identify unused imports, but the removals etc. were done manually. We could introduce (or just use internally) IWYU (include what you use) to automate header locating, inclusion and redundant header removal, I haven't used it (yet) but it seems like a great tool!

.headers.h
dime10: They are commented out via preprocessor macro, so we are already not using the associated code ^^
https://github.com/PennyLaneAI/catalyst/blob/66a46c43ed11c9116e71b167913b64223047d752/frontend/catalyst/utils/jax_cpu_lapack_kernels/lapack_kernels.cpp#L46-L48

dime10:
> > I'm curious if you used some tool for this or were superfluous headers manually removed / swapped out
> 
> @dime10 I was using the `clangd` language server to identify unused imports, but the removals etc. were done manually. We could introduce (or just use internally) IWYU (include what you use) to automate header locating, inclusion and redundant header removal, I haven't used it (yet) but it seems like a great tool!

Cool cool, yeah something for another pr perhaps 😄

mlir/include/CAPI/Dialects.h
dime10: Yeah all good 👍 Just weird because you'd think that registration.h is more specific than IR.h? (or at least more meaningfully named) But I'm not going to lose any sleep over it 😅

PR #2674, opened 2026-04-09
[qref 1.1] Verify that qubits used by gates have not been dealloced already

paul0403:
**Context:**
One of the things that is much easier to do in reference semantics is verifying that there are no "use-after-free" qubits.

In value semantics, this was difficult to do, because one would need to walk back the qubit value chain to get to the register (and this walk back could even cross control flow boundaries), in order to look for the dealloc. In reference semantics this is not necessary.

Currently, we do not do this verification anywhere regarding the implicit global initial register from device init. And regarding the dynamically allocated registers, we do this "verification" in the frontend, as part of the plxpr conversion logic: https://github.com/PennyLaneAI/catalyst/blob/9e4b582ecf2b84a7d69bcc9c348febde43d84ac4/frontend/catalyst/from_plxpr/qfunc_interpreter.py#L179

But it is through the hacky `QubitHandler` object that manages the plxpr conversion's qubit flow: https://github.com/PennyLaneAI/catalyst/blob/9e4b582ecf2b84a7d69bcc9c348febde43d84ac4/frontend/catalyst/from_plxpr/qubit_handler.py#L180

Such an obligation should really be done in mlir, instead of a side-quest of the frontend's plxpr conversion.

Add an mlir pass to verify that the qref gate-like operations' qubit operands have not been deallocated.

**Benefits:**
- We can catch use-after-free errors!
- One step towards cleaner plxpr conversion in the frontend

**Possible Drawbacks:**
We compute the dominance info to see if the dealloc op dominates the use. Computing the dominance info is not a trivial process, and doing it in the verifier might drop performance by quite a bit. This is why we do it as a one-time pass (to be placed at the start of the pipeline), instead of in the ops' verifiers themselves. 

But this will only catch bad programs from the frontend. and any mlir passes that produce use-after-free is still uncatched. 

Once we finish the better-container-for-quantum-program epic, we should move this verification to be an op-verification on the quantum kernel op.


mlir/lib/QRef/Transforms/verify_no_quantum_use_after_free.cpp
```diff
@@ -0,0 +1,88 @@
...
+#include "QRef/IR/QRefOps.h"
+
+using namespace mlir;
+using namespace catalyst;
+
+namespace {
+bool hasUseAfterFree(Value qubit, Operation *gate, DominanceInfo &domInfo)
+{
+    if (qubit.getDefiningOp() && isa<qref::GetOp>(qubit.getDefiningOp())) {
+        auto getOp = cast<qref::GetOp>(qubit.getDefiningOp());
```
rniczh: I think you can do this in one line. What do you think?

```cpp
    if (auto getOp = qubit.getDefiningOp<qref::GetOp>()) {
```

rniczh (commented):
Nice @paul0403 ! Just only one comment from my side.

mlir/lib/QRef/Transforms/verify_no_quantum_use_after_free.cpp
```diff
@@ -0,0 +1,88 @@
...
+                if (hasUseAfterFree(qubit, qOp, domInfo)) {
+                    qOp.emitOpError("Detected use of a qubit after deallocation");
+                    return WalkResult::interrupt();
+                }
+            }
+            return WalkResult::advance();
+        });
+        if (wr.wasInterrupted()) {
+            return signalPassFailure();
+        }
```
kipawaa, line 82: Rather than an op-agnostic pass that then walks the module for the op type it needs, can we change this to a filtered operation pass? See https://mlir.llvm.org/docs/PassManagement/#filtered-operation-pass for details

mlir/lib/QRef/Transforms/verify_no_quantum_use_after_free.cpp
paul0403: Nice suggestion! https://github.com/PennyLaneAI/catalyst/pull/2674/commits/dd496207c1f905e739d31d02c72eb230bcf2e7b4

rniczh (approved):
🥳

mlir/test/QRef/VerifyNoQuantumUseAfterFreeTest.mlir
kipawaa, line 1: Could you add a test that ensures no errors are raised when there are no use-after-free issues?

mlir/lib/QRef/Transforms/verify_no_quantum_use_after_free.cpp
paul0403: Discussed, it makes more sense to use as a module pass so that the dominance info is not recomputed for each gate.

mlir/test/QRef/VerifyNoQuantumUseAfterFreeTest.mlir
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2674/commits/649bda0edd709336ea9f95218c64c90b8da026c1

kipawaa (approved):
This is great, thank you!

PR #2675, opened 2026-04-09
Update the resource analysis pass to produce a format more easily ingested by specs

jzaia18:
**Context:**
Integrating the Catalyst resource analysis pass with `qp.specs` is a goal for the next release of PennyLane. Some changes to the resource analysis pass will be necessary to facilitate this integration

* Updates the resource analysis pass to be able to write out to a file, rather than dumping to stdout
* Adds wire information to measurement results
* Adds information about dynamic control flow to resource analysis output
* Renames "ResourceTrackerPass" to "ResourceAnalysisPass"

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-115506]


sengthai (approved):
Since the most of code are already approved in #[2695](https://github.com/PennyLaneAI/catalyst/pull/2695), Happy to approve this one!

mlir/lib/Catalyst/Transforms/ResourceTrackerPass.cpp
```diff
@@ -109,64 +127,109 @@ struct ResourceTrackerPass : public impl::ResourceTrackerPassBase<ResourceTracke
...
 
-        // TODO: write to file, when called from frontend. Then, frontend read and delete the file.
-        llvm::json::Value jsonValue(std::move(root));
-        llvm::outs() << llvm::formatv("{0:2}", jsonValue) << "\n";
+        std::string file_name = "__mlir_resources_";
+        file_name += jit_fn_name;
+
+        auto now = std::chrono::system_clock::now();
+        std::time_t now_time_t = std::chrono::system_clock::to_time_t(now);
+        std::tm *now_tm = std::localtime(&now_time_t);
```
maliasadi: This is not thread safe. As we're building with C++20, consider

```suggestion
        auto now = std::chrono::system_clock::now();
        auto zt = std::chrono::zoned_time{std::chrono::current_zone(), now};
        auto now_tm = zt.get_local_time();

```

mlir/lib/Catalyst/Transforms/ResourceTrackerPass.cpp
```diff
@@ -109,64 +127,109 @@ struct ResourceTrackerPass : public impl::ResourceTrackerPassBase<ResourceTracke
...
 
-            funcObj["num_qubits"] = static_cast<int64_t>(result.numQubits());
-            funcObj["num_alloc_qubits"] = static_cast<int64_t>(result.numAllocQubits);
-            funcObj["num_arg_qubits"] = static_cast<int64_t>(result.numArgQubits);
-            funcObj["device_name"] = result.deviceName;
+    /// Write JSON to a file.
+    static void writeJsonToFile(const std::string &jsonStr, const std::string &fileName)
+    {
+        std::ofstream ofile(fileName);
+        assert(ofile.is_open() && "Invalid file to store resource results");
```
maliasadi: Could we use `llvm::errs` instead of assert for a clearer compiler error message here?
 
```suggestion
        if (!ofile.is_open()) {
            llvm::errs() << "Error: could not open resource output file: " << fileName << "\n";
            return;
        }
```

mlir/lib/Catalyst/Utils/ConstantResolve.cpp
```diff
@@ -0,0 +1,166 @@
...
+    }
+    if (auto denseInt = dyn_cast<DenseIntElementsAttr>(attr)) {
+        if (denseInt.isSplat() || denseInt.getNumElements() == 1) {
+            return static_cast<double>(denseInt.getSplatValue<APInt>().getSExtValue());
+        }
+    }
+    return std::nullopt;
+}
+
+inline std::optional<double> resolveConstantArithmetic(Value val, Operation *op)
```
maliasadi: `inline` has no effect in C++ files

```suggestion
std::optional<double> resolveConstantArithmetic(Value val, Operation *op)
```

mlir/include/Catalyst/Transforms/Passes.td
```diff
@@ -270,6 +270,20 @@ def ResourceTrackerPass : Pass<"resource-tracker"> {
             /*CLI arg name=*/"output-json",
```
maliasadi, line 270: `output-json`, `print-json` and `output-fname` are a bit confusing! could we reduce the number of options to two? For example, `print-json` and `output-file`

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -168,6 +203,9 @@ ResourceAnalysis::ResourceAnalysis(Operation *op)
             }
         }
 
+        // TODO: deprecate `qnode` once it is fully deprecated.
```
maliasadi, line 206: A shortcut ticket for this TODO?

mlir/lib/Catalyst/Utils/ConstantResolve.cpp
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    if (opName == "stablehlo.add" || opName == "stablehlo.subtract") {
+        auto lhs = resolveConstant(op->getOperand(0));
+        auto rhs = resolveConstant(op->getOperand(1));
+        if (!lhs || !rhs) {
+            return std::nullopt;
+        }
+        return (opName == "stablehlo.add") ? *lhs + *rhs : *lhs - *rhs;
+    }
```
maliasadi, line 144: Could you fix the CodeFactor "Complex Method" warning?

mlir/include/Catalyst/Transforms/Passes.td
jzaia18: This is similar to my original thought as well. Originally I planned to use just `output-json` and `json-fname`, where we would write to stdout if the name is left blank. Thoughts on switching to this style @sengthai ?

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
jzaia18: CC @sengthai

mlir/include/Catalyst/Transforms/Passes.td
sengthai: these three flags give users options to print json or write to json with/without filepath. we can reduce that options to `output-json and json-fname`. For me, it seems both are dependent to each other example users may only use `json-fname` then which mean `output-json` will be activate. 

To be more concrete, I think `print-json` and `output-file` are also good idea. The thing is that user need to pass the file path to `output-file` in order to get a json file. if that okay in frontend, we can reduce to that options.

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: Let's me ask Jukebox to create it one. 😅

mlir/include/Catalyst/Transforms/Passes.td
jzaia18: For specs we will always provide a filename, so no concerns there. Are there any situations where we would want the pass to choose the filename automatically?

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: https://app.shortcut.com/xanaduai/story/117597/deprecate-qnode-attribute-check-in-resourceanalysis

mlir/include/Catalyst/Transforms/Passes.td
sengthai: Ahh then it would be fine.

maliasadi (approved):
Thanks @jzaia18! 🏖️

doc/releases/changelog-dev.md
```diff
@@ -183,15 +183,17 @@
...
   cacheable MLIR analysis class (`ResourceAnalysis`) that other transformation passes can query
   via `getAnalysis<ResourceAnalysis>()`, avoiding redundant recomputation.
   [(#2479)](https://github.com/PennyLaneAI/catalyst/pull/2479)
+  [(#2675)](https://github.com/PennyLaneAI/catalyst/pull/2675)
+  [(#2695)](https://github.com/PennyLaneAI/catalyst/pull/2695)
 
   ```bash
-  quantum-opt --resource-tracker='output-json=true' input.mlir
-  quantum-opt --resource-tracker -mlir-pass-statistics input.mlir
+  quantum-opt --resource-analysis='print-json=true' input.mlir
```
```suggestion
  quantum-opt --resource-analysis='output-json=true' input.mlir
```

PR #2676, opened 2026-04-10
Ignore deprecation warning in upstream

dime10:
Most recent version of clang warns about deprecated code in the toml dependency, which breaks our build (warnings as errors). Since we have no control over upstream, we silence it for affected modules only.


lazypanda10117 (approved):

paul0403 (commented):
I think there's two of them, the other is in PBC dialect?

kipawaa (approved):
Thanks for fixing this!

dime10:
> I think there's two of them, the other is in PBC dialect?

apparently only this one triggers it (@kipawaa confirmed this)

PR #2678, opened 2026-04-10
Add a pass to lower `qecp.noise` to subroutine in `qecp` dialect

multiphaseCFD:
### Before submitting

**Context:**
[sc-116060]

- This pass lowers `qecp.noise` operations in the `qecl` layer to subroutine calls in the `qecp` layer. 
- Users are allowed to pass the number of errors (`number_errors`) per noiseop injected into a `codeblock`. The other option is the number of physical data qubits (`n`) in the codeblock and is expected to be specified when lowering the `qecl` layer to the  `qecp` layer.
- Noises are mimicked by injecting `qecp.rot` gates with random angle parameters and random qubit indices into codeblocks for each `qecl.noise` operation.
-  The randomness is from the python module. Note that it's feasible to make use of `jax.random.rng.uniform` module to generate those random numbers but it's out of scope for the first prototyping.
- The integration tests are skipped here, as we can't do it util `quantum-to-qecl` conversion is ready. 

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


frontend/catalyst/python_interface/transforms/qecp/convert_noiseop_to_subroutine.py
```diff
@@ -0,0 +1,296 @@
...
+    """Get the symbol name of the noise injection subroutine for a given codeblock type.
+    Args:
+        k (int): The number of logical qubits of a physical codeblock.
+        n (int): The number of physical data qubits of a physical codeblock.
+        number_errors (int): The number of errors to be injected.
+    Returns:
+        str: The symbol name of the noise injection subroutine.
+    """
+
+    return "noise_subroutine_code" + str(k) + "x" + str(n) + "x" + str(number_errors)
```
joeycarter: f-strings are generally preferred these days.

[Optional] I also added an underscore before `k` for better readability.

```suggestion
    return f"noise_subroutine_code_{k}x{n}x{number_errors}"
```

frontend/catalyst/python_interface/transforms/qecp/convert_noiseop_to_subroutine.py
```diff
@@ -0,0 +1,296 @@
...
+
+
+class ConvertNoiseOpToSubroutinePattern(
+    pattern_rewriter.RewritePattern
+):  # pylint: disable=too-few-public-methods
+    """RewritePattern for converting to the MBQC formalism."""
```
joeycarter: Looks like you forgot to update the docstring 😉

frontend/catalyst/python_interface/transforms/qecp/convert_noiseop_to_subroutine.py
```diff
@@ -0,0 +1,296 @@
...
+        self,
+        op: qecl.NoiseOp,
+        rewriter: pattern_rewriter.PatternRewriter,
+        /,
+    ):
+        k = op.in_codeblock.type.k.value.data
+
+        in_block_cast = builtin.UnrealizedConversionCastOp.get(
+            op.in_codeblock, qecp.PhysicalCodeblockType(k, self._n)
+        )
```
joeycarter, line 70: `builtin.UnrealizedConversionCastOp.get` expects sequences for its arguments:

```suggestion
        in_block_cast = builtin.UnrealizedConversionCastOp.get(
            (op.in_codeblock,), (qecp.PhysicalCodeblockType(k, self._n),)
        )
```

frontend/catalyst/python_interface/transforms/qecp/convert_noiseop_to_subroutine.py
```diff
@@ -0,0 +1,296 @@
...
+        arguments = [
+            in_block_cast.results[0],
+            qubit_indices_constantop.results[0],
+            rotation_params_constantop.results[0],
+        ]
+
+        return_types = self.noise_subroutine.function_type.outputs.data
+        callOp = func.CallOp(callee, arguments, return_types)
+        rewriter.insert_op(callOp, InsertPoint.before(op))
+        cast_op = builtin.UnrealizedConversionCastOp.get(callOp.results[0], op.out_codeblock.type)
```
joeycarter: Same here

```suggestion
        cast_op = builtin.UnrealizedConversionCastOp.get((callOp.results[0],), (op.out_codeblock.type,))
```

frontend/catalyst/python_interface/transforms/qecp/convert_noiseop_to_subroutine.py
```diff
@@ -0,0 +1,296 @@
...
+        rewriter.insert_op(cast_op, InsertPoint.before(op))
+        rewriter.replace_all_uses_with(op.out_codeblock, cast_op.results[0])
+        rewriter.erase_op(op)
+
+
+class ConvertNoiseOpToSubroutinePass(passes.ModulePass):
```
joeycarter, line 126: `ModulePass` is a frozen `dataclass` so we should add the decorator here too:

```suggestion
@dataclass(frozen=True)
class ConvertNoiseOpToSubroutinePass(passes.ModulePass):
```

frontend/catalyst/python_interface/transforms/qecp/convert_noiseop_to_subroutine.py
```diff
@@ -0,0 +1,296 @@
...
+        The options include:
+        1. number_errors: the number of errors to be injected for each noise operation.
+        Defaults to 1, which is the number of errors that can be corrected by the Steane
+        code.
+        2. n: the number of physical data qubits of per codeblocks. Defaults to 7, which
+        is the number of physical data qubits for the Steane code. NOTE: this option is
+        expected to be specified in the `qecl-to-qecp` skeleton pass.
+        """
+        self._number_errors = options.get("number_errors", 1)
+        self._n = options.get("n", 7)
```
joeycarter: Since the pass is a dataclass, we don't need the constructor:

```suggestion
    n: int  # The number of physical data qubits per codeblock [...]

    number_of_errors: int = 1  # Number of errors to inject [...]
```

I dropped the default value for `n`; that seems error-prone for when we eventually move beyond the Steane code. Defaulting to 1 error per noise op seems sensible, though.

frontend/catalyst/python_interface/transforms/qecp/convert_noiseop_to_subroutine.py
```diff
@@ -0,0 +1,296 @@
...
+
+    def _create_noise_subroutine(self, k, n, number_errors):
```
joeycarter: We probably don't need the `n` and `number_errors` arguments here, it seems more appropriate to use the class member variables directly.

```suggestion
    def _create_noise_subroutine(self, k):
```

frontend/catalyst/python_interface/transforms/qecp/convert_noiseop_to_subroutine.py
```diff
@@ -0,0 +1,296 @@
...
+
+        Returns:
+            The corresponding subroutine (func.FuncOp).
+        """
+        # 1. Define the input and output types of the subroutine
+        # The input types include:
+        # 1, a physical codeblock;
+        # 2, number of errors (rot operations);
+        # 3, a tensor containing rotation parameters for rot operations (errors).
+        codeblock_type = qecp.PhysicalCodeblockType(k, n)
```
joeycarter: e.g.

```suggestion
        codeblock_type = qecp.PhysicalCodeblockType(k, self.n)
```

frontend/catalyst/python_interface/transforms/qecp/convert_noiseop_to_subroutine.py
```diff
@@ -0,0 +1,296 @@
...
+                    indices=[index_var_int, two],
+                    result_type=rotation_params.type.element_type,
+                )
+
+                # Extract a phyiscal qubit from the current codeblock
+                extracted_phyiscal_qubit = qecp.ExtractQubitOp(current_codeblock, qubit_index)
+                # Create the Rot operation with the extracted qubit for error injection
+                rot_op = qecp.RotOp(phi, theta, omega, extracted_phyiscal_qubit)
+
+                # Insert the phyiscal qubit with noise back to the codeblock
```
joeycarter:
```suggestion
                # Extract a physical qubit from the current codeblock
                extracted_physical_qubit = qecp.ExtractQubitOp(current_codeblock, qubit_index)
                # Create the Rot operation with the extracted qubit for error injection
                rot_op = qecp.RotOp(phi, theta, omega, extracted_physical_qubit)

                # Insert the physical qubit with noise back to the codeblock
```

frontend/catalyst/python_interface/transforms/qecp/convert_noiseop_to_subroutine.py
```diff
@@ -0,0 +1,296 @@
...
+                    if isinstance(op_, qecl.NoiseOp):
+                        k = op_.in_codeblock.type.k.value.data
+                        break
+        # Skip the conversion if there is no noise operation in the module.
+        if k is None:
+            return
+
+        # Insert a noise injection subroutine into the module.
+        noise_subroutine = self._create_noise_subroutine(k, self._n, self._number_errors)
+        op.regions[0].blocks.first.add_op(noise_subroutine)
```
joeycarter, line 278: Mind adding this assert? `RegionBlocks.first` can return None.

```suggestion
        assert op.regions[0].blocks.first is not None
        op.regions[0].blocks.first.add_op(noise_subroutine)
```

frontend/catalyst/python_interface/transforms/qecp/convert_noiseop_to_subroutine.py
```diff
@@ -0,0 +1,296 @@
...
+
+        pattern_rewriter.PatternRewriteWalker(
+            pattern_rewriter.GreedyRewritePatternApplier(
+                [ConvertNoiseOpToSubroutinePattern(noise_subroutine, self._n, self._number_errors)]
+            ),
+            apply_recursively=False,
+        ).rewrite_module(op)
```
joeycarter, line 283: I don't think we need a greedy rewrite-pattern applier here, since the pattern only needs to be applied once. Greedy is used in cases where there are multiple patterns or where applying the rewrite pattern puts the IR into a state where the same pattern can be applied again and again until some condition is hit (like when the pattern no longer modifies the IR). Might want to test this first before committing the change 🙂 

```suggestion
        pattern_rewriter.PatternRewriteWalker(
            ConvertNoiseOpToSubroutinePattern(noise_subroutine, self._n, self._number_errors),
            apply_recursively=False,
        ).rewrite_module(op)
```

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_noiseop_to_subroutine.py
```diff
@@ -0,0 +1,216 @@
...
+import pytest
+
+from catalyst.python_interface.transforms.qecp import (
+    ConvertNoiseOpToSubroutinePass,
+)
+
+pytestmark = pytest.mark.xdsl
+
+
+class TestConvertNoiseOpToSubroutinePass:
```
joeycarter, line 28: Mind adding this warning filter here? (Since we expect these tests to leave behind `builtin.unrealized_conversion_cast` ops that can't be reconciled)

```suggestion
@pytest.mark.filterwarnings("ignore:Unable to remove cast UnrealizedConversionCastOp")
class TestConvertNoiseOpToSubroutinePass:
```

frontend/catalyst/python_interface/transforms/qecp/convert_noiseop_to_subroutine.py
```diff
@@ -0,0 +1,296 @@
...
+        # pass was added to the pipeline.
+        symbol_name = _get_noise_subroutine_name(k, n, number_errors)
+        funcOp = func.FuncOp(
+            symbol_name,
+            (input_types, output_types),
+            visibility="private",
+            region=region,
+        )
+        # Add an attribute to the noise injection subroutine
+        funcOp.attributes[_get_noise_subroutine_name(k, n, number_errors)] = builtin.NoneAttr()
```
joeycarter: Maybe a `UnitAttr` would be more appropriate here? Like we have for the `quantum.node` attribute. I think this should work:

```suggestion
        funcOp.attributes[_get_noise_subroutine_name(k, n, number_errors)] = UnitAttr.get()
```

frontend/catalyst/python_interface/transforms/qecp/convert_noiseop_to_subroutine.py
```diff
@@ -0,0 +1,296 @@
...
+        cast_op = builtin.UnrealizedConversionCastOp.get(callOp.results[0], op.out_codeblock.type)
+        rewriter.insert_op(cast_op, InsertPoint.before(op))
+        rewriter.replace_all_uses_with(op.out_codeblock, cast_op.results[0])
+        rewriter.erase_op(op)
+
+
+class ConvertNoiseOpToSubroutinePass(passes.ModulePass):
+    """Pass that converts qecl.noise operations to subroutines in the qecp layer."""
+
+    name = "convert-noiseop-to-subroutine"
```
joeycarter: What do you think about renaming this pass to something like `convert-qecl-noise-to-qecp-noise`? "`to-subroutine`" is fairly generic and someone from outside might reasonably wonder what the subroutine is doing, so maybe this name is more descriptive.

joeycarter (commented):
Looks great, thanks @multiphaseCFD! This is a very clearly written pass 🚀 

I only had a few minor comments, happy to approve once resolved.

frontend/catalyst/python_interface/transforms/qecp/convert_noiseop_to_subroutine.py
```diff
@@ -0,0 +1,296 @@
...
+    ):
+        k = op.in_codeblock.type.k.value.data
+
+        in_block_cast = builtin.UnrealizedConversionCastOp.get(
+            op.in_codeblock, qecp.PhysicalCodeblockType(k, self._n)
+        )
+        rewriter.insert_op(in_block_cast, InsertPoint.before(op))
+
+        # Create random qubit indices and rotation parameters for error injection, which are
+        # generated randomly from a `qnode` function.
```
lillian542: what does it mean that its "from a `qnode` function"?

frontend/catalyst/python_interface/transforms/qecp/convert_noiseop_to_subroutine.py
multiphaseCFD: ooops, I forgot to update it. Thanks

frontend/catalyst/python_interface/transforms/qecp/convert_noiseop_to_subroutine.py
multiphaseCFD: Nice catch!

frontend/catalyst/python_interface/transforms/qecp/convert_noiseop_to_subroutine.py
multiphaseCFD: agree, would `convert-qecl-noise-to-qecp-noise-subroutine` be more clear?

multiphaseCFD:
Will update the module name to avoid review info loss

frontend/catalyst/python_interface/transforms/qecp/convert_noiseop_to_subroutine.py
joeycarter: Phew that's a pretty long pass name :sweat_smile: 

I wouldn't say we need `subroutine` in the pass name, that's an implementation detail rather than a description of what the pass is for. I might even be in favour of `convert-qecl-noise-to-qecp`; that's more concise and still descriptive enough to understand what the pass is for.

Naming things is hard! I'll leave it to you :slightly_smiling_face:

joeycarter (approved):
Looks great, thanks @multiphaseCFD! 💯 

I trust your judgement on the pass name 😉

frontend/catalyst/python_interface/transforms/qecp/convert_noiseop_to_subroutine.py
```diff
@@ -0,0 +1,292 @@
...
+        return funcOp
+
+    def apply(self, _ctx: context.Context, op: builtin.ModuleOp) -> None:
+        """Apply the convert-qecl-noise-to-qecp-noise-subroutine pass."""
+        # pylint: disable=line-too-long
+        # Insert a noise subroutine to the module in the qecp layer.
+        # Note that the visibility of those subroutines are set as private, which ensure the
+        # ["symbol-dce"](https://github.com/PennyLaneAI/catalyst/blob/372c376eb821e830da778fdc8af423eeb487eab6/frontend/catalyst/pipelines.py#L248)_
+        # pass could eliminate the unreferenced subroutines.
+        # Collect different types of codeblocks in a module.
```
lillian542, line 256: It looks like this comment might be left over from a previous version of this

frontend/catalyst/python_interface/transforms/qecp/convert_noiseop_to_subroutine.py
multiphaseCFD: Let's go for `convert-qecl-noise-to-qecp`. Looks like it would be too long to append `subroutine` here

lillian542 (approved):
A couple questions about the comments but nothing blocking - thanks @multiphaseCFD 🎉

frontend/catalyst/python_interface/transforms/qecp/convert_noiseop_to_subroutine.py
multiphaseCFD: Nice catch! Thanks @lillian542 ! Should be `the python random module`, updated

frontend/catalyst/python_interface/transforms/qecp/convert_noiseop_to_subroutine.py
multiphaseCFD: nice spot! Removed! Thanks @lillian542

Issue #2681, opened 2026-04-13
Controlled pauli rot crashes with capture


paul0403:
Controlled pauli rot crashes with capture:
```python
@qml.qjit(capture=True)
@qml.qnode(qml.device('lightning.qubit', wires=3))
def aloha(i: int):
    qml.ctrl(qml.PauliRot(0.1, ["Z", "X"],wires=[1, 2]), control=0)
    return qml.probs()

print(aloha(0))
```
```
catalyst.utils.exceptions.CompileError: catalyst failed with error code -6: catalyst: /home/paul.wang/catalyst_new/catalyst/mlir/llvm-project/mlir/lib/Transforms/Utils/DialectConversion.cpp:2159: 
virtual void mlir::ConversionPatternRewriter::replaceOp(mlir::Operation *, mlir::ValueRange): Assertion `op->getNumResults() == newValues.size() && "incorrect # of replacement values"' failed.
```

Works fine without capture, or without control

isaacdevlugt:
With https://github.com/PennyLaneAI/catalyst/pull/2683 providing a more helpful error message—thanks @lazypanda10117!—being able to execute the example in this issue boils down to either decomposing `Controlled(PauliRot)` earlier or supporting `Controlled(PauliRot)` in lightning.

PR #2683, opened 2026-04-13
Fixing controlled PauliRot to cAPI lowering bug

lazypanda10117:
**Context:** Issue #2681 raised an issue regarding a compilation error when using controlled PauliRot with capture enabled. This PR addresses the compilation issue and now indicates to the user with the proper error:

```
RuntimeError: [/Users/runner/work/pennylane-lightning/pennylane-lightning/pennylane_lightning/core/simulators/lightning_qubit/catalyst/LightningSimulator.cpp:457][Function:NamedOperation] Error in Catalyst Runtime: Controlled PauliRot is not supported
``` 

**Related GitHub Issues:**
Issue #2681 
[sc-116463]


mlir/include/Quantum/Utils/QIRPauliRot.h
```diff
@@ -0,0 +1,35 @@
+// Copyright 2026 Xanadu Quantum Technologies Inc.
```
paul0403, line 1: Does this header file need to exist? We can probably just add the definitions of these headers to the `.cpp` file directly.

In general we only need a header file if (a) the functions need to be publicly exposed, either due to mlir boilerplate requirements or the helper has value to be reused by multiple passes, or (b) the functions for some reason needs to be forward declared, neither of which is the case here I think.

mlir/lib/PBC/Transforms/ConversionPatterns.cpp
```diff
@@ -113,27 +78,14 @@ template <typename T> struct PPRotationBasedPattern : public OpConversionPattern
...
-        // Build the arguments: pauliStr, theta, modifiers, numQubits, qubits...
-        int64_t numQubits = adaptor.getInQubits().size();
-        SmallVector<Value> args;
-        args.push_back(pauliWordPtr);
-        args.push_back(thetaValue);
-        args.push_back(LLVM::ZeroOp::create(rewriter, loc, ptrType));
-        args.push_back(cond);
-        args.push_back(
-            LLVM::ConstantOp::create(rewriter, loc, rewriter.getI64IntegerAttr(numQubits)));
-        args.append(adaptor.getInQubits().begin(), adaptor.getInQubits().end());
```
paul0403, line 133: Ah I see, the old conversion just completely didn't do the control qubits! 

(Probably because it's not in the quantum dialect so you didn't know to use the `getModifiersPtr` helper 😆 )

mlir/test/Quantum/ConversionTest.mlir
```diff
@@ -331,6 +331,21 @@ func.func @multirz(%q0 : !quantum.bit, %p : f64) -> (!quantum.bit, !quantum.bit,
...
+    // CHECK-DAG: [[ctrue:%.+]] = llvm.mlir.constant(true) : i1
+    // CHECK-DAG: llvm.mlir.addressof @pauli_word_X : !llvm.ptr
+    // CHECK-DAG: [[pauliPtr:%.+]] = llvm.getelementptr inbounds {{.*}}[0, 0] : (!llvm.ptr) -> !llvm.ptr, !llvm.array<2 x i8>
+    // CHECK-DAG: [[numQubits:%.+]] = llvm.mlir.constant(1 : i64) : i64
+    // CHECK: llvm.call @__catalyst__qis__PauliRot([[pauliPtr]], {{%.+}}, {{%.+}}, [[ctrue]], [[numQubits]], %arg0)
+    %out = quantum.paulirot ["X"](%angle) %q0 : !quantum.bit
+    return %out : !quantum.bit
+}
+
+// -----
```
paul0403, line 347: Shouldn't we add a lit test with control qubits? 😅

mlir/test/PBC/ConversionTest.mlir
```diff
@@ -20,35 +20,18 @@
...
-        // CHECK: [[pauliPtr:%.+]] = llvm.getelementptr inbounds {{.*}}[0, 0] : (!llvm.ptr) -> !llvm.ptr, !llvm.array<4 x i8>
+        // CHECK-DAG: [[ctrue:%.+]] = llvm.mlir.constant(true) : i1
         // CHECK-DAG: [[theta:%.+]] = llvm.mlir.constant(1.5
+        // CHECK-DAG: llvm.mlir.addressof @pauli_word_XIZ : !llvm.ptr
+        // CHECK-DAG: [[pauliPtr:%.+]] = llvm.getelementptr inbounds {{.*}}[0, 0] : (!llvm.ptr) -> !llvm.ptr, !llvm.array<4 x i8>
+        // CHECK-DAG: [[zero:%.+]] = llvm.mlir.zero : !llvm.ptr
         // CHECK-DAG: [[numQubits:%.+]] = llvm.mlir.constant(3 : i64) : i64
-        // CHECK-DAG: [[true:%.+]] = llvm.mlir.constant(true) : i1
-        // CHECK: llvm.call @__catalyst__qis__PauliRot([[pauliPtr]], [[theta]], {{%.+}}, [[true]], [[numQubits]], %arg0, %arg1, %arg2)
+        // CHECK: llvm.call @__catalyst__qis__PauliRot([[pauliPtr]], [[theta]], [[zero]], [[ctrue]], [[numQubits]], %arg0, %arg1, %arg2)
```
paul0403, line 34: Do you know what caused the reordering here?

mlir/lib/PBC/Transforms/ConversionPatterns.cpp
```diff
@@ -39,51 +39,16 @@ using namespace catalyst::quantum;
...
-Value getPauliProductPtr(Location loc, ConversionPatternRewriter &rewriter, ModuleOp mod,
-                         ArrayAttr pauliProduct)
-{
-    std::string pauliWord = pauliProductToString(pauliProduct);
-    std::string pauliWordKey = "pauli_word_" + pauliWord;
-    return catalyst::quantum::getGlobalString(
-        loc, rewriter, pauliWordKey, StringRef(pauliWord.c_str(), pauliWord.length() + 1), mod);
-}
-
 template <typename T> struct PPRotationBasedPattern : public OpConversionPattern<T> {
```
paul0403, line 41: A little bit confused as to why the PPR pattern in the `PBC` dialect is only partially deleted. I thought you said you are moving the entire pattern to the `quantum` dialect? Or am I missing anything?

Is the idea that `pbc.ppr` will convert to `quantum.pauli_rot` by `--convert-pbc-to-llvm`, then `quantum.pauli_rot` will get converted to the llvm dialect (runtime call) by  `--convert-quantum-to-llvm`? But I don't see the creation of `Quantum::PauliRotOp` here. 

Also you are deleting the pattern from `populateConversionPatterns()` entirely? 🤔

mlir/test/PBC/ConversionTest.mlir
lazypanda10117: Not entirely sure here, but it seems only the ordering changed.

mlir/lib/PBC/Transforms/ConversionPatterns.cpp
lazypanda10117: I think we ignored that because PPR/PPMs don't have control, and the support for PauliRot lowering was added later, so it didn't occur to us to add support for controlled PauliRot.

mlir/test/Quantum/ConversionTest.mlir
lazypanda10117: Oh lemme add one. I added a pytest but forgot the lit test.

mlir/lib/PBC/Transforms/ConversionPatterns.cpp
lazypanda10117: As you said, the `PPRotationOp` and `PPRotationArbitraryOp` are part of the PBC dialect, so I keep them in the `--convert-pbc-to-llvm` pass here. 

> I don't see the creation of Quantum::PauliRotOp here.

I am not sure I understand what you mean by this? Quantum PauliRotOp is created elsewhere (e.g, at the to-ppr pass) and then the `--convert-quantum-to-llvm` pass would lower that into the proper runtime call.

paul0403: I see, thanks!

mlir/include/Quantum/Utils/QIRPauliRot.h
paul0403: Discussed, this is needed by both the quantum and the pbc conversion.

mlir/test/Quantum/ConversionTest.mlir
```diff
@@ -331,6 +331,38 @@ func.func @multirz(%q0 : !quantum.bit, %p : f64) -> (!quantum.bit, !quantum.bit,
...
+
+// CHECK: llvm.func @__catalyst__qis__PauliRot(!llvm.ptr, f64, !llvm.ptr, i1, i64, ...)
+
+// CHECK-LABEL: @controlled_paulirot
+func.func @controlled_paulirot(%q0 : !quantum.bit, %q1 : !quantum.bit, %angle : f64) -> (!quantum.bit) {
+    // CHECK: [[alloca:%.+]] = llvm.alloca %5 x !llvm.struct<(i1, i64, ptr, ptr)> : (i64) -> !llvm.ptr
```
paul0403:
```suggestion
    // CHECK: [[alloca:%.+]] = llvm.alloca {{%.+}} x !llvm.struct<(i1, i64, ptr, ptr)> : (i64) -> !llvm.ptr
```

paul0403 (approved):
🥳

sengthai (approved):
Thanks @lazypanda10117 for fixing this. 👍

PR #2684, opened 2026-04-13
Adjoint Callable with lazy=False is QJIT compatible

rniczh:
**Context:**

Previously, the eager adjoint of a callable will raise a `ValueError` inside `@qjit` because eagerly computing the adjoint of a callable was not supported. This is needed for QJIT compatibility of MultiControlledX (used in `_mcx_one_worker` and `_mcx_two_workers`), whose decomposition rules use `adjoint(fn, lazy=False)`.

The new `_expand_adjoint_callable` method handles this by executing the callable, collecting the resulting ops, then processing each one in reverse order:

1. Simple gates (RX, Toffoli, etc.) are eagerly adjointed via `create_adjoint_op(op, lazy=False)`, so the tape contains concrete adjoint gates
2. For HybridOps (ForLoop, Cond, WhileLoop) are individually wrapped in HybridAdjoint, delegating their reversal to the MLIR adjoint-lowering pass.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-116178]

### DIFF
--- a/frontend/catalyst/api_extensions/quantum_operators.py
+++ b/frontend/catalyst/api_extensions/quantum_operators.py
@@ -391,23 +391,29 @@ class AdjointCallable:
         else:
             raise ValueError(f"Expected a callable or a qml.Operator, not {target}")
 
-        if not self.lazy and not self.single_op:
-            # Supporting this for a qfunc is technically possible by invoking the decomposition on
-            # HybridAdjoint with laza=False, however one would need to outline the classical jaxpr
-            # into the outer scope, and possibly re-mapping tracers in the quantum operators,
-            # in order to avoid escaped tracers which indicate an invalid program.
-            raise ValueError(
-                "Eagerly computing the adjoint (lazy=False) is only supported on single operators."
-            )
-
     def __call__(self, *args, **kwargs):
         if self.single_op:
             base_op = self.target if self.instantiated else self.target(*args, **kwargs)
             return create_adjoint_op(base_op, self.lazy)
 
+        if not self.lazy:
+            return self._expand_adjoint_callable(args, kwargs)
+
         tracing_artifacts = self.trace_body(args, kwargs)
         dbg = tracing_artifacts[3]
         return HybridAdjoint(*tracing_artifacts[0:3], debug_info=dbg)
+
+    def _expand_adjoint_callable(self, args, kwargs):
+        """Expand a callable into individual ops, then reverse and adjoint each one."""
+        with QueuingManager.stop_recording(), QuantumTape() as tape:
+            self.target(*args, **kwargs)
+
+        for op in reversed(tape.operations):
+            # HybridAdjoint and HybridCtrl can be handled by create_adjoint_op directly
+            if isinstance(op, HybridOp) and not isinstance(op, (HybridAdjoint, HybridCtrl)):
+                AdjointCallable(lambda bound_op=op: qml.apply(bound_op) and None, lazy=True)()
+            else:
+                create_adjoint_op(op, lazy=False)
 
     def trace_body(self, args, kwargs):
         """Generate a HybridOpRegion for use by Catalyst."""
--- a/frontend/catalyst/api_extensions/quantum_operators.py
+++ b/frontend/catalyst/api_extensions/quantum_operators.py
@@ -409,8 +409,9 @@ class AdjointCallable:
             self.target(*args, **kwargs)
 
         for op in reversed(tape.operations):
-            # HybridAdjoint and HybridCtrl can be handled by create_adjoint_op directly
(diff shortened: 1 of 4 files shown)


frontend/test/pytest/test_preprocess.py
```diff
@@ -632,5 +632,51 @@ def test_grad_method_with_target_gates_provided_error(self):
...
+
+        @qjit(capture=False)
+        @qml.qnode(dev)
+        def circuit():
+            def sub():
+                qml.Hadamard(0)
+
+                @for_loop(0, 3, 1)
+                def loop(i):
+                    qml.RX(0.5, wires=i)
```
dwierichs: Let's do 
```suggestion
                    qml.RY(0.5, wires=i)
```
to avoid all commutativity :)

frontend/catalyst/api_extensions/quantum_operators.py
```diff
@@ -391,24 +391,30 @@ def __init__(self, target, lazy):
...
 
+    def _expand_adjoint_callable(self, args, kwargs):
+        """Expand a callable into individual ops, then reverse and adjoint each one."""
+        with QueuingManager.stop_recording(), QuantumTape() as tape:
+            self.target(*args, **kwargs)
+
+        for op in reversed(tape.operations):
+            # HybridAdjoint and HybridCtrl can be handled by create_adjoint_op directly
+            if isinstance(op, HybridOp) and not isinstance(op, (HybridAdjoint, HybridCtrl)):
+                AdjointCallable(lambda bound_op=op: qml.apply(bound_op) and None, lazy=True)()
```
dwierichs, line 415: Just to double check: `lazy=True` is the correct thing to pass here? :)

dwierichs (commented):
Looks good to me, thanks @rniczh 👌 
Just one tiny request for one test, and a question due to my unfamiliarity with the Catalyst stack (I might leave approving to people with expertise here :) )

frontend/catalyst/api_extensions/quantum_operators.py
rniczh: For the HybridOp, it will create HybridAdjoint and let MLIR to handle this. `lazy=True` is to avoid this op run into infinite recursion here, since it will be required to expand it again.

frontend/catalyst/api_extensions/quantum_operators.py
```diff
@@ -391,24 +391,30 @@ def __init__(self, target, lazy):
...
         return HybridAdjoint(*tracing_artifacts[0:3], debug_info=dbg)
 
+    def _expand_adjoint_callable(self, args, kwargs):
+        """Expand a callable into individual ops, then reverse and adjoint each one."""
+        with QueuingManager.stop_recording(), QuantumTape() as tape:
+            self.target(*args, **kwargs)
+
+        for op in reversed(tape.operations):
+            # HybridAdjoint and HybridCtrl can be handled by create_adjoint_op directly
+            if isinstance(op, HybridOp) and not isinstance(op, (HybridAdjoint, HybridCtrl)):
```
dime10: I don't think we can exclude these here, as we don't handle `Adjoint(HybridAdjoint)`. Unlike gates, there is no attribute we can generate for the MLIR op (in this case `AdjointOp`) to indicate its adjoint version.

See this example:
```py
def g():
    qml.H(0)
    qml.adjoint(lambda: qml.Z(0) and None)()
    qml.X(0)

@qjit
@qml.qnode(qml.device('null.qubit', wires=1))
def f():
    qml.adjoint(g, lazy=False)()
    return qml.probs()

f()
```

dime10 (changes requested):

frontend/catalyst/api_extensions/quantum_operators.py
rniczh: Thanks! You're right, thanks for the clarification. I updated in this commit, and I used your example as a test 👍, but since the adjoint of Z is still Z, I modified it to RX here. https://github.com/PennyLaneAI/catalyst/pull/2684/commits/379bdb2f84484b2ac457ddc948f253569448c241 https://github.com/PennyLaneAI/catalyst/pull/2684/commits/123c7f7e3ff9360501c6b7a910866dc16de75880

dime10: Perfect thanks :)

dime10 (approved):

frontend/catalyst/api_extensions/quantum_operators.py
dwierichs: Ah, makes sense 👍 Thank you!

paul0403 (approved):

PR #2690, opened 2026-04-14
Improve test coverage for specs

jzaia18:
**Context:**
The integration tests for specs are lacking some coverage of edge and corner cases. This presents a worry that if code changes, some edge cases may silently stop working correctly

* Increases integration test coverage of backends of `qp.specs` which directly use code within Catalyst
* Adds new tests for the features added in https://github.com/PennyLaneAI/pennylane/pull/9307
* Removes a non-actionable warning that frequently pollutes specs output

**Benefits:**
More confidence when changes are made that will impact specs

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-113517]
Relies on https://github.com/PennyLaneAI/pennylane/pull/9307 being merged first


frontend/catalyst/python_interface/inspection/specs.py
```diff
@@ -88,13 +87,6 @@ def mlir_specs(
             "The provided `qnode` argument does not appear to be a valid QJIT compiled QNode."
         )
 
-    if args or kwargs:
-        warnings.warn(
```
AntonNI8, line 92: So long as we raise more actionable warnings downstream, no issues with removing this one.

sengthai (approved):
LGTM! Thanks @jzaia18

frontend/test/pytest/test_specs.py
```diff
@@ -334,6 +342,40 @@ def test_basic_passes_multi_level(self, simple_circuit):
             check_specs_header_same(actual, single_level_specs, skip_level=True)
             check_specs_resources_same(res, single_level_specs["resources"])
 
+    @pytest.mark.usefixtures("use_both_frontend")
```
andrijapau: Can we update this to use the `capture_mode` fixture instead?

jzaia18: Any reason we don't want it to test both?

andrijapau: What do you mean?

andrijapau: OH. Yeah, the fixture name is a bit weird - `capture_mode` is a flag that produces a function param that's either `True` or `False`.

jzaia18: Oh nevermind, I see what you mean. Yes, I will update this

andrijapau: See this [example](https://github.com/PennyLaneAI/catalyst/blob/1b067e0f29f76148ebdb55a5fa2697c7d39ecc80/frontend/test/pytest/test_state_prep.py#L53).

jzaia18: [9ac099d](https://github.com/PennyLaneAI/catalyst/pull/2690/commits/9ac099d214f65a33c11417d02568629691a7848d)

frontend/test/pytest/test_specs.py
```diff
@@ -262,32 +261,38 @@ def circ():
...
+        ):
             qml.specs(no_passes, level=10)()
 
-        with pytest.raises(check=ValueError, match="Requested specs levels 10, 11"):
+        with pytest.raises(
+            check=ValueError,
+            match="The 'level' argument to qml.specs for QJIT'd "
+            "QNodes is out of bounds, got 10.",
+        ):
             qml.specs(no_passes, level=[10, 11])()
```
andrijapau, line 287: Non-blocking but can't we parameterize this test?

frontend/test/pytest/test_specs.py
```diff
@@ -723,6 +947,8 @@ def subroutine():
         @qml.qjit(autograph=True)
         @qml.qnode(dev)
         def circuit():
+            qml.PauliX(wires=1)
```
andrijapau, line 950: Why was this added?

andrijapau (approved):
Looks good, just minor, non-blocking questions :)

frontend/test/pytest/test_specs.py
jzaia18: To make sure that gates that are _not_ within the loop aren't counted multiple times

frontend/test/pytest/test_specs.py
jzaia18: Good point. I'll do this in the next PR in the sequence so we can merge this one

doc/releases/changelog-dev.md
```diff
@@ -624,6 +624,9 @@
 
 <h3>Internal changes ⚙️</h3>
 
+* Additional integration tests have been added for the pass-by-pass version of `qp.specs`.
```
dime10, line 627: Just some general guidelines regarding our changelog:
- We don't add entries that only concern tests, build scripts, or dev configuration stuff, because none of those are part of the released package.
- The changelog should reflect changes in the next version of the package _as compared to the previously released version_, that is we are not recording a PR-by-PR log of changes. Instead, when multiple PRs work on the same "feature", or a PR fixes a bug from another PR in the same release cycle, there would be a single changelog entry for all of them. (For instance, it wouldn't make sense to have a bugfix entry for a feature that was newly introduced in the same cycle because a user would have never encountered that bug.)
- The warning removal on the other hand is a change in observable behaviour compared the previous release but is not mentioned in the changelog entry.

dime10 (commented):
Thanks for fixing this @jzaia18, no worries about the comment below since we'll the edit the changelog again before release, but for the next time :)

PR #2692, opened 2026-04-15
[qref 2.5.1] Skip transform sequences when looking for subroutine functions

paul0403:
**Context:**
When determining which functions are qref subroutines, there's a unconditional cast to `func::FuncOp` in the code. 

As it turns out, in a `mlir::CallGraph`, each node is not guaranteed to be a `func::FuncOp`. Things like `transform.named_sequence` are also considered call graph nodes. 

Which makes sense: each call graph node is only supposed to be a callable, which are all things `FunctionOpInterface`.

Early exit if the call graph node is not a `func::FuncOp`.

**Benefits:**
Correctness.

[sc-115119]


kipawaa (approved):
good catch, thanks!

mehrdad2m (approved):

PR #2693, opened 2026-04-15
Fix double extracts when multiple compbasis obs are present

paul0403:
**Context:**
In #2641, we fixed double extracts when multiple named observables or Hermitian observables are present. 

We should do the same for compbasis observable.

Insert qubits extracted for compbasis observables back into the cache.

**Benefits:**
No double extracts when multiple compbasis observables are present (in legacy frontend).

[sc-115297]


lillian542 (approved):
LGTM and fixes the problem I was seeing in #2656! Thanks 🚀

multiphaseCFD (approved):
Thanks @paul0403 ! LGTM!

PR #2694, opened 2026-04-15
[qref 3.3] Migrate for loop's plxpr conversion to reference semantics

paul0403:
**Context:**
Migrate for loop's plxpr conversion to reference semantics

All quantum reference values are captured now via closure from above automatically by jax infra.
No need to keep track of quantum value args to the for loop anymore :rocket: :fire: :partying_face: 

**Benefits:**
Much cleaner conversion of plxpr to catalys jaxpr. 

[sc-115870]


frontend/catalyst/from_plxpr/control_flow.py
```diff
@@ -40,62 +40,11 @@
...
 
-def _calling_convention(
-    interpreter, closed_jaxpr, *args_plus_qregs, outer_dynqreg_handlers=(), return_qreg=True
-):
-    # Arg structure (all args are tracers, since this function is to be `make_jaxpr`'d):
-    # Regular args, then dynamically allocated qregs, then global qreg
-    # TODO: merge dynamically allocaed qregs into regular args?
-    # But this is tricky, since qreg arguments need all the SSA value semantics conversion infra
-    # and are different from the regular plain arguments.
-    *args_plus_dynqregs, global_qreg = args_plus_qregs
```
paul0403, line 51: You can't imagine how happy I am to see the register tracking gone 😭

frontend/catalyst/jax_primitives.py
```diff
@@ -2616,7 +2616,7 @@ def _for_loop_lowering(
     assert [lower_bound.type, upper_bound.type, step.type] == all_param_types_plus_consts[
         body_nconsts + num_implicit_inputs : body_nconsts + num_implicit_inputs + 3
     ]
-    assert [val.type for val in body_consts] == all_param_types_plus_consts[:body_nconsts]
+    # assert [val.type for val in body_consts] == all_param_types_plus_consts[:body_nconsts]
```
paul0403: As it turns out, this assertion will break because it is somehow comparing `qref.reg<?>` and `qref.reg<3>`. 

In the definition for the register type
```python
class QrefQreg(AbstractValue):
    """Abstract quantum register."""

    def __init__(self, num_qubits=None):
        self.num_qubits = num_qubits
        self.hash_value = hash("QrefQreg") + hash(num_qubits)

    def __eq__(self, other):
        return isinstance(other, QrefQreg) and self.hash_value == other.hash_value

    def __hash__(self):
        return self.hash_value


def _qref_qreg_lowering(aval):
    assert isinstance(aval, QrefQreg)
    if aval.num_qubits is None:
        tag = "?"
    else:
        tag = str(aval.num_qubits)
    return ir.OpaqueType.get("qref", "reg<" + tag + ">")
```

and the def and bind call for the alloc primitive
```python
# qref_jax_primitives.py
@qref_alloc_p.def_abstract_eval
def _qref_alloc_abstract_eval(num_qubits=None):
    return QrefQreg(num_qubits)

# from_plxpr.py
qreg = qref_alloc_p.bind(len(device.wires))
```

Because I binded the primitive as a positional arg instead of a kwarg, the `num_qubits` is actually always dynamic 😅 , leading to the assertion failure.

So, there should be better handling regarding static-vs-dynamic allocation size. 

We have dragged on for quite some time regarding implementing qubit-invariant compilation in plxpr path. This is a good opportunity to actually do it. This is why I labeled this PR as 3.4. Part 3.3 will be qubit-invariant compilation.

frontend/catalyst/from_plxpr/control_flow.py
```diff
@@ -40,62 +40,11 @@
...
-    num_dynamic_alloced_qregs = len(outer_dynqreg_handlers)
-    args, dynalloced_qregs = (
-        args_plus_dynqregs[: len(args_plus_dynqregs) - num_dynamic_alloced_qregs],
-        args_plus_dynqregs[len(args_plus_dynqregs) - num_dynamic_alloced_qregs :],
-    )
-
+def _calling_convention(interpreter, closed_jaxpr, *args):
```
albi3ro, line 40: Do we even need `_calling_convention` anymore?

paul0403: I think we can get rid of it, the inner and outer interpreters don't have any difference now

paul0403: Let me see

frontend/test/lit/test_qref/test_control_flow.py
```diff
@@ -0,0 +1,190 @@
+# Copyright 2026 Xanadu Quantum Technologies Inc.
```
paul0403, line 1: changelog

frontend/catalyst/from_plxpr/control_flow.py
paul0403: We don't need `workflow_for_loop` or `handle_for_loop` anymore! We literally only need the lowering rule from plxpr directly, since all loop arguments are now classical.

frontend/catalyst/jax_primitives.py
```diff
@@ -2688,6 +2689,64 @@ def _for_loop_lowering(
     return for_op_scf.results
 
 
+# pylint: disable=too-many-arguments
+def _pl_for_loop_lowering(
```
paul0403, line 2693: This is the same lowering @albi3ro added in #2603

frontend/catalyst/from_plxpr/control_flow.py
JerryChen97: :partying_face:

frontend/catalyst/from_plxpr/control_flow.py
```diff
@@ -31,70 +30,17 @@
...
```
albi3ro: Should we then leave it be in this PR and remove it once the last thing that uses it has been removed?

paul0403: https://github.com/PennyLaneAI/catalyst/pull/2694/commits/602763dfaecec6a0dd4a7e37874e8981901201b4 restored

albi3ro (approved):
Beautiful! ❤️ 

Going to trust you that all the tests will pull together, but looks good 👍 So wonderful to get rid of all of that.

JerryChen97 (approved):

PR #2695, opened 2026-04-15
Fix resource-tracker measurement names, loop resolution, and qnode marking

sengthai:
**Context:**
The Catalyst `ResourceAnalysis` is now used by PennyLane's `qp.specs` for MLIR-level 
resource tracking (via the `resource-tracker` pass). Several features that the 
Python xDSL-based `specs_collect` handled correctly were missing from the C++ 
implementation, causing test failures in `test_specs.py`.

- Fix measurement names in `ResourceAnalysis` to resolve observable chains 
  (e.g. `expval(Hamiltonian(num_terms=2))`, `expval(PauliZ)`, `sample(1 wires)`)
  instead of the incorrect `expval(all wires)` that was produced by counting 
  qubit operands on the measurement op itself.
- Add `resolveConstantIndex` to handle for-loop trip count resolution.
- Add `isQnode` field to `ResourceResult` and emit `"qnode": true/false` per  function.

[[sc-115506]]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -429,7 +429,7 @@
 <h3>Bug fixes 🐛</h3>
 
 * Fixed incorrect measurement names and missing loop/subroutine resource counts in the
-  `resource-tracker` MLIR pass used by :func:`~pennylane.specs`.
+  `resource-tracker` MLIR pass.
   [(#2695)](https://github.com/PennyLaneAI/catalyst/pull/2695)
 
 * :func:`~pennylane.adjoint` can now be used on subroutines with classical arguments.
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -428,6 +428,10 @@
 
 <h3>Bug fixes 🐛</h3>
 
+* Fixed incorrect measurement names and missing loop/subroutine resource counts in the
+  `resource-tracker` MLIR pass used by :func:`~pennylane.specs`.
+  [(#2695)](https://github.com/PennyLaneAI/catalyst/pull/2695)
+
 * :func:`~pennylane.adjoint` can now be used on subroutines with classical arguments.
   [(#2590)](https://github.com/PennyLaneAI/catalyst/pull/2590)
 
--- a/mlir/include/Catalyst/Transforms/Passes.td
+++ b/mlir/include/Catalyst/Transforms/Passes.td
@@ -278,6 +278,13 @@ def ResourceTrackerPass : Pass<"resource-tracker"> {
             /*type=*/"std::string",
             /*default=*/"",
             /*description=*/"The output JSON's filename"
+        >,
+        Option<
+            /*C++ var name=*/"printJson",
+            /*CLI arg name=*/"print-json",
+            /*type=*/"bool",
+            /*default=*/"false",
+            /*description=*/"Print resource counts as JSON to stdout only"
         >
     ];
 }
--- a/mlir/include/Catalyst/Utils/ConstantResolve.h
+++ b/mlir/include/Catalyst/Utils/ConstantResolve.h
@@ -28,8 +28,8 @@ namespace catalyst {
 /// Returns std::nullopt if the value cannot be statically resolved.
 std::optional<double> resolveConstant(mlir::Value val);
 
-/// Convenience wrapper that resolves to int64_t.
-/// Returns std::nullopt if the value cannot be resolved or is not integral.
+/// Convenience wrapper that resolves to int64_t by truncating the result.
+/// Returns std::nullopt if the value cannot be resolved.
 std::optional<int64_t> resolveConstantInt(mlir::Value val);
 
 } // namespace catalyst
--- a/mlir/include/Catalyst/Utils/ConstantResolve.h
+++ b/mlir/include/Catalyst/Utils/ConstantResolve.h
@@ -1,0 +1,35 @@
+// Copyright 2026 Xanadu Quantum Technologies Inc.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+
+//     http://www.apache.org/licenses/LICENSE-2.0
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <optional>
+
+#include "mlir/IR/Value.h"
+
+namespace catalyst {
+
+/// Recursively resolve a Value to a numeric constant by walking through
+/// common MLIR op chains (arith.constant, arith.index_cast, arith.addi/subi/muli,
+/// arith.addf, tensor.extract, tensor.from_elements,
+/// stablehlo.constant/convert/broadcast_in_dim/add/subtract, etc.).
+///
+/// Returns std::nullopt if the value cannot be statically resolved.
+std::optional<double> resolveConstant(mlir::Value val);
+
+/// Convenience wrapper that resolves to int64_t.
+/// Returns std::nullopt if the value cannot be resolved or is not integral.
+std::optional<int64_t> resolveConstantInt(mlir::Value val);
+
+} // namespace catalyst
--- a/mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
+++ b/mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
@@ -249,6 +249,9 @@ void ResourceAnalysis::analyzeForLoop(scf::ForOp forOp, ResourceResult &result, 
             int64_t tripCount = (*ub - *lb + *step - 1) / *step;
             bodyResult.multiplyByScalar(tripCount);
         }
+        else {
+            result.hasDynLoop = true;
+        }
     }
     result.mergeWith(bodyResult);
 }
@@ -263,12 +266,17 @@ void ResourceAnalysis::analyzeWhileLoop(scf::WhileOp whileOp, ResourceResult &re
         int64_t iters = estAttr.getValue().getSExtValue();
         bodyResult.multiplyByScalar(iters);
     }
+    else {
+        result.hasDynLoop = true;
+    }
 
     result.mergeWith(bodyResult);
 }
 
 void ResourceAnalysis::analyzeIfOp(scf::IfOp ifOp, ResourceResult &result, bool isAdjoint)
 {
+    result.hasBranches = true;
+
     ResourceResult thenResult;
     analyzeRegion(ifOp.getThenRegion(), thenResult, isAdjoint);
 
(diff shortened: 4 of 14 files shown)


doc/releases/changelog-dev.md
```diff
@@ -428,6 +428,10 @@
 
 <h3>Bug fixes 🐛</h3>
 
+* Fixed incorrect measurement names and missing loop/subroutine resource counts in the
+  `resource-tracker` MLIR pass used by :func:`~pennylane.specs`.
```
```suggestion
* Fixed incorrect measurement names and missing loop/subroutine resource counts in the
  `resource-tracker` MLIR pass.
```

Since these are just changes to the pass itself, we can avoid mentioning specs directly

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -178,11 +216,83 @@ ResourceAnalysis::ResourceAnalysis(Operation *op)
...
+    for (auto &funcEntry : funcResults) {
+        resolveFunctionCalls(funcEntry.getKey());
+    }
 
     entryFuncName = entryFunc.str();
 }
 
+/// Recursively resolve a Value to an integer constant.
+/// Mirrors the Python `resolve_constant_params` in xdsl_conversion.py.
+static std::optional<int64_t> resolveConstantIndex(Value val)
```
jzaia18: Would it be better to put this somewhere more general? Is there a place within Catalyst for general utility functions? I can imagine this being useful for other MLIR passes as well

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -178,11 +216,83 @@ ResourceAnalysis::ResourceAnalysis(Operation *op)
...
+{
+    Operation *op = val.getDefiningOp();
+    if (!op) {
+        return std::nullopt;
+    }
+
+    StringRef opName = op->getName().getStringRef();
```
jzaia18: Let's move this down past line 270 so we only compute it in the cases where its needed

jzaia18 (commented):
Awesome work! Just a couple comments

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: Yes, we should do!

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: Done!

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: Done

[813eb1b](https://github.com/PennyLaneAI/catalyst/pull/2695/commits/813eb1b671b3d7e2c4a49994d5b3e41496cc672c)

jzaia18 (approved):
LGTM

jzaia18:
Note to reviewers: Don't worry about these failing unit tests for now, they should be intentionally failing in this branch which has the updated unit tests but not yet the updated code in PennyLane (and this branch will not be merged to main)

mlir/lib/Catalyst/Transforms/ResourceTrackerPass.cpp
```diff
@@ -192,6 +193,7 @@ struct ResourceTrackerPass : public impl::ResourceTrackerPassBase<ResourceTracke
             std::strftime(buffer, sizeof(buffer), "_%Y%m%d_%H%M%S", now_tm);
 
             file_name += std::string(buffer);
+            llvm::outs() << llvm::formatv("{0:2}", jsonValue).str() << "\n";
```
lazypanda10117: Why do we want to do this?

mlir/include/Catalyst/Utils/ConstantResolve.h
```diff
@@ -0,0 +1,35 @@
...
+/// Recursively resolve a Value to a numeric constant by walking through
+/// common MLIR op chains (arith.constant, arith.index_cast, arith.addi/subi/muli,
+/// arith.addf, tensor.extract, tensor.from_elements,
+/// stablehlo.constant/convert/broadcast_in_dim/add/subtract, etc.).
+///
+/// Returns std::nullopt if the value cannot be statically resolved.
+std::optional<double> resolveConstant(mlir::Value val);
+
+/// Convenience wrapper that resolves to int64_t.
+/// Returns std::nullopt if the value cannot be resolved or is not integral.
```
lazypanda10117: I am not sure if this description is correct? It is always going to be integral because in the implementation you are just casting it from a double. Is that the intended behaviour?

lazypanda10117 (commented):
Just two minor comments. Otherwise looks good to me :)

mlir/lib/Catalyst/Transforms/ResourceTrackerPass.cpp
sengthai: it is needed for MLIR test.

lazypanda10117: But in general this is not used right? Is there a better way to do this? This feels more like a debug statement to me.

sengthai: This prints the JSON to stdout, so the MLIR lit test works.

jzaia18: I agree, I think at the very least, printing to stdout should be an option that's mutually exclusive with writing a file. I can make this happen if no filename is given in the base PR

mlir/lib/Catalyst/Utils/ConstantResolve.cpp
```diff
@@ -0,0 +1,155 @@
...
+    }
+    if (auto denseInt = dyn_cast<DenseIntElementsAttr>(attr)) {
+        if (denseInt.isSplat() || denseInt.getNumElements() == 1) {
+            return static_cast<double>(denseInt.getSplatValue<APInt>().getSExtValue());
+        }
+    }
+    return std::nullopt;
+}
+
+std::optional<double> resolveConstant(Value val)
```
rniczh, line 49: I think handle the rules manually may not a good idea...It's hard to enumerate all the possibility if we make IR somewhat different in the future.

A way to handle this is use dataflow solver. And a very good example you can follow is this: `mlir/lib/Dialect/Arith/Transforms/IntRangeOptimizations.cpp`.  So it includes`IntegerRangeAnalysis`,`SparseConstantPropagation` and `DeadCodeAnalysis`.

Btw, because your implementation is analysis pass, so you shouldn't use this pass to simplify the code in your analysis pass. Otherwise, it will break the rule in [MLIR specfication](https://mlir.llvm.org/docs/PassManagement/#analysis-management): **Must not modify the given operation.**

```
An important concept, along with transformation passes, are analyses. These are conceptually similar to transformation passes, except that they compute information on a specific operation without modifying it. In MLIR, analyses are not passes but free-standing classes that are computed lazily on-demand and cached to avoid unnecessary recomputation. An analysis in MLIR must adhere to the following:

Provide a valid constructor taking either an Operation* or Operation* and AnalysisManager &.
The provided AnalysisManager & should be used to query any necessary analysis dependencies.
Must not modify the given operation.
```

rniczh (commented):
Just a fly-by comment, but yeah, it's okay to ignore the following comment and do whatever you want 👍

mlir/lib/Catalyst/Utils/ConstantResolve.cpp
paul0403: 👍 definitely don't evaluate the entire computational tree if there's already an MLIR tool to do it!

sengthai: Thanks @jzaia18! That's really helpful! 

> I think handle the rules manually may not a good idea...It's hard to enumerate all the possibility if we make IR somewhat different in the future.

That's a good idea to use build-in functions if they exists. I wonder if they resolve `stablehlo` as well?

sengthai: >Btw, because your implementation is analysis pass, so you shouldn't use this pass to simplify the code in your analysis pass. Otherwise, it will break the rule in [MLIR specfication](https://mlir.llvm.org/docs/PassManagement/#analysis-management): Must not modify the given operation.

Yes, probably `IntegerRangeAnalysis` and `SparseConstantPropagation` can be used in pass?

mlir/lib/Catalyst/Transforms/ResourceTrackerPass.cpp
sengthai: now I separate both printing (`print-json`) and writing json (`output-json`). 
[3f0dd40](https://github.com/PennyLaneAI/catalyst/pull/2695/commits/3f0dd40f87740686fbc8672779a174624855abe9)

mlir/include/Catalyst/Utils/ConstantResolve.h
sengthai: Improved! 
https://github.com/PennyLaneAI/catalyst/blob/3f0dd40f87740686fbc8672779a174624855abe9/mlir/include/Catalyst/Utils/ConstantResolve.h#L31-L32

mlir/include/Catalyst/Transforms/Passes.td
```diff
@@ -278,6 +278,13 @@ def ResourceTrackerPass : Pass<"resource-tracker"> {
             /*type=*/"std::string",
             /*default=*/"",
             /*description=*/"The output JSON's filename"
+        >,
+        Option<
+            /*C++ var name=*/"printJson",
+            /*CLI arg name=*/"print-json",
+            /*type=*/"bool",
+            /*default=*/"false",
```
jzaia18, line 286: Don't we want this to default as true?

mlir/lib/Catalyst/Utils/ConstantResolve.cpp
rniczh: yeah, you can run it at initial stage
```
solver = std::make_unique<DataFlowSolver>();
solver->load<DeadCodeAnalysis>();
solver->load<SparseConstantPropagation>();
solver->load<IntegerRangeAnalysis>();
if (failed(solver->initializeAndRun(op))) {
     ...
}
```

Then you can take the integer lattice later

mlir/include/Catalyst/Transforms/Passes.td
sengthai: I think the default should be false because the json is not the default format for mlir stdout.

jzaia18 (approved):
Looks good from my POV 🚀

mlir/lib/Catalyst/Utils/ConstantResolve.cpp
sengthai: > yeah, you can run it at initial stage

We still need to support with `stablehlo`. For now, we can use the `ConstantResolve`. However, of course, we can improve that later by supporting sablehlo ops. I drop the task in https://github.com/PennyLaneAI/catalyst/issues/2707

mlir/lib/Catalyst/Transforms/ResourceTrackerPass.cpp
lazypanda10117: Thanks!

lazypanda10117 (approved):

mlir/lib/Catalyst/Utils/ConstantResolve.cpp
rniczh: @sengthai shouldn't the stablehlo already support the fold function? That what SCCP needs

rniczh: Okay, I see, stablehlo only support fold on constant 🤔

PR #2696, opened 2026-04-15
Remove dynamic qubit allocation from device capabilities in legacy frontend

paul0403:
**Context:**
We unconditionally add `qml.allocate()` and `qml.deallocate()` to the device's capability set, regardless of what the device toml says.

This is obviously wrong. In fact, we don't actually need to register these capabilities, because dynamic qubit allocation is only implemented for the plxpr frontend. Catalyst sees the primitives, not the operators.

Remove the registry of `qml.allocate()` and `qml.deallocate()` from the `qjit_device`'s capabilities.

### DIFF
--- a/frontend/catalyst/device/qjit_device.py
+++ b/frontend/catalyst/device/qjit_device.py
@@ -238,19 +238,6 @@ def get_qjit_device_capabilities(target_capabilities: DeviceCapabilities) -> Dev
         target_capabilities.measurement_processes, RUNTIME_MPS
     )
 
-    # Enable dynamic qubit allocation with qml.allocate and qml.deallocate
-    # if target_capabilities.dynamic_qubit_management:
-    # qjit_capabilities.operations.update(
-    #     {
-    #         "allocate": OperatorProperties(
-    #             invertible=False, controllable=False, differentiable=False
-    #         ),
-    #         "deallocate": OperatorProperties(
-    #             invertible=False, controllable=False, differentiable=False
-    #         ),
-    #     }
-    # )
-
     # Control-flow gates to be lowered down to the LLVM control-flow instructions
     qjit_capabilities.operations.update(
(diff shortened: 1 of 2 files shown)


frontend/catalyst/device/qjit_device.py
```diff
@@ -239,16 +239,17 @@ def get_qjit_device_capabilities(target_capabilities: DeviceCapabilities) -> Dev
...
-        {
-            "allocate": OperatorProperties(
-                invertible=False, controllable=False, differentiable=False
-            ),
-            "deallocate": OperatorProperties(
-                invertible=False, controllable=False, differentiable=False
-            ),
-        }
-    )
+    # if target_capabilities.dynamic_qubit_management:
```
paul0403: @dime10 It just occurred to me that dynamic allocation isn't even implemented in the legacy frontend. We only implemented it for capture, at which point it sees the plxpr alloc primitive instead of the `qml.allocate` operator

So this might not even be needed in the first place 😅 I'm commenting them out to see if anything breaks.

dime10: I suppose that's true. Yeah maybe we just remove them then 👍

paul0403: deleted https://github.com/PennyLaneAI/catalyst/pull/2696/commits/2a8b0bb917da99947c980fd13298b031dd9bbe68

dime10 (approved):

joeycarter (commented):
Changelog?

paul0403:
> Changelog?

Added https://github.com/PennyLaneAI/catalyst/pull/2696/commits/8a1d17dbd439208d98f161d3d0d063a316780bfe

joeycarter (approved):
Looks good, thanks @paul0403!

PR #2703, opened 2026-04-16
Add --debug-pipeline option for run_filecheck tests

joeycarter:
**Context:** Often a compiler-pass developer wants to see the full IR before and after applying some pass or pipeline, but it is not currently convenient to do so for xDSL tests in the Catalyst Python interface that use the `run_filecheck` fixture.

**Description of the Change:** This PR adds a pytest command line option `--debug-pipeline` that prints out the full IR before and after the given pass pipeline is applied. For example:

```console
$ pytest frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_cancel_inverses.py -k test_simple_self_inverses -s --debug-pipeline
=== test session starts ===
...

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_cancel_inverses.py 
[DEBUG] ========== RUNNING FILECHECK ==========
[DEBUG] Initial xDSL module:
builtin.module {
  func.func @test_func() {
    %0 = "test.op"() : () -> !quantum.bit
    %1 = quantum.custom "PauliX"() %0 : !quantum.bit
    %2 = quantum.custom "PauliX"() %1 : !quantum.bit
    quantum.dealloc_qb %2 : !quantum.bit
    func.return
  }
}
[DEBUG] Applying pass pipeline 'PassPipeline(passes=(IterativeCancelInversesPass(),), callback=None)'... OK
[DEBUG] After:
builtin.module {
  func.func @test_func() {
    %0 = "test.op"() : () -> !quantum.bit
    quantum.dealloc_qb %0 : !quantum.bit
    func.return
  }
}
.
```

Note that the pytest `--capture=no` or `-s` option (or similar) must also be supplied in order to view the output in the terminal.

[sc-116855]

### DIFF
--- a/frontend/test/pytest/python_interface/conftest.py
+++ b/frontend/test/pytest/python_interface/conftest.py
@@ -39,19 +39,7 @@ except (ImportError, ModuleNotFoundError):
     deps_available = False
 
 
-def pytest_addoption(parser: pytest.Parser):
-    """Register custom command line options to pytest."""
-    parser.addoption(
-        "--debug-pipeline",
-        action="store_true",
-        help=(
-            "For tests that use the run_filecheck fixture, display the full xDSL module IR before "
-            "and after applying a compilation pipeline",
-        ),
-    )
-
-
-@pytest.fixture(name="debug_pipeline", scope="session")
+@pytest.fixture(scope="session", name="debug_pipeline")
 def debug_pipeline_fixture(request: pytest.FixtureRequest):
     """Returns whether the --debug-pipeline command line option has been given."""
     return request.config.getoption("--debug-pipeline")
--- a/frontend/test/pytest/python_interface/conftest.py
+++ b/frontend/test/pytest/python_interface/conftest.py
@@ -53,11 +53,16 @@ def pytest_addoption(parser: pytest.Parser):
 
 @pytest.fixture(scope="session")
 def debug_pipeline(request: pytest.FixtureRequest):
+    """Returns whether the --debug-pipeline command line option has been given."""
     return request.config.getoption("--debug-pipeline")
 
 
 @contextmanager
 def debug_step(message, enabled=True):
+    """A convenience context manager that prints a debug message before running a block of code.
+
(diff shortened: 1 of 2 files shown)


frontend/test/pytest/python_interface/conftest.py
```diff
@@ -38,10 +39,44 @@
...
 
 
+def pytest_addoption(parser: pytest.Parser):
+    """Register custom command line options to pytest."""
+    parser.addoption(
+        "--debug-pipeline",
+        action="store_true",
+        help=(
+            "For tests that use the run_filecheck fixture, display the full xDSL module IR before "
+            "and after applying a compilation pipeline",
```
mehrdad2m: Can we also mention here that one needs to use `-s` to get to see the IR?

frontend/test/pytest/python_interface/conftest.py
```diff
@@ -77,8 +105,14 @@ def _run_filecheck_impl(
...
-    pipeline = PassPipeline(pipeline)
-    pipeline.apply(ctx, xdsl_module)
+    pipeline = PassPipeline(passes)
+
+    with debug_step(f"Applying pass pipeline '{pipeline}'", debug_pipeline):
+        pipeline.apply(ctx, xdsl_module)
+
+    if debug_pipeline:
+        print("[DEBUG] After:")
+        print(xdsl_module)
```
mehrdad2m, line 115: Would it make sense to hide this within the `debug_step`?

mehrdad2m (approved):
Thanks @joeycarter, This is great!

frontend/test/pytest/python_interface/conftest.py
joeycarter: Probably not, since `debug_step` prints "OK" at the end of the debug message, which wouldn't be applicable here.

frontend/test/pytest/python_interface/conftest.py
joeycarter: Good call, done: e4f971a8c

mudit2812 (approved):

PR #2708, opened 2026-04-17
[Test] Phase Gradient Alloc

JerryChen97:
**Context:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/frontend/catalyst/from_plxpr/qfunc_interpreter.py
+++ b/frontend/catalyst/from_plxpr/qfunc_interpreter.py
@@ -393,15 +393,19 @@ _special_op_bind_call = {
 }
 
 
-# pylint: disable=unused-argument
 @PLxPRToQuantumJaxprInterpreter.register_primitive(qml.allocation.allocate_prim)
-def handle_qml_alloc(self, *, num_wires, state=None, restored=False):
+def handle_qml_alloc(self, *, num_wires, state=None, restored=False, precision=None):
     """Handle the conversion from plxpr to Catalyst jaxpr for the qml.allocate primitive"""
 
     self.has_dynamic_allocation = True
 
+    # Normalize AllocateState enum to its string value so qalloc_p receives a plain string.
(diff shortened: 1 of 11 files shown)


frontend/catalyst/from_plxpr/qfunc_interpreter.py
```diff
@@ -393,15 +393,19 @@ def _pauli_rot_bind_call(*invals, op, qubits_len, params_len, ctrl_len, adjoint,
...
     """Handle the conversion from plxpr to Catalyst jaxpr for the qml.allocate primitive"""
 
     self.has_dynamic_allocation = True
 
+    # Normalize AllocateState enum to its string value so qalloc_p receives a plain string.
+    state_str = state.value if hasattr(state, "value") else state
+
     new_qreg = QubitHandler(
-        qalloc_p.bind(num_wires), self.qubit_index_recorder, dynamically_alloced=True
+        qalloc_p.bind(num_wires, state=state_str, precision=precision),
```
paul0403, line 406: You can add a state prep op after alloc so you don't have to change the underlying alloc op's definition. I don't think alloc op should take on responsibilities outside just allocation. 

Also, just defining a new mlir op/attribute won't make anything execute, you would need to add a mlir lowering to runtime for your new op.

JerryChen97: Yep right now the researcher's request is explicitly only a placeholder; but I am indeed also exploring how to add the corresponding runtime afterwards, will let you know if that starts to happen!

paul0403: > how to add the corresponding runtime afterwards

I think it will be preferable to use a stateprep op after the alloc, to not overload the alloc op's responsibilities. While the user might request a simple UI to indicate the state being allocated from inside the `qml.allocate` call, the implementation does not need to be bound by the UI.

PR #2711, opened 2026-04-17
Move DecompGraph library to MLIR

maliasadi:
**Context:**
DecompGraph is a standalone library required solely for the `graph_decomposition` MLIR pass. We move this library to the MLIR subdir closer to where is used/reqiured.

- Move the library from `./runtime` to `./mlir`

**Benefits:**
- Uncouple the graph decomposition pass and the runtime/lib build 

**Possible Drawbacks:**
- The Graph Boost Library dependency of `quantum-transforms`. 

**Related GitHub Issues:**
[sc-116823]


.github/workflows/check-catalyst.yaml
dime10: This was added in the previous pr I believe

.github/workflows/check-catalyst.yaml
```diff
@@ -600,12 +596,6 @@ jobs:
         name: quantum-build-${{ matrix.compiler }}
         path: quantum-build
 
-    - name: Download Catalyst-Runtime Artifact
```
dime10, line 603: why are we moving this one step down? probably doesn't make a difference but just checking

doc/releases/changelog-dev.md
```diff
@@ -612,9 +614,12 @@
 
 <h3>Internal changes ⚙️</h3>
 
+* Moved `DecompGraphSolver` to the core compiler to improve code locality, as it is used exclusively by the `graph_decomposition` pass.
+  [(#2711)](https://github.com/PennyLaneAI/catalyst/pull/2711)
```
dime10: This can be just be added to the existing entry form the PR we just merged :)
The changelog is meant to describe any changes with respect to the prior release, so PRs that keep updating the same component within one release cycle all go into one entry that describes the combined change compared to the previous release.

mlir/lib/Quantum/Transforms/DecompGraphSolver/CMakeLists.txt
```diff
@@ -0,0 +1,89 @@
...
+endif()
+
+
+add_library(decompsolver OBJECT DGBuilder.cpp DGSolver.cpp)
+
+install(TARGETS decompsolver
+        EXPORT CatalystTargets
+        LIBRARY DESTINATION lib
+        ARCHIVE DESTINATION lib
+        RUNTIME DESTINATION bin)
```
dime10: We don't really have install directives in our cmake files, can you describe the build strategy here?

mlir/lib/Quantum/Transforms/DecompGraphSolver/DGBuilder.hpp
kipawaa, line 1: To follow convention in `mlir/`, it would likely be best to move the headers to `mlir/include/`

mlir/lib/Quantum/Transforms/DecompGraphSolver/DGBuilder.cpp
kipawaa, line 1: I wonder if the decomp solver is better located in `Quantum/Utils` rather than `transforms`, since it doesn't implement a pass?

kipawaa (commented):
Thanks! Just a couple comments regarding organization here :)

mlir/lib/Quantum/Transforms/DecompGraphSolver/DGBuilder.hpp
maliasadi: Is this convention applied to C++ headers?

mlir/lib/Quantum/Transforms/DecompGraphSolver/CMakeLists.txt
```diff
@@ -0,0 +1,89 @@
+cmake_minimum_required(VERSION 3.26)
+
+set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
dime10: Not sure what standard procedure for cmake is, does it make sense to have different values for this within the same build tree? 
If not we can just set this in the top-level cmakelists. The CXX standard is already the same, so these lines would be redundant. The minimum cmake version is currently set to 3.20, if you need 3.26 I think we can update that for the entire project :)

mlir/lib/Quantum/Transforms/DecompGraphSolver/DGBuilder.cpp
maliasadi: I added the solver to `Transforms` to keep the solver close to the decomposition passes and impl. cc: @dime10

mlir/lib/Quantum/Transforms/CMakeLists.txt
```diff
@@ -37,16 +39,24 @@ set(LIBS
...
     MLIRQuantumPassIncGen
 )
 
-add_mlir_library(${LIBRARY_NAME} STATIC ${SRC} LINK_LIBS PRIVATE ${LIBS} DEPENDS ${DEPENDS})
+add_mlir_library(${LIBRARY_NAME} STATIC
+    ${SRC}
+    LINK_LIBS
+    PRIVATE
+    ${LIBS}
+    $<TARGET_OBJECTS:decompsolver>
```
dime10, line 53: I'm not familiar with this formulation, is it not sufficient to list the target (`decompsolver`) as a dependency?

dime10 (commented):
Thanks for moving this @maliasadi !

mlir/lib/Quantum/Transforms/DecompGraphSolver/DGBuilder.hpp
dime10: Hmm, we don't have strict rules here. Generally, reusable components do end up in the `mlir/include` folder, but we also have a few instances of "private" headers only used by source files in the same directory (e.g. see gradient transforms and quantum transforms).

Most MLIR code is C++ btw, including all those `.h` headers (that's just their convention).

mlir/lib/Quantum/Transforms/DecompGraphSolver/DGBuilder.cpp
dime10: I don't think directories named `utils` is something we should encourage except in very few special cases. For instance, a better name for a directory here is simply `Decomposition` or something similarly meaningful (`DecompGraphSolver` is also good imo).

mlir/lib/Quantum/Transforms/DecompGraphSolver/DGBuilder.hpp
maliasadi: I'll keep them in one place similar to gradient algo implementations.

.github/workflows/check-catalyst.yaml
maliasadi: Fixed [689b765](https://github.com/PennyLaneAI/catalyst/pull/2711/commits/689b7651cca3bd7bfd6d7b8c74b3f5d42894d51f) (old commit)

.github/workflows/check-catalyst.yaml
maliasadi: It's caused while fixing git merge conflicts with the other branch.  Fixed [c40a4ef](https://github.com/PennyLaneAI/catalyst/pull/2711/commits/c40a4efd27be10f0900bf62a9b193957eaaf7aa0)

mlir/lib/Quantum/Transforms/DecompGraphSolver/CMakeLists.txt
maliasadi: I've added the `decompsolver` to the Catalyst export set to ensure this will be built before `quantum-transforms` (the outer cmake) and is available at link time. But this didn't fix the problem:
```
CMake Error in cmake/modules/CMakeLists.txt:

  export called with target "quantum-transforms" which requires target

  "decompsolver" that is not in any export set.
```

I tried different things to fix this. tl;dr I learnt that if I want to tell a specific cmake target in MLIR to swallow a specific object wihtout creating a install directive that affects the whole project, I can always compile the code to an object file and then link/load it at link time instead of build time. 

This install directive is no longer needed and I removed it [c40a4ef](https://github.com/PennyLaneAI/catalyst/pull/2711/commits/c40a4efd27be10f0900bf62a9b193957eaaf7aa0)

mlir/lib/Quantum/Transforms/DecompGraphSolver/CMakeLists.txt
maliasadi: Fixed [c40a4ef](https://github.com/PennyLaneAI/catalyst/pull/2711/commits/c40a4efd27be10f0900bf62a9b193957eaaf7aa0)

mlir/lib/Quantum/Transforms/CMakeLists.txt
maliasadi: By using `$<TARGET_OBJECTS:decompsolver>`, I'm telling CMake to Take the compiled code from `decompsolver` and merge it directly into the outer scope cmake target `quantum-transforms`. I had to do this because the code is now inside `quantum-transforms`, CMake no longer sees `decompsolver` as an external dependency that needs to be exported separately. It effectively disappears into the final MLIR library, satisfying the strict requirements of `add_mlir_library`.

kipawaa (approved):
This is great, thanks so much!

dime10 (approved):
Thanks 🙂

PR #2713, opened 2026-04-20
Improve precompiled rule infrastructure

kipawaa:
**Context:**
The current system for precompiling decomposition rules appears flaky in CI.

Precompiles the rules at build time rather than import time, using bootstrapping to compile the decomp rules to bytecode during `make wheel`.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-117517]


dime10:
>The current system for precompiling decomposition rules appears flaky in CI.

Any clue as to why ? 👀

kipawaa:
> Any clue as to why ? 👀

Not 100% sure yet, but it's definitely to do with the bytecode file. It seems to me that the generation is being triggered, but somewhere between generating the bytecode and accessing it something goes wrong. My best guess is that the multiple-worker shared environment is causing one worker to access while another tries to re-compile and this in general causes issues. This is somewhat backed up by the success of the runners when the wheel tests are run one-by-one, but it's somewhat difficult to confirm. Another weird aspect of this is the failures only appear on macOS, not sure what that's about...

dime10:
> > Any clue as to why ? 👀
> 
> Not 100% sure yet, but it's definitely to do with the bytecode file. It seems to me that the generation is being triggered, but somewhere between generating the bytecode and accessing it something goes wrong. My best guess is that the multiple-worker shared environment is causing one worker to access while another tries to re-compile and this in general causes issues. This is somewhat backed up by the success of the runners when the wheel tests are run one-by-one, but it's somewhat difficult to confirm.

ohh yeah concurrency could be an issue 🤔 

> Another weird aspect of this is the failures only appear on macOS, not sure what that's about...

That tracks actually, the test schedule when running tests concurrently is dependent on the number of cores available, and the macos runners have a different core count than the linux ones. In the past we've often observed concurrency issues on the mac runner because of the different test schedule.
Sometimes you can replicate the CI issue locally merely by changing the number of cores used by pytest.

paul0403:
Any status updates here? A few things are waiting on catalyst package push to testpypi

kipawaa:
> Any status updates here? A few things are waiting on catalyst package push to testpypi

I'm not convinced that this solution is the correct approach. I had hoped this would be simpler, but using the partial install to precompile the rules during setup is proving not-so-successful. I think that we should re-consider tracking the bytecode file, but it might be best to have an in-depth discussion about this offline.

.github/workflows/build-wheel-linux-arm64.yaml
```diff
@@ -174,7 +175,7 @@ jobs:
 
     - name: Install Dependencies (Python)
       run: |
-        python${{ matrix.python_version }} -m pip install numpy nanobind pybind11 PyYAML cmake ninja
+        python${{ matrix.python_version }} -m pip install numpy nanobind pybind11 PyYAML cmake ninja jax==0.7.1 jaxlib==0.7.1 pennylane==0.45.0-dev84 setuptools scipy-openblas32>=0.3.26 xdsl==0.59.0 xdsl-jax==0.5.0
```
kipawaa: I think rather than this, we might be able to `pip install -r requirements.txt`? If we can't provide this via some file, we should likely have a reference to this somewhere as a reminder when updating dependencies.

dime10: So some action manually installed a (small) number of dependencies in order to save time. pip can be slow for a lot of packages, and when the main action on PRs was still running in 5min this could be a non-trivial portion of it. Not sure if we still need this strategy, especially for the wheels.

`requirements.txt` should never install jax, pennylane, or other runtime dependencies though, because our requirements file is intended to set up a build environment not a runtime environment.

dime10: The above is mostly general context. Looking at this specific line I don't think I would be comfortable merging hardcoded version numbers in the workflow files like this.

Makefile
```diff
@@ -258,9 +259,18 @@ wheel:
 		    cp $(COPY_FLAGS) $$file $$dest_dir; \
 	    done' sh {} +
 
+	$(PYTHON) -m pip wheel --no-deps . -w bootstrap_dist
+	$(PYTHON) -m pip install bootstrap_dist/*.whl
```
dime10: I'm wondering if we can just add `frontend` as a dependent target of `wheel`? Would that not be even simpler, since we have to have a working build already, just not installed, and `make frontend` would take care of that (hopefully including the dependency stuff you had to deal with?)

.github/workflows/build-wheel-linux-arm64.yaml
kipawaa: I guess that there is some argument to be made that these are build dependencies now that the bytecode generation uses the runtime... I see the idea though, and I certainly agree that hardcoding version numbers in the workflow files is not something we want to do. Would it be reasonable to change the format of `.dep-versions` so that it can be used for the pip runtime requirements and versions? Or some other central source of truth for dependency versions?

Makefile
kipawaa: This failed in [b152fd7](https://github.com/PennyLaneAI/catalyst/pull/2713/commits/b152fd70f2559a7c8934a02ca5acc1d4ca52ea8b) because the editable install wasn't able to access the dialects bindings, but this is an elegant solution. I'll keep working on this to see if it can be fixed.

.github/workflows/build-wheel-linux-arm64.yaml
dime10: > Would it be reasonable to change the format of .dep-versions so that it can be used for the pip runtime requirements and versions? Or some other central source of truth for dependency versions?

I think we could do this if necessary. I do have to ask though, why was the --no-deps required which consequently need to be installed manually in advance?

kipawaa: I was just following the format of the existing wheel infrastructure

dime10: I'm not sure I follow, where does the existing wheel infra rely on `--no-deps`? 

Maybe asked another way, if you didn't use that flag, wouldn't that make your addition on this line obsolete?

Makefile
dime10: Oh I think that's just because the discovery for `mlir_quantum` doesn't consider a custom build directory structure like we have in the CI:
https://github.com/PennyLaneAI/catalyst/blob/b152fd70f2559a7c8934a02ca5acc1d4ca52ea8b/frontend/catalyst/__init__.py#L55-L62

Should be an easy fix by looking at the appropriate environment variable first. I don't think we have a "run-time" env var for the mlir_quantum directory / the dialect build directory specifically, although we do have several already setup in a few locations in the frontend, for example here:
https://github.com/PennyLaneAI/catalyst/blob/b152fd70f2559a7c8934a02ca5acc1d4ca52ea8b/frontend/catalyst/utils/runtime_environment.py#L83

The DIALECTS_BUILD_DIR that we set in the CI is technically a build env var, but we could reuse or remap it as a run-time var (for use in the Python code itself). There is somewhat related open issue to clean up the path discovery a bit: https://github.com/PennyLaneAI/catalyst/issues/2730

.github/workflows/check-catalyst.yaml
```diff
@@ -474,26 +474,6 @@ jobs:
       with:
         python-version: ${{ needs.constants.outputs.primary_python_version }}
 
-    - name: Install Deps
```
paul0403, line 477: Sorry, why are these moved to after the build?

Makefile
```diff
@@ -122,6 +122,7 @@ frontend:
 	# versions of a package with the same version tag (e.g. 0.38-dev0).
 	$(PYTHON) -m pip uninstall -y pennylane
 	$(PYTHON) -m pip install -e . --extra-index-url https://test.pypi.org/simple $(PIP_VERBOSE_FLAG)
+	$(PYTHON) -m catalyst.utils.precompile_decomposition_rules
```
paul0403, line 125: Apologies, I finally have some time to look at some of the suggestions here in depth

The original flakiness in CI seems to be caused by different pytest runners having race condition regarding building the rules during `import catalyst`, so that will be solved now that we build the rules during `make frontend`, ala build time instead of user program runtime, so that's good.

But this introduces a circular dependency: to build catalyst, you need to already have catalyst (to build the rules). 

This situation is analogous to a very dumb confusion I had when I first started learning llvm/clang 6 years ago: it's a compiler for C++, but it itself is also written in C++! What do you mean! How am I compiling my compiler??

Eventually I realized the point is, there is an existing, fully-built clang, that is compiling a new version of clang that I am developing. That existing clang does not need to have any new code I wrote.

May I offer a suggestion that mirrors this realization? Because when running your `precompile_decomp_rules` script, the catalyst running that script does _not_ need to have the rules built. In that script, we start a temporary virtual enviornment, `pip install -e .` to install catalyst without rules (hence not invoking any makefile commands, but a raw `pip install -e .`).

The makefile looks the same as your changes: `make frontend` just runs the `precompile_decomp_rules` as an extra step.

The script would look something like this:
1. Save the path to the environment that it was launched from (call this environment X). 
2. Launch a new environment (call this environment Y)
3. In Y, `pip install -e . --extra-index-url https://test.pypi.org/simple`. This will install the catalyst package whose content is however the project that contains the script looks like, and automatically pull in all *runtime* requirements from testpypi. Note that these are runtime requirements managerd by the pypi index (i.e. `setup.py`), not from `requirements.txt`!
4. In Y, execute the `qjit` that will dump the rules
5. Move the rule files to appropriate locations in X.
6. Delete Y.

I think this way, we do not need to manage any packages/dependencies in X, or in any build scripts, unless I'm missing something.

What do you guys think?

.github/workflows/check-catalyst.yaml
kipawaa: Compiling the rules requires a complete build, i.e. we need dialects + frontend + runtime etc. To keep all of the frontend related steps together (building + compiling) I moved the frontend steps to the end.

.github/workflows/build-wheel-linux-arm64.yaml
kipawaa: Fixed in [9f0a607](https://github.com/PennyLaneAI/catalyst/pull/2713/commits/9f0a607c723946d35f25c6e28661d3228ad25463)

kipawaa: > where does the existing wheel infra rely on --no-deps?

Both `make wheel` and `make plugin-wheel` use `--no-deps`. I agree that it's nicer (especially in this scenario) to let pip resolve the dependencies. I just wanted to have the initial implementation be consistent with the existing implementations in case there was a reason I hadn't considered.

Makefile
kipawaa: Ah, good catch! I'd love to make this change, though in the interest of fixing wheels on main and other priority items, I'd like to stick to the current wheels approach if that's alright. I'd be happy to revisit the wheels process once the path discovery changes are implemented.

Makefile
dime10: > But this introduces a circular dependency: to build catalyst, you need to already have catalyst (to build the rules).

This is only sort-of true, make frontend is more like an install step than a build step 🤔 The only caveat to this is the two c++ files we build as part of the installation of the catalyst package (from setup.py), although ideally I'd like to see those moved into cmake at some point. 

> This situation is analogous to a very dumb confusion I had when I first started learning llvm/clang 6 years ago: it's a compiler for C++, but it itself is also written in C++! What do you mean! How am I compiling my compiler??

Indeed, you are describing the bootstrapping problem 😅 

The solution you describe seems equivalent though to what River is already doing though (from quick look), he's installing the package once (without rules), then building the rules, then building the package again.

paul0403: River just explained what he's doing to do in depth :+1: Thanks

.github/workflows/build-wheel-linux-arm64.yaml
dime10: Oh that's different, we added that so that pip doesn't _build_ wheels for our dependencies (which it otherwise would, so the dist folder would be full of wheels from all our dependencies which we don't need). For installation (which you're doing) you'd definitely want dependencies :)

Makefile
dime10: Logged in this existing issue: https://github.com/PennyLaneAI/catalyst/issues/2730

.github/workflows/build-wheel-linux-arm64.yaml
kipawaa: Ahhh okay, that was my misunderstanding! Thanks for the explanation :)

Makefile
dime10, line 125: I can't remember, is the script itself conditional or does rebuild every time (fine if it does)?

dime10 (approved):
Thanks @kipawaa !

Makefile
kipawaa: It was conditional when implemented on import, but the new implementation will recompile every time. I think it will be managed relatively naturally since it's added to `make frontend` and `make clean`.

paul0403 (approved):

PR #2714, opened 2026-04-20
Add noiseop qecl-to-qecp lowering pass to skeleton pass

multiphaseCFD:
### Before submitting

Please complete the following checklist when submitting a PR:

- [ ] All new functions and code must be clearly commented and documented.

- [x] Ensure that code is properly formatted by running `make format`.
      The latest version of black and `clang-format-20` are used in CI/CD to check formatting.

- [x] All new features must include a unit test.
      Integration and frontend tests should be added to [`frontend/test`](../frontend/test/),
      Quantum dialect and MLIR tests should be added to [`mlir/test`](../mlir/test/), and
      Runtime tests should be added to [`runtime/tests`](../runtime/tests/).

When all the above are checked, delete everything above the dashed
line and fill in the pull request template.

------------------------------------------------------------------------------------------------------------

**Context:**

[sc-117194]

This PR adds a few things:
- update the qecl-nosie-to-qecp-noise pass naming
- add the qecl-nosie-to-qecp-noise pass to the qecl-to-qecp skeleton pass
- add integration tests for the `quantum-to-qecl`, `inject-noiseop-to-qecl` and `qecl-to-qecp`
- update the data type of `qecl-nosie-to-qecp-noise` pass to ensure the inserted IR working with the integration tests
- update the unit tests for the qecl-nosie-to-qecp-noise pass to keep them up with the data type changes made in `qecl-nosie-to-qecp-noise` pass 

**Description of the Change:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

PR #2716, opened 2026-04-20
Miscellaneous improvements to qecp dialect, QEC code UI and qecp noise pass

joeycarter:
- Remove `ContainerOf` attribute aliases in qecp dialect (these were incorrect) and replace with `base(...)`.
- Remove `QecCodeKey` class from `convert-qecl-to-qecp` pass and replace with new `QecCode` interface.
- Add type-conversion patterns to the `convert-qecl-to-qecp` pass.
- Commits the changes made in #2714


frontend/catalyst/python_interface/dialects/qecp.py
```diff
@@ -242,10 +242,6 @@ def parse_parameters(cls, parser: AttrParser) -> Sequence[Attribute]:
 PhysicalHyperRegisterSSAValue: TypeAlias = SSAValue[PhysicalHyperRegisterType]
 TannerGraphSSAValue: TypeAlias = SSAValue[TannerGraphType]
 
-anyPhysicalQubit = ContainerOf(QecPhysicalQubitType)
-anyPhysicalCodeblock = ContainerOf(PhysicalCodeblockType)
-anyPhysicalHyperRegister = ContainerOf(PhysicalHyperRegisterType)
```
joeycarter, line 247: See discussion here for context on this change: https://github.com/PennyLaneAI/catalyst/pull/2589/changes#r3090214623

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -18,26 +18,48 @@
...
+@dataclass
+class CodeblockTypeConversion(TypeConversionPattern):
+    """Codeblock type conversion pattern from qecl.codeblock -> qecp.codeblock."""
+
+    qec_code: QecCode = field(kw_only=True)
+
+    @attr_type_rewrite_pattern
+    def convert_type(self, typ: qecl.LogicalCodeblockType) -> qecp.PhysicalCodeblockType:
+        """Type conversion rewrite pattern for logical codeblock types."""
+        return qecp.PhysicalCodeblockType(typ.k, self.qec_code.n)
```
lillian542, line 50: Should we check that `typ.k` and `self.qec_code.k` match?

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -51,22 +73,20 @@ class ConvertQecLogicalToQecPhysicalPass(ModulePass):
...
-                    # Insert conversion patterns here
+                    CodeblockTypeConversion(qec_code=self.qec_code),
+                    HyperRegisterTypeConversion(qec_code=self.qec_code),
                 ]
             )
         ).rewrite_module(op)
 
-        # Certain patterns leave behind `builtin.unrealized_conversion_cast` ops;
-        # this pass removes them
-        ReconcileUnrealizedCastsPass().apply(ctx, op)
```
lillian542, line 69: So the idea here is that we won't need the casts anymore because we will use the patterns above instead, right? And that means that other patterns will come after and will already have the correct code-block types, they'll just still be `qecl` operations.

frontend/catalyst/python_interface/transforms/qecp/qec_code_lib.py
```diff
@@ -0,0 +1,54 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""This module contains a library of QEC codes."""
+
+from dataclasses import dataclass
+
+_CODE_REGISTRY: dict[str, tuple[int, int, int]] = {
+    "Steane": (7, 1, 3),
```
lillian542, line 20: I'm going to add an array for an `x_tanner` graph and an array for a `z_tanner` graph here as well in the `qecl.encode` lowering PR (I'll leave discussion of the exact format for that PR), and that should fit well with this format, but I think we'll hit a limitation when its time to add lowering for gates. 

Not something to worry about now, but just mentioning it as something to mull over, because I imagine we'll need to figure it out next iteration.

lillian542 (approved):
Looks great, @joeycarter 🎉  My comments are non-blocking, so I'll go ahead and approve.

multiphaseCFD (approved):
LGTM! Thanks @joeycarter

multiphaseCFD:
I tested this branch in the integration tests [here](https://github.com/PennyLaneAI/catalyst/blob/1a2fe2bdd64f7c0a65295b656dbcb6b4b2cfee64/frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py#L49) . It failed locally, `QecCode` is not compatible with catalyst. here is the error msg:  `catalyst.utils.exceptions.CompileError: Cannot convert Python type <class 'catalyst.python_interface.transforms.qecp.qec_code_lib.QecCode'> to an MLIR attribute.`

joeycarter:
> I tested this branch in the integration tests [here](https://github.com/PennyLaneAI/catalyst/blob/1a2fe2bdd64f7c0a65295b656dbcb6b4b2cfee64/frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py#L49) . It failed locally, `QecCode` is not compatible with catalyst. here is the error msg: `catalyst.utils.exceptions.CompileError: Cannot convert Python type <class 'catalyst.python_interface.transforms.qecp.qec_code_lib.QecCode'> to an MLIR attribute.`

This error will be fixed with #2719.

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -51,22 +73,20 @@ class ConvertQecLogicalToQecPhysicalPass(ModulePass):
 
     name = "convert-qecl-to-qecp"
 
-    qec_code: QecCodeKey
+    qec_code: QecCode
```
multiphaseCFD, line 90: With the latest change in the dataclass support for the options, qec_code here is not a QecCode anymore and is converted to a dict instead.

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
multiphaseCFD, line 58: consequently, we can't do attribute access with `.n` here, instead we need to access it with  `self.qec_code['n']`

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -18,26 +18,48 @@
...
+@dataclass
+class HyperRegisterTypeConversion(TypeConversionPattern):
+    """Hyper-register type conversion pattern from qecl.hyperreg -> qecp.hyperreg."""
+
+    qec_code: QecCode = field(kw_only=True)
+
+    @attr_type_rewrite_pattern
+    def convert_type(self, typ: qecl.LogicalHyperRegisterType) -> qecp.PhysicalHyperRegisterType:
+        """Type conversion rewrite pattern for physical codeblock types."""
+        return qecp.PhysicalHyperRegisterType(typ.width, typ.k, self.qec_code.n)
```
multiphaseCFD, line 76: same here

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
multiphaseCFD: I guess this pass would be only applied for the integration tests? or we want to do unit test for this skeleton pass as well?

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
joeycarter: Good call, added here: f08e6d4ac

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
joeycarter: That's it exactly. We can get away with this here since there is sufficient information in the IR to do the type conversion automatically from qecl types to qecp types (in part since the QEC code is known). In the quantum -> qecl dialect conversion, there was insufficient information, since `!quantum.reg` does not have an attribute specifying its width, which is required by `qecl.hyperreg`. This meant that we needed a chain of `unrealized_conversion_cast` ops stretching back to an `alloc` op to get the register width, which we resolve by applying `ReconcileUnrealizedCastsPass`.

frontend/catalyst/python_interface/transforms/qecp/qec_code_lib.py
joeycarter: Agreed, we can find a better approach when it becomes necessary!

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
joeycarter: Shoot, that shouldn't be happening; something is overwriting `qec_code` with the dictionary. I'm investigating...

joeycarter: Aha, it's taking the dict attribute from the `transform.apply_registered_pass` op and creating a new instance of the pass with that dictionary:

https://github.com/PennyLaneAI/catalyst/blob/359c1d13c751824db75ad781311abbf569178767/frontend/catalyst/python_interface/pass_api/apply_transform_sequence.py#L250-L263

We could probably fix this by adding a constructor to `QecCode` to build an instance from a dictionary. Let me try that out.

multiphaseCFD: > We could probably fix this by adding a constructor to QecCode to build an instance from a dictionary. Let me try that out.

Sounds like a good idea!

joeycarter: Hopefully this is fixed now: eb5c6d0c2

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
joeycarter: Following up in https://github.com/PennyLaneAI/catalyst/pull/2716#discussion_r3119259774

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
joeycarter: Following up in https://github.com/PennyLaneAI/catalyst/pull/2716#discussion_r3119259774

frontend/test/pytest/python_interface/transforms/qecl/test_xdsl_convert_quantum_to_qecl.py
```diff
@@ -824,7 +824,7 @@ def test_circuit_basic(self, run_filecheck_qjit):
         """Test the convert-quantum-to-qecl pass on the simplest possible, non-trivial circuit."""
         dev = qml.device("null.qubit", wires=1)
 
-        @qml.qjit(target="mlir", keep_intermediate=True)
+        @qml.qjit(target="mlir")
```
joeycarter, line 827: Another opportunistic fix: we shouldn't have `keep_intermediate=True` in tests (that was my bad).

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
joeycarter: I'd say this is resolved now :relieved:

PR #2717, opened 2026-04-21
[qref 3.4] Migrate while loop's plxpr conversion to reference semantics

paul0403:
**Context:**
Migrate while loop's plxpr conversion to reference semantics

Delete the conversion from plxpr while loop primitive to catalyst while loop primitive, so we just use the lowering of plxpr while loop primitive directly.

**Benefits:**
Much cleaner conversion of plxpr to catalys jaxpr.

[sc-115871]


albi3ro (approved):
❤️ 

Similar comment as to for_loop PR. This looks amazing! And going to trust you that the CI will all come together.

JerryChen97 (approved):

PR #2718, opened 2026-04-21
fix: 2 improvements across 2 files

tomaioo:
## Summary

fix: 2 improvements across 2 files

## Problem

**Severity**: `High` | **File**: `bin/utils.py:L40`

The function `get_cpp_files_from_path` assigns `exts = HEADERFILE_EXT` and then does `exts += SRCFILE_EXT` when `header_only` is false. Because `exts` references the global list, this permanently mutates `HEADERFILE_EXT`, causing later calls (including `header_only=True`) to incorrectly include source files.

## Solution

Use a copy when building the extension list, e.g. `exts = list(HEADERFILE_EXT)` and then `exts.extend(SRCFILE_EXT)`.

## Changes

- `bin/utils.py` (modified)
- `.github/workflows/set_nightly_version.py` (modified)

### DIFF
--- a/.github/workflows/set_nightly_version.py
+++ b/.github/workflows/set_nightly_version.py
@@ -29,7 +29,7 @@ with open(version_file_path, "r+", encoding="UTF-8") as f:
     version_line = lines[-1]
     assert "__version__ = " in version_line
 
-    pattern = r"(\d+).(\d+).(\d+)-dev(\d+)"
+    pattern = r"^(\d+)\.(\d+)\.(\d+)-dev(\d+)$"
     match = re.search(pattern, version_line)
     assert match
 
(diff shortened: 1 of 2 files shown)


.github/workflows/set_nightly_version.py
```diff
@@ -29,7 +29,7 @@
     version_line = lines[-1]
     assert "__version__ = " in version_line
 
-    pattern = r"(\d+).(\d+).(\d+)-dev(\d+)"
+    pattern = r"^(\d+)\.(\d+)\.(\d+)-dev(\d+)$"
```
dime10, line 32: What is this change about?

dime10: @tomaioo could you elaborate on this fix here, what is it doing?

tomaioo:
Great question — the fix prevents accidental mutation of the global `HEADERFILE_EXT` list.  
Previously, `exts = HEADERFILE_EXT` created an alias, so `exts += SRCFILE_EXT` modified the original global list and caused later calls (even `header_only=True`) to include source extensions.  
Now we build `exts` from a copy (`list(HEADERFILE_EXT)`) and only extend that local list, so each call gets the correct, isolated extension set.

PR #2719, opened 2026-04-21
Add support for conversion of dataclasses to MLIR dict attributes

joeycarter:
The function [`get_mlir_attribute_from_pyval()`](https://github.com/PennyLaneAI/catalyst/blob/1b067e0f29f76148ebdb55a5fa2697c7d39ecc80/frontend/catalyst/jax_extras/lowering.py#L191-L236), which converts Python types to MLIR attributes, does not currently support Python `dataclass`es. A `dataclass` is easily convertible to a `dict` object via [`dataclasses.asdict()`](https://docs.python.org/3/library/dataclasses.html#dataclasses.asdict), therefore we can easily support the conversion of a `dataclass` object to an MLIR dictionary attribute.

Having this conversion is particularly useful for developers who wish to use a `dataclass` object as an xDSL pass option.

[sc-117368]

### DIFF
--- a/frontend/catalyst/jax_extras/lowering.py
+++ b/frontend/catalyst/jax_extras/lowering.py
@@ -233,12 +233,12 @@ def get_mlir_attribute_from_pyval(value):
                 named_attrs[k] = get_mlir_attribute_from_pyval(v)
             attr = ir.DictAttr.get(named_attrs)
 
+        case _ if dataclasses.is_dataclass(value):
+            attr = get_mlir_attribute_from_pyval(dataclasses.asdict(value))
+
         case _:
-            if dataclasses.is_dataclass(value):
-                attr = get_mlir_attribute_from_pyval(dataclasses.asdict(value))
-            else:
-                raise CompileError(
-                    f"Cannot convert Python type {type(value)} to an MLIR attribute."
-                )
+            raise CompileError(
+                f"Cannot convert Python type {type(value)} to an MLIR attribute."
+            )
 
     return attr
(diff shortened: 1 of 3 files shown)


frontend/catalyst/jax_extras/lowering.py
```diff
@@ -233,6 +234,11 @@ def get_mlir_attribute_from_pyval(value):
             attr = ir.DictAttr.get(named_attrs)
 
         case _:
-            raise CompileError(f"Cannot convert Python type {type(value)} to an MLIR attribute.")
+            if dataclasses.is_dataclass(value):
+                attr = get_mlir_attribute_from_pyval(dataclasses.asdict(value))
```
multiphaseCFD: just wondering if it still works for the integration tests if there is an variable in the dataclass is a `list`, which is not hashable?

frontend/catalyst/jax_extras/lowering.py
```diff
@@ -233,6 +234,11 @@ def get_mlir_attribute_from_pyval(value):
...
+            else:
+                raise CompileError(
+                    f"Cannot convert Python type {type(value)} to an MLIR attribute."
+                )
```
paul0403:
```python3
    match value:
        case _ if is_dataclass(value):
            print("This is a dataclass!")
        case _:
            print("default")
```

There's a nice syntax sugar to save some indenting

multiphaseCFD (commented):
LGTM! Thanks @joeycarter ! Just one question

paul0403 (approved):
Nice!

frontend/catalyst/jax_extras/lowering.py
paul0403: Yes, this helper function is recursive and you can see a case for list

@joeycarter can you add a test?

multiphaseCFD (approved):

frontend/catalyst/jax_extras/lowering.py
joeycarter: Done: a5d15a795

frontend/catalyst/jax_extras/lowering.py
joeycarter: Nice, thanks! 847845d90

Issue #2516, opened 2026-02-25
Do not resolve adjoints on the gate operations' cannonicalization themselves; use a pass instead


paul0403:
When a gate operation carries the `adjoint` attribute, we cannonicalize it away in the operation's canonicalizer itself. For example, [this](https://github.com/PennyLaneAI/catalyst/blob/788b78eb94999bca06b34f8133691d2b300afe7d/mlir/lib/Quantum/IR/QuantumOps.cpp#L48) is the one on `quantum.custom` op. We see that Hermitian gates have their adjoint attribute removed, and rotation gates have the rotation angle negated.

However, maybe we shouldn't automatically (because canonicalizations happen very frequently) optimize the adjoint of operations away, because:

- quantum compilation is often a bit more deliberate from the user side
- a user may also expect to see an adjointed op when inspecting a circuit (because that's what they wrote)
- the adjoint carries semantic meaning

A more controllable approach would simply wrap this is a pass.

This was proposed by Romain when we first started to implement merge rotations (for OQD). At the time we didn't want to deal with all the adjoint possibilities in the merge rotation pass, so we just decided to canonicalize the adjoint away on the gate itself.

But of course, pulling it out into its own pass makes the most sense. The merge rotation pass can just call that pass before merging, if needed.

PR #2720, opened 2026-04-21
[qref 3.5] Migrate adjoint op's plxpr conversion to reference semantics

paul0403:
**Context:**
Migrate adjoint op's plxpr conversion to reference semantics

Add qref adjoint primitive and update conversion to bind qref adjoint primitive.

[sc-115873]
[sc-102216]


doc/releases/changelog-dev.md
```diff
@@ -761,6 +761,7 @@
   [(#2672)](https://github.com/PennyLaneAI/catalyst/pull/2672)
   [(#2694)](https://github.com/PennyLaneAI/catalyst/pull/2694)
   [(#2717)](https://github.com/PennyLaneAI/catalyst/pull/2717)
+  [(#2720)](https://github.com/PennyLaneAI/catalyst/pull/2720)
```
paul0403, line 774: self reminder to add ctrl+adjoint tests

frontend/test/pytest/test_dynamic_qubit_allocation.py
```diff
@@ -534,6 +534,25 @@ def circuit():
     assert np.allclose(observed, expected)
 
 
+@pytest.mark.usefixtures("use_capture")
+def test_adjoint(backend):
+    """
+    Test adjoints work.
+    """
+
+    @qjit
```
albi3ro:
```suggestion
def test_adjoint(backend):
    """
    Test adjoints work.
    """

    @qjit(capture=True)
```

frontend/test/pytest/test_dynamic_qubit_allocation.py
```diff
@@ -534,6 +534,25 @@ def circuit():
...
+    @qp.qnode(qp.device(backend, wires=2))
+    def circuit():
+        with qp.allocate(1) as q:
+            qp.adjoint(qp.X)(q[0])
```
albi3ro: Should we test with an operator that isn't self adjoint if we are checking the result? Maybe an `RX` with checking `qml.expval(Y)`?

paul0403: So when we print `QJIT.mlir`, the printed IR actually has canonicalization run on it, and the Hermitian gates have canonicalization pattern to remove the adjoint, and rotations have canonicalization pattern to negate the angles.

We do want to move this out of the canonicalization, due to issues like the one you're describing, but also just due to pre mature optimization changing semantics of the program https://github.com/PennyLaneAI/catalyst/issues/2516

paul0403: But you're right, changing this to a non Hermitian gate would be better for this test

paul0403: Changed to 
```python
        with qp.allocate(1) as q:
            qp.adjoint(qp.RX)(0.12, q[0])
            qp.RX(0.12, q[0])
            qp.CNOT(wires=[q[0], 0])
        return qp.probs(wires=[0, 1])
```

so the cancelling will test the correctness of the adjoint

albi3ro (approved):
❤️

PR #2721, opened 2026-04-21
[qref 2.5.2] Small follow-ups to the ref-to-v MLIR pass

paul0403:
**Context:**
3 small fixes to the reference-to-value semantics conversion pass.

1. `qref.alloc` ops are cleaned at the end of the conversion explicitly, instead of relying on the canonicalizer
2. Functions explicitly marked with `quantum.node` attribute will not be mistakenly classified as qref subroutines
3. `qref.adjoint` ops will always be handled, regardless of whether it uses any quantum values from outside. This is because unlike the scf regioned operations, `qref.adjoint` is a quantum operation itself, so it will never be "purely classical".  

**Benefits:**
Correctness


rniczh (approved):
👍

mehrdad2m (approved):

PR #2722, opened 2026-04-21
refactor: all passes in `builtin_passes.py` are now `Transform` objects

### LINKED ISSUE #2766: Cannot use integer weight in `gate_set`
One design choice in graph decomposition feels a little unfriendly: weights in gate_set cannot be integers, so in the situation below it only works if you change 1 to 1.0:
```python
@graph_decomposition(
    gate_set={qp.RX: 1.0, qp.RY: 1, qp.Rot: 5.0},
    fixed_decomps={qp.PauliX: x_to_rx, qp.PauliY: y_to_ry},
    alt_decomps={qp.H: [h_to_rx_ry]},
)
```

Reproduce
```python
import pennylane as qp
import pennylane.numpy as np

from catalyst import qjit
from catalyst.jax_primitives import decomposition_rule
from catalyst.passes import cancel_inverses, graph_decomposition, merge_rotations


@decomposition_rule(op_type=qp.PauliX)
def x_to_rx(wire: int):
    qp.RX(np.pi, wire)


@decomposition_rule(op_type=qp.PauliY)
def y_to_ry(wire: int):
    qp.RY(np.pi, wire)


@decomposition_rule(op_type=qp.Hadamard)
def h_to_rx_ry(wire: int):
    qp.RX(np.pi / 2, wire)
    qp.RY(np.pi / 2, wire)


@qjit(capture=True)
@graph_decomposition(gate_set={qp.Rot})
@merge_rotations
@graph_decomposition(
    gate_set={qp.RX: 1.0, qp.RY: 1, qp.Rot: 5.0},
    fixed_decomps={qp.PauliX: x_to_rx, qp.PauliY: y_to_ry},
    alt_decomps={qp.H: [h_to_rx_ry]},
)
@cancel_inverses
def circuit(x: float, y: float):
    qp.H(0)
    qp.H(0)
    qp.RX(x, wires=0)
    qp.PauliX(0)
    qp.RY(y, wires=0)
    qp.PauliY(0)
    qp.RY(x + y, wires=0)

    # register custom decomposition rules, required
    # when using the decomposition_rule decorator
    x_to_rx(int)
    y_to_ry(int)
    h_to_rx_ry(int)

    return qp.state()

print(qp.specs(circuit, level="device")(1.23, 4.56).resources.gate_types)
```

andrijapau:
**Context:**

Right now, we can't do this,
```python
>>> from catalyst.passes import to_ppr
>>> import pennylane as qp
>>> qp.CompilePipeline(to_ppr)
TypeError: CompilePipeline can only be constructed with a series of transforms or compile pipelines, or with a single list of transforms.
```

because these passes (although act like `Transform` objects), are not true `Transform` objects.

- Instead of defining them as decorators, they are now true `Transform` objects.
- Added a testing file for these passes where we assert that they are `Transform` objects.

**Benefits:** Behave truly as `Transform` objects. 

**Possible Drawbacks:** 

- In the [render](https://xanaduai-pennylane--2722.com.readthedocs.build/projects/catalyst/en/2722/code/__init__.html#module-catalyst-passes), passes no longer have a link to their source code. This isn't that alarming as these transforms, when viewed from PennyLane docs, also don't have a source code link. This is simply due to the implementation of docstrings being attached to the pass' setup inputs function.
- Users must manually remember to add their new pass to `__all__` or else it won't render on the website.
- IDEs can't pick up the docstring or the signature of these objects

Fixes #2766 

[sc-117354] [sc-118014]


doc/code/__init__.rst
```diff
@@ -22,14 +22,18 @@ Module: catalyst.debug
...
-.. automodapi:: catalyst.passes
-    :no-heading:
-    :no-inheritance-diagram:
-    :fullname:
+.. currentmodule:: catalyst.passes
+
+.. autosummary:: 
+    :toctree: api
+
+    ~to_ppr
```
andrijapau:
```suggestion
    ~to_ppr
    ~commute_ppr
```

doc/conf.py
```diff
@@ -163,6 +163,8 @@ def __getattr__(cls, name):
     "READTHEDOCS_CANONICAL_URL", "https://docs.pennylane.ai/projects/catalyst/"
 )
 
+os.environ["SPHINX_BUILD"] = "1"
```
andrijapau, line 166: Because these `Transform` objects are passes without a tape transform, this is needed to be able to display the signature. We do the same thing over in PennyLane.

frontend/catalyst/passes/utils.py
```diff
@@ -49,19 +49,24 @@ def prepare_decomposition_options(
...
-        "bytecode-rules": str(_builtin_rule_path),
+        "gate_set": gate_set,
+        "bytecode_rules": str(_builtin_rule_path),
     }
 
     if fixed_decomps:
         options |= {
-            "fixed-decomps": {to_name(op): rule.__name__ for op, rule in fixed_decomps.items()}
+            "fixed_decomps": {
+                to_name(op): (rule if isinstance(rule, str) else rule.__name__)
```
andrijapau: This makes the `setup_inputs` idempotent. This is required for the current implementation of `Transform` as during execution the outputs are fed as inputs for some reason.

doc/code/__init__.rst
```diff
@@ -22,14 +22,33 @@ Module: catalyst.debug
...
+    ~cancel_inverses
+    ~merge_rotations
+    ~decompose_lowering
+
 
 Module: catalyst.third_party.cuda
 ---------------------------------
 
+.. currentmodule:: catalyst
+
```
andrijapau:
```suggestion
```

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -435,10 +450,15 @@ def circuit(x: float):
...
+    return (), {}
 
 
-def decompose_lowering(qnode):
+merge_rotations = qml.transform(
+    pass_name="merge-rotations", setup_inputs=merge_rotations_setup_inputs
+)
+
+
+def decompose_lowering_setup_inputs():  # pragma: no cover
```
albi3ro, line 461: Why can't we test this?

albi3ro: And doesn't decompose have a whole bunch of options?

andrijapau: I think you're thinking of `graph_decomposition` (also in the same file)

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -1552,17 +1584,21 @@ def circuit():
...
 
-def graph_decomposition(
-    qnode=None,
-    *,
+decompose_arbitrary_ppr = qml.transform(
+    pass_name="decompose-arbitrary-ppr", setup_inputs=decompose_arbitrary_ppr_setup_inputs
+)
+
+
+def graph_decomposition_setup_inputs(
```
albi3ro, line 1595: What's the difference between this and decompose? I'm really confused. Also, where did bytecode_rules come from?

andrijapau: > Also, where did bytecode_rules come from?

It comes from `prepare_decomposition_options` in `frontend/catalyst/passes/utils.py`. 

That helper adds `bytecode_rules` to the options so to make it idempotent you need to add it also to the signature of `setup_inputs`

andrijapau: > What's the difference between this and decompose? 

I thought this is the MLIR pass?

frontend/test/pytest/test_builtin_passes.py
albi3ro, line 1: I'm not sure this should actually be something we need to test.  I think we should just be unit testing each transform to make sure it actually is a Transform and call it good.

andrijapau: True, that makes more sense. I can update it.

andrijapau: Should I just manually import all of the passes we have at the moment and pass them into a test that essentially does `isinstance(pass, Transform)`? 

If someone adds another pass, they would have to manually add it to the test ... any way to make this dynamic?

frontend/test/pytest/test_builtin_passes.py
```diff
@@ -0,0 +1,36 @@
...
+}
+
+
+@pytest.mark.parametrize("pass_", AVAILABLE_PASSES)
+def test_integration_with_compile_pipeline(pass_):
+    """Tests that these passes can be fed into a 'qp.CompilePipeline' object."""
+
+    pipeline = qp.CompilePipeline(pass_)
+    assert len(pipeline) == 1
+    assert pipeline[0].pass_name == pass_.pass_name
```
albi3ro: So these tests wouldn't have caught the bug, since the fundamental problem is that they *weren't* `Transform` objects, not that anything was going wrong with things that were.  Hence why I'm kinda thinking we need some sort of `assert_valid_transform` that we test on each thing that's a transform.

Perhaps instead we just track down the unit tests for each pass, and just make sure `isinstance(thing_that_should_be_transform, Transform)`?

andrijapau: There are no unit tests for each pass :\

andrijapau: Let me try something out ...

doc/code/__init__.rst
```diff
@@ -23,9 +23,7 @@ Module: catalyst.passes
 ---------------------------------
 
 .. automodapi:: catalyst.passes
-    :no-heading:
-    :no-inheritance-diagram:
-    :fullname:
+    :members:
```
andrijapau, line 28:
```suggestion
    :no-heading:
    :no-inheritance-diagram:
    :members:
```

frontend/test/pytest/test_builtin_passes.py
andrijapau: What do you think?

frontend/test/pytest/test_builtin_passes.py
```diff
@@ -0,0 +1,53 @@
...
+    # Must be insertable into a qp.CompilePipeline
+    try:
+        pipeline = CompilePipeline(obj)
+    except Exception as e:
+        raise AssertionError(f"Cannot be inserted into a CompilePipeline: {e}")
+    assert len(pipeline) == 1
+    assert pipeline[0].pass_name == pass_name
+
+
+@pytest.mark.parametrize("name", builtin_passes.__all__)
```
andrijapau, line 48: Devs would just need to make sure `__all__` is updated whenever another pass is introduced.

frontend/test/pytest/test_builtin_passes.py
andrijapau, line 1: Added this file as it should exist for unit testing these objects.

frontend/catalyst/passes/builtin_passes.py
kipawaa: > doesn't decompose have a whole bunch of options?

@albi3ro `decompose_lowering` is used by `graph_decomposition`, but itself just searches for functions with a `target_gate` attribute and replaces all instances of `target_gate` with those functions, so it takes no options. `graph_decomposition` creates + solves the graph with all available (precompiled + user) decomposition rules, then applies `decompose_lowering` with the chosen optimal-cost decomposition rules. It takes a number of options (`gate_set`, `fixed`/`alt_decomps` and `bytecode_rules`).

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -578,7 +603,12 @@ def circuit():
...
-    return qp.transform(pass_name="ions-decomposition")(qnode)
+    return (), {}
+
+
+ions_decomposition = qp.transform(
+    pass_name="ions-decomposition", setup_inputs=ions_decomposition_setup_inputs
+)
 
 
 def gridsynth(qnode=None, *, epsilon=1e-4, ppr_basis=False):
```
kipawaa: Should `gridsynth` be refactored as well?

frontend/catalyst/passes/builtin_passes.py
kipawaa: > What's the difference between this and decompose?

I saw the other comment first so I [replied there](https://github.com/PennyLaneAI/catalyst/pull/2722/changes#r3154594970) explaining the difference between `decompose_lowering` and `graph_decomposition`

frontend/test/lit/test_decomposition.py
```diff
@@ -1654,7 +1654,7 @@ def h_to_rx_ry(wire):
     # CHECK-DAG: "gate-set" = {Hadamard = 1.000000e+00 : f64, RX = 1.000000e+00 : f64, RY = 1.000000e+00 : f64}
     # CHECK-DAG: "fixed-decomps" = {PauliX = "x_to_rx", PauliY = "y_to_ry"}
     # CHECK-DAG: "alt-decomps" = {Hadamard = ["h_to_rx_ry"]}
-    # CHECK-DAG: "bytecode-rules" = "/decomp_rules.mlirbc"
+    # CHECK-DAG: "bytecode-rules" = "{{.*}}decomposition_rules_{{.*}}.mlirbc"
```
kipawaa, line 1657: Thanks! I definitely should have written this in the first place 😅

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -1554,17 +1586,21 @@ def circuit():
...
+decompose_arbitrary_ppr = qp.transform(
+    pass_name="decompose-arbitrary-ppr", setup_inputs=decompose_arbitrary_ppr_setup_inputs
+)
+
+
+def graph_decomposition_setup_inputs(
     gate_set: Iterable[type | str] | dict[type | str, float],
     fixed_decomps: dict | None = None,
     alt_decomps: dict | None = None,
+    bytecode_rules: str | None = None,
```
kipawaa, line 1599: It could be nice to unify the names across the transform, pass, `setup_options`, and `prepare_options`. I think this would remove the need for the duplicate parameter, which would feel much better :)

kipawaa (commented):
Looks great, thank you for updating this!

frontend/catalyst/passes/builtin_passes.py
andrijapau: Whoops. Good catch! 😓 

Fixed here: https://github.com/PennyLaneAI/catalyst/pull/2722/commits/fe28ab25d40f4eccbed3311b022477e41b584aa3

frontend/catalyst/passes/builtin_passes.py
andrijapau: I feel like I can just move that logic into setup_options and delete that utills.py file? What do you think?

kipawaa: Now that you mention it, they're definitely redundant! Good call!

andrijapau: Trying it out here: https://github.com/PennyLaneAI/catalyst/pull/2722/commits/ad6911fbfd30da81fd9fd85e0b55335c30ca5294

frontend/catalyst/passes/utils.py
andrijapau, line 1: No longer needed as this is taken care of by the setup_inputs function of graph_decomposition

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -1653,20 +1689,57 @@ def circuit(x: float, y: float):
...
-    options = prepare_decomposition_options(
-        gate_set=gate_set,
-        fixed_decomps=fixed_decomps,
-        alt_decomps=alt_decomps,
-        _builtin_rule_path=_builtin_rule_path,
-    )
+    if not isinstance(gate_set, dict):
+        gate_set = {to_name(op): 1.0 for op in gate_set}
+    else:
+        gate_set = {to_name(op): cost for op, cost in gate_set.items()}
```
kipawaa:
```suggestion
        gate_set = {to_name(op): float(cost) for op, cost in gate_set.items()}
```
Since we're here anyways, would you mind adding this to close https://github.com/PennyLaneAI/catalyst/issues/2766? 👼

frontend/catalyst/passes/builtin_passes.py
kipawaa: Great, thanks!

frontend/catalyst/passes/builtin_passes.py
kipawaa: Looks great, thanks! Do you think we could combine the `_builtin_rule_path` and `bytecode_rules` since we don't need idempotence through the ~setup~ prepare function anymore? We would need to update the MLIR pass parameter as well, since the names need to match, but this would be very nice!

frontend/catalyst/passes/builtin_passes.py
dime10, line 1: I haven't checked this for this PR, but in general I am concerned that we have moved and keep moving in a direction where our code base is less and less friendly for developer tooling. Since you mention something along those lines in the PR description, it would be good assess whether there is any impact on developer tooling before merging.
For instance, what is the type of `cancel_inverses`, will an IDE pick up and display its docstring correctly, will it able to show you the expected signature as you're typing it out?

If the answer is negative to any of these questions I would strongly advise to reconsider this system. As a user of PennyLane/Catalyst it is super frustrating when these quality of life tools don't work when we've come to expect them. For context, I reported a few other ways in which the tool-friendliness has been degrading in our libraries [here](https://xanaduhq.slack.com/archives/C06VCNYC3LH/p1772660644341259).

frontend/catalyst/passes/__init__.py
andrijapau, line 1: Is this a breaking change @isaacdevlugt @kipawaa?

Idk why it was added here, it was only used internally in `builtin_passes.py`

kipawaa: This was added this release cycle, so not a breaking change. I think it was added here to avoid a circular dependency issue 😕 very happy to have it removed from here!

frontend/catalyst/passes/builtin_passes.py
andrijapau: > will an IDE pick up and display its docstring correctly, will it able to show you the expected signature as /you're typing it out?

With `Transform`'s current implementation, no. ⬇️ 

image

Although, FWIW, I agree. Although, I feel like the fix here (getting these passes to be true `Transform` objects) is more important for the release.  

I wonder if this can be prioritized next quarter, CC @isaacdevlugt

frontend/catalyst/passes/__init__.py
andrijapau: Nice!!

frontend/catalyst/passes/builtin_passes.py
andrijapau: Sure! Why not.

andrijapau: Done. Also, I added this PR as an entry to the `graph_decomposition` entry as its a bug fix for that feature.

kipawaa: Thanks a ton!

frontend/catalyst/passes/builtin_passes.py
albi3ro: Sounds like a problem with `Transform`, and not with things being `Transform`.

frontend/catalyst/passes/builtin_passes.py
andrijapau: Sure! As discussed offline, feel free to open a PR based on this branch (ideally that change should be separate from this PR) :)

frontend/catalyst/passes/builtin_passes.py
dime10: Definitely, but it is a regression being introduced here.

kipawaa (approved):
Thanks!

frontend/catalyst/passes/builtin_passes.py
albi3ro: How? We can now identify them as transforms and integrate them with the ecosystem.

doc/releases/changelog-dev.md
```diff
@@ -639,6 +640,11 @@
 
 <h3>Internal changes ⚙️</h3>
 
+* Refactored all passes in `catalyst.passes.builtin_passes.py` to be `pennylane.transforms.core.Transform` objects
+  rather than decorators. This allows them to be used as standard transforms, enabling full compatibility with
+  `pennylane.CompilePipeline`.
+  [(#2722)](https://github.com/PennyLaneAI/catalyst/pull/2722)
```
isaacdevlugt: I think we can move this to the bug fixes section 🤔

isaacdevlugt (approved):
As far as I'm concerned, the bug is fixed ✅ and the docs still display properly (built locally) ✅. I do agree with @dime10's gripe with how parts of our infra hinder the experience in an IDE... I'll start a conversation about this internally 😄. Sounds like this is worth resolving.

doc/releases/changelog-dev.md
andrijapau: Done!

albi3ro (approved):

frontend/catalyst/passes/builtin_passes.py
andrijapau: Resolving this thread.

PR #2723, opened 2026-04-21
Add constant index folder to qref.get op

paul0403:
**Context:**
Like the `quantum.extract` op, the `qref.get` op can take in static attribute or dynamic SSA value as the extract index.

However I forgot to add the folder for folding dynamic indices from constant-ops into attributes.  

Add a folder on `qref.get` op to convert seemingly dynamic indices from constant-like operations into static attributes.
This is a mirror of [the folder on `quantum.extract`](https://github.com/PennyLaneAI/catalyst/blob/359c1d13c751824db75ad781311abbf569178767/mlir/lib/Quantum/IR/QuantumOps.cpp#L144).

**Benefits:**
When the ref-to-value conversion is judging whether a region needs some reference semantics quantum SSA values from above, it decides single qubits are needed only if the `qref.get` op in the region is indexing with a static attribute.

This folder allows more cases to work with simply taking in qubits from above, instead of entire qregs.

### DIFF
--- a/mlir/test/QRef/CanonicalizationTest.mlir
+++ b/mlir/test/QRef/CanonicalizationTest.mlir
@@ -107,6 +107,21 @@ func.func @test_get_no_user_fold(%r: !qref.reg<3>) {
 
 // -----
 
+// CHECK-LABEL: test_get_constant_index
+func.func @test_get_constant_index(%r: !qref.reg<3>) -> (!qref.bit, !qref.bit) {
+    // CHECK: qref.get %arg0[ 0] : !qref.reg<3> -> !qref.bit
+    %0 = arith.constant 0 : i64
+    %q0 = qref.get %r[%0] : !qref.reg<3>, i64 -> !qref.bit
+
+    // CHECK: qref.get %arg0[ 1] : !qref.reg<3> -> !qref.bit
+    %1 = stablehlo.constant dense<1> : tensor<i64>
+    %extracted = tensor.extract %1[] : tensor<i64>
+    %q1 = qref.get %r[%extracted] : !qref.reg<3>, i64 -> !qref.bit
+    return %q0, %q1 : !qref.bit, !qref.bit
+}
+
+// -----
+
 // CHECK-LABEL: test_get_cse
 func.func @test_get_cse(%r: !qref.reg<3>, %i: i64) {
     // CHECK: [[q0:%.+]] = qref.get %arg0[ 0] : !qref.reg<3> -> !qref.bit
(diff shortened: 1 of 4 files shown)


mlir/test/QRef/CanonicalizationTest.mlir
```diff
@@ -107,6 +107,21 @@ func.func @test_get_no_user_fold(%r: !qref.reg<3>) {
 
 // -----
 
+// CHECK-LABEL: test_get_constant_index
+func.func @test_get_constant_index(%r: !qref.reg<3>) -> (!qref.bit, !qref.bit) {
+    // CHECK: qref.get %arg0[ 0] : !qref.reg<3> -> !qref.bit
```
joeycarter: Worth adding a `CHECK-NOT` here?

```suggestion
    // CHECK-NOT: arith.constant
    // CHECK: qref.get %arg0[ 0] : !qref.reg<3> -> !qref.bit
```

mlir/test/QRef/CanonicalizationTest.mlir
```diff
@@ -107,6 +107,21 @@ func.func @test_get_no_user_fold(%r: !qref.reg<3>) {
...
+    %0 = arith.constant 0 : i64
+    %q0 = qref.get %r[%0] : !qref.reg<3>, i64 -> !qref.bit
+
+    // CHECK: qref.get %arg0[ 1] : !qref.reg<3> -> !qref.bit
```
joeycarter: As above

```suggestion
    // CHECK-NOT: stablehlo.constant
    // CHECK-NOT: tensor.extract
    // CHECK: qref.get %arg0[ 1] : !qref.reg<3> -> !qref.bit
```

joeycarter (commented):
Looks great! Just one quick comment on the tests.

mlir/test/QRef/CanonicalizationTest.mlir
paul0403: Yes! I'll add it

paul0403: https://github.com/PennyLaneAI/catalyst/pull/2723/commits/43032affaaf5f9e2430c3aab00eacf618fc648f4

Also checked the new get op's results are correctly wired up to the return op, because well why not

joeycarter (approved):
💯

mehrdad2m (approved):
The PR looks good, but I noticed this pattern is repeated at least 3 times in different dialects already and maybe more to come with the enhanced register ops. So I am wondering if there is a better solution to avoid this duplication. Maybe by adding a trait/interface for indexing ops like this that we can use everywhere or some other way. I don't want to block you PR for this but maybe a issue page would be nice to track it if you agree.

PR #2731, opened 2026-04-21
Add pattern for lowering qecl.encode

lillian542:
**Context:**
We want to be able to lower `qecl.encode [zero]` for CSS codes based on the X-check measurement pattern, as is done in the [Steane code demo](https://pennylane.ai/qml/demos/tutorial_bp_catalyst)

- Add `x_tanner` and `z_tanner` to the `QecCode` data class fields
- Add a function to the pass for adding the operations for performing X/Z checks based on the tanner graphs of a CSS code
- Add a function to the pass that uses the X-check to perform encoding to zero on a `qecp` codeblock, and replace all `qecl.encode [zero]` ops with a call to this subroutine

**Benefits:**

- We can lower the `qecl.encode` op based on the tanner graph of a CSS code
- The function for checks used for the subroutine for this pattern should also be useful for adding lowering for QEC cycles

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A

[sc-115298]


frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -109,14 +155,155 @@ def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
...
+            visibility="private",
+            region=Region([block]),
+        )
+
+        return funcOp
+
+    def check_pattern(
+        self,
+        aux_qubits: Iterable[qecp.QecPhysicalQubitSSAValue],
+        codeblock: qecp.PhysicalCodeBlockSSAValue,
```
lillian542: I'm not sure if I did this right in terms of type hints

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -109,14 +155,155 @@ def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
...
+            # allocate auxiliary qubits
+            aux_allocate_ops = (qecp.AllocAuxQubitOp() for row in self.qec_code.x_tanner)
+            aux_qubits = [op.results[0] for op in aux_allocate_ops]
+
+            # apply X-check gate+measurement pattern
+            measure_ops, encoded_codeblock = self.check_pattern(
+                aux_qubits, codeblock, check_type=CheckType.X
+            )
+
+            # ToDo: should we also be applying the Z corrections?
```
lillian542: As I understand it (and IIRC from our previous discussions), this would be redundant because we will be following this with a cycle of error correction. But the demo adds corrections here, so just wanted to highlight the difference in case there are any concerns.

frontend/catalyst/python_interface/transforms/qecp/qec_code_lib.py
```diff
@@ -17,8 +17,16 @@
 from dataclasses import dataclass, fields
 from typing import Self
 
+import numpy as np
+
 _CODE_REGISTRY: dict[str, tuple[int, int, int]] = {
```
multiphaseCFD: do we want to update the type annotation as well?

frontend/catalyst/python_interface/transforms/qecp/qec_code_lib.py
```diff
@@ -17,8 +17,16 @@
...
-    "Steane": (7, 1, 3),
+    "Steane": (
+        7,
+        1,
+        3,
+        np.array([[1, 1, 1, 1, 0, 0, 0], [0, 1, 1, 0, 1, 1, 0], [0, 0, 1, 1, 0, 1, 1]]),
+        np.array([[1, 1, 1, 1, 0, 0, 0], [0, 1, 1, 0, 1, 1, 0], [0, 0, 1, 1, 0, 1, 1]]),
```
multiphaseCFD, line 28: Would these two arrays be the single source of initializing the tanner graph? If yes, I believe we need alignment on how to build a tanner graph @joeycarter .

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -76,6 +92,40 @@ def convert_type(self, typ: qecl.LogicalHyperRegisterType) -> qecp.PhysicalHyper
...
 
+# MARK: Encode Op Pattern
+
+
+class EncodeOpConversion(RewritePattern):
+    """Converts qecl.encode [zero] to the equivalent subroutine of qecp gates"""
+
+    def __init__(self, qec_code: QecCode, encode_subroutine: func.FuncOp):
+        self.qec_code = qec_code
+        self.encode_subroutine = encode_subroutine
```
joeycarter: We could make this a `dataclass` to skip the manual `__init__()`. This is common in xDSL, I've learned, e.g. https://github.com/xdslproject/xdsl/blob/main/xdsl/transforms/convert_ptr_to_llvm.py

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
multiphaseCFD: > As I understand it (and IIRC from our previous discussions), this would be redundant because we will be following this with a cycle of error correction.

I agree. Compile time noiseops are injected right before `qecp.qec` operations.

frontend/catalyst/python_interface/transforms/qecp/qec_code_lib.py
multiphaseCFD: I thought we would use the arrays here to build a tanner graph operations. Looks like tanner graph operation injection is a separate thing

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
```diff
@@ -117,12 +133,148 @@ def test_hyperreg_conversion_with_k_mismatch(self, run_filecheck):
...
+        two auxiliary qubits (set by the number of rows in the z_tanner graph)"""
+
+        n = 2
+        qec_code = QecCode(code_name, n=n, k=k, d=1, x_tanner=np.eye(n), z_tanner=np.eye(n))
+
+        program = f"""
+        builtin.module @module_circuit {{
+                func.func @test_func() attributes {{quantum.node}} {{
+                    // CHECK: [[codeblock:%.+]] = "test.op"() : () -> !qecp.codeblock<{k} x {n}>
+                    // CHECK-NEXT: [[codeblock2:%.+]] = func.call @encode_zero_{code_name}([[codeblock]]) : (!qecp.codeblock<{k} x {n}>) -> !qecp.codeblock<{k} x {n}>
```
multiphaseCFD, line 163: Maybe we could add an audit story to update the naming and unit tests conventions. The `qecl.noise` lowering pass hardcoded the program str.

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -119,14 +169,156 @@ def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
...
+    def check_pattern(
+        self,
+        aux_qubits: Iterable[qecp.QecPhysicalQubitSSAValue],
+        codeblock: qecp.PhysicalCodeBlockSSAValue,
+        check_type: CheckType,
+    ) -> (Iterable[qecp.MeasureOp], qecp.PhysicalCodeblockType):
+        """Contains the ops to perform a QEC check on the provided auxiliary qubits and codeblock.
+        Intended to be called inside `builder.ImplicitBuilder` to add these operations to a block.
+
+        This pattern includes measurement of the auxiliary qubits, and returns the MeasureOps, as
```
multiphaseCFD: I would suggest to add a reference to the implementations here since there are many difference approaches to do zero encoding.

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -119,14 +169,156 @@ def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
...
+        )
+
+        return funcOp
+
+    def check_pattern(
+        self,
+        aux_qubits: Iterable[qecp.QecPhysicalQubitSSAValue],
+        codeblock: qecp.PhysicalCodeBlockSSAValue,
+        check_type: CheckType,
+    ) -> (Iterable[qecp.MeasureOp], qecp.PhysicalCodeblockType):
```
joeycarter: But the return type hint should be:

```suggestion
    ) -> tuple[Iterable[qecp.MeasureOp], qecp.PhysicalCodeBlockSSAValue]:
```

multiphaseCFD (commented):
Nice work! Thanks @lillian542 . Just a few Qs.

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
joeycarter: The input type hints look right to me.

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -119,14 +169,156 @@ def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
...
+            # re-set auxiliary qubits
+            hadamard_ops = [qecp.HadamardOp(aux) for aux in aux_qbs_out]
+            aux_qbs_out = [h_op.results[0] for h_op in hadamard_ops]
+
+        # measure auxilary qubits
+        measure_ops = [qecp.MeasureOp(qb) for qb in aux_qbs_out]
+
+        return measure_ops, codeblock
+
+    def _get_cnot_and_tanner_graph(self, check_type: CheckType) -> (np.ndarray, Callable):
```
joeycarter:
```suggestion
    def _get_cnot_and_tanner_graph(self, check_type: CheckType) -> tuple[np.ndarray, Callable]:
```

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -119,14 +169,156 @@ def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
...
+        """Get the appropriate tanner graph and the function for applying CNOTs in an
+        QEC check based on the check type."""
+
+        if check_type == "X":
```
joeycarter: Might as well use the enum here:

```suggestion
        if check_type == CheckType.X:
```

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -119,14 +169,156 @@ def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
...
+            # we use CNOT(aux, data) and X-tanner graph
+            tanner_graph = self.qec_code.x_tanner
+
+            def cnot_fn(aux_qb, data_qb):
+                cnot_op = qecp.CnotOp(aux_qb, data_qb)
+                aux_qb, data_qb = cnot_op.results
+                return aux_qb, data_qb
+
+        elif check_type == "Z":
```
joeycarter: and here:

```suggestion
        elif check_type == CheckType.Z:
```

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -119,14 +169,156 @@ def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
...
+            tanner_graph = self.qec_code.z_tanner
+
+            def cnot_fn(aux_qb, data_qb):
+                cnot_op = qecp.CnotOp(data_qb, aux_qb)
+                data_qb, aux_qb = cnot_op.results
+                return aux_qb, data_qb
+
+        else:
+            raise CompileError(
+                f"Only CSS codes are supported, check_type must be X or Z but recieved {check_type}"
```
joeycarter:
```suggestion
                f"Only CSS codes are supported, check_type must be X or Z but received {check_type}"
```

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -119,14 +169,156 @@ def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
...
+            check_type (CheckType): Which check pattern will be performed.
+
+        Returns:
+            Iterable[qecp.MeasureOp]: a list of the ops measuring the auxiliary qubits
+            qecp.PhysicalCodeblockType: the codeblock after the check pattern has been applied
+        """
+
+        tanner_graph, cnot_fn = self._get_cnot_and_tanner_graph(check_type)
+
+        if check_type == "X":
```
joeycarter:
```suggestion
        if check_type == CheckType.X:
```

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -119,14 +169,156 @@ def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
...
+        for i in range(self.qec_code.n):
+            insert_op = qecp.InsertQubitOp(codeblock, i, data_qbs_out[i])
+            codeblock = insert_op.results[0]
+
+        if check_type == "X":
+            # re-set auxiliary qubits
+            hadamard_ops = [qecp.HadamardOp(aux) for aux in aux_qbs_out]
+            aux_qbs_out = [h_op.results[0] for h_op in hadamard_ops]
+
+        # measure auxilary qubits
```
joeycarter:
```suggestion
        # measure auxiliary qubits
```

frontend/catalyst/python_interface/transforms/qecp/qec_code_lib.py
joeycarter: Since the dict no longer contains a tuple of three ints :)

```suggestion
_CODE_REGISTRY: dict[str, tuple[Any, ...]] = {
```

frontend/catalyst/python_interface/transforms/qecp/qec_code_lib.py
joeycarter, line 28: So the plan is to start with human-readable parity check matrices and then eventually migrate to a sparse-matrix representation of the Tanner graph?

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
```diff
@@ -117,12 +133,148 @@ def test_hyperreg_conversion_with_k_mismatch(self, run_filecheck):
...
+        )
 
         with pytest.raises(CompileError, match="Failed to convert type"):
             run_filecheck(program, pipeline)
 
 
+class TestLoweringEncode:
+    """Test lowering the qecl.EncodeOp to a subroutine of qecp gates"""
+
+    @pytest.mark.parametrize("code_name", ["test_name", "abcd"])
```
joeycarter, line 147: Mind if I ask what the purpose of parameterizing the `code_name` variable is? Wouldn't the test be essentially the same for both cases, just with a different subroutine name?

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
joeycarter: I was wondering this as well. I think this is fine for now, and is in line with what we proposed originally. In the _Quantum Error Correction for Beginners_ paper ([arXiv: 0905.2794](https://arxiv.org/abs/0905.2794)), Fig. 8, the authors also include the "half" cycle of QEC  to apply Z corrections given the X checks that were just performed, but if encoding is immediately following by a full cycle of QEC then I think it's safe to omit the half cycle here.

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -119,14 +169,156 @@ def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
...
+            Iterable[qecp.MeasureOp]: a list of the ops measuring the auxiliary qubits
+            qecp.PhysicalCodeblockType: the codeblock after the check pattern has been applied
+        """
+
+        tanner_graph, cnot_fn = self._get_cnot_and_tanner_graph(check_type)
+
+        if check_type == "X":
+            # auxiliary qubits are prepared in the |+> state
+            hadamard_ops = [qecp.HadamardOp(qb) for qb in aux_qubits]
+            aux_qubits = [h_op.results[0] for h_op in hadamard_ops]
```
joeycarter, line 269: [Optional] This line modifies the input `aux_qubits` argument. Probably not an issue, but might be safer to use a new variable and treat the input `aux_qubits` as constant.

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -119,14 +169,156 @@ def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
...
+                if val:
+                    aux_qb, data_qb = cnot_fn(aux_qb, data_qb)
+                data_qbs_out.append(data_qb)
+            data_qubits = data_qbs_out
+            aux_qbs_out.append(aux_qb)
+
+        # insert data qubits back into the codeblock
+        for i in range(self.qec_code.n):
+            insert_op = qecp.InsertQubitOp(codeblock, i, data_qbs_out[i])
+            codeblock = insert_op.results[0]
```
joeycarter, line 290: [Optional] Same here: this line modifies the input `codeblock` argument.

joeycarter (commented):
Looks great, thanks @lillian542! Just a few, mostly minor comments and questions to start.

frontend/catalyst/python_interface/transforms/qecp/qec_code_lib.py
joeycarter: It's a good question. I think having the Tanner graphs in this format is a good starting point for the Steane code since it's more human-readable and easier to reason about. But, when we start inserting Tanner graph _types_ into the IR, we'll need a method to convert these dense parity-check matrices to the sparse matrix representation that the `qecp.assemble_tanner` op expects.

The other option is to use the sparse matrix representation as the source of truth and write utility functions to make it easier to iterate over them in the body of the pass.

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
lillian542: Good point!

frontend/catalyst/python_interface/transforms/qecp/qec_code_lib.py
lillian542: Updated!

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
lillian542: Sure - maybe I can add it as a requirement to the over all "A pass to lower qec [...] is implemented" story? It's basically our 'wrapping up' story at this point.

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
lillian542: updated :)

frontend/catalyst/python_interface/transforms/qecp/qec_code_lib.py
lillian542: That's what I was thinking - this is easier to input based on a paper, and easier to handle when iterating over it in Python, and we can convert it to the more efficient format when we want to insert it in the IR

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
lillian542: updated :)

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
lillian542: updated :)

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
lillian542: I added a brief note and a reference

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
lillian542: I've added a reference to the QEC for Beginners paper for context on what implementation we are using for encoding, and a note about this being omitted.

lillian542: (and I removed this comment)

frontend/catalyst/python_interface/transforms/qecp/qec_code_lib.py
lillian542: > Would these two arrays be the single source of initializing the tanner graph?

Yes, my thinking was that, since this is the UI for adding new QEC codes, this would be the source of truth for initializing, and we would go with a human-friendly format that more closely follows what is used in textbooks and papers. It will be easier to read, and less error-prone when we need to enter data for a new code. 

Then (as Joey already said), we'll need to use a helper function that converts it to the sparse representation to assemble it for the IR.

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
lillian542: Yes, all it really does is test that it pulls the name from the `QecCode` when we define one, instead of being hard-coded.

In my head it's just like the parameterization for `k` - there's information in the `QecCode` that goes into the IR, and this is confirming that the information from a generic `QecCode` ends up in the IR where its expected.

frontend/catalyst/python_interface/transforms/qecp/qec_code_lib.py
joeycarter: Sounds good 👍

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
joeycarter: Sounds good 👍

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
joeycarter: Looks good, I think this is the best approach for now.

joeycarter (approved):
Nice work! 💯

multiphaseCFD (approved):
Great work!

PR #2733, opened 2026-04-22
Clean some unused leftover headers in pipelines.cpp

paul0403:
**Context:**
#2259 removed the pipeline definition from `mlir/lib/Driver/pipelines.cpp`, to have a single source of truth pipeline registry :partying_face:

There are some left over dialect and pass headers that are no longer needed.

Delete unused headers.

**Benefits:**
Cleaner code.


dime10 (approved):
Thanks!

mehrdad2m (approved):

PR #2747, opened 2026-04-23
`qml` -> `qp` in `utils`

kipawaa:
**Context:**
In order to reposition PennyLane and Catalyst for general quantum computing (rather than quantum machine learning) PennyLane will be imported as `qp` (rather than `qml`). These changes will be made per-module in Catalyst.

Update references to PennyLane in `utils`.

**Benefits:**
Improve user perception of the uses of PennyLane + Catalyst.

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-117380]


kipawaa:
There is only one reference to `qml` in this module, which is in an exception message. Since we'll be handling exceptions and associated tests in a separate PR, I'll close this PR.

PR #2749, opened 2026-04-24
Update pl-compat-workflow for precompiled decomp rules

kipawaa:
**Context:**
The introduction of precompiled decomp rules for the `graph-decomposition` pass required a change in build order for most scripts (since `make frontend` now requires a complete build of catalyst).

Updates the build order in `check-pl-compat.yaml` for LLL testing. The relevant workflow run is available [here](https://github.com/PennyLaneAI/catalyst/actions/runs/24900906312). NOTE: This was run with patched tests that match pennylane exceptions which have been updated from `qml` to `qp`. These changes have since been reverted in this PR to match Catalyst main ([fe24754](https://github.com/PennyLaneAI/catalyst/pull/2749/commits/fe247544964d1e864743550cc5240dda91b031fc))

**Benefits:**
LLL is tested correctly.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A

### DIFF
--- a/.github/workflows/check-pl-compat.yaml
+++ b/.github/workflows/check-pl-compat.yaml
@@ -132,12 +132,12 @@ jobs:
         DIALECTS_BUILD_DIR="$(pwd)/quantum-build" \
         ENABLE_OPENQASM=ON \
         make runtime dialects plugin
-        echo "PYTHONPATH=$PYTHONPATH:$(pwd)/quantum-build/python_packages/quantum" >> $GITHUB_ENV
-        echo "RUNTIME_LIB_DIR=$(pwd)/runtime-build/lib" >> $GITHUB_ENV
-        echo "MLIR_LIB_DIR=$(pwd)/llvm-build/lib" >> $GITHUB_ENV
-        echo "CATALYST_BIN_DIR=$(pwd)/quantum-build/bin" >> $GITHUB_ENV
-        echo "CATALYST_LIB_DIR=$(pwd)/quantum-build/lib" >> $GITHUB_ENV
-        chmod +x $(pwd)/quantum-build/bin/catalyst  # artifact upload does not preserve permissions
+        PYTHONPATH=$PYTHONPATH:$(pwd)/quantum-build/python_packages/quantum \
+        RUNTIME_LIB_DIR=$(pwd)/runtime-build/lib \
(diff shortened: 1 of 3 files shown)


.github/workflows/check-pl-compat.yaml
```diff
@@ -131,7 +131,15 @@ jobs:
         ENZYME_BUILD_DIR="$(pwd)/enzyme-build" \
         DIALECTS_BUILD_DIR="$(pwd)/quantum-build" \
         ENABLE_OPENQASM=ON \
-        make catalyst
+        make runtime dialects plugin
+        echo "PYTHONPATH=$PYTHONPATH:$(pwd)/quantum-build/python_packages/quantum" >> $GITHUB_ENV
```
kipawaa: Why do we use these paths (that differ from where the frontend typically searches)? This requires the `make catalyst` command be split so that we can `make dialects` and then add update the path before `make frontend`.

kipawaa:
I think that this is the only other script that we hadn't updated, outside of the `check-jax-release` workflow. That workflow is currently disabled, should I update it as well or is it out of commission?
EDIT: offline discussion determined it doesn't need to be updated.

.github/workflows/check-pl-compat.yaml
```diff
@@ -131,7 +140,11 @@ jobs:
         ENZYME_BUILD_DIR="$(pwd)/enzyme-build" \
         DIALECTS_BUILD_DIR="$(pwd)/quantum-build" \
         ENABLE_OPENQASM=ON \
-        make catalyst
+        make runtime dialects plugin
+        
+        chmod +x $(pwd)/quantum-build/bin/catalyst  # artifact upload does not preserve permissions
```
dime10: Thanks River! Minor clarification, is `catalyst` coming from the make command or an artifact upload?
- if the former, the comment doesn't really make sense, and you'd think it would already be executable anyway
- if the latter, we can run chmod first without needing the split

kipawaa: [Seems to pass](https://github.com/PennyLaneAI/catalyst/actions/runs/24900906312) with `make catalyst`, good catch!

dime10 (approved):
Easy peasy 🎉

mehrdad2m (approved):

PR #2752, opened 2026-04-24
Remove include<format> in the compiler driver

maliasadi:
**Context:**
Removing `include<format>` from the core compiler is to ensure maximum portability and ABI stability. While `std::format` is a modern C++20 convenience, it often acts as a "gateway to linker hell" in systems like Ubuntu 22.04; its absence in older standard libs forces developers to either switch to `libc++`. As it is used in one single place, this PR rewrites the use case with C++14 compatible impl avoiding the "Undefined Symbol" and "Missing header" errors caused by using older ABI or C++ compilers.

**Related GitHub Issues:**
[sc-117824]

### DIFF
--- a/mlir/include/Driver/DefaultPipelines/DefaultPipelines.h
+++ b/mlir/include/Driver/DefaultPipelines/DefaultPipelines.h
@@ -14,7 +14,6 @@
 
 #pragma once
 
-#include <format>
 #include <numeric>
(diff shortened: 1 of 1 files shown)


kipawaa (approved):
seems great, thanks!

paul0403 (approved):

PR #2753, opened 2026-04-24
Split lowering rules out of jax_primitives.py

paul0403:
**Context:**
The `jax_primitives.py` file was getting too heavy.

Split the lowering rules out into their own file, `primitive_lowering_rules.py`.
Remove all `NotImplementedError()` on the primitives' unused `def_impl`.
Slightly improve import structure.
Delete a unused function `catalyst.Pass.get_options()`. 

**Benefits:**
Better organization.

### DIFF
--- a/frontend/catalyst/jax_extras/lowering.py
+++ b/frontend/catalyst/jax_extras/lowering.py
@@ -126,7 +126,7 @@ def custom_lower_jaxpr_to_module(
     # Create a keepalives list that will be mutated during the lowering.
     keepalives = []
     host_callbacks = []
-    custom_lowering_rules = get_custom_lowering_rules()
+    custom_lowering_rules = tuple(get_custom_lowering_rules().items())
     lowering_params = LoweringParameters(override_lowering_rules=custom_lowering_rules)
     ctx = ModuleContext(
         backend=None,
--- a/frontend/catalyst/jax_extras/lowering.py
+++ b/frontend/catalyst/jax_extras/lowering.py
@@ -15,9 +15,7 @@
 
 from __future__ import annotations
 
-import dataclasses
 import logging
-import textwrap
 
 import jax
 from jax._src import core
@@ -39,6 +37,7 @@ from jaxlib.mlir.dialects.func import FuncOp
 
 import catalyst
 from catalyst.logging import debug_logger
+from catalyst.primitive_lowering_rules import get_custom_lowering_rules
 from catalyst.utils.exceptions import CompileError
 from catalyst.utils.patching import Patcher
 
@@ -129,7 +128,7 @@ def custom_lower_jaxpr_to_module(
     # Create a keepalives list that will be mutated during the lowering.
     keepalives = []
     host_callbacks = []
-    custom_lowering_rules = catalyst.jax_primitives.CUSTOM_LOWERING_RULES
+    custom_lowering_rules = get_custom_lowering_rules()
     lowering_params = LoweringParameters(override_lowering_rules=custom_lowering_rules)
     ctx = ModuleContext(
         backend=None,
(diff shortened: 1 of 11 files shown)


frontend/catalyst/jax_extras/lowering.py
```diff
@@ -187,56 +184,3 @@ def custom_lower_jaxpr_to_module(
                 worklist += [*op.body.operations]
 
     return ctx.module, ctx.context
-
-
-def get_mlir_attribute_from_pyval(value):
```
albi3ro, line 192: unused?

paul0403: Moved to a more appropriate place (`jax_primitives_utils.py`).

I wrote this helper function a while ago, and placed it here in `lowering.py` somewhat arbitrarily. This `lowering.py` file needs the `CUSTOM_LOWERING_RULES` registry, so I was moving it to break out of circular imports.

frontend/catalyst/primitive_lowering_rules.py
```diff
@@ -0,0 +1,2013 @@
...
+from pennylane.capture.primitives import while_loop_prim as pl_while_loop_prim
+
+from catalyst.compiler import get_lib_path
+from catalyst.utils.extra_bindings import FromElementsOp, TensorExtractOp
+from catalyst.utils.types import convert_shaped_arrays_to_tensors
+
+# pylint: disable=unused-argument,too-many-lines,too-many-statements,protected-access
+
+
+CUSTOM_LOWERING_RULES = ()
```
albi3ro: What would you think of having this be a dictionary and then just converting it via `tuple(CUSTOM_LOWERING_RULES.items())` when needed?  It would make it so we easily fetch the corresponding lowering rule when we need it for a given primitive.

paul0403: Ah! Good point, I kept it as a tuple because the connection with jax, `jax.interpreters.mlir.LoweringParameters(override_lowering_rules=CUSTOM_LOWERING_RULES)`

But it completely didn't occur to me to just convert :sweat_smile: I'll add it

paul0403: https://github.com/PennyLaneAI/catalyst/pull/2753/commits/464e8eeca8fb38f40dba294ef3d8350ffc9a147f

paul0403: Could also get rid of the need for `global` keyword since the registry is mutable now.

dime10 (commented):
Thanks @paul0403! This is actually a great opportunity to move these files into the `tracing` module, which was intended to be sub-module for everything jax/tracing related.

The root-level files stem from when we first started the repo and didn't have any subdirectories. After a while we refactored things into proper sub-modules like api-extensions, passes, autograph, etc., but didn't get to refactor the jax code. As a result, it's split across a growing number of files/folders:
- jax_tracer/primitives/primitives_utils.py
- jax_extras (for patches/functionality considered "core jax")
- the started but never completed tracing module
- and maybe even some things in utils/ ?
 
If we're able move all of that into one sub-module that would be amazing in terms of code org 😍
As usual, a general reminder to avoid "utils"-style files and folders, in favour of meaningfully grouped code by purpose/functionality.

paul0403:
> Thanks @paul0403! This is actually a great opportunity to move these files into the `tracing` module, which was intended to be sub-module for everything jax/tracing related.
> 
> The root-level files stem from when we first started the repo and didn't have any subdirectories. After a while we refactored things into proper sub-modules like api-extensions, passes, autograph, etc., but didn't get to refactor the jax code. As a result, it's split across a growing number of files/folders:
> 
> * jax_tracer/primitives/primitives_utils.py
> * jax_extras (for patches/functionality considered "core jax")
> * the started but never completed tracing module
> * and maybe even some things in utils/ ?
> 
> If we're able move all of that into one sub-module that would be amazing in terms of code org 😍 As usual, a general reminder to avoid "utils"-style files and folders, in favour of meaningfully grouped code by purpose/functionality.

Ah, nice! Them being at root-level and not having their own subdirectory (and hence not having their own `__init__.py`) was actually causing a lot of troubles for me and @kipawaa , it caused a bunch of circular imports. A reorganization would certainly be most wonderful!

I'll see what I can come up with 👍

paul0403:
Note that I can do the file reorganization after finishing reference semantics. In the meantime, this PR (which just splits `jax_primitives.py`) can probably go in first.

kipawaa (approved):
This is so great, thanks for doing this!

PR #2755, opened 2026-04-24
Assemble Ops (nWires, nParams) in ResourceAnalysis and RegisterDecompRuleResource passes

maliasadi:
**Context:**
Include Ops (nWires, nParams) in `ResourceAnalysis` and `RegisterDecompRuleResource` passes. We primarily need this info for the graph-decomposition pass; GraphBuilder & Solver use this info to find the most "optimized" decomposition pathway. 

**Benefits:**
- number of params information is available along with number of wires and adjoint for Ops visited by `ResourceAnalysis` and `RegisterDecompRuleResource` passes

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-117856]


mlir/lib/Quantum/Transforms/graph_decomposition.cpp
```diff
@@ -288,19 +288,59 @@ struct GraphDecompositionPass : public impl::GraphDecompositionPassBase<GraphDec
...
+    //     if (name.consume_front("Adjoint(")) {
+    //         node.adjoint = true;
+    //         name = name.split(')').first;
+    //     }
+
+    //     // Extract base name before any parameters
+    //     // (e.g., "GateName" from "Adjoint(GateName)(metadata)")
+    //     node.name = name.split('(').first.trim().str();
+    //     return node;
+    // }
```
sengthai:
```suggestion
```

maliasadi: old commit -- it's removed in the latest

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
```diff
@@ -288,19 +288,44 @@ struct GraphDecompositionPass : public impl::GraphDecompositionPassBase<GraphDec
...
+
+        // Parse "(w,p)" (new) or "(w)" (legacy) suffix.
+        if (raw.consume_front("(") && raw.consume_back(")")) {
+            llvm::StringRef wStr, pStr;
+            std::tie(wStr, pStr) = raw.split(',');
+            int w = -1, p = -1;
+            if (!wStr.getAsInteger(10, w))
+                node.numWires = w;
+            if (!pStr.empty() && !pStr.getAsInteger(10, p))
+                node.numParams = p;
```
sengthai:
```suggestion
            if (!wStr.getAsInteger(10, w)){
                node.numWires = w;
             }
            if (!pStr.empty() && !pStr.getAsInteger(10, p)){
                node.numParams = p;
            }
```

mlir/test/Catalyst/RegisterDecompRuleResourceTest.mlir
```diff
@@ -148,3 +148,33 @@ func.func @rule_with_nested_loop(%arg0: !quantum.bit) -> !quantum.bit attributes
 
     return %q : !quantum.bit
 }
+
+
+// -----
+
+// Rules with parametric ops
+// CHECK:  operations = {"Adjoint(Rot)(4,3)" = 1 : i64, "Rot(4,3)" = 1 : i64}}
```
sengthai, line 156: nit: here we phase it into string that storing op type, wires, and params. that is easy for human-read and frontend, but it would be little challenge to pass to phase or write.  it would be nice if we have any structural way to manipulate it easily. However, this version is good enough.

sengthai (approved):
Look good to me! Thanks @maliasadi 👍

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
```diff
@@ -288,19 +288,46 @@ struct GraphDecompositionPass : public impl::GraphDecompositionPassBase<GraphDec
...
     {
         OperatorNode node;
 
-        // Check for adjoint pattern
-        if (name.consume_front("Adjoint(")) {
+        // Unwrap "Adjoint(GateName)"
+        if (raw.consume_front("Adjoint(")) {
             node.adjoint = true;
-            name = name.split(')').first;
+            auto closeIdx = raw.find(')');
```
kipawaa:
```suggestion
            auto closeIdx = raw.rfind(')');
```
This searches for the last occurrence rather than the first, to ensure it's always the matching bracket (in the case that we have nested operations)

kipawaa: I see that the rest of this isn't set up to support nested operations either, should we implement that now or leave that for a follow-up PR when it's needed?

mlir/test/Catalyst/RegisterDecompRuleResourceTest.mlir
maliasadi: Agreed -- happy to update this as we continue the integration with specs and graph.

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
maliasadi: chatted offline -- there are missing features from the capture mechanism (PLxPR) if we want t support nested ops. we'll get back to this later after supporting `qp.ctrl` and other symbolic ops PL -> PLxPR -> MLIR lowering from the frontend.

kipawaa (approved):
💯

PR #2757, opened 2026-04-25
[qref 3.7] xdsl tools convert `QJIT.mlir` to value semantics

paul0403:
**Context:**
Catalyst's xdsl tools all assume their input is in value semantics, but the frontend produces reference semantics now.

Call the `--convert-to-value-semantics` pass at the beginning of various xdsl tools.

**Benefits:**
Things work.

**Possible Drawbacks:**
When we have time, we should add the actual reference semantics ops to xdsl, so xdsl tools can parse `qref` mlir outputted from the capture frontend pipeline directly. But for now, calling the semantics conversion pass is good enough.

[sc-118243]


frontend/catalyst/python_interface/inspection/xdsl_conversion.py
```diff
@@ -77,21 +79,49 @@ def dont_do_anything():
...
     with Patcher((workflow, "jaxpr", jaxpr)):
         mlir_module = workflow.generate_ir()
 
-    return mlir_module
+    value_semantics_mlir = _quantum_opt_stderr(
+        '--catalyst-pipeline="pipe(canonicalize;convert-to-value-semantics;canonicalize)"',
+        stdin=str(mlir_module),
+    )
+
+    return value_semantics_mlir
```
mudit2812: This function is supposed to return an MLIR module, not an MLIR string

frontend/test/pytest/python_interface/transforms/qecl/test_xdsl_inject_noise_to_qecl.py
```diff
@@ -75,9 +75,9 @@ def circuit():
             # CHECK: qecl.noise
             # CHECK: qecl.qec
             # CHECK: qecl.measure {{%.+}}[0]
+            # CHECK: qecl.insert_block
             # CHECK: quantum.mcmobs
             # CHECK: quantum.sample
-            # CHECK: qecl.insert_block
             # CHECK: qecl.dealloc
```
mudit2812, line 81: Does this impact the correctness of the circuit?

paul0403: No, the DAG in the IR is the same. You can see that in all instances where this change happens, the observable is always an mcm, instead of the qubit values.

frontend/catalyst/python_interface/inspection/xdsl_conversion.py
paul0403: Ah true, do you know where I can find a sample usage of xdsl's parser?

mudit2812: MLIR module, not xDSL module

paul0403: https://github.com/PennyLaneAI/catalyst/pull/2757/commits/57da3d2b12cd850b353c8aa505940c57afadb43e

paul0403: and https://github.com/PennyLaneAI/catalyst/pull/2757/commits/2cd8ef4401ed1c125455a898dd12bc8f06a0c49e

mudit2812 (approved):

PR #2758, opened 2026-04-25
[qref 2.6] Reference semantics MBQC dialect is defined, and convertable to regular MBQC dialect

paul0403:
**Context:**
Create the reference semantics mirror for MBQC dialect, and define conversion rules to regular value semantics MBQC dialect.

**Benefits:**
Unblocks the migration of the entire capture frontend to reference semantics (due to the `measurement_in_basis` primitive).

**Possible Drawbacks:**
None.

[sc-117892]


mlir/include/MBQC/IR/MBQCAttrDefs.h
```diff
@@ -0,0 +1,21 @@
+// Copyright 2026 Xanadu Quantum Technologies Inc.
```
paul0403, line 1: changelog

paul0403: (and shortcut story)

mlir/lib/QRef/IR/QRefOps.cpp
```diff
@@ -282,10 +283,32 @@ LogicalResult ComputationalBasisOp::verify()
...
+    QuregType qregType = getQreg().getType();
+    if (qregType.isDynamic()) {
+        return emitOpError() << "expected static allocation size";
+    }
+
+    size_t qregSize = qregType.getSize().getInt();
+    size_t expectedAdjMatrixSize = qregSize * (qregSize - 1) / 2;
+    if (adjMatrixSize != expectedAdjMatrixSize) {
+        return emitOpError() << "mismatch between allocation size and size of dense adjacency "
+                                "matrix. For an allocation size of "
```
joeycarter: [Minor] We've been calling this format a "densely packed adjacency matrix" to distinguish it from the more familiar "dense matrix".

```suggestion
        return emitOpError() << "mismatch between allocation size and size of densely packed "
                                "adjacency matrix. For an allocation size of "
```

joeycarter (approved):
Looks great, thanks @paul0403! I only had one minor comment on the text of an error message.

mlir/lib/QRef/IR/QRefOps.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2758/commits/2dad6d96494b151b08b2c9936847ca214059a165

mehrdad2m (approved):

PR #2761, opened 2026-04-27
Fix precompile flaky tests

kipawaa:
**Context:**
precompiled decomposition rules tests were [failing flaky tests](https://github.com/PennyLaneAI/catalyst/actions/runs/24943656658/job/73041227432) because the decomposition rules were being added to the graph multiple times, causing warnings other than those expected.

Ensures the tests and their iterations are independent via the [`local_decomps` context manager from PL](https://github.com/PennyLaneAI/pennylane/blob/00c2ffd6437a2d6c50af299c2d7d924dd9328fe2/tests/workflow/interfaces/execute/test_autograd.py#L580).

**Benefits:**
Improved pytest infrastructure, [fixes precompile decomposition rules flaky failures](https://github.com/PennyLaneAI/catalyst/actions/runs/25011556236/job/73248342930).

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/frontend/test/conftest.py
+++ b/frontend/test/conftest.py
@@ -35,6 +35,7 @@ def skip_flaky_tests(config, items):
         if "skip_flaky" in item.keywords:
             item.add_marker(pytest.mark.skip(reason="don't test flakiness"))
 
+
 def is_cuda_available():
     """Checks if cuda is available by trying an import.
 
--- a/frontend/test/conftest.py
+++ b/frontend/test/conftest.py
@@ -26,16 +26,6 @@ import platform
 import pytest
 
 
-def skip_flaky_tests(config, items):
-    """Skip marked tests when running with `--force-flaky`."""
-    if not config.getoption("--force-flaky"):
-        return
-
-    for item in items:
-        if "skip_flaky" in item.keywords:
-            item.add_marker(pytest.mark.skip(reason="don't test flakiness"))
-
-
 def is_cuda_available():
     """Checks if cuda is available by trying an import.
 
@@ -152,7 +142,6 @@ def pytest_collection_modifyitems(config, items):
     """A pytest items modifier method"""
 
     skip_cuda_tests(config, items)
-    skip_flaky_tests(config, items)
 
     # skip braket tests
     skipper = pytest.mark.skip()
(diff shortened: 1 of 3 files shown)


dime10:
Is the flaky testing itself that was causing the rules to be created multiple times, or did it just reveal that this was possible?

kipawaa:
> Is the flaky testing itself that was causing the rules to be created multiple times, or did it just reveal that this was possible?

@dime10 The PL graph is global and hence persistent across multiple runs of the test, so the issue is the flaky testing itself. I discussed with @paul0403 and we agreed that these tests likely don't need to be tested for flakiness. A part of this decision was also that the only way to remove rules from the graph is to use PL internals (see [PL tests](https://github.com/PennyLaneAI/pennylane/blob/00c2ffd6437a2d6c50af299c2d7d924dd9328fe2/tests/decomposition/test_decomposition_rule.py#L283)).

dime10:
> @dime10 The PL graph is global and hence persistent across multiple runs of the test, so the issue is the flaky testing itself. I discussed with @paul0403 and we agreed that these tests likely don't need to be tested for flakiness. A part of this decision was also that the only way to remove rules from the graph is to use PL internals (see [PL tests](https://github.com/PennyLaneAI/pennylane/blob/00c2ffd6437a2d6c50af299c2d7d924dd9328fe2/tests/decomposition/test_decomposition_rule.py#L283)).

I see, so the issue is that we're adding the same op to the graph database multiple times, which is not our intention. I think it is okay to avoid flaky testing then.

However, separate from the flaky testing is the issue of global state changes. That is a big concern in tests because it breaks the independence assumption! Tests can affect each other and this effect can be sensitive to test ordering, or whether a given test fails or succeeds, causing difficult to reproduce issues. We've dealt with this a couple of times already when global state was involved (e.g. capture on/off), so I would say it's imperative to figure out a way to test these features without mutating (or at least restoring) global state.

kipawaa:
> However, separate from the flaky testing is the issue of global state changes.

Thanks @dime10 that's a great point, I'll look into how PL manages the global state and apply a better solution!

paul0403 (approved):
Nice that they have a fixture we can use directly!

joeycarter (approved):
🟢

PR #2762, opened 2026-04-27
Bump PL version to 0.45.0.dev89

paul0403:
**Context:**
Bump PL version for something needed in the reference semantics work (https://github.com/PennyLaneAI/pennylane/pull/9336)

I wanted to separate this bump so that any potential issues can be dealt with separately, instead of being squashed into the already big reference semantics PR.


JerryChen97 (approved):
Thanks for fixing!!!!

kipawaa (approved):
Thanks!

PR #2768, opened 2026-04-28
Add bufferization interfaces to qecp ops

joeycarter:
**Context:** Catalyst uses MLIR's built-in bufferization pipeline to convert operations with `tensor` semantics to `memref` semantics. The heavy lifting of this pipeline is performed by the `-one-shot-bufferize` MLIR pass, which acts on all operations that implement a `BufferizableOpInterface`. For more information, see the MLIR [Bufferization](https://mlir.llvm.org/docs/Bufferization/) documentation.

All `qecp` ops that have `tensor` operands and/or results will need to implement a `BufferizableOpInterface`. 

**Description of the Change:** This PRs implements the `BufferizableOpInterface` for the following qecp ops:

- `qecp.assemble_tanner`
- `qecp.decode_esm_css`
- `qecp.decode_physical_meas`

Some TableGen op definitions also needed to be updated to include a new input operand that acts as an output parameter in place of the returned value; the updates were made to both the MLIR and xDSL op definitions.
 
**Benefits:** This is a necessary component of the QEC compilation pipeline for end-to-end execution.

[sc-117446]


mlir/test/QecPhysical/BufferizationTest.mlir
```diff
@@ -0,0 +1,54 @@
...
+    func.return
+}
+
+// -----
+
+// CHECK-LABEL: decode_esm_css
+// CHECK-SAME: [[esm:%.+]]: tensor<3xi1>
+func.func @decode_esm_css(%esm : tensor<3xi1>) {
+    // CHECK: [[tanner:%.+]] = "test.op"() : () -> !qecp.tanner_graph<12, 8, i32>
+    %tanner = "test.op"() : () -> !qecp.tanner_graph<12, 8, i32>
```
paul0403:
```suggestion
func.func @decode_esm_css(%esm : tensor<3xi1>, %tanner: !qecp.tanner_graph<12, 8, i32>) {
```

I usually like doing this in arguments, but up to you, no strong preference

mlir/lib/QecPhysical/Transforms/BufferizableOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,212 @@
...
+#include "QecPhysical/Transforms/BufferizableOpInterfaceImpl.h"
+
+#include "mlir/Dialect/Bufferization/IR/BufferizableOpInterface.h"
+#include "mlir/Dialect/Bufferization/IR/Bufferization.h"
+#include "mlir/Dialect/MemRef/IR/MemRef.h"
+
+#include "QecPhysical/IR/QecPhysicalDialect.h"
+#include "QecPhysical/IR/QecPhysicalOps.h"
+
+#include <mlir/IR/Builders.h>
```
paul0403: Slightly messed up include position : )

mlir/lib/QecPhysical/Transforms/BufferizableOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,212 @@
+// Copyright 2024-2025 Xanadu Quantum Technologies Inc.
```
paul0403:
```suggestion
// Copyright 2026 Xanadu Quantum Technologies Inc.
```

mlir/lib/QecPhysical/Transforms/BufferizableOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,212 @@
...
+        return true;
+    }
+
+    bool bufferizesToMemoryWrite(Operation *op, OpOperand &opOperand,
+                                 const bufferization::AnalysisState &state) const
+    {
+        return false;
+    }
+
+    bool bufferizesToAllocation(Operation *op, Value value) const { return true; }
```
paul0403: Where does `AssembleTannerGraphOp` allocate? It just reads in two tensors/memrefs right?

mlir/test/QecPhysical/BufferizationTest.mlir
```diff
@@ -0,0 +1,54 @@
...
+// CHECK-LABEL: assemble_tanner
+func.func @assemble_tanner() {
+    // CHECK: [[row_idx:%.+]] = memref.get_global {{.*}} : memref<12xi32>
+    // CHECK: [[col_ptr:%.+]] = memref.get_global {{.*}} : memref<8xi32>
+    %row_idx = arith.constant dense<[0, 0, 1, 0, 1, 2, 0, 2, 1, 1, 2, 2]> : tensor<12xi32>
+    %col_ptr = arith.constant dense<[0, 1, 3, 6, 8, 9, 11, 12]> : tensor<8xi32>
+
+    // CHECK: [[tanner:%.+]] = qecp.assemble_tanner [[row_idx]], [[col_ptr]] : memref<12xi32>, memref<8xi32> -> !qecp.tanner_graph<12, 8, i32>
+    %0 = qecp.assemble_tanner %row_idx, %col_ptr : tensor<12xi32>, tensor<8xi32> -> !qecp.tanner_graph<12, 8, i32>
+    func.return
```
paul0403: Can you return the result here (as a use) to check the bufferization's "replace result with result of buffered new op" is correct?

paul0403: Same with the other tests

paul0403: Actually especially true with the other two ops, since their result tensors should be replaced with new memref buffers (which are now operands on the new op!)

paul0403 (commented):
Thanks @joeycarter ! Very straightfoward PR, just a couple small things 💯

mlir/test/QecPhysical/BufferizationTest.mlir
```diff
@@ -0,0 +1,54 @@
...
+
+// CHECK-LABEL: decode_esm_css
+// CHECK-SAME: [[esm:%.+]]: tensor<3xi1>
+func.func @decode_esm_css(%esm : tensor<3xi1>) {
+    // CHECK: [[tanner:%.+]] = "test.op"() : () -> !qecp.tanner_graph<12, 8, i32>
+    %tanner = "test.op"() : () -> !qecp.tanner_graph<12, 8, i32>
+
+    // CHECK-DAG: [[esm_buf:%.+]] = bufferization.to_buffer [[esm]] : tensor<3xi1> to memref<3xi1>
+    // CHECK-DAG: [[idx_buf:%.+]] = memref.alloc() : memref<1xindex>
+    // CHECK: qecp.decode_esm_css([[tanner]] : !qecp.tanner_graph<12, 8, i32>) [[esm_buf]] in([[idx_buf]] : memref<1xindex>) : memref<3xi1>
```
multiphaseCFD: since `tanner_graph` is not bufferize, I'm wondering how I can access the row_idx an col_ptr data could be used to feed the runtime?

mlir/lib/QecPhysical/Transforms/BufferizableOpInterfaceImpl.cpp
joeycarter: Thanks! de7e8cb83

mlir/test/QecPhysical/BufferizationTest.mlir
joeycarter: Fair, that also seems to be the preference in MLIR's [Testing Guidelines](https://mlir.llvm.org/getting_started/TestingGuide/#test-formatting-best-practices). Updated: 9d7efbfa5

mlir/test/QecPhysical/BufferizationTest.mlir
joeycarter: Good point! Done: 9d7efbfa5

mlir/lib/QecPhysical/Transforms/BufferizableOpInterfaceImpl.cpp
joeycarter: Thanks, my editor has a habit of inserting includes for me :upside_down_face: 

fa83535a0

mlir/lib/QecPhysical/Transforms/BufferizableOpInterfaceImpl.cpp
joeycarter: Yes, it only reads in two already-allocated tensors/buffers, it doesn't do the allocation itself. This behaviour is similar to the [`sparse_tensor.assemble`](https://mlir.llvm.org/docs/Dialects/SparseTensorOps/#sparse_tensorassemble-sparse_tensorassembleop) op. The return type is effectively a struct wrapper around these two buffers.

mlir/test/QecPhysical/BufferizationTest.mlir
joeycarter: Discussed offline. @multiphaseCFD anything else to add here or happy to resolve?

mlir/test/QecPhysical/BufferizationTest.mlir
```diff
@@ -0,0 +1,67 @@
...
+// CHECK-LABEL: func.func @assemble_tanner()
+func.func @assemble_tanner() -> !qecp.tanner_graph<12, 8, i32> {
+    // CHECK: [[row_idx:%.+]] = memref.get_global {{.*}} : memref<12xi32>
+    // CHECK: [[col_ptr:%.+]] = memref.get_global {{.*}} : memref<8xi32>
+    %row_idx = arith.constant dense<[0, 0, 1, 0, 1, 2, 0, 2, 1, 1, 2, 2]> : tensor<12xi32>
+    %col_ptr = arith.constant dense<[0, 1, 3, 6, 8, 9, 11, 12]> : tensor<8xi32>
+
+    //      CHECK: [[tanner:%.+]] = qecp.assemble_tanner [[row_idx]], [[col_ptr]] :
+    // CHECK-SAME:   memref<12xi32>, memref<8xi32> -> !qecp.tanner_graph<12, 8, i32>
+    %0 = qecp.assemble_tanner %row_idx, %col_ptr : tensor<12xi32>, tensor<8xi32> -> !qecp.tanner_graph<12, 8, i32>
```
multiphaseCFD, line 26: One last Q: Correct me if I'm wrong. I believe we could access the  `row_idx` and `col_ptr` a `qecp.tanner_graph` SSA value wrapped with the following steps: 1. get `qecp.assemble_tanner` operation with the `getDefiningOp` method; 2. Access the `row_idx` and `col_ptr` from the defining operation?

joeycarter: Yes and no. If we're still in the MLIR layer, e.g. in a compiler pass, and if both `row_idx` and `col_ptr` originate from `arith.constant` ops, then yes, we could get the arrays as array attributes and we could use them however we like _in the compiler pass_. But, the SSA values might also originate from block arguments, whose values wouldn't be known at compile time.

But in the context of bufferization, I don't think this is relevant either way, since bufferization itself doesn't need to know the values in these arrays. At runtime, when these values _do_ matter, we don't have access to the compiler methods like `getDefiningOp` since we're no longer in the MLIR layer.

multiphaseCFD: Thanks Joey! Well explained!

multiphaseCFD (approved):
LGTM! Thanks @joeycarter

mlir/lib/QecPhysical/Transforms/BufferizableOpInterfaceImpl.cpp
paul0403: Yes, so the op itself doesn't allocate right? 

For reference, the `sparse_tensor.assemble` you are quoting has allocation set to false. 
https://github.com/llvm/llvm-project/blob/fc77aa9f72a2928321691b8747da9eab52aa2008/mlir/lib/Dialect/SparseTensor/Transforms/BufferizableOpInterfaceImpl.cpp#L147

mlir/lib/QecPhysical/Transforms/BufferizableOpInterfaceImpl.cpp
```diff
@@ -0,0 +1,205 @@
...
+        bufferization::replaceOpWithBufferizedValues(rewriter, op, newAssembleTannerOp.getResult());
+        return success();
+    }
+};
+
+// Bufferization of qecp.decode_esm_css.
+//   - Convert tensor of ESMs to memref.
+//   - Bufferize result tensor of error indicies with a corresponding memref.alloc; users of the
+//     result tensor are updated to use the new memref.
+struct DecodeEsmCssOpInterface
```
paul0403, line 94: Sorry, why did you remove the `bufferizesToAllocation` methods on the other two ops? 😅

joeycarter: From https://discourse.llvm.org/t/psa-bufferization-new-buffer-deallocation-pipeline/73375#p-302980-migration-guide-1:

> If you own `BufferizableOpInterface` implementations: `bufferizesToAllocation` will be deleted and is no longer necessary, as One-Shot Bufferize no longer deals with deallocations.

mlir/lib/QecPhysical/Transforms/BufferizableOpInterfaceImpl.cpp
joeycarter: Yes, this was a mistake: I've deleted the `bufferizesToAllocation` method altogether. See my comment here: https://github.com/PennyLaneAI/catalyst/pull/2768#discussion_r3169009784

mlir/lib/QecPhysical/Transforms/BufferizableOpInterfaceImpl.cpp
paul0403: Oh, we haven't had time to migrate to the new deallocation yet :sweat_smile: 

https://github.com/PennyLaneAI/catalyst/blob/main/mlir/lib/Catalyst/Transforms/BufferDeallocation.cpp#L24

paul0403: Discussed in person, it's fine to remove the `bufferizesToAllocation` methods. It is only used during one shot bufferization when there are alisaing op operands (https://github.com/llvm/llvm-project/blob/b6aa86a3b18d63ababe57b4012a6854dfd41ad35/mlir/lib/Dialect/Bufferization/Transforms/OneShotAnalysis.cpp#L1062), which our ops do not have.

paul0403 (approved):

PR #2476, opened 2026-02-10
Add a fixture to pass `capture` kwarg to qjit automatically

JerryChen97:
**Context:**
We are about to start the audit of tests. For convenience, a tiny nice fixture to automatically apply the two values of `capture` kwarg locally is good.

**Description of the Change:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-111330]

PR #2771, opened 2026-04-28
Use `qjit(capture=)` locally in tests

JerryChen97:
**Context:**

Tests previously relied on the `use_both_frontend` fixture to toggle `qp.capture` globally, then used `@qjit` with the default `capture="global"` to inherit that state. This couples test behavior to fragile global state and is unsafe under `pytest -n auto` (xdist), where leaked state on one worker poisons subsequent tests.

- `use_both_frontend` → `capture_mode` fixture (parametrizes `True`/`False` without touching global state; tests pass it to `qjit(capture=capture_mode)`)
- Bare `capture.enable()`/`capture.disable()` pairs → `qjit(capture=True)` (locally scoped via `_ensure_capture_mode` with try/finally)
- Removed redundant standalone `capture.disable()` calls that were no-ops after `qjit`'s own state restore
- 40 files changed,,, but lots of unnecessary inline global toggles were eliminated

**Benefits:**

- Capture mode is now controlled locally per-`qjit` call, not via global state
- Eliminates a class of xdist-dependent CI failures caused by capture state leaks
- Safer by construction — no try/finally discipline needed in test code

**Possible Drawbacks:**

- None. All modified tests pass individually and under `-n auto`. Full codebase audit confirms no remaining unguarded toggles.

**Related GitHub Issues:**
[sc-117344]


JerryChen97:
Can't reproduce failures locally. Smells like some dangerous status leak happening in multi-processing
P.S. fixed at https://github.com/PennyLaneAI/catalyst/pull/2771/commits/0a068ff97c2140b3f640e5b92445014bd59f47eb

rniczh (approved):
💯

frontend/test/pytest/python_interface/inspection/test_construct_circuit_dag.py
```diff
@@ -210,12 +209,12 @@ def test_external_empty_function_visualization_error(self):
         with pytest.raises(VisualizationError, match=expected_error):
             utility.construct(module)
 
-    def test_standard_qnode(self):
+    def test_standard_qnode(self, capture_mode):
```
comp-phys-marc, line 212: Where is capture mode coming from? A fixture?

frontend/test/pytest/python_interface/inspection/test_draw_unified_compiler.py
```diff
@@ -561,7 +563,7 @@ def circuit():
             print(draw(circuit)())
 
 
-@pytest.mark.usefixtures("use_both_frontend", "skip_no_graph_deps")
```
comp-phys-marc, line 564: Why did we delete this?

frontend/test/pytest/python_interface/inspection/test_mlir_specs.py
```diff
@@ -76,7 +76,6 @@ def make_static_resources(
     return res
 
 
-@pytest.mark.usefixtures("use_both_frontend")
```
comp-phys-marc, line 79: Same question here? Why do we delete these?

frontend/test/pytest/test_quantum_control.py
```diff
@@ -45,17 +45,17 @@
 
 
 def verify_catalyst_ctrl_against_pennylane(
-    quantum_func: Callable, device, *args, with_adjoint_arg=False
+    quantum_func: Callable, device, *args, with_adjoint_arg=False, capture_mode="global"
```
comp-phys-marc, line 48: shouldn't capture_mode be true or false?

frontend/test/pytest/python_interface/inspection/test_construct_circuit_dag.py
JerryChen97: Yes. They were introduced at https://github.com/PennyLaneAI/catalyst/pull/2476

comp-phys-marc: Nice thanks.

frontend/test/pytest/python_interface/inspection/test_mlir_specs.py
JerryChen97: In Catalyst, `capture` is basically only used by `qjit`; that is also why months ago we introduced the new kwarg `capture=true/false/global` for `qjit`.
However, what `use_both_frontend` does is that it will turn on capture for the entire scope it acts on. This leads to many cases where one does not want to have capture enabled anymore (e.g., in a test you need an inplace DQ execution using tapes) and then has to manually turn it off and on etc. Some cases you will see that people forgot to write a `try finally` for such scenario, which caused quite some mysterious global state leak.
So we were thinking about moving to the pure parametrization fixture `capture_mode` which will automatically assign a new param to the decorated test function to pass the values for `qjit(capture=...)` in, without altering the other parts of any tests.

comp-phys-marc: Ok thanks.

comp-phys-marc: But in the `capture_mode` fixture it seems that it is only `True` or `False`. I don't see the other option "global"

frontend/test/pytest/test_quantum_control.py
JerryChen97: yeah this place is a very special case I forgot to mention: `verify_catalyst_ctrl_against_pennylane` is a testing helper here, and its purpose is to accept whatever input value from external, and pass the value to `qjit` internal. Therefore I set it up as the default value `"global"`, so that if someone happens to use this helper on their own tests or whatever other tasks without specifying a capture value, this overall helper function behaves just like a pure wrapper over `qjit` without unnecessary assumption.
For this file, what is supposed to do is any test should always pass this kwarg in

comp-phys-marc: Thanks for the explanation. So what does `@qjit(capture="global")` do?

frontend/test/pytest/python_interface/inspection/test_mlir_specs.py
JerryChen97: Yeah the fixture is definitely only True or False, because we should not let any test set in the volatile initial state.

frontend/test/pytest/test_quantum_control.py
JerryChen97: A `qjit(capture="global")` will do nothing but accept the "global" state of capture (not really the true "global" as we used to mention in a random script, but in the sense that whoever external to `qjit` that finally decided the value of `capture.enabled`)

comp-phys-marc: ok

comp-phys-marc (approved):
Thanks for answering my questions!

frontend/test/pytest/test_quantum_control.py
JerryChen97: So by design, `qjit(capture=)` is tri-state: global means “*inherit* the current 'global' capture setting”, while true and false are local overrides. In other words, by default qjit passively follows the ambient capture mode, but an explicit capture=True or capture=False forces the behavior for this qjit call only (and also will reset capture.enabled to before at exit).

PR #2773, opened 2026-04-28
Reference semantics `pbc.ref.ppm` op is defined, and convertable to regular `pbc.ppm` op

paul0403:
**Context:**
Define PPM in reference semantics, and add conversion pattern to the value semantics `pbc.ppm` op.

**Benefits:**
Unblocks the reference semantics lowering of `pauli_measure_p` primitive in frontend. 
Technically it's not too much work to define `qref.pbc.*` for all pbc ops, but I want to finish the frontend work first, so I'm only adding the blocker op first.

[sc-117901]


mlir/lib/QRef/Transforms/value_semantics_conversion.cpp
```diff
@@ -1061,6 +1062,23 @@ void handleMeasureInBasis(IRRewriter &builder, qref::MeasureInBasisOp rMeasureIn
     builder.eraseOp(rMeasureInBasisOp);
 }
 
+void handlePPM(IRRewriter &builder, qref::PPMeasurementOp rPPMOp, QubitValueTracker &tracker)
+{
+    OpBuilder::InsertionGuard guard(builder);
+    MLIRContext *ctx = rPPMOp.getContext();
+
+    SmallVector<Type> qubitResultsType;
+    for (size_t i = 0; i < rPPMOp.getQubits().size(); i++) {
```
sengthai, line 1072:
```suggestion
    for (size_t i = 0; i < rPPMOp.getQubits().size(); ++i) {
```

mlir/lib/QRef/Transforms/value_semantics_conversion.cpp
```diff
@@ -1061,6 +1062,23 @@ void handleMeasureInBasis(IRRewriter &builder, qref::MeasureInBasisOp rMeasureIn
...
+        qubitResultsType.push_back(quantum::QubitType::get(ctx));
+    }
+
+    auto vPPMOp =
+        migrateOpToValueSemantics<pbc::PPMeasurementOp>(builder, rPPMOp, tracker, qubitResultsType);
```
sengthai, line 1077: Just to clarify, does `migrateOpToValueSemantics` handle the `negated` attribute as well?

sengthai (approved):
It seems the conversion framework is well constructed, so adding new ops looks clean. Good job! Thanks @paul0403

mlir/lib/QRef/Transforms/value_semantics_conversion.cpp
paul0403: It is constructed to move all attributes over: https://github.com/PennyLaneAI/catalyst/blob/dead506a0ce5ff6b591db06268a0d056ae28be91/mlir/lib/QRef/Transforms/value_semantics_conversion.cpp#L592

>  We create the corresponding value semantics operation, with exactly the same operands and
 attributes, except we replace the rQubit (and rQreg) Values with the corresponding current
 vQubit (and vQreg) Values.

In code: https://github.com/PennyLaneAI/catalyst/blob/dead506a0ce5ff6b591db06268a0d056ae28be91/mlir/lib/QRef/Transforms/value_semantics_conversion.cpp#L647

mlir/lib/QRef/Transforms/value_semantics_conversion.cpp
paul0403: https://github.com/PennyLaneAI/guidance-docs/blob/master/development/cpp-dev-guidelines.md#design-practices

> Prefer post-increment for all iterated indices & iterators.

mlir/test/QRef/VerifierTests.mlir
```diff
@@ -241,3 +241,11 @@ func.func @test_mbqc_graph_state_prep_invalid_size() {
     %graph_reg = qref.mbqc.graph_state_prep (%adj_matrix : tensor<1xi1>) [init "Hadamard", entangle "CZ"] : !qref.reg<4>
     func.return
 }
+
+// -----
+
+func.func @test_pbc_ppm_size_mismatch(%q: !qref.bit) {
+    // expected-error@below {{Number of qubits must match number of pauli operators}}
+    %m = qref.pbc.ppm ["Z", "X"] %q : i1
```
dime10: The naming structure here is a bit unfortunate, does this imply there is a single `qref` dialect that encompasses all other dialects? Doesn't really make too much sense to me 🤔 

Might be worth discussing an alternative that enables the "primary" dialect to be what we expect it to be (quantum, pbc, mbqc, etc). If we consider that `qref` was supposed to be an abbreviation of the actual dialect name, not the name itself (i.e. "QuantumRef", indicating it's a variation of the existing "Quantum" dialect), we could introduce something like this hierarchy: `quantum.ref.custom`, `pbc.ref.ppm`, `mbqc.ref.measure_in_basis`, ...

But I'm also curious to hear other ideas that don't involve `qref` being a single amalgamation of all other dialects :)

paul0403: I went with this because the alternative is for each dialect to have its own "ref" dialect, which sounds even more horrible : (

paul0403: The other question regarding organization is what to do with the `--convert-to-value-semantics` pass. If we have multiple ref ops scattered around all the dialects, then it kinda doesn't make sense for that pass to live in qref and contain convertion logic for ops in other quantum dialects. Maybe we need to do something similar to bufferization, where the qref dialect only has the "main converter", and each quantum dialect define their own "op interface methods" that define the conversion....

dime10: > I went with this because the alternative is for each dialect to have its own "ref" dialect, which sounds even more horrible : (

Disagree, I think that would be preferable. We don't have a single dialect for all the existing dialects either just because they share "value semantics".

While MLIR has two distinct dialects for tensors (value-sem) and memrefs (memory-sem), the difference is that these dialects specifically focus on those (i.e. what can you do with tensors or memrefs). Domain specific dialects seem either work with both (e.g. linalg can operate on both tensors and memrefs) or chose to explicitly work with only one of those (e.g. stablehlo).

dime10: > The other question regarding organization is what to do with the --convert-to-value-semantics pass. If we have multiple ref ops scattered around all the dialects, then it kinda doesn't make sense for that pass to live in qref and contain convertion logic for ops in other quantum dialects. Maybe we need to do something similar to bufferization, where the qref dialect only has the "main converter", and each quantum dialect define their own "op interface methods" that define the conversion....

Yes I don't think the pass has to be specific to a dialect in this case. It could be implemented via an extensible interface like you say, but we can also just have the pass contain all the conversions directly. The point of the extensible interface is that the people writing the pass don't yet know what ops or dialects they will encounter, we do.

dime10: So I think we can go about this in two/three ways:
- Each dialect has reference- and value-semantic versions of their ops, in one dialect.
- Like above, but using two dialects for each current dialect.
- Be stricter about the dialects we are targeting from the frontend. In particular, we disallow any direct generation of lower-level abstractions like PBC/MBQC etc. Instead, Quantum(Ref) contains everything we wish to target from the frontend. Other abstraction levels exclusively pick one of reference or value semantics.

lazypanda10117 (commented):
Looks good to me, thanks @paul0403 :) Will approve once the other comments, specifically David's, are addressed.

mlir/lib/QRef/Transforms/value_semantics_conversion.cpp
sengthai: Nice!

mlir/test/QRef/VerifierTests.mlir
paul0403: I ended up adding the reference semantics ppm op to the pbc dialect itself.

mehrdad2m (approved):

PR #2776, opened 2026-04-29
Add lowering for qecl gates to the qecp dialect

lillian542:
**Context:**
We want to lower transversal logical gates on logical qubits to single `qecp` gates applied on individual physical qubits in the codeblock.

We add lowering for the qecl gates if a mapping is provided in the `QecCode` for how to apply the gate transversally. 

**Benefits:**
We can lower the `qecl` gates to the `qecp` dialect for the Steane code.

**Possible Drawbacks:**
This is probably the part of the pass that will need the most substantial reworking to extend to k>1. Most of this will likely need to change.

**Related GitHub Issues:**
N/A


frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -243,6 +243,109 @@ def match_and_rewrite(self, op: qecl.MeasureOp, rewriter: PatternRewriter):
...
+
+
+@dataclass(frozen=True)
+class Transversal1QGateConversion(RewritePattern):
+    """Converts transversal single-qubit `qecl` gate ops specified in the QEC code definition to
+    a call to a subroutine that performs the transversal gate on the physical codeblock.
+
+    As implemented, this pattern does not use the `idx` attribute/operand of the logical gate
+    op, which indicates the position of the logical qubit within the codeblock to act on. The
+    current implementation for specifying gates also doesn't allow for specifying differen patterns
```
multiphaseCFD:
```suggestion
    current implementation for specifying gates also doesn't allow for specifying different patterns
```

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -243,6 +243,109 @@ def match_and_rewrite(self, op: qecl.MeasureOp, rewriter: PatternRewriter):
...
+@dataclass(frozen=True)
+class Transversal2QGateConversion(RewritePattern):
+    """Converts transversal two-qubit `qecl` gate ops specified in the QEC code definition to
+    a call to a subroutine that performs the transversal gate on the physical codeblock. Note that
+    this is currently limited to the `qecl.cnot` gate, so the pattern as implemented matches to
+    that rather than a generalised `qecl.TwoQubitLogicalGateOp`.
+
+    As implemented, this pattern does not use the `idx` attribute/operand of the logical gate
+    op, which indicates the position of the logical qubit within the codeblock to act on. The
+    current implementation for specifying gates also doesn't allow for specifying differen patterns
```
multiphaseCFD:
```suggestion
    current implementation for specifying gates also doesn't allow for specifying different patterns
```

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -243,6 +243,109 @@ def match_and_rewrite(self, op: qecl.MeasureOp, rewriter: PatternRewriter):
...
+
+        if k > 1:
+            raise NotImplementedError(
+                "Lowering logical gates to the `qecp` dialect is not implemented for k > 1 codes."
+            )
+
+        gate_name = op.name.split(".")[1]
+        gate_subroutine = self.gate_subroutines.get(gate_name, None)
+
+        if gate_subroutine is None:
```
multiphaseCFD, line 337: I'm wondering why we silently skip the conversion here?

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -243,6 +243,109 @@ def match_and_rewrite(self, op: qecl.MeasureOp, rewriter: PatternRewriter):
...
+
+        gate_name = op.name.strip("qecl.")
+        gate_subroutine = self.gate_subroutines.get(gate_name, None)
+
+        if gate_subroutine is None:
+            return
+
+        subroutine_call_op = func.CallOp(
+            callee=SymbolRefAttr(gate_subroutine.sym_name),
+            arguments=(op.in_codeblock,),
```
multiphaseCFD: Is `op.in_codeblock` in the `qecp` dialect or `qecl`? I guess it's already converted with `CodeblockTypeConversion` pattern.

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -440,6 +559,124 @@ def create_encode_subroutine(self) -> func.FuncOp:
...
+
+            with ImplicitBuilder(block):
+                (in_codeblock,) = block.args
+
+                # extract qubits
+                extract_ops = [qecp.ExtractQubitOp(in_codeblock, i) for i in range(self.qec_code.n)]
+                qubits = [ext_op.results[0] for ext_op in extract_ops]
+
+                transversal_gate = [
+                    gate_op(qb) if idx in gate_indices else qecp.IdentityOp(qb)
```
multiphaseCFD, line 668: Oh, I see, for Pauli gates, we only apply physical gates to the data qubits on an edge in the geometrical representation of Steane code. I thought we would apply Paulis to all data qubits instead.

frontend/catalyst/python_interface/transforms/qecp/qec_code_lib.py
```diff
@@ -15,21 +15,41 @@
...
         np.array([[1, 1, 1, 1, 0, 0, 0], [0, 1, 1, 0, 1, 1, 0], [0, 0, 1, 1, 0, 1, 1]]),
         np.array([[1, 1, 1, 1, 0, 0, 0], [0, 1, 1, 0, 1, 1, 0], [0, 0, 1, 1, 0, 1, 1]]),
+        {
+            # keys need to match the names of the corresponding qecl.gate gates
+            # values are a tuple of the qecp gate, and the indices its applied at in the codeblock
+            # will need to be refactored for k>1
+            "x": (qecp.PauliXOp, [4, 5, 6]),
+            "y": (
+                qecp.PauliYOp,
+                [4, 5, 6],
```
multiphaseCFD, line 39: I remembered @joeycarter previously suggested to traverse all data qubits for the Z, X basis measurement operation. I believe we need alignments on this.

multiphaseCFD (commented):
Look good to me! Thanks @lillian542 !

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -243,6 +243,109 @@ def match_and_rewrite(self, op: qecl.MeasureOp, rewriter: PatternRewriter):
...
+    current implementation for specifying gates also doesn't allow for specifying differen patterns
+    for different qubits within a codeblock. We assume `k=1` and therefore `idx=0`.
+    """
+
+    qec_code: QecCode
+
+    gate_subroutines: dict[str, func.FuncOp]
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: qecl.SingleQubitLogicalGateOp, rewriter: PatternRewriter):
```
joeycarter: [Optional] There's a decent amount of code reuse here and in `Transversal2QGateConversion`, so you might be able to do:

```suggestion
    def match_and_rewrite(self, op: qecl.SingleQubitLogicalGateOp | qecl.CnotOp, rewriter: PatternRewriter):
```

and then insert the necessary `isinstance(op, <type>)` to handle the one-qubit or two-qubit ins/outs.

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
joeycarter: In practice it will be the `qecp` codeblock, since the type conversion happens first, but in principle it can be either (since it's equally valid to run the type conversion pattern at the end).

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
joeycarter: @lillian542 can correct me if I'm wrong, but I think it's to handle the case where the pattern encounters a logical gate that is not transversal, which would need to be lowered to the `qecp` dialect via a separate rewrite pattern.

Maybe a comment here would be helpful.

frontend/catalyst/python_interface/transforms/qecp/qec_code_lib.py
joeycarter: @multiphaseCFD I think this is fine as is. The protocol we're using for logical measurements is a transversal measurement of all the data qubits in the physical codeblock, which as we have it is a fully separate process from applying logical gates to the codeblock. We would need to rework the measurement protocol a bit to support measurements in a basis other than the computational basis, but the principle still holds. It's true that another way to perform logical measurements on the codeblock is to do an ESM-style entangling of an aux qubit with the codeblock according to the gate operations defined here (this would be a non-destructive measurement protocol). However, I don't believe we have plans to support a protocol like that at the moment.

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
joeycarter, line 668: Mind if I ask, do we need the identity ops here? Is the goal just to have an op applied to every physical qubit in the codeblock for better symmetry? easier testing?

frontend/catalyst/python_interface/transforms/qecp/qec_code_lib.py
```diff
@@ -15,21 +15,41 @@
...
         7,
         1,
         3,
         np.array([[1, 1, 1, 1, 0, 0, 0], [0, 1, 1, 0, 1, 1, 0], [0, 0, 1, 1, 0, 1, 1]]),
         np.array([[1, 1, 1, 1, 0, 0, 0], [0, 1, 1, 0, 1, 1, 0], [0, 0, 1, 1, 0, 1, 1]]),
+        {
+            # keys need to match the names of the corresponding qecl.gate gates
+            # values are a tuple of the qecp gate, and the indices its applied at in the codeblock
+            # will need to be refactored for k>1
+            "x": (qecp.PauliXOp, [4, 5, 6]),
```
joeycarter, line 36: I know you had concerns about this interface but I think this is as clean as we could reasonable make it for now! 💯

joeycarter (commented):
Looks great, thanks @lillian542! 🚀 

I don't have much to add, just a few questions and suggestions.

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
lillian542: Exactly :) Good point, I'll add a comment.

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
lillian542: I went back and forth on this a bit, but opted to add them. 

Identity might well not actually mean "do nothing" low level, but rather "actively preserve current state for this cycle", so I thought it was preferable to be explicit.

But we might not want to have these in here at all - or if we do, we might want to "pad" with identity everywhere nothing is happing later, after scheduling.

lillian542: For this particular code, I think we could do either. Since I've only added [4, 5, 6] as the indices for the Pauli gates, that's what it will do for the Steane code as currently defined. It's based on the Ryan-Anderson et. al. paper we read/drew from earlier this year, which defined the transversal gates as:

image

joeycarter: Sounds good, I think it's fine to leave them in for now for the sake of being explicit and not knowing what the target architecture is at the moment. For the prototype where we convert qecp gate ops back to the quantum dialect, we can always remove them their for better simulation efficiency.

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
lillian542: Reorganized a bit to combine the two patterns, it does seem a bit tidier that way!

doc/releases/changelog-dev.md
```diff
@@ -15,8 +15,15 @@
 - Update RC nightly builds to read version number from the `_version.py` file 
   [(#2797)](https://github.com/PennyLaneAI/catalyst/pull/2797)
 
+* A the experimental compiler pass `convert-qecl-to-qecp` has been extended to lower
```
lillian542:
```suggestion
* The experimental compiler pass `convert-qecl-to-qecp` has been extended to lower 
```

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -56,6 +56,7 @@
 
 from .convert_qecl_noise_to_qec_noise import ConvertQECLNoiseOpToQECPNoisePass
 from .qec_code_lib import QecCode
+from ..qecl.convert_quantum_to_qecl import _get_idx_value_or_attr_from_extract_or_insert_op
```
lillian542: Added for a bug fix for the insert/extract pattern that came up when testing the (quantum--> qecl --> qecp) pipeline with this PR. 

Not sure if we want to move this function out of the `qecl` dialect and into a more general location. Maybe the `python_interface.utils` folder? It might be too specific for that.

joeycarter: Feels like we might need a common `qec` module for all things that have some overlap between `qecl` and `qecp`. I think this is fine for now while we're still prototyping.

frontend/catalyst/python_interface/transforms/qecp/qec_code_lib.py
```diff
@@ -65,7 +105,16 @@ def from_dict(cls, data: dict) -> Self:
...
+        >>> QecCode.from_dict({
+        >>>    'name': "Steane",
+        >>>    'n': 7,
+        >>>    'k': 1,
+        >>>    'd': 3,
+        >>>    "x_tanner": np.eye(7),
+        >>>    "z_tanner": np.eye(7),
+        >>>    "transversal_1q_gates": {},
+        >>>    "transversal_2q_gates": {},
+        >>>    })
```
joeycarter:
```suggestion
        >>> QecCode.from_dict({
        ...    'name': "Steane",
        ...    'n': 7,
        ...    'k': 1,
        ...    'd': 3,
        ...    "x_tanner": np.eye(7),
        ...    "z_tanner": np.eye(7),
        ...    "transversal_1q_gates": {},
        ...    "transversal_2q_gates": {},
        ...    })
```

joeycarter (approved):
Looks good, nice work @lillian542! 💯

multiphaseCFD (approved):
Great one! Thanks @lillian542

lillian542:
[sc-115308]

PR #2777, opened 2026-04-30
Add `combine_global_phase` to `passes`

isaacdevlugt:
**Context:** https://github.com/PennyLaneAI/catalyst/pull/2604 added the pass to the CLI, but we should have the pass available in the ``passes`` module.

**Description of the Change:** Adds `combine_global_phase` to ``passes`` module.

**Benefits:** Better access for users (don't have to use the pass via ``qp.transforms(pass_name="combine-global-phases")``)

**Possible Drawbacks:**

**Related GitHub Issues:** [sc-113790](https://app.shortcut.com/xanaduai/story/113790/parity-synth-and-combine-global-phases-in-the-passes-module)

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -291,9 +291,10 @@
   [(#2660)](https://github.com/PennyLaneAI/catalyst/pull/2660)
 
 * A new optimization pass has been added to reduce the number of instructions in a quantum program,
-  `--merge-global-phase`, which safely combines global phase instructions for each region in the
+  `--combine-global-phase`, which safely combines global phase instructions for each region in the
   program. The xDSL version written in Python has been removed.
   [(#2604)](https://github.com/PennyLaneAI/catalyst/pull/2604)
+  [(#???)](https://github.com/PennyLaneAI/catalyst/pull/???)
 
 * A new native-MLIR graph-based decomposition framework is now available. This system
   migrates the graph-decomposition logic from Python into the Catalyst compiler as a
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -294,7 +294,7 @@
   `--combine-global-phase`, which safely combines global phase instructions for each region in the
   program. The xDSL version written in Python has been removed.
   [(#2604)](https://github.com/PennyLaneAI/catalyst/pull/2604)
-  [(#???)](https://github.com/PennyLaneAI/catalyst/pull/???)
+  [(#2777)](https://github.com/PennyLaneAI/catalyst/pull/2777)
 
 * A new native-MLIR graph-based decomposition framework is now available. This system
   migrates the graph-decomposition logic from Python into the Catalyst compiler as a
--- a/frontend/catalyst/passes/builtin_passes.py
+++ b/frontend/catalyst/passes/builtin_passes.py
@@ -609,6 +609,111 @@ ions_decomposition = qp.transform(
 )
 
 
+def combine_global_phases_setup_inputs():  # pragma: nocover
+    r"""A quantum compilation pass that combines global phase instructions for each region in the
+    program.
+
+    Args:
+        qnode (QNode): the QNode to apply the compiler pass to
+
+    Returns:
+        :class:`QNode <pennylane.QNode>`
+
+    **Example**
+
+    In this example, the ``GlobalPhase`` gates are merged together into one ``GlobalPhase``
+    operation.
+
+    .. code-block:: python
+
+        import pennylane as qp
+        import catalyst
+        from catalyst import qjit, for_loop
+        from catalyst.passes import combine_global_phases
+        from catalyst.debug import get_compilation_stage
+
+        n = 10
+        dev = qp.device('null.qubit', wires=n)
+
+        @qjit(keep_intermediate=True, capture=True)
+        @combine_global_phases
+        @qp.qnode(dev)
+        def circuit(n):
+            qp.GlobalPhase(0.1, wires = n-1)
+            qp.X(n-1)
+            qp.GlobalPhase(0.1, wires = n-2)
+            qp.H(n-2)
+
+            @qp.for_loop(0, n)
+            def loop(i):
+                qp.GlobalPhase(0.1967, wires=i)
+                qp.GlobalPhase(0.7691, wires=i)
+
+            loop()
+
+            qp.GlobalPhase(0.1, wires=n-3)
+            qp.GlobalPhase(0.1, wires=0)
+
+            return qp.expval(qp.Z(0))
+
+    >>> circuit(n)
+    0.0
+
+    The ``GlobalPhase`` operations surrounding control flow operations will be merged, while those
+    within the control flow are merged together separately (i.e., no formal loop-boundary
+    optimizations).
+
+    Example MLIR Representation:
+
+    >>> print(get_compilation_stage(circuit, stage="QuantumCompilationStage"))
+
+    .. code-block:: mlir
+
+        . . .
+        %extracted_4 = tensor.extract %3[] : tensor<i64>
+        %5 = quantum.extract %4[%extracted_4] : !quantum.reg -> !quantum.bit
+        %out_qubits = quantum.custom "PauliX"() %5 : !quantum.bit
+        %6 = stablehlo.subtract %arg0, %c_1 : tensor<i64>
+        %extracted_5 = tensor.extract %3[] : tensor<i64>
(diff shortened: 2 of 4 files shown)


doc/releases/changelog-dev.md
```diff
@@ -291,9 +291,10 @@
   [(#2660)](https://github.com/PennyLaneAI/catalyst/pull/2660)
 
 * A new optimization pass has been added to reduce the number of instructions in a quantum program,
-  `--merge-global-phase`, which safely combines global phase instructions for each region in the
+  `--combine-global-phase`, which safely combines global phase instructions for each region in the
   program. The xDSL version written in Python has been removed.
   [(#2604)](https://github.com/PennyLaneAI/catalyst/pull/2604)
+  [(#???)](https://github.com/PennyLaneAI/catalyst/pull/???)
```
```suggestion
  [(#2777)](https://github.com/PennyLaneAI/catalyst/pull/2777)
```

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -609,6 +609,111 @@ def circuit():
...
+    Args:
+        qnode (QNode): the QNode to apply the compiler pass to
+
+    Returns:
+        :class:`QNode <pennylane.QNode>`
+
+    **Example**
+
+    In this example, the ``GlobalPhase`` gates are merged together into one ``GlobalPhase``
+    operation.
```
```suggestion
    Consider the following example:
```

Misleading to say that they're all merged together.

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -609,6 +609,111 @@ def circuit():
...
+        quantum.gphase(%cst)
+        %19 = quantum.namedobs %18[ PauliZ] : !quantum.obs
+        %20 = quantum.expval %19 : f64
+        . . .
+    """
+    return (), {}
+
+
+combine_global_phases = qp.transform(
+    pass_name="combie-global-phases", setup_inputs=combine_global_phases_setup_inputs
```
andrijapau:
```suggestion
    pass_name="combine-global-phases", setup_inputs=combine_global_phases_setup_inputs
```

andrijapau: I feel like CI should have caught this 🤔

isaacdevlugt: :oh-maaaan:

andrijapau: What do you think we enforce the convention of pass name (snake case) = variable name? This will allow us to catch these in CI? @dime10 @isaacdevlugt

andrijapau: I'm also wondering if we should just add a simple smoke integration test to the new `test_builtin_passes.py` like,
```python
@qjit
@pass_obj
def circuit():
    qp.Hadamard(0)
    return qp.expval(qp.Z(0))

circuit()  # triggers compilation - a bad pass_name should be raised here
```

andrijapau: If you want I can push those changes to this PR just lmk.

dime10: > What do you think we enforce the convention of pass name = variable name? This will allow us to catch these in CI? @dime10 @isaacdevlugt

Sounds good to me 👍

andrijapau: Tests are up 🤞🏼

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -609,6 +609,110 @@ def circuit():
 )
 
 
+def combine_global_phases_setup_inputs():  # pragma: nocover
```
andrijapau: Don't think we need this 🤔 
```suggestion
def combine_global_phases_setup_inputs(): 
```

andrijapau (approved):

paul0403:
Just wondering, isn't the whole point that we want to unify calling passes as `qp.transforms(pass_name="combine-global-phases")`? 

Do we need to specifically expose this in the legacy frontend?

isaacdevlugt:
> Just wondering, isn't the whole point that we want to unify calling passes as `qp.transforms(pass_name="combine-global-phases")`?
> 
> Do we need to specifically expose this in the legacy frontend?

@paul0403 thanks for the question! Yes and no 😅. Yes: we want users to access any and all transforms via PennyLane. No: ``combine_global_phases`` has a proper MLIR implementation in Catalyst, which requires that it have documentation that lives alongside it (in the Catalyst repo), and I would argue that that should be next to other compilation passes in Catalyst (in the ``passes`` module). 

We are still in a bit of a grey area with Catalyst having a Python frontend that users can access, but PennyLane is the frontend we want users to use for everything. So, just trying to straddle the line the best we can for now 🚀

frontend/test/pytest/test_builtin_passes.py
```diff
@@ -45,9 +46,47 @@ def assert_valid_transform(obj: Any) -> None:
     assert pipeline[0].pass_name == pass_name
 
 
+_PASS_NAME_SPECIAL_CASES = {"diagonalize_measurements": "diagonalize-final-measurements"}
```
paul0403: Can you put these special cases inside the tests? Since they're only used by one test, I don't think it's necessary to make them global

paul0403: Same with the other dictionary

frontend/test/pytest/test_builtin_passes.py
```diff
@@ -45,9 +46,47 @@ def assert_valid_transform(obj: Any) -> None:
...
+
+_PASSES_WITH_ARGS_KWARGS = {"graph-decomposition": ((), {"gate_set": ()})}
+
+
+@pytest.mark.parametrize("name", builtin_passes.__all__)
+def test_pass_compiles_with_qjit(name):
+    """Basic smoke test to ensure proper compilation. For example, if a pass name
+    was incorrect this would fail."""
+
+    obj = getattr(builtin_passes, name)
```
paul0403:
```suggestion
    transform = getattr(builtin_passes, name)
```

renaming a bit to make the tests more readable : )

frontend/test/pytest/test_builtin_passes.py
```diff
@@ -45,9 +46,47 @@ def assert_valid_transform(obj: Any) -> None:
...
+
+    args, kwargs = _PASSES_WITH_ARGS_KWARGS.get(obj.pass_name, ((), {}))
+
+    bound_obj = obj
+    if args or kwargs:
+        bound_obj = obj(*args, **kwargs)
+
+    @qp.qjit(target="mlir")
+    @bound_obj
```
paul0403:
```suggestion
    @qp.qjit(target="mlir")
    @transform(*args, **kwargs)
```

This check does not need to exist right? Since unpacking empty tuples/dictionaries is fine?

paul0403:
```python
>>> def f(*args, **kwargs):
...    return *args, **kwargs

>>> f(*(), **{})
((), {})
```

frontend/test/pytest/test_builtin_passes.py
andrijapau: Sure

andrijapau: Done!

frontend/test/pytest/test_builtin_passes.py
andrijapau: If you don't do this you get a bunch of [failures](https://github.com/PennyLaneAI/catalyst/actions/runs/25185523201/job/73842424177) 🤔

paul0403: https://github.com/PennyLaneAI/pennylane/blob/f7c09964a811db3095eb6226e267ea65296f946d/pennylane/transforms/core/transform.py#L635

ah ok we explicitly check for it

All good then

frontend/test/pytest/test_builtin_passes.py
```diff
@@ -45,9 +46,43 @@ def assert_valid_transform(obj: Any) -> None:
...
     obj = getattr(builtin_passes, name)
     assert_valid_transform(obj)
+
+
+@pytest.mark.parametrize("name", builtin_passes.__all__)
+def test_pass_compiles_with_qjit(name):
+    """Basic smoke test to ensure proper compilation. For example, if a pass name
+    was incorrect this would fail."""
+
+    transform = getattr(builtin_passes, name)
```
paul0403, line 72: Well, remember to change the name where it's used 😅

andrijapau: Yea oops. Just pushed a commit :)

paul0403 (approved):
Nice :100:

PR #2772, opened 2026-04-28
Add qecp.decode_esm_css to llvm lowering pass

multiphaseCFD:
### Before submitting

Please complete the following checklist when submitting a PR:

- [x] All new functions and code must be clearly commented and documented.

- [x] Ensure that code is properly formatted by running `make format`.
      The latest version of black and `clang-format-20` are used in CI/CD to check formatting.

- [x] All new features must include a unit test.
      Integration and frontend tests should be added to [`frontend/test`](../frontend/test/),
      Quantum dialect and MLIR tests should be added to [`mlir/test`](../mlir/test/), and
      Runtime tests should be added to [`runtime/tests`](../runtime/tests/).

When all the above are checked, delete everything above the dashed
line and fill in the pull request template.

------------------------------------------------------------------------------------------------------------

**Context:**

[sc-117444]

This PR lowers the `qecp.decode_esm_css` operation to runtime decoder.

before

```mlir
qecp.decode_esm_css(%tanner : !qecp.tanner_graph<8, 6, i32>) %esm in(%err_buf : memref<2xindex>) : memref<3xi1>
```
after

```mlir
....
llvm.call @__catalyst__qecp__lut_decoder(%17, %21, %19) : (!llvm.ptr, !llvm.ptr, !llvm.ptr) -> ()
```

**Description of the Change:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

PR #2780, opened 2026-05-01
MLIR to LLVM lowering for `qecp.assemble_tanner`

multiphaseCFD:
### Before submitting

**Context:**

[sc-118325]

 In #2768, we added bufferization interfaces to the `qecp.assemble_tanner` operation (among others). Now, this operation requires a lowering pattern for conversion to the LLVM dialect.

This PR includes a new type converter for `!qecp.tanner_graph` types as well as an op conversion pattern for `qecp.assemble_tanner` ops to the LLVM dialect:

* _Tanner graph type converter:_ Converts MLIR `!qecp.tanner_graph` types to [LLVM identified structs](https://llvm.org/docs/LangRef.html#structure-types) of the form
    ```mlir
    !llvm.struct<"TannerGraph", (memref<...xi32)>, memref<...xi32>)>
    ```
    With the `memref`s expanded out into their `memref` descriptor form, this struct is:
    ```mlir
    !llvm.struct<"TannerGraph", (
        struct<(ptr, ptr, i64, array<1 x i64>, array<1 x i64>)>,
        struct<(ptr, ptr, i64, array<1 x i64>, array<1 x i64>)>)>
    ```

* _Op conversion pattern for assemble-Tanner ops:_ After the input `row_idx` and `col_ptr` `memref`s have been expanded out into their `memref` descriptor form, `struct<(ptr, ptr, i64, array<1 x i64>, array<1 x i64>)>`, the rewrite pattern simply inserts these `memref` structs into the `TannerGraph` struct and replaces the `qecp.assemble_tanner` op with the assembled `TannerGraph` struct value.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


joeycarter (approved):

mlir/lib/QecPhysical/Transforms/qecp_to_llvm.cpp
```diff
@@ -0,0 +1,102 @@
...
+    {
+        addConversion([&](TannerGraphType type) { return convertTannerGraphType(type); });
+    }
+
+  private:
+    std::optional<Type> convertTannerGraphType(TannerGraphType mlirType)
+    {
+        auto *ctx = &getContext();
+
+        auto llvmStruct = LLVM::LLVMStructType::getIdentified(ctx, "TannerGraph");
```
rniczh, line 51: During the second conversion, the same `TannerGraph` gets overwritten. If we must use Identified, should we also mangle the parameters into the name?

mlir/lib/QecPhysical/Transforms/qecp_to_llvm.cpp
```diff
@@ -0,0 +1,102 @@
...
+#include "QecPhysical/IR/QecPhysicalOps.h"
+#include "QecPhysical/Transforms/Patterns.h"
+
+using namespace mlir;
+using namespace catalyst::qecp;
+
+namespace catalyst {
+namespace qecp {
+
+#define GEN_PASS_DECL_QECPHYSICALCONVERSIONPASS
```
rniczh: Why we need the `...DECL...` here?

mlir/lib/QecPhysical/Transforms/ConversionPatterns.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+    LogicalResult matchAndRewrite(AssembleTannerGraphOp op, AssembleTannerGraphOpAdaptor adaptor,
+                                  ConversionPatternRewriter &rewriter) const override
+    {
+        Location loc = op.getLoc();
+        const TypeConverter *conv = getTypeConverter();
+
+        // Both row_idx and col_ptr should be bufferized first.
+        if (!op.isBufferized()) {
+            return op.emitOpError("op must be bufferized before lowering to LLVM");
```
rniczh, line 42: Is your intention to `emitError` and then keep trying other patterns, or do you want to assert? Because this only skips this pattern.

mlir/lib/QecPhysical/Transforms/ConversionPatterns.cpp
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        auto mlirTannerType = op.getTannerGraph().getType();
+        auto tannerStructType = conv->convertType(mlirTannerType);
```
rniczh, line 46: Is this always convert successfully? do you need the null check?

mlir/lib/QecPhysical/Transforms/ConversionPatterns.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+        Value tannerStructValue = LLVM::UndefOp::create(rewriter, loc, tannerStructType);
```
rniczh, line 48: `Undef` is deprecated, should use `PoisonOp` instead

mlir/lib/QecPhysical/Transforms/ConversionPatterns.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+        tannerStructValue = LLVM::InsertValueOp::create(
+            rewriter, loc, tannerStructValue, adaptor.getRowIdx(), SmallVector<int64_t>{0});
```
rniczh: you don't need to construct a vector here, could probably just pass `{0}` directly.

mlir/include/QecPhysical/Transforms/Passes.td
```diff
@@ -0,0 +1,29 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef QECPHYSICAL_PASSES
+#define QECPHYSICAL_PASSES
+
+include "mlir/Pass/PassBase.td"
+
+def QecPhysicalConversionPass : Pass<"convert-qecp-to-llvm"> {
```
rniczh: Because you added the module level pass in your `runOperation`, I think you should specify the pass as Module pass here.

rniczh (commented):
Thanks! Just some comments from my side.

mlir/lib/QecPhysical/Transforms/qecp_to_llvm.cpp
rniczh: Or just use `getLiteral`

mlir/lib/QecPhysical/Transforms/ConversionPatterns.cpp
rniczh: Oh, sorry, it's not deprecated yet. Still in RFC discussion.

mlir/lib/QecPhysical/Transforms/qecp_to_llvm.cpp
joeycarter: Maybe I've misunderstood, is this case not handled by the `if (!llvmStruct.isInitialized()) {...}` block below?

mlir/lib/QecPhysical/Transforms/ConversionPatterns.cpp
joeycarter: This `if (!op.isBufferized())` block is used elsewhere in Catalyst, e.g. in a number of the `quantum` -> LLVM dialect-conversion rewrite patterns. It should be fine as is, it's more of a sanity check for the compilation pipeline being used, rather than the conversion pattern itself.

mlir/lib/QecPhysical/Transforms/ConversionPatterns.cpp
joeycarter: I think as we've defined it, the type conversion should always be successful (or at least, I can't come up with a test case that would trigger a failure here). I'm not too concerned here, this project is still in the prototyping phase.

mlir/lib/QecPhysical/Transforms/ConversionPatterns.cpp
joeycarter: The MLIR docs still use `undef` in a number of places, e.g. here: https://mlir.llvm.org/docs/TargetLLVMIR/#function-result-packing. We also use it elsewhere in Catalyst. Are the MLIR docs out of date? Where does it say that `llvm.poison` is preferred over `llvm.undef`?

mlir/lib/QecPhysical/Transforms/ConversionPatterns.cpp
joeycarter: Seems we can't:

```
.../dev/catalyst/mlir/lib/QecPhysical/Transforms/ConversionPatterns.cpp:50:29: error: call to 'create' is ambiguous
   50 |         tannerStructValue = LLVM::InsertValueOp::create(
      |                             ^~~~~~~~~~~~~~~~~~~~~~~~~~~
.../dev/catalyst/mlir/llvm-project/build/tools/mlir/include/mlir/Dialect/LLVMIR/LLVMOps.h.inc:14538:24: note: candidate function
 14538 |   static InsertValueOp create(::mlir::OpBuilder &builder, ::mlir::Location location, ::mlir::Value container, ::mlir::Value value, ::mlir::DenseI64ArrayAttr position);
       |                        ^
.../dev/catalyst/mlir/llvm-project/build/tools/mlir/include/mlir/Dialect/LLVMIR/LLVMOps.h.inc:14547:24: note: candidate function
 14547 |   static InsertValueOp create(::mlir::OpBuilder &builder, ::mlir::Location location, ::mlir::Value container, ::mlir::Value value, ::llvm::ArrayRef<int64_t> position);
       |                        ^
```

mlir/lib/QecPhysical/Transforms/qecp_to_llvm.cpp
rniczh: But once it need to construct that again, it will just overwrite the same struct?

mlir/lib/QecPhysical/Transforms/ConversionPatterns.cpp
rniczh: Sorry, just tried it, I think it should wrap with ArrayRef `ArrayRef<int64_t>{0}`

mlir/lib/QecPhysical/Transforms/ConversionPatterns.cpp
rniczh: Sorry, forgot to submit the comment. I misremembered this thing, it's still under discussion in some other RFCs, no any action needed. 😅

mlir/lib/QecPhysical/Transforms/ConversionPatterns.cpp
rniczh: Thanks for explanation

mlir/include/QecPhysical/Transforms/Passes.td
joeycarter: ~Is that necessary here? The `qecp` ops being converted to LLVM aren't anchored to a `builtin.module` op, i.e. it's not modifying any global symbols, function signatures, external linkages, etc.~

~Is there an issue in the `runOnOperation()` definition that makes it appear as a module-level pass?~

Discussed offline, the potential issue is the inclusion of the other conversion patterns, e.g. `populateFuncToLLVMConversionPatterns`. I'll look into this...

rniczh: One of the pattern running on the `runOperation` will try to modify the symbol tables.

It's in the `catalyst/mlir/llvm-project/mlir/lib/Conversion/FuncToLLVM/FuncToLLVM.cpp::convertFuncOpToLLVMFuncOp`

```cpp
if (symbolTables && symbolTableOp) {
    auto ip = rewriter.getInsertionPoint();
    SymbolTable &symbolTable = symbolTables->getSymbolTable(symbolTableOp);
    symbolTable.insert(newFuncOp, ip);
  }
```

mlir/lib/QecPhysical/Transforms/qecp_to_llvm.cpp
joeycarter: Good catch, looks like this was copied over from the `quantum-to-llvm` pass, which has pass options and so requires the `GEN_PASS_DECL_`. But you're right, it's not needed here. Removed: 9f89b3084

mlir/lib/QecPhysical/Transforms/ConversionPatterns.cpp
joeycarter: Yes `ArrayRef` works here. Fixed: 604a1fc61

mlir/include/QecPhysical/Transforms/Passes.td
rniczh: But I think if there is no tanner_graph involve in the function signature would probably fine 🤔 (e.g. `func.func @foo(%g: !qecp.tanner_graph<...>)`

mlir/lib/QecPhysical/Transforms/qecp_to_llvm.cpp
joeycarter: My understanding is that it doesn't overwrite the first struct, it returns the existing one: https://github.com/llvm/llvm-project/blob/db436f826bb97a1a36d300eaa8f8f5317fb66686/mlir/include/mlir/Dialect/LLVMIR/LLVMTypes.td#L184-L189

```cpp
/// Gets or creates an identified struct with the given name in the provided
/// context. Note that unlike llvm::StructType::create, this function will
/// _NOT_ rename a struct in case a struct with the same name already exists
/// in the context. Instead, it will just return the existing struct,
/// similarly to the rest of MLIR type ::get methods.
static LLVMStructType getIdentified(MLIRContext *context, StringRef name);
```

mlir/include/QecPhysical/Transforms/Passes.td
rniczh: So for this case, 

```mlir
module {
    memref.global "private" constant @__constant_8xi32 : memref<8xi32> = dense<[3, 3, 4, 4, 0, 1, 1, 2]> {alignment = 64 : i64}
    memref.global "private" constant @__constant_6xi32 : memref<6xi32> = dense<[0, 1, 3, 4, 6, 8]> {alignment = 64 : i64}

    func.func @returns_tanner() -> !qecp.tanner_graph<8, 6, i32> {
        %row_idx = memref.get_global @__constant_8xi32 : memref<8xi32>
        %col_ptr = memref.get_global @__constant_6xi32 : memref<6xi32>
        %tanner = qecp.assemble_tanner %row_idx, %col_ptr : memref<8xi32>, memref<6xi32> -> !qecp.tanner_graph<8, 6, i32>
        func.return %tanner : !qecp.tanner_graph<8, 6, i32>
    }

    func.func @takes_tanner(%row_idx: memref<8xi32>, %col_ptr: memref<6xi32>) -> !qecp.tanner_graph<8, 6, i32> {
        %tanner = qecp.assemble_tanner %row_idx, %col_ptr : memref<8xi32>, memref<6xi32> -> !qecp.tanner_graph<8, 6, i32>
        func.return %tanner : !qecp.tanner_graph<8, 6, i32>
    }
}
```

It will run into a segfault if running the following command:
```
quantum-opt test.mlir --pass-pipeline='builtin.module(finalize-memref-to-llvm, func.func(convert-qecp-to-llvm))'
```
Fine if running as module pass:
```
quantum-opt  test.mlir --pass-pipeline='builtin.module(finalize-memref-to-llvm, convert-qecp-to-llvm)
```

mlir/lib/QecPhysical/Transforms/qecp_to_llvm.cpp
rniczh: I probably see your point, so the type wouldn't change right? For example: it's what I get from `!qecp.tanner_graph<8, 6, i32>` and `!qecp.tanner_graph<16, 10, i64>`, both return the following type in the same llvm module.
```
%TannerGraph = type { { ptr, ptr, i64, [1 x i64], [1 x i64] }, { ptr, ptr, i64, [1 x i64], [1 x i64] } }
```

mlir/include/QecPhysical/Transforms/Passes.td
joeycarter: Interesting, thanks for looking into this. Sounds like the safer option is to make it a module-level pass. I've updated here: 1e68f170b

mlir/lib/QecPhysical/Transforms/qecp_to_llvm.cpp
joeycarter: That was my thinking, yes. At the MLIR level it's useful to know the size of the underlying `row_idx` and `col_ptr` buffers, but at the LLVM level, the structs that define the memrefs are the same, and hence the wrapper "TannerGraph" structs are also the same, regardless of their size.

Another way to look at it is in the runtime, every TannerGraph types is represented by the struct

```cpp
struct TannerGraph_CSC_int32 {
    struct MemRefT_int32_1d row_idx;
    struct MemRefT_int32_1d col_ptr;
};
```

where `MemRefT_int32_1d` is:

```cpp
struct MemRefT_int32_1d {
    int32_t *data_allocated;
    int32_t *data_aligned;
    std::size_t offset;
    std::size_t sizes[1];
    std::size_t strides[1];
};
```

(Side note: this is something we're adding in #2772)

rniczh (approved):
🎉

mlir/lib/QecPhysical/Transforms/qecp_to_llvm.cpp
rniczh: I see. Thank you for your explanation 👍

PR #2782, opened 2026-05-04
Restructure ResourceAnalysis around lifted loops and a flatten result object

sengthai:
**Context:**
Previously, ResourceAnalysis folded scf.for bodies into the caller (gates × trip count) and inlined func.call callees into the caller's resource map. With dynamic loops on the way, we want each loop body and each callee as a first-class entry, with one consistent rule for aggregation.

* Each loop body is now its own entry: static loops appear as for_loop_<N> with their trip count, and input-driven dynamic loops appear as dyn_for_loop_<N> with a stable identifier. Other dynamic loops keep the existing "count body once" fallback.
* Subroutines keep all their resource counts (gates, measurements, qubit allocations, classical instructions, and direct call references) local to themselves instead of being merged into the caller.
* The JSON output gains a small key for dynamic-loop identifiers; existing keys are unchanged. Synthetic loop names are guaranteed not to collide with user functions named e.g. for_loop_3.
* Qnode functions that use {auto_qubit_management} (where the device allocates qubits dynamically on first access rather than statically upfront) now emit "auto_qubit_management": true in the JSON. Functions without a quantum.device op do not emit the field at all. The field is modelled as std::optional<bool> in ResourceResult to distinguish three states: no device (nullopt), device without the flag (false), and device with the flag (true).

[sc-117893]


mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -226,42 +300,69 @@ ResourceAnalysis::ResourceAnalysis(Operation *op)
...
+{
+    // Bump `counter` until the resulting name does not collide with an
+    // existing entry. This protects against user functions named e.g.
+    // `for_loop_3` shadowing or being shadowed by a lifted body.
+    while (true) {
+        std::string candidate = prefix.str() + std::to_string(++counter);
+        if (!funcResults.contains(candidate)) {
+            return candidate;
+        }
+    }
```
jzaia18:
```suggestion
    std::string candidate;
    do {
        candidate = prefix.str() + std::to_string(++counter);
    } while (!funcResults.contains(candidate));

    return candidate;
```

Maybe something like this would be cleaner?

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -226,42 +300,69 @@ ResourceAnalysis::ResourceAnalysis(Operation *op)
...
-    else if (auto tripCount = forOp.getStaticTripCount()) {
-        bodyResult.multiplyByScalar(tripCount->getSExtValue());
+    else if (auto sTrip = forOp.getStaticTripCount()) {
+        tripCount = sTrip->getSExtValue();
     }
     else {
         auto lb = resolveConstantInt(forOp.getLowerBound());
         auto ub = resolveConstantInt(forOp.getUpperBound());
         auto step = resolveConstantInt(forOp.getStep());
         if (lb && ub && step && *step != 0 && *ub > *lb) {
```
jzaia18, line 279: I'm just realizing, is it necessarily the case that `*ub > *lb`? What about loops that count downward?

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -226,42 +300,69 @@ ResourceAnalysis::ResourceAnalysis(Operation *op)
...
+        result.varFunctionCalls[name] =
+            static_cast<uint64_t>(llvm::hash_value(forOp.getOperation()));
+        result.hasDynLoop = true;
+        return;
+    }
+
+    // Fallback: trip count is unknown and the loop is not input-driven.
+    // Count the body once and flag the function as containing a dyn loop.
+    result.hasDynLoop = true;
     result.mergeWith(bodyResult);
```
jzaia18: I'm confused why we treat input-driven loops differently. I think we should treat all dynamic loops the same --  a user might not directly control the input but they may still want to know how the loop iterations affect the number of resources.

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -226,42 +300,69 @@ ResourceAnalysis::ResourceAnalysis(Operation *op)
...
+        return;
+    }
+
+    if (hasInputDrivenBounds(forOp)) {
+        // Loop trip count comes from the user's inputs. Record the body under dyn_for_loop_<N>
+        // and store a fixed number (hash) so each such loop has its own id in the output.
+        std::string name = makeUniqueSyntheticName("dyn_for_loop_", dynForLoopCounter);
+        funcResults[name] = std::move(bodyResult);
+        result.varFunctionCalls[name] =
+            static_cast<uint64_t>(llvm::hash_value(forOp.getOperation()));
```
jzaia18: Are these guaranteed to be unique or could there be a collision?

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -463,40 +562,85 @@ void ResourceAnalysis::collectOperation(Operation *op, ResourceResult &result, b
...
- * by inlining callee resources into the caller.
+ * @brief Merge `child`'s quantum content, classical content, and transitive
+ * call counts into `flat`, scaled by `count`. Used to fold callee/loop-body
+ * contributions into a flattened view.
  *
- * @param funcName The name of the function to resolve calls for.
- * This will be called recursively on callees.
+ * @param flat The ResourceResult to accumulate counts into.
+ * @param child The ResourceResult to merge.
+ * @param count The scalar to multiply the counts by.
```
jzaia18:
```suggestion
 * @param count A scalar to multiply the child's counts by.
```

Can we also make this default to 1?

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -463,40 +562,85 @@ void ResourceAnalysis::collectOperation(Operation *op, ResourceResult &result, b
...
- * @brief Resolve function calls in the resource results
- * by inlining callee resources into the caller.
+ * @brief Merge `child`'s quantum content, classical content, and transitive
+ * call counts into `flat`, scaled by `count`. Used to fold callee/loop-body
+ * contributions into a flattened view.
  *
- * @param funcName The name of the function to resolve calls for.
- * This will be called recursively on callees.
+ * @param flat The ResourceResult to accumulate counts into.
+ * @param child The ResourceResult to merge.
```
jzaia18:
```suggestion
 * @param dest The ResourceResult to accumulate counts into (modified).
 * @param source The ResourceResult to merge into `dest` (unmodified).
```

I think the names "flat" and "child" are a bit confusing, how about something like this?

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -463,40 +562,85 @@ void ResourceAnalysis::collectOperation(Operation *op, ResourceResult &result, b
...
+static void accumulateScaled(ResourceResult &flat, const ResourceResult &child, int64_t count)
 {
-    auto it = funcResults.find(funcName);
-    if (it == funcResults.end()) {
-        return;
+    for (const auto &opEntry : child.operations) {
+        auto &innerDst = flat.operations[opEntry.getKey()];
+        for (const auto &sizeEntry : opEntry.getValue()) {
+            innerDst[sizeEntry.first] += sizeEntry.second * count;
+        }
```
jzaia18: I'm confused, why is `flat` also being scaled by `count`?

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -463,40 +562,85 @@ void ResourceAnalysis::collectOperation(Operation *op, ResourceResult &result, b
...
         }
-        resolveFunctionCalls(calledFunc);
-
-        // Scale and merge callee resources
-        ResourceResult calleeResources = calleeIt->second;
-        calleeResources.multiplyByScalar(callCount);
-        resources.mergeWith(calleeResources);
+        accumulateScaled(flat, *child, fc.getValue());
     }
+    return &flat;
```
jzaia18, line 584: Is it okay to use a raw pointers here? Can we guarantee that this memory will not become invalid before the pointer is used?

mlir/lib/Catalyst/Analysis/ResourceResult.cpp
```diff
@@ -36,39 +36,39 @@ static int64_t applyMerge(int64_t a, int64_t b, MergeMethod method)
...
-        functionCalls[entry.getKey()] =
-            applyMerge(functionCalls[entry.getKey()], entry.getValue(), method);
-    }
-
-    for (const auto &entry : other.unresolvedFunctionCalls) {
-        unresolvedFunctionCalls[entry.getKey()] =
-            applyMerge(unresolvedFunctionCalls[entry.getKey()], entry.getValue(), method);
+    // varFunctionCalls hold SSA-derived hashes; first writer wins on
+    // conflict. The `method` parameter is intentionally ignored — Sum/Max/Min
+    // are not meaningful on identifiers, and keys are minted uniquely by
```
jzaia18: I think we should generate a new unique hash instead. Long term these hashes are supposed to have meaning, each one maps to an (unknown) unique value. But if we keep the first one, we are essentially saying `x + y = x`, which may cause issues if another loop had the same hash intentionally.

This will not be an issue for the current implementation where each loop always has a unique hash, but the plan going forward will be that distinct loops with the same bounds may reuse the same hash.

mlir/lib/Catalyst/Analysis/ResourceResult.cpp
```diff
@@ -150,6 +146,16 @@ std::string ResourceResult::toJson(int indent) const
     }
     root["function_calls"] = std::move(fcObj);
 
+    // Variable function calls (dynamic-loop hashes). Emitted as fixed-width
+    // hex strings so that the high-bit portion of the 64-bit hash space
+    // round-trips losslessly through JSON (which would otherwise sign-flip
+    // when cast to int64 or lose precision past 2^53 as a double).
```
jzaia18:
```suggestion
    // round-trips losslessly through JSON.
```

I don't think the int64 issue is true, since each int is still unique. But either way JavaScript doesn't have `int`s, only the `Number` type, so a hex string is best here

jzaia18 (commented):
Some initial comments, will take another look soon

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: Nice!

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: Actually, MLIR already check if the [` assert .. ub>=lb` ](https://github.com/llvm/llvm-project/blob/8f264586d7521b0e305ca7bb78825aa3382ffef7/mlir/include/mlir/Interfaces/ControlFlowInterfaces.h#L305-L306) so it is unnecessary here, but here we do little check if when `ub==lb` then we skip it. In MLIR, scf.for only support possitive `step` which mean only count forward, so we are good here.

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: Yes, `makeUniqueSyntheticName` would guarantees that the name will always be unique.

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: > I think we should treat all dynamic loops the same -- a user might not directly control the input but they may still want to know how the loop iterations affect the number of resources.

if frontend is okay to support any dynamic loops, then it would be easy here. we can remove the `hasInputDrivenBounds` entirely.

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: > Can we also make this default to 1? 

Yes, we can~

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: `flat` itself is not being scaled. Only the incoming child contribution is scaled before being added into `flat`.
The reason for multiplying `child` by `count` is that `count` represents how many times that child contribution occurs, for example a calle called multiple times or a loop body repeated statically. 

actually, that equivalent to flat_existing_value = flat_existing_value + (child_value * count);

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: Yes, it should be safe. The pointer is either `nullptr` for an unknown/external function, or it points to a `ResourceResult` owned by `flattenedCache`, not a local temporary. And entries are not erased during flattening.

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
jzaia18: Sorry, I should've clarified. I meant for the hash value, is it guaranteed that these are always unique? Or could 2 different functions potentially have the same hash value?

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
jzaia18: Yes I think we prefer to treat all dynamic loops the same in the frontend

rniczh (approved):
💯

mlir/lib/Catalyst/Analysis/ResourceResult.cpp
sengthai: Agreed. First-writer-wins can combine different dynamic loops on future intentional hash reuse, so I updated this to mint a unique entry on conflict instead.

[26f5850](https://github.com/PennyLaneAI/catalyst/pull/2782/commits/26f58505794bcb2480eae9d851d5fea75ae1370c)

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: Ahh i see. and yes, the hash_value of op in MLIR is always unique.

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: That also works

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
jzaia18: Oh I see what's happening here. I misread.

doc/releases/changelog-dev.md
```diff
@@ -8,6 +8,12 @@
 
 <h3>Improvements 🛠</h3>
 
+* The `ResourceAnalysis` pass now reports each loop body and each subroutine as its own entry
+  instead of folding their gate counts into the caller. Static loops appear as `for_loop_<N>`
+  with their trip count, input-driven dynamic loops appear as `dyn_for_loop_<N>` with a stable
+  identifier, and totals across the call graph are computed on demand.
```
jzaia18, line 20:
```suggestion
* The `ResourceAnalysis` pass now reports each loop body and each subroutine as its own entry
  instead of folding their gate counts into the caller. Loops with constant bounds appear as `for_loop_<N>`
  with their trip count. Loops with dynamic bounds appear as `dyn_for_loop_<N>` with a stable
  identifier, and totals across the call graph are computed on demand.
```

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -463,40 +489,85 @@ void ResourceAnalysis::collectOperation(Operation *op, ResourceResult &result, b
...
- * by inlining callee resources into the caller.
+ * @brief Merge `source`'s quantum content, classical content, and transitive
+ * call counts into `dest`, scaled by `count`. Used to fold callee/loop-body
+ * contributions into a flattened view.
  *
- * @param funcName The name of the function to resolve calls for.
- * This will be called recursively on callees.
+ * @param dest The ResourceResult to accumulate counts into (modified).
+ * @param source The ResourceResult to merge into `dest` (unmodified).
+ * @param count A scalar to multiply the child's counts by (defaults to 1).
```
jzaia18:
```suggestion
 * @param count A scalar to multiply the `source`'s counts by (defaults to 1).
```

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -463,40 +489,85 @@ void ResourceAnalysis::collectOperation(Operation *op, ResourceResult &result, b
...
     }
+    for (const auto &m : source.measurements) {
+        dest.measurements[m.getKey()] += m.getValue() * count;
+    }
+    for (const auto &ci : source.classicalInstructions) {
+        dest.classicalInstructions[ci.getKey()] += ci.getValue() * count;
+    }
+    for (const auto &fc : source.functionCalls) {
+        dest.functionCalls[fc.getKey()] += fc.getValue() * count;
+    }
```
jzaia18, line 530: There should also be a loop for the `varFunctionCalls`, right?

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -463,40 +489,85 @@ void ResourceAnalysis::collectOperation(Operation *op, ResourceResult &result, b
...
+    ResourceResult &flat = flattenedCache[funcName];
+
+    // Self-contributions (no scaling).
+    accumulateScaled(flat, r);
+    flat.numArgQubits = r.numArgQubits; // own arg qubits only
+    flat.deviceName = r.deviceName;
+    flat.isQnode = r.isQnode;
+
+    // Any var_function_calls implies a dynamic loop with no static count.
+    flat.hasDynLoop = flat.hasDynLoop || !r.varFunctionCalls.empty();
```
jzaia18: Are these lines needed? I'm pretty sure you already set `flat.hasDynLoop` to true if we encounter a dynamic loop (up in line 293)

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -213,56 +225,72 @@ ResourceAnalysis::ResourceAnalysis(Operation *op)
...
+        // Later, classical ops from that body are added into the parent, multiplied by that count.
+        // The name is always new, so we don't overwrite an old entry.
+        std::string name = makeUniqueSyntheticName("for_loop_", forLoopCounter);
+        funcResults[name] = std::move(bodyResult);
+        result.functionCalls[name] = tripCount.value();
+        return;
+    }
+
+    // Loop trip count comes from the user's inputs. Record the body under dyn_for_loop_<N>
+    // and store a fixed number (hash) so each such loop has its own id in the output.
```
jzaia18:
```suggestion
    // Loop trip count is dynamic. Record the body under dyn_for_loop_<N>
    // and store a fixed number (hash) so each such loop has its own id in the output.
```

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -213,56 +225,72 @@ ResourceAnalysis::ResourceAnalysis(Operation *op)
...
+    std::string name = makeUniqueSyntheticName("dyn_for_loop_", dynForLoopCounter);
+    funcResults[name] = std::move(bodyResult);
+    result.varFunctionCalls[name] = static_cast<uint64_t>(llvm::hash_value(forOp.getOperation()));
+    result.hasDynLoop = true;
```
jzaia18, line 300: We mark this function as having a dynamic loop here (context for other comment)

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -463,40 +489,85 @@ void ResourceAnalysis::collectOperation(Operation *op, ResourceResult &result, b
...
+
+    // Any var_function_calls implies a dynamic loop with no static count.
+    flat.hasDynLoop = flat.hasDynLoop || !r.varFunctionCalls.empty();
+
+    // Pull in each direct callee's flattened counts, × how often we call it.
+    for (const auto &fc : r.functionCalls) {
+        const ResourceResult *child = getFlattenedResource(fc.getKey());
+        if (!child || child == &flat) {
+            continue;
         }
```
jzaia18, line 581: I think just skipping when we get recursion isn't ideal. This hides information from the user. We should at the very least raise a warning when we do this since it will affect the gate counts

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -202,9 +215,8 @@ ResourceAnalysis::ResourceAnalysis(Operation *op)
             analyzeRegion(region, result, /*isAdjoint=*/false);
         }
 
-        // count qubit arguments only for the entry function (first non-declaration).
-        // callee qubit args are not counted to avoid double-counting after
-        // function call resolution merges callee resources into the caller.
+        // count qubit arguments only for the entry function (first non-declaration)
+        // to avoid double-counting when callees are folded into a flattened view.
```
jzaia18: Is the first (non-declaration) function always guaranteed to be the entrypoint?

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: Good spot! varFunctionCalls should be propagated here too, but copied rather than scaled.

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: Removed.

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: Updated. I kept the flattening guard as a safety net, but the warning is now emitted earlier at the recursive func.call site, so it points to the actual call causing incomplete counts.

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: Good question.  it not guaranteed by MLIR in general, but it matches an existing Catalyst convention. jit.py already assumes in here. 

Not a general MLIR guarantee, but we already rely on the same convention in `jit.py`

https://github.com/PennyLaneAI/catalyst/blob/81d6022f2fb7a08352e3a932d7262bd7bbd21b9f/frontend/catalyst/jit.py#L883-L885

 the first function in the module is the entrypoint. I mirrored that here so num_arg_qubits is only counted on the entry, not on callees when flattening.

dime10: No this is not guaranteed and, in general, we shouldn't assume this!
The passage that Sengthai points out is the only exception to this; we can assume it here because it is right after tracing, and the tracing process is implemented such that the first function is indeed the entry point, but even this is not ideal and we have to hope jax never changes this, but we don't have another way currently.

dime10: Actually, this is the location I was thinking of that is right after tracing:
https://github.com/PennyLaneAI/catalyst/blob/5aeedb236d5e5a34a4447c1229d9d213c368c015/frontend/catalyst/utils/gen_mlir.py#L66

It adds the `llvm.emit_c_interface` attribute which could be used for detection. The name of the function is always prefixed with `jit_`. 
But we can make this more explicit though with something like a `catalyst.main` attribute or something.

sengthai: @dime10 Yes, we have talked this before but I didn't logged it. Now I open issue for adding or renaming attribute for entry function.  https://github.com/PennyLaneAI/catalyst/issues/2854 

Also, I updated to use  `llvm.emit_c_interface` for now in this PR.

jzaia18:
I commented out the integration tests since they're failing until the PennyLane frontend for this is merged as well. I uncomment these in #2900 where I add integration tests for this new feature

jzaia18 (approved):
Looking good. We should be able to merge this once the 2 followup PRs are approved and ready to merge as well.

The order should be
1. This PR
2. https://github.com/PennyLaneAI/pennylane/pull/9507
3. #2900

PR #2786, opened 2026-05-04
Dynamic one shot pass works with dynamically shaped samples

paul0403:
**Context:**
In dynamic one shot pass, fix a small logic error when collecting the SSA values for the dynamic dimensions on sample ops.

I have no idea what the old logic was doing :sweat_smile: (yeah I know I wrote it)


rniczh (approved):
💯

sengthai (approved):
Nice generalization!

PR #2789, opened 2026-05-04
Prepare for Catalyst RC v0.15.0 (update lightning dev)

sengthai:
lightning=0.45.0-dev43 is broken during uploading, so using dev44 instead.


JerryChen97 (approved):
🍀

lazypanda10117 (approved):

Issue #2540, opened 2026-03-03
MLIR pass plugin registration from Python no longer works


dime10:
This concerns the [apply_pass_plugin](https://github.com/PennyLaneAI/catalyst/blob/b15b417a4fe852d706137095e504e866822aca8f/frontend/catalyst/passes/pass_api.py#L212), which is supposed to load an MLIR plugin and schedule a pass from it with Catalyst. For starters, the docstring and [plugin docs](https://docs.pennylane.ai/projects/catalyst/en/latest/dev/plugins.html) are currently outdated and full of errors. However, ignoring those, the example does not appear to compile (requires that the plugin is built and installed):

```py
from catalyst import *
from standalone_plugin import getStandalonePluginAbsolutePath

@passes.apply_pass_plugin(getStandalonePluginAbsolutePath(), "standalone-switch-bar-foo")
@qml.qnode(qml.device("lightning.qubit", wires=1))
def qnode():
    return qml.state()

@qjit
def module():
    return qnode()

print(module.mlir)
print(module.mlir_opt)
```

```
%0 = "transform.apply_registered_pass"(%arg0) {options = {}, pass_name = "standalone-switch-bar-foo"} : (!transform.op<"builtin.module">) -> !transform.op<"builtin.module">
             ^
While processing 'transform_standalone-switch-bar-foo' pass While processing 'ApplyTransformSequencePass' pass While processing 'mlir::detail::OpToOpPassAdaptor' pass Failed to lower MLIR module
```

rniczh:
Inside `apply_pass_plugin`, the decorator only passes `pass_name` into `transform(...)`, and the `path_to_plugin` is never forwarded:

```python
def apply_pass_plugin(...)
    ...
	def decorator(obj):
	        return transform(pass_name=pass_name)(obj, *flags, **valued_options)
```

Because the path is dropped here, `EvaluationContext.add_plugin(path)` is never called and the compiler subprocess is never invoked with `--load-pass-plugin`, which is why catalyst reports unknown pass: `standalone-switch-bar-foo`. We can add a wrapper here to ensure the `EvaluationContext.add_plugin(path_to_plugin)` be executed before transform, what do you think? @dime10

dime10:
Thanks for looking into this @rniczh! That sounds reasonable :)

PR #2790, opened 2026-05-05
Fix that pass-plugin path never be forwarded

rniczh:
**Context:**

Decorator in `apply_pass_plugin` only passes `pass_name` into `transform(...)`, and the `path_to_plugin` is never forwarded.

Add a wrapper to ensure the `EvaluationContext.add_plugin(path_to_plugin)` be executed before transform. 

TODO:
~- [ ] Add tests~ We already have the tests in `standalone_plugin_wheel/standalone_plugin/test/test_mlir_plugin.py`
- [x] Changelog

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
https://github.com/PennyLaneAI/catalyst/issues/2540
[sc-112983]

### DIFF
--- a/doc/releases/changelog-0.15.0.md
+++ b/doc/releases/changelog-0.15.0.md
@@ -693,6 +693,10 @@ codes only.
 
 <h3>Bug fixes 🐛</h3>
 
+* Fixed a bug where the `path_to_plugin` never be forwarded in :func:`~.passes.apply_pass_plugin`. The plugin path is now registered with the compiler during tracing.
+  [(#2790)](https://github.com/PennyLaneAI/catalyst/pull/2790)
+
+
 * Refactored all passes in `catalyst.passes.builtin_passes.py` to be `pennylane.transforms.core.Transform` objects
   rather than decorators. This allows them to be used as standard transforms, enabling full compatibility with
   `pennylane.CompilePipeline`.
--- a/Makefile
+++ b/Makefile
@@ -348,9 +348,6 @@ endif
 ifeq ($(ENABLE_OQD), ON)
 	COVERAGE_FILE=$(MK_DIR)/.coverage.pytest $(ASAN_COMMAND) $(PYTHON) -m pytest frontend/test/test_oqd/oqd $(PYTEST_FLAGS) --cov=catalyst --cov-append --tb=native --cov-report=
 endif
-	$(MAKE) plugin-wheel
-	$(PYTHON) -m pip install --force-reinstall --no-deps $(MK_DIR)/standalone_plugin_wheel/dist/*.whl
-	COVERAGE_FILE=$(MK_DIR)/.coverage.pytest $(ASAN_COMMAND) $(PYTHON) -m pytest standalone_plugin_wheel/standalone_plugin/test $(PYTEST_FLAGS) --cov=catalyst --cov-append --tb=native --cov-report=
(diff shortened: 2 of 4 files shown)


paul0403:
We just found out that the plugin tests `standalone_plugin_wheel/standalone_plugin/test/test_mlir_plugin.py` are actually not covered under `make coverage-frontend`, which means they are not covered in CI!

@rniczh let's add it to `make coverage-frontend` in this PR!

dime10:
> We just found out that the plugin tests `standalone_plugin_wheel/standalone_plugin/test/test_mlir_plugin.py` are actually not covered under `make coverage-frontend`, which means they are not covered in CI!
> 
> @rniczh let's add it to `make coverage-frontend` in this PR!

That's because they are only tested once a week, I think they have a separate CI setup

paul0403:
> > We just found out that the plugin tests `standalone_plugin_wheel/standalone_plugin/test/test_mlir_plugin.py` are actually not covered under `make coverage-frontend`, which means they are not covered in CI!
> > @rniczh let's add it to `make coverage-frontend` in this PR!
> 
> That's because they are only tested once a week, I think they have a separate CI setup

:exploding_head:

paul0403:
Ok we found the runs in CI, but isn't it strange that these tests passed without @rniczh 's fix? https://github.com/PennyLaneAI/catalyst/actions/runs/25149431251/job/73717875061#step:14:1

rniczh:
@dime10 @paul0403 So it's because the `apply_pass_plugin(...)` never be covered if we're compiling with `@qjit(target='mlir')`. I add one more test to remove the `target='mlir'` to ensure we cover the failure test. https://github.com/PennyLaneAI/catalyst/pull/2790/commits/0d1f4255ec9a1f2a72a083988ff6adee17e279cc

dime10:
> @dime10 @paul0403 So it's because the `apply_pass_plugin(...)` never be covered if we're compiling with `@qjit(target='mlir')`. I add one more test to remove the `target='mlir'` to ensure we cover the failure test. [0d1f425](https://github.com/PennyLaneAI/catalyst/commit/0d1f4255ec9a1f2a72a083988ff6adee17e279cc)

Excellent, thank you :)

rniczh:
@dime10 I think it's because the test only run on every Thursday, so this line will not be covered by codecov
```
+             if EvaluationContext.is_tracing():
```

dime10:
I think that's fine, we do allow exceptions to the coverage where it makes sense. Feel free to merge if everything else looks good and is approved 👌

mehrdad2m (approved):
Thanks @rniczh

doc/releases/changelog-0.15.0.md
```diff
@@ -693,6 +693,10 @@ codes only.
 
 <h3>Bug fixes 🐛</h3>
 
+* Fixed a bug where the `path_to_plugin` never be forwarded in :func:`~.passes.apply_pass_plugin`. The plugin path is now registered with the compiler during tracing.
```
paul0403, line 701: Is this going into v0.15? If so remember to update the target branch of this PR and add to the milestone : )

paul0403 (approved):

doc/releases/changelog-0.15.0.md
rniczh: Sure, thanks :)

PR #2792, opened 2026-05-05
fix changelog render

paul0403:
**Context:**
The changelog section on decomposition renders broken.

image

We fix it.


kipawaa (approved):
It's unfortunate that the RTD renderer fails here, but the github renderer + my local renderer are both fine. Anyways, good catch! Thank you!

sengthai (approved):

PR #2793, opened 2026-05-05
Drop quantum.node from caller in split-multiple-tapes

sengthai:
**Context:**
SplitMultipleTapesPass outlines each tape into its own qnode function and turns the original into a classical wrapper. Per Catalyst's qnode contract the wrapper must drop quantum.node (see [qnode-transform-guide](https://github.com/PennyLaneAI/catalyst/blob/main/frontend/catalyst/python_interface/transforms/qnode-transform-guide.md)).

Downstream, the `resource-analysis` pass then misidentified the empty wrapper as an additional qnode, surfacing as a spurious empty column in `qp.specs` at MLIR levels. The wrapper now drops the attribute, matching `split-non-commuting` and `split-to-single-terms`.

- Remove `quantum.node` in `SpliMultipleTapes.cpp`.

Before fix:
```mlir
module @module_circuit {
    // BUG: wrapper is now classical but still marked `qnode`
    func @circuit() attributes {qnode} {
        a = call @circuit_tape_0()
        b = call @circuit_tape_1()
        return a, b
    }
    func @circuit_tape_0() attributes {qnode} { device; gates_a; a = MeasurementProcessA; device_release; return a }
    func @circuit_tape_1() attributes {qnode} { device; gates_b; b = MeasurementProcessB; device_release; return b }
}

```
After fix:

```mlir
module @module_circuit {
    // wrapper is now classical, no `qnode`
    func @circuit() {
        a = call @circuit_tape_0()
        b = call @circuit_tape_1()
        return a, b
    }
    func @circuit_tape_0() attributes {qnode} { device; gates_a; a = MeasurementProcessA; device_release; return a }
    func @circuit_tape_1() attributes {qnode} { device; gates_b; b = MeasurementProcessB; device_release; return b }
}
```

[sc-118570]

### DIFF
--- a/doc/releases/changelog-0.15.0.md
+++ b/doc/releases/changelog-0.15.0.md
@@ -697,7 +697,10 @@ codes only.
   the `quantum.node` attribute. Downstream, the `resource-analysis` pass then misidentified
   the empty wrapper as an additional qnode, causing an empty column in `qp.specs` at MLIR levels.
   [(#2793)](https://github.com/PennyLaneAI/catalyst/pull/2793)
-
+* Fixed a bug in the `split-multiple-tapes` pass where the post-split classical wrapper kept
+  the `quantum.node` attribute. Downstream, the `resource-analysis` pass would identify
+  the empty wrapper as an additional qnode, causing `qp.specs` to return invalid results at MLIR levels.
+  [(#2793)](https://github.com/PennyLaneAI/catalyst/pull/2793)
 * Refactored all passes in `catalyst.passes.builtin_passes.py` to be `pennylane.transforms.core.Transform` objects
   rather than decorators. This allows them to be used as standard transforms, enabling full compatibility with
   `pennylane.CompilePipeline`.
--- a/doc/releases/changelog-0.15.0.md
+++ b/doc/releases/changelog-0.15.0.md
@@ -693,6 +693,11 @@ codes only.
 
 <h3>Bug fixes 🐛</h3>
 
+* Fixed a bug in the `split-multiple-tapes` pass where the post-split classical wrapper kept
+  the `quantum.node` attribute. Downstream, the `resource-analysis` pass then misidentified
+  the empty wrapper as an additional qnode, causing an empty column in `qp.specs` at MLIR levels.
+  [(#2793)](https://github.com/PennyLaneAI/catalyst/pull/2793)
+
 * Refactored all passes in `catalyst.passes.builtin_passes.py` to be `pennylane.transforms.core.Transform` objects
   rather than decorators. This allows them to be used as standard transforms, enabling full compatibility with
   `pennylane.CompilePipeline`.
--- a/mlir/lib/Quantum/Transforms/SplitMultipleTapes.cpp
+++ b/mlir/lib/Quantum/Transforms/SplitMultipleTapes.cpp
@@ -323,6 +323,11 @@ struct SplitMultipleTapesPass : public impl::SplitMultipleTapesPassBase<SplitMul
                 outlinedfunc->moveAfter(func);
                 outlinedfunc->setAttrs(OutlinedFuncAttrs);
             }
+
+            // The caller function is now a classical wrapper calling the outlined
+            // tape functions, so it must drop `quantum.node`
+            // (matches split_non_commuting and split_to_single_terms).
+            func->removeAttr("quantum.node");
         }
     } // runOnOperation()
 };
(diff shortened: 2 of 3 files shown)


mlir/lib/Quantum/Transforms/SplitMultipleTapes.cpp
```diff
@@ -323,6 +323,11 @@ struct SplitMultipleTapesPass : public impl::SplitMultipleTapesPassBase<SplitMul
                 outlinedfunc->moveAfter(func);
                 outlinedfunc->setAttrs(OutlinedFuncAttrs);
             }
+
+            // The caller function is now a classical wrapper calling the outlined
+            // tape functions, so it must drop `quantum.node`
+            // (matches split_non_commuting and split_to_single_terms).
```
jzaia18, line 329: Would it be good to link [this page](https://github.com/PennyLaneAI/catalyst/blob/main/frontend/catalyst/python_interface/transforms/qnode-transform-guide.md) in the comment here?

doc/releases/changelog-0.15.0.md
```diff
@@ -693,6 +693,11 @@ codes only.
 
 <h3>Bug fixes 🐛</h3>
 
+* Fixed a bug in the `split-multiple-tapes` pass where the post-split classical wrapper kept
+  the `quantum.node` attribute. Downstream, the `resource-analysis` pass then misidentified
+  the empty wrapper as an additional qnode, causing an empty column in `qp.specs` at MLIR levels.
+  [(#2793)](https://github.com/PennyLaneAI/catalyst/pull/2793)
+
```
AntonNI8:
```suggestion
* Fixed a bug in the `split-multiple-tapes` pass where the post-split classical wrapper kept
  the `quantum.node` attribute. Downstream, the `resource-analysis` pass would identify
  the empty wrapper as an additional qnode, causing `qp.specs` to return invalid results at MLIR levels.
  [(#2793)](https://github.com/PennyLaneAI/catalyst/pull/2793)
```

AntonNI8 (approved):
Thanks for the quick fix @sengthai! 🚀

doc/releases/changelog-0.15.0.md
paul0403: I actually don't think we need to explicitly link this to the specs passes, since this is a valid bug fix in itself, even in the sense of just better conforming to the catalyst quantum pass guidelines. The current wording also makes it sound like `resource-analysis` is a pass that is always run, which is not true. 

We can just describe the bug fix as its own. If we really want to describe its impact with specs, we can use specs as an example.

mlir/lib/Quantum/Transforms/SplitMultipleTapes.cpp
paul0403: I don't think we need to link, there's plenty of places where we do this.

paul0403 (approved):
Nice! Thanks for catching this!

I have some comments but all non blocking 👍

mlir/lib/Quantum/Transforms/SplitMultipleTapes.cpp
sengthai: > I don't think we need to link, there's plenty of places where we do this.
Yes I do agree.

doc/releases/changelog-0.15.0.md
```diff
@@ -693,6 +693,14 @@ codes only.
...
 <h3>Bug fixes 🐛</h3>
 
+* Fixed a bug in the `split-multiple-tapes` pass where the post-split classical wrapper kept
+  the `quantum.node` attribute. Downstream, the `resource-analysis` pass then misidentified
+  the empty wrapper as an additional qnode, causing an empty column in `qp.specs` at MLIR levels.
+  [(#2793)](https://github.com/PennyLaneAI/catalyst/pull/2793)
+* Fixed a bug in the `split-multiple-tapes` pass where the post-split classical wrapper kept
+  the `quantum.node` attribute. Downstream, the `resource-analysis` pass would identify
+  the empty wrapper as an additional qnode, causing `qp.specs` to return invalid results at MLIR levels.
+  [(#2793)](https://github.com/PennyLaneAI/catalyst/pull/2793)
```
paul0403, line 703: @sengthai I think you accidentally duplicated the entry

paul0403: Deduplicating it for you in #2792

AntonNI8: @paul0403 I'll fix this in the Catalyst release notes PR, good catch.

AntonNI8: #2794

paul0403: I already fixed it in #2792 ! I was fixing some changelog rendering there anyway

PR #2798, opened 2026-05-06
Update iterator to vector in DOS to avoid gcc14 stdlib strictness changes

mlxd:
**Context:** Fixes an incompatibility between LLVM iterator and GCC14+ stdlib. Missing default ctor's cause a failure due to changes in strictness. 

**Description of the Change:** Create a vector instead of using an iterator for the DOS.

**Benefits:** Will allow comp with newer GCC/stdlib variants on Linux machines (vetted against gcc14.3)

**Possible Drawbacks:** Potential for perf change in DOS due to vec creation instead of iterator use. Unlikely to show up in practice though.

**Related GitHub Issues:**


mlxd:
Do not merge until the RC are created

PR #2799, opened 2026-05-06
Detect MCMs nested in higher order primitives

### LINKED ISSUE #2382: Dynamic one-shot execution with "hw-like" post-selection does not filter out shots with MCM in higher-order operator
In addition to issue #2381:
```py
@qjit
@qml.qnode(
    qml.device("lightning.qubit", wires=1),
    shots=10,
    mcm_method="one-shot",
    postselect_mode="hw-like",
)
def circuit(b: bool):
    qml.H(0)

    @cond(b)
    def nested():
        measure(0, postselect=1)
    nested()

    return qml.sample(wires=0)

print(circuit(True))
```
produces unfiltered shots
```
[[0]
 [1]
 [0]
 [0]
 [0]
 [0]
 [0]
 [1]
 [1]
 [0]]
```
whereas
```py
@qjit
@qml.qnode(
    qml.device("lightning.qubit", wires=1),
    shots=10,
    mcm_method="one-shot",
    postselect_mode="hw-like",
)
def circuit(b: bool):
    qml.H(0)

    measure(0, postselect=1)

    return qml.sample(wires=0)

print(circuit(True))
```
produces correctly filtered shots
```
[[-2147483648]
 [          1]
 [-2147483648]
 [-2147483648]
 [          1]
 [-2147483648]
 [          1]
 [          1]
 [          1]
 [          1]]
```

paul0403:
**Context:**
In legacy frontend, when detecting whether a quantum tape has MCMs, the check was a simple top-level (pseudocode)
```
if any(op is MCM for op in tape.ops)
``` 
This means any nested MCMs when `op` is a higher-order primitive will not be detected, hence failing to trigger any MCM processing!

Recursively check for MCMs acorss all nested tapes.

**Benefits:**
Correctness.

**Related GitHub Issues:**
Also needs PennyLane PR https://github.com/PennyLaneAI/pennylane/pull/9431
closes #2382
[sc-107747]


paul0403:
@kipawaa I'm basing this off your branch that propages mcm config into higher-order primtitives, since it needs that

dime10:
Is this sufficient to execute one-shot correctly with nested mcms? I originally imagined this would be quite difficult because the one-shot method relies on returning mcm values from the qnode, so you'd have to propagate the results from inner scopes to the outer one?

paul0403:
> Is this sufficient to execute one-shot correctly with nested mcms? I originally imagined this would be quite difficult because the one-shot method relies on returning mcm values from the qnode, so you'd have to propagate the results from inner scopes to the outer one?

The PL one-shot method is also missing the collection of nested MCMs, which I'm fixing as well https://github.com/PennyLaneAI/pennylane/pull/9431

With these two branches, and slightly modifying the circuit in the original issue (to make the conditional return the mcm), the IR and results do look correct:

```python
@qml.qjit
@qml.qnode(
    qml.device("lightning.qubit", wires=2),
    shots=10,
    mcm_method="one-shot",
    postselect_mode="hw-like",
)
def circuit(b: bool):
    qml.H(0)

    @cond(b)
    def nested():
        _ = measure(0, postselect=1)
        return _

    @nested.otherwise
    def cond_else():
        return 1

    m = nested()
    qp.RX(m*0.1, wires=1)

    return qml.sample(wires=0)

print(circuit.mlir)
print(circuit(True))
```

```
....
      %3:2 = scf.if %extracted -> (tensor<i64>, !quantum.reg) {
        %13 = quantum.extract %2[ 0] : !quantum.reg -> !quantum.bit
        %mres, %out_qubit = quantum.measure %13 : i1, !quantum.bit
        %from_elements = tensor.from_elements %mres : tensor<i1>
        %14 = quantum.insert %2[ 0], %out_qubit : !quantum.reg, !quantum.bit
        %15 = stablehlo.convert %from_elements : (tensor<i1>) -> tensor<i64>
        scf.yield %15, %14 : tensor<i64>, !quantum.reg
      } else {
        scf.yield %c, %2 : tensor<i64>, !quantum.reg
      }
      %4 = stablehlo.convert %3#0 : (tensor<i64>) -> tensor<f64>
      %5 = stablehlo.multiply %4, %cst : tensor<f64>
      %6 = quantum.extract %3#1[ 1] : !quantum.reg -> !quantum.bit
      %extracted_0 = tensor.extract %5[] : tensor<f64>
      %out_qubits_1 = quantum.custom "RX"(%extracted_0) %6 : !quantum.bit
....

```

```
[[          1]
 [          1]
 [          1]
 [-2147483648]
 [-2147483648]
 [-2147483648]
 [          1]
 [          1]
 [          1]
 [-2147483648]]
```

dime10:
> The PL one-shot method is also missing the collection of nested MCMs, which I'm fixing as well [PennyLaneAI/pennylane#9431](https://github.com/PennyLaneAI/pennylane/pull/9431)

To be fair, operators with tapes attached to them is not a thing in PennyLane, I don't think we can fault them for that. Any functionality related to that was implemented in the Catalyst frontend I believe. 

> With these two branches, and slightly modifying the circuit in the original issue (to make the conditional return the mcm), the IR and results do look correct:

Nice! But unfortunately there is no guarantee that the user will choose to return the mcm result from their nested scope.

I think it would be useful to assess the following before proceeding:

- how much work is it to get this working in the current frontend, considering the above points?
- does this work with the new one-shot pass and program capture?

paul0403:
> To be fair, operators with tapes attached to them is not a thing in PennyLane, I don't think we can fault them for that. 

Oh yes I remember now, it was us who exposed things like `Cond` as "Operator"s.


> Nice! But unfortunately there is no guarantee that the user will choose to return the mcm result from their nested scope.

Oh, I was just responding to your comment regarding returning the mcm from inner scope. Not returning MCM (i.e. the code snippet in the issue as-is) works too.


> * how much work is it to get this working in the current frontend, considering the above points?

I actually think it's already working. 


> * does this work with the new one-shot pass and program capture?

Capture can't handle hw-like post select mode yet, but the mlir one shot pass can handle control flow no problem.

paul0403:
> the one-shot method relies on returning mcm values from the qnode, so you'd have to propagate the results from inner scopes to the outer one

I think this is exactly what I did in https://github.com/PennyLaneAI/pennylane/pull/9431! i.e. collecting MCMs in nested tapes

dime10:
> > the one-shot method relies on returning mcm values from the qnode, so you'd have to propagate the results from inner scopes to the outer one
> 
> I think this is exactly what I did in [PennyLaneAI/pennylane#9431](https://github.com/PennyLaneAI/pennylane/pull/9431)! i.e. collecting MCMs in nested tapes

I guess we should consider whether we want to introduce this concept of "nested tapes" to PennyLane given that one, it doesn't exist there yet, and two, we're getting rid of it in Catalyst this year 🤔

paul0403:
Discussed offline, this is deprioritized.

PR #2801, opened 2026-05-06
Fix build failures with clang + GCC ≤ 13 libstdc++ due to broken range views

mehrdad2m:
**Context:**
`DefaultPipelines.h` recent changes use `std::views::filter` and `std::views::transform` via the | pipe syntax which fails when compiling from source using clang with GCC 12 or GCC 13's `libstdc++`.

replaced those methods with `std::copy_if` and  `std::transform` from `<algorithm>` to dodge these failures.  

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-119527]


paul0403 (approved):

kipawaa (approved):
Nice fix, thanks!

PR #2803, opened 2026-05-06
Fixing PPR decomposition in `qml.transforms.decompose`

lazypanda10117:
Adding support for decomposing PauliRot using the (python-based) graph decomposition system. Previously, the following example would not work, leaving the PauliRot op unchanged in the IR.

```python
qp.capture.enable()
qp.decomposition.enable_graph()

@qp.qjit(keep_intermediate=2)
@qp.decompose(gate_set=[qml.RZ, qml.MultiRZ, qml.RX, qml.GlobalPhase])
@qp.qnode(qml.device("lightning.qubit", wires=1))
def circuit():
  qp.PauliRot(0.5, pauli_word="X", wires=[0])
  return qp.expval(qml.Z(0))

print(circuit.mlir_opt)
```

This PR also re-enables skipped tests that previously failed due to PauliRot decomposition.

[[sc-119011]]


mlir/lib/Quantum/Transforms/decompose_lowering.cpp
```diff
@@ -109,6 +119,16 @@ struct DecomposeLoweringPass : impl::DecomposeLoweringPassBase<DecomposeLowering
 
                     decompositionRegistry[newTargetOpStr] = func;
                 }
+                else if (targetOp == "PauliRot") {
+                    // Create a new target op name with the number of wires and the Pauli word
+                    StringRef pauliWord = DecompUtils::getPauliWord(func);
+                    if (!pauliWord.empty()) {
+                        std::string newTargetOpStr =
+                            targetOp.str() + "_" + std::to_string(DecompUtils::getNumWires(func)) +
```
kipawaa, line 127: Do we need the number of wires if we already have the pauliword? I have no issue with this, it just seems redundant.

mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
```diff
@@ -91,6 +91,94 @@ struct DLCustomOpPattern : public OpRewritePattern<CustomOp> {
...
+            return failure();
+        }
+
+        // Find the corresponding decomposition function for the op
+        auto numQubits = op.getInQubits().size();
+        std::string pauliWord;
+        for (auto pauli : op.getPauliProduct()) {
+            pauliWord += cast<StringAttr>(pauli).getValue();
+        }
+        std::string lookupKey = gateName + "_" + std::to_string(numQubits) + "_" + pauliWord;
```
kipawaa, line 124: It might be nice to move this name generation to a helper function to standarize and reduce code duplication here and [here](https://github.com/PennyLaneAI/catalyst/pull/2803/changes#diff-d5cea81e148b8f765ef08a837addfb7461a27b76e6dfc7c95dc5e0dba7f30adfR127).

frontend/test/lit/test_decomposition.py
```diff
@@ -1013,6 +991,36 @@ def circuit_20(x: float):
 test_decompose_lowering_multirz()
 
 
+def test_decompose_lowering_pauli_rot():
```
kipawaa, line 1020: Could you include a `CHECK` for the required decomp rule attributes here?

frontend/test/lit/test_decomposition.py
kipawaa, line 1: I think it would be nice to test a longer Pauli word, a Pauli word containing an identity gate, and perhaps with out-of-order qubits as well (either here or in the MLIR lit tests).

kipawaa (changes requested):
Great to see this support added! It looks good, I'd just like to see a couple more tests

mlir/lib/Quantum/Transforms/decompose_lowering.cpp
lazypanda10117: Yes you are right, I can change that.

maliasadi:
@lazypanda10117 please create a PR targeting the rc branch to raise an error :bug:

mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
lazypanda10117: I think this is pretty clear and there's only two usage of this pattern. This also follows how the other decompose lowering pattern works.

frontend/test/lit/test_decomposition.py
lazypanda10117: What do you mean by "with out-of-order qubits"?

mlir/lib/Quantum/Transforms/decompose_lowering.cpp
lazypanda10117: This actually comes from here: https://github.com/PennyLaneAI/catalyst/blob/lazypanda10117/ppr-bug-fix/frontend/catalyst/from_plxpr/decompose.py#L343-L345

I am not sure if we want to remove this logic because the current naming would be more consistent with the other type of multi-wires ops.

frontend/test/lit/test_decomposition.py
kipawaa: something with (for example) `wires=[0, 2, 1]` to ensure that the qubits are re-ordered correctly

frontend/test/lit/test_decomposition.py
kipawaa: [1d58612](https://github.com/PennyLaneAI/catalyst/pull/2803/commits/1d586127015d8fa23da1756b5534ed351fcdafdb)

lazypanda10117: thanks!

frontend/test/lit/test_decomposition.py
lazypanda10117: Done cf1f19d00ac3e1da688c2aa6d1b813e0cd870593

kipawaa (approved):
Great, thank you!

PR #2805, opened 2026-05-07
Fix xdsl void returning funcs

mehrdad2m:
**Context:**

When an xDSL pass is applied to a function with no return value, compilation crashes with an LLVM assertion in FuncToLLVM.cpp:
```                                                                       
assert(!resAttrDicts.empty() && "expected array to be non-empty")
```            
The reason this is triggered is that [JAX unconditionally sets result_attrs for void functions](https://github.com/jax-ml/jax/blob/aafe16fcd0276daa5d54bd540b2a999a9824c14e/jax/_src/interpreters/mlir.py#L1862-L1863) when when `Compiler.run` serializes the JAX module to generic text via `get_asm(print_generic_op_form=True)`. Note that when no xdsl pass is applied `get_asm` prints in assembly format which indeed gets rid of the result_attrs.
However xDSL's generic parser preservesthe empty array through the round-trip in generic format and the generic printer then emits it into the IR fed to the C++ driver, where FuncToLLVM eventually asserted on it.                                           

walk the IR and strip the empty res_attr out. Note that printing assembly format was used as an initial fix but turns out that xDSL's assembly printer generates SSA naming that MLIR's C++ parser cannot re-parse for programs with many block arguments.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-113431]


mehrdad2m:
Seems like assembly format cannot be used as a general fix here because xDSL's assembly printer generates SSA naming that MLIR's C++ parser cannot re-parse for programs with many block arguments. Will walk the IR and strip the empty res_attr out.

rniczh (approved):

frontend/catalyst/python_interface/compiler.py
```diff
@@ -67,7 +67,15 @@ def run(
...
-        # Convert back to string
+        # JAX serialises void func.func ops with `res_attrs = []` in generic form
+        # triggering an assertion in FuncToLLVM lowering.
+        # Remove empty arrays in-place so the generic printer omits them.
+        for op in xmod.walk():
+            for key in ("res_attrs", "arg_attrs"):
+                val = op.properties.get(key)
+                if isinstance(val, ArrayAttr) and len(val) == 0:
+                    del op.properties[key]
+
```
maliasadi: Only `func.func` ops carry `res_attrs`/`arg_attrs`. right?

What do you think about the following suggestion?


```suggestion
        from xdsl.dialects.func import FuncOp

        for op in xmod.walk():
            if not isinstance(op, FuncOp):
                continue
            for key in ("res_attrs", "arg_attrs"):
                val = op.properties.get(key)
                if isinstance(val, ArrayAttr) and len(val) == 0:
                    del op.properties[key]
```

maliasadi (approved):
The fix seems straightforward! Just one single performance-related comment 1️⃣

frontend/catalyst/python_interface/compiler.py
mehrdad2m: Yeah, that's a good suggestion. Thanks @maliasadi

PR #2811, opened 2026-05-08
[WIP] Add runtime_call for pre-compiled external kernels

mehrdad2m:


frontend/catalyst/python_interface/transforms/qecp/qec_code_lib.py
```diff
@@ -14,6 +14,7 @@
 
 """This module contains a library of QEC codes."""
 
+import math
```
mehrdad2m:
```suggestion
```

PR #2735, opened 2026-04-23
Add pattern for qecl.measure to convert-qecl-to-qecp pass

joeycarter:
**Context:** The `convert-qecl-to-qecp` dialect-conversion pass requires a rewrite pattern for `qecl.measure` operations.

**Description of the Change:** This pass adds the `MeasureOpConversion` rewrite pattern to the `convert-qecl-to-qecp` dialect-conversion pass. This pattern replaces `qecl.measure` ops with a call to a subroutine that performs a transversal (destructive) measurement of each QEC physical qubit in the codeblock. It then inserts a `qecp.decode_physical_meas` op to decode the _n_ physical measurements into the equivalent _k_ logical measurements. This pattern is currently limited to QEC codes with _k_ = 1.

**Example:**

Before:

```mlir
builtin.module {
  func.func @test_program() {
    %0 = "test.op"() : () -> !qecl.codeblock<1>
    %mres0, %1 = qecl.measure %0[0] : i1, !qecl.codeblock<1>
    %mres1 = "test.op"(%mres0) : (i1) -> i1
    %2 = "test.op"(%1) : (!qecl.codeblock<1>) -> !qecl.codeblock<1>
    func.return
  }
}
```

After (using Steane code):

```mlir
builtin.module {
  func.func @test_program() {
    %0 = "test.op"() : () -> !qecp.codeblock<1 x 7>
    %1, %2 = func.call @measure_transversal_Steane(%0) : (!qecp.codeblock<1 x 7>) -> (tensor<7xi1>, !qecp.codeblock<1 x 7>)
    %3 = qecp.decode_physical_meas %1 : tensor<7xi1> -> tensor<1xi1>
    %4 = arith.constant 0 : index
    %mres0 = tensor.extract %3[%4] : tensor<1xi1>
    %mres1 = "test.op"(%mres0) : (i1) -> i1
    %5 = "test.op"(%2) : (!qecp.codeblock<1 x 7>) -> !qecp.codeblock<1 x 7>
    func.return
  }
  func.func private @measure_transversal_Steane(%0: !qecp.codeblock<1 x 7>) -> (tensor<7xi1>, !qecp.codeblock<1 x 7>) {
    %1 = tensor.empty() : tensor<7xi1>
    %2 = arith.constant 0 : index
    %3 = arith.constant 7 : index
    %4 = arith.constant 1 : index
[...]

PR #2813, opened 2026-05-08
Improve lowering of the `qecl.measure` op in the `convert-qecl-to-qecp` pass

joeycarter:
**Context:**

In #2735, we added a rewrite pattern to the `convert-qecl-to-qecp` pass that converts `qecl.measure` operations to transversal measurements over all qubits in the physical codeblock, packs the _n_ measurement results into a `tensor`, and then inserts a `qecp.decode_physical_meas` op that acted as a placeholder for a runtime function that decodes the _n_ physical measurements into the corresponding _k_ logical measurements.

However, the `qecp.decode_physical_meas` op, as currently implemented, is limited in how it can interpret the physical measurements and decode them. Specifically, it must know the QEC code configuration in order to correctly interpret the _n_ physical measurements, but this information is not passed to the op.

Instead of updating the `qecp.decode_physical_meas` op, we can insert the operations that perform the physical-measurement decoding directly into the IR, since these instructions are reasonably straightforward (for CSS codes, the logical measurement in the computational basis is the XOR product all the measurement results from the qubits acted upon by the logical Pauli Z observable).

This PR removes the insertion of the `qecp.decode_physical_meas` op in the `qecl.measure` rewrite pattern in the `convert-qecl-to-qecp` pass, and instead inserts a call to a subroutine, `decode_physical_measurements_<code_name>`, which performs the physical-measurement decoding. This subroutine pulls the qubit indices in the physical codeblock that correspond to the logical Pauli Z observable from the `qec_code` pass option. This change now puts a hard requirement that the logical Pauli Z gate op be defined in the `transversal_1q_gates` field of the `QecCode` dataclass.

[sc-119006]


frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -512,6 +521,77 @@ def create_measure_subroutine(self) -> func.FuncOp:
...
+            f"'{self.qec_code}' does not specify a logical Pauli Z operation"
+        )
+
+        if pauli_z_gate_data is None:
+            raise CompileError(err_msg)
+
+        _, pauli_z_indices = pauli_z_gate_data
+
+        if len(pauli_z_indices) == 0:
+            raise CompileError(err_msg)
```
joeycarter, line 628: I'm not 100% satisfied with putting this error-checking here: if a program contains no measurements (like in a number of our unit tests), it would still require the logical Pauli Z gate to be defined in the `QecCode` options, even though this subroutine would never be called. But, this might be a non-issue in "real" workloads when all the QEC code information should be well-defined.

lillian542: As the `QecCode` class has been growing, I've been thinking we might want to add something like a `create_generic_qec_code` test fixture that accepts the same inputs, but has default values for everything (or default behaviours based on the values of n  and k). In this case, there will always be an index 0, so the default could just be `{"z": [0]}`.

It's increasingly irritating to make a generic code for a test -  it would be nice to just say `qec_code=create_generic_qec_code(n=10, name="test1")` and not worry about gate dictionaries or pretend tanner graphs for unit tests 😅

joeycarter: I think that's a good idea, I can add it in this PR.

joeycarter: Done

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
multiphaseCFD, line 450: Would it possible to add a pattern to remove `unrealized_cast_op` in the IR?

multiphaseCFD (commented):
Thanks @joeycarter ! Looks good to me!

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
dime10: I'm not sure of the context, but usually if there are leftover `unrealized_conversion_cast` after your passes ran, that means one side of a conversion wasn't completed or not fully completed. Just deleting it is not usually possible because they exist to change the type of values while you're working on a conversion. For example:
```
// %0 is coming from somewhere that we're not converting at the same time as dialect.op
%1 = dialect.op %0 : tensor<f64>
```
gets converted to:
```
%1 = builtin.unrealized_conversion_cast %0 : tensor<f64> to memref<f64>
dialect.op %1 : memref<f64>  // in place version
```
But at some point the source for `%0` will have to be converted as well, then the `unrealized_conversion_cast` can be eliminated.

If you're talking about a situation like this though:
```
%1 = builtin.unrealized_conversion_cast %0 : tensor<f64> to memref<f64>
%2 = builtin.unrealized_conversion_cast %1 : memref<f64> to tensor<f64>
```
then it should already be covered by existing canonicalization patterns.

multiphaseCFD: The reason I asked is that xdsl threw a related warning for the end-to-end test:
```
reconcile_unrealized_casts.py:75: UserWarning: Unable to remove cast UnrealizedConversionCastOp(%0 = builtin.unrealized_conversion_cast %1 : !qecl.codeblock<1> to !qecp.codeblock<1 x 7>) because it is not unifiable with its uses
  warn(
reconcile_unrealized_casts.py:75: UserWarning: Unable to remove cast UnrealizedConversionCastOp(%0 = builtin.unrealized_conversion_cast %1 : !qecp.codeblock<1 x 7> to !qecl.codeblock<1>) because it is not unifiable with its uses
```

joeycarter: @multiphaseCFD I don't think that warning is related to the `qecl-to-qecp` pass, which doesn't insert any `builtin.unrealized_conversion_cast` ops. We can discuss offline.

multiphaseCFD: Then all good!

multiphaseCFD (approved):

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
```diff
@@ -44,6 +46,33 @@ def fixture_qecl_to_qecp_steane_pipeline():
...
+                "x": (qecp.PauliXOp, list(range(n))),
+                "y": (qecp.PauliXOp, list(range(n))),
+                "z": (qecp.PauliXOp, list(range(n))),
+                "hadamard": (qecp.HadamardOp, list(range(n))),
+                "s": (partial(qecp.SOp, adjoint=True), list(range(n))),
+            },
+            transversal_2q_gates={"cnot": qecp.CnotOp},
+        )
+
+    return _make_qec_code
```
lillian542, line 73: 🤩

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
```diff
@@ -671,26 +595,139 @@ def test_measure(self, run_filecheck, qecl_to_qecp_steane_pipeline):
...
+        //       CHECK:     [[mres_t_1:%.+]] = tensor.insert [[mres0]] into [[mres_t_arg]][[[idx]]] : tensor<7xi1>
+        //       CHECK:     scf.yield [[mres_t_1]], [[cb2]] : tensor<7xi1>, !qecp.codeblock<1 x 7>
+        //       CHECK:   }
+        //       CHECK:   func.return [[mres_t_out]], [[cb_out]] : tensor<7xi1>, !qecp.codeblock<1 x 7>
+        //       CHECK: }
+
+        // CHECK-LABEL: func.func private @decode_physical_measurements_Steane
+        //  CHECK-SAME:     [[in_mres_t:%.+]]: tensor<7xi1>) -> tensor<1xi1> {
+        //       CHECK:   [[c4:%.+]] = arith.constant 4 : index
+        //       CHECK:   [[m4:%.+]] = tensor.extract [[in_mres_t]][[[c4]]] : tensor<7xi1>
```
lillian542, line 619: Sometimes in the IR we have `extract %1[%0]` (where %0 is some constant op), and sometimes we have  `extract %0[3]`. Why, and what are the pros/cons/restrictions on these options?

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
```diff
@@ -671,26 +595,139 @@ def test_measure(self, run_filecheck, qecl_to_qecp_steane_pipeline):
...
+        //       CHECK:   [[c0:%.+]] = arith.constant 0 : index
+        //       CHECK:   [[out_mres_t1:%.+]] = tensor.insert [[xor1]] into [[out_mres_t0]][[[c0]]] : tensor<1xi1>
+        //       CHECK:   func.return [[out_mres_t1]] : tensor<1xi1>
+        //       CHECK: }
         }
         """
         run_filecheck(program, qecl_to_qecp_steane_pipeline)
 
+    def test_measure_toy_code(self, run_filecheck):
+        """Test the lowering pattern for `qecl.measure` ops with a toy QEQ code.
```
lillian542:
```suggestion
        """Test the lowering pattern for `qecl.measure` ops with a toy QEC code.
```

lillian542 (approved):
Looks good @joeycarter, thanks!

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
lillian542: (I don't mean to suggest that we change anything, I'm just curious)

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
joeycarter: Thanks! 🙈

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
joeycarter: Good question. In the case of `tensor.extract` ops, the index (or indices) has to be given as an SSA value. More info in the docs here: https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorextract-tensorextractop

For Catalyst ops that take either an index _value_ or an index _attribute_, generally speaking providing it as an attribute is preferred. In fact, we have canonicalization patterns that do exactly this, e.g.

https://github.com/PennyLaneAI/catalyst/blob/11ba3a2c40bc3f7a2395caf4b67fb001bed1e1e1/mlir/lib/Quantum/IR/QuantumOps.cpp#L128-L151

I think in the regular Catalyst pipeline, the extract/insert indices are initially given as values, since that's what happens when lowered through JAX and stablehlo, but then running `-canonicalize` folds these constant ops into attributes on the extract/insert ops.

lillian542: Thanks!

PR #2814, opened 2026-05-08
Emit error when decomposing PauliRot

lazypanda10117:
Emit an error when decomposing a PauliRot.

Proper fix here: https://github.com/PennyLaneAI/catalyst/pull/2803

### DIFF
--- a/frontend/catalyst/from_plxpr/decompose.py
+++ b/frontend/catalyst/from_plxpr/decompose.py
@@ -276,6 +276,9 @@ def _create_decomposition_rule(
             with a variable number of wires (e.g., MultiRZ, GlobalPhase).
     """
 
+    if op_name == "PauliRot":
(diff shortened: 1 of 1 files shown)


kipawaa (approved):
Thank you! 🚀

paul0403 (commented):
Nice, I'll approve after tests are updated to reflect the error 👍

frontend/catalyst/from_plxpr/decompose.py
```diff
@@ -276,6 +276,9 @@ def _create_decomposition_rule(
             with a variable number of wires (e.g., MultiRZ, GlobalPhase).
     """
 
+    if op_name == "PauliRot":
```
maliasadi, line 279: Could you add a unittest?

PR #2795, opened 2026-05-05
Register all `QuantumGate`s to the MLIR graph

kipawaa:
**Context:**
The current MLIR graph registers only `quantum.custom` operations, meaning that gates like `quantum.paulirot`, `quantum.multirz` etc are ignored. This allows examples like the following:

```python
@qp.qjit
@graph_decomposition(gate_set={qp.X})
@qp.qnode(qp.device("lightning.qubit", wires=2))
def circuit(x: float, y: float):
    qp.PauliRot(0.1, "ZZ", wires=[0, 1])
    return
```

Users would expect this to fail, since a `ZZ` `paulirot` can't be decomposed to `X`, but it passes the graph since it is unregistered.

**Description of the Change:**
Registers all `QuantumGate` operations to the graph.

**Benefits:**
Improved graph behaviour.

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-118714]

PR #2815, opened 2026-05-08
Brace Style in `.clang-format`

kipawaa:
**Context:**
Catalyst's style guide suggests the LLVM/K&R style brace format, but the `.clang-format` config file uses the `stroustrup` style. [This can  lead to confusion](https://github.com/PennyLaneAI/catalyst/pull/2795#discussion_r3203792742) over style preferences.

Similarly, the style guide requires braces over in-line or trailing statements, but doesn't enforce this.

Lastly, this downgrades the `pre-commit` mirror of `clang-format` to match the current `clang-format` version used by the makefile.

Update `.clang-format` brace style to `Attach`, and sets `InsertBraces` to `true` to enforce correct style.

**Benefits:**
LLVM-style brace format, cross-org consistency with [Lightning](https://github.com/PennyLaneAI/pennylane-lightning), and coherence with the style guide.

**Possible Drawbacks:**
Reduced familiarity for those not using auto-formatters.

**Related GitHub Issues:**


zsb-xqc:
Let's think about this some more before committing to it.

zsb-xqc:
@kipawaa please go ahead with this brace style change.

runtime/include/QuantumDevice.hpp
```diff
@@ -119,8 +118,7 @@ struct QuantumDevice {
      *
      * @return `QubitIdType` Qubit ID.
      */
-    virtual auto AllocateQubit() -> QubitIdType
-    {
+    virtual auto AllocateQubit() -> QubitIdType {
```
paul0403, line 122: The only concern I have is changing the `QuantumDevice.hpp` source file, which is the base class (aka device interface) for downstream devices....

Although this shouldn't affect anything, I'm still a bit cautious. We have seen instances in the past where simply reordering the methods in this base class leads to downstream devices breaking, due to changed layout in the compiled object file for the virtual `QuantumDevice` base class. Although that specific problem is not going to happen here, I wouldn't be surprised if some downstream devices build with debug info and suddenly the source info is mismatching what code is on what line :confused: 

That being said, this is just braces, so we should be fine? 🤔 @maliasadi

dime10: Yes, method ordering does affect the vtable of compiled C++ objects, however whitespace should not have any effect on the compiled code.

paul0403 (approved):

mehrdad2m (approved):

PR #2816, opened 2026-05-08
Support PPM execution with legacy frontend

lazypanda10117:
To support pauli measurement using the legacy frontend, we introduce a new function `catalyst.pauli_measure` as a wrapper for pauli measurement due to the legacy frontend's limitations. Users using the legacy frontend should use `catalyst.pauli_measure` instead of `qp.pauli_measure`. 

For example, this change would enable users to execute the following circuit using the legacy frontend.

```python
@qp.qjit(capture=False, autograph=True)
@qp.set_shots(1000)
def circuit():
    m = catalyst.pauli_measure(wires=[0], pauli_word="X")
    if m:
        qp.PauliRot(np.pi / 4, "XX", wires=[0, 1])
    else:
        qp.PauliRot(np.pi / 4, "YY", wires=[0, 1])
    return qp.probs()

print(circuit())
```


frontend/catalyst/jax_tracer.py
```diff
@@ -882,6 +888,27 @@ def bind_native_operation(qrp, op, controlled_wires, controlled_values, adjoint=
...
+                *qubits,
+                op.parameters[0],
+                pauli_word=op.hyperparameters["pauli_word"],
+                qubits_len=len(qubits),
+                params_len=len(op.parameters),
+                ctrl_len=len(controlled_wires),
+                adjoint=adjoint,
+            )
+            qrp.insert(op.wires, qubits2[: len(qubits)])
+        elif isinstance(op, PauliMeasure):
```
dime10: Is this not a `HybridOp`? If so it would be covered by the branch below, similar the regular measure op 🤔

frontend/catalyst/jax_tracer.py
```diff
@@ -882,6 +888,27 @@ def bind_native_operation(qrp, op, controlled_wires, controlled_values, adjoint=
...
+            qubits2 = pauli_rot_p.bind(
+                *qubits,
+                op.parameters[0],
+                pauli_word=op.hyperparameters["pauli_word"],
+                qubits_len=len(qubits),
+                params_len=len(op.parameters),
+                ctrl_len=len(controlled_wires),
+                adjoint=adjoint,
+            )
+            qrp.insert(op.wires, qubits2[: len(qubits)])
```
dime10, line 903: I think controlled_wires need to be reinserted as well? Actually, they weren't extracted either, so I think that's also missing

dime10 (commented):
Thanks @lazypanda10117, with a couple of tests we can merge this in 👍

frontend/catalyst/jax_tracer.py
lazypanda10117: Fixed.

frontend/catalyst/jax_tracer.py
lazypanda10117: Also done!

dime10 (approved):
🎉

lazypanda10117:
The failing CI tests stem from a mismatch in the handling of identity Pauli words in lightning. This will be fixed in another PR in Lightning https://github.com/PennyLaneAI/pennylane-lightning/pull/1390.

frontend/catalyst/device/qjit_device.py
```diff
@@ -263,6 +265,16 @@ def get_qjit_device_capabilities(target_capabilities: DeviceCapabilities) -> Dev
             }
         )
 
+    # Optionally enable runtime-powered Pauli-product mid-circuit measurements
+    if "PauliMeasure" in target_capabilities.operations:
```
Qottmann, line 269: here we should also account for the case when ppm_compilation is in the pipeline (or some other, broader condition) because for the old frontend we need the device preprocessing to accept MidCircuitPauliMeasure regardless of the backend device's native PauliMeasure support.

runtime/lib/backend/null_qubit/null_qubit.toml
```diff
@@ -42,7 +42,8 @@ IsingXY                = { properties = [ "invertible", "controllable", "differe
 IsingYY                = { properties = [ "invertible", "controllable", "differentiable" ] }
 IsingZZ                = { properties = [ "invertible", "controllable", "differentiable" ] }
 MultiRZ                = { properties = [ "invertible", "controllable", "differentiable" ] }
-PauliRot               = { properties = [ "invertible",                                  ] }
+PauliRot               = { properties = [ "invertible", "controllable",                  ] }
+PauliMeasure           = { properties = [                                                ] }
```
Qottmann, line 46: this doesnt seem to cut it because in the old frontend it still gets hung up on the device preprocessing

Qottmann: Actually nevermind, needed to re-make catalyst and now seems to work with null.qubit

frontend/catalyst/device/qjit_device.py
dime10: Unfortunately I don't think we can resolve this in the legacy frontend. I don't want to break "regular" workflows by unconditionally allowing `MidCircuitPauliMeasure` through when a device couldn't execute it, and generate a confusing error somewhere down below. The legacy pipeline is set up in way where device preprocessing and final validation of the program happens in the frontend, and I don't want to compromise that process.

However, for our current workflows this should be fine because both null.qubit and lightning.qubit support this op.

In the upcoming redesign, even device preprocessing will be something that's overridable by advanced users.

Qottmann (approved):

frontend/catalyst/device/qjit_device.py
Qottmann: sounds good! thanks!

frontend/catalyst/api_extensions/quantum_operators.py
```diff
@@ -189,6 +189,60 @@ def reset_fn():
...
+    )
+
+    if postselect is not None:
+        raise NotImplementedError(
+            "Postselection on catalyst.pauli_measure is not supported under qjit."
+        )
+
+    cur_trace = EvaluationContext.get_current_trace()
+    wires_list = list(wires) if isinstance(wires, (list, tuple)) else [wires]
+    if len(pauli_word) != len(wires_list):
```
sengthai, line 228: do you think we should to check if the each of Pauli string is {X, Y, Z} as well?

frontend/catalyst/jax_primitives.py
```diff
@@ -1618,7 +1618,7 @@ def _pauli_rot_lowering(
 # pauli measure operation
 #
 @pauli_measure_p.def_abstract_eval
-def _pauli_measure_abstract_eval(*qubits, pauli_word=None, qubits_len=0, adjoint=False):
```
sengthai, line 1621: why adjoint is there at the first place? 🤔

sengthai (commented):
Thanks @lazypanda10117. Well done!

PR #2817, opened 2026-05-09
Fix failing tests in RC branch

JerryChen97:
**Context:**
rc/rc/rc is failing

Removes an XPASS'ing unit test, as well as another test of a deprecated feature.

**Benefits:**
Unblocks rc/rc/rc

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/frontend/test/pytest/python_interface/inspection/test_mlir_specs.py
+++ b/frontend/test/pytest/python_interface/inspection/test_mlir_specs.py
@@ -598,31 +598,6 @@ class TestMLIRSpecs:
             res = mlir_specs(circ, 0, n)
             assert resources_equal(res, expected)
 
-    def test_tape_transforms(self, skip_preprocess, capture_mode):
-        """Test that tape transforms are handled correctly."""
-        if capture_mode:
-            pytest.xfail("Currently broken with plxpr enabled.")
-        if not capture_mode and skip_preprocess:
-            pytest.skip(reason="skip_preprocess ignored without program capture.")
-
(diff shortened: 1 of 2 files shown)


sengthai (approved):
Thanks for quickly fixing this

frontend/test/pytest/python_interface/inspection/test_mlir_specs.py
jzaia18: Honestly, can we just remove this test entirely? This (internal) feature is no longer used within Catalyst and removing it will avoid further issues

jzaia18 (approved):
Approving so you can merge this when ready. Just don't forget the changelog

JerryChen97:
> Approving so you can merge this when ready. Just don't forget the changelog

Thanks for removing redundant test! In Catalyst we typically dont update log for pure test update

paul0403:
Sorry, what happened with the combine global phase tape transform?

paul0403:
We should fix the test instead of just deleting it.

JerryChen97:
> We should fix the test instead of just deleting it.

OK. I'm not aware of the related context too much. So this fix PR was kinda handed over to @jzaia18 and @sengthai who knew better about it.
My first version of fixing actually did try splitting different cases of the original test and aligning the results with what actually happen for `specs` now. However, that was essentially only my cosmetic change based on superficial behavior observation. Guess you guess better discuss about what should be expected for this removed test. @paul0403

PR #2829, opened 2026-05-12
Move reference semantics MBQC ops to MBQC dialect

paul0403:
**Context:**
When the MBQC reference semantics ops were first added, I added them to the overarching `qref` dialect. 

This was not the best organization. See [discussion in thread](https://github.com/PennyLaneAI/catalyst/pull/2773#discussion_r3168963212)

In particular, 

> We don't have a single dialect for all the existing dialects either just because they share "value semantics". While MLIR has two distinct dialects for tensors (value-sem) and memrefs (memory-sem), the difference is that these dialects specifically focus on those (i.e. what can you do with tensors or memrefs). Domain specific dialects seem either work with both (e.g. linalg can operate on both tensors and memrefs) or chose to explicitly work with only one of those (e.g. stablehlo).

As a domain specific dialect, MBQC is now hosting both value and reference semantics ops. 

Move reference semantics mbqc ops from `qref` dialect to `mbqc` dialect.

**Benefits:**
Better organization.

[sc-117892]


mlir/include/MBQC/IR/MBQCRefOps.td
```diff
@@ -0,0 +1,62 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+include "QRef/IR/QRefTypes.td"
+
+def RefMeasureInBasisOp : MBQC_Op<"ref.measure_in_basis"> {
+    let summary = "A parametric single-qubit projective measurement in an arbitrary basis.";
+    let description = [{
+        Identical to mbqc.measure_in_basis operation, except in reference semantics.
```
joeycarter:
```suggestion
        Identical to the `mbqc.measure_in_basis` operation, except in reference semantics.
```

mlir/include/MBQC/IR/MBQCRefOps.td
```diff
@@ -0,0 +1,62 @@
...
+
+    let assemblyFormat = [{
+        `[` $plane `,` $angle `]` $in_qubit (`postselect` $postselect^)? attr-dict `:` type(results)
+    }];
+}
+
+def RefGraphStatePrepOp : MBQC_Op<"ref.graph_state_prep"> {
+    let summary = "Allocate resources for a new graph state.";
+    let description = [{
+        Identical to mbqc.graph_state_prep operation, except in reference semantics.
```
joeycarter:
```suggestion
        Identical to the `mbqc.graph_state_prep` operation, except in reference semantics.
```

joeycarter (approved):
Thanks @paul0403 💯 

Just had a couple of suggestions to polish up the docs while we're here, otherwise looks good.

mlir/lib/QRef/Transforms/value_semantics_conversion.cpp
mehrdad2m, line 1: Now that we have a mbqc ref operations in a separate dialect, we should ideally have their lowering to semantic implementation under the same dialect folder. I guess this would be challenging without major refactor at this point, but as an idea, maybe we could define an OpInterface (e.g. `ReferenceToValueOpInterface`) which would then allow us to define the lowering in a better place. It may also help with the extensive if-else situation in this file. What do you think?

paul0403: So David and I actually discussed this a bit already https://github.com/PennyLaneAI/catalyst/pull/2773#discussion_r3173982804

Essentially an interface is nice to have but not necessary.

mehrdad2m: I see, yeah I agree that it is not necessary but still a nice code quality improvement to have. All good for now.

mehrdad2m (approved):
Thanks! Don't forget to add changelog!

PR #2830, opened 2026-05-12
Prepare to merge RC 0.15.0 branch to main

sengthai:
**Context:**
Config the upstream version and undo some comment

### DIFF
--- a/.dep-versions
+++ b/.dep-versions
@@ -8,8 +8,8 @@ enzyme=v0.0.238
 
 # For a custom PL version, update the package version here and at
 # 'doc/requirements.txt'
-# pennylane=0.45.0-dev94
+pennylane=0.45.0
 
 # For a custom LQ/LK version, update the package version here and at
 # 'doc/requirements.txt'
-# lightning=0.45.0-dev44
+lightning=0.45.0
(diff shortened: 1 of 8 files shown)


mehrdad2m:
Why do we need to modify the rc branch?  Shouldn't this PR target the main branch instead?

dime10:
> Why do we need to modify the rc branch? Shouldn't this PR target the main branch instead?

I suppose we need to undo some changes before merging back to main. We could create a new branch for that off of the rc, but I think it's okay to modify the rc branch as we'll delete it after merging back to main anyway.

paul0403:
Usually what I do is just wait for the auto action to sync from rc branch back to main tomorrow (aka day after the release); but of course we can do this in this PR as well

sengthai:
I am able to push into rc branch directly. So, here is the way to update the rc branch. Once it merge, then I will open PR of RC branch to merge into main.

mehrdad2m:
I see, yeah I guess as this branch will be deleted, this can work too. Just seemed unnecessary as we have the rc sync PR. Both are fine 👍

sengthai:
Let's stick with release guideline, so I will close this one.

PR #2832, opened 2026-05-13
Adding  'debug.compile_mlir` to support compiling a standalone MLIR module through Catalyst pipeline 

mehrdad2m:
**Context:**
Currently, there was no way to compile and run a standalone MLIR file directly from Python without a full @qjit-decorated function. This could be useful for debugging and also for prototyping an MLIR functionality for which frontend lowering doesn't exist yet. The existing but rather hacky workaround for such use-cases is to use `debug.replace_ir` to create a dummy MLIR and replace the whole file with the desired one.   

This is essentially a QoL improvement where we add `debug.compile_mlir`, which compiles an MLIR string or file through the full Catalyst pipeline and returns a callable CompiledMLIR object. The entry-point function name and output types are supplied explicitly via `func_name` and `result_types`. The result can be called like a regular Python function and passed to `get_compilation_stage` to inspect intermediate IR.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


frontend/catalyst/debug/compiler_functions.py
```diff
@@ -307,3 +318,93 @@ def f(x):
...
+        self.compiled_function = compiled_function
+        self.compiler = compiler
+        self.workspace = workspace
+
+    def __call__(self, *args):
+        return self.compiled_function(*args)
+
+
+@debug_logger
+def compile_mlir(mlir_source, *, func_name, result_types=(), **options):
```
joeycarter:
```suggestion
def compile_mlir(mlir_source, *, func_name, result_types, **options):
```

Thoughts on making `result_types` a required option? When playing around with the `identity` example in the tests, I forgot to provide `result_types=[jax.ShapeDtypeStruct((), np.float64)]`, which led to a segfault when it attempted to use the default empty tuple.

joeycarter (commented):
Wow this is great @mehrdad2m! 🤩 

I think this would be a great debugging feature to have, I'm happy to approve if the rest of the team agrees.

frontend/catalyst/debug/compiler_functions.py
```diff
@@ -307,3 +318,93 @@ def f(x):
...
+        that do not match the compiled function's signature could produce incorrect results
+        or undefined behavior.
+
+    Args:
+        mlir_source (str | pathlib.Path): Path to a ``.mlir`` file, or the raw MLIR text.
+        func_name (str): Name of the entry-point function (without ``@``), e.g. ``"jit_my_func"``.
+            The function must be annotated with ``{llvm.emit_c_interface}``.
+        result_types (Sequence[jax.ShapeDtypeStruct]): Output tensor types, in order.
+            Each element must have ``.shape`` (tuple of ints) and ``.dtype`` (numpy dtype).
+            Defaults to ``()`` for functions with no return values.
```
mehrdad2m:
```suggestion
```

PR #2834, opened 2026-05-13
[qref 3.9] Migrate `handle_decomposition_rule` to reference semantics

paul0403:
**Context:**
Migrate `handle_decomposition_rule` to reference semantics.

Mostly, the change is to just get rid of the `QubitHandler`. This was the machinery that converted reference semantics plxpr to value semantics catalyst jaxpr. But now catalyst jaxpr is also reference semantics, so this conversion in the frontend is not needed anymore.

This makes the code significantly simpler.

[sc-119731]


frontend/catalyst/from_plxpr/qfunc_interpreter.py
```diff
@@ -69,8 +68,6 @@
 from catalyst.utils.exceptions import CompileError
 
 from .qubit_handler import (
-    QubitHandler,
```
paul0403, line 72: YES!!!! :fireworks: :fire:

kipawaa (approved):
Congrats! No more `qubit_handler`! 🎉

frontend/catalyst/from_plxpr/qubit_handler.py
```diff
@@ -1,457 +0,0 @@
-# Copyright 2025 Xanadu Quantum Technologies Inc.
```
rniczh, line 1: Nice

rniczh (approved):
Awesome! @paul0403

PR #2836, opened 2026-05-13
Decompose lowering can handle qreg argument to the rule function being at any position

paul0403:
**Context:**
In `--decompose-lowering` pass, there is an assumption that when a quantum register is present on a decomposition rule function, the register has to be the 0-th (colloquially first) argument to the rule function.

This assumption is unnecessarily strict. Other passes (for example the conversion from reference to value semantics) can choose to insert register Values wherever they want on a funcop.

Implement logic to handle when the qreg is at an arbitrary location in the arg list on the decomp rule functions.

**Benefits:**
Code is more generalized and robust.
Unblocks reference semantics work.


kipawaa (approved):
Nice fix, thanks @paul0403 !

rniczh (approved):
🎉

PR #2837, opened 2026-05-13
FIX `convert-qecl-to-qecp` qjit lowering

multiphaseCFD:
### Before submitting

**Context:**

[sc-119708]

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -359,7 +359,7 @@ class ConvertQecLogicalToQecPhysicalPass(ModulePass):
 
     name = "convert-qecl-to-qecp"
 
-    qec_code: QecCode
+    qec_code: QecCode | str
```
joeycarter:
```suggestion
    qec_code: QecCode
```

There's no great way to do this with `dataclass`es, that I'm aware of, but what I have found works best is to use the type hint for the type that gets _used_ rather than the type that initializes the variable. Otherwise, every time you do, for example

```python
self.qec_code.k
```

A static type checker would give an error like "Attribute `k` is not defined on `str` in union `QecCode | str`"

doc/releases/changelog-dev.md
```diff
@@ -22,6 +22,10 @@
   folder being created and the files printed outside in the main directory.
   [(#2807)](https://github.com/PennyLaneAI/catalyst/pull/2807)
 
+* Resolved a bug in the `qeccycle` subroutine within the `convert_qecl_to_qecp_pass` where the `SSA` values
+  of `scf.YieldOp` were incorrectly returned instead of `scf.ForOp` results.
+  [(#2837)](https://github.com/PennyLaneAI/catalyst/pull/2837)
```
joeycarter: I'd leave this in "Internal changes", personally. This is likely not a bug that a user would have encountered.

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
```diff
@@ -1014,4 +1014,5 @@ def circuit():
             m0 = qp.measure(0)
             return qp.sample([m0])
 
-        run_filecheck_qjit(circuit)
+        with pytest.raises(CompileError, match="error: 'qecp.decode_esm_css' op"):
+            run_filecheck_qjit(circuit)
```
joeycarter: I would hope that an integration test with valid input wouldn't raise an error. Maybe we should leave this as `xfail`.

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
```diff
@@ -981,7 +982,6 @@ def test_nontransveral_ops_ignored(self, run_filecheck):
 
 
 # We can remove this xfail and warning filter once `convert_qecl_to_qecp_pass` is complete
-@pytest.mark.xfail(reason="The `convert_qecl_to_qecp_pass` is incomplete")
 @pytest.mark.filterwarnings("ignore:Unable to remove cast UnrealizedConversionCastOp")
```
joeycarter, line 986: We should also check if we can remove the warning filter.

joeycarter (changes requested):
Thanks @multiphaseCFD! Just a few things first.

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
multiphaseCFD: The warning can be removed after qecp-to-quantum lowering is in

doc/releases/changelog-dev.md
```diff
@@ -44,6 +44,10 @@
   [(#2780)](https://github.com/PennyLaneAI/catalyst/pull/2780)
   [(#2772)](https://github.com/PennyLaneAI/catalyst/pull/2772)
 
+* Resolved a bug in the `qeccycle` subroutine within the `convert_qecl_to_qecp_pass` where the `SSA` values
+  of `scf.YieldOp` were incorrectly returned instead of `scf.ForOp` results. Ensure the pass access QEC information with its name in `str` to ensure the compatibility with the compiler infrastructure.
```
joeycarter:
```suggestion
* Resolved a bug in the QEC-cycle subroutine within the `convert-qecl-to-qecp` pass where the SSA
  values of the `scf.yield` op were incorrectly returned instead of the `scf.for` op results. Also,
  the `qec_code` pass option is now given as a `str` rather than a `QecCode` object to ensure
  compatibility with Catalyst's compiler infrastructure.
```

joeycarter (approved):

lillian542 (approved):

PR #2838, opened 2026-05-13
Support skipping inlining annotated nested modules

mehrdad2m:
**Context:**
Currently `InlineNestedModules` pass inline all the modules. We may wish to not inline a module if it is aimed to be compile for a different target. 
 
skip inlining if a module is annotated with `catalyst.target`

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp
```diff
@@ -513,22 +519,19 @@ struct InlineNestedSymbolTablePass : PassWrapper<InlineNestedSymbolTablePass, Op
...
-            }
-        }
+        symbolTable->walk<WalkOrder::PreOrder>([&](Operation *nested) {
+            if (nested == symbolTable)
+                return WalkResult::advance();
+            // Any module still nested at this point is a catalyst.target module.
+            if (isa<ModuleOp>(nested))
+                return WalkResult::skip();
+            if (!isa<SymbolOpInterface>(nested) || !nested->hasAttr(fullyQualifiedNameAttr))
+                return WalkResult::advance();
```
rniczh:
```suggestion
            if (nested == symbolTable) {
                return WalkResult::advance();
            }
            // Any module still nested at this point is a catalyst.target module.
            if (isa<ModuleOp>(nested)) {
                return WalkResult::skip();
            }
            if (!isa<SymbolOpInterface>(nested) || !nested->hasAttr(fullyQualifiedNameAttr)) {
                return WalkResult::advance();
            }
```

rniczh (commented):
don't forget the braces

rniczh (approved):

mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp
```diff
@@ -395,13 +398,12 @@ struct NestedToFlatCallPattern : public OpRewritePattern<catalyst::LaunchKernelO
...
     {
-        auto found = _map->find(op.getCallee()) != _map->end();
-        if (!found) {
+        auto it = _map->find(op.getCallee());
+        if (it == _map->end()) {
             return failure();
         }
 
-        auto newSymbolRefAttr = _map->find(op.getCallee())->getSecond();
-        rewriter.replaceOpWithNewOp<func::CallOp>(op, newSymbolRefAttr, op.getResultTypes(),
```
paul0403, line 404: 😆

mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp
```diff
@@ -429,6 +432,9 @@ struct CleanupPattern : public RewritePattern {
             if (hasQualifiedName) {
```
paul0403, line 432: What is this change about?

mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp
```diff
@@ -513,22 +519,22 @@ struct InlineNestedSymbolTablePass : PassWrapper<InlineNestedSymbolTablePass, Op
...
-                    auto hasQualifiedName = op.hasAttr(fullyQualifiedNameAttr);
-                    if (!hasQualifiedName)
-                        continue;
-
-                    SymbolRefAttr old = op.getAttrOfType<SymbolRefAttr>(fullyQualifiedNameAttr);
-                    auto symbol = cast<SymbolOpInterface>(op);
-                    SymbolRefAttr _new = SymbolRefAttr::get(symbol);
-                    old_to_new.insert({old, _new});
-                }
+        symbolTable->walk<WalkOrder::PreOrder>([&](Operation *nested) {
```
paul0403: I think Erick left a comment somewhere in this file saying this was an explicit loop instead of a walk on purpose.

paul0403: Where does the extra nestedness com from?

dime10:
I know we want to get this in fast, but have we considered or tried not inlining them at all (at least not until llvm lowering)?

mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp
mehrdad2m: That's fair, the comment in on another loop that I didn't see, but I guess it applies here as well. I'll revert it back to the old loop.

mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/2838/commits/a407c579479171d86cb17911a308a97239b00212

mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp
mehrdad2m: do you mean this
```
if (hasBeenRenamed) {
    op->removeAttr(hasBeenRenamedAttrName);
}
```
if so, this is a housekeeping attribute that is set within the pass and need to be cleaned up at the end. the reason it was not there before was that all the modules we inlined so it would be cleaned up automatically as opposed to now where we'll have nested modules surviving the pass.

dime10 (approved):
👍 

One question: What is the `catalyst.target` naming supposed to represent?

mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp
paul0403: Yes this is what I meant. Thanks for explaining!

mehrdad2m:
> I know we want to get this in fast, but have we considered or tried not inlining them at all (at least not until llvm lowering)?

Yes, It should be possible. However there are some choices on how we want to handle them. One is to simply have a similar inliner pass but at a lower level. The other is to compile each module separately and statically link them afterwards which in fact is supported by https://github.com/PennyLaneAI/catalyst/pull/3033 where you can define local targets and have them statically linked. In theory we should support both but not yet sure which is better for the default path. either way, I think we should do that transition once the rest of the PRs are merged, so the goal here was to not disturb the hot path for now.

mehrdad2m:
> One question: What is the catalyst.target naming supposed to represent?

`catalyst.target` was originally defined to annotate a module to target another architecture. So in the later passes, it is used to identify modules for cross-compilation. However, the naming was never discussed in any detail, so we could change it if needed.

dime10:
> > One question: What is the catalyst.target naming supposed to represent?
> 
> `catalyst.target` was originally defined to annotate a module to target another architecture. So in the later passes, it is used to identify modules for cross-compilation. However, the naming was never discussed in any detail, so we could change it if needed.

Ah so it has a value that specifies the chosen target, makes sense

mehrdad2m:
> Ah so it has a value that specifies the chosen target, makes sense

yes, so can pass things like the target triple or even the lowering pipeline you want run on the module.

PR #2839, opened 2026-05-14
An experimental pattern that lowers qecp hyperregister to quantum

multiphaseCFD:
### Before submitting

**Context:**

[sc-119021]

This PR add an experimental pattern for converting hyperregister related types and operations. 
- The `qecp.alloc` is lowered to a serial of `quantum.alloc` and each  `quantum.alloc` operation allocates a `quantum.reg` for a `codeblock`.
- The `qecp.extract_codeblock`  is then removed from the IR
- The `qecp.insert_codeblock` is replaced with `quantum.dealloc` operations
- The `qecp.dealloc` is removed.
- The encoding for-loop is unrolled with a specific pattern. This indicates that `1-logical` qubit circuit (no for-loop encoding in those IRs) lowering is not supported yet. We can come back to this later if needed.

Note that the current implementation is specifically for the `3-Logical qubit GHZ` circuit only. 

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -437,6 +438,7 @@ def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
                         qec_code=self.qec_code,
                         gate_subroutines=transversal_gate_subroutines,
                     ),
+                    ReconcileUnrealizedCastsPattern(),
```
joeycarter: It was puzzling to me why we need to apply `reconcile-unrealized-casts` within the `convert-qecl-to-qecp` pass when there are no patterns in this pass that insert `builtin.unrealized_conversion_cast` ops. But then I was able to track it down to the `ConvertQECLNoiseOpToQECPNoisePass().apply()` call from within the pass. I think it's *this* pass that is inserting `unrealized_conversion_cast` ops where it shouldn't. I'll need to look into this a bit more, but I don't think we want to add `ReconcileUnrealizedCastsPattern` here.

multiphaseCFD: removed!

>  ConvertQECLNoiseOpToQECPNoisePass().apply() call from within the pass. I think it's this pass that is inserting unrealized_conversion_cast ops where it shouldn't. 

Resolved here [8d6f71f](https://github.com/PennyLaneAI/catalyst/pull/2839/commits/8d6f71f0ee90d848df98dd738030062eb222701f)

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_noise_to_qec_noise.py
joeycarter: We should be able to remove this pass now that we no longer insert `builtin.unrealized_conversion_cast` ops.

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_noise_to_qecp_noise.py
joeycarter: We can probably also remove this warning filter now.

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_noise_to_qec_noise.py
joeycarter, line 272: I'm wondering, maybe we don't want to expose this pass to the user if it's being called from the `convert-qecl-to-qecp` pass. Otherwise, if we wanted to run this pass in isolation, strictly speaking I think we would need codeblock and hyperregister type-conversion patterns in this pass as well, like we currently have in `convert-qecl-to-qecp` pass. Not a blocking issue though since this work is all still experimental.

frontend/catalyst/python_interface/transforms/qecp/convert_qecp_to_quantum.py
```diff
@@ -15,13 +15,20 @@
 """QEC Physical to Quantum dialect conversion.
 
 This module contains the implementation of the xDSL convert-qecp-to-quantum dialect-conversion pass.
+
+Known Limitations:
+* The hyperregister related lowering is experimental and the target is for circuits with 1 more
+logical codeblocks, where there is a loop for encoding each logical codeblocks. It's sufficient for
+the GHZ circuit. We might have to come back to this later.
+
```
joeycarter: Just to match the style in other modules 🙂

Also I changed the wording a bit; do I understand correctly that the hyper-register lowering only works when there is more than one logical codeblock? Since it assumes that there is a loop to unroll?

```suggestion
Known Limitations
-----------------

  * The hyper-register lowering is experimental can only target programs with more than one logical
    codeblock, where there is a loop for encoding each logical codeblock. It's sufficient for the
    GHZ circuit. We might have to come back to this later.
```

frontend/catalyst/python_interface/transforms/qecp/convert_qecp_to_quantum.py
```diff
@@ -213,6 +222,66 @@ def match_and_rewrite(self, op: qecp.MeasureOp, rewriter: PatternRewriter):
...
+        # 3. Verify the callee target name
+        _, call_op, _, _ = block.ops
+        if "encode_zero_" not in call_op.callee.string_value():
+            return False, None
+
+        return True, call_op.callee.string_value()
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, op: scf.ForOp | qecp.ExtractCodeblockOp, rewriter: PatternRewriter):
```
joeycarter: This rewrite pattern doesn't do anything if `op` is a `qecp.extract_block` op, correct?

```suggestion
    def match_and_rewrite(self, op: scf.ForOp, rewriter: PatternRewriter):
```

frontend/catalyst/python_interface/transforms/qecp/convert_qecp_to_quantum.py
```diff
@@ -213,6 +222,66 @@ def match_and_rewrite(self, op: qecp.MeasureOp, rewriter: PatternRewriter):
...
+                op.results[0].replace_all_uses_with(init_hreg)
+
+                n_regs = init_hreg.type.width.value.data
+                num_qubits = init_hreg.type.n.value.data
+
+                for _ in range(n_regs):
+                    reg = quantum.AllocOp(num_qubits)
+                    rewriter.insert_op(reg)
+
+                    encode_subroutine_symbol = self._is_exact_encode_zero_steane_loop(op)[1]
```
joeycarter: You could save a call to `self._is_exact_encode_zero_steane_loop` if you pull them out of the `if isinstance(op, scf.ForOp) ...` block.

frontend/catalyst/python_interface/transforms/qecp/convert_qecp_to_quantum.py
```diff
@@ -221,6 +290,66 @@ class ConvertQecPhysicalToQuantumPass(ModulePass):
...
+        module_op = op
+        for op_ in module_op.walk():
+            if isinstance(op_, func.FuncOp) and "quantum.node" in op_.attributes:
+                qecp_alloc_op = None
+                regs = []
+                reg_idx = 0
+                qecp_ops_to_remove = []
+
+                for quantum_op in op_.walk():
+                    if isinstance(quantum_op, qecp.AllocOp):
```
joeycarter: Maybe a match-case block would be more appropriate here?

frontend/catalyst/python_interface/transforms/qecp/convert_qecp_to_quantum.py
```diff
@@ -221,6 +290,66 @@ class ConvertQecPhysicalToQuantumPass(ModulePass):
...
+                        quantum_op.results[0].replace_all_uses_with(qecp_alloc_op.results[0])
+                    if isinstance(quantum_op, qecp.DeallocOp):
+                        rewriter = pattern_rewriter.PatternRewriter(quantum_op)
+                        rewriter.erase_op(quantum_op)
+
+                for quantum_op in reversed(qecp_ops_to_remove):
+                    rewriter = pattern_rewriter.PatternRewriter(quantum_op)
+                    rewriter.erase_op(quantum_op)
+                # Remove dead code
+                DeadCodeElimination().apply(ctx, op_)
```
joeycarter: `DeadCodeElimination()` is meant as a module pass (meaning that the op you supply to the `apply()` method should be a `module` op). The way you have it, this is applying DCE on every `func` op (that has the `quantum.node` attribute) within the module.

What you have works because `DeadCodeElimination.apply()` just calls `region_dce(op.body)`, so I think you should be able to do this instead:


```suggestion
                region_dce(op_.body)
```

where the import is:

```python
from xdsl.transforms.dead_code_elimination import region_dce
```

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecp_to_quantum.py
```diff
@@ -479,3 +487,111 @@ def test_subroutine_qecp_qubit_conversion(self, run_filecheck):
...
+                    %0 = qecp.alloc() : !qecp.hyperreg<2 x 1 x 7>
+                    %1 = arith.constant 0 : index
+                    %2 = arith.constant 2 : index
+                    %3 = arith.constant 1 : index
+                    %4 = scf.for %5 = %1 to %2 step %3 iter_args(%6 = %0) -> (!qecp.hyperreg<2 x 1 x 7>) {
+                    %7 = qecp.extract_block %6[%5] : !qecp.hyperreg<2 x 1 x 7> -> !qecp.codeblock<1 x 7>
+                    %8 = func.call @encode_zero_Steane(%7) : (!qecp.codeblock<1 x 7>) -> !qecp.codeblock<1 x 7>
+                    %9 = qecp.insert_block %6[%5], %8 : !qecp.hyperreg<2 x 1 x 7>, !qecp.codeblock<1 x 7>
+                    scf.yield %9 : !qecp.hyperreg<2 x 1 x 7>
+                    }
```
joeycarter, line 524: (For readability)

```suggestion
                    %4 = scf.for %5 = %1 to %2 step %3 iter_args(%6 = %0) -> (!qecp.hyperreg<2 x 1 x 7>) {
                        %7 = qecp.extract_block %6[%5] : !qecp.hyperreg<2 x 1 x 7> -> !qecp.codeblock<1 x 7>
                        %8 = func.call @encode_zero_Steane(%7) : (!qecp.codeblock<1 x 7>) -> !qecp.codeblock<1 x 7>
                        %9 = qecp.insert_block %6[%5], %8 : !qecp.hyperreg<2 x 1 x 7>, !qecp.codeblock<1 x 7>
                        scf.yield %9 : !qecp.hyperreg<2 x 1 x 7>
                    }
```

joeycarter (commented):
Thanks @multiphaseCFD, great work! 🚀 

Most of comments are pretty minor, happy to approve once resolved.

frontend/catalyst/python_interface/transforms/qecp/convert_qecp_to_quantum.py
multiphaseCFD: Nice

frontend/catalyst/python_interface/transforms/qecp/convert_qecp_to_quantum.py
multiphaseCFD: Forgot to remove it. Nice catch!

frontend/catalyst/python_interface/transforms/qecp/convert_qecp_to_quantum.py
multiphaseCFD: Done!

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_noise_to_qec_noise.py
multiphaseCFD: Nice catch!

multiphaseCFD: I remember we discussed whether we want to keep it as a separate pass or not before. I believe the conclusion is to keep this pass isolated as it would be removed eventually whether we will be targeting at the noisy device or the real hardware later. Let's come back to it later.

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_noise_to_qecp_noise.py
multiphaseCFD: Done!

frontend/catalyst/python_interface/transforms/qecp/convert_qecp_to_quantum.py
multiphaseCFD: Done! [b3115d6](https://github.com/PennyLaneAI/catalyst/pull/2839/commits/b3115d6c44bcd6f7c400e453367db20e1bbe66f4)

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_noise_to_qec_noise.py
joeycarter: Yes that part I agree with; in other words we should keep the `ConvertQECLNoiseOpToQECPNoisePass` class. My comment was more about whether we can get rid of the `convert_qecl_noise_to_qecp_noise_pass` object, since we shouldn't be applying it on the qnode directly as it gets applied indirectly in the `convert-qecl-to-qecp` pass. For example, I don't see any tests that use the `convert_qecl_noise_to_qecp_noise_pass` object, nor should we have any, since this is not the interface we were aiming to support, so perhaps we can remove it.

joeycarter (approved):
Nice work! 💯 

The comment on `convert_qecl_noise_to_qecp_noise_pass` is non-blocking, although I do think it would be safe (and better) to remove it.

frontend/catalyst/python_interface/transforms/qecp/convert_qecp_to_quantum.py
```diff
@@ -213,6 +223,66 @@ def match_and_rewrite(self, op: qecp.MeasureOp, rewriter: PatternRewriter):
         rewriter.replace_op(op, measure_op)
 
 
+@dataclass(frozen=True)
+class UnrollEncodeLoop(RewritePattern):
+    """Op conversion pattern for removing scf.ForOp with encode operations."""
+
+    def _is_exact_encode_zero_steane_loop(self, for_op: scf.ForOp) -> tuple[bool, str | None]:
```
lillian542: I don't see where this is hardcoded to the Steane code specifically, only to the encode loop

```suggestion
    def _is_encode_zero_loop(self, for_op: scf.ForOp) -> tuple[bool, str | None]:
```

frontend/catalyst/python_interface/transforms/qecp/convert_qecp_to_quantum.py
```diff
@@ -221,6 +291,64 @@ class ConvertQecPhysicalToQuantumPass(ModulePass):
...
+                    match quantum_op:
+                        case qecp.AllocOp():
+                            qecp_alloc_op = quantum_op
+                            qecp_ops_to_remove.append(quantum_op)
+                        case func.CallOp() if "encode_zero_" in quantum_op.callee.string_value():
+                            regs.append(quantum_op.results[0])
+                        case qecp.ExtractCodeblockOp():
+                            qecp_ops_to_remove.append(quantum_op)
+                            quantum_op.codeblock.replace_all_uses_with(regs[reg_idx])
+                            reg_idx += 1
```
lillian542: Could we use the `idx` from the `ExtractCodeblockOp` instead of tracking and incrementing `reg_idx`, like we do with the `InsertcodeblockOp`?

frontend/catalyst/python_interface/transforms/qecp/convert_qecp_to_quantum.py
```diff
@@ -221,6 +291,64 @@ class ConvertQecPhysicalToQuantumPass(ModulePass):
...
+                        case qecp.InsertCodeblockOp():
+                            qecp_ops_to_remove.append(quantum_op)
+                            rewriter = pattern_rewriter.PatternRewriter(quantum_op)
```
lillian542: This is already imported at the top of the file

```suggestion
                            rewriter = PatternRewriter(quantum_op)
```

frontend/catalyst/python_interface/transforms/qecp/convert_qecp_to_quantum.py
```diff
@@ -15,13 +15,21 @@
...
+
+  * The hyper-register lowering is experimental can only target programs with more than one logical
+    codeblock, where there is a loop for encoding each logical codeblock. It's sufficient for the
+    GHZ circuit. We might have to come back to this later.
 """
 
 from dataclasses import dataclass
 from typing import cast
 
+from xdsl import pattern_rewriter
```
lillian542:
```suggestion
```

This is only used to access `PatternRewriter`, which is imported in line 38 below

frontend/catalyst/python_interface/transforms/qecp/convert_qecp_to_quantum.py
```diff
@@ -15,13 +15,21 @@
 """QEC Physical to Quantum dialect conversion.
 
 This module contains the implementation of the xDSL convert-qecp-to-quantum dialect-conversion pass.
+
+Known Limitations
+-----------------
+
+  * The hyper-register lowering is experimental can only target programs with more than one logical
+    codeblock, where there is a loop for encoding each logical codeblock. It's sufficient for the
+    GHZ circuit. We might have to come back to this later.
```
lillian542, line 24: It also doesn't support any control flow that iterates over hyper registers (except for encoding), right?

frontend/catalyst/python_interface/transforms/qecp/convert_qecp_to_quantum.py
lillian542: Also pretty minor, but what would happen if we just make 1 `rewriter` at the top and give it the `module_op`?

lillian542 (commented):
Thanks Shuli! Still going over the tests now, but here are my comments/questions for the code

frontend/catalyst/python_interface/transforms/qecp/convert_qecp_to_quantum.py
multiphaseCFD: yep!

frontend/catalyst/python_interface/transforms/qecp/convert_qecp_to_quantum.py
multiphaseCFD: Nice catch! Removed!

frontend/catalyst/python_interface/transforms/qecp/convert_qecp_to_quantum.py
multiphaseCFD: Done!

frontend/catalyst/python_interface/transforms/qecp/convert_qecp_to_quantum.py
multiphaseCFD: Done!

frontend/catalyst/python_interface/transforms/qecp/convert_qecp_to_quantum.py
multiphaseCFD: yes, updated!

frontend/catalyst/python_interface/transforms/qecp/convert_qecp_to_quantum.py
lillian542: Let's note that here as separate bullet point

lillian542 (approved):
Thanks Shuli, it's awesome to see this working! 🚀

PR #2841, opened 2026-05-14
Remove `mlir_specs` as it is now unused

jzaia18:
[sc-116165]

**Context:**
`mlir_specs` no longer serves as the backend for `qp.specs` since the integration with the resource analysis pass in https://github.com/PennyLaneAI/pennylane/pull/9279. Since this is an undocumented internal function, we will remove it to avoid having dead code within Catalyst.

Removes `mlir_specs` from Catalyst

**Benefits:**
Less dead code sitting in the repo

**Possible Drawbacks:**

**Related GitHub Issues:**


frontend/catalyst/python_interface/inspection/draw.py
```diff
@@ -58,6 +57,10 @@
 _cache_store: dict[Callable, dict[int, tuple[str, str]]] = {}
 
 
+class StopCompilation(Exception):
+    """Custom exception to stop compilation early when the desired specs level is reached."""
```
jzaia18:
```suggestion
    """Custom exception to stop compilation early when the desired level is reached."""
```

paul0403 (approved):
Beautiful 💯 

I know this is not a publically documented feature, so we can probably just add a changelog entry to the internal changes section (as opposed to the breaking changes section)

sengthai (approved):
Thanks @jzaia18

PR #2842, opened 2026-05-14
Test cleanup

JerryChen97:
1. `pytest.raises` use error type as the first positional arg instead of `check=` kwarg. `check` won't work if pytest upgrades to a newer version.
2. `id` has been removed on Pennylane side; caused two of the failures in https://github.com/PennyLaneAI/catalyst/actions/runs/25856613784/job/75975962909
   - [x] test_unsupported_measurement_types_rejected

[sc-119766]


JerryChen97:
Let me split them up into two PR's since fixing `test_prod` requires source code change

JerryChen97:
The state prep failure of LLL turns out to be a pennylane bug. I'm fixing on that side.

paul0403 (approved):

sengthai (approved):

PR #2846, opened 2026-05-14
Ensure constructors of xDSL ops with `idx_attr` produce correct attribute type

joeycarter:
**Context:** Certain xDSL ops with an `idx` attribute, such as

* `quantum.{extract, insert}`
* `qecl.{extract_block, insert_block, measure, <gate>`}
* `qecp.{extract_block, insert_block}`, etc.

can be built with incorrect types. For instance, the `qecl.extract_block` expects an integer attribute of type `index` for its `idx` attribute, however, it can be built with other integer attribute types, such as `i64`. This is normally not an issue when using the custom assembly format, since the attribute type is not displayed and can therefore be correctly interpreted during parsing. However, when using generic MLIR format, ops with the incorrect attribute would fail to parse.

**Description of the Change:** The constructors of these xDSL ops have been updated to ensure the `idx` attributes are constructed with the correct types.

[sc-119832]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -47,6 +47,12 @@
   [(#2780)](https://github.com/PennyLaneAI/catalyst/pull/2780)
   [(#2772)](https://github.com/PennyLaneAI/catalyst/pull/2772)
 
+* The constructors of xDSL ops that accept index attributes have been updated to ensure that the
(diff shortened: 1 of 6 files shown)


paul0403 (approved):
Nice! Thanks for the quick fix 💯

mudit2812 (approved):

PR #2847, opened 2026-05-15
Remove rt_decoder from build targets

joeycarter:
**Context:** PR #2772 added the new LUT decoder runtime lib, `rt_decoder`, as a build target. However, on Linux arm64 builds, this triggers a `stringop-overflow` compiler warning (promoted to error). See here for example: https://github.com/PennyLaneAI/catalyst/actions/runs/25882806007/job/76067179962.

[sc-119884]

**Description of the Change:** This is a difficult compiler warning to debug as it only appears on Linux arm64 builds. Therefore as a stopgap measure while we work on a proper fix, this PR removes `rt_decoder` from the build targets to unblock the wheels builds.


paul0403 (approved):

multiphaseCFD (approved):
Thanks Joey!

PR #2848, opened 2026-05-15
Update tests that were affected by Pennylane deprecation/removal v0.46

JerryChen97:
It's causing trouble after any breaking changes from Pennylane: https://github.com/PennyLaneAI/catalyst/actions/runs/25914582605/job/76167967768#step:25:306
Now it's updated to reflect the actual logic from Pennylane side without assuming mocked base class property will always be used.

Also,
 - [x] bumped the dep versions of Pennylane and Lightning.
 - [x] fixed other tests that are using deprecated `op.basis` API
 - [x] fixed two OQD tests that have fewer decomposed gates now after we improved the `BasisState` decomps

[sc-119833]

### DIFF
--- a/.dep-versions
+++ b/.dep-versions
@@ -8,7 +8,7 @@ enzyme=v0.0.238
 
 # For a custom PL version, update the package version here and at
 # 'doc/requirements.txt'
-pennylane=0.45.0
+pennylane=0.46.0.dev9
 
 # For a custom LQ/LK version, update the package version here and at
 # 'doc/requirements.txt'
--- a/.dep-versions
+++ b/.dep-versions
@@ -8,7 +8,7 @@ enzyme=v0.0.238
 
 # For a custom PL version, update the package version here and at
 # 'doc/requirements.txt'
-pennylane=0.46.0.dev9
+pennylane=0.46.0.dev11
 
 # For a custom LQ/LK version, update the package version here and at
 # 'doc/requirements.txt'
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -32,6 +32,6 @@ lxml_html_clean
 
 # Pre-install PL development wheels
 --extra-index-url https://test.pypi.org/simple/
-pennylane-lightning-kokkos==0.45.0-dev44
-pennylane-lightning==0.45.0-dev44
-pennylane==0.45.0-dev94
+pennylane-lightning-kokkos==0.45.0
+pennylane-lightning==0.45.0
+pennylane==0.46.0.dev9
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -34,4 +34,4 @@ lxml_html_clean
 --extra-index-url https://test.pypi.org/simple/
 pennylane-lightning-kokkos==0.45.0
 pennylane-lightning==0.45.0
-pennylane==0.46.0.dev9
+pennylane==0.46.0.dev11
--- a/frontend/test/pytest/test_adjoint.py
+++ b/frontend/test/pytest/test_adjoint.py
@@ -1082,19 +1082,6 @@ class TestAdjointOperation:
         for angle1, angle2 in zip(angles, reversed(base_angles)):
             assert angle1 == -angle2
 
-    @pytest.mark.parametrize(
-        "base, basis",
-        (
-            (qp.RX(1.234, wires=0), "X"),
-            (qp.PauliY("a"), "Y"),
-            (qp.PhaseShift(4.56, wires="b"), "Z"),
-            (qp.SX(-1), "X"),
-        ),
-    )
-    def test_basis_property(self, base, basis):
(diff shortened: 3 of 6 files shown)


JerryChen97:
~? Are we not using latest PL~ No worry I forgot Catalyst pin PL

JerryChen97:
OK, `BasisEmbedding` fix will be in `dev10` :sweat: it only got merged this morning. Since nothing is like urgent bug let's push it to next week

doc/requirements.txt
```diff
@@ -32,6 +32,6 @@ lxml_html_clean
 
 # Pre-install PL development wheels
 --extra-index-url https://test.pypi.org/simple/
-pennylane-lightning-kokkos==0.45.0-dev44
-pennylane-lightning==0.45.0-dev44
-pennylane==0.45.0-dev94
+pennylane-lightning-kokkos==0.45.0
+pennylane-lightning==0.45.0
+pennylane==0.46.0.dev9
```
JerryChen97:
```suggestion
pennylane==0.46.0.dev11
```

.dep-versions
```diff
@@ -8,7 +8,7 @@ enzyme=v0.0.238
 
 # For a custom PL version, update the package version here and at
 # 'doc/requirements.txt'
-pennylane=0.45.0
+pennylane=0.46.0.dev9
```
JerryChen97:
```suggestion
pennylane=0.46.0.dev11
```

frontend/test/pytest/test_quantum_control.py
```diff
@@ -912,7 +912,7 @@ class DummyOp(Operator):
             _queue_category = value
 
         op = C_ctrl(DummyOp(1), 0)
-        assert op._queue_category == value
+        assert op._queue_category == "_ops"
```
kipawaa, line 908: Maybe I don't understand the changes correctly, but is this not an error? Shouldn't we be (as the test description says) deferring to the op's `_queue_category`?

JerryChen97: Good catch — the hardcoded `"_ops"` does look confusing. Here's the context:

PL PR [#9470](https://github.com/PennyLaneAI/pennylane/pull/9470) (commit `4020a93`) removed the `_queue_category` property from `SymbolicOp`. Previously, `SymbolicOp.__init__` would copy `base._queue_category` into `self.__queue_category`, and its `_queue_category` property would return that cached value — so `Controlled` would "defer" to the base operator.

After that PL change, `SymbolicOp` (and therefore `Controlled`) no longer has its own `_queue_category` property and falls back to `Operator._queue_category`, which unconditionally returns `"_ops"`. Setting `_queue_category = None` is also no longer allowed (raises `ValueError`).

So the test is technically correct: `C_ctrl(anything, 0)._queue_category` is always `"_ops"` now regardless of the base operator. But the parametrization over `("_ops", "_measurements")` is misleading — it suggests we're testing that the value varies with the base, when in fact the assertion is the same hardcoded `"_ops"` either way.

Better options:
1. **Remove the test** — this tests a PL internal property, not Catalyst-specific behavior
2. **Simplify** — drop the parametrize decorator, update the docstring to say something like "Test that `C_ctrl` always returns `'_ops'` for `_queue_category`"

I'd lean toward option 1 since `_queue_category` is explicitly described as a temporary internal solution that shouldn't be relied on externally. What do you think? @kipawaa @paul0403

kipawaa: I think I lean option 2, since it's kind of good to verify that we don't alter PL internal members. That said, I'm happy with either of these solutions :)

frontend/test/pytest/test_adjoint.py
```diff
@@ -1082,19 +1082,6 @@ def test_single_qubit_rot_angles(self):
...
-    @pytest.mark.parametrize(
-        "base, basis",
-        (
-            (qp.RX(1.234, wires=0), "X"),
-            (qp.PauliY("a"), "Y"),
-            (qp.PhaseShift(4.56, wires="b"), "Z"),
-            (qp.SX(-1), "X"),
-        ),
-    )
-    def test_basis_property(self, base, basis):
```
paul0403, line 1094: What happened here?

JerryChen97: https://github.com/PennyLaneAI/pennylane/pull/9476

frontend/test/pytest/test_quantum_control.py
JerryChen97: https://github.com/PennyLaneAI/catalyst/pull/2848/commits/8d262e40c4b6e5d777a462e375f5e8ca47a2f918

kipawaa (approved):
Looks good, thanks @JerryChen97 !

paul0403 (approved):

PR #2852, opened 2026-05-17
Add qecp-to-llvm pass to pipeline

multiphaseCFD:
### Before submitting

**Context:**

[sc-119971] & [sc-111769]

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -16,11 +16,11 @@
 
 <h3>Internal changes ⚙️</h3>
 
-- Update RC nightly builds to read version number from the `_version.py` file 
+* Update RC nightly builds to read version number from the `_version.py` file
   [(#2797)](https://github.com/PennyLaneAI/catalyst/pull/2797)
 
-* The experimental compiler pass `convert-qecl-to-qecp` has been extended to lower 
-  transversal gate operations from the QEC Logical (`qecl`) dialect into the QEC 
+* The experimental compiler pass `convert-qecl-to-qecp` has been extended to lower
+  transversal gate operations from the QEC Logical (`qecl`) dialect into the QEC
   Physical (`qecp`) dialect.
   [(#2776)](https://github.com/PennyLaneAI/catalyst/pull/2776)
 
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -24,11 +24,17 @@
   Physical (`qecp`) dialect.
   [(#2776)](https://github.com/PennyLaneAI/catalyst/pull/2776)
 
+* The strategy to decode physical measurements in the `convert-qecl-to-qecp` pass has been updated
+  to perform the decoding directly in the IR rather than offloading to a pre-compiled runtime
+  function.
+  [(#2813)](https://github.com/PennyLaneAI/catalyst/pull/2813)
+
 <h3>Documentation 📝</h3>
 
 <h3>Contributors ✍️</h3>
 
 This release contains contributions from (in alphabetical order):
 
+Joey Carter,
 Lillian Frederiksen,
 Shuli Shu,
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -9,6 +9,7 @@
   [(#2824)](https://github.com/PennyLaneAI/catalyst/pull/2824)
   [(#2835)](https://github.com/PennyLaneAI/catalyst/pull/2835)
   [(#2839)](https://github.com/PennyLaneAI/catalyst/pull/2839)
+  [(#2849)](https://github.com/PennyLaneAI/catalyst/pull/2849)
 
 
 <h3>Improvements 🛠</h3>
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -33,6 +33,9 @@
 
 <h3>Internal changes ⚙️</h3>
 
+* Fixed ``KeyError`` in autograph when using ``qp.prod`` as a decorator with PennyLane >= 0.45.
+  [(#2844)](https://github.com/PennyLaneAI/catalyst/pull/2844)
+
 * Update RC nightly builds to read version number from the `_version.py` file
   [(#2797)](https://github.com/PennyLaneAI/catalyst/pull/2797)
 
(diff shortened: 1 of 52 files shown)


frontend/catalyst/ftqc/pipelines.py
```diff
@@ -82,3 +82,64 @@ def workload():
...
+
+    # Insert (in-place) the "convert-qecp-to-llvm" pass immediately before the
+    # "convert-quantum-to-llvm" pass in the MLIRToLLVMDialectConversion pipeline
+    insert_pass_before(
+        pipeline, ref_pass="convert-quantum-to-llvm", new_pass="convert-qecp-to-llvm"
+    )
+
+    # Enable array-backed registers via the "convert-quantum-to-llvm" pass
+    quantum_to_llvm_pass = pipeline.index("convert-quantum-to-llvm")
+    pipeline[quantum_to_llvm_pass] = "convert-quantum-to-llvm{use-array-backed-registers=true}"
```
multiphaseCFD: Do we need this? @joeycarter

joeycarter: No I don't think so. We needed array-backed registers for the mbqc pipeline since we were inserting qubits into a register that the qubit did not originate from (e.g. it came from another register or a single-qubit allocation). But we don't have that in the QEC pipeline so it should be safe to remove it.

multiphaseCFD: Done! Removed

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecp_to_quantum.py
```diff
@@ -594,3 +597,6 @@ def ghz():
             return qp.sample([m0, m1, m2])
 
         run_filecheck_qjit(ghz)
+        results = ghz()
+        target_res_set = [np.array([0, 0, 0]), np.array([1, 1, 1])]
+        assert np.all(res in target_res_set for res in results)
```
joeycarter: This numpy syntax looks off to me, since you can't check that an array is in a list of arrays:

```pycon
>>> x = np.array([0, 1, 2])
>>> targets = [np.array([0, 0, 0]), np.array([0, 1, 2])]
>>> x in targets
  File "<python-input-20>", line 1, in <module>
    x in targets
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

Maybe what you want is this:


```suggestion
        assert all(any(np.array_equal(res, target) for target in target_res_set) for res in results)
```

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecp_to_quantum.py
```diff
@@ -549,14 +551,15 @@ class TestQECPassIntegration:
     # pylint: disable=line-too-long
     def test_qec_pass_ghz_integration(self, run_filecheck_qjit):
         """Integration tests for lowering a 3-logical qubit GHZ circuit."""
-        dev = qp.device("null.qubit", wires=3)
+        dev = qp.device("lightning.qubit", wires=3)
```
joeycarter, line 555: This is great that it runs on Lightning, but I'm a little concerned how long the runtime is. It takes about 40 s on my machine, which is a little on the long side for a pytest. That being said, I don't have a good alternative to suggest since this is already a very minimal circuit with only a single shot. Maybe we can add a `@pytest.mark.slow`?

doc/releases/changelog-dev.md
```diff
@@ -11,6 +11,9 @@
   [(#2839)](https://github.com/PennyLaneAI/catalyst/pull/2839)
   [(#2849)](https://github.com/PennyLaneAI/catalyst/pull/2849)
 
+* A new, experimental compiler pipeline `qec_pipeline` has been added to the `ftqc.pipelines` module.
+  [(#2852)](https://github.com/PennyLaneAI/catalyst/pull/2852)
```
joeycarter: I wouldn't say this is a feature we want to advertise to users, can we move this entry to the _Internal changes_ section?

joeycarter (changes requested):
Thanks @multiphaseCFD, it's all coming together! 🤩 

Just a few suggestions to start.

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecp_to_quantum.py
multiphaseCFD: Good one!

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecp_to_quantum.py
```diff
@@ -549,14 +551,15 @@ class TestQECPassIntegration:
...
 
-        @qp.qjit(capture=True, target="mlir")
+        @qp.qjit(capture=True, pipelines=qec_pipeline())
         @convert_qecp_to_quantum_pass
         @convert_qecl_to_qecp_pass(qec_code="Steane", number_errors=1)
         @inject_noise_to_qecl_pass
         @convert_quantum_to_qecl_pass(k=1)
-        @qp.qnode(dev, shots=1, mcm_method="one-shot")
+        @qp.set_shots(1)
```
lillian542, line 562: Is it very slow? Could we make it 10 shots?

lillian542: Oh, I see the above. What if we run it with 10 shots on `null.qubit`, and a single shot on lightning?

lillian542: Just to confirm that everything integrates well with `dynamic-one-shot` and having multiple shots

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecp_to_quantum.py
```diff
@@ -545,12 +545,13 @@ def test_hyperregister_lowering(self, run_filecheck):
         run_filecheck(program, (ConvertQecPhysicalToQuantumPass(),))
 
 
+@pytest.mark.slow
```
joeycarter, line 548: Forgot to mention, you'll also need to add the following line to `pytest.ini`:

```ini
slow: marks tests as slow (deselect with '-m "not slow"')
```

Otherwise, we get a `PytestUnknownMarkWarning: Unknown pytest.mark.slow` warning.

joeycarter (approved):
Looks good! 💯 

Just had one final additional to suppress a pytest warning, but happy to approve.

lillian542 (approved):
Looks good, thanks @multiphaseCFD!

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecp_to_quantum.py
multiphaseCFD: Thanks Joey! Never touch that file before 😢

PR #2853, opened 2026-05-19
Delete plxpr transform integration

albi3ro:
**Context:**

We no longer want to support or develop applying transforms to the plxpr itself. Instead, we want to force all transforms to be applied in MLIR/ XDSL with the capture pipeline.

But we still had some relatively dead code handing around to support them. This removes them.

Deletes some code for handling `Transform.plxpr_transform` in the conversion process from plxpr to catalyst-jaxpr.

**Benefits:**

Prunes out features we don't actually want anymore.

**Possible Drawbacks:**

This was always experimental, so if anyone is wanting it or depending on it, they are asking for this type of thing to happen.

Is a little sad to delete all the code that took so long to develop though.

**Related GitHub Issues:**

[sc-120025]


JerryChen97 (approved):
🚀 🚀 🚀

paul0403 (approved):

PR #2855, opened 2026-05-19
Decompose lowering pass can handle a null decomp rule

paul0403:
**Context:**
The decompose lowering pass makes certain assumptions about the signature of the decomp rule. In particular, it assumes that the signature has either qreg or qubit values.

This is a natural assumption. However, a special kind of decomp rules are the null rules, which is just empty. These rule functions will not have quantum types in their signature. 

In decompose lowering pass, if the decomp rule has no quantum values in the function signature, simply replace qubit operands of the quantum gate op with the qubit results. 

**Benefits:**
Null decomp rule works.

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -14,11 +14,16 @@
   argument is at an arbitrary position in the argument list.
   [(#2836)](https://github.com/PennyLaneAI/catalyst/pull/2836)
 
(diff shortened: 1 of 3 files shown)


kipawaa (approved):
I'm surprised we never ran into this before, thanks for the fix!

rniczh (approved):
💯

PR #2857, opened 2026-05-20
bump PL dev12 and fix test failures

JerryChen97:
Fixing https://github.com/PennyLaneAI/catalyst/actions/runs/26159251434/job/76946605862

## Changes

1. https://github.com/PennyLaneAI/pennylane/pull/9494 changed the mismatch type of `cond` (now, within the same test, `cond` dropped the `AbstractOperator` outvar so comes across number mismatch)
2. use Python `hash()` instead of `op.hash` as the latter is [being deprecated](https://github.com/PennyLaneAI/pennylane/pull/9488)

[sc-120115]

### DIFF
--- a/.dep-versions
+++ b/.dep-versions
@@ -8,7 +8,7 @@ enzyme=v0.0.238
 
 # For a custom PL version, update the package version here and at
 # 'doc/requirements.txt'
-pennylane=0.46.0.dev11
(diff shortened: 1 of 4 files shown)


paul0403 (approved):
Thanks!

astralcai (approved):

PR #2858, opened 2026-05-20
Depth-1: PBC layer partitioning via `disjoint-qubit`

sengthai:
**Context:**
PBC depth analysis was implemented in https://github.com/PennyLaneAI/catalyst/pull/2014. The depth (layer) definition is based on Figure 6 [GoSC](https://arxiv.org/pdf/1808.02892) paper. 

In previous implementation, a PPR(pi/8) operation is assigned to an existing layer if it commutes with all other operations in that layer or acts on non-overlapping (disjoint) qubits. 

Here, we want to disable commutation rule, and only enable operations that acts on non-overlapping qubits in layer.

This PR adds a `disjoint-qubit` flag to `--partition-layers` pass that disables commutation-based merging. With the flag enabled, ops are grouped into the same layer only when they act on disjoint qubits. 

```bash
# Default: merge ops if they commute or act on disjoint qubits
quantum-opt --partition-layers input.mlir
# Disjoint-qubit only: ignore commutation when grouping
quantum-opt --partition-layers=disjoint-qubit input.mlir
```

[sc-120052]


mlir/lib/PBC/Utils/PBCLayer.cpp
```diff
@@ -202,11 +247,11 @@ bool PBCLayer::insert(PBCOpInterface op)
         return true;
     }
 
-    if (commuteToLayer(op)) {
+    // If onlyOnDisjointQubit is true, we only check the disjoint qubit condition
+    if (commuteToLayer(op) && !onlyDisjointQubit) {
```
dime10: More efficient due to Boolean short-circuiting :)
```suggestion
    if (!onlyDisjointQubit && commuteToLayer(op)) {
```

sengthai: Sweet! Thanks

mlir/lib/PBC/Transforms/PartitionLayers.cpp
dime10, line 1: Can you explain the thinking behind these changes here? :)

mlir/include/PBC/Transforms/Passes.td
```diff
@@ -131,6 +139,8 @@ def CountPPMSpecsPass : Pass<"ppm-specs"> {
 
 def PartitionLayersPass : Pass<"partition-layers"> {
     let summary = "Partition layers of Pauli Product Measurement operations.";
+
+    let options = [OnlyDisjointQubitOption];
 }
 
 def TLayerReductionPass : Pass<"reduce-t-depth"> {
```
dime10, line 146: Would the TLayerReductionPass benefit from the same option? I'm assuming it considers commutativity in its algorithm somehow, or not really?

sengthai: Yes, eventually but not as a simple flag reuse. TLayerReductionPass builds and optimizes layers purely via Pauli commutation (normalizePPROps -> commute -> merge).

dime10: That commutation criteria can be restricted to qubit-disjoint commutativity no? I think without the layer optimization the depth calculation seems less useful.

mlir/lib/PBC/Transforms/PartitionLayers.cpp
sengthai: here we just want to split gathering _then_ construct rather then gathering and construct in the same type. The main reason is to make layer grouping reusable for PBC depth analysis. However, this change requires to iterate twice once to group and once to construct.

mlir/include/PBC/Transforms/Passes.td
sengthai: Yes, it can be restricted to qubit-disjoint commutativity as well.

sengthai: We can set restricted TLayerReductionPass by just updating this func
https://github.com/PennyLaneAI/catalyst/blob/70622420cfaed5f3182fdc7b6f99c248bafe94f4/mlir/lib/PBC/Transforms/TLayerReduction.cpp#L44-L45

dime10: Nice, let's do it :)

sengthai: Sure will do after PBC depth analysis. 🫡

dime10 (approved):

mlir/lib/PBC/Utils/PBCLayer.cpp
```diff
@@ -25,6 +26,53 @@ using namespace catalyst::pbc;
 namespace catalyst {
 namespace pbc {
 
+bool isParentLayerOp(PBCOpInterface op)
+{
+    // Skip ops nested inside an existing pbc.layer region
+    auto parentOp = op->getParentOp();
+    while (parentOp != nullptr) {
+        if (isa<LayerOp>(parentOp))
+            return true;
```
rniczh:
```suggestion
        if (isa<LayerOp>(parentOp)) {
            return true;
        }
```

rniczh (approved):
👍  Just a syntax issue.

mlir/lib/PBC/Utils/PBCLayer.cpp
```diff
@@ -25,6 +26,54 @@ using namespace catalyst::pbc;
...
+        if (layer.insert(op, onlyOnDisjointQubit)) {
+            return WalkResult::skip();
+        }
+
+        // insert() can fail on an empty layer (e.g. block ordering); do not push [].
+        if (!layer.empty()) {
+            groups.emplace_back(layer.getOps());
+        }
+        layer = PBCLayer(this);
+        layer.insert(op);
```
lazypanda10117: `layer.insert(op, onlyOnDisjointQubit)`

mlir/lib/PBC/Utils/PBCLayer.cpp
```diff
@@ -25,6 +26,54 @@ using namespace catalyst::pbc;
...
+    root->walk([&](PBCOpInterface op) {
+        if (isParentLayerOp(op)) {
+            return WalkResult::skip();
+        }
+
+        if (layer.insert(op, onlyOnDisjointQubit)) {
+            return WalkResult::skip();
+        }
+
+        // insert() can fail on an empty layer (e.g. block ordering); do not push [].
```
lazypanda10117: I think this is dead code.

mlir/lib/PBC/Utils/PBCLayer.cpp
```diff
@@ -25,6 +26,54 @@ using namespace catalyst::pbc;
...
+// Partition PBC ops into layer groups using commutation/disjoint-qubit rules.
+// Only op membership is recorded; operand/result bookkeeping is deferred until
+// construction so that SSA values reflect any layers already materialized in IR.
+llvm::SmallVector<std::vector<PBCOpInterface>>
+PBCLayerContext::groupLayers(mlir::Operation *root, bool onlyOnDisjointQubit)
+{
+    llvm::SmallVector<std::vector<PBCOpInterface>> groups;
+    PBCLayer layer(this);
+    root->walk([&](PBCOpInterface op) {
+        if (isParentLayerOp(op)) {
```
lazypanda10117: I don't think this is needed as well since we are not constructing the layer at the same time. this is just for analysis. If we want to make sure we cannot do nested layers, we should probably assert that no such layer operations exist in the IR at this stage.

mlir/lib/PBC/Utils/PBCLayer.cpp
sengthai: Actually, since that first op in layer, it doesn't matter to pass the `onlyOnDisjointQubit`

mlir/lib/PBC/Utils/PBCLayer.cpp
sengthai: Fixed! [3df21ee](https://github.com/PennyLaneAI/catalyst/pull/2858/commits/3df21eed59c265d807ace82f7f4de6963027ad96)

mlir/lib/PBC/Utils/PBCLayer.cpp
sengthai: Fixed! [3df21ee](https://github.com/PennyLaneAI/catalyst/pull/2858/commits/3df21eed59c265d807ace82f7f4de6963027ad96)

mlir/lib/PBC/Utils/PBCLayer.cpp
sengthai: Fixed! [3df21ee](https://github.com/PennyLaneAI/catalyst/pull/2858/commits/3df21eed59c265d807ace82f7f4de6963027ad96)

PR #2861, opened 2026-05-20
Decompose lowering pass can handle gates on different qreg SSA values

paul0403:
**Context:**
Consider the following case
```mlir
module @circuit {
  module @module_circuit {
    func.func public @circuit() attributes {quantum.node} {
      %cst_0 = arith.constant 0.78539816339744828 : f64
      %0 = quantum.alloc( 4) : !quantum.reg
      %1 = quantum.extract %0[ 1] : !quantum.reg -> !quantum.bit

      %2 = quantum.extract %0[ 0] : !quantum.reg -> !quantum.bit
      %out_qubits = quantum.custom "Hadamard"() %2 : !quantum.bit
      %3 = quantum.insert %0[ 0], %out_qubits : !quantum.reg, !quantum.bit

      %4 = quantum.extract %3[ 2] : !quantum.reg -> !quantum.bit
      %out_qubits_0:2 = quantum.custom "ControlledPhaseShift"(%cst_0) %4, %1 : !quantum.bit, !quantum.bit
      %13 = quantum.insert %3[ 2], %out_qubits_0#0 : !quantum.reg, !quantum.bit
      %14 = quantum.insert %13[ 1], %out_qubits_0#1 : !quantum.reg, !quantum.bit
      quantum.dealloc %14 : !quantum.reg
      return
    }
    func.func private @_cphase_to_ppr(%arg0: tensor<1xf64>, %arg1: tensor<2xi64>, %arg2: !quantum.reg) -> !quantum.reg attributes {num_wires = 2 : i64, target_gate = "ControlledPhaseShift"} {
      return %arg2 : !quantum.reg
    }
  }
}
```

The decomposed gate 
```mlir
%out_qubits_0:2 = quantum.custom "ControlledPhaseShift"(%cst_0) %4, %1 : !quantum.bit, !quantum.bit
```
act on `%4` and `%1`. These qubit values are extracted from 2 different qreg SSA values, namely `%3` and `%0` (linked via an insert op). 

The decompose lowering pass only works if the source qreg of all qubit operands of the gate are exactly the same SSA value https://github.com/PennyLaneAI/catalyst/blob/45745897d3da547023761adb605741ba36c1e47b/mlir/lib/Quantum/Transforms/DecomposeLoweringImpl.hpp#L161

So the above case currently fails. This assumption is overly strict.

When collecting the qreg values used by a gate, try to see if they live on the same def-use chain (via insert ops). 
If all qreg values live on the same def-use chain, identify the latest qreg value as the operand to the call to the decomp rule func.
If any qreg values trace back to a different allocation, i.e. they are on multiple def-use chains, we raise an error.

Note that this was the intention of the original error.

**Benefits:**
Enhanced functionality.
Decompose lowering pass works with general qubit extract/insert structures, like those produced by the qref-to-value semantics conversion.


rniczh (approved):
💯 Thanks @paul0403

kipawaa (approved):
Great fix, thanks @paul0403 !

PR #2863, opened 2026-05-21
PBC Depth: Support disjoint-qubit depth in ppm-specs

sengthai:
**Context:**
Previously, `ppm-spec` reports the depth count based on the https://github.com/PennyLaneAI/catalyst/pull/2014 which later will added a new metric depth-1 as shown https://github.com/PennyLaneAI/catalyst/pull/2858. 

- Add `disjoint-qubit` to the `ppm-specs` MLIR pass and `only_disjoint_qubit` to Python `ppm_specs()`.
- Report circuit depth as `depth` (layer count) and `depth_type`:
  - `0` (default): commuting PBC ops on overlapping qubits may share a layer
  - `1` (`disjoint-qubit`): only ops with disjoint qubit support may share a layer
- Replace per-kind depth keys (`depth_pi8_ppr`, `depth_ppm`, etc.) with the aggregate `depth` metric.
- Improve AOT `ppm_specs`: build pipelines when missing, replace existing `ppm-specs` pass options, write MLIR to `/dev/null`, parse JSON from stdout only.

[sc-120052]


frontend/test/pytest/test_peephole_optimizations.py
```diff
@@ -619,20 +650,6 @@ def jit_circuit(x):  # JIT mode since x is unknown
 
             ppm_specs(jit_circuit)
 
-    def test_no_pipeline_error(self):
```
sengthai, line 622: Since PPM execution is now supported, `ppm_spec` no longer requires a custom pipeline.

rniczh (approved):
👍

mlir/include/PBC/Utils/PBCLayer.h
```diff
@@ -49,6 +50,22 @@ class PBCLayerContext {
     // Returns op lists only; operand/result bookkeeping is deferred until layer construction.
     llvm::SmallVector<std::vector<PBCOpInterface>> groupLayers(mlir::Operation *root,
                                                                bool onlyOnDisjointQubit = false);
+
+    // Worst-case PBC layer depth across `scf.if` branches within `block`
+    // (typically a function body's entry block). `scf.for` / `scf.while` are not supported yet.
```
paul0403: Reminder to update this comment to include index switch and for loops.

sengthai: done! 
https://github.com/PennyLaneAI/catalyst/blob/fd2b43e07a8fe1f26284e887b5ec6222bd4f7fc4/mlir/include/PBC/Utils/PBCLayer.h#L54-L56

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -1423,7 +1424,7 @@ def circuit():
 )
 
 
-def ppm_specs(fn):
+def ppm_specs(fn, *, only_disjoint_qubit: bool = False):
```
lazypanda10117: why allowing arbitrary argument here?

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -1457,69 +1465,57 @@ def ppm_specs(fn):
...
+            ]
 
         # add ppm-spec pass at the end to existing pipeline
         _, pass_list = new_options.pipelines[0]  # first pipeline runs the user passes
         # check if ppm-specs is already in the pass list
-        if "ppm-specs" not in pass_list:  # pragma: nocover
-            pass_list.append("ppm-specs")
+        ppm_specs_pass = "ppm-specs{disjoint-qubit=true}" if only_disjoint_qubit else "ppm-specs"
+        pass_list[:] = [p for p in pass_list if not p.startswith("ppm-specs")]
+        if ppm_specs_pass not in pass_list:  # pragma: nocover
```
lazypanda10117: I am confused here, aren't you just filtering it out and addint it back in? What's the purpose of this?

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -1457,69 +1465,57 @@ def ppm_specs(fn):
...
+            pass_list.append(ppm_specs_pass)
```
paul0403: Doesn't the above line guarantee that `ppm_specs_pass` (which always atrts with `:ppm-specs"`) will never be in `pass_list`?


```suggestion
        ppm_specs_pass = "ppm-specs{disjoint-qubit=true}" if only_disjoint_qubit else "ppm-specs"
        pass_list = [p for p in pass_list if not p.startswith("ppm-specs")]  + [ppm_specs_pass]
```

paul0403 (approved):

frontend/catalyst/passes/builtin_passes.py
sengthai: agree! the `if` is redundant. Simplified to your suggestion, with one tweak: pass_list[:] instead of pass_list = ..., since pass_list is unpacked from new_options.pipelines[0] and we need to mutate the pipeline's list in place.

mlir/lib/PBC/Transforms/CountPPMSpecs.cpp
```diff
@@ -226,22 +191,19 @@ struct CountPPMSpecsPass : public impl::CountPPMSpecsPassBase<CountPPMSpecsPass>
...
 
-        // Add the last layer if it is not empty.
-        if (!currentLayer.empty()) {
-            layers.emplace_back(std::move(currentLayer));
-        }
-
-        countDepths(layers, PPMSpecs, stringAllocator);
+        // Compute depth, but still emit JSON even if depth analysis fails for some
+        // functions (counts may have succeeded). Pass failure is reported at the end.
+        LogicalResult depthResult = computeAllFuncDepth(onlyDisjointQubit, PPMSpecs);
```
lazypanda10117, line 196: Why do we need this logicalResult type here? What's the purpose?

mlir/lib/PBC/Utils/PBCLayer.cpp
```diff
@@ -14,18 +14,168 @@
...
+            continue;
+        }
+
+        // plain ops
+        if (op.getNumRegions() == 0) {
+            continue;
+        }
+
+        // multi-region ops
+        if (op.getNumRegions() != 1) {
```
lazypanda10117, line 152: what are some examples of these multi regions ops you have in mind?

mlir/lib/PBC/Utils/PBCLayer.cpp
```diff
@@ -14,18 +14,168 @@
...
+            return op.emitOpError(
+                "worst-case depth cannot analyze operations with multiple regions");
+        }
+
+        // Recurse into the op's single-block region body (e.g. quantum.adjoint)
+        Region &region = op.getRegion(0);
```
lazypanda10117, line 158: So you are treating adjoint as a "blocking" condition here? If something is in adjoint, then it cannot be grouped into the previous layer?

mlir/lib/PBC/Utils/PBCLayer.cpp
sengthai: It is future proof, just in case we have a new operation that has more multi regions.

mlir/lib/PBC/Utils/PBCLayer.cpp
sengthai: That's right. for those quantum.adjoint, we treat those in separate layer.

sengthai: Handling such quantum.adjoint might be non trivial so i think it should be in separate PR, if it neccessary.

frontend/catalyst/passes/builtin_passes.py
sengthai: That's old one here is the update one.
https://github.com/PennyLaneAI/catalyst/pull/2863/changes#diff-186908b65c27487fa7c61d9e586ded18d641b08965e2a35ebd457ab35d20c116R1505

mlir/lib/PBC/Transforms/CountPPMSpecs.cpp
sengthai: LogicalResult is used to propagate failure in computeAllFuncDepth. 🤔

frontend/catalyst/passes/builtin_passes.py
sengthai: The `*` only means everything after `fn`, but remove this also reasonable since only one flag in this function now.

sengthai: Removed!

lazypanda10117 (approved):
🚀

PR #2864, opened 2026-05-21
Small follow up to ref semantics ops in mbqc and pbc

paul0403:
**Context:**
Forgot to register them as quantum gates when collecting closure qref values from inside regions.


mehrdad2m (approved):

rniczh (approved):
Awesome 👍

PR #2865, opened 2026-05-21
Add the `qecl.fabricate` op

joeycarter:
**Context:** In order to support T gates and π/8 PPMs in the experimental QEC pipeline, we require an operation to represent the fabrication of magic states in the QEC logical layer.

**Description of the Change:** Mirroring the `pbc.fabricate` op, this PR adds the `qecl.fabricate` op, which fabricates a logical codeblock in a specified initial state (typically a magic state). This operation is available in both MLIR and in the xDSL interface to Catalyst.

For example, the following operation represents the fabrication of a logical codeblock in the magic state (where the codeblock encodes _k_ = 1 logical qubits):

```mlir
%0 = qecl.fabricate [magic] : !qecl.codeblock<1>
```

The `LogicalCodeblockInitState` enum attribute has been updated to include the "magic" option for the fabrication of magic states. The  `qecl.fabricate` op also includes a verifier that raises an error when attempting to fabricate a logical codeblock in the logical 'zero' state (instead, the `qecl.encode` op should be used). Similarly, a verifier has been added to the `qecl.encode` op that raises an error when attempting to encode an already-allocated codeblock in the magic state.

[sc-119923]

PR #2866, opened 2026-05-22
Add the `qecl.dealloc_cb` op

joeycarter:
**Context:** In order to support T gates and π/8 PPMs in the experimental QEC pipeline, in #2865 we added support to represent the fabrication of magic states in the QEC logical layer. Now, we need the ability to also deallocate single logical codeblocks.

**Description of the Change:** This PR adds the `qecl.dealloc_cb` op, which deallocates a single logical codeblock. This operation is available in both MLIR and in the xDSL interface to Catalyst.

This operation is generally used in conjunction with the `qecl.fabricate` op. For example,

```mlir
%0 = qecl.fabricate [magic] : !qecl.codeblock<1>
// %1 = operations on the magic state...
qecl.dealloc_cb %1 : !qecl.codeblock<1>
```

In principle, it is also possible to deallocate a single logical codeblock extracted from a hyper-register, e.g.

 ```mlir
%0 = qecl.alloc() : !qecl.hyperreg<3 x 1>
%1 = qecl.extract %0[0] : !qecl.hyperreg<3 x 1> -> !qecl.codeblock<1>
qecl.dealloc_cb %1 : !qecl.codeblock<1>
```

However, IR patterns like this should be used with caution for now since they require careful consideration in the `qecp` layer below and in the runtime (for instance, similar patterns of single-qubit deallocation in the `quantum` dialect generally require the use of array-backed registers to allow for execution of the compiled workload).

[sc-119924]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -110,6 +110,8 @@
   - `qecl.fabricate`, which fabricates a logical codeblock in a specified initial state (typically a
     magic state).
     [(#2865)](https://github.com/PennyLaneAI/catalyst/pull/2865)
+  - `qecl.dealloc_cb`, which deallocates a single logical codeblock.
+    [(#2866)](https://github.com/PennyLaneAI/catalyst/pull/2866)
 
 <h3>Documentation 📝</h3>
 
--- a/frontend/catalyst/python_interface/dialects/qecl.py
+++ b/frontend/catalyst/python_interface/dialects/qecl.py
@@ -341,6 +341,22 @@ class FabricateOp(IRDLOperation):
 
 
 @irdl_op_definition
+class DeallocCodeblockOp(IRDLOperation):
+    """Deallocate a single logical codeblock."""
+
+    name = "qecl.dealloc_cb"
+
+    assembly_format = """
+            $codeblock attr-dict `:` type($codeblock)
+        """
+
+    codeblock = operand_def(base(LogicalCodeblockType))
+
+    def __init__(self, codeblock: LogicalCodeBlockSSAValue | Operation):
+        super().__init__(operands=(codeblock,))
+
+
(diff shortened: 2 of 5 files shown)


lillian542:
> However, IR patterns like this should be used with caution for now since they require careful consideration in the qecp layer below and in the runtime (for instance, **similar patterns of single-qubit deallocation in the quantum dialect generally require the use of array-backed registers to allow for execution of the compiled workload**).

How much is this an "implementation of the `quantum` dialect" challenge versus a more fundamental challenge with how the runtime works? If we want to use the procedure in Nielsen & Chuang we will be deallocating the initial codeblock (from the hyper-register) and doing a switcheroo, although that won't be obvious in the main module. (Maybe easier to take offline on Monday.)

joeycarter:
> > However, IR patterns like this should be used with caution for now since they require careful consideration in the qecp layer below and in the runtime (for instance, **similar patterns of single-qubit deallocation in the quantum dialect generally require the use of array-backed registers to allow for execution of the compiled workload**).
> 
> How much is this an "implementation of the `quantum` dialect" challenge versus a more fundamental challenge with how the runtime works? If we want to use the procedure in Nielsen & Chuang we will be deallocating the initial codeblock (from the hyper-register) and doing a switcheroo, although that won't be obvious in the main module. (Maybe easier to take offline on Monday.)

I wouldn't say this is a fundamental challenge, just something to be aware of as we build out the pipeline. If we're continuing to validate by re-lowering to the the `quantum` dialect, then it should be as simple as enabling the `array-backed-registers` option in the `quantum-to-llvm` pass, but we'll need to verify.

paul0403 (approved):

sengthai (approved):
Straightforward because of well organized PR. Thanks @joeycarter

lillian542 (approved):

PR #2868, opened 2026-05-22
Adjoint lowering pass can handle control flow with arbitrary quantum operands

paul0403:
**Context:**
Recently we extended the `quantum.adjoint` operation to handle arbitrary quantum operand structures (as opposed to being restricted to have a single qreg arg) #2590

In adjoint lowering pass, when the adjoint region contains other control flow regions, the code is still assuming the control flow regions have a single qreg argument (at index `0`).

Update adjoint lowering pass's control flow part to deal with arbitrary quantum operand structure.

**Benefits:**
Adjoint lowering pass works with more general programs.
In particular, those generated by the ref-to-value semantics conversion will in general have unconstrained quantum operand structures on control flow ops.

### DIFF
--- a/mlir/lib/Quantum/Transforms/AdjointPatterns.cpp
+++ b/mlir/lib/Quantum/Transforms/AdjointPatterns.cpp
@@ -613,41 +613,58 @@ class AdjointGenerator {
 
     void visitOperation(scf::IndexSwitchOp switchOp, OpBuilder &builder)
     {
-        std::optional<Value> qureg = getQuantumReg(switchOp.getResults());
-        if (!qureg.has_value()) {
+        SmallVector<Value> yieldedQValues = getQuantumValues(switchOp.getResults());
+        if (yieldedQValues.empty()) {
+            // This operation is purely classical
             return;
         }
 
         Value tape = cache.controlFlowTapes.at(switchOp.getOperation());
         Value index = ListPopOp::create(builder, switchOp.getLoc(), tape);
-        Value reversedResult = remappedValues.lookup(getQuantumReg(switchOp.getResults()).value());
 
-        auto findRootQuregInRegion = [&](Region &region) {
+        SmallVector<Value> reversedResults;
+        for (auto v : getQuantumValues(switchOp.getResults())) {
+            reversedResults.push_back(remappedValues.lookup(v));
+        }
+
+        auto findRootQvaluesInRegion = [&](Region &region) -> SetVector<Value> {
+            SetVector<Value> qvalues;
             for (Operation &innerOp : region.getOps()) {
                 for (Value operand : innerOp.getOperands()) {
-                    if (isa<quantum::QuregType>(operand.getType())) {
-                        return operand;
+                    bool isDefinedFromOutsideRegion =
+                        operand.getParentRegion()->isProperAncestor(&region);
+                    if (isa<quantum::QuregType, quantum::QubitType>(operand.getType()) &&
+                        isDefinedFromOutsideRegion) {
+                        qvalues.insert(operand);
                     }
                 }
             }
-            llvm_unreachable("failed to find qureg in scf.index_switch region");
+            assert(!qvalues.empty() && "failed to find quantum values in scf.index_switch region");
+            return qvalues;
         };
 
-        auto newSwitchOp = scf::IndexSwitchOp::create(builder, switchOp.getLoc(),
-                                                      TypeRange{reversedResult.getType()}, index,
-                                                      switchOp.getCases(), switchOp.getNumCases());
+        auto newSwitchOp =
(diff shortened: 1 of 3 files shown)


rniczh (approved):
Nice work 💯 ! Would you have a chance to add a test for this improvement?

mlir/lib/Quantum/Transforms/AdjointPatterns.cpp
```diff
@@ -452,75 +454,111 @@ class AdjointGenerator {
...
-        remappedValues.map(startingThenQureg, reversedIf.getResult(0));
+
+        SetVector<Value> startingThenQvalues = findOldestQvaluesInRegion(ifOp.getThenRegion());
+        SetVector<Value> startingElseQvalues = findOldestQvaluesInRegion(ifOp.getElseRegion());
+
+        int res_idx = 0;
+        for (auto [t, e] : llvm::zip_equal(startingThenQvalues, startingElseQvalues)) {
+            assert(t == e && "Expected the same input quantum values for both scf.if branches");
+            remappedValues.map(t, reversedIf.getResult(res_idx++));
+        }
```
paul0403: A bit unhappy that all of these are kinda assuming there's no classical operands/results to these scf ops. But at least it's an assumption in the original code already. 

I'll see what I can do regarding this after adding tests

rniczh (approved):
💯

mlir/lib/Quantum/Transforms/AdjointPatterns.cpp
```diff
@@ -452,75 +454,111 @@ class AdjointGenerator {
...
+        // The quantum values are captured from outside rather than passed in through a
         // basic block argument. We thus need to traverse the region to look for it.
-        auto findOldestQuregInRegion = [&](Region &region) {
+        auto findOldestQvaluesInRegion = [&](Region &region) -> SetVector<Value> {
+            SetVector<Value> qvalues;
             for (Operation &innerOp : region.getOps()) {
                 for (Value operand : innerOp.getOperands()) {
-                    if (isa<quantum::QuregType>(operand.getType())) {
-                        return operand;
+                    bool isDefinedFromOutsideRegion =
```
mehrdad2m, line 511: why is this newly added check needed now and not before?

mlir/lib/Quantum/Transforms/AdjointPatterns.cpp
mehrdad2m: Are you saying that this is a bug in the current state of the code? in that case it would be nice to take a note of this for future reference as in either a TODO entry in the code or an issue page.

mehrdad2m: suggestion:
```suggestion
        for (auto [idx, pair] : llvm::enumerate(llvm::zip_equal(startingThenQvalues, startingElseQvalues))) {
          auto [t, e] = pair;
          assert(t == e && "Expected the same input quantum values for both scf.if branches");
          remappedValues.map(t, reversedIf.getResult(idx));
        }
```
I think `res_idx` should be an unsigned looking at the `getResult` definition.

mehrdad2m (approved):
Looks good! just a minor comment.

mlir/lib/Quantum/Transforms/AdjointPatterns.cpp
paul0403: This lambda is trying to find all "quantum arguments" to a region via clousure. This is because `scf.if` does not have operands. The lanbda does so by going through all operands of all ops in the region, and just record the values with qreg/qubit type. Kinda like what we did in the ref-to-val semantics conversion for regioned-ops.

The old code only has qreg type, and returns at the first instance, which would be the qreg operand of the first extract op in the region. This `Value` would be the same latest qreg `Value` from above the `if` op.

But now we include qubits as well. Unlike registers, many ops can have qubit operands, e.g. 
```mlir
%1 = quantum.custom %0
%2 = quantum.custom %1
```
we must not count qubit `Value`s that are defined from inside the region, like `%1` and `%2`.

mlir/lib/Quantum/Transforms/AdjointPatterns.cpp
paul0403: Nice! https://github.com/PennyLaneAI/catalyst/pull/2868/commits/5622fcdc46a1cab62e11a1654caacdf13d7a5be7

paul0403: I thought about it a bit more, classical return values from scf ops are actually already supported in the current logic. The container here iterated over are actually already quantum-only, so the remap is correct even when there are classical results/operands.

mlir/lib/Quantum/Transforms/AdjointPatterns.cpp
mehrdad2m: Ah I see, thanks for the explanation.

PR #2870, opened 2026-05-22
Lower to subroutine with magic state for T

lillian542:
**Context:**
Current `convert-quantum-to-qecl` pass is limited to Clifford gates

We add lowering for T-gates, following the procedure outlined in Nielsen & Chuang (Section 10.6.2), and using the new `qecl.fabricate [magic]` operation. 

**Benefits:**

We can lower circuits of arbitrary gates, expressed in the frontend, down to the `qecp` layer (for k=1 codes) and incorporate magic states.


lillian542:
[sc-119926]

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
```diff
@@ -616,5 +639,86 @@ def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
...
+        """Create a subroutine that takes in a codeblock in state |φ>, and outputs a codeblock
+        in state T|φ>. The subroutine includes instructions to fabricate a codeblock in the
+        magic state, entangle it with the input codeblock and perform measurement and corrections.
+        As the gate application includes teleportation of the state from the input codeblock to
+        the codeblock containing the magic state, the output codeblock is the one generated by
+        the fabricate op.
+
+        This procedure is outlined in Nielsen & Chuang, (Section 10.6.2), and is as follows:
+
+            |magic_state> ──╭●───────SX─┤ output_state
```
lillian542:
```suggestion
            |magic_state> ──╭●───────SX─┤ |output_state>
```

lillian542:
We're getting a couple of integration test failures because this adds `qecl.fabricate` and `qecl.deallocate_cb` (inside a subroutine) that don't lower to LLVM. I'm thinking I'll just mark them as xfail until we can "lower" this through `qecp` back to the `quantum` dialect, unless there's a simpler way of handling that.

joeycarter:
> We're getting a couple of integration test failures because this adds `qecl.fabricate` and `qecl.deallocate_cb` (inside a subroutine) that don't lower to LLVM. I'm thinking I'll just mark them as xfail until we can "lower" this through `qecp` back to the `quantum` dialect, unless there's a simpler way of handling that.

Agreed, that's probably the best option for now. Another option is to keep the original integration test the same but rename it to "Clifford-only" and then add another integration test for Clifford+T, and mark that one as xfail.

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
```diff
@@ -599,6 +618,11 @@ def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
                 f"codeblock, k, is 1, but got k = {self.k}"
             )
 
+        module_block = op.regions[0].blocks.first
+        assert module_block is not None, "Module has no block"
+        t_subroutine = self.create_t_subroutine()
```
joeycarter, line 626: One thing that worked well in the `qecl-to-qecp` pass was that we passed the subroutine object to the rewrite-pattern class, that way we could get the subroutine name directly with `subroutine.sym_name`, rather writing it out twice, first when creating it and then again when calling it.

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
```diff
@@ -616,5 +640,86 @@ def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
         # this pass removes them
         ReconcileUnrealizedCastsPass().apply(ctx, op)
 
+    def create_t_subroutine(self):
+        """Create a subroutine that takes in a codeblock in state |φ>, and outputs a codeblock
+        in state T|φ>. The subroutine includes instructions to fabricate a codeblock in the
+        magic state, entangle it with the input codeblock and perform measurement and corrections.
```
joeycarter:
```suggestion
        magic state, entangle it with the input codeblock and perform measurements and corrections.
```

(or in the singular, "a measurement and correction")

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
```diff
@@ -616,5 +640,86 @@ def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
...
+            fabricate_op = qecl.FabricateOp("magic", qecl.LogicalCodeblockType(k=self.k))
+            magic_state_block = fabricate_op.results[0]
+
+            # apply cnot between aux codeblock and input data codeblock
+            cnot_op = qecl.CnotOp(
+                in_ctrl_codeblock=magic_state_block,
+                idx_ctrl=0,
+                in_trgt_codeblock=in_codeblock,
+                idx_trgt=0,
+            )
```
joeycarter, line 685: Maybe worth adding a note that we've also assumed k = 1 here by using `idx=0`.

frontend/test/pytest/python_interface/transforms/qecl/test_xdsl_convert_quantum_to_qecl.py
```diff
@@ -633,6 +633,43 @@ def test_gate_s_k_1(self, run_filecheck, quantum_to_qecl_pipeline_k_1):
...
+            // CHECK-NEXT: [[magic_cb2:%.+]], [[in_codeblock2:%.+]] = qecl.cnot [[magic_cb]][0], [[in_codeblock]][0]
+            // CHECK-NEXT: [[mres:%.+]], [[in_codeblock3:%.+]] = qecl.measure [[in_codeblock2]][0]
+            // CHECK-NEXT: qecl.dealloc_cb [[in_codeblock3]]
+            // CHECK-NEXT: [[out_codeblock:%.+]] = scf.if [[mres]] -> (!qecl.codeblock<1>)
+            // CHECK-NEXT: [[s_corrected_cb:%.+]] = qecl.s [[magic_cb2]][0]
+            // CHECK-NEXT: [[corrected_cb:%.+]] = qecl.x [[s_corrected_cb]]
+            // CHECK-NEXT: scf.yield [[corrected_cb]]
+            // CHECK-NEXT: else
+            // CHECK-NEXT: scf.yield [[magic_cb2]]
+            // CHECK: func.return [[out_codeblock]]
```
joeycarter: (For readability)

```suggestion
            //      CHECK: func.func private @apply_T([[in_codeblock:%.+]]: !qecl.codeblock<1>)
            // CHECK-NEXT: [[magic_cb:%.+]] = qecl.fabricate[magic] : !qecl.codeblock<1>
            // CHECK-NEXT: [[magic_cb2:%.+]], [[in_codeblock2:%.+]] = qecl.cnot [[magic_cb]][0], [[in_codeblock]][0]
            // CHECK-NEXT: [[mres:%.+]], [[in_codeblock3:%.+]] = qecl.measure [[in_codeblock2]][0]
            // CHECK-NEXT: qecl.dealloc_cb [[in_codeblock3]]
            // CHECK-NEXT: [[out_codeblock:%.+]] = scf.if [[mres]] -> (!qecl.codeblock<1>)
            // CHECK-NEXT:     [[s_corrected_cb:%.+]] = qecl.s [[magic_cb2]][0]
            // CHECK-NEXT:     [[corrected_cb:%.+]] = qecl.x [[s_corrected_cb]]
            // CHECK-NEXT:     scf.yield [[corrected_cb]]
            // CHECK-NEXT: else
            // CHECK-NEXT:     scf.yield [[magic_cb2]]
            //      CHECK: func.return [[out_codeblock]]
```

frontend/test/pytest/python_interface/transforms/qecl/test_xdsl_convert_quantum_to_qecl.py
```diff
@@ -835,12 +872,14 @@ def circuit():
...
             # CHECK: qecl.hadamard {{%.+}}[0]
             # CHECK: qecl.qec
+            # CHECK: apply_T
             # CHECK: qecl.measure {{%.+}}[0]
             # CHECK: quantum.mcmobs
             # CHECK: quantum.sample
             # CHECK: qecl.insert_block
             # CHECK: qecl.dealloc
             qp.H(0)
+            qp.T(0)
```
joeycarter, line 882: I commented this elsewhere, but another option is to have separate integration tests for a Clifford-only circuit and a Clifford+T circuit.

joeycarter (commented):
Looks great! 💯 

Just a few suggestions, nothing major. Happy to approve once CI is green.

frontend/test/pytest/python_interface/transforms/qecl/test_xdsl_convert_quantum_to_qecl.py
lillian542: Very good point, let's do that 😅

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
```diff
@@ -616,5 +640,86 @@ def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
...
+        block = Block(arg_types=input_types)
+
+        with ImplicitBuilder(block):
+            (in_codeblock,) = block.args
+
+            # fabricate aux codeblock in magic state
+            fabricate_op = qecl.FabricateOp("magic", qecl.LogicalCodeblockType(k=self.k))
+            magic_state_block = fabricate_op.results[0]
+
+            # apply cnot between aux codeblock and input data codeblock
```
lillian542, line 678:
```suggestion
            # apply cnot between aux codeblock and input data codeblock
            # hardcode idx=0 for now, based on the current assumption that k=1
```

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
lillian542: added a comment above :)

frontend/test/pytest/python_interface/transforms/qecl/test_xdsl_convert_quantum_to_qecl.py
lillian542: Actually this seems to be fine since the target is `"mlir"`, the problem is coming from the tests for the `qecp-to-quantum` pass integration tests, when it tries to lower to LLVM and execute. I thought `dce` would have cleared this out by then, since the subroutine isn't used, but I guess it happens lower than I thought?

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
lillian542: Updated here https://github.com/PennyLaneAI/catalyst/pull/2870/commits/964a0e1c3b08a6fa1156b7a059f73349d8367c2c

frontend/test/pytest/python_interface/transforms/qecl/test_xdsl_convert_quantum_to_qecl.py
joeycarter: I checked, the problem is that `symbol-dce` isn't run until the end of the `quantum-compilation` stage in the pipeline. But, by the end of the `convert-qecp-to-quantum` pass (which is applied within this stage), we're left with the subroutine

```mlir
func.func private @apply_T(%0: !quantum.reg) -> !qecl.codeblock<1> {
  ...
  func.return %9 : !quantum.reg
}
```

There's a mismatch in the return type (`!qecl.codeblock<1>` vs. `!quantum.reg`), so compilation fails. We should be able to resolve the issue by updating the pipeline to include a `symbol-dce` pass after the `convert-qecl-to-qecp` pass. (We could probably also add another after the `convert-quantum-to-qecl` pass). I'm looking to see what the best way to do this is.

joeycarter: A fast fix would be to add

```python
@qp.transform(pass_name="symbol-dce")
```

to the pipeline of the `test_qec_pass_ghz_nullqubit_integration` test after the `convert_quantum_to_qecl_pass` (i.e., immediately before it in the list of decorators). I would consider this a stopgap measure until we have proper lowering of the `apply_T` subroutine to the `qecp` dialect, then we should be able to remove it.

lillian542: Awesome, I'll add that for now, and note that it should be removed in the relevant shortcut story :)

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
```diff
@@ -1075,6 +1076,7 @@ def test_convert_qecl_noise_to_qecp_noise_pass_integration(self, run_filecheck_q
         @qp.qjit(target="mlir", capture=True)
         @convert_qecl_to_qecp_pass(qec_code="Steane", number_errors=1)
         @inject_noise_to_qecl_pass
+        @qp.transform(pass_name="symbol-dce")
```
lillian542, line 1079: These can be removed once the qecl-to-qecp lowering is in place (noted in shortcut)

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecp_to_quantum.py
```diff
@@ -611,6 +612,7 @@ def test_qec_pass_ghz_nullqubit_integration(self):
         @convert_qecp_to_quantum_pass
         @convert_qecl_to_qecp_pass(qec_code="Steane", number_errors=1)
         @inject_noise_to_qecl_pass
+        @qp.transform(pass_name="symbol-dce")
```
lillian542, line 615: These can be removed once lowering of `qecl.allocate/deallocate_cb` to `quantum` dialect are in place (noted in shortcut)

joeycarter (approved):
💯

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
```diff
@@ -616,5 +643,87 @@ def apply(self, ctx: Context, op: builtin.ModuleOp) -> None:
...
+        As the gate application includes teleportation of the state from the input codeblock to
+        the codeblock containing the magic state, the output codeblock is the one generated by
+        the fabricate op.
+
+        This procedure is outlined in Nielsen & Chuang, (Section 10.6.2), and is as follows:
+
+            |magic_state> ──╭●───────SX─┤ |output_state>
+                            |        ║
+            |input_state> ──╰X──┤↗├──║── deallocate
+                                 ╚═══╝
```
multiphaseCFD, line 659: Wondering if the assumption is still k=1 for now?

multiphaseCFD (approved):
LGTM! Thanks @lillian542

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
multiphaseCFD: Thanks for putting the assumption here

PR #2872, opened 2026-05-25
Add support for `scf.if` ops in the `convert-quantum-to-qecl` pass

joeycarter:
**Context:** In order to support more general input programs in the experimental QEC compilation pipeline, we must add support for control-flow operations.

**Description of the Change:** This PR adds support for conditional blocks (via [`scf.if`](https://mlir.llvm.org/docs/Dialects/SCFDialect/#scfif-scfifop) ops) in the `convert-quantum-to-qecl` pass. For example, an input program such as

```mlir
func.func @test_program(%arg0: tensor<i1>) {
    %1 = quantum.extract %0[0] : !quantum.reg -> !quantum.bit
    %2 = quantum.custom "PauliX"() %1 : !quantum.bit
    %cond = tensor.extract %arg0[] : tensor<i1>
    %3 = scf.if %cond -> (!quantum.bit) {
        %4 = quantum.custom "PauliZ"() %2 : !quantum.bit
        scf.yield %4 : !quantum.bit
    } else {
        scf.yield %2 : !quantum.bit
    }
    %5 = quantum.insert %0[0], %3 : !quantum.reg, !quantum.bit
    func.return
}
```

is lowered to

```mlir
func.func @test_program(%arg0: tensor<i1>) {
    %1 = qecl.extract_block %0[0] : !qecl.hyperreg<1 x 1> -> !qecl.codeblock<1>
    %2 = qecl.encode[zero] %1 : !qecl.codeblock<1>
    %3 = qecl.insert_block %0[0], %2 : !qecl.hyperreg<1 x 1>, !qecl.codeblock<1>
    %4 = qecl.extract_block %3[0] : !qecl.hyperreg<1 x 1> -> !qecl.codeblock<1>
    %5 = qecl.qec %4 : !qecl.codeblock<1>
    %6 = qecl.x %5[0] : !qecl.codeblock<1>
    %7 = qecl.qec %6 : !qecl.codeblock<1>
    %cond = tensor.extract %arg0[] : tensor<i1>
    %8 = scf.if %cond -> (!qecl.codeblock<1>) {
        %9 = qecl.z %7[0] : !qecl.codeblock<1>
        %10 = qecl.qec %9 : !qecl.codeblock<1>
        scf.yield %10 : !qecl.codeblock<1>
    } else {
        scf.yield %7 : !qecl.codeblock<1>
    }
    %11 = qecl.insert_block %3[0], %8 : !qecl.hyperreg<1 x 1>, !qecl.codeblock<1>
    qecl.dealloc %11 : !qecl.hyperreg<1 x 1>
    func.return
}
```

Support for `scf.if` ops that return `!quantum.reg` values, and a mix of quantum and classical types, has also been added.

[sc-120230]


frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
```diff
@@ -245,7 +247,9 @@ def match_and_rewrite(self, op: quantum.ExtractOp, rewriter: PatternRewriter):
         # Recall that the `alloc` conversion pattern inserts a builtin.unrealized_conversion_cast
         # op immediately after the alloc that converts from qecl.hyperreg -> quantum.reg.
         qreg_owner_op = op.qreg.owner
-        if not _is_type_convertible(qreg_owner_op, qecl.LogicalHyperRegisterType):
+        if not _is_type_convertible(
+            qreg_owner_op, qecl.LogicalHyperRegisterType
+        ):  # pragma: no cover
```
joeycarter, line 252: I'm sneaking in some opportunistic 'no cover' pragmas for blocks that are generally unreachable to improve the coverage score 🙈

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
```diff
@@ -494,6 +500,175 @@ def match_and_rewrite(self, op: quantum.MeasureOp, rewriter: PatternRewriter):
...
+                conv_cast_op = self._insert_conv_cast_op(operand_owner, rewriter)
+                new_yield_operands.append(conv_cast_op.results[0])
+                self._notify_parent_op_modified(op, rewriter)
+
+            else:
+                # Not a quantum type, no need to convert
+                new_yield_operands.append(operand)
+
+        # Update the yield op's operands
+        op.operands = new_yield_operands
```
lillian542, line 565: this doesn't need a `notify_op_modified`?

frontend/test/pytest/python_interface/transforms/qecl/test_xdsl_convert_quantum_to_qecl.py
```diff
@@ -813,6 +815,171 @@ def test_unconvertible_measure(self, run_filecheck, quantum_to_qecl_pipeline_k_1
...
+        """Test a simple program with an `scf.if` conditional, which applies gates conditional on
+        some input value.
+        """
+        program = """
+        builtin.module @top {
+        // CHECK-LABEL: func.func @test_program(
+        func.func @test_program(%arg0: tensor<i1>) {
+            // CHECK: qecl.alloc() : !qecl.hyperreg<1 x 1>
+            //   COM: <qecl.extract_block>
+            //   COM: <qecl.encode>
```
lillian542, line 871: why `COM` instead of `CHECK`?

lillian542 (commented):
Looks good to me! I just left a couple questions.

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
joeycarter: No, only if we change the input(s)/output(s) of an op do we need to notify the rewriter that the op was modified. In the case where the operand type is not a `quantum` type, we push the original operand to the list of new operands. If all the operands are not `quantum` types, then the op hasn't changed, so no need to notify.

frontend/test/pytest/python_interface/transforms/qecl/test_xdsl_convert_quantum_to_qecl.py
joeycarter: Good question, since I was debating what to do here. For reference, the [`COM:` directive](https://llvm.org/docs/CommandGuide/FileCheck.html#the-com-directive) in FileCheck is a comment, so these lines are not checked. I opted to make these comments here in order to remind the reader of the test what's happening in the dialect conversion, without explicitly testing these components since they're beyond the scope of the unit test. In this case, we're testing the lowering of conditional blocks, so don't want to test the conversion pattern for `quantum.alloc` ops (in case we change it in the future, we wouldn't want this unit test to fail).

lillian542: Thanks!

lillian542 (approved):
Looks good to me!

frontend/test/pytest/python_interface/transforms/qecl/test_xdsl_convert_quantum_to_qecl.py
```diff
@@ -850,6 +852,171 @@ def test_unconvertible_measure(self, run_filecheck, quantum_to_qecl_pipeline_k_1
...
+            %1 = quantum.extract %0[ 0] : !quantum.reg -> !quantum.bit
+            %2 = quantum.custom "PauliX"() %1 : !quantum.bit
+
+            // CHECK: [[cond:%.+]] = tensor.extract
+            %cond = tensor.extract %arg0[] : tensor<i1>
+
+            // CHECK: [[out_cb:%.+]] = scf.if [[cond]] -> (!qecl.codeblock<1>)
+            // CHECK:     [[cb3:%.+]] = qecl.z [[cb2]]
+            // CHECK:     [[cb4:%.+]] = qecl.qec
+            // CHECK:     scf.yield [[cb4]] : !qecl.codeblock<1>
```
multiphaseCFD, line 944: Curious if lowering nested `scf.if` is supported.

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
```diff
@@ -516,6 +522,175 @@ def match_and_rewrite(self, op: quantum.MeasureOp, rewriter: PatternRewriter):
...
+
+        return conv_cast_op
+
+    @classmethod
+    def _notify_parent_op_modified(cls, op: scf.YieldOp, rewriter: PatternRewriter):
+        """Notify the rewriter that the parent op of the matched `scf.yield` op (typically another
+        `scf` op, like `scf.if`, `scf.for`, etc.) was modified (this triggers the rewrite pattern
+        for the parent op, which updates the return types(s) corresponding to the new yield types).
+        """
+        parent_op = op.parent_op()
```
multiphaseCFD, line 588: So nested `scf.if` is supported?

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
```diff
@@ -516,6 +522,175 @@ def match_and_rewrite(self, op: quantum.MeasureOp, rewriter: PatternRewriter):
...
+        rewriter.notify_op_modified(parent_op)
+
+
+# MARK: SCF Base Pattern
+
+
+class ScfConversionPattern(RewritePattern):
+    """Base class for conversion of `scf` ops (except `scf.yield`)."""
+
+    def match_and_rewrite(self, op: Operation, rewriter: PatternRewriter):
```
multiphaseCFD, line 599: Per `SCF.IF` operation, it wouldn't accept `Quantum Type` SSA values are operands, hence there is no need to deal with the type conversion for its operand and we only need to care about the yield operation, right?

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
```diff
@@ -516,6 +522,175 @@ def match_and_rewrite(self, op: quantum.MeasureOp, rewriter: PatternRewriter):
...
+
+    @classmethod
+    def _insert_conv_cast_op(
+        cls, operand_owner: Operation, rewriter: PatternRewriter
+    ) -> builtin.UnrealizedConversionCastOp:
+        """Helper function to insert a `builtin.unrealized_conversion_cast` op after
+        `operand_owner`, which converts the result of `operand_owner` to the type of its operand.
+        """
+        conv_cast_op = builtin.UnrealizedConversionCastOp.get(
+            (operand_owner.results[0],), (operand_owner.operands[0].type,)
```
multiphaseCFD, line 575: I assume that patterns applied before have converted quantum type to qecl types already, so we can rely on the operand_owner's type for type conversion

frontend/test/pytest/python_interface/transforms/qecl/test_xdsl_convert_quantum_to_qecl.py
```diff
@@ -850,6 +852,171 @@ def test_unconvertible_measure(self, run_filecheck, quantum_to_qecl_pipeline_k_1
...
+            // CHECK:     scf.yield [[mres_true]], [[hreg_true]] : i1, !qecl.hyperreg<1 x 1>
+            // CHECK: else
+            // CHECK:     qecl.extract_block [[hreg0]][0]
+            //   COM:     <qec cycle>
+            // CHECK:     [[mres_false:%.+]], {{%.+}} = qecl.measure
+            //   COM:     <qec cycle>
+            // CHECK:     [[hreg_false:%.+]] = qecl.insert_block [[hreg0]][0]
+            // CHECK:     scf.yield [[mres_false]], [[hreg_false]] : i1, !qecl.hyperreg<1 x 1>
+            %1, %2 = scf.if %extracted -> (i1, !quantum.reg) {
+                %5 = quantum.extract %0[ 0] : !quantum.reg -> !quantum.bit
```
multiphaseCFD: why we skip %3, %4 and jump to %5 directly? 🤔

multiphaseCFD (approved):
Looks good to me! Nice work Joey!

frontend/test/pytest/python_interface/transforms/qecl/test_xdsl_convert_quantum_to_qecl.py
joeycarter: Haha this is because I copied the output from QJIT and cut out a few ops not needed in the test :sweat_smile:

Updated here: 087e78aed

frontend/test/pytest/python_interface/transforms/qecl/test_xdsl_convert_quantum_to_qecl.py
joeycarter: Yes nested `scf.if`s are also supported. I added a test here: 7d5539540

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
joeycarter: Yes, see https://github.com/PennyLaneAI/catalyst/pull/2872#discussion_r3356961177

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
joeycarter: That's right, the only operand the `scf.if` op accepts is the `condition` value: https://mlir.llvm.org/docs/Dialects/SCFDialect/#operands-3

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
joeycarter: Exactly, this is a great case that illustrates how the greedy rewrite-pattern applier works. On the first walk through the IR, it reaches the `scf.if` op and sees that its return type and `yield` operand are both `quantum` types, so nothing to do. It then recurses into the `true` region of the `if` op and converts all the ops therein, with the conversion of the `yield` op at the end. If marks the `if` op as modified, so on the next walk through the IR, it now sees that the return type is a `quantum` type, but the `yield` operand now has a `qecl` type, so it updates the return type of the `scf.if` op to match. It recurses into the `false` region next, and at the end the `yield` operand types from both regions and the `scf.if` return type should all agree.

It takes a number of walks through the IR to do all the conversions, but it gets the job done :wink:

PR #2873, opened 2026-05-26
Replace Op strings with Op types in decompose.py

### LINKED ISSUE #2875: [BUG] `DecompRuleInterpreter` performs string check rather than type check during `cleanup` method
Because of this hard string match [here](https://github.com/PennyLaneAI/catalyst/blob/main/frontend/catalyst/from_plxpr/decompose.py#L244) for "Controlled", this MWE raises an unexpected error,

```python 
import pennylane as qml
from pennylane.decomposition import controlled_resource_rep

def _resources():
    return {
        controlled_resource_rep(qml.BasisEmbedding, {"num_wires": 1}, num_control_wires=1): 1,
        qml.resource_rep(qml.GlobalPhase): 1,
    }


@qml.register_resources(_resources)
def my_rz(phi, wires):
    qml.GlobalPhase(phi / 2)
    qml.ctrl(qml.BasisEmbedding([1], wires=[1]), control=wires)


@qml.qjit(capture=True)
@qml.decompose(gate_set={"CNOT", "PauliX", "GlobalPhase", "MultiControlledX"},
               fixed_decomps={qml.RZ: my_rz})
def f(phi):
    qml.RZ(phi, 0)
    return qml.state()
```

```pycon
>>> f(0.5)
ValueError: Could not capture _OperatorNode(op=Controlled(BasisState(num_wires=1), num_control_wires=1, num_work_wires=0, num_zero_control_values=0, work_wire_type=borrowed), num_work_wire_not_available=0, work_wire_dependent=False, min_work_wires=0, reachable=True) without the number of wires.
```

maliasadi:
**Context:**
Replace Op strings with Op types in decompose.py

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-120705]
Fixes #2875

### DIFF
--- a/frontend/catalyst/from_plxpr/decompose.py
+++ b/frontend/catalyst/from_plxpr/decompose.py
@@ -237,14 +237,17 @@ class DecompRuleInterpreter(qp.capture.PlxprInterpreter):
                     requires_copy=requires_copy,
                     pauli_word=pauli_word,
                 )
-            elif not any(
-                keyword in getattr(op.op, "name", "")
-                for keyword in (
-                    "Adjoint",
-                    "Controlled",
-                    "TemporaryAND",
-                    "ChangeOpBasis",
-                    "Prod",
+            elif not (
+                (op_type := getattr(op.op, "op_type", None)) is not None
+                and issubclass(
+                    op_type,
+                    (
+                        qp.ops.Adjoint,
+                        qp.ops.Controlled,
+                        qp.ops.ChangeOpBasis,
+                        qp.ops.Prod,
+                        qp.TemporaryAND,
+                    ),
                 )
             ):  # pragma: no cover
                 # Note that the graph-decomposition returns abstracted rules
--- a/frontend/catalyst/from_plxpr/decompose.py
+++ b/frontend/catalyst/from_plxpr/decompose.py
@@ -255,7 +255,7 @@ class DecompRuleInterpreter(qp.capture.PlxprInterpreter):
                 # These abstracted rules cannot be captured and lowered.
                 # We use MLIR AdjointOp and ControlledOp primitives
                 # to deal with decomposition of symbolic operations at PLxPR.
-                raise ValueError(f"Could not capture {op} without the number of wires.")
+                raise ValueError(f"Could not capture {op.op.name} without the number of wires.")
 
 
 # pylint: disable=too-many-arguments, too-many-positional-arguments
--- a/frontend/catalyst/from_plxpr/decompose.py
+++ b/frontend/catalyst/from_plxpr/decompose.py
@@ -249,7 +249,7 @@ class DecompRuleInterpreter(qp.capture.PlxprInterpreter):
                         qp.TemporaryAND,
                     ),
                 )
-            ):  # pragma: no cover
+            ):
                 # Note that the graph-decomposition returns abstracted rules
                 # for Adjoint and Controlled operations, so we skip them here.
                 # These abstracted rules cannot be captured and lowered.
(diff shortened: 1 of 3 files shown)


frontend/catalyst/from_plxpr/decompose.py
astralcai:
```suggestion
                raise ValueError(f"Could not capture {op.op} without the number of wires.")
```
`op` is an `_OperatorNode`, so the error message often turns out to be confusing.

frontend/catalyst/from_plxpr/decompose.py
```diff
@@ -237,14 +237,17 @@ def cleanup(self):
...
+                    op_type,
+                    (
+                        qp.ops.Adjoint,
+                        qp.ops.Controlled,
+                        qp.ops.ChangeOpBasis,
+                        qp.ops.Prod,
+                        qp.TemporaryAND,
+                    ),
                 )
             ):  # pragma: no cover
```
andrijapau: Why do we have a pragma no cover here? 🤔

frontend/catalyst/from_plxpr/decompose.py
maliasadi: ah this is outdated ! let me fix in a followup push

maliasadi: Fixed [d2cd267](https://github.com/PennyLaneAI/catalyst/pull/2873/commits/d2cd2672de38630260d8a54407051f90e48f8c8e)

frontend/catalyst/from_plxpr/decompose.py
maliasadi: Added [17f2d55](https://github.com/PennyLaneAI/catalyst/pull/2873/commits/17f2d55105b364a177b13c2ab539fe5d4e38fcec)

frontend/test/pytest/from_plxpr/test_decompose_transform.py
```diff
@@ -1085,6 +1086,88 @@ def c():
...
+                controlled_resource_rep(
+                    qp.BasisEmbedding, {"num_wires": 1}, num_control_wires=1
+                ): 1,
+                qp.resource_rep(qp.GlobalPhase): 1,
+            }
+
+        @qp.register_resources(_resources)
+        def my_rz(phi, wires):
+            qp.GlobalPhase(phi / 2)
+            qp.ctrl(qp.BasisEmbedding, control=wires)([1], wires=[1])
```
maliasadi, line 1155: `qml.ctrl(qml.BasisEmbedding([1], wires=[1]), control=wires)`  now triggers a capture issue. But the issue related to the string-based checking in decmpose.py is fixed in both cases

astralcai (approved):

andrijapau (commented):
My only concern is here is that we're technically missing coverage for `Adjoint`, `ChangeOpBasis`, `Prod` and `TemporaryAND`, no?

maliasadi:
> My only concern is here is that we're technically missing coverage for `Adjoint`, `ChangeOpBasis`, `Prod` and `TemporaryAND`, no?

@andrijapau each edge case is tested by the `TestPlxPRDecomposition` suite (just a few: `test_adjoint`, `test_ctrl`, `test_template_qft`, `test_multi_qubits`). Removing any of these Ops will break tests in `from_plxpr`.

Note this is a *temporary* workaround introduced a few cycles ago for `Adjoint`/`Ctrl` and later expanded for other ops! We should not rely on it long-term and will completely revisit this class after the new Operator is developed.

andrijapau (approved):

PR #2874, opened 2026-05-26
Fix for macos local wheels issue

kipawaa:
**Context:**
MacOS locally-built wheels would fail thanks to `install_name_tool` preferring the wrong dylib.

Prefers the correct dylib.

**Benefits:**
working wheels

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -331,13 +331,13 @@ class CustomBuildExtMacos(UnifiedBuildExt):
         build_path = glob.glob(os.path.join("build", "**", library_name), recursive=True)
         lib_with_r_path = "@rpath/libcustom_calls.so"
 
-        original_path = build_path[0] if build_path else frontend_path[0]
-
         # Run install_name_tool to modify LC_ID_DYLIB(other the rpath stays in vars/folder)
-        subprocess.run(
-            ["/usr/bin/install_name_tool", "-id", lib_with_r_path, original_path],
-            check=False,
-        )
+        for path in (build_path, frontend_path):
+            if path:
+                subprocess.run(
+                    ["/usr/bin/install_name_tool", "-id", lib_with_r_path, path[0]],
+                    check=False,
+                )
 
 
 Py_LIMITED_API_macros = [("Py_LIMITED_API", "0x030C0000")]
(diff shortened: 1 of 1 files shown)


paul0403 (approved):

setup.py
```diff
@@ -331,7 +331,7 @@ def run(self):
         build_path = glob.glob(os.path.join("build", "**", library_name), recursive=True)
         lib_with_r_path = "@rpath/libcustom_calls.so"
 
-        original_path = frontend_path[0] if frontend_path else build_path[0]
+        original_path = build_path[0] if build_path else frontend_path[0]
```
mehrdad2m: I wondering if this would fail when the order of installation is reversed? i.e. installing the wheel first and then make all. 
Would it make sense to run install_name_tool on both? 🤔

kipawaa: When building from source the `INSTALLED` python variable is set to `false`, and the path resolution searches static relative paths. I think this would cause the dylib resolution to fail, but maybe I'm wrong. Can test this tomorrow and report back.

kipawaa: @mehrdad2m it seems that running `install_name_tool` on both succeeds! ([4fb384f](https://github.com/PennyLaneAI/catalyst/pull/2874/commits/4fb384f8cb86624aa76a33e72bcb1cd5d737dbb9))

dime10: Isn't this writing to the same field twice?
EDIT: oh nvm it's the opposite, you're writing the same value to different libraries

dime10: Good catch on this build folder thing! I wonder if instead of us searching particular paths, shouldn't we able to find the lib we just built directly (since this is the code that builds it in the first place)?

mehrdad2m: > shouldn't we able to find the lib we just built directly

coming back to this, I think this makes sense to me. Looking at the `build_ext` class, it has a `get_ext_filename` method that we could potentially use but I am not familiar with that module enough to know for sure without testing.

dime10: I was wondering why I was seeing this line in River's PR, I thought we had already merged this 😅

dime10: But yeah I think it's worth updating this properly for good, if it doesn't take too long

kipawaa: I likely won't be able to get back to this today, but I can try to check that out tomorrow/monday.

PR #2876, opened 2026-05-26
PBC Depth: Support depth for ppm-specs across `scf.if` branches

sengthai:
**Context:**
Today the `--ppm-specs` pass reports a depth field by using the number of layers in the PBC circuit. When a function contained an `scf.if`, the previous implmentation counted the ops in the THEN and ELSE branches sequentially, as if both branches always run. 

**Description of the Change:**
In this PR, the depth analysis the worst-case across scf.if branches:
`depth(region) = layers_outside_if + Σ max(depth(then), depth(else))`

For example:

```
pbc.ppm ["X","X","Y"]         # 1 layer
scf.if:
  then: pbc.ppr["Y","X"](4)   # 2 ops on overlapping qubits
           pbc.ppr["Z","X"](4)   # don't commute -> 2 layers
  else: pbc.ppr["X","X"](4)   # 1 layer
```
The previously reported depth was 4 (everything serialized). The correct worst-case is 3 = 1 + max(2, 1).

**Related GitHub Issues:**
* Follows up #2863, which introduced the depth / depth_type fields.

[sc-120053]

PR #2877, opened 2026-05-26
PBC Depth: Support depth for ppm-specs across `scf.index_switch` branches 

sengthai:
**Context:**
This PR extends the worst-case depth analysis in `--ppm-specs` (introduced in #2876) to also handle `scf.index_switch`. Catalyst emits `scf.index_switch` from the user-facing catalyst.switch primitive and from the gridsynth PPR-basis decomposition. Previously, a function containing any of these would fail the depth analysis with worst-case depth is not available when PBC ops are inside scf.index_switch.

This PR implement to support switch-case op for worst-case depth analysis.
A switch is a multi-way scf.if. The worst-case depth across all branches (default + cases) is:
`depth(switch) = max(depth(default), depth(case_0), ..., depth(case_{N-1}))`

For example:
```
pbc.ppm ["Z"]                        # 1 layer
scf.index_switch:
  case 0: pbc.ppr ["Z"](4)           # 2 non-commuting PPRs
          pbc.ppr ["X"](4)           #   -> 2 layers
  case 1: pbc.ppr ["Y"](4)           # 1 layer
  default: pbc.ppr ["X"](4)          # 1 layer
```
Reported depth is now 1 + max(2, 1, 1) = 3 same shape as the scf.if case in the previous PR, just with N+1 regions.

**Related GitHub Issues:**
Follows up #2876  (worst-case depth across scf.if), which built on #2863 .

[sc-120053]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -43,7 +43,7 @@
   ``depth(region) = layers_outside_branch + Σ max(depth(branch_i))``. Previously branches
   were counted sequentially.
   [(#2876)](https://github.com/PennyLaneAI/catalyst/pull/2876)
-  [(#XXXX)](https://github.com/PennyLaneAI/catalyst/pull/XXXX)
+  [(#2877)](https://github.com/PennyLaneAI/catalyst/pull/2877)
 
 * The `--decompose-lowering` pass can now handle cases where the decomposed gate act on qubit values
   extracted from different quantum register SSA values, as long as all these quantum register values
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -38,12 +38,14 @@
   longer requires an explicit pipeline and no longer mixes MLIR into the JSON output.
   [(#2863)](https://github.com/PennyLaneAI/catalyst/pull/2863)
 
-* The ``depth`` field reported by :func:`~.passes.ppm_specs` is now the **worst-case quantum
-  depth** across ``scf.if`` and ``scf.index_switch`` branches:
-  ``depth(region) = layers_outside_branch + max(depth(branch_i))``. Previously branches
-  were counted sequentially.
+* The ``depth`` field reported by :func:`~.passes.ppm_specs` is now the worst-case depth
+  across ``scf.if`` and ``scf.index_switch`` branches (taking the maximum over all branches)
+  and across statically-bounded ``scf.for`` loops (multiplied by the trip count).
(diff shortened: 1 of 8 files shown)


doc/releases/changelog-dev.md
```diff
@@ -39,9 +39,11 @@
   [(#2863)](https://github.com/PennyLaneAI/catalyst/pull/2863)
 
 * The ``depth`` field reported by :func:`~.passes.ppm_specs` is now the **worst-case quantum
-  depth** across ``scf.if`` branches: ``depth(region) = layers_outside_if + Σ max(depth(then),
-  depth(else))``. Previously THEN and ELSE were counted sequentially.
+  depth** across ``scf.if`` and ``scf.index_switch`` branches:
+  ``depth(region) = layers_outside_branch + Σ max(depth(branch_i))``. Previously branches
```
```suggestion
  ``depth(region) = layers_outside_branch + max(depth(branch_i))``. Previously branches
```

Same comment as the `if` PR

mlir/lib/PBC/Utils/PBCLayer.cpp
```diff
@@ -62,25 +84,28 @@ FailureOr<int64_t> PBCLayerContext::computeWorstCaseDepth(Block *block, bool onl
...
             }
-
             depth += *branchDepth;
             continue;
         }
 
-        if (isa<scf::ForOp>(&op) || isa<scf::WhileOp>(&op)) {
-            return op.emitOpError(
-                "worst-case depth is not available when PBC ops are inside scf.for or scf.while");
-        }
```
paul0403, line 79: Let's restore the for and while guards in this PR, and only remove them in their own PRs! This way we minimize the chance of random git errors as we move across the chain of PRs.

paul0403 (approved):

mlir/lib/PBC/Utils/PBCLayer.cpp
paul0403: Especially since I assume you will not deal with while loops in this series of PRs due to its dynamic nature.

mlir/include/PBC/Utils/PBCLayer.h
paul0403, line 54: Reminder to update this description to include switch

mlir/lib/PBC/Utils/PBCLayer.cpp
sengthai: the code hasn't removed, i just moved it up. 😅
https://github.com/PennyLaneAI/catalyst/blob/025cffdf8f131bcd869b044a41c898819504ec2f/mlir/lib/PBC/Utils/PBCLayer.cpp#L87

PR #2879, opened 2026-05-26
PBC Depth: Support depth for ppm-specs on static `scf.for`

sengthai:
**Context:**
Following #2876 and #2877, the --ppm-specs pass still errored when any PBC op appeared inside `scf.for`.

In this PR, static `scf.for` loops (all three bounds are arith.constant) contribute to worst-case depth as:
`depth(scf.for, static N) = N × depth(body)`

Dynamic-bound `scf.for` and `scf.while` still error. `scf.while` is intentionally out of scope. its trip count is data-dependent.

**Related GitHub Issues:**
Follows up #2876  (worst-case depth across scf.if), #2877 (scf.index_switch), and #2863  (original depth/depth_type fields).

### DIFF
--- a/mlir/include/PBC/Utils/PBCLayer.h
+++ b/mlir/include/PBC/Utils/PBCLayer.h
@@ -62,6 +62,9 @@ class PBCLayerContext {
     // Worst-case depth of an `scf.index_switch`: `max(depth(default), depth(case_i))`
     mlir::FailureOr<int64_t> switchWorstCaseDepth(mlir::scf::IndexSwitchOp switchOp,
                                                   bool onlyOnDisjointQubit);
+    // Worst-case depth of a static `scf.for`: `N * depth(body)`.
+    // Returns failure if the trip count is dynamic.
+    mlir::FailureOr<int64_t> forWorstCaseDepth(mlir::scf::ForOp forOp, bool onlyOnDisjointQubit);
 };
 
 class PBCLayer {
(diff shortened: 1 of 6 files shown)


dime10:
The title says depth-1, but hopefully this address both depth-0 and 1?

sengthai:
> The title says depth-1, but hopefully this address both depth-0 and 1?

@dime10  title is confusing. Actually, this PR works both depth types. 😅

mlir/include/PBC/Utils/PBCLayer.h
paul0403, line 55: Reminder to update this description to include for ops.

mlir/lib/PBC/Utils/PBCLayer.cpp
```diff
@@ -71,6 +72,21 @@ FailureOr<int64_t> PBCLayerContext::switchWorstCaseDepth(scf::IndexSwitchOp swit
...
 
+FailureOr<int64_t> PBCLayerContext::forWorstCaseDepth(scf::ForOp forOp, bool onlyOnDisjointQubit)
+{
+    int64_t iterations = countStaticForOpIterations(forOp);
+    if (iterations == -1) {
+        return forOp.emitOpError(
+            "worst-case depth is not available when there are dynamically sized for loops");
+    }
+
+    FailureOr<int64_t> bodyDepth = computeWorstCaseDepth(forOp.getBody(), onlyOnDisjointQubit);
```
paul0403, line 83: Based on the recursive structure this seems to work for nested for loops. Can you add a test for nested for loops?

sengthai: Yes, we already had a nested for-loop. 🙌
https://github.com/PennyLaneAI/catalyst/blob/0bfb99d6204d6066c81974898bf78b1edf94ae60/mlir/test/PBC/PPMSpecsTest.mlir#L696-L712

PR #2883, opened 2026-05-27
Add `quantum.operator` to Catalyst [Operator2 work]

dime10:
Provides the new MLIR operation for advanced quantum operators as defined by this [ADR](https://github.com/PennyLaneAI/adrs/blob/master/documents/094-new-operator-abstractions.md). The following is provided in this PR:
- the operation definition
- op verifiers
- custom assembly syntax
- integration with the `ParametrizedGate` interface

Future PRs will tackle:
- updates to existing passes (if necesary)
- reference-semantic version of the op + conversion
- lowering to the op from the frontend

[sc-120860]


mlir/include/Quantum/IR/QuantumOps.td
```diff
@@ -534,6 +536,170 @@ def CustomOp : UnitaryGate_Op<"custom", [DifferentiableGate, NoMemoryEffect,
...
+        strings). Frontend-level data arguments are compressed into a unique identifier, since
+        in general they may not be lowerable to MLIR.
+
+        The operation has two mutually exclusive representations:
+
+        * **Qubit mode** (`in_qubits` / `out_qubits`):
+          Use this when the operator acts on a statically known set of wires. Qubits are
+          threaded through the op explicitly and can carry per-control values through
+          `in_ctrl_qubits` / `in_ctrl_values` (with matching `out_ctrl_qubits`).
+          This is the natural representation for fixed-arity gates and operators.
```
mudit2812: arity?

dime10: Generally, the number of inputs a mathematical operator acts on, in this case wires. Honestly we can just remove that sentence though:
```suggestion
```

mlir/lib/Quantum/IR/QuantumOps.cpp
mudit2812, line 1: Printer and parser are thick 😬

mlir/test/Quantum/VerifierTest.mlir
```diff
@@ -522,3 +522,330 @@ func.func @qnode_without_measurement(%q : !quantum.bit) attributes {quantum.node
...
+// -----
+
+func.func @operator_basic_qubits(%q0 : !quantum.bit, %q1 : !quantum.bit) {
+    %o0, %o1 = "quantum.operator"(%q0, %q1) <{op_name = "basic_qubits", operandSegmentSizes = array<i32: 0, 0, 2, 0, 0, 0, 0, 0, 0>, resultSegmentSizes = array<i32: 2, 0, 0>}> : (!quantum.bit, !quantum.bit) -> (!quantum.bit, !quantum.bit)
+    return
+}
+
+//////////////////////
+// quantum.operator //
+//////////////////////
```
mudit2812:
```suggestion
//////////////////////
// quantum.operator //
//////////////////////

func.func @operator_basic_qubits(%q0 : !quantum.bit, %q1 : !quantum.bit) {
    %o0, %o1 = "quantum.operator"(%q0, %q1) <{op_name = "basic_qubits", operandSegmentSizes = array<i32: 0, 0, 2, 0, 0, 0, 0, 0, 0>, resultSegmentSizes = array<i32: 2, 0, 0>}> : (!quantum.bit, !quantum.bit) -> (!quantum.bit, !quantum.bit)
    return
}
```

mudit2812: Just swapping test and the /////// comment

mlir/lib/Quantum/IR/QuantumOps.cpp
```diff
@@ -568,3 +705,492 @@ LogicalResult AdjointOp::verify()
...
+                llvm::zip(getForwardArgs(), getForwardArgs().getTypes()), p,
+                [&](auto pair) { p << std::get<0>(pair) << ": " << std::get<1>(pair); });
+            p << ")";
+        }
+    }
+
+    // 7. Quantum register
+    if (getInQreg()) {
+        p.printNewline();
+        p << "quregs(" << getInQreg() << ") indices(";
```
mudit2812, line 808: I've always found the inconsistency of using both `qureg` and `qreg` to be kind of weird. Though I haven't seem `qureg(s)` for operand/result names or in assembly formats before, so maybe this should just be `qregs`?
```suggestion
        p << "qregs(" << getInQreg() << ") indices(";
```

mudit2812: This is purely stylistic though, so it's a non-blocking comment

mlir/test/Quantum/VerifierTest.mlir
```diff
@@ -522,3 +522,330 @@ func.func @qnode_without_measurement(%q : !quantum.bit) attributes {quantum.node
...
+func.func @operator_with_decompositions(%q : !quantum.bit) {
+    %out = "quantum.operator"(%q) <{op_name = "with_decompositions", decompositions = [@decomp_a, @decomp_b], operandSegmentSizes = array<i32: 0, 0, 1, 0, 0, 0, 0, 0, 0>, resultSegmentSizes = array<i32: 1, 0, 0>}> : (!quantum.bit) -> !quantum.bit
+    return
+}
+
+// -----
+
+func.func @operator_custom_with_decompositions(%q : !quantum.bit) {
+    %out = quantum.operator "custom_with_decompositions"() qubits(%q)
+      decompositions = [@decomp_a, @decomp_b]
```
mudit2812: I would've expected MLIR to have automatic verification about whether symbols exist in the IR or not. Should that be included in the verifier?

mlir/test/Quantum/VerifierTest.mlir
```diff
@@ -522,3 +522,330 @@ func.func @qnode_without_measurement(%q : !quantum.bit) attributes {quantum.node
...
+    %out = quantum.operator "custom_bad_register_map_entry"() quregs(%r)
+      indices(%idx0 : tensor<2xi64>, %idx1 : tensor<1xi64>)
+      param_map = {}
+      qubit_map = {qi0 = [0, 1], qi1 = [2]}
+    return
+}
+
+// -----
+
+func.func @operator_invalid_forward_args_without_uid(%fwd : i64, %q0 : !quantum.bit, %q1 : !quantum.bit) {
```
mudit2812, line 760: Should there be assembly format tests for this and following tests?

mudit2812 (commented):
Looks great @dime10 . My only blocking comment is about testing invalid operators with the custom format

mlir/lib/Quantum/IR/QuantumOps.cpp
dime10: Yeah I share the frustration. On the one hand, "qureg" is more analogous to "qubit", on the other it is perhaps a bit more unconventional? If we look at precedent in Catalyst, the official name in the type system is `QubitType` and `QuregType` (written as quantum.bit and quantum.reg in the IR, avoiding the conundrum). `qreg` does appear as certain op argument/result names though, and maybe as program variables.

mlir/test/Quantum/VerifierTest.mlir
dime10: Good idea! There is also a SymbolUser interface, although I'd have to check to see if it allows for optional symbols

mlir/test/Quantum/VerifierTest.mlir
dime10: I don't think we gain from testing invalid states in custom assembly necessarily, unless the custom assembly enables invalid states that the generic form doesn't allow?

mlir/include/Quantum/IR/QuantumOps.td
```diff
@@ -534,6 +536,170 @@ def CustomOp : UnitaryGate_Op<"custom", [DifferentiableGate, NoMemoryEffect,
...
 }
 
+def OperatorOp : UnitaryGate_Op<"operator", [ParametrizedGate, NoMemoryEffect,
+                                             AttrSizedOperandSegments, AttrSizedResultSegments]> {
+    let summary = "A generalized quantum operation with arbitrary input signature.";
+    let description = [{
+        The `quantum.operator` is a generic representation for arbitrary quantum operations
+        that cannot be expressed by the simpler `quantum.custom` op and don't have a dedicated
+        operation.
+        It's purpose is to represent high-level operations from frontend languages such as
```
kipawaa:
```suggestion
        Its purpose is to represent high-level operations from frontend languages such as
```

mlir/include/Quantum/IR/QuantumAttrDefs.td
```diff
@@ -40,4 +40,20 @@ def NamedObservableAttr : EnumAttr<QuantumDialect, NamedObservable, "named_obser
...
+class DictValuesOfType<string attrCppType> : AttrConstraint<
+    CPred<"::llvm::all_of("
+            "::llvm::cast<::mlir::DictionaryAttr>($_self),"
+            "[](::mlir::NamedAttribute na) {"
+              "return ::llvm::isa<::mlir::" # attrCppType # ">(na.getValue());"
+            "}"
+          ")">,
+    "dictionary with " # attrCppType # " values"
+>;
+def I64ArrayDictAttr : ConfinedAttr<DictionaryAttr, [DictValuesOfType<"DenseI64ArrayAttr">]>;
```
paul0403, line 53:
```suggestion
>;

def I64ArrayDictAttr : ConfinedAttr<DictionaryAttr, [DictValuesOfType<"DenseI64ArrayAttr">]>;
```

paul0403:
Do you mind running `make dialect-docs` locally and commit the built doc files (just for `QuantumOps.md` is fine) too? It would be much easier to look at when reviewing :sweat_smile: 

Non blocking if it turns out to be too much trouble

dime10:
> Do you mind running make dialect-docs locally and commit the built doc files (just for QuantumOps.md is fine) too? It would be much easier to look at when reviewing 😅

Great idea, thanks!

https://github.com/PennyLaneAI/catalyst/pull/2883/commits/5937954d2a2777f6a1ee3d8d295970c2573ed2b8, it hadn't been updated in a while, so there are few additional changes in there.

mlir/test/Quantum/VerifierTest.mlir
mudit2812: Gemini says yes, but I always take AI answers about MLIR with a grain of salt

kipawaa: It might be preferable to _not_ verify that the decomposition rule symbols exist in the IR? I think in contexts like the graph solver, we may want to provide the name of a precompiled decomposition rule to indicate that it's applicable to this op, but it won't exist in the IR until the `graph-decomposition` pass is run.

dime10: I think the idea is that the symbols won't be there unless the rules have been generated already. Especially since there can be multiple and they need to be uniqued. I would have to examine this workflow more closely though:
> I think in contexts like the graph solver, we may want to provide the name of a precompiled decomposition rule to indicate that it's applicable to this op

mlir/include/Quantum/IR/QuantumOps.td
```diff
@@ -534,6 +536,170 @@ def CustomOp : UnitaryGate_Op<"custom", [DifferentiableGate, NoMemoryEffect,
     let hasVerifier = 1;
 }
 
+def OperatorOp : UnitaryGate_Op<"operator", [ParametrizedGate, NoMemoryEffect,
```
paul0403, line 539: Is this `UnitaryGate_Op` or just `GateOp`? I would think not all templates will be unitary, e.g. state prep templates exist.

paul0403: This might be quite tricky... since I agree you do want all the `QuantumGate` interface methods on here on top of the basic ones in `QuantumOperation` 🤔

mlir/include/Quantum/IR/QuantumOps.td
```diff
@@ -534,6 +536,170 @@ def CustomOp : UnitaryGate_Op<"custom", [DifferentiableGate, NoMemoryEffect,
...
+
+        An operator acting on a dynamic number of qubits, expressed in register mode,
+        with explicit name-to-position maps for its parameters and wire arguments:
+
+        ```mlir
+        %out_qreg = quantum.operator "MyOp"(%state: tensor<?xcomplex<f64>>)
+            quregs(%qreg) indices(%w0: tensor<?xi64>, %w1: tensor<1xi64>)
+            param_map = {state_vector = [0]}
+            qubit_map = {wires = [0], precision_wire = [1]}
+        ```
```
paul0403, line 606: I assume the names like `state_vector` and `precision_wire` will be generated in the frontend during plxpr conversion, and here we are just defining the op to accept them? And I guess the idea is to allow arbitrary python template argument names, so we can deal with a general template?

mlir/include/Quantum/IR/QuantumOps.td
paul0403, line 539: The `NoMemoryEffect` trait also scares me a bit. Templates might have qubit allocations inside?

But, as long as the template allocs and deallocs within its scope, probably this is still fine? Because I guess the spirit of `NoMemoryEffect` is "op can be removed if no user in dataflow exists", which is still true if the temporary wires allocated within the template are freed in the same template?

mlir/include/Quantum/IR/QuantumOps.td
```diff
@@ -534,6 +536,170 @@ def CustomOp : UnitaryGate_Op<"custom", [DifferentiableGate, NoMemoryEffect,
...
+
+        An operator carrying frontend-level data (summarized by `UID`), together with
+        pre-traced decomposition rules referenced through `decompositions` and extra
+        dynamic data forwarded into those rules via `forward(...)`:
+
+        ```mlir
+        %out:3 = quantum.operator "MyOp"(%state: tensor<8xcomplex<f64>>) qubits(%q0, %q1, %q2)
+            UID(324958) forward(%phi: f64, %delta: f64)
```
paul0403, line 613: The UID will be a hash from frontend? I guess in situations where an operator take in unlowerable arbitrary python objects as inputs, e.f. other operators as inputs like QSVT?

In that case, probably we need to make sure in frontend that different python objects hash to different numbers, i.e. the hash function needs to be one-to-one. If hash collisions exist (i.e. two python objects have the same hash), two `quantum.operator` ops here might be mistakenly judged as equivalent, despite having different other operators as args!

paul0403, line 613: Just trying to understand what `forward` means. Does it mean that `%phi` and `%delta` are both (a) operands of the current `quantum.operator` op, and (b) args to the decomp function referenced?

mlir/lib/Quantum/IR/QuantumOps.cpp
```diff
@@ -568,3 +705,492 @@ LogicalResult AdjointOp::verify()
...
+    return success();
+}
+
+ParseResult OperatorOp::parse(OpAsmParser &parser, OperationState &result)
+{
+    Builder &builder = parser.getBuilder();
+    MLIRContext *ctx = parser.getContext();
+
+    // 1. Parse operation name string: "foo"
+    std::string opName;
```
paul0403, line 895: Wow, `StringProp` allows using the field as `std::string` directly? 🤩 

Probably the idea upstream was to save the construction of an attribute object?

mlir/include/Quantum/IR/QuantumOps.td
```diff
@@ -534,6 +536,170 @@ def CustomOp : UnitaryGate_Op<"custom", [DifferentiableGate, NoMemoryEffect,
...
+        Variadic<AnyType>:$forward_args,
+
+        // traditional qubit specification
+        Variadic<QubitType>:$in_qubits,
+        Variadic<QubitType>:$in_ctrl_qubits,
+        Variadic<I1>:$in_ctrl_values,
+
+        // alternative spec for acting on a dynamic number of qubits
+        Optional<QuregType>:$in_qreg,
+        Variadic<1DTensorOf<[I64]>>:$arr_qubit_indices,
```
paul0403, line 629: Why is it variadic of tensor? Wouldn't there be only one tensor, with each entry of the tensor being a wire index?

mlir/test/Quantum/VerifierTest.mlir
```diff
@@ -522,3 +522,330 @@ func.func @qnode_without_measurement(%q : !quantum.bit) attributes {quantum.node
...
+    %o0, %o1 = "quantum.operator"(%p0, %p1, %q0, %q1) <{op_name = "qubit_maps", param_map = {p0 = array<i64: 0>, p1 = array<i64: 1>}, qubit_map = {pair = array<i64: 0, 1>}, operandSegmentSizes = array<i32: 2, 0, 2, 0, 0, 0, 0, 0, 0>, resultSegmentSizes = array<i32: 2, 0, 0>}> : (f64, i64, !quantum.bit, !quantum.bit) -> (!quantum.bit, !quantum.bit)
+    return
+}
+
+// -----
+
+func.func @operator_custom_qubits_with_maps(%p0 : f64, %p1 : i64, %q0 : !quantum.bit, %q1 : !quantum.bit) {
+    %o0, %o1 = quantum.operator "custom_qubit_maps"(%p0 : f64, %p1 : i64) qubits(%q0, %q1)
+      param_map = {p0 = [0], p1 = [1]}
+      qubit_map = {pair = [0, 1]}
```
paul0403, line 604: Confirming my understanding again, this means:
- The frontend variable named `p0` corresponds to index 0 in the param list, i.e. `%p0`; same for `p1`
- The frontend variable named `pair` corredponds to indices `0, 1` in the qubits list, i.e. `%q0, %q1`?

mlir/test/Quantum/VerifierTest.mlir
```diff
@@ -522,3 +522,330 @@ func.func @qnode_without_measurement(%q : !quantum.bit) attributes {quantum.node
...
+    return
+}
+
+// -----
+
+func.func @operator_custom_registers_with_maps(%p0 : f64, %p1 : i64, %r : !quantum.reg, %idx0 : tensor<2xi64>, %idx1 : tensor<1xi64>) {
+    %out = quantum.operator "custom_reg_maps"(%p0 : f64, %p1 : i64)
+      quregs(%r) indices(%idx0 : tensor<2xi64>, %idx1 : tensor<1xi64>)
+      param_map = {p0 = [0], p1 = [1]}
+      qubit_map = {qi0 = [0], qi1 = [2]}
```
paul0403: How come there's a index `2` in the `qubit_map`, when there's only 2 items in `arr_qubit_indices` here?

mlir/test/Quantum/VerifierTest.mlir
```diff
@@ -522,3 +522,330 @@ func.func @qnode_without_measurement(%q : !quantum.bit) attributes {quantum.node
...
+    %out = quantum.operator "custom_with_decompositions"() qubits(%q)
+      decompositions = [@decomp_a, @decomp_b]
+    return
+}
+
+// -----
+
+func.func @operator_custom_with_uid_and_forward(%fwd : i64, %q0 : !quantum.bit, %q1 : !quantum.bit) {
+    %o0, %o1 = quantum.operator "custom_uid_forward"() qubits(%q0, %q1)
+      UID(7) forward(%fwd : i64)
```
paul0403, line 662: This got me thinking, are python hashes guaranteed to be within `i64`?

paul0403: Because if not, overflowing during lowering, when we lower big python ints to `I64Prop` in mlir, might lead to extra collisions.

mlir/include/Quantum/IR/QuantumOps.td
paul0403: Oh! Is it because a template can have multiple python fields that will be interpreted as (different kinds of) wires(, e.g. data wires and work wires)?

mlir/test/Quantum/VerifierTest.mlir
```diff
@@ -522,3 +522,330 @@ func.func @qnode_without_measurement(%q : !quantum.bit) attributes {quantum.node
...
+func.func @operator_with_registers(%r : !quantum.reg, %idx0 : tensor<2xi64>, %idx1 : tensor<1xi64>) {
+    %out = "quantum.operator"(%r, %idx0, %idx1) <{op_name = "with_registers", operandSegmentSizes = array<i32: 0, 0, 0, 0, 0, 1, 2, 0, 0>, resultSegmentSizes = array<i32: 0, 0, 1>}> : (!quantum.reg, tensor<2xi64>, tensor<1xi64>) -> !quantum.reg
+    return
+}
+
+// -----
+
+func.func @operator_custom_with_registers(%r : !quantum.reg, %idx0 : tensor<2xi64>, %idx1 : tensor<1xi64>) {
+    %out = quantum.operator "custom_with_registers"()
+      quregs(%r) indices(%idx0 : tensor<2xi64>, %idx1 : tensor<1xi64>)
```
paul0403, line 572: Since qreg is `Optional<QuregType>:$in_qreg`, instead of `Variadic`, should we just do `qureg` in the custom format, i.e. no plural?

mlir/lib/Quantum/IR/QuantumOps.cpp
```diff
@@ -212,6 +213,142 @@ LogicalResult CustomOp::verify()
...
+
+    if (hasQregInput != hasQregOutput) {
+        return emitOpError() << "in_qreg and out_qreg must either both be present or absent";
+    }
+
+    if (getInQubits().size() != getOutQubits().size()) {
+        return emitOpError() << "number of input qubits (" << getInQubits().size()
+                             << ") and output qubits (" << getOutQubits().size()
+                             << ") must be the same";
+    }
```
paul0403: Isn't this verifier already on the `QuantumOperation` interface?

mlir/lib/Quantum/IR/QuantumOps.cpp
```diff
@@ -212,6 +213,142 @@ LogicalResult CustomOp::verify()
...
+
+    if (!getForwardArgs().empty() && !getUID()) {
```
paul0403, line 236: To me XOR is more readable, but it's probably just preference. Feel free to leave as is.

mlir/lib/Quantum/IR/QuantumOps.cpp
```diff
@@ -212,6 +213,142 @@ LogicalResult CustomOp::verify()
...
+        return emitOpError() << "number of input control qubits (" << getInCtrlQubits().size()
+                             << ") and output control qubits (" << getOutCtrlQubits().size()
+                             << ") must be the same";
+    }
+
+    if (getInCtrlQubits().size() != getInCtrlValues().size()) {
+        return emitOpError() << "number of input control qubits (" << getInCtrlQubits().size()
+                             << ") and control values (" << getInCtrlValues().size()
+                             << ") must be the same";
+    }
```
paul0403: Same comment as above, these are already checked on the interface right?

mlir/lib/Quantum/IR/QuantumOps.cpp
```diff
@@ -212,6 +213,142 @@ LogicalResult CustomOp::verify()
...
+            for (int64_t idx : denseArray.asArrayRef()) {
+                if (idx < 0 || idx >= static_cast<int64_t>(numParams)) {
+                    return emitOpError() << "param_map index is out of bounds with respect to "
+                                            "params: "
+                                         << idx << " is not in [0, " << numParams << ")";
+                }
+                if (!coveredParams[static_cast<size_t>(idx)]) {
+                    coveredParams[static_cast<size_t>(idx)] = true;
+                    ++coveredCount;
+                }
```
paul0403, line 281: Just use std::set? Saves an awkward `coveredCount` variable.Same for qubit map cover below.

mlir/test/Quantum/VerifierTest.mlir
```diff
@@ -522,3 +522,330 @@ func.func @qnode_without_measurement(%q : !quantum.bit) attributes {quantum.node
...
+    %o0, %o1 = quantum.operator "custom_bad_qubit_map_union"() qubits(%q0, %q1)
+      param_map = {}
+      qubit_map = {a = [0], b = [0]}
+    return
+}
+
+// -----
+
+func.func @operator_invalid_register_qubit_map_entry_width(%r : !quantum.reg, %idx0 : tensor<2xi64>, %idx1 : tensor<1xi64>) {
+    // expected-error@+1 {{each qubit_map entry can only contain one index in register mode}}
```
paul0403: Why is this the case? Why can qubit mode have multiple indices for each python wire variable?

mlir/lib/Quantum/IR/QuantumOps.cpp
```diff
@@ -568,3 +705,492 @@ LogicalResult AdjointOp::verify()
...
+    NamedAttrList genericAttrs;
+    if (parser.parseOptionalAttrDict(genericAttrs)) {
+        return failure();
+    }
+    result.addAttributes(genericAttrs);
+
+    // 6. Optional Python-only metadata: UID(...) [forward(...)].
+    if (succeeded(parser.parseOptionalKeyword("UID"))) {
+        int64_t uid = 0;
+        if (parser.parseLParen() || parser.parseInteger(uid) || parser.parseRParen()) {
```
paul0403, line 962: Why do some fields have the left/right parenthesis parser, but others do not? 🤔

mlir/lib/Quantum/IR/QuantumOps.cpp
```diff
@@ -568,3 +705,492 @@ LogicalResult AdjointOp::verify()
...
+                return success();
+            };
+            if (parser.parseCommaSeparatedList(OpAsmParser::Delimiter::Paren, parseCtrlValue)) {
+                return failure();
+            }
+        }
+    }
+
+    // 9. Optional inherent metadata blocks.
+    while (true) {
```
paul0403: Do we need the while loop? Is there an issue with just 5 if blocks? Because it would be a binary "either exists or not" for all 5 optional fields, with no chance of repeating, right?

paul0403: As in, there will not be two `decomposition` attributes, for example

paul0403 (commented):
What a PR! 💪 💪 

Majority of my comments are just me being new to the operator 2.0 work and trying to understand things. Overall code structure is very nice already!

mlir/include/Quantum/IR/QuantumOps.td
dime10: It's true that not all PL ops lowered to this operation might be unitary, but we decided to allow direct `adjoint` and `control` attributes/operands on the MLIR op, which means that it satisfies the interface in principle even if not on all instances.
We haven't really decided how we would handle those cases, but the `UnitaryGate_Op` base class should be there as long as the `adjoint` and `control` arguments are there.

mlir/include/Quantum/IR/QuantumOps.td
dime10: Yes this is mostly meta-info that we may or may not find useful. It should help passes identify relevant arguments whose name they know from the frontend definition.

mlir/include/Quantum/IR/QuantumOps.td
dime10: > Templates might have qubit allocations inside?

The template wouldn't, a decomposition certainly could, but that applies to any gate like quantum.custom or quantum.unitary etc. The `NoMemoryEffect` means that from the exterior, this op has no side-effects, and if the results are unused it can be removed (for example). All our gates are modelled in this way (value-semantics).

mlir/include/Quantum/IR/QuantumOps.td
dime10: > The UID will be a hash from frontend? 

The implementation is not prescribed (its generation may or may not involve hashing), but the UID is indeed provided from the frontend as a differentiator of Python-only data arguments between instances of the same op.

This is why I insisted on renaming this field (originally called hash), because conceptually we are talking about a UID and not a hash. The latter has collisions but the former does not. Whatever value is provided here by the frontend will be treated as the source of truth in the compiler. Whether the frontend provides a good UID or not (e.g. one that is frequently not unique) is a separate concern.

dime10: These are program values that conceptually aren't part of the operator itself, but need to be forwarded to its decomposition function. It contains any dynamic data that was embed in the Python-only data args.

mlir/lib/Quantum/IR/QuantumOps.cpp
dime10: I believe regular attribute fields typically already generate both an `Attribute` accessor and "C++ type" accessor. For example these methods for an optional i64 attribute in the `AllocOp`:
```
  ::mlir::IntegerAttr getNqubitsAttrAttr();
  ::std::optional<uint64_t> getNqubitsAttr();
```
The attribute version is always appended with `Attr` (note in this case the actual name also had "attr" in it (nqubits_attr), slightly confusing).

The idea with properties instead of attributes is that they would be stored inline in the op storage, rather than uniqued in a global attribute storage. Since they don't use the attribute system I'm guessing the accessors always produce the C++ type directly, but don't quote me on that.

mlir/include/Quantum/IR/QuantumOps.td
dime10: Exactly 👍 This means wires specified to the operator as separate arrays don't need to be merged during the lowering

mlir/test/Quantum/VerifierTest.mlir
dime10: correct 👍

mlir/test/Quantum/VerifierTest.mlir
dime10: That's a great question, I thought I added some bounds checking as well for these indices, let me double check.

mlir/test/Quantum/VerifierTest.mlir
dime10: Let's forget the notion of hashing at this stage, all we have is a unique 64-bit integer representing the unrepresentable data. 64-bits should definitely be enough to cover all unique parameter instantiations in a program.

mlir/test/Quantum/VerifierTest.mlir
paul0403: Ah, I looked at it closer, the qubit map index check was only there for qubit mode!

mlir/test/Quantum/VerifierTest.mlir
dime10: Yeah I brought this up as well, the register argument should be variadic imo.
I guess I'm doing a little forsighting here to allow quregs to have multiple values in the future without changing the assembly format.

doc/code/dialects/Quantum/QuantumAttributes.md
```diff
@@ -0,0 +1,19 @@
+<!-- Autogenerated by mlir-tblgen; don't manually edit -->
```
paul0403, line 1: Remember to add a changelog entry (and your name)!

mlir/lib/Quantum/IR/QuantumOps.cpp
dime10: What do you mean? I don't think `!A and !B` has the same truth table as `A xor B`

mlir/lib/Quantum/IR/QuantumOps.cpp
dime10: True I considered it, this should be faster but I can use a set

mlir/test/Quantum/VerifierTest.mlir
dime10: In qubit mode, the Python operator will typically only have one name for the wires (e.g. wires), but we map them to multiple ssa args. 
For register mode I thought this would not be the case originally (since we get all the indices in one array), but I didn't consider that we could have a Pytree of qubit indices, which in fact would mean multiple ssa values for one named arg. I need to update this.

mlir/lib/Quantum/IR/QuantumOps.cpp
dime10: Hmm, it's true we do use two styles here:
- named separator for additional args we've traditionally used: `name(args)`
- list attributes (no SSA values) for this op: `name = [values]`

UID doesn't wrap SSA values, so I'm not sure why I used `()` for it instead of ` = `, probably because it wasn't a list?

Do you have any suggestions for what could be improved? 😅

mlir/lib/Quantum/IR/QuantumOps.cpp
dime10: The while loop allows parsing these in any order, not sure whether we need that though 🤔

mlir/lib/Quantum/IR/QuantumOps.cpp
paul0403: Nevermind, I somehow forgot they could both be absent 🤦

mlir/test/Quantum/VerifierTest.mlir
dime10: So given what we discussed, the attached rules are going to be removed entirely.

dime10: 4b3312f6fe3955477d0253c3d420af6ad21affbc

mlir/test/Quantum/VerifierTest.mlir
dime10: 5b1caae51987ac69f8c9aecaace7d5d9c43feb51

mlir/include/Quantum/IR/QuantumOps.td
dime10: 5b1caae51987ac69f8c9aecaace7d5d9c43feb51

mlir/include/Quantum/IR/QuantumOps.td
dime10: 5b1caae51987ac69f8c9aecaace7d5d9c43feb51

mlir/include/Quantum/IR/QuantumAttrDefs.td
dime10: 5b1caae51987ac69f8c9aecaace7d5d9c43feb51

mlir/lib/Quantum/IR/QuantumOps.cpp
paul0403: ~~Hmm that's true, I think the default printer will always print attributes in alphabetical order (don't quote me on that), but of course we should allow arbitrary order because any pass can create these ops with for example attributes in some custom-ordered containers.~~ Never mind this is parser not printer. All good then :+1:

mlir/lib/Quantum/IR/QuantumOps.cpp
dime10: 04ec0bace7a14b3fd51f955b07b2f6976e846b90

mlir/lib/Quantum/IR/QuantumOps.cpp
dime10: 04ec0bace7a14b3fd51f955b07b2f6976e846b90

mlir/lib/Quantum/IR/QuantumOps.cpp
dime10: 04ec0bace7a14b3fd51f955b07b2f6976e846b90

mlir/lib/Quantum/IR/QuantumOps.cpp
dime10: Removed it anyway since it's probably overkill: 04ec0bace7a14b3fd51f955b07b2f6976e846b90

mlir/test/Quantum/VerifierTest.mlir
dime10: 307f72b8fedafda60e90ba89422d50ac246b6686

mlir/test/Quantum/VerifierTest.mlir
dime10: 307f72b8fedafda60e90ba89422d50ac246b6686

doc/code/dialects/Quantum/QuantumAttributes.md
dime10: cd210bada4ec1e6ed991e088c4bebea6fd9d62c9

mudit2812 (approved):

mlir/lib/Quantum/IR/QuantumOps.cpp
paul0403: No I think this is fine 👍

paul0403 (approved):
🥳

PR #2884, opened 2026-05-28
Fix `ppm_specs` depth when extract is independent of prior PBC op

sengthai:
**Context:**
`ppm_specs` depth depends on how PBC ops are grouped into layers. The old layer check rejected a merge whenever a `quantum.extract` appeared after a layer op in the IR, even when the extract read the original register and had no real dependency on that op.

Replace textual extract/insert ordering checks with an SSA backward walk: if a new op’s quantum.extract operand reads a register that traces back to a layer op result (via insert -> extract), the layers split.

**Example:**
Now, %4 and %6 are merged in the same layer (depth=1).
```mlir
%4 = pbc.ppr ... %1, %2, %3
%5 = quantum.extract %0[3] ...   // original register, not updated by %4
%6 = pbc.ppr ... %4#1, %4#2, %5
```

Still splits (depth = 2)
```mlir
%2 = pbc.ppr ... %1
%3 = quantum.insert %0[0], %2
%4 = quantum.extract %3[0] ...   // register updated by %2
%5 = pbc.ppr ... %4
```

[sc-120054]


doc/releases/changelog-dev.md
```diff
@@ -76,6 +76,11 @@
   folder being created and the files printed outside in the main directory.
   [(#2807)](https://github.com/PennyLaneAI/catalyst/pull/2807)
 
+* Fixed incorrect ``depth`` in :func:`~.passes.ppm_specs` when a ``quantum.extract`` appeared
+  after a PBC op but read from a register not updated by that op. Layer grouping now checks
+  data dependencies through insert to extract chains instead of textual op ordering.
+  [(#XXXX)](https://github.com/PennyLaneAI/catalyst/pull/XXXX)
```
```suggestion
  [(#2884)](https://github.com/PennyLaneAI/catalyst/pull/2884)
```

paul0403 (approved):
Nice 👍

doc/releases/changelog-dev.md
```diff
@@ -76,6 +76,11 @@
   folder being created and the files printed outside in the main directory.
   [(#2807)](https://github.com/PennyLaneAI/catalyst/pull/2807)
 
+* Fixed incorrect ``depth`` in :func:`~.passes.ppm_specs` when a ``quantum.extract`` appeared
```
rniczh, line 83: Have we need to consider the following case?
```
%4:3 = pbc.ppr ["X","X","X"](4) %1, %2, %3
%inserted  = quantum.insert %0[0], %4#0
%5   = quantum.extract %inserted[3] : !quantum.reg -> !quantum.bit
%6   = pbc.ppr ["Y"](4) %5
```
EDITED: where %1, %2, %3 are wire 0, 1, 2, respectively and %5 is from wire 3

mlir/lib/PBC/Utils/PBCLayer.cpp
```diff
@@ -331,40 +342,47 @@ bool PBCLayer::isSameBlock(PBCOpInterface op) const
     return op->getBlock() == ops.back()->getBlock();
 }
 
-// Check if the op has extract op that must be occurred before the operations in layers
-bool PBCLayer::extractsAreBeforeExistingOps(PBCOpInterface op) const
+bool PBCLayer::dependsOnLayerOps(mlir::Value value) const
```
rniczh, line 345: How you handle the region?  (e.g. scf.if)

rniczh (commented):
Thanks @sengthai, just two questions.

doc/releases/changelog-dev.md
sengthai: Yes we consider that case. I don't think the frontend generate that case, but I think we other passes in MLIR may do.

mlir/lib/PBC/Utils/PBCLayer.cpp
sengthai: each region in contains it own layer, so we start a new layer when entering the region.

rniczh: For region contains no pbc op, the `flushLayer` will help in this case.

doc/releases/changelog-dev.md
rniczh: @sengthai  Has been discussed in offline, can take it as an enhancement, nothing have to do for this PR

rniczh (approved):
💯

PR #2887, opened 2026-05-28
Fix --cp-global-memref dropping non-memref return operands

mehrdad2m:
**Context:**
`cp-global-memref` pass guards every returned memref with an `scf.if` that checks the descriptor's allocated pointer against the `0xdeadbeef` sentinel that upstream MLIR stamps onto `memref.get_global`. If the sentinel is hit, the memref is copied into a fresh `memref.alloc` so the Python host can safely free the returned buffer.
However the pass has an implicit assumption that all return values are memrefs and drops any non-memref return operands when rebuilding `func.return`, producing invalid IR. 

Note that Catalyst should not generate an IR where this pattern is exposed in the first place. This fix is just for making the pass more robust in case that pattern emerges by another pass downstream.

We now iterate positionally thorough all results where non-memrefs pass through but memrefs go through the existing copy.

**Benefits:**
Correct for mixed-type return signatures

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-120922]


mlir/lib/Quantum/Transforms/cp_global_buffers.cpp
```diff
@@ -112,19 +112,27 @@ Value allocCopyMemrefDyn(Location loc, Value memref, PatternRewriter &rewriter)
  */
 void applyCopyGlobalMemRefToReturnOp(func::ReturnOp op, PatternRewriter &rewriter)
 {
-    llvm::SmallVector<Value> memrefs = getReturnMemRefs(op);
-    if (memrefs.empty())
+    if (getReturnMemRefs(op).empty())
         return;
```
paul0403:
```suggestion
    if (getReturnMemRefs(op).empty()) {
        return;
    }
```

I like fixing these as we encounter them. https://github.com/PennyLaneAI/guidance-docs/blob/master/development/cpp-dev-guidelines.md#brace-style

paul0403 (approved):
Curious as to what prompted this, as at the end of the buffer pipeline everything should be memrefs. Is it just scalars?

mehrdad2m:
> Curious as to what prompted this, as at the end of the buffer pipeline everything should be memrefs. Is it just scalars?

Yes, that's my exception for entry point functions. In the QEC pipeline, there is a random function that returns a mix of a memref and a !quantum.reg. There was a mistake in a pass that added `llvm.emit_c_interface` to that function triggering this pass to be applied on it.

PR #2888, opened 2026-05-28
Add qecp.t op

lillian542:
**Context:**

To facilitate creating magic states on codeblocks in the QECP layer, we want to be able to apply a physical T gate on a single qubit within the codeblock

We add a `qecp.t` op


lillian542:
[sc-120969]

joeycarter (approved):
🦖

multiphaseCFD (approved):
👍

Issue #2885, opened 2026-05-28
Uncaught exception from setup/teardown functions


rniczh:
Now `SharedObjectManager.__enter__`/`__exit__` invoke the JIT-compiled `setup` an `teardown` symbols directly via `ctypes`

```python
self.setup(ctypes.c_int(argc), array_of_char_ptrs)
...
self.teardown()
```

If either function throws a C++ exception (e.g. `RT_FAIL`) the unwinder hits the ctypes C frame, which has no C++ exception handler. libc++abi falls back to `std::terminate()` and the interpreter aborts:

```
libc++abi: terminating due to uncaught exception of type
Catalyst::Runtime::RuntimeException: ...
```

An obvious way to let the exception handler can capture this error is to make nanobind wrappers for both `setup` and `teardown` just like what we did for `wrapper.wrap` for launching the entry function of the circuit.

PR #2889, opened 2026-05-28
Catch exceptions thrown from setup/teardown

rniczh:
**Context:**

If either function, setup and teardown ,throws a C++ exception like `RT_FAIL` the unwinder hits the ctypes C frame, which has no C++ exception handler. `libc++abi` falls back to `std::terminate()` and the interpreter aborts:

```
libc++abi: terminating due to uncaught exception of type
Catalyst::Runtime::RuntimeException: ...
```

Add two nanobind bindings (`wrapper.invoke_setup` / `wrapper.invoke_teardown`) that call the underlying C function pointers, and route `__enter__/__exit__` through them.

- [x] Add test 

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

https://github.com/PennyLaneAI/catalyst/issues/2885
[sc-119222]


paul0403 (approved):
Nice!

Remember to add a changelog entry (probably under improvements)

frontend/catalyst/utils/wrapper.cpp
```diff
@@ -230,13 +234,49 @@ nb::list wrap(nb::object func, nb::tuple py_args, nb::object result_desc, nb::ob
...
+    setup_fn_t fn = extract_fn_ptr<setup_fn_t>(setup_fn);
+
+    std::vector<char *> argv_c;
+    argv_c.reserve(argv.size());
+    for (auto &s : argv) {
+        argv_c.push_back(s.data());
+    }
+
+    nb::gil_scoped_release lock;
+    fn(static_cast<int>(argv.size()), argv_c.data());
```
dime10: Just in case, a little safer (against future additions) & more explicit
```suggestion
    {
        nb::gil_scoped_release lock;
        fn(static_cast<int>(argv.size()), argv_c.data());
    }
```

frontend/catalyst/utils/wrapper.cpp
```diff
@@ -230,13 +234,49 @@ nb::list wrap(nb::object func, nb::tuple py_args, nb::object result_desc, nb::ob
...
+}
+
+void invoke_teardown(nb::object teardown_fn)
+{
+    using teardown_fn_t = void (*)();
+    teardown_fn_t fn = extract_fn_ptr<teardown_fn_t>(teardown_fn);
+
+    nb::gil_scoped_release lock;
+    fn();
```
dime10:
```suggestion
    {
        nb::gil_scoped_release lock;
        fn();
    }
```

dime10 (approved):
💯

PR #2892, opened 2026-05-28
Conservative array casting in conditional

dime10:
In the context of JAX tracing, casting values to arrays automatically converts them to tracers, which means this value, and everything it interacts with, is now considered a dynamic program value. Being more conservative in this casting process, thus ensuring that static values remain static for longer, can be beneficial, e.g. for trace-time optimizations. This is especially true given that this is generally a one-way process, i.e. it is much easier to make a static value dynamic, than a dynamic value static.

[sc-121161]

### DIFF
--- a/frontend/catalyst/api_extensions/control_flow.py
+++ b/frontend/catalyst/api_extensions/control_flow.py
@@ -810,7 +810,10 @@ class CondCallable:
 
         if not self._is_any_boolean(pred):
             try:
-                pred = jnp.astype(pred, bool, copy=False)
+                if isinstance(pred, jax.core.Tracer):
+                    pred = jnp.astype(pred, bool, copy=False)
+                else:
+                    pred = bool(pred)
             except TypeError as e:
                 raise TypeError(
                     "Conditional predicates are required to be of bool, integer or float type"
--- a/frontend/catalyst/api_extensions/control_flow.py
+++ b/frontend/catalyst/api_extensions/control_flow.py
@@ -814,9 +814,9 @@ class CondCallable:
                     pred = jnp.astype(pred, bool, copy=False)
                 else:
                     pred = bool(pred)
-            except TypeError as e:
+            except (TypeError, ValueError) as e:
                 raise TypeError(
-                    "Conditional predicates are required to be of bool, integer or float type"
+                    "Conditional predicates are required to be cast-able to bool."
                 ) from e
 
         return pred
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -126,6 +126,9 @@
   - `qecp.dealloc_cb`, which deallocates a single physical codeblock.
     [(#2867)](https://github.com/PennyLaneAI/catalyst/pull/2867)
 
+* More conservative casting to tracer arrays in conditionals to preserve constant (static) values
+  better. This can be useful for optimizations that depend on values being static.
+  [(#2892)](https://github.com/PennyLaneAI/catalyst/pull/2892)
 
 
 <h3>Documentation 📝</h3>
--- a/frontend/test/pytest/test_conditionals.py
+++ b/frontend/test/pytest/test_conditionals.py
@@ -1079,34 +1079,8 @@ class TestCondPredicateConversion:
 
         assert workflow(3) == 9
 
-    def test_string_conversion_failed(self, capture_mode):
-        """Test failure at converting string to bool using Autograph."""
-
-        if capture_mode:
-            pytest.skip("works with program capture.")
-
-        @qjit(autograph=True, capture=capture_mode)
-        def workflow(x):
-            n = "fail"
-
-            if n:
-                y = x**2
-            else:
-                y = 0
-
-            return y
-
-        with pytest.raises(
-            TypeError,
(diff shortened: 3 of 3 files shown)


paul0403 (approved):

dime10:
Ah, this change has more impact than I thought. Whereas before the following:
```
        @qjit(autograph=True)
        def workflow(x):
            n = "fail"

            if n:
                y = x**2
            else:
                y = 0
```
would raise an error, because we were trying to cast a string to an array, it now succeeds because `bool("fail")` as a constant input to the catalyst if statement is totally fine.

Not sure if there are other opinions on this, but I'm leaning towards this change being fine?

paul0403:
> Ah, this change has more impact than I thought. Whereas before the following:
> 
> ```
>         @qjit(autograph=True)
>         def workflow(x):
>             n = "fail"
> 
>             if n:
>                 y = x**2
>             else:
>                 y = 0
> ```
> 
> would raise an error, because we were trying to cast a string to an array, it now succeeds because `bool("fail")` as a constant input to the catalyst if statement is totally fine.
> 
> Not sure if there are other opinions on this, but I'm leaning towards this change being fine?

I think it's fine since this is just how regular python would work too? a.k. strings being used as some sort of condition

dime10:
> > would raise an error, because we were trying to cast a string to an array, it now succeeds because `bool("fail")` as a constant input to the catalyst if statement is totally fine.
> > Not sure if there are other opinions on this, but I'm leaning towards this change being fine?
> 
> I think it's fine since this is just how regular python would work too? a.k. strings being used as some sort of condition

Turns out program capture already behaves this way too, so this is actually unifying the behaviour 👍

frontend/catalyst/api_extensions/control_flow.py
```diff
@@ -810,7 +810,10 @@ def _convert_predicate_to_bool(self, pred):
 
         if not self._is_any_boolean(pred):
             try:
-                pred = jnp.astype(pred, bool, copy=False)
+                if isinstance(pred, jax.core.Tracer):
+                    pred = jnp.astype(pred, bool, copy=False)
+                else:
+                    pred = bool(pred)
             except TypeError as e:
                 raise TypeError(
```
rniczh, line 818: `bool()` will throw an ValueError if it cannot convert to bool, we also probably need to catch the ValueError here?
```
>>> import numpy as np
>>> bool(np.array([1, 2])) 
  File "<python-input-1>", line 1, in <module>
    bool(np.array([1, 2]))
    ~~~~^^^^^^^^^^^^^^^^^^
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() ora.all()
```

rniczh (commented):
👍

frontend/catalyst/api_extensions/control_flow.py
paul0403: Or we can actually just use what plxpr is already using, by using `jnp.bool`?

https://github.com/PennyLaneAI/pennylane/blob/265354fa7fe9ab9bac27c0fc862a9c22afb727eb/pennylane/ops/op_math/condition.py#L313-L314

rniczh: looks great
```
>>> import jax
>>> jax.numpy.bool(np.array([1, 2]))
Array([ True,  True], dtype=bool)
>>> 
```

dime10: The value will still be ambiguous though, you just can't use an array as input to a conditional unless it only has one element  🤔

dime10: > Or we can actually just use what plxpr is already using, by using jnp.bool?

Doing this would defeat the entire point of this PR, which is to not cast things to jax arrays (i.e. tracers) unnecessarily.

dime10: Good catch about the ValueError though @rniczh! I wonder if we need try - catch at all, I guess it provides additional context to the error message?

paul0403: > > Or we can actually just use what plxpr is already using, by using jnp.bool?
> 
> Doing this would defeat the entire point of this PR, which is to not cast things to jax arrays (i.e. tracers) unnecessarily.

Ah yes, sorry :sweat_smile: 

We can leave this as is then, the cases uncastable to a single bool are the cases that won't work in base python too.

frontend/catalyst/api_extensions/control_flow.py
```diff
@@ -810,7 +810,10 @@ def _convert_predicate_to_bool(self, pred):
...
                     "Conditional predicates are required to be of bool, integer or float type"
```
dime10:
```suggestion
            except (TypeError, ValueError) as e:
                raise TypeError(
                    "Conditional predicates are required to be cast-able to bool."
```

frontend/catalyst/api_extensions/control_flow.py
rniczh: https://github.com/PennyLaneAI/catalyst/pull/2892/commits/74bb9006d27ef4ff4f7b21822aa1346d90b9a4e1
@dime10 Thanks!

rniczh (approved):

PR #2894, opened 2026-05-29
Add qecl -> qecp lowering for `fabricate [magic]` and associated ops

lillian542:
**Context:**

We introduce `qecl.fabricate [magic]` in the QEC Logical layer, and want to lower it to a subroutine for encoding a codeblock in the magic state in the QEC Physical dialect. This subroutine will eventually be performed by a magic state factory rather than being included in the IR, but including it allows us to validate our processes via simulator. We will not inject noise into the magic state fabrication subroutine, so this does not need to be implemented in a fault-tolerant manner.

We add lowering for putting a code-block in the magic state based on the unitary encoding for the ground-state of the codeblock. Note that the unitary rather than measurement-based encoding must be used, because the measurement-based encoding projects the codeblock into the code space, but the magic state exists outside the code space.

We also fix a bug from the previous (QEC Logical level) PR for `fabricate [magic]`, where the op is used in the `apply_T` subroutine. The conditional `SX` correction was added in the wrong order.

**Benefits:**

We can lower `qecl.fabricate` to something that can be executed on a simulator to create a codeblock in the magic state.

**Possible drawbacks:**

We don't lower `fabricate` to a magic-state distillation protocol, so its not mimicking the instructions for hardware; this could be a source of confusion if not communicated clearly. 

Also, it's likely that we will eventually lower this op to `qecp.fabricate` instead of including instructions for "fabrication" in the IR - the fabrication instructions will already be known in the magic-state factory, and we will be interfacing with it. However, for exploration and validation, this is the solution most suited to our current needs.


lillian542:
[sc-119927]

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
```diff
@@ -692,8 +692,8 @@ def create_t_subroutine(self):
 
             with ImplicitBuilder(if_apply_corr_op.true_region):
                 # This branch is for the case where a correction is needed
-                corrected_cb1 = qecl.SOp(magic_state1, idx=0)
-                corr_cb_out = qecl.PauliXOp(corrected_cb1, idx=0)
+                corrected_cb1 = qecl.PauliXOp(magic_state1, idx=0)
+                corr_cb_out = qecl.SOp(corrected_cb1, idx=0)
```
lillian542, line 987: these were incorrectly ordered in the previous PR

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecp_to_quantum.py
```diff
@@ -610,6 +611,7 @@ def test_qec_pass_ghz_nullqubit_integration(self):
 
         @qp.qjit(capture=True, pipelines=qec_pipeline())
         @convert_qecp_to_quantum_pass
+        @qp.transform(pass_name="symbol-dce")
         @convert_qecl_to_qecp_pass(qec_code="Steane", number_errors=1)
         @inject_noise_to_qecl_pass
         @qp.transform(pass_name="symbol-dce")
```
lillian542:
```suggestion
```

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecp_to_quantum.py
```diff
@@ -556,6 +556,7 @@ def test_qec_pass_ghz_lightning_integration(self, run_filecheck_qjit):
...
```
lillian542:
```suggestion
```

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecp_to_quantum.py
```diff
@@ -556,9 +556,9 @@ def test_qec_pass_ghz_lightning_integration(self, run_filecheck_qjit):
 
         @qp.qjit(capture=True, pipelines=qec_pipeline())
         @convert_qecp_to_quantum_pass
+        @qp.transform(pass_name="symbol-dce")
         @convert_qecl_to_qecp_pass(qec_code="Steane", number_errors=1)
         @inject_noise_to_qecl_pass
-        @qp.transform(pass_name="symbol-dce")
```
lillian542, line 561: There are still ops that aren't unsupported for "lowering" with `qecp-to-quantum` (fixed in next PR), but should now work through the `qecp` layer.

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -363,6 +405,20 @@ def match_and_rewrite(
...
+
+@dataclass(frozen=True)
+class UpdateFunctionSignatures(RewritePattern):
+    """Update types for any FuncOps in the IR whose input/output types have changed."""
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, funcop: func.FuncOp, rewriter: PatternRewriter):
+        """Match all FuncOps and call their update_function_type method"""
+
+        funcop.update_function_type()
```
lillian542, line 420: Needed for `apply_T`, maybe we should also be matching to the name? But if we add more subroutines at the `qecl` layer, this way it will also cover those. 

Could also move this into the main function instead of making it a rewrite pattern, it's not using the rewriter anyway.

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -744,6 +808,86 @@ def create_encode_subroutine(self) -> func.FuncOp:
...
+        """
+        unitary_encoding_info = self.qec_code.unitary_encoding
+
+        required_keys = {"state_prep_index", "hadamard_indices", "cnot_indices"}
+        if not required_keys.issubset(unitary_encoding_info):
+            raise CompileError(
+                f"QEC code '{self.qec_code.name}' does not define a unitary encoding "
+                f"(missing {required_keys - set(unitary_encoding_info)}); cannot lower "
+                f"qecl.fabricate [magic] for this code."
+            )
```
lillian542, line 843: Maybe we only want to generate this subroutine if there are `fabricate` ops in the code. Otherwise we can't play around with simple Clifford circuits for a code without defining the `unitary_encoding` 😅 Would go along with something like this proposed change: https://github.com/PennyLaneAI/catalyst/pull/2917

Thoughts?

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
joeycarter: Interesting, so the type-conversion patterns aren't applied to input-arg and return types in the `function_type` property of a `FuncOp` (the function's signature, in other words). It _does_ look like it's converting the block arguments in the function body (which are distinct from the input args in the `function_type` property), and then `funcop.update_function_type()` converts the signature types to match.

This is probably the best approach for now, the alternative would likely be adding a `walk` over the IR after calling `PatternRewriteWalker.rewrite_module()`, matching on `func.FuncOp`s and calling `funcop.update_function_type()` there instead. I think the two approaches amount to the same thing, and having it as a `match_and_rewrite()` pattern is probably the more idiomatic way so I think it's fine as is.

Maybe just throw in a `#pylint: disable=unused-argument` 😅

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
multiphaseCFD: In the docstr for this method, `SX` is applied. It would be nice to add some explaination why we apply `X` first instead of `S`.

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -363,6 +405,20 @@ def match_and_rewrite(
...
+# MARK: Update Function Signatures
+
+
+@dataclass(frozen=True)
+class UpdateFunctionSignatures(RewritePattern):
+    """Update types for any FuncOps in the IR whose input/output types have changed."""
+
+    @op_type_rewrite_pattern
+    def match_and_rewrite(self, funcop: func.FuncOp, rewriter: PatternRewriter):
+        """Match all FuncOps and call their update_function_type method"""
```
multiphaseCFD, line 418: Is this type update API only for the return type?

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -744,6 +808,86 @@ def create_encode_subroutine(self) -> func.FuncOp:
 
         return funcOp
 
+    # MARK: Fabricate subroutine
+
+    def create_fabricate_subroutine(self) -> func.FuncOp:
+        """Create a subroutine that allocates a codeblock and encodes it in the magic state for
+        the QEC code (based on the tanner graph), and returns the encoded codeblock. This is a
+        non-fault tolerant encoding intended for use on a simulator, and not a distillation process
```
multiphaseCFD, line 817: Thanks for adding this context.

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -744,6 +808,86 @@ def create_encode_subroutine(self) -> func.FuncOp:
...
+        for generating a magic state from many noisy copies.
+
+        The encoding process involves putting the initial QEC physical qubit in the desired state
+        via application of a Hadamard and physical T gate, and then using the unitary encoding for
+        the zero state create the desired state for the codeblock. This is not the same procedure
+        as the syndrome-measurement based procedure for encoding the zero state; encoding via the
+        syndrome-measurement procedure would force the input back into the code-space, and destroy
```
multiphaseCFD: Could we add a reference to this statement here?

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
multiphaseCFD: > Would go along with something like this proposed change: #2917 

Looks good to me

multiphaseCFD (approved):
Thanks @lillian542 . Nice work, happy to approve

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
lillian542: I believe so

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
lillian542: I'll link the paper that describes the overall procedure we are using here, and it does say "apply gate encoding", but it doesn't explicitly warn _against_ using the stabilizer-based encoding. That we can't use the stabilizer measurements in its place like we can for `encode [zero]` is something found through trial-and-error and reasoning about the system, and as I think about it more now I'm not actually 100% confident the reasoning is correct. 

Maybe I'll avoid making any claims about what is going on here, and just say I tried it and it didn't work, and doing exactly what they say in the paper did work 😅

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
joeycarter: Yeah, I've been thinking that it might make sense to do a single walk through the IR at the beginning of the pass to figure out which subroutines should be added, and only insert the ones that are actually used, rather than inserting them all and then relying on `symbol-dce` to remove the unused ones. If you want set that up in this PR, please do!

frontend/catalyst/python_interface/transforms/qecp/qec_code_lib.py
```diff
@@ -45,6 +45,27 @@
...
+                [3, 4],
+                [1, 5],
+                [1, 6],
+                [3, 0],
+            ),
+            # The state_prep_index is the index of the physical qubit that the state is
+            # injected on (i.e. for a magic state, -H-T is applied here pre-encoding).
+            # Must be consistent with the qubit treated as the encoding "input" by the
+            # cnot_indices ordering above.
+            "state_prep_index": 6,
```
joeycarter, line 69: Can you remind me where these number come from?

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
lillian542: I'm going to think about this a bit more

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
```diff
@@ -1015,6 +1029,129 @@ def test_nontransveral_ops_ignored(self, run_filecheck):
...
+            // Unitary encoding
+            //       CHECK-NEXT:   [[q0_1:%.+]] = qecp.hadamard [[q0]] : !qecp.qubit<data>
+            //       CHECK-NEXT:   [[q2_1:%.+]] = qecp.hadamard [[q2]] : !qecp.qubit<data>
+            //       CHECK-NEXT:   [[q0_2:%.+]], [[q1_out:%.+]] = qecp.cnot [[q0_1]], [[q1_2]] : !qecp.qubit<data>, !qecp.qubit<data>
+            //       CHECK-NEXT:   [[q2_out:%.+]], [[q0_out:%.+]] = qecp.cnot [[q2_1]], [[q0_2]] : !qecp.qubit<data>, !qecp.qubit<data>
+            // Insert qubits and return
+            //       CHECK-NEXT:   [[cb_1:%.+]] = qecp.insert [[cb]][0], [[q0_out]]
+            //       CHECK-NEXT:   [[cb_2:%.+]] = qecp.insert [[cb_1]][1], [[q1_out]]
+            //       CHECK-NEXT:   [[cb_3:%.+]] = qecp.insert [[cb_2]][2], [[q2_out]]
+            //       CHECK-NEXT:   func.return [[cb_3:%.+]] : !qecp.codeblock<1 x 3>
```
joeycarter: Thinking about the recent issue we had with unit lit tests being too restrictive, I might change some of these `CHECK-NEXT` directives to either `CHECK-DAG` or just `CHECK`. For example, the data flow would still be correct if we extracted qubits from the codeblock in some other permutation of {0, 1, 2}, so they could use `CHECK-DAG`. Similarly, the data flow of the state-prep ops `H -> T` would still correct if, for example, some unrelated op were inserted between them, since we check that the output qubit of the `H` gate is used as input to the `T`, so they could use `CHECK`. And so on with the others. Writing flexible lit tests is more of an art than a science, since it's hard to predicate what changes might be made in the future that are still valid; these are just suggestions, I leave it to you 🙂

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
```diff
@@ -1015,6 +1029,129 @@ def test_nontransveral_ops_ignored(self, run_filecheck):
...
+            }
+            // CHECK-LABEL: func.func private @fabricate_magic_state_Steane() -> !qecp.codeblock<1 x 7>
+            //       CHECK:   [[cb:%.+]] = qecp.alloc_cb : !qecp.codeblock<1 x 7>
+            //       CHECK-NEXT:   [[q0:%.+]] = qecp.extract [[cb]][0] : !qecp.codeblock<1 x 7> -> !qecp.qubit<data>
+            //       CHECK-NEXT:   [[q1:%.+]] = qecp.extract [[cb]][1] : !qecp.codeblock<1 x 7> -> !qecp.qubit<data>
+            //       CHECK-NEXT:   [[q2:%.+]] = qecp.extract [[cb]][2] : !qecp.codeblock<1 x 7> -> !qecp.qubit<data>
+            //       CHECK-NEXT:   [[q3:%.+]] = qecp.extract [[cb]][3] : !qecp.codeblock<1 x 7> -> !qecp.qubit<data>
+            //       CHECK-NEXT:   [[q4:%.+]] = qecp.extract [[cb]][4] : !qecp.codeblock<1 x 7> -> !qecp.qubit<data>
+            //       CHECK-NEXT:   [[q5:%.+]] = qecp.extract [[cb]][5] : !qecp.codeblock<1 x 7> -> !qecp.qubit<data>
+            //       CHECK-NEXT:   [[q6:%.+]] = qecp.extract [[cb]][6] : !qecp.codeblock<1 x 7> -> !qecp.qubit<data>
```
joeycarter: Same as above.

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
```diff
@@ -1015,6 +1029,129 @@ def test_nontransveral_ops_ignored(self, run_filecheck):
...
+            }
+            //      CHECK-LABEL: func.func private @fabricate_magic_Steane
+            // CHECK: qecp.alloc_cb
+            // CHECK: qecp.h
+            // CHECK: qecp.t [[qb:%.+]]
+            // CHECK-NOT: qecp.t [[qb:%.+]] adj
+            // CHECK: qecp.h
+            // CHECK: qecp.cnot
+        """
+
```
joeycarter: Are you missing the `run_filecheck` here?

joeycarter (commented):
Looks great, thanks @lillian542! 🪄 

Just a few comments and suggestions to start.

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecp_to_quantum.py
joeycarter: :it:

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -203,6 +213,38 @@ def match_and_rewrite(self, op: qecl.EncodeOp, rewriter: PatternRewriter):
...
+            raise NotImplementedError(
+                "Lowering qecl.FabricateOp to the qecp dialect is only implemented "
+                "for init_state 'magic'"
+            )
+
+        if (k := op.out_codeblock.type.k.value.data) != self.qec_code.k:
+            raise CompileError(
+                f"Circuit expressed in the qecl dialect with k={k} is not compatible with "
+                f"lowering to a code with k={self.qec_code.k}"
+            )
```
joeycarter: I'm not too concerned with testing coverage of these errors, so maybe throw a `# pragma: no cover` at the end of the `if` statement lines?

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -744,6 +808,86 @@ def create_encode_subroutine(self) -> func.FuncOp:
...
+        syndrome-measurement procedure would force the input back into the code-space, and destroy
+        our magic state.
+
+        Note that this method does not insert the subroutine into the module op. Instead it
+        returns the built func.FuncOp object that can then be subsequently inserted where desired.
+        """
+        unitary_encoding_info = self.qec_code.unitary_encoding
+
+        required_keys = {"state_prep_index", "hadamard_indices", "cnot_indices"}
+        if not required_keys.issubset(unitary_encoding_info):
```
joeycarter: (Same here)

```suggestion
        if not required_keys.issubset(unitary_encoding_info):  # pragma: no cover
```

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -744,6 +808,91 @@ def create_encode_subroutine(self) -> func.FuncOp:
...
 
+    # MARK: Fabricate subroutine
+
+    def create_fabricate_subroutine(self) -> func.FuncOp:
+        """Create a subroutine that allocates a codeblock and encodes it in the magic state for
+        the QEC code (based on the tanner graph), and returns the encoded codeblock. This is a
+        non-fault tolerant encoding intended for use on a simulator, and not a distillation process
+        for generating a magic state from many noisy copies.
+
+        The encoding process follows the third option for magic state encodingdescribed in
```
lillian542:
```suggestion
        The encoding process follows the third option for magic state encoding described in 
```

frontend/catalyst/python_interface/transforms/qecp/qec_code_lib.py
lillian542: They're from the main Steane code implementation paper we've been using (https://arxiv.org/pdf/2107.07505). I'll add a note, probably at the top since its true for all the insides in the whole code definition 😅

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
lillian542: I wasn't aware of `CHECK-DAG`, thanks! I'll update them to that, sounds strictly better to me.

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
lillian542: I'd prefer to wrap this one up and open a separate PR that deals with all the subroutines at once. I'll add a shortcut story.

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
lillian542: Updated with a reference for what procedure we are using, and what I believe is a more accurate explanation of why 😅

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
lillian542: I absolutely was, yes

frontend/catalyst/python_interface/transforms/qecp/qec_code_lib.py
joeycarter: Of course, thanks!

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
```diff
@@ -1015,6 +1029,131 @@ def test_nontransveral_ops_ignored(self, run_filecheck):
...
+            // Unitary encoding
+            //       CHECK-DAG:   [[q0_1:%.+]] = qecp.hadamard [[q0]] : !qecp.qubit<data>
+            //       CHECK-DAG:   [[q2_1:%.+]] = qecp.hadamard [[q2]] : !qecp.qubit<data>
+            //       CHECK-DAG:   [[q0_2:%.+]], [[q1_out:%.+]] = qecp.cnot [[q0_1]], [[q1_2]] : !qecp.qubit<data>, !qecp.qubit<data>
+            //       CHECK-DAG:   [[q2_out:%.+]], [[q0_out:%.+]] = qecp.cnot [[q2_1]], [[q0_2]] : !qecp.qubit<data>, !qecp.qubit<data>
+            // Insert qubits and return
+            //       CHECK-DAG:   [[cb_1:%.+]] = qecp.insert [[cb]][0], [[q0_out]]
+            //       CHECK-DAG:   [[cb_2:%.+]] = qecp.insert [[cb_1]][1], [[q1_out]]
+            //       CHECK-DAG:   [[cb_3:%.+]] = qecp.insert [[cb_2]][2], [[q2_out]]
+            //       CHECK-DAG:   func.return [[cb_3:%.+]] : !qecp.codeblock<1 x 3>
```
joeycarter: Ah sorry if my previous comment wasn't very clear, the `qecp.extract` ops can use `CHECK-DAG` because they can occur in any order, but I think it would make more sense for these checks to use the `CHECK` directive, since the order in which the ops occur _does_ matter (e.g. we couldn't apply the T gate on qubit `q1_1` until that SSA value has been defined from the output of the previous `H` gate op).

More info on CHECK-DAG here: https://llvm.org/docs/CommandGuide/FileCheck.html#the-check-dag-directive

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
```diff
@@ -1015,6 +1029,131 @@ def test_nontransveral_ops_ignored(self, run_filecheck):
...
+            //       CHECK:   [[cb:%.+]] = qecp.alloc_cb : !qecp.codeblock<1 x 7>
+            //       CHECK-DAG:   [[q0:%.+]] = qecp.extract [[cb]][0] : !qecp.codeblock<1 x 7> -> !qecp.qubit<data>
+            //       CHECK-DAG:   [[q1:%.+]] = qecp.extract [[cb]][1] : !qecp.codeblock<1 x 7> -> !qecp.qubit<data>
+            //       CHECK-DAG:   [[q2:%.+]] = qecp.extract [[cb]][2] : !qecp.codeblock<1 x 7> -> !qecp.qubit<data>
+            //       CHECK-DAG:   [[q3:%.+]] = qecp.extract [[cb]][3] : !qecp.codeblock<1 x 7> -> !qecp.qubit<data>
+            //       CHECK-DAG:   [[q4:%.+]] = qecp.extract [[cb]][4] : !qecp.codeblock<1 x 7> -> !qecp.qubit<data>
+            //       CHECK-DAG:   [[q5:%.+]] = qecp.extract [[cb]][5] : !qecp.codeblock<1 x 7> -> !qecp.qubit<data>
+            //       CHECK-DAG:   [[q6:%.+]] = qecp.extract [[cb]][6] : !qecp.codeblock<1 x 7> -> !qecp.qubit<data>
+            // State injection on the state_prep_index (qubit 6): H then T
+            //       CHECK-DAG:   [[h_inj:%.+]] = qecp.hadamard [[q6]] : !qecp.qubit<data>
```
joeycarter: Same here and below.

joeycarter (approved):
One last clarification from my previous review, but otherwise looks good! Happy to provisionally approve 🚀

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
lillian542: Ooooh, I misunderstood it to mean it checks what is next _on the graph_ for the SSA value, regardless of the written input. That's less helpful, then. I will put most of them back 😅

lillian542: Though now that I'm thinking about it more, that doesn't make a lot of sense 🙈

PR #2898, opened 2026-06-01
fix: use `QuantumScript.from_queue` instead of `process_queue`

andrijapau:
**Context:**

`process_queue` was recently moved to `pennylane/tape/qscript` which caused a breaking change here.

Can simply use `QuantumScript.from_queue` instead to make it simpler.

**Benefits:**

Code is more maintainable.

**Possible Drawbacks:** None identified.

[[sc-120925](https://app.shortcut.com/xanaduai/story/120925)]

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -34,4 +34,4 @@ lxml_html_clean
 --extra-index-url https://test.pypi.org/simple/
 pennylane-lightning-kokkos==0.46.0-dev10
 pennylane-lightning==0.46.0-dev10
-pennylane==0.46.0.dev18
+pennylane==0.46.0.dev20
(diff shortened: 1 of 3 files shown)


doc/requirements.txt
andrijapau, line 1: @paul0403 , if I don't *have* to bump the PL version, should I just leave it?

paul0403: ~~Sorry what do you mean? 😅 You can bump the version if you need it~~

paul0403: Oh I see, we can probably just bump it, even if just catching errors early

paul0403 (approved):

albi3ro (approved):

PR #2899, opened 2026-06-01
Add mark-entry-points pass to annotate nested-module entry functions

mehrdad2m:
**Context:**
When the compiler emits nested modules, the functions that the host actually dispatches into need to be distinguishable from purely-internal helpers. Downstream passes may use this as a signal for "this function is an externally-reachable entry point". 

Introduces a `catalyst.entry_point` annotation marking the functions that are externally callable, produced from either:
1. frontend: for the main dispatch function
2. `MarkEntryPoints` pass for nested modules.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/frontend/catalyst/utils/gen_mlir.py
+++ b/frontend/catalyst/utils/gen_mlir.py
@@ -75,4 +75,3 @@ def inject_functions(module, ctx, seed):
     teardown_module = gen_teardown(ctx)
     teardown_func = teardown_module.body.operations[0]
     module.body.append(teardown_func)
-
--- a/frontend/catalyst/utils/gen_mlir.py
+++ b/frontend/catalyst/utils/gen_mlir.py
@@ -64,6 +64,7 @@ def inject_functions(module, ctx, seed):
     """
     # Add C interface for the quantum function and annonate it as the main entry_point.
     mlir_qfunc = module.body.operations[0]
+    assert isinstance(mlir_qfunc, FuncOp)
     mlir_qfunc.attributes["catalyst.entry_point"] = ir.UnitAttr.get(context=ctx)
     mlir_qfunc.attributes["llvm.emit_c_interface"] = ir.UnitAttr.get(context=ctx)
 
@@ -75,4 +76,3 @@ def inject_functions(module, ctx, seed):
     teardown_func = teardown_module.body.operations[0]
     module.body.append(teardown_func)
 
-    assert isinstance(mlir_qfunc, FuncOp)
--- a/frontend/catalyst/utils/gen_mlir.py
+++ b/frontend/catalyst/utils/gen_mlir.py
@@ -62,8 +62,10 @@ def inject_functions(module, ctx, seed):
     """
     This function appends functions to the input module.
     """
-    # Add C interface for the quantum function.
-    module.body.operations[0].attributes["llvm.emit_c_interface"] = ir.UnitAttr.get(context=ctx)
+    # Add C interface for the quantum function and annonate it as the main entry_point.
+    mlir_qfunc = module.body.operations[0]
+    mlir_qfunc.attributes["catalyst.entry_point"] = ir.UnitAttr.get(context=ctx)
+    mlir_qfunc.attributes["llvm.emit_c_interface"] = ir.UnitAttr.get(context=ctx)
 
     setup_module = gen_setup(ctx, seed)
     setup_func = setup_module.body.operations[0]
@@ -73,5 +75,4 @@ def inject_functions(module, ctx, seed):
     teardown_func = teardown_module.body.operations[0]
     module.body.append(teardown_func)
 
-    mlir_qfunc = module.body.operations[0]
     assert isinstance(mlir_qfunc, FuncOp)
--- a/mlir/lib/Catalyst/Transforms/MarkEntryPoints.cpp
+++ b/mlir/lib/Catalyst/Transforms/MarkEntryPoints.cpp
@@ -1,0 +1,68 @@
+// Copyright 2026 Xanadu Quantum Technologies Inc.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+
+//     http://www.apache.org/licenses/LICENSE-2.0
+
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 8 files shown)


frontend/catalyst/utils/gen_mlir.py
```diff
@@ -73,5 +75,4 @@ def inject_functions(module, ctx, seed):
     teardown_func = teardown_module.body.operations[0]
     module.body.append(teardown_func)
 
-    mlir_qfunc = module.body.operations[0]
     assert isinstance(mlir_qfunc, FuncOp)
```
paul0403: Move this assert up to after `mlir_qfunc = module.body.operations[0]` as well?

dime10:
Thanks @mehrdad2m! I have a couple of questions:
- Do we want to use the same attribute for the program entry point (analogous to a `main` function), as for the quantum kernel entry points?
- For the kernels, is there a reason that this is more suitable as a pass rather than an attribute assigned during lowering?

mehrdad2m:
> * Do we want to use the same attribute for the program entry point (analogous to a `main` function), as for the quantum kernel entry points?

That's a very good question. Currently we do reuse `catalyst.entry_point` for both, but they are a slightly different. The semantic of `catalyst.entry_point` allows for multiple entry points for a module. while something like `catalyst.main` should be singular in a program. We could have both, but since I am not adding `catalyst.entry_point` to any other functions (you may argue we should add it for `setup` and `teardown` functions as well), we can reuse it as "catalyst.main". but If you prefer having that explicitly, I don't mind adding that as a separate attribute.

> * For the kernels, is there a reason that this is more suitable as a pass rather than an attribute assigned during lowering?

This was not a clear choice for me as I think both are possible. The semantics of `catalyst.entry_point`, allows for the pass to infer it assuming the kernels entry points are used in the main program. The only downside of adding it from frontend is that every downstream pass that restructures the call graph would have to move the attribute correctly (e.g. if a qnode is cloned and called from a workflow function, you have to remove the attribute from the qnode and attach it to the workflow function) which is a maintenance burden.

mlir/lib/Catalyst/Transforms/MarkEntryPoints.cpp
```diff
@@ -0,0 +1,68 @@
...
+        // stamp matching functions inside each nested module. The host's own
+        // entry is annotated by the frontend, not this pass.
+        llvm::SmallSet<StringRef, 8> exposed;
+        for (Operation &topOp : root.getBody()->getOperations()) {
+            if (isa<ModuleOp>(&topOp)) {
+                continue;
+            }
+            if (auto uses = SymbolTable::getSymbolUses(&topOp)) {
+                for (SymbolTable::SymbolUse use : *uses) {
+                    exposed.insert(use.getSymbolRef().getLeafReference().getValue());
```
paul0403, line 48: What is `getLeafReference()`?

mehrdad2m: this deals with nested symbolic references. e.g. running it on `@workflow_module::@circuit` would result in `@circuit`. The reason is that the Host call references the nested symbol but when tryign to match against the function inside the module there is no reference of the parent module in the symbol

frontend/catalyst/utils/gen_mlir.py
```diff
@@ -62,8 +62,10 @@ def inject_functions(module, ctx, seed):
     """
     This function appends functions to the input module.
     """
-    # Add C interface for the quantum function.
-    module.body.operations[0].attributes["llvm.emit_c_interface"] = ir.UnitAttr.get(context=ctx)
+    # Add C interface for the quantum function and annonate it as the main entry_point.
+    mlir_qfunc = module.body.operations[0]
```
mehrdad2m, line 66:
```suggestion
    mlir_qfunc = module.body.operations[0]
    assert isinstance(mlir_qfunc, FuncOp)
```

frontend/catalyst/utils/gen_mlir.py
mehrdad2m:
```suggestion
```

paul0403 (approved):

dime10:
> That's a very good question. Currently we do reuse catalyst.entry_point for both, but they are a slightly different. The semantic of catalyst.entry_point allows for multiple entry points for a module. while something like catalyst.main should be singular in a program. We could have both, but since I am not adding catalyst.entry_point to any other functions (you may argue we should add it for setup and teardown functions as well), we can reuse it as "catalyst.main". but If you prefer having that explicitly, I don't mind adding that as a separate attribute.

Hmm, putting the question of having the same attribute for program and kernel entry points aside, why do we actually need multiple entry points for a kernel? I've never seen us make use of it, and even conceptually I would think the kernel should have one entry point. 
Although things have a changed a little now that we separated out QNode functions from the kernel concept, because it's actually the QNode functions that need some unique processing at the execution boundary (device setup and measurement processes for instance). I feel like our abstractions aren't as clean as they could be atm 🤔 

Coming back to the original question, if we make it so the kernel has exactly one entry point, then I think reusing the attribute is fine. If they do in fact have different semantics separating them out might make sense instead.

dime10:
> The semantics of catalyst.entry_point, allows for the pass to infer it assuming the kernels entry points are used in the main program. The only downside of adding it from frontend is that every downstream pass that restructures the call graph would have to move the attribute correctly (e.g. if a qnode is cloned and called from a workflow function, you have to remove the attribute from the qnode and attach it to the workflow function) which is a maintenance burden.

This is true, there is a bit of maintenance burden, although this applies to all invariants of our IR that passes need to respect. They also need to make sure to propagate, add, and remove the quantum.node attribute as needed as well. I actually had a [WIP branch](https://github.com/PennyLaneAI/catalyst/compare/main...entry-point-attribute) to formalize the entry point attribute similar to qnode attribute, and the idea was to use the verifier to ensure that passes are not violating our invariants.

Relying on this pass alone for example doesn't fix the problem, unless you always run it right before you need it (otherwise passes can still invalidate the attribute invariant after the pass was run). Even then you'd have to make sure the pass deletes any existing ones too, not just adding them for the current launch_kernel ops (as the pass currently does).

Overall I would lean towards an "invariants" approach rather than an "analysis" approach, but I'm curious to hear what you think.

mehrdad2m:
Closing as it is not needed in this form.

PR #2902, opened 2026-06-02
Make a header for QubitIndex

rniczh:
**Context:**

QubitIndex was originally defined in the decomposition lowering pass (`DecomposeLoweringImpl.hpp`), but it is a general-purpose utility for representing a qubit index that may be either a dynamic SSA Value or a compile-time IntegerAttr, together with the qreg it belongs to. Other developers may need to reuse it for tracing the index.

Extract QubitIndex into its own header at `mlir/include/Quantum/Utils/QubitIndex.h`, in the `catalyst::quantum` namespace.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/mlir/include/Quantum/Utils/QubitIndex.h
+++ b/mlir/include/Quantum/Utils/QubitIndex.h
@@ -1,0 +1,81 @@
+// Copyright 2026 Xanadu Quantum Technologies Inc.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


kipawaa (approved):
Good choice moving this out!

sengthai (approved):
Thanks @rniczh!

paul0403:
Nice! Can I screenshot this PR in my presentation about qref to show that the walkback is really annoying? :laughing:

rniczh:
@paul0403 sure, no problem

PR #2904, opened 2026-06-02
Handle dlopen if rtcapi is built but used on different os

rniczh:
**Context:**

When the rt-capi was built on one OS and is being run on another.

Fall back to the platform-native extension (`.so/.dylib/.dll`) when dlopen fails.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-119223]


runtime/lib/capi/ExecutionContext.hpp
```diff
@@ -81,15 +83,37 @@ class SharedLibraryManager final {
...
+            return;
+        }
+
+        // dlerror() is destructive: it returns the pending error and clears the state. Capture it
+        // into a local before any further use.
+        const char *primary_dlerror = dlerror();
+        std::string primary_error = primary_dlerror ? primary_dlerror : "unknown dlopen error";
+        auto basename = std::filesystem::path(filename).filename().string();
+        
+        // rewrite the extension if it differs from the platform one.
```
dime10: Should we just do this first, rather than waiting on an error? 🤔

dime10 (approved):
Thanks!

runtime/lib/capi/ExecutionContext.hpp
rniczh: Yes for the extension rewrite, we can normalize `.so/.dylib` up front since a wrong native extension won't load anyway. I updated this in this commit:
https://github.com/PennyLaneAI/catalyst/pull/2904/commits/233bda689b3919a3a9af4a32a74ff001f3606f9b

mehrdad2m (approved):

runtime/lib/capi/ExecutionContext.hpp
dime10: Ah I'm just realizing your fallback was doing two things at once, fixing the extension _and_ attempt loading via the libname only (no path). Is the latter a sensible approach? If so feel free to resolve after responding.

rniczh: This is because the current path for rt_capi is fixed and inherited from the attribute. This fallback mainly handles the case when the library is not provided in the given path; then it tries again using only the filename to search directly from LD_LIBRARY_PATH.

I think the simplest approach would be to just take the rt_capi lib directly from LD_LIBRARY_PATH and never pass this path via the attribute. However, this change is relatively large and would have a wider impact.

PR #2905, opened 2026-06-02
Add support for `scf.while` ops in the `convert-quantum-to-qecl` pass

joeycarter:
**Context:** In order to support more general input programs in the experimental QEC compilation pipeline, we must add support for control-flow operations.

**Description of the Change:** This PR adds support for `while` loops (via [`scf.while`](https://mlir.llvm.org/docs/Dialects/SCFDialect/#scfwhile-scfwhileop) ops) in the `convert-quantum-to-qecl` pass. For example, an input program such as

```mlir
func.func @test_program(%arg0: f64) {
    %c2 = arith.constant 2.000000e+00 : f64
    %c9 = arith.constant 9.000000e+00 : f64
    %1 = quantum.extract %0[0] : !quantum.reg -> !quantum.bit
    %2, %3 = scf.while (%arg1 = %arg0, %arg2 = %1) : (f64, !quantum.bit) -> (f64, !quantum.bit) {
        %4 = arith.cmpf olt, %arg1, %c9 : f64
        scf.condition(%4) %arg1, %arg2 : f64, !quantum.bit
    } do {
    ^bb0(%arg1_1: f64, %arg2_1: !quantum.bit):
        %5 = quantum.custom "PauliX"() %1 : !quantum.bit
        %6 = arith.addf %arg1_1, %c2 : f64
        scf.yield %6, %5 : f64, !quantum.bit
    }
    %7 = quantum.insert %0[0], %3 : !quantum.reg, !quantum.bit
    quantum.dealloc %7 : !quantum.reg
    func.return
}
```

is lowered to

```mlir
func.func @test_program(%arg0: f64) {
    %c2 = arith.constant 2.000000e+00 : f64
    %c9 = arith.constant 9.000000e+00 : f64
    %1 = arith.constant 0 : index
    %2 = arith.constant 1 : index
    %3 = arith.constant 1 : index
    %4 = scf.for %5 = %1 to %2 step %3 iter_args(%6 = %0) -> (!qecl.hyperreg<1 x 1>) {
        %7 = qecl.extract_block %6[%5] : !qecl.hyperreg<1 x 1> -> !qecl.codeblock<1>
        %8 = qecl.encode[zero] %7 : !qecl.codeblock<1>
        %9 = qecl.insert_block %6[%5], %8 : !qecl.hyperreg<1 x 1>, !qecl.codeblock<1>
        scf.yield %9 : !qecl.hyperreg<1 x 1>
    }
    %10 = qecl.extract_block %4[0] : !qecl.hyperreg<1 x 1> -> !qecl.codeblock<1>
    %11 = qecl.qec %10 : !qecl.codeblock<1>
    %12, %13 = scf.while (%arg1 = %arg0, %arg2 = %11) : (f64, !qecl.codeblock<1>) -> (f64, !qecl.codeblock<1>) {
        %14 = arith.cmpf olt, %arg1, %c9 : f64
        scf.condition(%14) %arg1, %arg2 : f64, !qecl.codeblock<1>
    } do {
    ^bb0(%arg1_1: f64, %arg2_1: !qecl.codeblock<1>):
        %15 = qecl.x %11[0] : !qecl.codeblock<1>
        %16 = qecl.qec %15 : !qecl.codeblock<1>
        %17 = arith.addf %arg1_1, %c2 : f64
        scf.yield %17, %16 : f64, !qecl.codeblock<1>
    }
    %18 = qecl.insert_block %4[0], %13 : !qecl.hyperreg<1 x 1>, !qecl.codeblock<1>
    qecl.dealloc %18 : !qecl.hyperreg<1 x 1>
    func.return
}
```

Support for `scf.while` ops that return `!quantum.reg` values, and a mix of quantum and classical types, has also been added.

This PR also refactors the other `scf` conversion patterns (for `scf.if` and `scf.for`) to improve code abstractions and reduce code duplication.

[sc-120267]


multiphaseCFD:
Would you please remove the `scf.for` operation in the target IR listed in the PR description? There is no `scf.for` in the original IR

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
```diff
@@ -513,16 +514,22 @@ def match_and_rewrite(self, op: quantum.MeasureOp, rewriter: PatternRewriter):
...
 
-    @op_type_rewrite_pattern
-    def match_and_rewrite(self, op: scf.YieldOp, rewriter: PatternRewriter):
-        """Rewrite pattern for `scf.yield` ops."""
-        new_yield_operands = []
+    def match_and_rewrite(self, op: Operation, rewriter: PatternRewriter):
+        """Rewrite pattern for terminator ops.
+
+        Each subclass of `TerminatorConversion` must implement its own `match_and_rewrite()` with
+        the `@op_type_rewrite_pattern` decorator and call `super().match_and_rewrite(...)`.
```
multiphaseCFD, line 531: Nice! In terms of this pattern for `func.return`, it would update the both the SSA value to be returned and the return type, while the update_func_... signature API only works for return type update?

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
```diff
@@ -571,70 +580,114 @@ def _insert_conv_cast_op(
...
-        assert parent_op is not None, "Op 'scf.yield': expected parent op, but found none"
+        assert parent_op is not None, f"Op '{op.name}': expected parent op, but found none"
         rewriter.notify_op_modified(parent_op)
 
 
-# MARK: SCF Base Pattern
+# MARK: SCF Yield Pattern
+
+
+class ScfYieldConversion(TerminatorConversion):
```
multiphaseCFD: Could we put `ScfYieldConversion` and `ScfConditionConversion` together?

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
```diff
@@ -571,70 +580,114 @@ def _insert_conv_cast_op(
...
         the `@op_type_rewrite_pattern` decorator and call `super().match_and_rewrite(...)`.
         """
-        yield_op = self._get_yield_op(op)
+        self._convert_op_results(op, rewriter)
 
-        for i_res, result in enumerate(op.results):
-            # The yield operand that corresponds to the current scf op's return value
-            yield_operand = yield_op.operands[i_res]
+    @classmethod
+    def _convert_op_results(cls, op: OpType, rewriter: PatternRewriter):
```
multiphaseCFD, line 627: Any design consideration on using classmethod to decorate methods instead of move them to an utils module as plain free functions? I feel like we need similar helper funcs for the lowering from qecl to qecp as well?

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
```diff
@@ -658,140 +711,147 @@ def _convert_quantum_type_result(
...
-        yield_op = yield_block.last_op
-        assert isinstance(yield_op, scf.YieldOp), (
-            f"Op 'scf.for': expected last op in `body` to be 'scf.yield', but got "
-            f"{type(yield_op).__name__}"
+        return first_block
+
+    @classmethod
+    def _get_terminator_operands_passed_to_results(cls, op: scf.WhileOp) -> Sequence[SSAValue]:
+        """Get the operands of the `scf.condition` op from the `before` region of an `scf.while` op.
+
```
multiphaseCFD, line 826: Thanks for adding notes here!

multiphaseCFD (commented):
LGTM! Thanks @joeycarter !

joeycarter:
> Would you please remove the `scf.for` operation in the target IR listed in the PR description? There is no `scf.for` in the original IR

That's the `for` loop over each codeblock in the hyper-register that performs encoding, which is lowered from `quantum.alloc` :slightly_smiling_face:

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
joeycarter: Yep that's the idea. This rewrite pattern would only modify the `func.return` op within the body of the `func.func` op. We would need to add another conversion pattern to convert the signature of the `func.func` op itself (e.g. based on the updated types of the `func.return` operands and the types of the values passed to `func.call`. We currently can't do the latter in xDSL because they haven't yet added support for finding the uses of a symbol, but that's a separate issue :slightly_smiling_face:)

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
joeycarter: I don't think we need similar helper functions elsewhere, since it's only in the `quantum-to-qecl` pass that we need to jump through these hoops of adding manual type conversions. Unless you can think of a specific example—otherwise all these type conversions are largely handled automatically by the `qecl-to-qecp` type converters. My feeling is that these helper methods are very much specific to the conversion of `scf` ops in this pass only, so probably makes most sense to leave them as class methods.

frontend/catalyst/python_interface/transforms/qecl/convert_quantum_to_qecl.py
joeycarter: Indeed we can! 0f34987aa

multiphaseCFD (approved):

mehrdad2m (approved):

PR #2909, opened 2026-06-03
[Remote 1.1] Add executor dialect

rniczh:
**Context:**

The executor dialect gives Catalyst a way to express the communication between executor (via ORC JIT) and the current program. The goal of the executor PRs is we want to be able to carve out selected kernels (e.g. a qnode), hand them to a separate executor process, and stitch the results back into the host program, and all represented explicitly in the IR.

Adds a new executor dialect (`::catalyst::executor`) that models host <-> executor interaction as four ops:

1. `executor.open(address)`: establish a session with the executor at address.
2. `exectuor.send_binary(address, binary_path)`:  ship a cross-compiled kernel object file (with a format selector) and expose its symbols or any arbitrary binary files.
3. `exectuor.launch(kernel_callee, address)(inputs) -> results`:  call a previously-shipped qnode kernel @ address
4. `exectuor.call(symbol, address)(inputs) -> results`: Invoke an arbitrary symbol in an already-loaded shared library, with an optional num_input_args boundary separating input buffers from output buffers
5. `exectuor.close(address)`: Tear down the session with an executor.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-119223]


dime10:
Please provide a PR description :)

mlir/include/Remote/IR/RemoteOps.td
```diff
@@ -0,0 +1,119 @@
...
+
+#ifndef REMOTE_OPS
+#define REMOTE_OPS
+
+include "mlir/IR/OpBase.td"
+include "mlir/IR/BuiltinAttributes.td"
+include "Remote/IR/RemoteDialect.td"
+
+//===----------------------------------------------------------------------===//
+// remote.open
```
maliasadi: What do you think of adding a `remote.close` to the dialect? THis is especially useful to express  the session teardown in the IR.

mlir/include/Executor/IR/ExecutorOps.td
maliasadi, line 1: My concern with this design is that we cannot represent the session lifecycle in the IR. `open` produces no SSA value, same as others so the open -> send -> launch ordering is purely positional and needs to be heavily checked at runtime. 

What can be done here is for `remote.open` to return a handle, for example `!remote.session` so that the other ops consume.

mlir/include/Remote/IR/RemoteOps.td
```diff
@@ -0,0 +1,119 @@
...
+def Remote_LaunchOp : Remote_Op<"launch"> {
+    let summary = "Dispatch a remote-compiled kernel and receive its results.";
+    let description = [{
+        Invoke the kernel previously registered via `remote.send_binary`.
+        The `inputs` operand list mirrors the original host-side `func.call`
+        to the kernel; `results` mirrors the kernel's return values.
+    }];
+
+    let arguments = (ins
+        Variadic<AnyType>:$inputs,
```
maliasadi: Shouldn't be memrefs for compatibility with the LLVM lowering (https://github.com/PennyLaneAI/catalyst/pull/2910) and the runtime?

maliasadi: maybe https://gitlab.stud.idi.ntnu.no/filipko/sedei/-/blame/4ba6a9c9f65bbc8bd06e3652cb20fd4dfc846137/mlir/test/mlir-tblgen/op-side-effects.td#L18

```suggestion
        Variadic<AnyMemRef>:$inputs,
```

mlir/test/Executor/ExecutorOps.mlir
```diff
@@ -0,0 +1,53 @@
...
+
+// CHECK-LABEL: func.func @call
+// CHECK:         remote.call("foo", "127.0.0.1:9000")
+// CHECK-SAME:    num_input_args = 1 : i32
+// CHECK-SAME:    (memref<4xf64>, memref<4xf64>) -> ()
+func.func @call(%arg0: memref<4xf64>, %arg1: memref<4xf64>) {
+  remote.call("foo", "127.0.0.1:9000") (%arg0, %arg1)
+      {num_input_args = 1 : i32} : (memref<4xf64>, memref<4xf64>) -> ()
+  return
+}
```
maliasadi, line 56: `remote`: "remote from whom?" revisiting the project docs/reports, this dialect defines an execute model and a strategy to launch kernels on either local or remote devices.  I'm a bit in favour of naming this dialect `executor` instead of `remote`; it's also the term that LLVM ORC uses for it (right?). This will be more aligned with the runtime implementation in PR https://github.com/PennyLaneAI/catalyst/pull/2915.

mlir/include/Remote/IR/RemoteOps.td
rniczh: Thanks! Yeah, that's reasonable. commit: https://github.com/PennyLaneAI/catalyst/pull/2909/commits/828eaaa9de9444afcba04b5d0f136dbeebcbd145

mlir/include/Remote/IR/RemoteOps.td
rniczh: https://github.com/PennyLaneAI/catalyst/pull/2909/commits/c5712f37f3b1375916dfb8374232b112e38765b7

mlir/include/Executor/IR/ExecutorOps.td
rniczh: Happy to add `remote.close` to the dialect. The problematic part is today `remote.open` is injected into setup(), and in Catalyst setup/teardown are per-JIT-call: host calls setup() -> qfunc -> teardown() on every invocation of the compiled function (which were called in gen_mlir.py, they are three independent entry points). In other words, the session's lifetime is process scoped, but setup/teardown are per-call. So this is why current address keyed, get-or-create session map was designed.

mlir/test/Executor/ExecutorOps.mlir
rniczh: Sounds good, I updated it in this commit, I'll also update it in other PRs. https://github.com/PennyLaneAI/catalyst/pull/2909/commits/19c861f4aad0f75fdd865da2bde95afdcd1dfdda

mlir/include/Executor/IR/ExecutorOps.td
maliasadi: What's the future plan for executor open/close lifetime? I see the benefits to tie the executor scope to per-JIT-call but this needs to eventually be decoupled, in particular when there'll be multiple device/kernel open/launch/close in one single JIT call. what do you think?

rniczh: I think the main issue would be more like you have multiple JIT calls (call a qnode twice)., then it will need to open and close the session multiple times.

maliasadi: But the lifetime of each open/close will be per kernel than per-jit

rniczh: Yeah, I agree, chaining a session handle can make the identity explicit in the IR (such as the following IR), which is nice. The catch is lifetime, if `executor.close` were a real disconnect, the calling this qfunc repeatedly (like calling it with different params) would tear the session down and force us to reconnect and re-send the same binary on every call, that's what we don't want, instead of reusing the original session.  

So I'd split the two concerns:

1. Identity: keep the SSA handles chained as in the IR below, so each kernel/executor is explicit and each launch consumes the right session.
2. Lifetime: keep it deferred. `executor.open` stays `get-or-create`, send_binary stays cached (skips if that object was already shipped), and `executor.close` is a release marker, not a hard disconnect. The runtime keeps the session alive and only really tears it down at process exit (atexit).

```mlir
func.func @qfunc(...) {
  %sX = executor.open("X")
  executor.launch %sX (kernelA ...)
  executor.close %sX
  %sY = executor.open("Y")
  executor.launch %sY (kernelC ...)
  executor.close %sY
}
```

maliasadi: Happy to go with this Identity-Lifetime split!

maliasadi: just checking: releasing the marker w/ `executor.close` w/o a hard disconnect shouldn't add unwanted overhead to multiple kernels launch scenarios. right?

maliasadi (commented):
Thanks @rniczh! Awesome! Happy to approve after resolving the remaining comment! Could you add a changelog entry?

mlir/include/Executor/IR/ExecutorOps.td
rniczh: Yes

mehrdad2m (approved):

mlir/include/Executor/IR/ExecutorOps.td
rniczh: Done https://github.com/PennyLaneAI/catalyst/pull/2909/commits/bf3b98fde65f48c730021e8d20455100a2e7bee5

maliasadi: Awesome!

mlir/include/Executor/IR/ExecutorDialect.td
```diff
@@ -0,0 +1,50 @@
...
+//===----------------------------------------------------------------------===//
+
+def Executor_Dialect : Dialect {
+    let summary = "Operations for dispatching kernels and calling symbols on an executor.";
+    let description = [{
+        The `executor` dialect models a small set of operations that describe how a host program
+        interacts with an executor service:
+
+        * Opening a session with an executor endpoint
+        * Shipping a cross-compiled kernel object file or assets to that endpoint
```
maliasadi:
```suggestion
        * Shipping a cross-compiled kernel object file (or other assets) to that endpoint
```

mlir/include/Executor/IR/ExecutorOps.td
```diff
@@ -0,0 +1,141 @@
...
+//===----------------------------------------------------------------------===//
+// executor.send_binary
+//===----------------------------------------------------------------------===//
+
+def Executor_SendBinaryOp : Executor_Op<"send_binary"> {
+    let summary = "Ship a compiled kernel object file to the executor.";
+    let description = [{
+        Transfer the compiled binary at `binary_path` over the `session` to the executor.
+        The executor loads the file and exposes all of its symbols.
+        `format` selects the wire encoding which is handled by the runtime.
```
maliasadi:
```suggestion
        `format` selects the wire encoding for the transfer, which is handled by the
        runtime. Defaults to `0` (the raw object file); other values are reserved for
        future encodings.
```

maliasadi:
Don't forget to update the changelog before merging the PR :slightly_smiling_face:

maliasadi (approved):
🚀

PR #2910, opened 2026-06-03
[Remote 1.2] Add executor to llvm transform

rniczh:
**Context:**

Converting the executor dialect to LLVM IR. Handling the memref operands/result and string endpoints. It rewrites each `executor.*` op into a concrete call into the `__catalyst__executor__*` C-ABI runtime (librt_executor), materializing the string globals, descriptor buffers, and per-argument metadata the runtime expects.

The conversion for each executor op:

1. `executor.open`: declares and calls `__catalyst__executor__open(!llvm.ptr) -> i64`
2. `executor.send_binary`: `__catalyst__executor__send_binary(ptr addr, ptr path, i32 format)`. The path global is keyed by the object-file name; the address global is keyed by the endpoint address, so every op targeting the same address shares one global.
3. `executor.launch`: `__catalyst__executor__launch(addr, sym, num_in, in_descs, in_ranks, in_sizes, num_out, out_descs, out_ranks, out_sizes)`.
	- Input memref descriptors are stored to stack slots and gathered into an `!llvm.array<N x ptr>`
	- output descriptors get fresh stack slots that the runtime fills
	- Per-arg ranks and element sizes are emitted as constant i64 array globals. The callee is mangled to the `_catalyst_pyface_<name>` symbol.
4. `executor.call`: packs input memrefs into a single contiguous byte buffer and calls `__catalyst__executor__call_wrapper` where we need to copy the returned buffer back into the output memrefs, then calls `__catalyst__executor__free_result` to release it.
5. `executor.close`: declares and calls `__catalyst__executor__close(ptr addr) -> i64`, tearing down the session for that address.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-119223]


mlir/test/Executor/ConvertExecutorToLLVM.mlir
maliasadi, line 1: Nice tests! As the lowering introduces lots of alloca/pointer logic, could you also test/check the logic  to ensure that the pass doesn't introduce any unwanted/unexpected instructions?

mlir/lib/Executor/Transforms/ExecutorToLLVM.cpp
```diff
@@ -0,0 +1,478 @@
...
+        // - num_outputs: the number of output arguments
+        // - output_descs: the output descriptor array
+        // - output_ranks: the output rank array
+        Type launchSig = LLVM::LLVMFunctionType::get(
+            voidTy, {ptrTy, ptrTy, i64Ty, ptrTy, ptrTy, ptrTy, i64Ty, ptrTy, ptrTy, ptrTy});
+        LLVM::LLVMFuncOp launchFn = catalyst::ensureFunctionDeclaration<LLVM::LLVMFuncOp>(
+            rewriter, op, "__catalyst__remote__launch", launchSig);
+
+        std::string callee = op.getKernelCallee().str();
+        Value addrPtr = getGlobalString(loc, rewriter, "remote_addr_" + callee,
```
maliasadi: Any risks of name collisions with `send_binary` keys? 🤔

mlir/lib/Executor/Transforms/ExecutorToLLVM.cpp
```diff
@@ -0,0 +1,478 @@
...
+        Type ptrTy = LLVM::LLVMPointerType::get(ctx);
+        Type i64Ty = rewriter.getI64Type();
+        ModuleOp mod = op->getParentOfType<ModuleOp>();
+
+        Type openSig = LLVM::LLVMFunctionType::get(i64Ty, {ptrTy});
+        LLVM::LLVMFuncOp openFn = catalyst::ensureFunctionDeclaration<LLVM::LLVMFuncOp>(
+            rewriter, op, "__catalyst__remote__open", openSig);
+
+        Value addrPtr =
+            getGlobalString(loc, rewriter, "remote_setup_addr", op.getAddress().str() + '\0', mod);
```
maliasadi: Shouldn't this be parameterized by the address instead of always creating it with the `remote_setup_addr` key?

mlir/lib/Executor/Transforms/ExecutorToLLVM.cpp
```diff
@@ -0,0 +1,478 @@
...
+
+// Calculate the byte size of one memref element.
+// For complex<T>, the size is 2 * sizeof(T). 1 for real, 1 for imaginary.
+int64_t memrefElemSizeBytes(MemRefType ty)
+{
+    Type elem = ty.getElementType();
+    if (auto cplx = dyn_cast<ComplexType>(elem)) {
+        return 2 * ((cplx.getElementType().getIntOrFloatBitWidth() + 7) / 8);
+    }
+    return (elem.getIntOrFloatBitWidth() + 7) / 8;
```
maliasadi: This removes the `getIntOrFloatBitWidth()` call entirely. `primitiveByteSize` already handles integer, float, index (8 bytes), and complex (2 * inner) and returns -1 for anything else.

```suggestion
    return primitiveByteSize(ty.getElementType());
```

mlir/lib/Executor/Transforms/ExecutorToLLVM.cpp
```diff
@@ -0,0 +1,478 @@
...
+    Value arr = LLVM::UndefOp::create(rewriter, loc, arrTy);
+    for (auto [i, p] : llvm::enumerate(ptrs)) {
+        arr = LLVM::InsertValueOp::create(rewriter, loc, arr, p, SmallVector<int64_t>{(int64_t)i});
+    }
+    LLVM::StoreOp::create(rewriter, loc, arr, alloca);
+    return alloca;
+}
+
+// Calculate the byte size of one memref element.
+// For complex<T>, the size is 2 * sizeof(T). 1 for real, 1 for imaginary.
```
maliasadi: Consider this patch over the current impl to avoid `IndexType` special logic:
```suggestion
// Calculate the byte size of one memref element.
// Byte size of one memref element, or -1 on unsupported element types.
// Delegates to primitiveByteSize so the launch and call paths agree
// (in particular, index is treated as 8 bytes, not asserted on).
```

mlir/lib/Executor/Transforms/ExecutorToLLVM.cpp
```diff
@@ -0,0 +1,478 @@
...
+        SmallVector<Value> inputDescPtrs;
+        SmallVector<int64_t> inputRanks, inputElemSizes;
+        for (auto [origInput, llvmInput] : llvm::zip(op.getInputs(), adaptor.getInputs())) {
+            auto memrefTy = cast<MemRefType>(origInput.getType());
+            inputRanks.push_back(memrefTy.getRank());
+            inputElemSizes.push_back(memrefElemSizeBytes(memrefTy));
+            Value alloca = getStaticAlloca(loc, rewriter, llvmInput.getType(), 1);
+            LLVM::StoreOp::create(rewriter, loc, llvmInput, alloca);
+            inputDescPtrs.push_back(alloca);
+        }
```
maliasadi, line 279:
```suggestion
        for (auto [origInput, llvmInput] : llvm::zip(op.getInputs(), adaptor.getInputs())) {
            auto memrefTy = cast<MemRefType>(origInput.getType());
            int64_t elemSz = memrefElemSizeBytes(memrefTy);
            if (elemSz < 0) {
                return op->emitOpError("unsupported memref element type for remote.launch: ")
                    << memrefTy.getElementType();
            }
            inputRanks.push_back(memrefTy.getRank());
            inputElemSizes.push_back(elemSz);
            Value alloca = getStaticAlloca(loc, rewriter, llvmInput.getType(), 1);
            LLVM::StoreOp::create(rewriter, loc, llvmInput, alloca);
            inputDescPtrs.push_back(alloca);
        }
```

mlir/lib/Executor/Transforms/ExecutorToLLVM.cpp
```diff
@@ -0,0 +1,478 @@
...
+        SmallVector<Value> outputDescPtrs;
+        SmallVector<int64_t> outputRanks, outputElemSizes;
+        for (Type resultTy : op.getResultTypes()) {
+            auto memrefTy = cast<MemRefType>(resultTy);
+            outputRanks.push_back(memrefTy.getRank());
+            outputElemSizes.push_back(memrefElemSizeBytes(memrefTy));
+            Type llvmDescTy = getTypeConverter()->convertType(resultTy);
+            Value alloca = getStaticAlloca(loc, rewriter, llvmDescTy, 1);
+            outputDescPtrs.push_back(alloca);
+        }
```
maliasadi, line 295:
```suggestion
        for (Type resultTy : op.getResultTypes()) {
            auto memrefTy = cast<MemRefType>(resultTy);
            int64_t elemSz = memrefElemSizeBytes(memrefTy);
            if (elemSz < 0) {
                return op->emitOpError("unsupported memref element type for remote.launch: ")
                    << memrefTy.getElementType();
            }
            outputRanks.push_back(memrefTy.getRank());
            outputElemSizes.push_back(elemSz);
            Type llvmDescTy = getTypeConverter()->convertType(resultTy);
            Value alloca = getStaticAlloca(loc, rewriter, llvmDescTy, 1);
            outputDescPtrs.push_back(alloca);
        }
```

rniczh: Thanks! I updated them in the same commit: https://github.com/PennyLaneAI/catalyst/pull/2910/commits/4b3b2b34da4180f97ecebc7c47afa0b3eb06fc9f

mlir/lib/Executor/Transforms/ExecutorToLLVM.cpp
rniczh: Thanks! I updated them in the same commit: https://github.com/PennyLaneAI/catalyst/pull/2910/commits/4b3b2b34da4180f97ecebc7c47afa0b3eb06fc9f

mlir/lib/Executor/Transforms/ExecutorToLLVM.cpp
rniczh: https://github.com/PennyLaneAI/catalyst/pull/2910/commits/9bfbfe24e4c632454318f6122ae1f1ab5ceffff3

mlir/lib/Executor/Transforms/ExecutorToLLVM.cpp
rniczh: Good catch! I added a function `addrGlobalKey` to get the global key https://github.com/PennyLaneAI/catalyst/pull/2910/commits/9bfbfe24e4c632454318f6122ae1f1ab5ceffff3

mlir/lib/Executor/Transforms/ExecutorToLLVM.cpp
rniczh: https://github.com/PennyLaneAI/catalyst/pull/2910#discussion_r3581582886
So With this fix ,now each address has its own unique key.

mlir/test/Executor/ConvertExecutorToLLVM.mlir
rniczh: Sure, no problem: https://github.com/PennyLaneAI/catalyst/pull/2910/commits/ca420c1e951ae0650f173f476a0a3366844be507

maliasadi (approved):
Thanks @rniczh

mehrdad2m (approved):

PR #2911, opened 2026-06-03
Quick fix for generic format MLIR sent into --convert-to-value-semantics

paul0403:
**Context:**
JAX serializes MLIR modules to textual MLIR in generic format. For functions with no return values, such as the decomposition rules in reference semantics (since they don't return quantum values, like those in value semantics), JAX adds the `res_attrs=[]` empty result attribute to the funcop.

When such a function is sent into `--convert-to-value-semantics`, and these decomp rule funcs are converted to value semantics, the new funcs now have return values, but the result attribute is still empty, leading to a mismatch and a crash. 

Use func op interface's methods to handle the addition and erasure of args and results. This ensures all relavant internals (like the aforementioned attributes) are appropriately handled by API contract.

**Benefits:**
Fixes XAS benchmark.

### DIFF
--- a/frontend/catalyst/python_interface/compiler.py
+++ b/frontend/catalyst/python_interface/compiler.py
@@ -80,13 +80,14 @@ class Compiler:
 
         # JAX serializes void func.func ops with `res_attrs = []` in generic form
         # triggering an assertion in FuncToLLVM lowering.
-        # Assert that the canonicalization above has removed these empty attributes
+        # Remove empty arrays in-place so the generic printer omits them.\
         for op in xmod.walk():
             if not isinstance(op, FuncOp):
                 continue
             for key in ("res_attrs", "arg_attrs"):
                 val = op.properties.get(key)
-                assert not (isinstance(val, ArrayAttr) and len(val) == 0)
+                if isinstance(val, ArrayAttr) and len(val) == 0:
+                    del op.properties[key]
         buffer = io.StringIO()
         Printer(stream=buffer, print_generic_format=True).print_op(xmod)
 
--- a/frontend/catalyst/python_interface/compiler.py
+++ b/frontend/catalyst/python_interface/compiler.py
@@ -80,7 +80,7 @@ class Compiler:
 
         # JAX serializes void func.func ops with `res_attrs = []` in generic form
         # triggering an assertion in FuncToLLVM lowering.
-        # Remove empty arrays in-place so the generic printer omits them.\
+        # Remove empty arrays in-place so the generic printer omits them.
         for op in xmod.walk():
             if not isinstance(op, FuncOp):
                 continue
--- a/frontend/catalyst/python_interface/compiler.py
+++ b/frontend/catalyst/python_interface/compiler.py
@@ -65,6 +65,7 @@ class Compiler:
         value_semantics_mlir = (
             catalyst.python_interface.inspection.xdsl_conversion._quantum_opt_stderr(
                 '--catalyst-pipeline="pipe(canonicalize;convert-to-value-semantics;canonicalize)"',
+                "--mlir-print-op-generic",
                 stdin=str(gentxtmod),
             )
         )
--- a/mlir/lib/QRef/Transforms/value_semantics_conversion.cpp
+++ b/mlir/lib/QRef/Transforms/value_semantics_conversion.cpp
@@ -1755,47 +1755,71 @@ void handleSubroutine(IRRewriter &builder, func::FuncOp f,
 
     // Add new quantum arguments
     QubitValueTracker regionTracker;
-    for (auto rValue : rValuesUsedBySubroutine) {
-        Value newArg;
+    size_t originalNumArgs = f.getFunctionType().getNumInputs();
+    SmallVector<unsigned> indicesToInsertArgs;
+    SmallVector<Type> typesToInsertArgs;
+    SmallVector<DictionaryAttr> attrsToInsertArgs;
+    SmallVector<Location> locsToInsertArgs;
+    SmallVector<unsigned> newVargIndices;
+    size_t numNewArgsAdded = 0;
+    for (Value rValue : rValuesUsedBySubroutine) {
         if (isa<qref::QubitType>(rValue.getType())) {
-            newArg = f.getBody().front().addArgument(quantum::QubitType::get(ctx), loc);
-            regionTracker.setCurrentVQubit(rValue, newArg);
+            typesToInsertArgs.push_back(quantum::QubitType::get(ctx));
+            newVargIndices.push_back(originalNumArgs + (numNewArgsAdded++));
         }
         else if (isa<qref::QuregType>(rValue.getType())) {
-            newArg = f.getBody().front().addArgument(quantum::QuregType::get(ctx), loc);
-            regionTracker.setCurrentVQreg(rValue, newArg);
+            typesToInsertArgs.push_back(quantum::QuregType::get(ctx));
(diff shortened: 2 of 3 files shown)


kipawaa (approved):

jzaia18 (approved):
LGTM, seems to be fixing one of the errors in the benchmarks

frontend/catalyst/python_interface/compiler.py
```diff
@@ -65,6 +65,7 @@ def run(
         value_semantics_mlir = (
             catalyst.python_interface.inspection.xdsl_conversion._quantum_opt_stderr(
                 '--catalyst-pipeline="pipe(canonicalize;convert-to-value-semantics;canonicalize)"',
+                "--mlir-print-op-generic",
```
dime10, line 68: why is this added?

mlir/lib/QRef/Transforms/value_semantics_conversion.cpp
```diff
@@ -1749,6 +1749,11 @@ void handleWhile(IRRewriter &builder, scf::WhileOp whileOp, QubitValueTracker &t
 void handleSubroutine(IRRewriter &builder, func::FuncOp f,
                       const SetVector<Value> &rValuesUsedBySubroutine)
 {
+    // Special: when calling this conversion pass with generic format MLIR, the input qref
+    // subroutine might have an empty `res_attr` attribtue, e.g. from JAX serialization.
+    // We just remove it, since the subroutine return signature is changing.
+    f->removeAttr("res_attrs");
```
dime10: Since this has given us a headache before, should we just remove these attributes across the board at some higher stage?

frontend/catalyst/python_interface/compiler.py
```diff
@@ -79,13 +80,14 @@ def run(
 
         # JAX serializes void func.func ops with `res_attrs = []` in generic form
         # triggering an assertion in FuncToLLVM lowering.
-        # Assert that the canonicalization above has removed these empty attributes
```
dime10, line 82: I'm guessing the canonicalization this is referencing was not successful then given the change? 😅

frontend/catalyst/python_interface/compiler.py
paul0403: xdsl parser fails without it on `catalyst.callback` op on the benchmark in question, not sure why it was never caught before.

mlir/lib/QRef/Transforms/value_semantics_conversion.cpp
paul0403: `QJIT.generate_ir()`?

frontend/catalyst/python_interface/compiler.py
dime10: I mean is there a reason we _should_ print a generic format here? What you're saying might point to a bug in the xdsl assembly format for the op for example.

frontend/catalyst/python_interface/compiler.py
paul0403: It was originally just a `del` on these attributes, but codecov on my main qref PR said the `del` was never hit because the condition was never true :shrug: here I just change it back

frontend/catalyst/python_interface/compiler.py
paul0403: Oh I see. This was always generic format. See the code just above: 
```python
        # Convert to generic text format
        is_jax_module = isinstance(module, jaxModule)
        if is_jax_module:
            gentxtmod = module.operation.get_asm(
                binary=False, print_generic_op_form=True, assume_verified=True
            )
        else:
            gentxtmod = module
```
 (@mudit2812 do you know why? I guess precisely because not all catalyst ops have custom formats, making xdsl parser unhappy?). 

The value semantics conversion pass was actually outputting pretty format, so I'm making it back to generic format explicitly.

mlir/lib/QRef/Transforms/value_semantics_conversion.cpp
dime10: I looked into this again and I think we might be going about this the wrong way. I mean there is nothing wrong with canonicalizing the empty attribute away (as it's redundant), but I think the issue is also pointing at possibly invalid IR mutations in our passes. If a function had legitimate (non-empty) result attributes, and we added some results to it, MLIR would expect that the mutation properly updates the attributes too.
Not sure how exactly we go about adding the results right now, but if we are mutating functions in place apparently there helper methods in the function interface that handle this for us, like `insertFunctionResults`: 
```
  // There are 2 things that need to be updated:
  // - Function type.
  // - Result attrs.
```

frontend/catalyst/python_interface/compiler.py
dime10: Ah got it, all good then 👌

mlir/lib/QRef/Transforms/value_semantics_conversion.cpp
paul0403: True, in the semantics conversion there could be existing classical returns on the subroutine func op

paul0403: https://github.com/PennyLaneAI/catalyst/pull/2911/commits/e55da2a38f9bee4cfef67e02f056015b2dab944b

paul0403: I did update the function type when I first wrote the pass, but I didn't update the result attributes. Hopefully that's the only 2 things necessary if the upstream func op interface says so :+1:

dime10: Why not use the interface functions to add/remove results instead of doing this manually?

frontend/catalyst/python_interface/compiler.py
mudit2812: It's not just catalyst ops that may not have generic formats that are mirrored in the xDSL dialects, but also for dialects like stablehlo. The xDSL stablehlo is still incomplete, and parsing would fail for IRs using assembly format that happen to have stablehlo ops that are not in xDSL. This is also the same reason why we set the xDSL parsing to allow unregistered dialects.

There are other quirks as well. For example, xDSL did not have custom directives for the longest time, and I'm not confident that there won't be other printer/parser related disparities between xDSL and MLIR.

dime10: Thanks!

mlir/lib/QRef/Transforms/value_semantics_conversion.cpp
paul0403: Ahh yes :sweat-smile: found the upstream [example use](https://github.com/llvm/llvm-project/blob/102d8583355c316618648de7d8537299f4357f95/mlir/test/lib/IR/TestFunc.cpp#L86) and [test](https://github.com/llvm/llvm-project/blob/102d8583355c316618648de7d8537299f4357f95/mlir/test/IR/test-func-insert-result.mlir#L16)

updating

paul0403: https://github.com/PennyLaneAI/catalyst/pull/2911/commits/3fd18c76b7939917b43a6a958255aafa982b412c

frontend/catalyst/python_interface/compiler.py
```diff
@@ -79,13 +80,14 @@ def run(
...
+        # Remove empty arrays in-place so the generic printer omits them.\
```
paul0403:
```suggestion
        # Remove empty arrays in-place so the generic printer omits them.
```

PR #2915, opened 2026-06-04
[Remote 1.3] Add executor runtime

rniczh:
**Context:**

The executor dialect and its `convert-executor-to-llvm` lowering emit calls to a set of `__catalyst__executor__*` symbols, but nothing implements them yet. This PR adds the host-side runtime that backs those symbols.

The things we need to have:
1. TCP connection to a executor on the executor
2. Ships kernel objects into the executor ORC JIT
3. Marshals memref arguments and results across the wire
4. Manages session lifetime

Adds a new runtime component in `runtime/include/executorCAPI.h` for implementing `__catalyst__executor__open`,
`__catalyst__executor__send_binary`, `__catalyst__executor__launch`, 1__catalyst__executor__call_wrapper` and
`__catalyst__executor__close()`

Two processes, one TCP socket, ORC v2 EPC as the wire protocol. The host running the qjit-compiled program, and the executor catalyst executor process listening on a TCP port. Everything between them is carried by LLVM's ORC JIT v2 EPC protocol.

The host walks through five stages, top to bottom:
Open Sessions. The host connects to one or more executor addresses. Each successful connect produces executorSession that owns its half of the EPC channel, its ObjectLinkingLayer, and a 1 GB slab (Currently, we fix it to 1GB) reserved on the executor side. The session also resolves four "exported" symbols from the executor that the rest of the stages will need.

1. Send Binaries (.o or assets). For each qnode, the host pushes the cross-compiled object file across the channel, ORC JITLink lays it into the slab and applies relocations on the executor. The same path also handles assets (data files the kernel will read by path), we use it for `.hsa` in the future, which is via catalyst_executor_store_asset. The executor writes those to ${tmp}/catalyst-assets/<basename>.
2. executor Lib Calls. Direct invocations of plugin function in `.so`, which is used when you need to orchestrate something from the host and require to apply the change on the target device.
3. executor Launch. The qnode invocation, broken into three sub-steps:
	- Prepare arguments. For every input memref, the host calls catalyst_executor_alloc once for the data buffer and once for the descriptor executorly, then write the host-side data to the target.
	- executor Invoke. The host calls catalyst_executor_invoke which takes in the executor entry address, executor return value address, executor argument address as inputs. The executor dispatches to _catalyst_pyface_<name> symbol. - The kernel runs in the executor’s address space. Any references to `.so` are resolved by executor dlsym(RTLD_DEFAULT,...) against the loaded libraries.
	- Get the result from the executor. For each output memref the host does allocation then uses executor Read to copy the kernel-produced data back. The descriptor's pointer slots are rewritten with the host addresses so downstream LLVM IR can dereference them as ordinary memrefs.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-119223]
[sc-119310]


runtime/include/RemoteCAPI.h
```diff
@@ -0,0 +1,46 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// Remote Runtime API. This is expected to be called by the host program to establish a remote
+// session. Including open a session, send the binary to the remote, launch the kernel on the remote
+// and close the session.
+int __catalyst__remote__open(const char *addr);
+int __catalyst__remote__send_binary(const char *addr, const char *path, uint32_t format);
```
maliasadi:
```suggestion
int64_t __catalyst__remote__open(const char *addr);
int64_t __catalyst__remote__send_binary(const char *addr, const char *path, uint32_t format);
```

runtime/include/RemoteCAPI.h
```diff
@@ -0,0 +1,46 @@
...
+void __catalyst__remote__launch(const char *addr, const char *entry_symbol, size_t num_inputs,
+                                void *const *input_descs, const size_t *input_ranks,
+                                const size_t *input_elem_sizes, size_t num_outputs,
+                                void *const *output_descs, const size_t *output_ranks,
+                                const size_t *output_elem_sizes);
+int __catalyst__remote__call_wrapper(const char *addr, const char *symbol, const char *args_buf,
```
maliasadi:
```suggestion
int64_t __catalyst__remote__call_wrapper(const char *addr, const char *symbol, const char *args_buf,
```

runtime/include/RemoteCAPI.h
```diff
@@ -0,0 +1,46 @@
...
+                                     size_t args_size, void **out_buf, size_t *out_size);
+void __catalyst__remote__free_result(void *buf);
+
+int __catalyst__remote__close();
```
maliasadi:
```suggestion
int64_t __catalyst__remote__close();
```

maliasadi: I didn't see any logic for `remote_close` in the lowering PR #2910. Could you add `close` semantic and lowering to the other ones?

runtime/lib/capi/RuntimeCAPI.cpp
maliasadi, line 170: Now with dep to `rt_remote`, I'd probably add a check to ensure CTX is always initialized


```suggestion
    RT_FAIL_IF(!CTX || !CTX->getMemoryManager(),
        "Managed memory allocation requires an initialized runtime context.");
    void *ptr = malloc(size);
```

runtime/lib/capi/RuntimeCAPI.cpp
```diff
@@ -169,6 +169,8 @@ void *_mlir_memref_to_llvm_alloc(size_t size)
     return ptr;
 }
 
+void *__catalyst__rt__alloc_managed(size_t size) { return _mlir_memref_to_llvm_alloc(size); }
+
 void *_mlir_memref_to_llvm_aligned_alloc(size_t alignment, size_t size)
 {
     void *ptr = aligned_alloc(alignment, size);
```
maliasadi, line 181:
```suggestion
    RT_FAIL_IF(!CTX || !CTX->getMemoryManager(),
        "Managed memory allocation requires an initialized runtime context.");
    void *ptr = aligned_alloc(alignment, size);
```

runtime/lib/executor/ExecutorRuntime.cpp
```diff
@@ -0,0 +1,267 @@
...
+    ~RemoteEntry()
+    {
+        if (session) {
+            catalyst::remote::close(session);
+            session = nullptr;
+        }
+    }
+};
+
+std::mutex g_map_mu;
```
maliasadi, line 47: missing code comment

runtime/lib/executor/ExecutorSession.cpp
```diff
@@ -0,0 +1,748 @@
...
+        if (copy_data) {
+            void *aligned_host = *reinterpret_cast<void **>(desc_host + kAlignedOff);
+            if (aligned_host) {
+                remote_write(s, data_remote, aligned_host, data_size);
+            }
+        }
+    }
+    std::memcpy(desc.data() + kAllocatedOff, &data_remote, sizeof(uintptr_t));
+    std::memcpy(desc.data() + kAlignedOff, &data_remote, sizeof(uintptr_t));
+    std::memset(desc.data() + kOffsetOff, 0, sizeof(int64_t));
```
maliasadi, line 575: this always zeros the offset. could you add this assumption to doc or honor the incoming offset?

runtime/lib/remote/RemoteSession.cpp
```diff
@@ -0,0 +1,748 @@
...
+/**
+ * @brief Pull a remote memref descriptor + its data back into the host descriptor.
+ *
+ * @param s the session object
+ * @param remote_desc the remote address of the memref descriptor
+ * @param host_desc the host memref descriptor
+ * @param rank the rank of the memref
+ * @param elem_size the element size of the memref
+ */
+void pull_memref(RemoteSession *s, ExecutorAddr remote_desc, void *host_desc, size_t rank,
```
maliasadi: This isn't used anywhere.

runtime/lib/remote/RemoteSession.cpp
```diff
@@ -0,0 +1,748 @@
...
+
+/**
+ * @brief Run the kernel as a main function (take argv as arguments, argc is the length of argv).
+ *
+ * @param s the session object
+ * @param entry the entry function address
+ * @param argv the command line arguments
+ * @return int32_t the exit code
+ */
+int32_t run_as_main(RemoteSession *s, uint64_t entry_addr, int argc, const char *const *argv)
```
maliasadi: Where/how is this used? 🤔

runtime/lib/executor/ExecutorRuntime.cpp
```diff
@@ -0,0 +1,267 @@
...
+{
+    static const bool v = []() {
+        const char *e = std::getenv("CATALYST_REMOTE_VERBOSE");
+        return e && *e && *e != '0';
+    }();
+    return v;
+}
+
+// Look up or create the entry for `addr`.
+RemoteEntry *find_or_create_entry(const char *addr, bool create_if_missing)
```
maliasadi: could you add more docs here?

maliasadi: If `find_or_create_entry` locks `g_map_mu`, grabs `it->second.get()`, releases `g_map_mu` on return, while `__catalyst__remote__close` does the caller lock `entry->mu` => there will be a race!

maliasadi (commented):
Nice work @rniczh! I've taken my first pass at reviewing the E2E Remote Executor PRs. Here are my high-level comments from this first round of review 🐎

runtime/lib/executor/ExecutorRuntime.cpp
rniczh: Thanks for pointing out! I changed to use `shared_ptr` to resolve this issue in this commit https://github.com/PennyLaneAI/catalyst/pull/2915/commits/caa39b6cf1f3e3c5a506c2be15511e0441eb6f41

runtime/include/RemoteCAPI.h
rniczh: https://github.com/PennyLaneAI/catalyst/pull/2915/commits/550c925822b4fa015637aac6660f3ba0edec8829

runtime/include/RemoteCAPI.h
rniczh: https://github.com/PennyLaneAI/catalyst/pull/2915/commits/caa39b6cf1f3e3c5a506c2be15511e0441eb6f41

runtime/include/RemoteCAPI.h
rniczh: https://github.com/PennyLaneAI/catalyst/pull/2915/commits/caa39b6cf1f3e3c5a506c2be15511e0441eb6f41

runtime/include/RemoteCAPI.h
rniczh: Sure, I supported the `executor.close` in [dialect PR](https://github.com/PennyLaneAI/catalyst/pull/2909) and the lowering in [this PR](https://github.com/PennyLaneAI/catalyst/pull/2910)

About the lit test `mlir/test/Executor/ConvertExecutorToLLVM.mlir`:

```mlir
// CHECK-DAG: llvm.mlir.global internal constant @executor_addr_127_0_0_1_9000("127.0.0.1:9000\00")
// CHECK-DAG: llvm.func @__catalyst__executor__close(!llvm.ptr) -> i64
// CHECK-LABEL: func.func @close
func.func @close() {
  // CHECK-NOT: llvm.alloca
  // CHECK: llvm.call @__catalyst__executor__close(%{{.*}}) : (!llvm.ptr) -> i64
  executor.close("127.0.0.1:9000")
  return
}
```

runtime/lib/executor/ExecutorRuntime.cpp
rniczh: Added in this commit: https://github.com/PennyLaneAI/catalyst/pull/2915/commits/caa39b6cf1f3e3c5a506c2be15511e0441eb6f41

runtime/lib/executor/ExecutorSession.cpp
rniczh: No, it's not the assumption, we are missing to update the rebased address. So we have to rebase it to correct offset first https://github.com/PennyLaneAI/catalyst/pull/2915/commits/14e5d7d5eff263750dfa1e478378a953b17ae17e

```cpp
if (aligned_host) {
    int64_t host_offset = 0;
    std::memcpy(&host_offset, desc_host + kOffsetOff, sizeof(int64_t));
    char *src = static_cast<char *>(aligned_host) +
                host_offset * static_cast<int64_t>(elem_size);
    remote_write(s, data_remote, src, data_size);
}
```

runtime/lib/remote/RemoteSession.cpp
rniczh: Yes, it's legacy, just removed it https://github.com/PennyLaneAI/catalyst/pull/2915/commits/e3a9be35ef6f1a86653f58776ac3fed22f6367c0

runtime/lib/remote/RemoteSession.cpp
rniczh: No, this is a legacy function. There are two paths to invoke a remote kernel: One is converted from executor.launch -> invoke_kernel, and the other one is exector.call -> call_wrapper_raw. So no one will touch this function, it's dead code now. Just removed it: https://github.com/PennyLaneAI/catalyst/pull/2915/commits/ea1e0b9e82c13edfb3d03ebd97e275192517b72b

mehrdad2m (approved):

runtime/include/ExecutorCAPI.h
```diff
@@ -0,0 +1,47 @@
...
+#define EXECUTORCAPI_H
+
+#include <stddef.h>
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// Executor Runtime API. This is expected to be called by the host program to establish a executor
```
maliasadi:
```suggestion
// Executor Runtime API. This is expected to be called by the host program to establish an executor
```

maliasadi (approved):
Nice work! Happy to approve 🐎

PR #2917, opened 2026-06-04
Skip generating the `apply_t` subroutine clifford-only circuits

mehrdad2m:
**Context:**
Currently, the `Apply_T` subroutine is generated even if the circuit consists of only clifford gates, which results in a redundant function in the IR. 

skip generating the circuit if it clifford-only.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


lillian542:
These changes are now included along with other changes to avoid generating unnecessary subroutines in https://github.com/PennyLaneAI/catalyst/pull/2929 🎉

PR #2918, opened 2026-06-04
Fix memory bugs in the PBC passes

dime10:
Primarily, iterator invalidation (mutation while iterating). Also one instance of using an object after deletion.
These issues are hard to surface through unit testing, but are reported by the address sanitizer.


paul0403 (approved):
Ah, the classic!

lazypanda10117 (approved):
Thanks @dime10!

PR #2897, opened 2026-05-29
Bugfix for deallocate in qecp -> quantum conversion

lillian542:
**Context:**
We were deallocating using the SSA value produced by the `encode_zero` subroutine instead of the final SSA value for the register. This caused errors when we wanted to deallocate an individual codeblock and replace it with a codeblock generated by `fabricate [magic]`.

**Description of the Change:**
We fix it, and update the relevant test so that the bug would be caught were it still present.

PR #2881, opened 2026-05-27
Add support for `scf.for` ops in the `convert-quantum-to-qecl` pass

joeycarter:
**Context:** In order to support more general input programs in the experimental QEC compilation pipeline, we must add support for control-flow operations.

**Description of the Change:** This PR adds support for `for` loops (via [`scf.for`](https://mlir.llvm.org/docs/Dialects/SCFDialect/#scffor-scfforop) ops) in the `convert-quantum-to-qecl` pass. For example, an input program such as

```mlir
func.func @test_program() {
    %c1 = arith.constant 1 : index
    %c5 = arith.constant 5 : index
    %c0 = arith.constant 0 : index
    %0 = quantum.alloc(1) : !quantum.reg
    %1 = quantum.extract %0[0] : !quantum.reg -> !quantum.bit
    %2 = quantum.custom "PauliX"() %1 : !quantum.bit
    %3 = scf.for %arg0 = %c0 to %c5 step %c1 iter_args(%arg1 = %2) -> (!quantum.bit) {
        %out_qubits = quantum.custom "PauliX"() %arg1 : !quantum.bit
        scf.yield %out_qubits : !quantum.bit
    }
    %4 = quantum.insert %0[0], %3 : !quantum.reg, !quantum.bit
    quantum.dealloc %4 : !quantum.reg
    func.return
}
```

is lowered to

```mlir
func.func @test_program() {
    %c1 = arith.constant 1 : index
    %c5 = arith.constant 5 : index
    %c0 = arith.constant 0 : index
    %0 = qecl.alloc() : !qecl.hyperreg<1 x 1>
    %1 = qecl.extract_block %0[0] : !qecl.hyperreg<1 x 1> -> !qecl.codeblock<1>
    %2 = qecl.encode[zero] %1 : !qecl.codeblock<1>
    %3 = qecl.insert_block %0[0], %2 : !qecl.hyperreg<1 x 1>, !qecl.codeblock<1>
    %4 = qecl.extract_block %3[0] : !qecl.hyperreg<1 x 1> -> !qecl.codeblock<1>
    %5 = qecl.qec %4 : !qecl.codeblock<1>
    %6 = qecl.x %5[0] : !qecl.codeblock<1>
    %7 = qecl.qec %6 : !qecl.codeblock<1>
    %8 = scf.for %arg0 = %c0 to %c5 step %c1 iter_args(%arg1 = %7) -> (!qecl.codeblock<1>) {
[...]

PR #2919, opened 2026-06-04
Fix broken quantum-to-qecl lit tests and make them more permissive

joeycarter:
**Context:** PRs #2897 and #2881 conflicted with each other, but they were merged at approximately the same time so CI didn't catch the failures! :boom: 

The issue was that the `convert-quantum-to-qecl` unit lit tests of the control-flow conversion patterns were too restrictive and implicitly tested other functionality of the pass, which was modified by the bug-fix PR #2897. Specifically, that PR changed how we encode the codeblocks in a hyper-register after allocation: where before we inlined the `encode` op for a single-codeblock hyper-register, we now always put the `encode` ops in a for loop that iterates over each codeblock in the hyper-register. The lit tests assumed the inlined IR, and broke when the pass switched to the for-loop format. See here for an example failure: https://github.com/PennyLaneAI/catalyst/actions/runs/26969667866/job/79582101140?pr=2291

**Description of the Change:** This is an example of how unit lit tests that are too restrictive in their assumptions about the IR can break when the logic of other conversion patterns changes. As such, this PRs tries to relax some of the other assumptions in the control-flow unit tests, in addition to fixing the main issue of encoding in a for loop. What we loose in explicitly testing the data flow, we gain in flexibility to modify this pass to better suit the needs of the QEC pipeline, hopefully without having to make more hot-fix PRs like this one!


dime10 (approved):
Can't really judge the contents but I'm happy to provide one of the two checkmarks to unblock us 😅

lillian542 (approved):

lillian542:
Thanks @joeycarter, looks good to me. I feel in hindsight like we should have realized this would happen, but I guess that's how hindsight goes 🙈 

Thanks for fixing it, something to think about for adding lit tests in the future :)

PR #2922, opened 2026-06-05
Bugfix: support adjoint S in qecl->qecp lowering

lillian542:
**Context:**

When translating to the `qecp` dialect, we currently only generate a subroutine for `S`, and not for adjoint S. 

We add support for adjoint S. We add a subroutine by adding an `s_adj` in the `QecCode` definition for `Steane`, and we map `qecl.s %0 adj` to `s_adj` instead of `s` when selecting a subroutine. 

Currently S is the only qecl logical op that can have adjoint=True, but if other ops are added in the future, they can similarly have their lowering defined by adding `opname_adj` to the `QecCode`.

**Benefits:**
We can use logical `adjoint(S)` and get correct results.

[sc-121126]


frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecp_to_quantum.py
```diff
@@ -709,6 +709,81 @@ def ghz():
...
+            # x after hadamard does nothing, y and z flip expval(X)
+            ([qp.H, qp.X], (1, 0, 0)),
+            ([qp.H, qp.Y], (-1, 0, 0)),
+            ([qp.H, qp.Z], (-1, 0, 0)),
+            # adjoint and S adjoint rotate onto y axis in expected directions
+            ([qp.H, qp.S], (0, 1, 0)),
+            ([qp.H, lambda w: qp.adjoint(qp.S(w))], (0, -1, 0)),
+            # hadamard is self-inverse
+            ([qp.H, qp.H], (0, 0, 1)),
+        ],
```
lillian542, line 731: Added these more generally for the gates we've implemented with the Steane code, since we don't test them all with the GHZ circuit

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
```diff
@@ -940,6 +940,48 @@ def test_single_qubit_gate_lowering_Steane(
 
         run_filecheck(program, qecl_to_qecp_steane_pipeline)
 
+    def test_adjoint_s_lowering_Steane(self, run_filecheck, qecl_to_qecp_steane_pipeline):
+        """Test that using the Steane code lowers S adjoint as expected the physical S operator
+        is applied to all qubits in the codeblock)."""
```
joeycarter: [Grammar] Is this what you meant (missing parenthesis)?

```suggestion
        """Test that using the Steane code lowers S adjoint as expected (the physical S operator
        is applied to all qubits in the codeblock)."""
```

joeycarter (approved):

multiphaseCFD (approved):
Happy to approve

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
lillian542: It is, thanks!

PR #2923, opened 2026-06-05
Add support for reference semantics to the `resource-analysis` pass

kipawaa:
**Context:**
In order to analyze resources the operations must be in value semantics, which requires a conversion step for any op in reference semantics.

Adds direct support for reference semantics to the `resource-analysis` pass and underlying class.

**Benefits:**
Conversions are no longer needed for resource analysis.

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-121611]


mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -44,10 +47,11 @@ namespace catalyst {
 
 static bool isSkippedOp(Operation *op)
 {
-    return isa<quantum::ComputationalBasisOp, quantum::DeallocOp, quantum::DeallocQubitOp,
-               quantum::DeviceReleaseOp, quantum::ExtractOp, quantum::FinalizeOp,
-               quantum::HamiltonianOp, quantum::HermitianOp, quantum::InitializeOp,
-               quantum::InsertOp, quantum::NamedObsOp, quantum::NumQubitsOp, quantum::TensorOp,
+    return isa<quantum::ComputationalBasisOp, quantum::DeallocOp, qref::DeallocOp,
```
paul0403: Missed qref compbasis op here

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -70,14 +75,17 @@ static std::string getGateOpName(Operation *op, bool isAdjoint)
...
-            .Case<quantum::CustomOp>([](auto customOp) { return customOp.getGateName().str(); })
-            .Case<quantum::PauliRotOp>([](auto) { return "PauliRot"; })
-            .Case<quantum::GlobalPhaseOp>([](auto) { return "GlobalPhase"; })
-            .Case<quantum::MultiRZOp>([](auto) { return "MultiRZ"; })
-            .Case<quantum::PCPhaseOp>([](auto) { return "PCPhase"; })
-            .Case<quantum::QubitUnitaryOp>([](auto) { return "QubitUnitary"; })
-            .Case<quantum::SetStateOp>([](auto) { return "SetState"; })
-            .Case<quantum::SetBasisStateOp>([](auto) { return "SetBasisState"; })
+            .Case<quantum::CustomOp, qref::CustomOp>(
+                [](auto customOp) { return customOp.getGateName().str(); })
```
paul0403, line 79: Wow, cpp supports duck typing like this? You didn't need to cast? 🤔

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -86,6 +94,10 @@ static std::string getGateOpName(Operation *op, bool isAdjoint)
         isAdjoint ^= gate.getAdjointFlag();
     }
 
+    if (auto gate = dyn_cast<qref::QuantumGate>(op)) {
+        isAdjoint ^= gate.getAdjointFlag();
+    }
```
paul0403, line 99: Like here, I assume duck typing on `getAdjointFlag` fails to compile.

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
paul0403: So the reference semantics pbc and mbqc ops are not in `qref` dialect, but in `pbc` and `mbqc` dialect:
- `pbc::RefPPMeasurementOp`
- `MBQC::RefMeasureInBasisOp`
- `MBQC::RefGraphStatePrepOp`

Can you update these three correspondingly too?

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -147,11 +165,11 @@ static std::string getObservableName(Operation *obsOp)
     }
 
     return llvm::TypeSwitch<Operation *, std::string>(obsOp)
-        .Case<quantum::ComputationalBasisOp>([](auto cbOp) {
+        .Case<quantum::ComputationalBasisOp, qref::ComputationalBasisOp>([](auto cbOp) {
             unsigned n = cbOp.getQubits().size();
```
paul0403, line 169: Same duck typing comment

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -147,11 +165,11 @@ static std::string getObservableName(Operation *obsOp)
...
             return n == 0 ? std::string("all wires") : std::to_string(n) + " wires";
         })
-        .Case<quantum::NamedObsOp>(
+        .Case<quantum::NamedObsOp, qref::NamedObsOp>(
             [](auto op) { return stringifyNamedObservable(op.getType()).str(); })
```
paul0403, line 173: Same duck typing comment

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -380,9 +398,12 @@ void ResourceAnalysis::analyzeRegion(Region &region, ResourceResult &result, boo
 {
     for (Block &block : region) {
         for (Operation &op : block) {
+            bool needsCollection = true;
+
             llvm::TypeSwitch<Operation &, void>(op)
-                .Case([&](quantum::AdjointOp adjOp) {
+                .Case<quantum::AdjointOp, qref::AdjointOp>([&](auto adjOp) {
                     analyzeRegion(adjOp.getRegion(), result, !isAdjoint);
```
paul0403, line 429: Same duck typing comment

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -380,9 +398,12 @@ void ResourceAnalysis::analyzeRegion(Region &region, ResourceResult &result, boo
 {
     for (Block &block : region) {
         for (Operation &op : block) {
+            bool needsCollection = true;
```
paul0403, line 425: What is this new flag for?

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -474,6 +499,19 @@ void ResourceAnalysis::collectOperation(Operation *op, ResourceResult &result, b
         return;
     }
 
+    // Metadata: qubit allocations
+    if (auto allocOp = dyn_cast<qref::AllocOp>(op)) {
+        uint64_t nqubits = allocOp.getNqubitsAttr().value_or(0);
+        result.numAllocQubits += nqubits;
+        return;
+    }
```
paul0403, line 526: You didn't duck type here... is it because llvm type switch provides duck typing sugars? 🤯

mlir/include/Catalyst/Analysis/ResourceResult.h
paul0403: There could also be qref bit/reg arguments.

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
kipawaa: This is specific to `llvm::TypeSwitch`; from https://llvm.org/doxygen/TypeSwitch_8h_source.html:
```
  /// Invoke a case on the derived class with multiple case types.
  template <typename CaseT, typename CaseT2, typename... CaseTs,
            typename CallableT>
```

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
kipawaa: This is to avoid reaching `collectOperation` on `AdjointOp`s (the region version of adjoint. When they reach `collectOperation` they effectively no-op, but they raise a warning which incorrectly suggests that they aren't handled correctly (see https://github.com/PennyLaneAI/catalyst/blob/157fb1558ec2d543e9acbd9b8312753b46ca42ce/mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp#L535).

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
paul0403: Ah, the cast is inside the typeswitch impl already! :+1:

mlir/include/Catalyst/Analysis/ResourceResult.h
kipawaa: Thanks! Added in [60bf494](https://github.com/PennyLaneAI/catalyst/pull/2923/commits/60bf494a4d9562e4fc25304aa63fd5eb929c20be)

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
kipawaa: Good catch, added in [5bac4e7](https://github.com/PennyLaneAI/catalyst/pull/2923/commits/5bac4e779224e3f8f2ca7e5f7d939caa0117b82d) (github really botched that diff 🥴)

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
kipawaa: Thanks! Added here: [b67eb41](https://github.com/PennyLaneAI/catalyst/pull/2923/commits/b67eb4135ed34c6d09ef207822d52ffdb465086f)

sengthai (approved):
Look good at ResourceAnalysis side! Thanks @kipawaa!

mlir/test/Catalyst/ResourceAnalysisTest.mlir
```diff
@@ -835,3 +835,51 @@ func.func @auto_qm_flag_unset() {
...
+
+    %meas = qref.measure %1 : i1
+
+    %pbc_meas = pbc.ref.ppm ["X", "Z"] %1, %2: i1
+
+    %angle = arith.constant 4.0 : f64
+    %mbqc_meas = mbqc.ref.measure_in_basis [ZX, %angle] %2 : i1
+
+    %adj_matrix = arith.constant dense<[1, 0, 1, 0, 0, 1]> : tensor<6xi1>
+    %graph_reg = mbqc.graph_state_prep (%adj_matrix : tensor<6xi1>) [init "Hadamard", entangle "CZ"] : !quantum.reg
```
paul0403: was this supposed to be the reference semantics version? 😅

paul0403: Note that if tensor size is 6, the reg size is 4

kipawaa: Woops 🤣  fixed in [dc55554](https://github.com/PennyLaneAI/catalyst/pull/2923/commits/dc55554eed361b4fbd22375d2a6ed65dca9b62a0)

kipawaa: I wonder if that should count as qubit allocation... 😅 maybe I'll check with Joey

paul0403 (approved):
Beautiful 💯

mlir/test/Catalyst/ResourceAnalysisTest.mlir
paul0403: True, I'll leave this conversation open pending on Joey's green light

paul0403: In all my semantics conversion logic, I did treat this as allocation, because unlike the `quantum.set_state` operations that operate on existing qubit values, this `mbqc.graph_state_prep` op creates a qreg out of nothing, suggesting an allocation semantic.

kipawaa: At the same time, the existing `mbqc.graph_state_prep` operation doesn't seem to count towards qubit allocation... @joeycarter @sengthai is this something I should add while I'm here making changes to `resource-analysis`, or should we leave as-is?

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -186,6 +204,52 @@ static std::string getMeasurementName(Operation *op)
...
+
+    int64_t m = shapedType.getDimSize(0);
+
+    int64_t discriminant = 1 + 8 * m;
+
+    int64_t sqrtDiscriminant = std::sqrt(discriminant);
+    llvm::errs() << "found mbqc graph num qubits: " << (1 + sqrtDiscriminant) / 2 << "\n";
+    return (1 + sqrtDiscriminant) / 2;
+}
```
joeycarter: Ah @paul0403 also added this formula in #2920 🙈 

Maybe it would be worth putting it in some common location. But, this is non-blocking from my perspective—I'll let you two coordinate.

joeycarter (approved):
Looks good, thanks @kipawaa!

Nothing blocking on my end, but I did note that we now have two PRs that are adding the same formula for the MBQC adjacency matrix!

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
paul0403: I'll do that

paul0403: https://github.com/PennyLaneAI/catalyst/pull/2923/commits/023148e7e533073df7ff8fa7bb1718b03a5e9145

PR #2924, opened 2026-06-05
Remove libboost-dev src installation dep

maliasadi:
**Context:**
Remove libboost-dev src installation dep. Tested locally and on manylinux_2_28 x86/aarch64

**Benefits:**
- Remove libboost-dev src installation dep on CIs
- No impacts on wheels recipes.

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-121685]


kipawaa (approved):
Thanks @maliasadi !

mlir/lib/Quantum/Transforms/DecompGraphSolver/CMakeLists.txt
```diff
@@ -41,6 +40,8 @@ else()
 
     FetchContent_MakeAvailable(Boost)
 
+    # Ensure the cache variable is consistent with the found branch
+    set(CATALYST_BOOST_FETCHED TRUE CACHE BOOL "Boost was fetched via FetchContent" FORCE)
```
paul0403: Oh! The cache string was never set? 😅 This is why locally boost was always making a new build every time people build dialects?

.github/workflows/check-catalyst.yaml
```diff
@@ -51,7 +51,7 @@ jobs:
       - name: Install dependencies
         run: |
           sudo apt-get update
-          sudo apt-get -y -q install ninja-build make cmake clang libboost-all-dev
```
paul0403, line 54: Just to confirm, the idea is to do the necessary boost install in cmake, so the runners don't need to install `libboost-all-dev` dependency explicitly themselves?

runtime/lib/CMakeLists.txt
```diff
@@ -37,6 +37,24 @@ else()
 
     FetchContent_MakeAvailable(Boost)
 
+    # Suppress -Werror on the fetched Boost targets so they build cleanly.
+    set(BOOST_TARGETS boost_container boost_random boost_core
+        boost_assert boost_throw_exception boost_config)
```
paul0403, line 42: Are these not fetched by `graph`, but only fetched by `multiprecision` and `random`? 🤔 Why didn't we need to silence the warnings in the graph solver cmake?

mlir/lib/Quantum/Transforms/DecompGraphSolver/CMakeLists.txt
dime10: Who is the consumer of this flag?

paul0403: Oh never mind, the checked flag is `Boost_FOUND`.

Apparantly this flag is only used in runtime? https://github.com/search?q=repo%3APennyLaneAI%2Fcatalyst%20CATALYST_BOOST_FETCHED&type=code

.github/workflows/check-catalyst.yaml
maliasadi: Yes! This will be taken care of by CMake on systems without Boost.

runtime/lib/CMakeLists.txt
maliasadi: We've already silenced the graph-solver warnings [here](https://github.com/PennyLaneAI/catalyst/blob/f78fb42c050ebb4a9051075d8561edf5d1234c6c/mlir/lib/Quantum/Transforms/DecompGraphSolver/CMakeLists.txt#L47).

paul0403 (approved):

PR #2926, opened 2026-06-08
Add integration tests to check control-flow in `convert-qecl-to-qecp`

joeycarter:
**Context:** Support for structured-control-flow operations in the `convert-quantum-to-qecl` pass were recently added in #2872 (`scf.if`), #2881 (`scf.for`) and #2905 (`scf.while`). Now we want to ensure that we support these control-flow operations lower in the QEC pipeline, in the `convert-qecl-to-qecp` pass.

**Description of the Change:** It turns out we get all the necessary `scf` conversions for free since we can map `qecl` types directly to `qecp` types! This PR only adds tests to ensure that an input PennyLane program containing control flow can be correctly lowered through the QEC pipeline to the `qecp` layer.

[sc-120287]


frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
```diff
@@ -1154,6 +1154,129 @@ def test_apply_t_steane(self, run_filecheck, qecl_to_qecp_steane_pipeline):
...
+class TestControlFlow:
+    """Test suite for control-flow conversion patterns in the convert-qecl-to-qecp pass."""
+
+    def test_scf_if_integration(self, run_filecheck_qjit):
+        """Test the convert-quantum-to-qecl and convert-qecl-to-qecp passes together on a simple
+        program with an if statement.
+        """
+        dev = qp.device("null.qubit", wires=1)
+
+        @qp.qjit(capture=True, target="mlir")
```
multiphaseCFD, line 1228: Just wondering if `autograph=True` is supported here? If yes, I'd recommend to add a few tests for that support. If not, it would be nice to add this feature support to TODOs for future reference. Thanks

multiphaseCFD (approved):
LGTM! Thanks @joeycarter  Just one comment

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
joeycarter: Hm, I'm on the fence about that. Autograph support should be unaffected by downstream compilation passes, so adding an equivalent test using autograph would amount to a duplicated test, I would say. That's because the capture of control-flow operations, whether using PennyLane's constructs like `qp.for_loop` and `qp.cond` or using autograph, is all happening in the frontend, well upstream of the QEC layer in the pipeline. I would expect the initial program representation seen by the `quantum-to-qecl` pass would be approximately the same, modulo a few tensor operations perhaps, but the `scf` ops at least should all be functionally the same.

As an example, these two programs, with and without autograph, both generate the same IR:

```python
@catalyst.qjit(keep_intermediate=True, autograph=True)
@qp.qnode(dev)
def circuit_yes_ag():
    for i in range(5):
        qp.H(0)
    
    m0 = qp.measure(0)
    if m0:
        qp.X(0)

    return qp.state()
```

```python
@catalyst.qjit(keep_intermediate=True, autograph=False)
@qp.qnode(dev)
def circuit_no_ag():
    @qp.for_loop(0, 5, 1)
    def loop(i):
        qp.H(0)
    
    loop()
    
    m0 = qp.measure(0)
    qp.cond(m0, qp.X)(0)

    return qp.state()
```

IR:

```mlir
func.func public @circuit_no_ag() -> tensor<2xcomplex<f64>> attributes {quantum.node} {
  %c1 = arith.constant 1 : index
  %c5 = arith.constant 5 : index
  %c0 = arith.constant 0 : index
  %1 = scf.for %arg0 = %c0 to %c5 step %c1 iter_args(%arg1 = %0) -> (!quantum.reg) {
    %7 = quantum.extract %arg1[ 0] : !quantum.reg -> !quantum.bit
    %out_qubits = quantum.custom "Hadamard"() %7 : !quantum.bit
    %8 = quantum.insert %arg1[ 0], %out_qubits : !quantum.reg, !quantum.bit
    scf.yield %8 : !quantum.reg
  }
  %2 = quantum.extract %1[ 0] : !quantum.reg -> !quantum.bit
  %mres, %out_qubit = quantum.measure %2 : i1, !quantum.bit
  %3 = quantum.insert %1[ 0], %out_qubit : !quantum.reg, !quantum.bit
  %4 = scf.if %mres -> (!quantum.reg) {
    %7 = quantum.extract %3[ 0] : !quantum.reg -> !quantum.bit
    %out_qubits = quantum.custom "PauliX"() %7 : !quantum.bit
    %8 = quantum.insert %3[ 0], %out_qubits : !quantum.reg, !quantum.bit
    scf.yield %8 : !quantum.reg
  } else {
    scf.yield %3 : !quantum.reg
  }
  %5 = quantum.compbasis qreg %4 : !quantum.obs
  %6 = quantum.state %5 : tensor<2xcomplex<f64>>
  quantum.device_release
  return %6 : tensor<2xcomplex<f64>>
}
```

mehrdad2m (approved):

PR #2927, opened 2026-06-08
Add integration tests to check control-flow in `convert-qecp-to-quantum`

joeycarter:
**Context:** Support for structured-control-flow operations in the `convert-quantum-to-qecl` pass were recently added in #2872 (`scf.if`), #2881 (`scf.for`) and #2905 (`scf.while`). We also confirmed support for these control-flow ops in the `convert-qecl-to-qecp` pass in #2926. Now we want to ensure that we support these control-flow operations in the QEC **validation** pipeline, by means of the `convert-qecp-to-quantum` pass.

**Description of the Change:** Like before, we get _most_ of the necessary `scf` conversions for free since we can map `qecp` types directly back to `quantum` types (with the exception of mapping `!qecp.hyperreg` types back to collections of `!quantum.reg` types, for which we do not currently have an abstraction in the `quantum` dialect). This PR mainly only adds tests to ensure that an input PennyLane program containing control flow can be correctly lowered through the QEC pipeline and back to the `quantum` dialect.

This PR also adds an error message when attempting to handle `qecp.extract_block` ops where the `idx` operand is not known statically.

[sc-120289]


frontend/catalyst/python_interface/transforms/qecp/convert_qecp_to_quantum.py
```diff
@@ -316,16 +319,17 @@ class ConvertQecPhysicalToQuantumPass(ModulePass):
...
         2. The encoding loop operation is unrolled.
         3. `qecp.extract_codeblock` operations are removed from the IR by replacing the uses with
-        the corresponding quantum.reg SSA value.
+           the corresponding quantum.reg SSA value.
         4. `qecp.insert_codeblock` operations are replaced with `quantum.dealloc` operation.
+
         NOTE: The current implementation only targets the 3-logical qubit GHZ circuit. The
         implementation is based on the IR structure of the specific circuit.
-        TODO: We might come back to update the logic below to support 1-logical qubit circuits,
-        where there is no ForOp encoding loop in the IR.
```
joeycarter, line 328: This should be resolved now since we always perform the encoding in a for loop.

frontend/catalyst/python_interface/transforms/qecp/convert_qecp_to_quantum.py
```diff
@@ -18,8 +18,11 @@
 
 Known Limitations
 -----------------
-  * The current hyper-register lowering implementation also does not support any control flow that
-    iterates over hyper registers, except for the encoding loop.
+
+  * The current hyper-register lowering implementation not support control-flow operations that
+    iterate over physical hyper-registers, except for the encoding loop after allocating the
+    hyper-register. However, control-flow ops that iterate over types, like physical codeblocks, are
+    supported.
```
joeycarter: The reference semantics migration came in at an opportune time, since now we don't pack qubits into the register before entering a control-flow region, which means `for` loops in the frontend take _qubits_ as iteration arguments, which are lowered to physical codeblocks in the `qecp` layer, which can be directly mapped back to `!quantum.reg` types in the `qecp-to-quantum` pass.

joeycarter: Just realized an exception to this rule: in the frontend, when the iteration index is used as the wire index, e.g.

```python
for i in range(10):
    qp.H(wires=i)
```

then the `scf.for` loop in the IR needs to take the register as an iteration argument, with `extract` and `insert` ops at that index within within the body of the `for` loop. I don't have a good workaround for this case, besides unrolling the loop like we do for `encode` ops, but I'd like to avoid that. Regardless, this point in the known limitations is still correct.

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecp_to_quantum.py
```diff
@@ -754,3 +883,28 @@ def circ():
...
+        @convert_qecp_to_quantum_pass
+        @convert_qecl_to_qecp_pass(qec_code="Steane", number_errors=1)
+        @inject_noise_to_qecl_pass
+        @convert_quantum_to_qecl_pass(k=1)
+        @qp.set_shots(10)
+        @qp.qnode(dev, mcm_method="one-shot")
+        def circ():
+            @qp.for_loop(0, 10, 1)
+            def loop(i):
+                qp.cond(i % 2 == 0, qp.X, qp.Z)(0)
```
multiphaseCFD, line 1215: 👍

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecp_to_quantum.py
```diff
@@ -624,6 +624,135 @@ def test_hyperregister_lowering(self, run_filecheck):
...
+
+
+class TestControlFlow:
+    """Test suite for control-flow conversion patterns in the convert-qecp-to-quantum pass."""
+
+    def test_scf_if_integration(self, run_filecheck_qjit):
+        """Test the QEC pipeline ending in the convert-qecp-to-quantum pass on a simple program with
+        an if statement.
+        """
+        dev = qp.device("null.qubit", wires=1)
```
multiphaseCFD, line 788: could we add a few cases on `>1` qubit circuit? Also add `CNOT` gate?

multiphaseCFD (commented):
Thanks @joeycarter ! LGTM, just a couple of Qs

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecp_to_quantum.py
joeycarter: This was a good suggestion because it revealed a bug in the `convert-qecp-to-quantum` pass (unrelated to control flow, _I think_) 🫠 

I'm still investigating, but long story short, if you have in the `qecp` layer, for example

```mlir
%13 = func.call @qec_cycle_Steane(%12) : (!qecp.codeblock<1 x 7>) -> !qecp.codeblock<1 x 7>
...
%16, %17 = scf.if %11 -> (!qecp.codeblock<1 x 7>, !qecp.codeblock<1 x 7>) {
  %18 = func.call @x_Steane(%13) : (!qecp.codeblock<1 x 7>) -> !qecp.codeblock<1 x 7>
  %19 = func.call @qec_cycle_Steane(%18) : (!qecp.codeblock<1 x 7>) -> !qecp.codeblock<1 x 7>
  scf.yield %19, %15 : !qecp.codeblock<1 x 7>, !qecp.codeblock<1 x 7>
} else {
  %20 = func.call @z_Steane(%15) : (!qecp.codeblock<1 x 7>) -> !qecp.codeblock<1 x 7>
  %21 = func.call @qec_cycle_Steane(%20) : (!qecp.codeblock<1 x 7>) -> !qecp.codeblock<1 x 7>
  scf.yield %13, %21 : !qecp.codeblock<1 x 7>, !qecp.codeblock<1 x 7>
}
%22 = qecp.insert_block %6[0], %16 : !qecp.hyperreg<2 x 1 x 7>, !qecp.codeblock<1 x 7>
%23 = qecp.insert_block %22[1], %17 : !qecp.hyperreg<2 x 1 x 7>, !qecp.codeblock<1 x 7>
%24 = qecp.extract_block %23[0] : !qecp.hyperreg<2 x 1 x 7> -> !qecp.codeblock<1 x 7>
%25 = func.call @qec_cycle_Steane(%24) : (!qecp.codeblock<1 x 7>) -> !qecp.codeblock<1 x 7>
```

converting back to the `quantum` dialect gives

```mlir
%3 = func.call @encode_zero_Steane(%2) : (!quantum.reg) -> !quantum.reg
...
%10, %11 = scf.if %7 -> (!quantum.reg, !quantum.reg) {
  %12 = func.call @x_Steane(%8) : (!quantum.reg) -> !quantum.reg
  %13 = func.call @qec_cycle_Steane(%12) : (!quantum.reg) -> !quantum.reg
  scf.yield %13, %9 : !quantum.reg, !quantum.reg
} else {
  %14 = func.call @z_Steane(%9) : (!quantum.reg) -> !quantum.reg
  %15 = func.call @qec_cycle_Steane(%14) : (!quantum.reg) -> !quantum.reg
  scf.yield %8, %15 : !quantum.reg, !quantum.reg
}
%16 = func.call @qec_cycle_Steane(%3) : (!quantum.reg) -> !quantum.reg
```

Notice the final QEC cycle isn't using the output of the `if` op, but the last `!quantum.reg` before the `if`. I think the `qecp-to-quantum` pass is getting tripped up on the series of `insert_block` and `extract_block` ops. I'm investigating...

joeycarter: So the issue appears to be that we replace the uses of `qecp.insert_block` ops' output hyper-register with the result of the last-matched `qecp.alloc` op and deallocate the codeblock. We'll need to fix this to support cases where there are `extract_block` ops after `insert_block` ops, like the above.

joeycarter: I have a bugfix PR here to address this issue: #2955. We can come back to this thread once that's merged.

joeycarter: Test with `scf.if` and two qubits added: 03b5f6b2b

mehrdad2m (approved):
LGTM Thanks @joeycarter

multiphaseCFD (approved):
Thanks @joeycarter for putting these together

PR #2928, opened 2026-06-09
Test generality with Shor [[9, 1, 3]]

lillian542:
**Context:**

The initial implementation is limited to CSS codes with k=1 (but written with extensibility to k>1 in mind). However, in-depth testing has focused on the Steane code, since we can define a complete Clifford+T gate set for this code, and it is small enough to run on `lightning.qubit` with noise and validate that everything is working. 

Because of this, there is a risk that we've made assumptions that are specific to the Steane code. To identify these, we would like to test with another small k=1 CSS code. 

The LUT decoder is hardcoded to [7, 1, 3] right now, so we can't validate executing a circuit now - that will come later. This PR focuses on validating that compilation works as expected.

We define Shor's 9-qubit repetition code in the `QecCode` registry, and test that compiling with it behaves as expected.

This identified a number of places where we've made incorrect assumptions:

- In the `QecCode` definition, the format for defining the unitary encoder is not flexible enough; it currently assumes every enocder will consist of Hadamards followed by CNOTs, rather than allowing these gates to be interspersed (or allowing additional gates)
- In the QEC code cycle, we allocate auxiliary qubits based on the number of rows in the X-tanner graph for every repetition (including Z checks). This fails for codes like the 9-qubit Shor code where a different number of auxiliary qubits is needed for X and Z checks.
- Need to support non-uniform transversal gate definitions. Given that the logical Pauli Y is `YXXZIIZII` in the Shor code, it is not compatible with the current system to define a transversal gate by a single gate and its indices in the codeblock, e.g. `{"x": (qecp.PauliXOp, [4, 5, 6])}`. We've therefore developed a new system to handle transversal gates with more flexibility, so now we can have, for example, `{"y": ("Y", "X", "X", "Z", "I", "I", "Z", "I", "I")}`.
- Need to support non-Pauli-Z ops in the logical Pauli Z observable for the transversal measurement protocol. Given that the logical Pauli Z is `XXXIIIIII` in the Shor code, this means that our current transversal measurement protocol, which measures all codeblock qubits in the computational basis, won't work. Instead, what we should be doing is measuring each codeblock qubit in the basis given by the corresponding physical Pauli op in the logical Pauli observable definition. In other words, to measure logical Pauli Z, we should measure physical codeblock qubits 0, 1, and 2 in the X basis. To do so, we've gone the route of inserting diagonalizing gates before applying the physical computational-basis measurement ops.
- The decode-physical-measurement subroutine required updating. Since we're no longer measuring each codeblock qubit, but instead only the non-identity elements in the logical Pauli observable, we altered the decode-physical-measurement subroutine to use the MLIR [`linalg.reduce`](https://mlir.llvm.org/docs/Dialects/Linalg/#linalgreduce-linalgreduceop) op to compute the XOR of all measurement results rather than picking out the bits at the indices given by the logical Pauli Z definition and inlining the XOR ops.

These assumptions are corrected in this PR.

**Benefits:**

The generality of our implementation for k=1 CSS codes is tested to some extent. 

**Possible Drawbacks:**

- Any assumptions that are true for Shor and Steane will not be uncovered by this
- This code doesn't have a transversal H or S gate. This means we can't test application of T gates (requires logical S for corrections in the protocol) in quantum-->qecl-->qecp compilation, and it also means that when we get it running end-to-end with the LUT decoder, we won't be able to test that phase-flip errors are caught and corrected as expected.
- We've used diagonalizing gates for non-Z-basis measurements. In the future, we may have native Pauli-observable primitives in the QEC pipeline that could be used instead.

Here we balance a desire for additional validation with time constraints.

[sc-119895]


joeycarter:
@lillian542 It turned out that some fairly large changes were required to support the Shor code in the QEC pipeline. Below is a quick summary to help navigate the three main changes I've made:

* **Needed to support non-uniform transversal gate definitions**
    * Given that the logical Pauli X is `ZIIZIIZII` and the logical Pauli Z is `XXXIIIIII` in the Shor code, this means we can also define the logical Pauli Y as `YXXZIIZII`.
    * This logical Pauli Y is not compatible with the current system to define a transversal gate by a single gate and its indices in the codeblock, e.g. `{"x": (qecp.PauliXOp, [4, 5, 6])}`.
    * I wrote a new system to handle transversal gates with more flexibility, so now we can have, for example, `{"y": ("Y", "X", "X", "Z", "I", "I", "Z", "I", "I")}`. I added a helper method to translate these strings into their corresponding `qecp` ops.
* **Needed to support non-Pauli-Z ops in the logical Pauli Z observable for the transversal measurement protocol**
    * Given that the logical Pauli Z is `XXXIIIIII` in the Shor code, this means that our current transversal measurement protocol, which measures all codeblock qubits in the computational basis, won't work.
    * Instead, what we should be doing is measuring each codeblock qubit in the basis given by the corresponding physical Pauli op in the logical Pauli observable definition. In other words, to measure logical Pauli Z, we should measure physical codeblock qubits 0, 1, and 2 in the X basis. To do so, I've gone the route of inserting diagonalizing gates before applying the physical computational-basis measurement ops.
    * I've also removed the measurement ops on the non-identity elements in the logical Pauli definition, since I think that strictly speaking, this is not correct. Really this amounts to a PPM on the codeblock, and a measurement in the `I` basis is equivalent to no measurement at all.
    * As an aside, this change should make it fairly easy to support logical Pauli X and Y terminal measurements, since the protocol is essentially the same as for Pauli Z, but just swap in the desired Pauli observable definition. I've left this for a future PR, though.
* **Updated the decode-physical-measurement subroutine**
    * Since we're no longer measuring each codeblock qubit, but instead only the non-identity elements in the logical Pauli observable, I had to tweak the decode-physical-measurement subroutine.
    * Before we passed in the measurement results of every codeblock qubit and picked out the bits at the indices given by the logical Pauli Z definition. This is no longer necessary since the transversal-measurement subroutine now only returns these measurement results.
    * This means that the decode subroutine can be rewritten as an XOR op applied over every element of the input tensor of bits.
    * Instead of inlining all of the XOR ops explicitly, I used the MLIR [`linalg.reduce`](https://mlir.llvm.org/docs/Dialects/Linalg/#linalgreduce-linalgreduceop) op to represent these operations more compactly.

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -640,75 +654,88 @@ def insert_tanner_graph_ops_into_block(
...
         """Create the subroutine that performs the transversal measurement of a physical codeblock
-        in the computational basis. All qubits in the codeblock are measured directly.
+        in the computational basis.
+
+        The physical measurement operations are applied on the codeblock qubits as defined in the
+        logical Pauli Z operation of the QEC code. For instance, if the QEC defines the logical
+        Pauli Z operation as "ZIZIZI", codeblock qubits 0, 2, and 3 are measured, and their outcomes
+        are returned as a tensor<3xi1>. For logical Pauli Z operations that contain _physical_ Pauli
+        gates other than Z, the appropriate diagonalizing gates are applied before measuring (i.e.,
+        H for Pauli X measurements and HS† for Pauli Y measurements).
```
lillian542, line 662: Longer-term I think we should separate out creating the measurement instructions and diagonalizing them, since we anticipate MCMs in other bases being a native instruction for the control hardware. But that can wait 😅

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
```diff
@@ -640,75 +654,88 @@ def insert_tanner_graph_ops_into_block(
...
+                gate_op = pauli_z_gate_ops[i]
+                match gate_op:
+                    case "Z":
+                        pass
+                    case "X":
+                        hadamard_op = qecp.HadamardOp(qubits[i])
+                        qubits[i] = hadamard_op.out_qubit
+                    case "Y":
+                        hadamard_op = qecp.HadamardOp(qubits[i])
+                        s_adj_op = qecp.SOp(hadamard_op.out_qubit, adjoint=True)
```
lillian542: I believe we need to reverse the order of these, HS† is applied right-to-left

frontend/catalyst/python_interface/transforms/qecp/qec_code_lib.py
```diff
@@ -15,62 +15,173 @@
...
-                qecp.PauliYOp,
-                [4, 5, 6],
-            ),
-            "z": (qecp.PauliZOp, [4, 5, 6]),
-            "hadamard": (qecp.HadamardOp, [0, 1, 2, 3, 4, 5, 6]),
-            "s": (partial(qecp.SOp, adjoint=True), [0, 1, 2, 3, 4, 5, 6]),
-            "s_adj": (qecp.SOp, [0, 1, 2, 3, 4, 5, 6]),
+            # Keys need to match the names of the corresponding qecl.gate gates; if any adjoint
+            # gates are supported, they should be included as a separate entry with key
+            # "gatename_adj". Values are a tuple of the qecp gate, and the indices its applied at in
```
lillian542:
```suggestion
            # "gatename_adj". Values are a tuple of gate string identifiers specifying the qecp ops
```

frontend/catalyst/python_interface/transforms/qecp/qec_code_lib.py
```diff
@@ -15,62 +15,173 @@
...
-                [2, 0],
-                [3, 5],
-                [6, 4],
-                [2, 6],
-                [3, 4],
-                [1, 5],
-                [1, 6],
-                [3, 0],
-            ),
+            # ops (in the form of a qecp operator and the indices of the codeblock
```
lillian542:
```suggestion
            # ops (in the form of a gate string identifier and the indices of the codeblock
```

frontend/test/pytest/python_interface/transforms/qecp/test_qec_code_lib.py
```diff
@@ -217,3 +226,92 @@ def test_get_unsupported_code(self, name):
...
+                x_tanner=np.ones(n, dtype=int),
+                z_tanner=np.ones(n, dtype=int),
+                transversal_1q_gates=transversal_1q_gates,
+                transversal_2q_gates={"cnot": "CNOT"},
+                unitary_encoding={},
+            )
+
+
+class TestGateStringIds:
+    """TODO"""
```
lillian542:
```suggestion
    """Test that gate string ID to qecp operation works as expected."""
```

lillian542: Or something 🤷‍♀️

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
```diff
@@ -74,21 +71,25 @@ def _make_qec_code(
...
+            hadamard_ops = tuple(("H", [i]) for i in range(n) if i % 2)
+            cnot_ops = tuple(("CNOT", [i, i + 1]) for i in range(n - 1))
+
             unitary_encoding = {
                 "state_prep_index": rng.integers(n),
-                "hadamard_indices": [i for i in range(n) if i % 2],
-                "cnot_indices": [[i, i + 1] for i in range(n - 1)],
+                "ops": [hadamard_ops, cnot_ops],
+                # "hadamard_indices": [i for i in range(n) if i % 2],
+                # "cnot_indices": [[i, i + 1] for i in range(n - 1)],
```
lillian542:
```suggestion
                "ops": [*hadamard_ops, *cnot_ops],
```

lillian542: Actually maybe we just delete this part? I guess now that we only generate the fabricate subroutines when they are used, we could just drop this from the QEC test code, because it will only be used in cases where the code defines unitary_encode. We never actually use this in the tests.

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
```diff
@@ -1475,3 +1502,484 @@ def circuit():
...
+            //   CHECK-DAG:   [[q0:%.+]] = qecp.extract [[cb]][0] : !qecp.codeblock<1 x 9> -> !qecp.qubit<data>
+            //   CHECK-DAG:   [[q1:%.+]] = qecp.extract [[cb]][1] : !qecp.codeblock<1 x 9> -> !qecp.qubit<data>
+            //   CHECK-DAG:   [[q2:%.+]] = qecp.extract [[cb]][2] : !qecp.codeblock<1 x 9> -> !qecp.qubit<data>
+            //   CHECK-DAG:   [[q3:%.+]] = qecp.extract [[cb]][3] : !qecp.codeblock<1 x 9> -> !qecp.qubit<data>
+            //   CHECK-DAG:   [[q4:%.+]] = qecp.extract [[cb]][4] : !qecp.codeblock<1 x 9> -> !qecp.qubit<data>
+            //   CHECK-DAG:   [[q5:%.+]] = qecp.extract [[cb]][5] : !qecp.codeblock<1 x 9> -> !qecp.qubit<data>
+            //   CHECK-DAG:   [[q6:%.+]] = qecp.extract [[cb]][6] : !qecp.codeblock<1 x 9> -> !qecp.qubit<data>
+            //   CHECK-DAG:   [[q7:%.+]] = qecp.extract [[cb]][7] : !qecp.codeblock<1 x 9> -> !qecp.qubit<data>
+            //   CHECK-DAG:   [[q8:%.+]] = qecp.extract [[cb]][8] : !qecp.codeblock<1 x 9> -> !qecp.qubit<data>
+            // COM: State injection on the state_prep_index (qubit 6): H then T
```
lillian542:
```suggestion
            // COM: State injection on the state_prep_index (qubit 0): H then T
```
6 is Steane

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
joeycarter: Yes for sure! Seemed like inserting the diagonalizing gates was the best approach for the time being.

frontend/test/pytest/python_interface/transforms/qecp/test_qec_code_lib.py
joeycarter: Thanks! Works with me 👍

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
joeycarter: Good catch, thanks!

frontend/catalyst/python_interface/transforms/qecp/convert_qecl_to_qecp.py
joeycarter: You're absolutely right, good eye! Fixed: ece927acf

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
joeycarter: Makes sense, dropped these lines here: ece927acf

joeycarter (approved):

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
```diff
@@ -60,9 +57,9 @@ def _make_qec_code(
         n_aux: int = 3,
         x_tanner=None,
         z_tanner=None,
-        transversal_1q_gates=None,
-        transversal_2q_gates=None,
-        unitary_encoding=None,
+        transversal_1q_gates: dict[str, tuple[str, ...]] | None = None,
+        transversal_2q_gates: dict[str, str] | None = None,
+        unitary_encoding: dict = {},
```
jzaia18: We should aim to remove this CodeFactor warning, calling this multiple times might pollute `unitary_encoding`. The easiest way would be to do this:


```suggestion
        unitary_encoding: dict | None = None,
```

and then have 
```python
if unitary_encoding is None:
    unitary_encoding = {}
```
in first line of the function body

jzaia18 (approved):

frontend/catalyst/python_interface/transforms/qecp/qec_code_lib.py
```diff
@@ -15,62 +15,173 @@
...
+        # CNOT is transversal for all CSS codes
+        # There are no transversal Hadamard or S gates for this code
+        {
+            "x": ("Z", "I", "I", "Z", "I", "I", "Z", "I", "I"),
+            "y": ("Y", "X", "X", "Z", "I", "I", "Z", "I", "I"),
+            "z": ("X", "X", "X", "I", "I", "I", "I", "I", "I"),
+        },
+        {"cnot": "CNOT"},
+        #### Unitary encoding circuit ####
+        # jukebox generated this and verified it in simulation
```
sengthai: Jukebox seems useful here. XD
Alternatively, we can cite https://journals.aps.org/pra/abstract/10.1103/PhysRevA.52.R2493 or https://onlinelibrary.wiley.com/doi/10.1155/2021/1101911.

sengthai (approved):
Nice!

frontend/test/pytest/python_interface/transforms/qecp/test_xdsl_convert_qecl_to_qecp.py
joeycarter: You're absolutely right (I fixed one issue only to introduce another :see_no_evil:)

Fixed: 499cd708f

frontend/catalyst/python_interface/transforms/qecp/qec_code_lib.py
joeycarter: Ha good point! Citing https://onlinelibrary.wiley.com/doi/10.1155/2021/1101911 as the source in public code might be better than citing Jukebox :sweat_smile:. This paper shows the exact encoding circuit (and is open access, unlike Shor's Phys Rev A paper).

I'll let @lillian542 comment here since this was her work.

lillian542: Looks great, let's go with that! How did you go about finding it?

sengthai: I used Gemini to find that sources. 😅

lillian542: Hahahaha, okay, I'll have to try that in the future. I only used Jukebox (and a Google search which turned up nothing), and Jukebox insisted on figuring it out itself and then confirming in simulation. When I asked it for sources, it pulled up links about how to derive it 😅 Seems like in this case it "overthinks" compared to Gemini.

joeycarter: 8bb4aa383

PR #2930, opened 2026-06-09
[qref 4.2] Value-to-ref conversion for adjoint ops

paul0403:
**Context:**
Add adjoint ops to `--convert-to-reference-semantics`.

[sc-105534]

### DIFF
--- a/mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp
+++ b/mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp
@@ -423,7 +423,7 @@ void handleAdjoint(IRRewriter &builder, quantum::AdjointOp vAdjointOp, QubitValu
 void handleRegion(IRRewriter &builder, Region &r, QubitValueTracker &tracker)
 {
     assert(r.hasOneBlock() && "Expected region to have a single block");
-    assert(r.front().getTerminator() && "Expected region to have a terminator");
+    assert(r.front().getTerminator() && "Expected the region body to have a terminator");
 
     SmallVector<Operation *> erasureWorklist;
     r.walk<WalkOrder::PreOrder>([&](Operation *op) {
--- a/mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp
+++ b/mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp
@@ -14,6 +14,9 @@
 
 #define DEBUG_TYPE "reference-semantics-conversion"
 
+#include "reference_semantics_conversion.h"
+
+#include "llvm/ADT/DenseMap.h"
 #include "llvm/ADT/SetVector.h"
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/ADT/TypeSwitch.h"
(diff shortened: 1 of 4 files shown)


mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp
```diff
@@ -39,8 +42,11 @@ using namespace catalyst;
 
 namespace {
 
-void eraseAllVOps(SmallVector<Operation *> &erasureWorklist)
+void eraseAllVOps(Region &r, SmallVector<Operation *> &erasureWorklist)
 {
+    if (isa<quantum::YieldOp>(r.front().getTerminator())) {
+        erasureWorklist.push_back(r.front().getTerminator());
+    }
```
rniczh: Here we assume every region has a terminator, because you're calling this on an adjointop's region, so that condition should always hold. But if that isn't true later, a region may not necessarily have a terminator. Also, this can only handle a single block. I'm wondering whether we should add asserts to constrain how handleRegion is used?

rniczh (approved):
Thanks @paul0403 !

mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp
paul0403: So for all the regions we are handling, i.e. value semantics adjoint, for, if, while and switch, they are all single block and has a terminator. Maybe I assert these points in `handleRegion`?

paul0403: https://github.com/PennyLaneAI/catalyst/pull/2930/commits/955209f939f50e4a9dd5a3a205abb34e4ce65c01

kipawaa: I think some documentation here would also be helpful! I had the same thought reading this block, and without the context of how it's called, this appears like it has a bug. I think it would be best to move the asserts here (since the assumption is made by this function). Alternatively, maybe this function could accept a block rather than a region, or handle any number of blocks?

mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp
```diff
@@ -376,8 +387,44 @@ void handleGraphStatePrep(IRRewriter &builder, mbqc::GraphStatePrepOp vGraphStat
...
+    // Remove the moved over value semantics block args
+    rAdjointOp.getRegion().front().eraseArguments([](BlockArgument arg) {
+        return isa<quantum::QubitType, quantum::QuregType>(arg.getType());
+    });
+}
+
 void handleRegion(IRRewriter &builder, Region &r, QubitValueTracker &tracker)
 {
+    assert(r.hasOneBlock() && "Expected region to have a single block");
+    assert(r.front().getTerminator() && "Expected region to have a terminator");
```
kipawaa: Isn't this checking that the _block_ has a terminator?

kipawaa (commented):
Thanks @paul0403 ! Just a couple questions about the `eraseAllVOps` helper function, the core functionality is great!

mehrdad2m (approved):

mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2930/commits/71375370d72cb509c4462fb1fb512fa58cd51a6e

mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2930/commits/71efbfe0bbc7c20bf63001ddc922d2c0b7538e49

PR #2933, opened 2026-06-10
Fix adjoint-lowering for dynamic allocation

### LINKED ISSUE #2932: `adjoint-lowering` pass is unable to handle dynamic qubit allocation
Issue as described in the title. This can arise whenever the functional `qp.adjoint` is used either with dynamic allocation directly, or with passes that introduce dynamic allocation (e.g. `ppr-to-ppm`).
It's worth noting that this pass runs automatically in the default pipeline, after user passes have been applied:
```
"quantum-compilation-pipeline":
     ...
     "builtin.module(apply-transform-sequence)",  // <--- user transforms like ppr-to-ppm
     "inline-nested-module",
     "lower-mitigation",
     "adjoint-lowering",                          // <--- adjoint lowering
     ...
```
Therefor, it's quite feasible to run into this issue in the right circumstances:
```py
@qp.qjit
@qp.transforms.ppr_to_ppm
@qp.transforms.to_ppr
def circuit():
   qp.adjoint(lambda: qp.S(0))()
   return qp.probs()
```

In principle, adjoint-ing through a dynamic alloc/dealloc pair shouldn't pose an issue, by swapping the allocation with the deallocation, ensuring the qubit exists before executing the first (formerly last) gate.

dime10:
closes #2932 

[sc-122115]

### DIFF
--- a/mlir/lib/Quantum/Transforms/AdjointPatterns.cpp
+++ b/mlir/lib/Quantum/Transforms/AdjointPatterns.cpp
@@ -49,6 +49,24 @@ using namespace catalyst::quantum;
 
 namespace {
 
+/// Trace the value-semantic register use-def chain backwards to the `quantum.alloc` operation that
+/// originally produced it, following straight-line `quantum.insert` operations. Returns a null op
+/// if the allocation cannot be determined statically, i.e. the register crosses the block boundary.
+quantum::AllocOp findSourceAllocOp(Value qreg)
+{
+    while (Operation *defOp = qreg.getDefiningOp()) {
+        if (auto allocOp = dyn_cast<quantum::AllocOp>(defOp)) {
+            return allocOp;
+        }
+        if (auto insertOp = dyn_cast<quantum::InsertOp>(defOp)) {
+            qreg = insertOp.getInQreg();
+            continue;
+        }
+        break;
+    }
+    return quantum::AllocOp();
+}
+
 /// Clone the region of the adjoint operation `op` to the insertion point specified by the
 /// `builder`. Build and return the value mapping `mapping`.
 void cloneAdjointRegion(AdjointOp op, OpBuilder &builder, IRMapping &mapping,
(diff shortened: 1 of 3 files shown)


mlir/lib/Quantum/Transforms/AdjointPatterns.cpp
```diff
@@ -49,6 +49,24 @@ using namespace catalyst::quantum;
 
 namespace {
 
+/// Trace the value-semantic register use-def chain backwards to the `quantum.alloc` operation that
+/// originally produced it, following straight-line `quantum.insert` operations. Returns a null op
+/// if the allocation cannot be determined statically, i.e. the register crosses the block boundary.
+quantum::AllocOp findSourceAllocOp(Value qreg)
+{
```
paul0403, line 56: No action needed for now, but I'm starting to think the adjoint lowering pass would be so much easier and less error prone in reference semantics, since there's no need to build the reverse extract/insert chain, which is where most of the trouble is.

mlir/lib/Quantum/Transforms/AdjointPatterns.cpp
```diff
@@ -49,6 +49,24 @@ using namespace catalyst::quantum;
...
+        if (auto allocOp = dyn_cast<quantum::AllocOp>(defOp)) {
+            return allocOp;
+        }
+        if (auto insertOp = dyn_cast<quantum::InsertOp>(defOp)) {
+            qreg = insertOp.getInQreg();
+            continue;
+        }
+        break;
+    }
+    return quantum::AllocOp();
```
paul0403, line 67: I prefer std::optional, feels safer than empty constructors and explicit null checks down the line, but up to you. Feel free to resolve.

paul0403 (approved):
Beautiful! Don't forget the changelog!

mlir/lib/Quantum/Transforms/AdjointPatterns.cpp
dime10: This is explicitly intended in MLIR's design. Classes like ops, attributes, and others come with a null initialization that replaces the need for optional.

mlir/lib/Quantum/Transforms/AdjointPatterns.cpp
dime10: Maybe, but the problem remains that we'd need to support these memory instructions explicitly.

paul0403: Yeah, my comment was not on the alloc and dealloc ops, I just thought of it off the top of my head as I was reading the PR and seeing the walk back haha

mlir/lib/Quantum/Transforms/AdjointPatterns.cpp
```diff
@@ -122,6 +140,35 @@ class AdjointGenerator {
...
+                quantum::AllocOp sourceAlloc = findSourceAllocOp(deallocOp.getQreg());
+                if (!sourceAlloc) {
+                    deallocOp.emitError("Unable to reverse dynamic register deallocation in the "
+                                        "adjoint region: allocation size could not be determined");
+                    generationFailed = true;
+                    return;
+                }
+                Value nqubits = sourceAlloc.getNqubits();
+                if (nqubits) {
+                    nqubits = remappedValues.lookupOrDefault(nqubits);
```
mehrdad2m, line 157: Could we add a test for this line? I don't find any quantum.alloc(%x) which I assume should be the case triggering this line but codecov is not complaining, so not sure what's going on. 🤔

mehrdad2m (approved):

mlir/lib/Quantum/Transforms/AdjointPatterns.cpp
dime10: https://github.com/PennyLaneAI/catalyst/pull/2933/commits/7d7152c4c2cf6f5e34d6c0d414dbac13e544da12 (codecov is not setup for the mlir component)

PR #2938, opened 2026-06-11
Allow quantum.adjoint op's assembly format to have no operands and results

### LINKED ISSUE #2935: `quantum.adjoint` with no external qubits generates invalid custom IR syntax
With the recent changes in the adjoint operation, allowing individual qubits to be passed on a need-only basis, instead of the entire default register, we can generate programs that lead to invalid custom syntax when printed (that is they cannot be round-tripped):

```py
import pennylane as qp

def fun():
    with qp.allocate(1) as qs:
        qp.S(qs[0])

@qp.qjit(capture=True)
def circuit():
    qp.adjoint(fun)()
    return qp.probs()
```

This produces the following intermediate IR (after value semantic conversion, the reference semantic version seems fine), which cannot be round-tripped:
```
      quantum.adjoint() :  {
        %3 = quantum.alloc( 1) : !quantum.reg
        %4 = quantum.extract %3[ 0] : !quantum.reg -> !quantum.bit
        %out_qubits = quantum.custom "S"() %4 : !quantum.bit
        %5 = quantum.insert %3[ 0], %out_qubits : !quantum.reg, !quantum.bit
      }
```

```
circuit/1_QuantumCompilationStage/2_ValueSemanticsConversionPass.mlir:22:26: error: expected non-function type
      quantum.adjoint() :  {
                         ^
```

paul0403:
**Context:**
`quantum.adjoint` ops's operand and results are a variadic of qreg or qubit types. This means they can be empty.

There is a hardcoded `:` in `quantum.adjoint` op's assembly format. When the operands and results are empty, mlir parser confuses it with a bad function type.

Anchor the `:` on whether the operands/result exist. If they don't, don't print the `:`.

**Benefits:**
Adjoint op can now ahave empty operands/results.

**Related GitHub Issues:**
Closes #2935 

[sc-122135]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -162,6 +162,10 @@
   data dependencies through insert to extract chains instead of textual op ordering.
   [(#2884)](https://github.com/PennyLaneAI/catalyst/pull/2884)
 
+* The `quantum.adjoint` op can now take in an empty list of qubit and qreg operands and results.
+  Previously, this was impossible due to a bad custom assembly format.
+  [(#2938)](https://github.com/PennyLaneAI/catalyst/pull/2938)
+
 <h3>Internal changes ⚙️</h3>
 
 * The frontend now generates MLIR in reference semantics when capture is enabled.
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -162,8 +162,7 @@
   data dependencies through insert to extract chains instead of textual op ordering.
   [(#2884)](https://github.com/PennyLaneAI/catalyst/pull/2884)
 
-* The `quantum.adjoint` op can now take in an empty list of qubit and qreg operands and results.
-  Previously, this was impossible due to a bad custom assembly format.
+* Fixed the assembly format for `quantum.adjoint` when it has no quantum operands/results.
   [(#2938)](https://github.com/PennyLaneAI/catalyst/pull/2938)
 
 <h3>Internal changes ⚙️</h3>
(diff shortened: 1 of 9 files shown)


mlir/include/Quantum/IR/QuantumOps.td
```diff
@@ -765,7 +765,7 @@ def AdjointOp : Region_Op<"adjoint", [SingleBlockImplicitTerminator<"YieldOp">,
     );
 
     let assemblyFormat = [{
-        `(` $args `)` attr-dict `:` type($results) $region
```
mehrdad2m, line 768: wouldn't adding functional-type resolve the issue? why do we need to remove `:`?

paul0403: True :sweat_smile: let me add it back

paul0403: https://github.com/PennyLaneAI/catalyst/pull/2938/commits/c3196e3c4ac7a98a1ac399e4bf2248979643548c

mehrdad2m (approved):

mlir/include/Quantum/IR/QuantumOps.td
paul0403: Updated, now just anchoring the `:` on whether there are args (like how we do with the `ctrls` string for custom op on whether the control qubits exist)

doc/releases/changelog-dev.md
```diff
@@ -162,6 +162,10 @@
   data dependencies through insert to extract chains instead of textual op ordering.
   [(#2884)](https://github.com/PennyLaneAI/catalyst/pull/2884)
 
+* The `quantum.adjoint` op can now take in an empty list of qubit and qreg operands and results.
+  Previously, this was impossible due to a bad custom assembly format.
```
dime10: It already could (that's how I found the error because it can be generated)
```suggestion
* Fixed the assembly format for `quantum.adjoint` when it has no quantum operands/results.
```

dime10 (approved):

PR #2944, opened 2026-06-12
Sara/quantum phase folding

sarababaeii:
### Before submitting

**Context:**

Phase-folding optimization [[Amy et. al.](https://arxiv.org/abs/1303.2042)] for pure quantum programs.

It reduces T-count —and more generally the number of Rz gates— of Clifford+Rz circuits, using a `SymbolicCircuit` data structure consisting of two main properties: `PhasePolynomial` and `AffineTransformation`, which store the transformations of a circuit in the phase space and state space, respectively. The underlying data structure for both `PhasePolynomial` and `AffineTransformation` is the `Parity` class, which is essentially an efficient bitvector, representing a binary affine functional, i.e. parity function.

The algorithm has two main parts:
1. Phase analysis, which scans the circuit and updates the `SymbolicCircuit` by each operation, accordingly.
2. Phase merge, which iterates through `PhasePolynomial` terms to merge Rz gates conditional on the same `Parity`.

Implements [sc-119891], [sc-119857], and [sc-119862]

**Benefits:**

- Supports multiple register allocations throughout the program 
- Supports qubit (ancilla) state initialization to computational basis states (by static values)
- Features efficient —in both time and memory— bitvector operations for `Parity` class

**Possible Drawbacks:**
- Does not support dynamic values
- Does not support classical control flow

**Related GitHub Issues:**


main.py
sengthai, line 1: Nice to show how we use the pass. We normally show the usage code such that in test (e.g in `/catalyst/frontend/test/lit`). So, the user could know how to use it while test the end-to-end pass as well.

PhaseFoldingTest.mlir
sengthai, line 1: Great to have a test file! Could you add this folder in `mlir/test/Quantum`

mlir/lib/Quantum/Transforms/phase_folding.cpp
sengthai, line 1: could you add the copyright text on the top of each files as well?

sengthai (commented):
Thanks @sarababaeii for PR. Here is my initialize comments. Could you add tests in pytest as well. it could compare the circuit with and without phase folding. Doing this, would convince the correctness of the pass, and also usually check that test to see how it use as well.

PR #2947, opened 2026-06-15
Improve benchmark trigger mechanism

jzaia18:
**Context:**
[sc-103160] plans to introduce support for extra arguments to the `/benchmark` trigger. To support this, we will have to change the `/benchmark` triggers in PennyLane, Lightning, and Catalyst to match on the beginning of the comment rather than the full comment.

Changes the check on the trigger condition for `/benchmark`, renames `/benchmark` to `!benchmark`

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-122206]
https://github.com/PennyLaneAI/pennylane/pull/9676
https://github.com/PennyLaneAI/pennylane-lightning/pull/1392

### DIFF
--- a/.github/workflows/trigger_benchmarks.yml
+++ b/.github/workflows/trigger_benchmarks.yml
@@ -5,7 +5,7 @@ on:
       - created
 
 env:
-  BENCHMARK_TRIGGER_COMMAND: /benchmark
+  BENCHMARK_TRIGGER_COMMAND: "!benchmark"
 
 permissions:
   contents: read
@@ -23,11 +23,12 @@ jobs:
         id: comment_body
         env:
           COMMENT_BODY: ${{ github.event.comment.body }}
+          BENCHMARK_TRIGGER_COMMAND: ${{ env.BENCHMARK_TRIGGER_COMMAND }}
         run: |
-          if [ "$COMMENT_BODY" = "$BENCHMARK_TRIGGER_COMMAND" ]; then
-            echo "triggers_benchmark=true" >> "$GITHUB_OUTPUT"
+          if [[ "$COMMENT_BODY" =~ ^[[:space:]]*"$BENCHMARK_TRIGGER_COMMAND" ]]; then
+            echo "triggers_benchmark=true" >> $GITHUB_OUTPUT
           else
-            echo "triggers_benchmark=false" >> "$GITHUB_OUTPUT"
+            echo "triggers_benchmark=false" >> $GITHUB_OUTPUT
           fi
 
       - name: Set Repository Info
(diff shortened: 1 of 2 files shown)


.github/workflows/trigger_benchmarks.yml
```diff
@@ -23,11 +23,12 @@ jobs:
         id: comment_body
         env:
           COMMENT_BODY: ${{ github.event.comment.body }}
+          BENCHMARK_TRIGGER_COMMAND: ${{ env.BENCHMARK_TRIGGER_COMMAND }}
         run: |
-          if [ "$COMMENT_BODY" = "$BENCHMARK_TRIGGER_COMMAND" ]; then
-            echo "triggers_benchmark=true" >> "$GITHUB_OUTPUT"
+          if [[ "$COMMENT_BODY" =~ ^[[:space:]]*"$BENCHMARK_TRIGGER_COMMAND" ]]; then
```
mehrdad2m, line 28: why do we need the leading space: `^[[:space:]] `for? shouldn't that be after the `!benchmark`?

.github/workflows/trigger_benchmarks.yml
```diff
@@ -63,21 +64,22 @@ jobs:
...
+
           # The URL of the actual actions run that triggered the webhook
           WEBHOOK_METADATA_ACTIONS_RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
-          
+
           # The username of the person who commented the trigger command on the pull request
           WEBHOOK_METADATA_GITHUB_ACTOR: ${{ github.event.comment.user.login }}
-          WEBHOOK_METADATA_COMMENT_LINK: ${{ github.event.comment.html_url }}
-          
+          WEBHOOK_METADATA_COMMENT_LINK_API: ${{ github.event.comment.url }}
```
mehrdad2m, line 80: Sorry not very familiar with this. how is the `github.event.comment.url` different that `github.event.comment.html_url`

mehrdad2m (approved):
Thanks @jzaia18, Looks great, just a couple of questions:

.github/workflows/trigger_benchmarks.yml
jzaia18: The HTML link is a link to a webpage (like this one) whereas the ordinary URL is a raw JSON without HTML e.g. https://api.github.com/repos/PennyLaneAI/pennylane-lightning/issues/comments/4708987804

.github/workflows/trigger_benchmarks.yml
jzaia18: In case the user accidentally types leading spaces, so for example ` !benchmark` will still work

runora95 (approved):

PR #2948, opened 2026-06-15
[qref 4.6] Value-to-ref conversion for subroutines

paul0403:
Context:
Add calls and subroutines to `--convert-to-reference-semantics`.

[sc-105534]

### DIFF
--- a/mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp
+++ b/mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp
@@ -844,7 +844,7 @@ std::optional<SmallVector<Operation *>> handleRegion(IRRewriter &builder, Region
     }
 }
 
-bool isQuantumSubroutine(func::FuncOp f)
+bool funcOpHasValueSemanticsOps(func::FuncOp f)
 {
     // quantum.node is not a subroutine
     if (f->hasAttrOfType<UnitAttr>("quantum.node")) {
@@ -869,7 +869,7 @@ bool isQuantumSubroutine(func::FuncOp f)
             auto funcOp = mlir::SymbolTable::lookupNearestSymbolFrom<func::FuncOp>(
                 callOp, callOp.getCalleeAttr());
             assert(funcOp && "calling a non-existent subroutine");
-            if (isQuantumSubroutine(funcOp)) {
+            if (funcOpHasValueSemanticsOps(funcOp)) {
                 return WalkResult::interrupt();
             }
         }
@@ -881,7 +881,6 @@ bool isQuantumSubroutine(func::FuncOp f)
 void handleSubroutine(IRRewriter &builder, func::FuncOp f,
                       const SmallVector<IntegerAttr> &qregSizesAtCallsite)
 {
-
     MLIRContext *ctx = f.getContext();
     OpBuilder::InsertionGuard guard(builder);
     Location loc = f->getLoc();
@@ -1003,7 +1002,7 @@ struct ReferenceSemanticsConversionPass
                 return signalPassFailure();
             }
 
-            if (!isQuantumSubroutine(subroutine)) {
+            if (!funcOpHasValueSemanticsOps(subroutine)) {
                 continue;
             }
 
--- a/mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp
+++ b/mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp
@@ -1001,8 +1001,8 @@ struct ReferenceSemanticsConversionPass
                                 qregSizesAtCallsite.push_back(qrefQuregType.getSize());
                             }
                         }
+                        break;
                     }
-                    break;
                 }
             }
             QubitValueTracker tracker;
--- a/mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp
+++ b/mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp
@@ -12,12 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <llvm/ADT/STLExtras.h>
 #define DEBUG_TYPE "reference-semantics-conversion"
+
+#include "reference_semantics_conversion.h"
 
 #include <optional>
 
 #include "llvm/ADT/DenseMap.h"
+#include "llvm/ADT/STLExtras.h"
 #include "llvm/ADT/SetVector.h"
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/ADT/TypeSwitch.h"
@@ -35,8 +37,6 @@
 #include "Quantum/IR/QuantumInterfaces.h"
 #include "Quantum/IR/QuantumOps.h"
 #include "Quantum/IR/QuantumTypes.h"
-
-#include "reference_semantics_conversion.h"
 
 using namespace mlir;
 using namespace catalyst;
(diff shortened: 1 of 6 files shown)


mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp
```diff
@@ -787,6 +811,109 @@ std::optional<SmallVector<Operation *>> handleRegion(IRRewriter &builder, Region
...
+        return true;
+    }
+
+    // If we don't know from the args, must look at the body
+    if (f.isDeclaration()) {
+        return false;
+    }
+
+    WalkResult walkResult = f.walk([](Operation *op) {
+        if (isa<quantum::QuantumDialect>(op->getDialect())) {
```
kipawaa: Should we also consider other dialects, like `PBC`, `MBQC` etc?

mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp
```diff
@@ -819,6 +946,55 @@ struct ReferenceSemanticsConversionPass
...
+            if (!isa<func::FuncOp>((*scc->begin())->getCallableRegion()->getParentOp())) {
+                continue;
+            }
+
+            func::FuncOp subroutine =
+                cast<func::FuncOp>((*scc->begin())->getCallableRegion()->getParentOp());
+            if (scc.hasCycle()) {
+                subroutine.emitOpError("Quantum subroutine call graphs must not have cycles");
+                return signalPassFailure();
+            }
```
kipawaa, line 1024: Is this an appropriate place to catch this? This should definitely raise an error, I'm just not sure it should be in this pass

mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp
```diff
@@ -819,6 +946,55 @@ struct ReferenceSemanticsConversionPass
...
+                for (auto use : *uses) {
+                    Operation *user = use.getUser();
+                    if (auto callOp = dyn_cast<func::CallOp>(user)) {
+                        for (Type t : callOp.getOperandTypes()) {
+                            if (auto qrefQuregType = dyn_cast<qref::QuregType>(t)) {
+                                qregSizesAtCallsite.push_back(qrefQuregType.getSize());
+                            }
+                        }
+                    }
+                    break;
```
kipawaa: Is there any value iterating if we `break` on the first iteration?

mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp
```diff
@@ -819,6 +946,55 @@ struct ReferenceSemanticsConversionPass
...
+        for (func::FuncOp subroutine : llvm::reverse(subroutinesPostOrder)) {
+            SmallVector<IntegerAttr> qregSizesAtCallsite;
+            auto uses = SymbolTable::getSymbolUses(subroutine, mod);
+            if (uses) {
+                for (auto use : *uses) {
+                    Operation *user = use.getUser();
+                    if (auto callOp = dyn_cast<func::CallOp>(user)) {
+                        for (Type t : callOp.getOperandTypes()) {
+                            if (auto qrefQuregType = dyn_cast<qref::QuregType>(t)) {
+                                qregSizesAtCallsite.push_back(qrefQuregType.getSize());
```
kipawaa: Do we ever allow the same subroutine to be called with different sized registers? I would imagine that value semantics doesn't care about the size of the register, so if I call the same function once with a register of 3 qubits and use 2 of them, then another time with a register of 4 qubits and use 2 of them it would be the same. I'm not sure that this conversion can handle a case like this.

mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp
kipawaa, line 1: Not sure where best to leave this comment, but I've been considering what might happen if we manage to mix value and reference semantics. It seems to me that this would result in an invalid dataflow graph, and I think we should assert after applying these passes that no operations in the originating semantics exist in the IR after applying the conversion, just to be safe?

kipawaa (commented):
Nice work! Just a couple of questions, since this one seems a bit more complicated than the others.

mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp
paul0403: Yes the register size is an asymmetry between the two semantics. For now `qref.reg<?>` strictly means "known to be dynamic", instead of "could be static, but different static sizes". 

This is not ideal, but we were not focusing on this. Much of the reference semantcis effort was to get the frontend to work...

mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp
paul0403: oops the break should be inside the if :sweat_smile: good catch!

(The idea is all the call ops' signatures will be the same, so no need to look at the others)

paul0403: https://github.com/PennyLaneAI/catalyst/pull/2948/commits/24335a4b7aa4e1fb7b06aa64658d13365472df28

mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp
paul0403: True, but these dialects contain ops in both semantics. I'll add a list of value semantics ops

paul0403: https://github.com/PennyLaneAI/catalyst/pull/2948/commits/339fb40cdcef11a185252cb91ae997ac0af0665b

mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp
paul0403: I think it's fine here, the alternative would be in some sort of verifier, but running call graph analysis in the verifier is way too expensive imo

kipawaa: That makes sense, just wanted to check in about it :) Thanks!

mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp
kipawaa: For the purpose of this `isSubroutine` function, shouldn't we return true on _any_ quantum operation, regardless of semantics?

paul0403: It's to check whether a func op needs to be converted, so it needs to be from the source semantics

kipawaa: It might be best to rename this function to reflect that, something like `funcOpNeedsHandling` and we can specify that this is true when both (isSubroutine) and (contains source semantics).

mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/2948/commits/659c58f9b4e51adde6d538b26a5e634c8d1615e3

paul0403: and https://github.com/PennyLaneAI/catalyst/pull/2948/commits/92479ca629684ad32757b3ff9a7850616caa773d

mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp
paul0403: renamed to `funcOpHasValueSemanticsOps` https://github.com/PennyLaneAI/catalyst/pull/2948/commits/080a6cefcad5e76652160aa9fd1265949609f288

kipawaa (approved):
Thanks @paul0403! 🏆

mehrdad2m (approved):
🎉 🎉

PR #2952, opened 2026-06-16
Support qreg + scalar qubit-index signatures in decompose-lowering (#2869)

### LINKED ISSUE #2869: `decompose-lowering`'s `prepareCallOperands` doesn't support qreg + multiple individual qubit indices
Signatures of the form `(qreg, *params, first_qubit_index, second_qubit_index, ...)` are not supported by decompose lowering. 

MWE:
```python
@qjit(capture=True)
@qnode(device("lightning.qubit", wires=2))
def circuit(wires):
    qp.PauliRot(0.2, "XZ", wires)
    return qp.probs()


circuit([0, 1])
print(circuit.mlir)
```
generates the IR
```mlir
    func.func public @circuit(%arg0: tensor<i64>, %arg1: tensor<i64>) -> tensor<4xf64> attributes {diff_method = "parameter-shift", llvm.linkage = #llvm.linkage<internal>, quantum.node} {
      %cst = arith.constant 2.000000e-01 : f64
      quantum.device shots(%c0_i64) ["/Users/river.mccubbin/Github/catalyst/.venv/lib/python3.14/site-packages/pennylane_lightning/liblightning_qubit_catalyst.dylib", "LightningSimulator", "{'mcmc': False, 'num_burnin': 0, 'kernel_name': None}"]
      %extracted = tensor.extract %arg0[] : tensor<i64>
      %1 = quantum.extract %0[%extracted] : !quantum.reg -> !quantum.bit
      %extracted_0 = tensor.extract %arg1[] : tensor<i64>
      %2 = quantum.extract %0[%extracted_0] : !quantum.reg -> !quantum.bit
      %out_qubits:2 = quantum.paulirot ["X", "Z"](%cst) %1, %2 : !quantum.bit, !quantum.bit
      %extracted_1 = tensor.extract %arg0[] : tensor<i64>
      %3 = quantum.insert %0[%extracted_1], %out_qubits#0 : !quantum.reg, !quantum.bit
      %extracted_2 = tensor.extract %arg1[] : tensor<i64>
      %4 = quantum.insert %3[%extracted_2], %out_qubits#1 : !quantum.reg, !quantum.bit
      %5 = quantum.compbasis qreg %4 : !quantum.obs
      %6 = quantum.probs %5 : tensor<4xf64>
      quantum.device_release
      return %6 : tensor<4xf64>
    }
```
which fails with decompose-lowering:
`Assertion failed: (idx < size()), function operator[], file SmallVector.h, line 295.`
because of the assumption that qubit indices with qregs will be represented as a tensor (eg. `tensor<2xi64>`).

jk20342:
**Context:**

`decompose-lowering`'s `prepareCallOperands` assumed qreg-mode rules pack all
wire indices into a single tensor (e.g. `tensor<2xi64>`). Signatures of the form
`(qreg, *params, first_qubit_index, second_qubit_index, ...)` - produced by ops
like `quantum.paulirot` - hit an out-of-bounds `SmallVector` access and crashed.

`prepareCallOperands` now maps wire indices to function inputs one at a time,
letting each input type's element count decide how many indices it consumes
handles both a single packed `tensor<Nxi64>` arg and separate scalar `tensor<i64>`
args. Also removed leftover debug `.dump()` calls.

**Benefits:**
Fixes the crash; qreg-mode rules with per-qubit index arguments now lower correctly.

**Possible Drawbacks:**
None known.

**Related GitHub Issues:**
Fixes #2869

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -190,6 +190,10 @@
 * Fixed the assembly format for `quantum.adjoint` when it has no quantum operands/results.
   [(#2938)](https://github.com/PennyLaneAI/catalyst/pull/2938)
 
+* The `--decompose-lowering` pass no longer crashes on qreg-mode decomposition rules whose wire
+  indices are passed as separate scalar arguments, e.g. `(qreg, *params, idx0, idx1, ...)`.
+  [(#2952)](https://github.com/PennyLaneAI/catalyst/pull/2952)
+
 <h3>Internal changes ⚙️</h3>
 
 * The `/benchmark` GitHub comment trigger can now accept additional arguments and has been renamed to `!benchmark`.
--- a/mlir/lib/Quantum/Transforms/DecomposeLoweringImpl.hpp
+++ b/mlir/lib/Quantum/Transforms/DecomposeLoweringImpl.hpp
@@ -220,11 +220,26 @@ class BaseSignatureAnalyzer {
         }
 
         if (hasQreg) {
+            // The wire indices may be packed into a single tensor argument (e.g.
+            // tensor<2xi64>) or spread across several scalar arguments, as in
+            // signatures of the form (qreg, *params, first_qubit_index,
+            // second_qubit_index, ...). Consume the function inputs one at a time,
+            // letting the number of elements each input type expects decide how many
+            // wire indices it absorbs.
             for (const auto &indices : {signature.inWireIndices, signature.inCtrlWireIndices}) {
-                if (!indices.empty()) {
+                size_t consumed = 0;
+                while (consumed < indices.size()) {
+                    assert(operandIdx < static_cast<int>(funcInputsNoQreg.size()) &&
+                           "ran out of function inputs while mapping wire indices");
+                    Type funcInputType = funcInputsNoQreg[operandIdx];
+                    size_t numToConsume =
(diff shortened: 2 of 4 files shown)


jk20342:
one thing I didnt change since it was already a pre existing contract is the fact that assert is loading bearing under NDEBUG but could add an emitError + failure() to the PauliRot//Custom paths since MultiRZ already does it

dime10:
Hi @jk20342, welcome to the Catalyst project 👋 And thank you for opening a PR!

Our team will have a look the submitted code, however one caveat is that this component (compiler decompositions) is heavily in flux right now, so we may or may not be able to merge your contribution. Please allow us some time to look into it. In the future, it would be safer to comment on the issue you are intending to tackle before work starts, that way we coordinate or provide missing context that may be necessary. 

Cheers :)

jk20342:
> Hi @jk20342, welcome to the Catalyst project 👋 And thank you for opening a PR!
> 
> Our team will have a look the submitted code, however one caveat is that this component (compiler decompositions) is heavily in flux right now, so we may or may not be able to merge your contribution. Please allow us some time to look into it. In the future, it would be safer to comment on the issue you are intending to tackle before work starts, that way we coordinate or provide missing context that may be necessary.
> 
> Cheers :)

thanks :) i normally contribute to pennylane rather then catalyst but this project seems quite interesting thanks for the info for future PRs :D

PR #2953, opened 2026-06-16
allow using AbstractArray for AOT compilation and specs

albi3ro:
**Context:**

I was messing around with AbstractArray, AbstractWires and resource estimation.  I'd like to be able to use AbstractArray and AbstractWires to trigger pass-by-pass specs.  I'd also like to use them to trigger AOT compilation via type-hints.

Various patches to treat AbstractArray and AbstractWires like abstract types.

**Benefits:**

Can now do:
```
@qp.qjit(capture=True)
@qp.qnode(qp.device('lightning.qubit', wires=4))
def c(x : qp.typing.AbstractArray((3,), float), wires : qp.wires.Wires[4]):
    @qp.for_loop(x.shape[0])
    def loop(i):
        qp.RX(x[i], wires[i])

    @qp.for_loop(wires.shape[0])
    def loop2(i):
        qp.X(i)

    loop()
    loop2()
    return qp.expval(qp.Z(0))

print(c.mlir)
```
```
@qp.qjit(capture=True)
@qp.qnode(qp.device('lightning.qubit', wires=4))
def c(x, wires):
    @qp.for_loop(x.shape[0])
    def loop(i):
        qp.RX(x[i], wires[i])

    @qp.for_loop(wires.shape[0])
    def loop2(i):
        qp.X(i)

    loop()
    loop2()
    return qp.expval(qp.Z(0))

s = qp.specs(c, level=0)(qp.typing.AbstractArray((3,), float), qp.wires.Wires[3])
```

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-122582]


doc/releases/changelog-dev.md
```diff
@@ -14,6 +14,9 @@
 
 <h3>Improvements 🛠</h3>
 
+* Adds the ability to use `pennylane.typing.AbstractArray` and `pennylane.wires.AbstractWires` as type hints for
+  AOT compilation and as arguments to `pennylane.specs` calculations.
```
albi3ro, line 25:
```suggestion
  AOT compilation and as arguments to `pennylane.specs` calculations.
  [(#2953)](https://github.com/PennyLaneAI/catalyst/pull/2953)
```

dime10:
Sounds like you might need to bump the PL dependency?

frontend/catalyst/jax_extras/patches.py
```diff
@@ -99,6 +101,8 @@ def _drop_unused_vars2(
 def get_aval2(x):
     """An extended version of `jax.core.get_aval` which also accepts AbstractValues."""
     # TODO: remove this patch when https://github.com/google/jax/pull/18579 is merged
```
comp-phys-marc, line 102: Are we still planning to remove this? Should we update this comment if we now need this in order to support `AbstractWires`, `AbstractArray`?

frontend/test/pytest/test_jit_behaviour.py
```diff
@@ -863,6 +863,28 @@ def workflow(phi: float):  # pylint: disable=function-redefined
...
+
+            @qp.for_loop(wires.shape[0])
+            def loop2(i):
+                qp.X(i)
+
+            loop()
+            loop2()
+            return qp.expval(qp.Z(0))
+
+        assert c.mlir
```
comp-phys-marc, line 886: What is this actually testing? Why would `c.mlir` be truthy or falsey?

frontend/catalyst/jax_extras/patches.py
albi3ro: @dime10 ?

frontend/test/pytest/test_jit_behaviour.py
albi3ro: This is testing that the mlir exists, simliar to the other tests in this class.

frontend/catalyst/jax_extras/patches.py
dime10: Getting things merged into jax proved quite difficult, hence our investigation into our own simplified fork of it.

frontend/test/pytest/test_jit_behaviour.py
comp-phys-marc: In what situation would the mlir not exist and how does that relate to whether AbstractArray and AbstractWires can be used to specify the input shapes for AOT compilation

frontend/catalyst/jax_extras/patches.py
comp-phys-marc: also it seems not to make sense anymore to merge this into jax since it depends on our internal classes

frontend/test/pytest/test_jit_behaviour.py
albi3ro: The mlir does not exist when AOT compilation is not triggered by the type hints.

frontend/catalyst/jax_extras/patches.py
dime10: Hmm that's probably not great. Is there another point we could hook into (in our own libraries), to achieve what you're doing?

dime10: One option I suppose would be to register our types as abstract jax values. Another might be to do this conversion further up in the call stack where we canonicalize the signature to check if AOT compilation is possible?

frontend/test/pytest/test_jit_behaviour.py
comp-phys-marc: Thanks, maybe a dev comment to that effect would make this test easier to understand

frontend/catalyst/jax_extras/patches.py
albi3ro: Jax itself doesn't care about the type as long as it has `shape` and `dtype` attributes. For example, this works fine:
```
def f(x):
    return 2 * x

jax.make_jaxpr(f)(qp.typing.AbstractArray((3,), float))
```
It's only this patch that actually cares.

frontend/test/pytest/test_jit_behaviour.py
albi3ro:
```suggestion

        # see qjit docs for more information
        assert c.mlir
```
?

comp-phys-marc: Honestly in my opinion that is not a very helpful comment because it doesn't tell me what I am looking for in the docs I am afraid. A very helpful comment in my opinion would be:

`# the mlir does not exist when AOT compilation is not triggered by the type hints`

frontend/catalyst/jax_extras/patches.py
dime10: Why does the patch care?

dime10: Though my point was more that if our types are abstract jax types as well this branch would catch it:
```
    if isinstance(x, AbstractValue):
        return x
```

dime10: To be fair it sounds like we shouldn't even be doing this (again mixing values with types): https://github.com/jax-ml/jax/pull/18561#issuecomment-1815540416

frontend/test/pytest/test_jit_behaviour.py
albi3ro: Given that this pattern occurs about a hundred times throughout this file and the qjit docstring is insufficient, I wonder if there is a better location to write more comprehensive onboarding information first. I wonder if we should fork off another task to improve the qjit docstring with the required explanations. @isaacdevlugt ?

comp-phys-marc: I suppose your point is that someone has to go through an onboarding process to do with working with Catalyst before they can understand the tests. In my opinion that is not ideal.

comp-phys-marc: Ideally we want our code to be self documenting, meaning that it is understandable to a reader without a lot of extra context.

comp-phys-marc: After reading the qjit docstring in this case I also unfortunately don't feel it gives me the helpful bit of information either. :/

comp-phys-marc: On the other hand, I understood the test well enough as soon as you were able to tell me that "The mlir does not exist when AOT compilation is not triggered by the type hints".

We should try to avoid the classic "hit by a bus" effect where information that is held in an expert's head is not written down and therefore the company stops working efficiently if that person is no longer available to ask questions to.

comp-phys-marc: I want to say though that you are very helpful! And it is great that you are so knowledgable about how things work and can answer questions like this. :)

comp-phys-marc: That said yeah let's just improve the qjit docstring! We can fork into a separate task if you like.

frontend/catalyst/jax_extras/patches.py
dime10: To be honest I don't have the full context for these patches anymore, and it would take a while to untangle. We do use exactly the pattern in Christina's code example, and it eventually bit us (probably in the context of dynamic shapes, which is what most of these patches were for). 

So while this is likely not a robust approach, and might further deepen the technical debt here, I'm also fine to say screw it, we can get back to this when we clean up this patch mess 😅

comp-phys-marc: I see :P

comp-phys-marc: I would ask whoever has the context on these patches to add some docs in the meantime if it were me :)

dime10: Unfortunately they are no longer here 😅

comp-phys-marc: This is exactly the hit by bus effect I was talking about 😭 😭

frontend/test/pytest/test_jit_behaviour.py
comp-phys-marc: Per the [slack](https://xanaduhq.slack.com/archives/C030UB33QH3/p1782155125265659) thread, @kipawaa is going to add a type hint and some documentation to the `mlir` property that says it returns `None` if compilation has not occurred. All good then!

frontend/catalyst/jax_extras/patches.py
comp-phys-marc: This is really not ideal but at the same time not something we can fully address in this PR. So all I ask is that there is a plan of attack for managing it at some point. 🙏

comp-phys-marc (approved):
Thanks for your patience as I learn the ropes.

frontend/catalyst/jax_extras/patches.py
```diff
@@ -46,6 +46,8 @@
 from jax._src.pjit import _out_type, _pjit_forwarding, jit_p
 from jax._src.sharding_impls import UnspecifiedValue
 from jax.core import AbstractValue, Tracer
+from pennylane.typing import AbstractArray
+from pennylane.wires import AbstractWires
```
albi3ro:
```suggestion
from pennylane.typing import AbstractArray, AbstractWires
```

frontend/catalyst/tracing/type_signatures.py
```diff
@@ -27,6 +27,8 @@
 from jax._src.pjit import _flat_axes_specs
 from jax.core import AbstractValue
 from jax.tree_util import tree_flatten, tree_unflatten
+from pennylane.typing import AbstractArray
+from pennylane.wires import AbstractWires
```
albi3ro:
```suggestion
from pennylane.typing import AbstractArray, AbstractWires
```

frontend/test/pytest/test_jit_behaviour.py
```diff
@@ -863,6 +863,28 @@ def workflow(phi: float):  # pylint: disable=function-redefined
...
 
 class TestDefaultAvailableIR:
+
+    def test_AbstractArray_AbstractWires_AOT(self):
+        """Test that AbstractArray and AbstractWires can be used to specify the input
+        shapes for AOT compilation."""
+
+        @qp.qjit(capture=True)
+        @qp.qnode(qp.device("lightning.qubit", wires=4))
+        def c(x: qp.typing.AbstractArray((3,), float), wires: qp.wires.Wires[4]):
```
albi3ro:
```suggestion
        def c(x: qp.typing.AbstractArray((3,), float), wires: qp.typing.Wire[4]):
```

frontend/test/pytest/test_specs.py
```diff
@@ -1474,5 +1474,28 @@ def test_marker(self, simple_circuit, capture_mode):
...
+
+        @qp.for_loop(wires.shape[0])
+        def loop2(i):
+            qp.X(i)
+
+        loop()
+        loop2()
+        return qp.expval(qp.Z(0))
+
+    s = qp.specs(c, level=0)(qp.typing.AbstractArray((3,), float), qp.wires.Wires[3])
```
albi3ro:
```suggestion
    s = qp.specs(c, level=0)(qp.typing.AbstractArray((3,), float), qp.typing.Wire[3])
```

frontend/test/pytest/test_specs.py
```diff
@@ -1474,5 +1474,28 @@ def test_marker(self, simple_circuit, capture_mode):
         check_specs_same(actual, expected)
 
 
+def test_abstract_array_inputs():
+    """Test that AbstractArray and AbstractWires can be used with specs when level!= device."""
```
jzaia18, line 1478:
```suggestion
    """Test that AbstractArray and AbstractWires can be used with pass-by-pass specs."""
```

More descriptive since we're planning to rework how these levels are used to dispatch

jzaia18 (commented):
Looks good, happy to approve once all tests are green

jzaia18 (approved):

Issue #755, opened 2024-05-20
Add local folding to ZNE error mitigation


josh146:
**Requirements**

The `mitigate_with_zne` function is modified to take a `folding` argument:

```python
mitigate_with_zne(f, *, scale_factors: jax.Array, folding: str, deg)
```

where `folding` accepts a string corresponding to:

- `global` (current behaviour, global folding)

- `random` (local random folding)

 - `left` (local folding from left)

 - `right` (local folding from right)

- `all` (all gates folded locally)

For more details on how these options behave, see the [mitiq documentation](https://mitiq.readthedocs.io/en/stable/guide/zne-3-options.html#unitary-folding).

The `mitigate_with_zne` docstring should be adequately updated.

**Beyond the MVP**

We may want `folding` to accept a callable, so that users can provide their own folding transform, akin to the [PennyLane version of the transform](https://docs.pennylane.ai/en/stable/code/api/pennylane.transforms.mitigate_with_zne.html#pennylane.transforms.mitigate_with_zne)

vprusso:
Hi @josh146, 

Thanks for putting this together. One point that may be relevant here is that `fold_from_left` and `fold_from_right` are deprecated within `mitiq`, so we may want to consider those options as out-of-scope. 

As for supporting local folding, the following snippet may be helpful as this applies local folding without the need to invoke `mitiq`: 

```py
from functools import reduce

from qiskit import QuantumCircuit
from qiskit.converters import circuit_to_dag, dag_to_circuit

# Define a basic Qiskit quantum circuit, for example.
qc = QuantumCircuit(3)
qc.h(0)
qc.cx(0, 1)
qc.cx(0, 2)

# Altering impacts the number of folds to perform.
num_folds = 0

# Deconstruct the circuit into layers as represented by DAG.
dag = circuit_to_dag(qc)
circuit_layers = [dag_to_circuit(layer["graph"]) for layer in dag.layers()]

# Construct a new circuit based on the number of folds requested.
layers: list[QuantumCircuit] = []
for i, layer in enumerate(circuit_layers):
    layers.append(layer)
    for _ in range(num_folds):
        layers.append(layer.inverse())
        layers.append(layer)

# Combine the layers into one circuit.
combined_circuit = reduce(lambda x, y: x.compose(y), layers)
print(combined_circuit)
```

So, for instance, if `num_folds=0`, then this is just the original circuit: 

```
     ┌───┐
q_0: ┤ H ├──■────■──
[...]

PR #2956, opened 2026-06-17
Implement ZNE local-random unitary folding

jk20342:
**Context:**
The `local-random` folding option for `mitigate_with_zne` was wired through the enum and op but the MLIR `randomLocalFolding` was a stub

Implemented `randomLocalFolding` in the `--lower-mitigation` pass: each gate is wrapped in a fold loop where every iteration flips a runtime coin (`__catalyst__rt__random_double() < 0.5`) deciding whether to insert a `G G^\dagger` pair. Added the `__catalyst__rt__random_double` runtime CAPI (backed by the ExecutionContext PRNG) and unblocked `folding="local-random"` in the frontend.

**Benefits:**
`local-random` folding now works end-to-end, with per-gate fold counts randomized at runtime and reproducible under `qjit(seed=...)`.

**Possible Drawbacks:**
Folding shares the single qjit `seed`, so it can't be seeded independently of measurement readout yet.

**Related GitHub Issues:**
#755


dime10:
Thank you for the PR @jk20342 🎉 

Before reviewing code, my main concern was figuring out what the feature is even _supposed_ to do.

> each gate is wrapped in a fold loop where every iteration flips a runtime coin (__catalyst__rt__random_double() < 0.5) deciding whether to insert a G G^\dagger pair

It sounds like in your PR each gate is wrapped in a loop that iterates `(scaling_factor-1)/2` number of times, and each iteration inserts an additional `GG†` pair with 50% probability. That implies that the total size of the circuit is roughly scaled by `1 + num_iterations*num_gates_inserted*insertion_probability` or `1 + (scaling_factor-1)/2 * 2 * 50%` which equals `(1+scale_factor)/2`.

---

The issue is pretty sparse about the desired behaviour, and mostly refers to the [UF documentation](https://mitiq.readthedocs.io/en/stable/guide/zne-3-options.html). Reading through it I'm getting the following relevant info:

> For any noise scaling function, if scale_factor is equal to 1, the input circuit is unchanged and it is subject to the base noise of the backend.

This is satisfied.

> When scale_factor is an odd integer, the number of gates is scaled exactly as dictated by the value of scale_factor.

Looking at the [odd integer scale factors section](https://mitiq.readthedocs.io/en/stable/guide/zne-3-options.html#the-special-case-of-odd-integer-scale-factors), this is not 100% clear to me, but it sounds like odd scaling factors might lead to a predictably uniformly scaled circuit? The example in that section, which also uses local-random folding, does not show any randomness either (admittedly it is very small).

Unfortunately, the [API docs](https://mitiq.readthedocs.io/en/stable/apidoc.html#mitiq.zne.scaling.folding.fold_gates_at_random) for the random folding function don't go into details on how the sampling is performed exactly, except for:

> Returns a new folded circuit by applying the map G -> G G^dag G to a subset of gates of the input circuit, different indices randomly sampled without replacement.

So gate indices (from the circuit) are sampled randomly (probability distribution not specified, although there is an optional fidelity mechanism that might affect this) without replacement.

I think the "index sampling without replacement" part could support the hypothesis that random folding with a scale factor of 3 would indeed lead to a uniformly scaled circuit.

> The folded circuit has a number of gates approximately equal to scale_factor * n where n is the number of gates in the input circuit.

What is clearer I think is that the proposed implementation doesn't meet this scaling factor target, since it generates `(1+scale_factor)/2 * n` many gates, rather than `scale_factor * n`. 

--- 

One interpretation is that the random folding for odd integer scale factors does not behave any differently than the "local-all" folding, and that the randomness is only meaningful for fractional scale_factors (currently not supported in the Catalyst implementation), or perhaps when the fidelity dict is used.

What are your thoughts on this? Perhaps the best thing to do would be to reach out to mitiq maintainers, dig into their source code, or play around with their tools to figure out the intended behaviour.

jk20342:
Hey @dime10 took a look at the mitiq docs https://mitiq.readthedocs.io/en/stable/guide/zne-3-options.html and https://mitiq.readthedocs.io/en/stable/_modules/mitiq/zne/scaling/folding.html from my understanding the rule is if the scale factor is an odd integer `1+2n` then all gates are folded `n` times -> `G - G (G G^\dagger)^((sf-1)/2)`: and a generic real `\lambda = 1 + 2(n + \\delta)`, `\\delta < 1` all gates folded n time and a subset is folded one more time. So the randomness only enters from the fractional remainder \\delta

going to make some changes to allow real `sf \ge 1` then pass `base` + `\\delta` to the `mitigation.zne` op and implement unconditional `base` folds + one `random_double() < \\delta` fold

that looks to be correct behavior to me if you could verify on your end :)

jk20342:
ie per gate: 
fold base = floor((sf-1)/2) times unconditionally then
fold one extra time with probability \\delta = (sf-1)/2 - base,
then the original gate.

jk20342:
I would leave the fidelity-weighted variant out though since catalyst doesnt expose per-gate fidelities yet

dime10:
Great, sounds like we have the same interpretation then :) Your proposal sounds reasonable!

jk20342:
@dime10 made the changes :)

jk20342:
hey @dime10 just repinging incase got lost in notifications :)

dime10:
Hi @jk20342, I apologize for the lack of response here, I'll review now!

frontend/catalyst/api_extensions/error_mitigation.py
```diff
@@ -75,6 +103,14 @@ def mitigate_with_zne(
         folding (str): Unitary folding technique to be used to scale the circuit. Possible values:
             - global: the global unitary of the input circuit is folded
             - local-all: per-gate folding sequences replace original gates in-place in the circuit
+            - local-random: reproduces Mitiq's ``fold_gates_at_random``. Every gate is folded
+              ``base = floor((scale_factor-1)/2)`` times, and then exactly
+              ``k = round(((scale_factor-1)/2 - base) * n)`` of the ``n`` gates are folded once
+              more, chosen uniformly at random *without replacement* (using the runtime PRNG,
```
dime10, line 109: I was thinking about this choice we have:
- make the random choice at runtime, like you did
- make the random choice at compile time

What are your thoughts on this choice?

Since this is a compilation technique, one might think this is a compile-time choice. On the other hand I realized that we can't guarantee the right scaling factor when making the random choice at compile because we don't control each gate in the program, only each gate _instruction_ (which may produce one or none or multiple gates).

frontend/test/pytest/test_mitigation.py
```diff
@@ -264,24 +264,90 @@ def mitigated_qnode():
...
+        qp.Hadamard(wires=0)
+        qp.RZ(0.3, wires=0)
+        qp.CNOT(wires=[1, 0])
+        qp.Hadamard(wires=1)
+        return qp.expval(qp.PauliY(wires=0))
+
+    @catalyst.qjit(seed=42)
+    def mitigated_qnode():
+        return catalyst.mitigate_with_zne(
+            circuit, scale_factors=[1.0, 2.0, 3.0], folding="local-random"
```
dime10: While these are technically floats they're not fractional, we should definitely have a test with fractional scaling factors.

mlir/include/Mitigation/IR/MitigationOps.td
```diff
@@ -48,7 +48,12 @@ def ZneOp : Mitigation_Op<"zne", [DeclareOpInterfaceMethods<CallOpInterface>,
         SymbolRefAttr:$callee,
         Variadic<AnyType>:$args,
         FoldingAttr:$folding,
-        RankedTensorOf<[AnySignlessIntegerOrIndex]>:$numFolds,
+        // Integer entries hold an exact per-circuit fold count `(scale_factor - 1) / 2`
+        // (used by `global`/`all`). Floating-point entries carry the same quantity but
+        // keep its fractional part so `random` folding can support non-integer
+        // scale factors (base folds + one extra fold sampled with probability = frac).
```
dime10: I think this comment could be moved into the description field.

mlir/include/Mitigation/IR/MitigationOps.td
```diff
@@ -48,7 +48,12 @@ def ZneOp : Mitigation_Op<"zne", [DeclareOpInterfaceMethods<CallOpInterface>,
...
+        AnyTypeOf<[RankedTensorOf<[AnySignlessIntegerOrIndex]>,
```
dime10: An alternative we could think about to simplify this parameter: always use floats but reject non-integral values for some of the methods, rather than dealing with two possible data types.

mlir/lib/Mitigation/Transforms/MitigationMethods/Zne.hpp
```diff
@@ -32,6 +32,15 @@ struct ZneLowering : public OpRewritePattern<mitigation::ZneOp> {
     LogicalResult matchAndRewrite(mitigation::ZneOp op, PatternRewriter &rewriter) const override;
 
   private:
+    static func::FuncOp getOrCreateFoldedCallee(Location loc, PatternRewriter &rewriter,
```
dime10, line 40: I know many existing modules here are missing this, but could you add docstrings to the new methods you added?

mlir/test/Mitigation/ZneFoldingRandomTest.mlir
dime10, line 1: This FileCheck test produces what looks like (most of) the IR for the function. This makes it difficult to read and maintain. If at all possible, it's better to perform surgical checks for just the relevant functionality, use multiple tests for different aspects of the functionality, and/or group check commands together with the source instructions they are testing.

dime10 (commented):
Thanks @jk20342, this looks fantastic 🎉 Last round of comments from me and I'm happy to approve :)

frontend/catalyst/api_extensions/error_mitigation.py
jk20342: I think runtime makes more sense since the semantics are defined over the executed gate sequence ie with a control flow an instruction can produce zero one or many gates so at compile time we can neither count n nor guarantee exactly k folds so an instruction picked inside a loop would get folded every iteration, one under a false branch never only the runtime selection would guarantee the scale factor for arbitrary programs

jk20342:
@dime10 there we go made the changes :)

mlir/lib/Mitigation/Transforms/MitigationMethods/Zne.hpp
jk20342: done :)

mlir/include/Mitigation/IR/MitigationOps.td
jk20342: updated this

dime10 (approved):
Nice work 🐐

paul0403 (approved):

runtime/lib/capi/ExecutionContext.hpp
```diff
@@ -294,6 +297,13 @@ class ExecutionContext final {
 
         if (this->seed != nullptr) {
             this->gen = std::mt19937(*seed);
+            // Derive a decoupled (but reproducible) stream for folding.
+            std::seed_seq foldSeq{static_cast<uint32_t>(*seed), static_cast<uint32_t>(0x9e3779b9U)};
```
paul0403, line 301: Nice! Just wondering where this magic number comes from. Is it a recommended magic number for seeding random sequences?

dime10: great question!

jk20342: hey @paul0403 and @dime10 its actually quite interesting its the 32 bit representation of the golden ratio which is the "conventional" bit mixing salt technically any "random looking" constant would achieve this same decorrelation since `seed_seq` does its own mixing but this magic number is considered standard, one fun fact its the salt that is used in boosts `hash_combine` :>

dime10: neat :D

jk20342:
added tests so codecov isnt failing

dime10:
Looks like you need to run clang-format first (we're setting up a pre-commit hook for this right now so these failures don't happen anymore). After that, I'll merge your PR as soon as the CI is green. No need to merge with main anymore (unless there are conflicts).

jk20342:
@dime10 there we go

PR #2961, opened 2026-06-17
Fix resource analysis pass to include functions in nested modules

sengthai:
Currently, `ResourceAnalysis` only looked at top-level func.func definitions, so nested kernels behind catalyst.launch_kernel were missing from the JSON output. It now walks nested modules recursively, so those functions are included in resource analysis results.


rniczh (approved):
💯

lazypanda10117 (approved):

Issue #2726, opened 2026-04-21
Canonicalization of inverse extract/insert pairs on the same index is obfuscated if there is a chain of inserts


paul0403:
Consider the circuit `CNOT(0, 1) - X(0)`

```mlir
  %0 = quantum.alloc( 2) : !quantum.reg

  %1 = quantum.extract %0[ 0] : !quantum.reg -> !quantum.bit
  %2 = quantum.extract %0[ 1] : !quantum.reg -> !quantum.bit
  %out_qubits:2 = quantum.custom "CNOT"() %1, %2 : !quantum.bit, !quantum.bit
  %3 = quantum.insert %0[ 0], %out_qubits#0 : !quantum.reg, !quantum.bit
  %4 = quantum.insert %3[ 1], %out_qubits#1 : !quantum.reg, !quantum.bit

  %5 = quantum.extract %4[ 0] : !quantum.reg -> !quantum.bit
  %out_qubits_0 = quantum.custom "PauliX"() %5 : !quantum.bit
  %6 = quantum.insert %4[ 0], %out_qubits_0 : !quantum.reg, !quantum.bit

  quantum.dealloc %6 : !quantum.reg
```

After `--canonicalize`, the IR stays unchanged. 

**Proposed improvement:**
By interchanging the order of the two inserts after the CNOT: 
```mlir
  %0 = quantum.alloc( 2) : !quantum.reg

  %1 = quantum.extract %0[ 0] : !quantum.reg -> !quantum.bit
  %2 = quantum.extract %0[ 1] : !quantum.reg -> !quantum.bit
  %out_qubits:2 = quantum.custom "CNOT"() %1, %2 : !quantum.bit, !quantum.bit
  %4 = quantum.insert %0[ 1], %out_qubits#1 : !quantum.reg, !quantum.bit
  %3 = quantum.insert %4[ 0], %out_qubits#0 : !quantum.reg, !quantum.bit

  %5 = quantum.extract %3[ 0] : !quantum.reg -> !quantum.bit
  %out_qubits_0 = quantum.custom "PauliX"() %5 : !quantum.bit
  %6 = quantum.insert %4[ 0], %out_qubits_0 : !quantum.reg, !quantum.bit

  quantum.dealloc %6 : !quantum.reg
```

we should be able to detect the inverse pair on index `0`:
```mlir
  %0 = quantum.alloc( 2) : !quantum.reg
[...]

PR #2893, opened 2026-05-28
Add PauliRot support to `decompose-lowering`

kipawaa:
**Context:**
Supporting `quantum.paulirot` in the `decompose-lowering` pass is essential for python-callbacks and `graph-decomposition` improvements.

**Description of the Change:**
Adds patterns for `quantum.paulirot` to `decompose-lowering`

[sc-121358]

PR #2965, opened 2026-06-18
Canonicalize extract past distinct-index insert

### LINKED ISSUE #2725: Canonicalization does not recognize extracts from different indices on the same qreg
Consider the circuit `X(0)-Z(1)`:

```mlir

  %out_qubits = quantum.custom "PauliX"() %1 : !quantum.bit
  %2 = quantum.insert %0[ 0], %out_qubits : !quantum.reg, !quantum.bit

  %3 = quantum.extract %2[ 1] : !quantum.reg -> !quantum.bit
  %out_qubits_0 = quantum.custom "PauliZ"() %3 : !quantum.bit
  %4 = quantum.insert %2[ 1], %out_qubits_0 : !quantum.reg, !quantum.bit

```

After `--canonicalize`, the IR stays unchanged.

**Proposed improvement:**
Since the 0th and 1st qubits are from the same register, it should be possible to make them extract from the same SSA qreg value:
```mlir

  %out_qubits = quantum.custom "PauliX"() %1 : !quantum.bit

  %3 = quantum.extract %0[ 1] : !quantum.reg -> !quantum.bit
  %out_qubits_0 = quantum.custom "PauliZ"() %3 : !quantum.bit

  %2 = quantum.insert %0[ 0], %out_qubits : !quantum.reg, !quantum.bit
  %4 = quantum.insert %2[ 1], %out_qubits_0 : !quantum.reg, !quantum.bit

```

This makes it easier to see if two qubits are from the same register. 

Also, in the original version's dataflow, the 1st qubit is in the forward slice of the 0th qubit (through the in-between insert `%2`), but this is not the most optimal description of the circuit: the two starting qubit values on the two wires should ideally be independent dataflow-wise.

rylanmalarchick:
**Context:**

`quantum.extract` / `quantum.insert` thread qubits through an SSA register value. When a circuit extracts qubit `b` after an insert at a distinct constant index `a`, the extract reads the post-insert register even though that insert never touched qubit `b`. That is a false data dependency between wires acting on different qubits of the same register (issue #2725).

Extend `ExtractOp` canonicalization to handle `%q = extract (insert %reg[a], v)[b]` where `a` and `b` are distinct static constants: rewrite the extract to read `%reg[b]` directly, and sink the bypassed insert down to just above the earliest remaining user of its result. Reading the pre-insert register drops the false dependency; sinking the insert keeps extracts grouped above the gates and inserts below them, which the downstream value-semantics and adjoint-lowering passes expect.

The canonicalizations are registered through `getCanonicalizationPatterns` with explicit `PatternBenefit` so that a same-index extract/insert round-trip cancels rather than being reshaped into a stranded pair when both patterns could apply: the same-index folds run at a higher benefit than the distinct-index redirect. The round-trip fold is rooted on the extract (replacing `InsertOp`'s cancellation canonicalizer), and the gradient-lowering and decompose-lowering cleanups pull `ExtractOp`'s patterns.

This targets `main` after the #2973 merge (decompose-lowering no longer runs `--canonicalize` internally). Verified locally: `check-dialects` 150/150, `test_xdsl_parity_synth` 55/55.

**Benefits:**

Breaks the false wire dependency the issue describes, exposing independent qubits to downstream analyses. In most cases it also simplifies the IR by collapsing inverse extract/insert round-trips: the `QRef/SemanticConversion` goldens shrink, and the long-standing TODO in `TestFlatCircuits.mlir` about distinct-index inverse pairs is now satisfied.

**Possible Drawbacks:**

One parity-synthesis test (`test_qjit_with_dynamic_allocation`) now synthesizes one extra CNOT (4 → 5) in its static-wire section. Parity synthesis is not monotone in how the gates are grouped, so the extracts-up reshaping lands a slightly worse grouping there; the circuit is value-identical and the pass's parity-equivalence checks still hold, so the golden was updated to match.

**Related GitHub Issues:**

Fixes #2725. Also satisfies the distinct-static-index case of the inverse extract/insert TODO in `TestFlatCircuits.mlir`; the insert-chain case (#2726) remains open.

Assisted-by: Claude Code


dime10:
Thanks @rylanmalarchick! We're currently resolving some CI issue and will get back to you afterwards.

mlir/lib/Quantum/IR/QuantumOps.cpp
```diff
@@ -162,13 +162,22 @@ LogicalResult ExtractOp::canonicalize(ExtractOp extract, mlir::PatternRewriter &
         bool staticallyEqual = bothStatic && extract.getIdxAttrAttr() == insert.getIdxAttrAttr();
         bool dynamicallyEqual = bothDynamic && extract.getIdx() == insert.getIdx();
 
+        bool staticallyDistinct = bothStatic && extract.getIdxAttrAttr() != insert.getIdxAttrAttr();
```
paul0403, line 248: We can just use the `staticallyEqual` from above right?

mlir/lib/Quantum/IR/QuantumOps.cpp
```diff
@@ -162,13 +162,22 @@ LogicalResult ExtractOp::canonicalize(ExtractOp extract, mlir::PatternRewriter &
...
         bool inSameBlock = extract->getBlock() == insert->getBlock();
 
         if ((staticallyEqual || dynamicallyEqual) && inSameBlock) {
             rewriter.replaceOp(extract, insert.getQubit());
             rewriter.replaceOp(insert, insert.getInQreg());
             return success();
         }
+
+        bool insertOnRegisterChain = llvm::any_of(
+            insert.getResult().getUsers(), [](Operation *user) { return !isa<ExtractOp>(user); });
```
paul0403: Would it be clearer to change the condition to be, perform the rewrite if the insert op has any users that is not the current extract op? Because the point is to prevent DCE eliminating the insert op if the extract op was the only user before the rewrite right?

mlir/test/QRef/SemanticConversion/TestFlatCircuits.mlir
```diff
@@ -280,23 +280,15 @@ func.func @test_namedobs_op() -> (!quantum.obs, !quantum.obs) attributes {quantu
...
 
-    // CHECK: [[extract:%.+]] = quantum.extract [[insertX]][ 1] : !quantum.reg -> !quantum.bit
-    // CHECK: [[obs_z:%.+]] = quantum.namedobs [[extract]][ PauliZ] : !quantum.obs
-    // CHECK: [[insertZ:%.+]] = quantum.insert [[insertX]][ 1], [[extract]] : !quantum.reg, !quantum.bit
+    // CHECK: [[obs_z:%.+]] = quantum.namedobs [[CNOT]]#1[ PauliZ] : !quantum.obs
+    // CHECK: [[insert1:%.+]] = quantum.insert [[insert0]][ 1], [[CNOT]]#1 : !quantum.reg, !quantum.bit
     %obs_z = qref.namedobs %q1 [ PauliZ] : !quantum.obs
 
-    // CHECK: quantum.dealloc [[insertZ]] : !quantum.reg
+    // CHECK: quantum.dealloc [[insert1]] : !quantum.reg
```
paul0403, line 291: Beautiful! 😻

mlir/test/Quantum/CanonicalizationTest.mlir
```diff
@@ -71,15 +71,14 @@ func.func @test_extract_insert_fold(%r1: !quantum.reg, %i: i64) -> !quantum.reg
...
 
-// CHECK-LABEL: test_extract_insert_no_fold_static
-func.func @test_extract_insert_no_fold_static(%r1: !quantum.reg, %i1: i64, %i2: i64) -> !quantum.reg {
-    // CHECK: quantum.extract
-    // CHECK: quantum.insert
+// CHECK-LABEL: test_extract_insert_distinct_static_folds
+func.func @test_extract_insert_distinct_static_folds(%r1: !quantum.reg, %i1: i64, %i2: i64) -> !quantum.reg {
+    // CHECK: %[[Q:.+]] = quantum.extract %arg0[ 0]
+    // CHECK: %[[R2:.+]] = quantum.insert %arg0[ 1], %[[Q]]
+    // CHECK: quantum.insert %[[R2]][%arg2], %[[Q]]
```
paul0403:
```suggestion
    // CHECK: [[Q:%.+]] = quantum.extract %arg0[ 0]
    // CHECK: [[R2:%.+]] = quantum.insert %arg0[ 1], [[Q]]
    // CHECK: quantum.insert [[R2]][%arg2], [[Q]]
```

We usually use this style

mlir/test/Quantum/CanonicalizationTest.mlir
```diff
@@ -90,6 +89,28 @@ func.func @test_extract_insert_no_fold_static(%r1: !quantum.reg, %i1: i64, %i2:
...
+func.func @test_extract_through_insert_distinct_index(%r0: !quantum.reg) -> !quantum.reg {
+    %q0 = quantum.extract %r0[0] : !quantum.reg -> !quantum.bit
+    %x = quantum.custom "PauliX"() %q0 : !quantum.bit
+    %r1 = quantum.insert %r0[0], %x : !quantum.reg, !quantum.bit
+    // CHECK: quantum.extract %arg0[ 1]
+    %q1 = quantum.extract %r1[1] : !quantum.reg -> !quantum.bit
+    %z = quantum.custom "PauliZ"() %q1 : !quantum.bit
+    %r2 = quantum.insert %r1[1], %z : !quantum.reg, !quantum.bit
+    return %r2 : !quantum.reg
+}
```
paul0403: Can you check the whole program for these tests? Just to make sure the extract/insert dataflow is checked.

mlir/test/Quantum/DecomposeLoweringTest.mlir
```diff
@@ -749,7 +749,9 @@ module @different_qreg_values{
     // CHECK: [[q0:%.+]] = quantum.extract [[reg]][ 0] : !quantum.reg -> !quantum.bit
     // CHECK: [[H:%.+]] = quantum.custom "Hadamard"() [[q0]] : !quantum.bit
     // CHECK: [[H_insert:%.+]] = quantum.insert [[reg]][ 0], [[H]] : !quantum.reg, !quantum.bit
-    // CHECK: [[full_insert:%.+]] = quantum.insert [[H_insert]][ 1], [[q1]] : !quantum.reg, !quantum.bit
+    // CHECK: [[q2_pre:%.+]] = quantum.extract [[reg]][ 2] : !quantum.reg -> !quantum.bit
+    // CHECK: [[insert_2:%.+]] = quantum.insert [[H_insert]][ 2], [[q2_pre]] : !quantum.reg, !quantum.bit
+    // CHECK: [[full_insert:%.+]] = quantum.insert [[insert_2]][ 1], [[q1]] : !quantum.reg, !quantum.bit
```
paul0403: Trying to understand what was adding these new ops. Can you explain a little bit? Because the new pattern should only be changing an operand on an extract op right?

paul0403 (commented):
Thanks for working on this!

paul0403:
The CI issues have been resolved. Can you merge from main again? Then we will run the CI and see the full impacts of this change in canonicalization :+1:

mlir/lib/Quantum/IR/QuantumOps.cpp
dime10: Do you mean `!staticallyEqual`? That wouldn't be equivalent since that could be triggered by them being dynamic as well, not just statically unequal

paul0403: I actually meant `bothStatic && !staticallyEqual`, but now that I look at it it's probably the same as what's here 😅 

Superficially looks like one less comparison but I can't imagine the compiler generating different code for these two. Feel free to resolve.

rylanmalarchick: Yeah, `bothStatic && !staticallyEqual` reduces to `bothStatic && unequal`, so it lands on the same condition as `staticallyDistinct`. I'll keep the direct form and resolve this. Thanks both.

mlir/lib/Quantum/IR/QuantumOps.cpp
rylanmalarchick: I tried that exact condition (rewrite when the insert has any user other than this extract) and it brings back the gate drop this guard prevents: `DecomposeLoweringTest` fails and an `RX` disappears, with the custom-op count going from 43 to 42. The reason is that "a user other than this extract" is also true when the only other user is another extract, which the greedy driver then redirects away as well; the insert ends up with no users and DCE removes it along with the gate it was feeding. Requiring a non-extract user keeps the insert on the register chain (a downstream insert or dealloc), so it can't be stranded. I kept that condition and renamed it to `insertHasNonExtractUser` to make the intent clear.

mlir/test/Quantum/CanonicalizationTest.mlir
rylanmalarchick: Done, switched to the `[[...]]` style.

mlir/test/Quantum/CanonicalizationTest.mlir
rylanmalarchick: Done. Both tests now check the full output, so the extract/insert dataflow is verified end to end: the redirected extract reads `%arg0` in `test_extract_through_insert_distinct_index`, and in the leaf case the extract still reads the insert result with the PauliX preserved.

mlir/test/Quantum/DecomposeLoweringTest.mlir
rylanmalarchick: You're right that the pattern only swaps the extract's `qreg` operand; it doesn't add ops. The extra extract/insert come from that swap turning off a downstream fold. In the `@my_cnot` decomposition, qubit 2 gets re-inserted into `%3` at index 2. `InsertOp::canonicalize` only folds an insert of an extracted qubit when the extract and insert share the register (its `sameQreg` check). Before this change qubit 2 was `extract %3[2]`, so `insert %3[2], extract(%3[2])` hit that case and folded away. After the redirect qubit 2 is `extract %0[2]`, so the extract reads `%0` while the insert targets `%3`, `sameQreg` is false, and the pair stays. It's semantically identical, since the only insert into `%3` is at index 0, so `%0[2]` and `%3[2]` are the same qubit, and the gate counts match baseline. So it trades the false dependency for two register ops that previously cancelled.

rylanmalarchick:
Pushed the review updates and merged `main`; should be ready for CI now. Thanks both for the careful look.

mlir/test/QRef/SemanticConversion/TestFlatCircuits.mlir
rylanmalarchick: thanks :)

rylanmalarchick:
Just wanted to nudge this. Merged main on the 19th as @paul0403 noted, so should be ready for CI now. No rush just wanted to make sure this was noted.

mehrdad2m:
Hi @rylanmalarchick, Thank you so much for working on this issue, and apologies on the delay in the review process. We'll try to complete the review as soon as possible. In the meanwhile, we ran the CI and it seems like there are a few minor failures. 
Cheers!

rylanmalarchick:
@mehrdad2m no worries, thanks! I looked into those 2 frontend failures locally. One is a pretty clear bug with a one line fix and the other is a more subtle interaction that you guys will probably want to steer me on.

1. SIGABRT in convert-to-value-semantics (test_dev_wires_have_pytree, both backends)

The distinct-index rewrite reassigns the extract's qreg operand but leaves the extract positioned after the insert. So it reads a register version a later insert has already superseded. This then trips IRMapping::lookup ("expected 'from' to be contained within the map") during value-semantics threading. Fix would be to also move the extract ahead of the insert:

  rewriter.modifyOpInPlace(extract, [&] {
      extract.getQregMutable().assign(insert.getInQreg());
      extract->moveBefore(insert);
  });

I verified locally and crash is gone with output numerically identical to baseline (maxdiff 0.0).
  
2. Parity-synth emits more CNOTs (test_xdsl_parity_synth::test_qjit_with_dynamic_allocation)

Structural thing, not just SSA renaming; the canonicalization runs before apply-transform-sequence, and hoisting the extract across the insert drops a quantum.insert mid-region. ParitySynthPattern treats InsertOp as a block boundary, so it synthesizes two smaller parity networks instead of one with 7 CNOTs vs the expected 4. I was able to confirm by toggling the pattern (off → 4, on → 7); numerics were correct either way. A local guard cannot fix this I think, as a canonicalizer can't see downstream block fragmentation, and the pattern fires in every canonicalize run.

Two paths I see here: 
- Move the hoisting into a dedicated pass after apply-transform-sequence. Would keep the optimization, stops it clobbering parity-synth.
- Keep it in canonicalize and update the CHECK lines. This would be more simple, but it ships the CNOT regression.

Or other options if you have better ideas than me :)

dime10:
Hi @rylanmalarchick, thanks for checking out those issues so quickly. 

> 1. SIGABRT in convert-to-value-semantics (test_dev_wires_have_pytree, both backends)

Good analysis! It looks like we are running into some hidden assumptions about the structure of inserts/extracts. The question boils down to whether these operations are pure (that is they may appear in any order as long as SSA dominance is satisfied), or whether their positions are relevant as well. Technically, they are declared side-effect free as seen here:
https://github.com/PennyLaneAI/catalyst/blob/a97d03a396381465acfd30cc4893a252032b5be1/mlir/include/Quantum/IR/QuantumOps.td#L195

I'd like to take a closer look before okaying the proposed solution. Swapping the extract and insert does solve their ordering, but it still leaves gates acting on the extracted qubit after the insert has "modified" the register. It might be fine (it should be if these operations were truly pure), but worth double checking.
The safest solution would be to follow the approach outlined in the issue:
```mlir
  %out_qubits = quantum.custom "PauliX"() %1 : !quantum.bit

  %3 = quantum.extract %0[ 1] : !quantum.reg -> !quantum.bit
  %out_qubits_0 = quantum.custom "PauliZ"() %3 : !quantum.bit

  %2 = quantum.insert %0[ 0], %out_qubits : !quantum.reg, !quantum.bit
  %4 = quantum.insert %2[ 1], %out_qubits_0 : !quantum.reg, !quantum.bit
```
We can see it's actually the insert that is moved down (delayed) as far as possible, which leads to a nice structure.

rylanmalarchick:
Thanks @dime10, I think that was the right cal here. I took the two issues above as separate, and rebuilt to test. They are less coupled than I implied above. 

For SIGABRT: since `extract`/`insert` are `NoMemoryEffect`, the rewrite is value-correct regardless of position; `extract %0[b]` reads the same qubit whether it sits before or after `insert %0[a]` at a distinct index, because the register is a functional SSA value rather than mutable storage. So I think the crash wasn't a semantic error, it was a stale-position assumption in `convert-to-value-semantics` (`IRMapping::lookup` expecting the extract to read the still-current register version). 

So the instinct to delay the insert downward rather than hoist the extract is the cleaner fix, because it satisfies that assumption and lands the nice extracts-up / inserts-down structure from the issue. I did a rewrite that reads the pre-insert register and then sinks the bypassed insert to just above the earliest user of its result, so inserts cluster after the gate run instead of being stranded mid-circuit.

I also isolated the cause on parity-synth. I think it is worth laying out because it bears on where this rewrite should live. `ParitySynthPattern` walks ops in textual order and flushes the accumulated phase-polynomial run whenever it hits an `InsertOp`, so the synthesized network (and its CNOT count) depends on exactly where inserts sit relative to the gates. I measured the checked circuit CNOTs with 3 separate ways. Baseline (no rewrite) is 8; the current operand-redirect is 11 (+3); the delay-insert form is 9 (+1). So the delay-insert structure you suggested cuts the regression substantially and fixes the crash, and 54/55 of the parity-synth tests pass with it.

I wanted input on the remaining +1. I think I had assumed a more contiguous gate run would always synthesize to fewer CNOTs, but that's not the case here. It seems that the baseline's particular insert placement happens to split the run at a point that synthesizes better than the fully-sunk form does. Looks like the synthesis isn't monotone in how the gates are grouped, so any reshape that runs before parity-synth carries some small, circuit-dependent delta. It seems the delay-insert form minimizes it but can't reliably drive it to zero. The only way I see to guarantee no parity-synth change is to not reshape the IR before it runs; i.e. move the distinct-index rewrite out of `canonicalize` into a pass scheduled after `apply-transform-sequence`. That keeps the #2725 false-dependency fix but trades away its benefit for any analysis (like parity-synth) that runs inside the transform sequence and actually wants the threaded form.

So this comes down a policy call (that I think you would rather take); keep it in `canonicalize` as the delay-insert form (fixes the crash, +1 on this one synthetic test, value-correct, and I would update the goldens), or move it to a dedicated post-transform pass for an exact-baseline parity-synth at the cost of the early-analysis benefit. I have the delay-insert version ready to push. But can take other direction if needed as well just lmk

mlir/lib/Quantum/IR/QuantumOps.cpp
mehrdad2m: Could we update this comment as well to reflect the new pattern?

mlir/test/Quantum/DecomposeLoweringTest.mlir
mehrdad2m: I agree that this is semantically equivalent, but it looks slightly awkward. If I understood correctly, this is a case where the pattern ordering is resulting in a different canonicalized form because if say the insert fold was triggered first, it would be folded away right? If that's the case, maybe it is worth having a similar mirrored pattern for the insertOp as well. That could ensure that no matter what is the order of applied canonicalization patterns, we would get the same result.

mehrdad2m (commented):
Thanks again @rylanmalarchick for the great work, I added two comments.

mlir/test/Quantum/DecomposeLoweringTest.mlir
rylanmalarchick: Yeah that is the cause of it. The leftover pair only survives because 'InsertOp::canonicalize' folds an inserted-extracted qubit only when the extract and insert name the same register value (its 'sameQreg' check), and the distinct-index rewrite makes the extract read the earlier register while the insert still targets the later one. So the result ends up depending on which pattern the greedy driven happens to apply first. 

A mirrored 'InsertOp' pattern is the correct fix in this case. Generalize that 'sameQreg' check from "same register" to "same register up to inserts at distinct static indicies" so 'insert(reg2, extract(reg1, i), i) folds when 'reg1' and 'reg2' agree at index 'i'. That makes the canonical form order-independent, and as a bonus it should (!) fold this awkward extract/insert pair away instead of leaving it, which should also remove the +2-ops drawback noted above.

One coupling I should flag is that this presupposes the rewrite stays in 'canonicalize'. (my open question from my reply last night regardging keeping in 'canonicalize' as the delay-insert form versus moving it to a pass after 'apply-transform-sequence') ; If it is kept in 'canonicalize' I would add the mirrored pattern with its own tests; if it moves to a dedicated pass the confluence concern mostly goes away. Either can be done

mehrdad2m: Yeah that makes sense. I think If making the existing passes' (including the parity-synth pass) assumptions inline with this canonical form then keeping it in the same pass is ideal, but I am not super familiar with the parity-synth implementation to know if it is trivial or not. I'll let @dime10 comment on that.

rylanmalarchick:
Pushed the delay-insert version (bd1655aa). The rewrite reads the pre-insert register and sinks the bypassed insert to just above its earliest remaining user, which lands the extracts-up / inserts-down structure you quoted; `test_extract_through_insert_distinct_index` pins it now. The SIGABRT is gone on both backends, and I checked it still aborts with the sink removed, so the test really covers the fix. Dialect suite is 147/147 with goldens updated for the new op order; gate counts and dataflow are unchanged in all of them. Also updated the comment @mehrdad2m flagged.

That leaves the parity golden as the one failure, the +1 CNOT case from before (the other 54 pass). Since @mehrdad2m leaned toward keeping this in `canonicalize`, my default is to update that golden and add the mirrored `InsertOp` pattern so the result is order-independent. If you'd rather parity-synth stay exactly at baseline I'll move the rewrite to a pass after `apply-transform-sequence` instead. Both are small from here, just lmk @dime10.

mlir/lib/Quantum/IR/QuantumOps.cpp
rylanmalarchick: Done in bd1655aa; the comment covers both patterns now (same-index fold and the distinct-index redirect + sink).

mlir/test/Quantum/DecomposeLoweringTest.mlir
dime10: Apologized for the wait, but I finally had the time to mull this over. The "problem" is indeed that more than one pattern could be applied to the given example code, and depending on which gets applied first the outcome is different. While I do believe the outcome is deterministic for any given example, it may not be the optimal one. The following minimal example demonstrates this:
```mlir
func.func public @circuit(%r: !quantum.reg, %q: !quantum.bit) -> (!quantum.reg) {

    %r1 = quantum.insert %r[ 0], %q : !quantum.reg, !quantum.bit
    %q1 = quantum.extract %r1[ 1] : !quantum.reg -> !quantum.bit
    %r2 = quantum.insert %r1[ 1], %q1 : !quantum.reg, !quantum.bit

    return %r2 : !quantum.reg
}

```
I don't think a mirrored pattern makes sense here, at least if by that you mean a pattern that would undo the swap of the insert-extract chain. Not only because that could lead to an infinite loop (I think), but also because we don't really ever want to "close up" a region of local qubit dataflow, we are always interested in expanding it to expose the dependencies between gates better.

I'm also not sure about the expanding the complexity of the cancellation pattern with the condition "same register up to inserts at distinct static indices". That could allow the cancellation to go through in more cases, but it also increase the complexity and begs the question how deep such chain should be followed up.

An alternative that I would like us to look into is specifying priorities for different transformations. For instance, cancellation should probably always have the highest priority, while reordering extracts/inserts for better dataflow would come second. MLIR has a "benefit" system in its patter rewriting infrastructure that was designed exactly for the purpose of prioritizing multiple possible transforms.

dime10:
> 1. SIGABRT in convert-to-value-semantics (test_dev_wires_have_pytree, both backends)

I was looking at the previous failure again and I'm a bit confused. You mention in your analysis that it's the `convert-to-value-semantics` pass that triggers the IRMap lookup error. Yet in the [CI runs](https://github.com/PennyLaneAI/catalyst/actions/runs/27846508725/job/83740587465) I can only see this error happening for the `adjoint-lowering` pass. We can see this in the traceback: 
```
IRMapping.h:74: auto mlir::IRMapping::lookup(T) const [with T = mlir::detail::TypedValue<catalyst::quantum::QuregType>]: Assertion `result && "expected 'from' to be contained within the map"' failed.
...
 #9 0x0000562036bfd15e auto mlir::IRMapping::lookup<mlir::detail::TypedValue<catalyst::quantum::QuregType>>(mlir::detail::TypedValue<catalyst::quantum::QuregType>) const AdjointPatterns.cpp:0:0
...
#16 0x0000562036bfb5f1 catalyst::quantum::AdjointLoweringPass::runOnOperation() adjoint_lowering.cpp:0:0
```

The reason I bring this up is because I want to make sure we look into this pass and fix the fact that it seems sensitive to valid insert/extract orderings (where I'm assuming these instructions are indeed pure), even if in this PR we've avoided the error. But fixing the pass is definitely a separate issue so it doesn't need to be solved here.

Just to recap the strategies we've considered so far:

1. V1: simply rewire the input of the extractOp without changing anything else
  - here we asserted that using a register value after it was "invalidated" by a consuming operation (insert) is not good
  - if the ops were truly pure (and fix the failing pass) this wouldn't be an issue
2. V2: rewire & reorder the extractOp with the insertOp
  - fixes the "invalidation" issue in V1 (even though the SSA/dataflow graph is identical to V1)
  - side-steps the pass failure
3. V3: rewire the extractOp and push the insertOp all the down to its next user
  - while I no longer think this is absolutely necessary anymore, it does have the added benefit of making the local dataflow region visually easier to parse
  - also identical SSA graph to V1

In short we can see that the dataflow graph is the same no matter what, and the differences are due to some positional side-effects (most likely unintended), and the not truly linear nature of these quantum values (extract can reuse, insert invalidates). Hopefully that's something we can improve going forward. I'm okay with V3 here, but I wanted to check with @paul0403 and @mehrdad2m about their thoughts on these different versions.

Regarding the CNOT count, I'm not too worried about it. I will loop in the author of that pass to have a look, but the simplifying the qubit/register has a lot of benefits so I think we should prioritize that. Adding an extra CNOT count to the test's expected output seems okay.

paul0403:
I agree that V3 is the best, although I still want the adjoint pass fixed (I don't know if it's the responsibility of this PR though).

I'm also curious as to the ref semantics conversion pass error, since we couldn't find it in CI

rylanmalarchick:
Thanks all, sounds like we're settled on V3 (rewire the extract + sink the bypassed insert down to its next user). That's what's on the branch so I'll leave it there.

Rebased on latest `main`, which pulled in #2973 (decompose-lowering doesn't run `inline`/`cse`/`canonicalize` internally anymore). Kind of a nice side effect: the +2 register ops thing we were going back and forth on in `@different_qreg_values` just goes away. The rewrite only ever fired in that pass via its old internal `--canonicalize`, and now that's gone, so it doesn't fire there at all and I reverted those goldens back to main's baseline. Rewrite itself is untouched and still exercised directly under `--canonicalize` in `CanonicalizationTest.mlir`. Dialect suite is 149/149 locally after the merge.

On the IRMapping crash, good catch @dime10, and yeah you're right, the CI failure is in `adjoint-lowering` (`AdjointPatterns.cpp`), not value-semantics. My bad on the earlier writeup, the `convert-to-value-semantics` one was a separate local repro I hit of the same `IRMapping::lookup` assertion, which is probably why @paul0403 couldn't find it in CI, it wasn't the CI failure. Same root cause either way though: the pass assumes an extract reads the current register version instead of treating extract/insert as position-independent, and V3's extracts-up/inserts-down form keeps both passes happy. So it's green now, but I agree the real fix is hardening those passes so they don't care about valid reorderings. Happy to open a separate issue for that, feels out of scope for this one.

For the CNOT count, went with @dime10's call, left the cancellation pattern alone (no mirrored `InsertOp` pattern) and just bumped the expected output in `test_qjit_with_dynamic_allocation`. Reshaped run synths to one extra CNOT (4 → 5) and the three `arg0` extracts group up above the run. Numbers-wise it's the same circuit, parity-equivalence checks still hold, all 55 parity-synth tests pass. And the SIGABRT test (`test_dev_wires_have_pytree`) passes now on both `lightning.qubit` and `lightning.kokkos`.

lmk if the separate issue for the adjoint/value-semantics ordering thing sounds good and I'll file it.

dime10:
Great, thanks for the response! Then we'll open an issue about hardening the adjoint-lowering / value-semantics-conversion passes against extract/insert positioning, or you are free to open it yourself if you'd like (since you offered) :)

I'm happy to provide final review and approval once the last comment about pattern ordering is addressed. While it may no longer lead to a pass failure in the current state, the underlying issue is still there so if we can implement preferential application of the old pattern vs the new, that would be better.

rylanmalarchick:
Sounds good, addressed the ordering. The distinct-index redirect now defers when the extract feeds a same-register, same-index insert, so the cancellation collapses that pair instead of the redirect stranding it. Your minimal example now folds down to the single `insert %r[0]`, and it's order-independent either way. Added a test for it in `CanonicalizationTest.mlir` (`test_extract_through_insert_prefers_cancellation`). Dialect suite 149/149, parity 55/55.

Noted on opening the adjoint-lowering / value-semantics hardening issue, will keep that in mind.

mlir/test/Quantum/DecomposeLoweringTest.mlir
dime10: > Sounds good, addressed the ordering. The distinct-index redirect now defers when the extract feeds a same-register, same-index insert, so the cancellation collapses that pair instead of the redirect stranding it. Your minimal example now folds down to the single insert %r[0], and it's order-independent either way. Added a test for it in CanonicalizationTest.mlir (test_extract_through_insert_prefers_cancellation). Dialect suite 149/149, parity 55/55.

Thanks, but it seems you've hardcoded this pattern. Could you look into the benefit mechanism please?

rylanmalarchick:
Good call, switched it over to the benefit mechanism. `ExtractOp` now registers its patterns through `getCanonicalizationPatterns` with explicit `PatternBenefit`: the same-index fold and the extract/insert round-trip cancellation at benefit 2, the distinct-index redirect at benefit 1. The cancellation and the redirect match the same op now, so the benefit ordering picks the cancellation when both apply, no hand-written guard. Your minimal example still folds to the single `insert %r[0]`.

(Kept `InsertOp`'s cancellation too since the gradient-lowering cleanup pulls it directly; the extract-side one is its mirror, just there so the benefit can outrank the redirect.)

Dialect 149/149, parity 55/55.

mlir/lib/Quantum/IR/QuantumOps.cpp
```diff
@@ -153,29 +153,136 @@ OpFoldResult ExtractOp::fold(FoldAdaptor adaptor)
...
+    patterns.add<CancelExtractInsertRoundTrip>(context, /*benefit=*/2);
+    patterns.add<RedirectExtractThroughDistinctInsert>(context, /*benefit=*/1);
 }
 
 LogicalResult InsertOp::canonicalize(InsertOp insert, mlir::PatternRewriter &rewriter)
 {
+    // Cancel an insert of a qubit that was just extracted from the same register at the same
+    // index. The gradient-lowering cleanup relies on this canonicalizer to remove round-trips
+    // (see gradient_lowering.cpp); the extract-rooted CancelExtractInsertRoundTrip above mirrors
+    // it so pattern benefits can prioritize it over the distinct-index redirect.
```
dime10: I don't think we should document all the users of a pass/pattern in the pattern itself.

mlir/lib/Quantum/IR/QuantumOps.cpp
```diff
@@ -153,29 +153,136 @@ OpFoldResult ExtractOp::fold(FoldAdaptor adaptor)
...
-{
-    // Handle the pattern: %reg2 = insert %reg1[idx], %qubit -> %q = extract %reg2[idx]
-    // Convert to: %q = %qubit, and replace other uses of %reg2 with %reg1
-    if (auto insert = dyn_cast_if_present<InsertOp>(extract.getQreg().getDefiningOp())) {
+namespace {
+
+// Fold an extract that reads back the qubit a same-index insert just wrote:
+//   %reg2 = insert %reg1[i], %q ; %out = extract %reg2[i]  ->  %out = %q
+// and forward %reg2's remaining users to %reg1.
+struct FoldExtractOfSameIndexInsert : public mlir::OpRewritePattern<ExtractOp> {
```
dime10, line 186: why is this one called fold and the other (reverse version) cancel? Let's stick with a common naming scheme :)

mlir/lib/Quantum/IR/QuantumOps.cpp
```diff
@@ -153,29 +153,136 @@ OpFoldResult ExtractOp::fold(FoldAdaptor adaptor)
...
+        rewriter.replaceOp(insert, insert.getInQreg());
+        return success();
+    }
+};
+
+// Cancel an extract/insert round-trip at a single index:
+//   %q = extract %reg[i] ; %reg2 = insert %reg[i], %q  ->  %reg2 = %reg
+// This is the extract-rooted mirror of InsertOp::canonicalize's cancellation. It exists so the
+// cancellation and the distinct-index redirect below match the same op (the extract), letting
+// pattern benefits arbitrate: at a higher benefit than the redirect, the cancellation wins when
```
dime10: Good point, I hadn't considered this caveat. I feel like there is no need to have the same pattern exist twice though, so either we:
- delete the insert canonicalization
- move the reorder to the insert

Since these are already written, I'm fine to just delete the insert one.

mlir/lib/Quantum/IR/QuantumOps.cpp
```diff
@@ -153,29 +153,136 @@ OpFoldResult ExtractOp::fold(FoldAdaptor adaptor)
...
+            return failure();
+        }
+        // Only fire while the bypassed insert still has a non-extract user, so it stays on the
+        // register chain (a downstream insert or dealloc) and cannot be stranded and DCE'd along
+        // with the gate feeding it.
+        bool insertHasNonExtractUser = llvm::any_of(
+            insert.getResult().getUsers(), [](Operation *user) { return !isa<ExtractOp>(user); });
+        if (!insertHasNonExtractUser) {
+            return failure();
+        }
```
dime10: Do we care? I guess we just need _a_ second user to move the op to, which is a simpler condition. Although, assuming it somehow didn't have another user, we could still proceed with the swap and leave the insert op where it is. Your loop below already handles this actually, so I think we can just delete these lines.

dime10 (commented):
Thanks @rylanmalarchick, the latest structure is great 💯 Final rounds of comments regarding the code itself and we should be good to go :)

rylanmalarchick:
Cleaned these up, thanks:

- **Dup cancellation / documenting pass users** — deleted the `InsertOp` one and kept just the extract-rooted round-trip fold (the one that competes with the redirect by benefit). The gradient-lowering cleanup now pulls `ExtractOp`'s canonicalizers instead of `InsertOp`'s; `Gradient/CleanupFunctions.mlir` still passes. That also drops the comment about who depends on it.
- **Naming** — renamed `CancelExtractInsertRoundTrip` -> `FoldExtractInsertRoundTrip` so both same-index folds share the `Fold*` scheme.

On the `insertHasNonExtractUser` check; I'd actually like to keep that one. It's not about finding a move target, it's guarding a gate drop: if the insert's only users are extracts, redirecting them all away leaves the insert with no users, and since it's `NoMemoryEffect` DCE deletes it *and* the gate feeding its qubit operand. So e.g.

```mlir
%q0 = quantum.extract %r0[0]
%x  = quantum.custom "PauliX"() %q0
%r1 = quantum.insert %r0[0], %x
%q1 = quantum.extract %r1[1]
return %q1
```

without the check, canonicalize drops the `PauliX` entirely and you're left with just `extract %r0[1]`. `test_extract_no_redirect_when_insert_is_leaf` pins this. The sink loop leaving the insert in place doesn't save it; it's dead either way once its last real user is gone. Lmk if you'd rather handle that a different way, but I don't think we can just drop the lines. hopefully this is our last bout :)

Dialect 149/149, parity 55/55.

dime10:
Thanks @rylanmalarchick, it would be easier if you replied in appropriate comment threads though.

mlir/lib/Quantum/IR/QuantumOps.cpp
dime10: I've had a look at your example:
```mlir
%q0 = quantum.extract %r0[0]
%x  = quantum.custom "PauliX"() %q0
%r1 = quantum.insert %r0[0], %x
%q1 = quantum.extract %r1[1]
return %q1
```

The thing is, this is not a valid program because it leaves the register dangling. All registers values should eventually be consumed (extract is a use but doesn't consume), eventually terminating in a dealloc instruction.

dime10 (approved):
Everything else looks good, I'm going to give preliminary approval. Well done! 🦖 

Note that to be compliant with our [policy](https://github.com/PennyLaneAI/catalyst/blob/main/AI_Policy.md) you must label the AI agent used to assist you in the PR description.

mlir/lib/Quantum/IR/QuantumOps.cpp
rylanmalarchick: Agreed, removed the guard along with test_extract_no_redirect_when_insert_is_leaf.

One thing this surfaced: test_paulirot in DecomposeLoweringTest.mlir then fails, because its final insert consumes %reg2 rather than %reg3, leaving the [1] insert chain unconsumed. That wiring goes back to #2893 and I couldn't find it discussed anywhere. I've fixed the test to consume %reg3 in this branch, but happy to revert that if needed. Dialect 150/150, parity 55/55.

mehrdad2m (approved):
Thanks @rylanmalarchick for addressing the review. It looks great! 🚀

mlir/lib/Quantum/IR/QuantumOps.cpp
dime10: Good catch, that was definitely a bug in the test!

PR #2967, opened 2026-06-18
Add PBC depth to resource-analysis JSON

sengthai:
Previously, the `resource-analysis` counts only gate and qubit number. The PBC depth analysis only exists in `ppm_specs` pass. 

In this PR, we  add PBC depth (`any_commuting_depth` / `qubit_disjoint_depth`) to `resource-analysis` JSON output. Also, report depth for both function bodies and lifted loop entries (`for_loop_*`, `dyn_for_loop_*`), same operation counts. For non-pbc circuit, the `depth` value is empty. 

Note:
`any_commuting_depth`: the PPR/PPM that has _non-overlapping qubit_ **or** _commute_ to all gates in the layer.
`qubit_disjoint_depth`: the PPR/PPM that has _non-overlapping qubit_ to all gates in the layer.

For the circuit something like this:

```mlir
scf.for ... to %n ... {
  pbc.ppr ["Z"](4) ...
}
pbc.ppr ["X"](4) ...
```

The output:

```json
{
  "circuit": {
    "depth": {
      "any_commuting_depth": 1,
      "qubit_disjoint_depth": 1
    },
    "var_function_calls": {
      "dyn_for_loop_1": "0x..."
    }
  },
  "dyn_for_loop_1": {
    "depth": {
      "any_commuting_depth": 1,
      "qubit_disjoint_depth": 1
    }
  }
}
```

See also: https://github.com/PennyLaneAI/pennylane/pull/9703


mlir/lib/Catalyst/Transforms/ResourceAnalysisPass.cpp
```diff
@@ -163,25 +203,40 @@ struct ResourceAnalysisPass : public impl::ResourceAnalysisPassBase<ResourceAnal
         if (result.autoQubitManagement.has_value()) {
             funcObj["auto_qubit_management"] = *result.autoQubitManagement;
         }
+        llvm::json::Object depthObj;
+        if (depth) {
+            depthObj["depth_0"] = depth->first;
+            depthObj["depth_1"] = depth->second;
+        }
+        funcObj["depth"] = std::move(depthObj);
```
jzaia18, line 209: Since this only relates to PBC, can we rename this field to something like "pbc_depth"?

mlir/include/PBC/Utils/PBCLayer.h
```diff
@@ -25,6 +28,9 @@
 namespace catalyst {
 namespace pbc {
 
+// (depth_0, depth_1), or nullopt if there is nothing to report.
+using BothWorstCaseDepth = std::optional<std::pair<int64_t, int64_t>>;
```
jzaia18:
```suggestion
using PBCDepths = std::optional<std::pair<int64_t, int64_t>>;
```

Maybe something like this for the name? And explain that it's an upper bound in the comment above

mlir/lib/Catalyst/Transforms/ResourceAnalysisPass.cpp
sengthai: Yes sounds good!

mlir/lib/Catalyst/Transforms/ResourceAnalysisPass.cpp
```diff
@@ -88,6 +90,43 @@ struct ResourceAnalysisPass : public impl::ResourceAnalysisPassBase<ResourceAnal
...
+                return;
+
+            pbc::PBCLayerContext layerContext;
+            depths[funcOp.getName()] =
+                layerContext.computeBothWorstCaseDepths(&funcOp.getBody().front());
+        });
+
+        // Handle dynamic loop bodies.
+        for (const auto &entry : analysis.getSyntheticLoopBodies()) {
+            llvm::outs() << "Synthetic loop body: " << entry.getKey() << "\n";
```
jzaia18:
```suggestion
```

mlir/lib/Catalyst/Transforms/ResourceAnalysisPass.cpp
```diff
@@ -113,7 +152,8 @@ struct ResourceAnalysisPass : public impl::ResourceAnalysisPassBase<ResourceAnal
     }
 
     /// Serialize a single ResourceResult into a JSON object.
-    static llvm::json::Object resultToJson(const ResourceResult &result)
+    static llvm::json::Object resultToJson(const ResourceResult &result,
+                                           const pbc::BothWorstCaseDepth &depth)
```
jzaia18: Is it necessary to pass this separately? It seems like it should be part of the `ResourceResult` object (especially if we're redesigning the frontend to explicitly include this data in the `SpecsResources` object)

mlir/include/PBC/Utils/PBCLayer.h
sengthai: Yes sure, no objection!

mlir/lib/Catalyst/Transforms/ResourceAnalysisPass.cpp
sengthai: Actually, i want to make ResourceResult becomse the base class, then PBCResourceResult should have depth. But it is also fine to make depth as part of ResourceResult.

mlir/include/PBC/Utils/PBCLayer.h
sengthai: [d30de16](https://github.com/PennyLaneAI/catalyst/pull/2967/commits/d30de161c17b97a4ae83d079f63f9a68de0f9e1c)

mlir/lib/Catalyst/Transforms/ResourceAnalysisPass.cpp
sengthai: [d30de16](https://github.com/PennyLaneAI/catalyst/pull/2967/commits/d30de161c17b97a4ae83d079f63f9a68de0f9e1c)

mlir/lib/Catalyst/Transforms/ResourceAnalysisPass.cpp
jzaia18: > Actually, i want to make ResourceResult becomse the base class, then PBCResourceResult should have depth. 

I think this is an even better solution. This is in line with what I'd expect we'd need to do in the frontend for the resource unification work next quarter as well.

sengthai: we now can set that depth belong to ResourceResult first, then we can unify resource analysis later next quarter.

sengthai: [ad23199](https://github.com/PennyLaneAI/catalyst/pull/2967/commits/ad23199466bd2c9ce76ea058767136c9eb4793bf)

mlir/include/PBC/Utils/PBCLayer.h
```diff
@@ -51,23 +57,29 @@ class PBCLayerContext {
...
+    // unless `skipDynamic` is true (dynamic loops then contribute 0).
     mlir::FailureOr<int64_t> computeWorstCaseDepth(mlir::Block *block,
-                                                   bool onlyOnDisjointQubit = false);
+                                                   bool onlyOnDisjointQubit = false,
+                                                   bool skipDynamic = false);
+
+    // Like `computeWorstCaseDepth`, but maps depth 0 to nullopt.
+    mlir::FailureOr<std::optional<int64_t>>
+    computeBlockWorstCaseDepth(mlir::Block *block, bool onlyOnDisjointQubit = false,
+                               bool skipDynamic = false);
```
jzaia18: Could we rename these functions? They both seem to work on blocks, maybe `computeBlockWorstCaseDepth` and `computeBlockWorstCaseDepthNonzero`?

sengthai: Yep, sure! 

[1b4b75f](https://github.com/PennyLaneAI/catalyst/pull/2967/commits/1b4b75fe1c08effa6df9643894d901f7897ad1bd)

mlir/lib/Catalyst/Transforms/ResourceAnalysisPass.cpp
```diff
@@ -88,6 +92,39 @@ struct ResourceAnalysisPass : public impl::ResourceAnalysisPassBase<ResourceAnal
...
+                    diag.str().find("worst-case depth") != std::string::npos) {
+                    return success();
+                }
+                return failure();
+            });
+
+        // Handle static loop bodies.
+        getOperation()->walk([&](func::FuncOp funcOp) {
+            if (funcOp.isDeclaration())
+                return;
```
rniczh:
```suggestion
            if (funcOp.isDeclaration()) {
                return;
            }
```

mlir/lib/PBC/Utils/PBCLayer.cpp
```diff
@@ -176,6 +212,43 @@ FailureOr<int64_t> PBCLayerContext::computeWorstCaseDepth(Block *block, bool onl
...
+        if (failed(p1) || !p1->has_value())
+            return std::nullopt;
+        return {{**p0, **p1}};
+    }
+    if (*d0 == 0)
+        return std::nullopt;
+    auto d1 = computeBlockWorstCaseDepth(block, /*onlyOnDisjointQubit=*/true);
+    if (failed(d1) || *d1 == 0)
+        return std::nullopt;
+    return {{*d0, *d1}};
```
rniczh: Forgot the { }

```
if (cond) {
      ...
}
```

rniczh (approved):
👍

mlir/lib/PBC/Utils/PBCLayer.cpp
sengthai: [d13bb64](https://github.com/PennyLaneAI/catalyst/pull/2967/commits/d13bb647163be9edcf5d34273166f854a6e45638)
done!

mlir/lib/PBC/Utils/PBCLayer.cpp
```diff
@@ -176,6 +206,47 @@ FailureOr<int64_t> PBCLayerContext::computeWorstCaseDepth(Block *block, bool onl
...
+        }
+        return {{**p0, **p1}};
+    }
+    if (*d0 == 0) {
+        return std::nullopt;
+    }
+    auto d1 = computeBlockWorstCaseDepth(block, /*onlyOnDisjointQubit=*/true);
+    if (failed(d1) || *d1 == 0) {
+        return std::nullopt;
+    }
```
lazypanda10117, line 230: why do we not care about `skipDynamic` in the else branch?

mlir/lib/PBC/Utils/PBCLayer.cpp
```diff
@@ -143,15 +171,16 @@ FailureOr<int64_t> PBCLayerContext::computeWorstCaseDepth(Block *block, bool onl
...
-        if (op.getNumRegions() == 0) {
+        // PPM variants without PBCOpInterface still contribute one layer each.
+        if (isa<SelectPPMeasurementOp>(&op)) {
+            flushLayer();
+            depth += 1;
             continue;
         }
 
-        // multi-region ops
-        if (op.getNumRegions() != 1) {
```
lazypanda10117, line 152: Can we analyze multi-region ops now?

mlir/lib/PBC/Utils/PBCLayer.cpp
```diff
@@ -143,15 +171,16 @@ FailureOr<int64_t> PBCLayerContext::computeWorstCaseDepth(Block *block, bool onl
             continue;
         }
 
-        // plain ops
-        if (op.getNumRegions() == 0) {
+        // PPM variants without PBCOpInterface still contribute one layer each.
+        if (isa<SelectPPMeasurementOp>(&op)) {
```
lazypanda10117, line 165: Is this only true for select PPM? What about PPM?

Additionally, how's PPR and cond PPR being handled in the depth calculation

mlir/test/Catalyst/ResourceAnalysisTest.mlir
```diff
@@ -147,6 +150,108 @@ func.func @dynamic_for_loop(%arg0: !quantum.bit, %n: index) -> !quantum.bit {
...
+// CHECK-DAG: "qubit_disjoint_depth": 1
+func.func private @depth_helper(%arg0: !quantum.bit) -> !quantum.bit {
+    %out = pbc.ppr ["Z"](4) %arg0 : !quantum.bit
+    return %out : !quantum.bit
+}
+
+func.func @depth_caller(%arg0: !quantum.bit) -> !quantum.bit {
+    %out = func.call @depth_helper(%arg0) : (!quantum.bit) -> !quantum.bit
+    return %out : !quantum.bit
+}
```
lazypanda10117, line 323: Would be good to have an explicit example lit test where the two depths differ

mlir/lib/PBC/Utils/PBCLayer.cpp
```diff
@@ -176,6 +206,47 @@ FailureOr<int64_t> PBCLayerContext::computeWorstCaseDepth(Block *block, bool onl
     return depth;
 }
 
+FailureOr<std::optional<int64_t>>
+PBCLayerContext::computeBlockWorstCaseDepthNonzero(Block *block, bool onlyOnDisjointQubit,
+                                                   bool skipDynamic)
+{
+    FailureOr<int64_t> depth = computeBlockWorstCaseDepth(block, onlyOnDisjointQubit, skipDynamic);
```
lazypanda10117: Why do we create this function here? It seems like we can just use `computeBlockWorstCaseDepth` and check directly using something similar to`(failed(d1) || *d1 == 0)`. Having this new function doesn't seem to add much to code clarity imo.

mlir/lib/Catalyst/Transforms/ResourceAnalysisPass.cpp
```diff
@@ -88,6 +92,40 @@ struct ResourceAnalysisPass : public impl::ResourceAnalysisPassBase<ResourceAnal
     }
 
   private:
+    /// Populate PBC worst-case depth on each function and lifted loop body entry.
+    void populatePBCDepths(llvm::StringMap<ResourceResult> &results,
+                           const ResourceAnalysis &analysis)
+    {
+        // Swallow expected errors
+        // Happens when the depth is counter in dynamic loops.
```
jzaia18: Does this mean the error happens if you try to compute the depth of a dynamic loop body? Or if you have a function which calls into a dynamic loop lifted function?

mlir/lib/PBC/Utils/PBCLayer.cpp
```diff
@@ -72,22 +74,45 @@ FailureOr<int64_t> PBCLayerContext::switchWorstCaseDepth(scf::IndexSwitchOp swit
...
+        }
+    }
+
+    if (!tripCount) {
+        if (skipDynamic) {
+            return 0;
+        }
         return forOp.emitOpError(
             "worst-case depth is not available when there are dynamically sized for loops");
     }
```
jzaia18: Should this be a utility function somewhere that we just call in both places? I think it'd be better not to have duplicated code if we can avoid it

mlir/lib/PBC/Utils/PBCLayer.cpp
jzaia18, line 151: Correct me if I'm misunderstanding how this works, but I think this is against our design for the resource analysis pass. For the other resources we don't include the resources of internal functions in the outer function, since those loops get outlined and the values are combined in the frontend. This seems like it counts the loop depth twice, both as a part of the caller and within the loop's outlined function itself

mlir/lib/PBC/Utils/PBCLayer.cpp
```diff
@@ -176,6 +206,47 @@ FailureOr<int64_t> PBCLayerContext::computeWorstCaseDepth(Block *block, bool onl
...
+    }
+    if (*depth == 0) {
+        return std::optional<int64_t>(std::nullopt);
+    }
+    return std::optional<int64_t>(*depth);
+}
+
+PBCDepths PBCLayerContext::computePBCDepth(Block *block)
+{
+    // Try to calcuate the depth with static first, then fallback to skip-dynamic.
```
jzaia18:
```suggestion
    // Try to calculate the depth with static first, then fallback to skip-dynamic.
```

mlir/lib/Catalyst/Transforms/ResourceAnalysisPass.cpp
```diff
@@ -74,7 +76,9 @@ struct ResourceAnalysisPass : public impl::ResourceAnalysisPassBase<ResourceAnal
         std::string jsonStr = "";
 
         if (outputJson) {
-            jsonStr = buildJsonString(results);
+            llvm::StringMap<ResourceResult> resultsWithDepth = results;
+            populatePBCDepths(resultsWithDepth, analysis);
+            jsonStr = buildJsonString(resultsWithDepth);
```
jzaia18, line 81:
```suggestion
            populatePBCDepths(results, analysis);
            jsonStr = buildJsonString(results);
```

Any reason not to just do this? Or move-construct?

mlir/lib/PBC/Utils/PBCLayer.cpp
```diff
@@ -72,22 +74,45 @@ FailureOr<int64_t> PBCLayerContext::switchWorstCaseDepth(scf::IndexSwitchOp swit
...
     }
 
-    FailureOr<int64_t> bodyDepth = computeWorstCaseDepth(forOp.getBody(), onlyOnDisjointQubit);
+    FailureOr<int64_t> bodyDepth =
+        computeBlockWorstCaseDepth(forOp.getBody(), onlyOnDisjointQubit, skipDynamic);
     if (failed(bodyDepth)) {
         return failure();
     }
-    return iterations * (*bodyDepth);
+    return *tripCount * (*bodyDepth);
```
jzaia18: I think we shouldn't use the tripCount here, since the depth should be per loop iteration to follow the same format as the frontend expects

mlir/lib/PBC/Utils/PBCLayer.cpp
sengthai: I made this less restrict, because there is multiple region in stablehlo dialect.

jzaia18: But we don't actually analyze the regions after the first one, do we?

mlir/lib/Catalyst/Transforms/ResourceAnalysisPass.cpp
sengthai: The `results` is const and `populatePBCDepths` update it, so we need to copy results to resultsWithDepth first.

mlir/lib/PBC/Utils/PBCLayer.cpp
sengthai: Good point! I made it change. 
[c31627b](https://github.com/PennyLaneAI/catalyst/pull/2967/commits/c31627b94513b602bf27d49aceb10da1cc0f68f0)

mlir/lib/PBC/Utils/PBCLayer.cpp
sengthai: @jzaia18 makes sense. we only care the first region here which is not right.

jzaia18: I think continuing to raise an error for multi-regions is still the best way to go. For ops with many regions I'm not sure it makes sense to analyze regions without knowing what they represent

sengthai: Okay make sense. Let's me bring this back.

sengthai: [f79a7ee](https://github.com/PennyLaneAI/catalyst/pull/2967/commits/f79a7ee7bd09a6f9203a598c9a679d0943e20782)

mlir/lib/Catalyst/Transforms/ResourceAnalysisPass.cpp
sengthai: Yes, `ScopedDiagnosticHandler` gonna swallow the error then when we try to track the dynamic wires in dynamic loop body. Initially, the pbc depth analysis function is not supposed to support the symbolic spec. so that the reason we need to handle that error and use symbolic result result.

mlir/lib/PBC/Utils/PBCLayer.cpp
sengthai: Nice spot! this is because we use that in ppm_specs as well where we accumulate all the depth for-loop. Let's me fix this one.

mlir/lib/PBC/Utils/PBCLayer.cpp
sengthai: For now, that is spacial case because SelectPPMeasurementOp does not implement PBCOpInterface. Everything else (PPM, PPR, cond PPR) does implement it, so they're all handled by branch above.

mlir/lib/PBC/Utils/PBCLayer.cpp
sengthai: skipDynamic  doesn't matter in else branch. the reason is that whether skipDynamic changes anything depends only on whether the block contains a dynamic loop and that is completely independent of `onlyOnDisjointQubit`.

mlir/lib/PBC/Utils/PBCLayer.cpp
sengthai: Fixed in [47ad86](https://github.com/PennyLaneAI/catalyst/pull/2967/commits/647ad864d42ae25ce67066f956c5f35caa55c414)

mlir/lib/PBC/Utils/PBCLayer.cpp
sengthai: Now fixed that in [47ad86](https://github.com/PennyLaneAI/catalyst/pull/2967/commits/647ad864d42ae25ce67066f956c5f35caa55c414)

mlir/test/Catalyst/ResourceAnalysisTest.mlir
sengthai: Thanks! I added more examples in here [47ad86](https://github.com/PennyLaneAI/catalyst/pull/2967/commits/647ad864d42ae25ce67066f956c5f35caa55c414)

mlir/lib/PBC/Utils/PBCLayer.cpp
sengthai: Yes good point! 
I added in SCFUtils then.
[b5d2d5c](https://github.com/PennyLaneAI/catalyst/pull/2967/commits/b5d2d5c3ead3fbb7876c3c1eadc67a0c961ffe6a)

mlir/lib/PBC/Utils/PBCLayer.cpp
```diff
@@ -74,45 +71,38 @@ FailureOr<int64_t> PBCLayerContext::switchWorstCaseDepth(scf::IndexSwitchOp swit
...
+    std::optional<int64_t> tripCount = resolveForLoopTripCount(forOp);
 
     if (!tripCount) {
-        if (skipDynamic) {
+        if (skipDynamic_) {
             return 0;
         }
         return forOp.emitOpError(
             "worst-case depth is not available when there are dynamically sized for loops");
     }
```
jzaia18, line 85: It seems like we only need the trip count if `skipDynamic_` is false, can we reorganize this to avoid computing it when it isn't needed?

mlir/lib/PBC/Utils/PBCLayer.cpp
```diff
@@ -194,7 +190,7 @@ FailureOr<int64_t> PBCLayerContext::computeBlockWorstCaseDepth(Block *block,
 
         flushLayer();
         FailureOr<int64_t> innerDepth =
-            computeBlockWorstCaseDepth(&region.front(), onlyOnDisjointQubit, skipDynamic);
+            worstCaseDepthOfBlock(&region.front(), /*liftForLoops=*/false);
```
jzaia18, line 193: Why is `liftForLoops` always set to false here rather than using the same value that was input to this function?

mlir/lib/PBC/Utils/PBCLayer.cpp
lazypanda10117: It is a little weird that select PPMs are not part of the PBCOpInterface, which likely indicates the interface is not properly capturing what a PBC op should be. I would suggest fixing that directly rather than a patch here, if it doesn't require not too much effort.

mlir/lib/PBC/Utils/PBCLayer.cpp
sengthai: The premise doesn't hold and we can't skip it. so when `skipDynamic_` is true, the dynamic loop returns 0, but a static loop still falls through to `worstCaseDepthOfBlock(forOp.getBody(), …)` so even in the skip-dynamic mode we must resolve the trip count to tell those two cases aparts.

mlir/lib/PBC/Utils/PBCLayer.cpp
sengthai: it would touch alot files, and let's put that into another PR.

mlir/lib/PBC/Utils/PBCLayer.cpp
sengthai: it is intentinoal. `liftForLoops` is only meant to apply at the current block being reported, typically the top-level function body or a lifted loop body. Once the code recurses into a nested region, it wants that region’s real local depth, so it calls `worstCaseDepthOfBlock(..., false)`.

mlir/lib/Catalyst/Transforms/ResourceAnalysisPass.cpp
jzaia18:
```suggestion
        // Errors arise when attempting to compute depth of dynamic loops.
```

jzaia18 (approved):
Looks good, nice work

mlir/include/Catalyst/Analysis/ResourceAnalysis.h
```diff
@@ -48,6 +48,12 @@ class ResourceAnalysis {
 
     const ResourceResult *getFlattenedResource(llvm::StringRef funcName) const;
 
+    // Maps synthetic loop names to their `scf::ForOp` for depth analysis in the pass.
```
dime10, line 51: What is a synthetic loop?

sengthai: synthetic loop names are auto-generated labels like `for_loop_N`, or `dyn_for_loop_N`.

PR #2969, opened 2026-06-19
add translation and lowering of OperatorOp

albi3ro:
**Context:**

Now that `Operator2` has been added to Pennylane and can be captured into plxpr, it's time to lower it to MLIR.

**Benefits:**

**Possible Drawbacks:**

Had to switch op_name from being a `StrProp` to being a `StrAttr`.  Seems to work now.

**Related GitHub Issues:**

[sc-121978] [sc-121484]


mlir/lib/Quantum/IR/QuantumOps.cpp
```diff
@@ -899,7 +899,7 @@ ParseResult OperatorOp::parse(OpAsmParser &parser, OperationState &result)
         return failure();
     }
     auto &opProperties = result.getOrAddProperties<OperatorOp::Properties>();
-    opProperties.setOpName(opName);
```
kipawaa, line 902: I think we can similarly remove these lines in [value-semantics-conversion](https://github.com/PennyLaneAI/catalyst/blob/229b1174eeb2be7293bf9331056005c7429c65e6/mlir/lib/QRef/Transforms/value_semantics_conversion.cpp#L1162) and [reference-semantics-conversion](https://github.com/PennyLaneAI/catalyst/blob/229b1174eeb2be7293bf9331056005c7429c65e6/mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp#L371). If we switch `adjoint` and `UID` to attributes as well, they should get the same treatment :) (I think we'll need to, the generated python bindings don't do anything with those parameters 😓 )

albi3ro: Thanks! removed just the op name in those spots.

frontend/test/lit/test_operator.py
```diff
@@ -0,0 +1,156 @@
...
+        super().__init__(wires=wires)
+
+
+@qp.qjit(target="mlir", capture=True)
+@qp.qnode(qp.device("null.qubit", wires=2))
+def c_no_params():
+    # CHECK: [[q0:%.+]] = qref.get {{%.+}}
+    # CHECK: qref.operator "NoParams"() qubits([[q0]])
+    # CHECK static_data = {}
+    # CHECK param_map = {} qubit_map = {wires = [0]}
```
kipawaa:
```suggestion
    # CHECK: static_data = {}
    # CHECK: param_map = {} qubit_map = {wires = [0]}
```

frontend/test/pytest/test_operator.py
```diff
@@ -0,0 +1,76 @@
...
+
+    with pytest.raises(ValueError, match="must occur inside a qnode."):
+
+        @qp.qjit(capture=True, target="jaxpr")
+        def f():
+            DummyOp(2)
+            return 2
+
+
+def test_hybrid_not_supported_yet():
```
kipawaa, line 29: A pattern I've seen is to write a test that we would like to pass in the future, and then `xfail` with a "not supported" type message. Not a requirement, but I think it's a nice way to indicate unsupported behaviour that will be added.

kipawaa (approved):
Looks great, thanks @albi3ro !

doc/requirements.txt
```diff
@@ -34,4 +34,4 @@ lxml_html_clean
 --extra-index-url https://test.pypi.org/simple/
 pennylane-lightning-kokkos==0.46.0-dev10
 pennylane-lightning==0.46.0-dev10
-pennylane==0.46.0.dev24
+pennylane==0.46.0.dev38
```
mudit2812, line 37: This should be bumped again

.dep-versions
```diff
@@ -8,7 +8,7 @@ enzyme=v0.0.238
 
 # For a custom PL version, update the package version here and at
 # 'doc/requirements.txt'
-pennylane=0.46.0.dev24
+pennylane=0.46.0.dev38
```
mudit2812, line 11: This should be bumped again

mlir/include/Quantum/IR/QuantumOps.td
```diff
@@ -613,7 +613,7 @@ def OperatorOp : UnitaryGate_Op<"operator", [ParametrizedGate, NoMemoryEffect,
     }];
 
     let arguments = (ins
-        StringProp:$op_name,
+        StrAttr:$op_name,
```
mudit2812, line 616: So only string properties are causing issue? All the other properties are fine? 🤔

mlir/lib/Quantum/IR/QuantumOps.cpp
mudit2812: We can make the changes to adjoint and UID as needed. We're not supporting them in the lowering yet

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -408,6 +409,11 @@ def _handle_decompose_transform(self, inner_jaxpr, consts, non_const_args, tkwar
     return next_eval.eval(inner_jaxpr, consts, *non_const_args)
 
 
+@WorkflowInterpreter.register_primitive(operator_p)
+def _error_on_operator(self, *args, op_cls, **kwargs):
+    raise ValueError(f"Operator {op_cls} must occur inside a qnode.")
```
mudit2812: Would be nice to support this in the future too. For example, in the case where we create an operator to get its matrix or to reuse it in multiple QNodes. That would be a new feature though, so not needed right now.

frontend/catalyst/from_plxpr/qfunc_interpreter.py
```diff
@@ -30,6 +30,7 @@
 from pennylane.capture.primitives import measure_prim as plxpr_measure_prim
 from pennylane.capture.primitives import pauli_measure_prim as plxpr_pauli_measure_prim
 from pennylane.capture.primitives import quantum_subroutine_prim, transform_prim
+from pennylane.core.operator.operator2 import operator_p
```
mudit2812, line 33: In the PR where I'm implementing, I'm also making `operator_p` visible from `qp.capture.primitives`, so we can update this after that change is made.
```suggestion
# TODO: Change import path for operator_p after it is available in qp.capture.primitives
from pennylane.core.operator.operator2 import operator_p
```

frontend/test/lit/test_operator.py
```diff
@@ -0,0 +1,156 @@
...
+@qp.qjit(target="mlir", capture=True)
+@qp.qnode(qp.device("null.qubit", wires=3))
+def c_single_param(x: float):
+
+    # CHECK: [[q0:%.+]] = qref.get {{%.+}}
+
+    # CHECK: qref.operator "SingleParam"({{%.+}}: tensor<f64>) qubits([[q0]])
+    # CHECK: static_data = {}
+    # CHECK: param_map = {x = [0]} qubit_map = {wires = [0]}
+    SingleParam(x, 0)
```
mudit2812, line 68: Why is the input a `tensor<f64>` instead of just an `f64`?

mudit2812: Oh it's because for `CustomOp` we explicitly add operations to extract a scalar from a 0-D tensor. It might be worth doing the same thing here. If any input operands are 0-D tensors, we extract the scalar from it and give _that_ as an input.

mudit2812 (commented):
Looks great Christina, just one blocking comment regarding how we're lowering scalar parameters of `qref.operator`

mlir/include/Quantum/IR/QuantumOps.td
mudit2812: I guess we don't know yet because adjoint and UID are not part of the lowering yet

frontend/catalyst/from_plxpr/qref_jax_primitives.py
```diff
@@ -164,6 +168,7 @@ class MeasurementPlane(Enum):
 qref_compbasis_p = Primitive("qref_compbasis")
 qref_namedobs_p = Primitive("qref_namedobs")
 qref_hermitian_p = Primitive("qref_hermitian")
+qref_operator_op = Primitive("qref_operator")
```
mudit2812:
```suggestion
qref_operator_p = Primitive("qref_operator")
```

frontend/catalyst/from_plxpr/qref_jax_primitives.py
```diff
@@ -798,6 +803,61 @@ def _qref_named_obs_lowering(jax_ctx: mlir.LoweringRuleContext, qubit: ir.Value,
     return NamedObsOp(result_type, qubit, obsId).results
 
 
+qref_operator_op.multiple_results = True
```
mudit2812:
```suggestion
qref_operator_p.multiple_results = True
```

frontend/catalyst/from_plxpr/qref_jax_primitives.py
```diff
@@ -798,6 +803,61 @@ def _qref_named_obs_lowering(jax_ctx: mlir.LoweringRuleContext, qubit: ir.Value,
...
+
+
+@qref_operator_op.def_abstract_eval
+def _qref_operator_op_abstract_eval(*args, **kwargs):
+    return []
+
+
+def _operator_op_lowering(
```
mudit2812:
```suggestion
@qref_operator_p.def_abstract_eval
def _qref_operator_p_abstract_eval(*args, **kwargs):
    return []


def _operator_p_lowering(
```

doc/requirements.txt
albi3ro: What's needed to bump it? Seems to be working on this version fine.

mudit2812: Not a blocker. I'll bump the version in my PR for adj/ctrl lowering anyway.

frontend/test/lit/test_operator.py
mudit2812: Had a chat offline. The plan is to discuss the necessity of this on Slack and handle it in a follow up if any changes are needed

mudit2812 (approved):

dime10:
I'd like to have a look over the changes before merging :) Is it correct that this is only part of the op that is supported? @albi3ro your goal is to a support in stages across multiple prs?

albi3ro:
> I'd like to have a look over the changes before merging :) Is it correct that this is only part of the op that is supported? @albi3ro your goal is to a support in stages across multiple prs?

So #2979 adds specialized lowerings for things that fit CustomOp and all the other specialized ops.  In https://github.com/PennyLaneAI/pennylane/pull/9729 and https://github.com/PennyLaneAI/pennylane/pull/9730 Mudit is adding specialized capturing for adjoint and controlled. Following those PR's, we will be able to add lowerings for adjoint and ctrl.  Then we will have a follow up for things with python-only data.

mlir/include/Quantum/IR/QuantumOps.td
albi3ro: We will also need to update hanlding of adjoint and UID when we start providing them, I think.

frontend/catalyst/from_plxpr/qfunc_interpreter.py
albi3ro: Generally skeptical of adding TODO's. I think it will be available soon enough I don't think we will forget.

frontend/catalyst/from_plxpr/qref_jax_primitives.py
albi3ro: changed locally, so I could change it everywhere together.

frontend/catalyst/from_plxpr/qref_jax_primitives.py
albi3ro: same comment.

frontend/catalyst/from_plxpr/qref_jax_primitives.py
albi3ro: same comment again.

frontend/test/lit/test_operator.py
dime10: Yeah I think for the generic op leaving the 0-D tensors might be simpler. A bit of an annoying difference w.r.t. to JAX, but scalar tensors are a huge resource waste when lowered to CPU, so it's definitely better to have scalars eventually.

doc/requirements.txt
dime10: Yeah generally these are bumped as needed, not with every new release on main :)

mlir/include/Quantum/IR/QuantumOps.td
dime10: Let me double check if there are any downstream effects from this since there were subtle differences in the APIs between attributes and props.

doc/releases/changelog-dev.md
```diff
@@ -5,6 +5,10 @@
 
 <h3>Improvements 🛠</h3>
 
+* The new `pennylane.core.Operator2` can now be lowered to MLIR with program capture for operators
+  without non-lowerable arguments.
```
dime10, line 9: These will probably all be pulled into single entry at sound point around operator2, but okay for now.

frontend/catalyst/from_plxpr/from_plxpr.py
dime10: Would such an operator A) have a primitive (i guess yes whenever capture is enabled)? and B) have its primitive evaluated (interpreted)?
To compute matrices and reuse across QNodes, we should already have the PL operator instance available anyway, right?

frontend/catalyst/from_plxpr/qref_jax_primitives.py
```diff
@@ -798,6 +803,61 @@ def _qref_named_obs_lowering(jax_ctx: mlir.LoweringRuleContext, qubit: ir.Value,
...
+    wire_lens,
+    **static_data,
+):
+    params = args[: len(op_cls.dynamic_argnames)]
+    qubits = args[len(op_cls.dynamic_argnames) :]
+
+    name_attr = get_mlir_attribute_from_pyval(op_cls.__name__)
+
+    repack_static_data = {k: unflatten(*v) for k, v in static_data.items()}
+    processed_static_data = get_mlir_attribute_from_pyval(repack_static_data)
```
dime10, line 829: I wonder how we want to handle error handling / verification about what kind of data is supported to be lowered like this.

frontend/catalyst/from_plxpr/qref_jax_primitives.py
```diff
@@ -798,6 +803,61 @@ def _qref_named_obs_lowering(jax_ctx: mlir.LoweringRuleContext, qubit: ir.Value,
...
+
+    param_map = {
+        name: ir.DenseI64ArrayAttr.get([ind]) for ind, name in enumerate(op_cls.dynamic_argnames)
```
dime10, line 832: Just to confirm, the `*args` are passed in the same order as the `dynamic_argnames`?

frontend/catalyst/from_plxpr/qref_jax_primitives.py
```diff
@@ -798,6 +803,61 @@ def _qref_named_obs_lowering(jax_ctx: mlir.LoweringRuleContext, qubit: ir.Value,
...
+        qubit_map[name] = ir.DenseI64ArrayAttr.get(list(range(ind, ind + size)))
+        ind += size
+
+    processed_qubit_map = get_mlir_attribute_from_pyval(qubit_map)
+
+    OperatorOp(
+        op_name=name_attr,
+        params=params,
+        qubits=qubits,
+        forward_args=[],
```
dime10, line 849: We might want to add the register arg here for completeness (setting it to None so it's explicit we're not using it).

dime10 (approved):
Thanks 🎉

frontend/catalyst/from_plxpr/qref_jax_primitives.py
albi3ro: Yep. Right here: https://github.com/PennyLaneAI/pennylane/blob/ab728f1929e54a96a0201979776e70a08cd580fb/pennylane/core/operator/operator2.py#L1195

doc/releases/changelog-dev.md
albi3ro: 👍

frontend/catalyst/from_plxpr/from_plxpr.py
albi3ro:
```
@qp.qjit(capture=True)
def forgot_my_qnode():
    qp.X(0)
    return qp.state()
```
for example.

dime10: But what do we do with such a function? I don't think that's one of the scenarios Mudit was describing 🤔

mudit2812: I was describing something along the lines of:

```python
@qp.qjit(capture=True)
def workflow():
    op = qp.X(0)

    @qp.qnode(dev)
    def f(mat):
        # Do something with mat
        ...
        # Support for qp.apply with program capture is being added right now
        qp.apply(op)
        ...
        return qp.state()

    res = f(op.matrix())
    # Do something with res
    return something
```

frontend/catalyst/from_plxpr/qfunc_interpreter.py
mudit2812: Sure. I added the change in my PR

frontend/catalyst/from_plxpr/from_plxpr.py
dime10: Nice thanks! In this case, `op` is a PL `Operator2` instance, so it can be freely used in PL functions. The generated equation should actually be removed upon use (`qp.apply(op)`) during the tracing process. When converting to catalyst / mlir, that equation would then already be gone right?

mudit2812: @dime10 `qp.apply` would be the one case where we don't delete the old equation. `apply` is not a primitive that is consuming operators as data, it treats the operator as an instruction that is needs to insert in the program.

I think we can implement some blanket logic in `from_plxpr` that just doesn't bind a new primitive if there are operators outside qnodes. So we would still have an equation in plxpr, but not in catalyst jaxpr. I'm not sure how to do this more elegantly though if we want to get rid of `from_plxpr` eventually.

dime10: > apply is not a primitive that is consuming operators as data

That's true, but a good way to look at it imo is that it's just another function that generates an operator / instruction, like `qp.adjoint`, except that it is the "identity" function. This treats it the same as another PL operator functions and gives more predictable behaviour. Perhaps better revisited elsewhere though.

> I think we can implement some blanket logic in from_plxpr that just doesn't bind a new primitive if there are operators outside qnodes. So we would still have an equation in plxpr, but not in catalyst jaxpr.

Makes sense, and is pretty much trivial 👍 

> I'm not sure how to do this more elegantly though if we want to get rid of from_plxpr eventually.

Hmm, regardless of what we do, at some point a lowering step will have to find a qubit value for it, and there won't be one (outside of a qnode/device/prior allocation), and we could ignore it there, but I definitely think it's better served via the first point.

frontend/test/pytest/test_operator.py
albi3ro: Comes down to personal preference, but I like it this way. That way we can write much more extensive tests when we actually have the feature added.

frontend/catalyst/from_plxpr/from_plxpr.py
albi3ro: I'll just drop these lines from this PR for now.

frontend/test/lit/test_operator.py
albi3ro: Let's come back to this later when more things are fleshed out.

frontend/catalyst/from_plxpr/qref_jax_primitives.py
albi3ro: added.

frontend/catalyst/from_plxpr/qref_jax_primitives.py
albi3ro: Am wondering if this helper utility could be fleshed out a bit more, but think it gets the job done for the limited cases we are looking at for the moment.

mlir/include/Quantum/IR/QuantumOps.td
albi3ro: Things seem to be working here, so I'm just going to merge for now so we don't block the follow on work.

PR #2970, opened 2026-06-19
Rework the `catalyst` CLI installation mechanism

dime10:
In order to better support dynamically linked or loaded libraries from the Catalyst executable (the CLI tool), we need to guarantee a predictable location. Currently, the binary is installed via the `data_files` mechanism which is not the most predictable (see current `get_cli_path` impl), and separates the CLI binary from other package libraries (typically under `site-packages/catalyst/lib`).

For dynamic linking of shared libs, we'd need the .so to be located on a fixed path relative to the executable, since this path needs to be set at build time (i.e. `RPATH`). For dynamically loaded libraries (dlopen), the fixed relative path removes the need for heuristics to compute the real path of a shared lib.

The new system leverages the [console_scripts entry point](https://packaging.python.org/en/latest/specifications/entry-points/) mechanism to install a tiny shim in a location automatically available on PATH for most Python installs. 
This is common practice among CLI tools distributed as a Python package (e.g. cmake) and comes with a few advantages:
- `catalyst` is available as a command for both wheel installations and editable installs (the current mechanism only works for wheel installs)
- predictable/fixed path of the CLI within the Catalyst package
- simpler `get_cli_path` implementation (fewer heuristics)
 
[sc-122013]

**Note on performance:**

The overhead of running the CLI through the shim in the first version was about 100x. While only a couple of seconds total (on an empty file / pipeline), when used interactively this is clearly noticeable and off-putting. The new structure avoids importing Catalyst except the necessary pieces to locate the binary. The final result is about 10x slower than when invoked directly, but this is fast enough to not be a burden (~0.3s).

### DIFF
--- a/frontend/catalyst/utils/runtime_environment.py
+++ b/frontend/catalyst/utils/runtime_environment.py
@@ -72,25 +72,27 @@ def get_libpython_path() -> str:  # pragma: no cover
 
 def get_lib_path(project, env_var):
     """Get the path to Catalyst's shared libraries."""
-    if INSTALLED:
-        return os.path.join(package_root, "lib")  # pragma: no cover
+    if INSTALLED:  # pragma: no cover
+        return os.path.join(package_root, "lib")
+
     return os.getenv(env_var, DEFAULT_LIB_PATHS.get(project, ""))
 
 
 def get_include_path():
     """Return the path to Catalyst's include directory."""
-    if INSTALLED:
-        return os.path.join(package_root, "include")  # pragma: no cover
+    if INSTALLED:  # pragma: no cover
+        return os.path.join(package_root, "include")
+
     return os.getenv("CATALYST_INCLUDE_DIRS", DEFAULT_INCLUDE_PATHS.get("mlir", ""))
 
 
-def get_cli_path() -> str:  # pragma: nocover
+def get_cli_path() -> str:
     """Method to obtain the Catalyst CLI path whether installed or locally built."""
     catalyst_cli = "catalyst"
 
-    if not INSTALLED:
-        return os.path.join(
-            os.getenv("CATALYST_BIN_DIR", DEFAULT_BIN_PATHS.get("cli", "")), catalyst_cli
-        )
+    if INSTALLED:  # pragma: nocover
+        return os.path.join(package_root, "bin", catalyst_cli)
 
-    return os.path.join(package_root, "bin", catalyst_cli)
+    return os.path.join(
+        os.getenv("CATALYST_BIN_DIR", DEFAULT_BIN_PATHS.get("cli", "")), catalyst_cli
+    )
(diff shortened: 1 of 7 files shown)


frontend/catalyst_cli_shim.py
```diff
@@ -0,0 +1,59 @@
...
+
+def _get_cli_path() -> str:
+    """Resolve the Catalyst CLI path without importing Catalyst via the usual machinery.
+
+    Relies on the runtime_environment module not importing Catalyst directly either.
+    """
+    spec = importlib.util.find_spec("catalyst")
+    if spec is None or spec.origin is None:
+        raise ImportError(
+            "Could not locate the 'catalyst' package, please make sure it is installed."
```
kipawaa: Should we provide an install hint here, ex. `via "pip install pennylane-catalyst"`?

frontend/catalyst/utils/runtime_environment.py
```diff
@@ -73,49 +86,27 @@ def get_libpython_path() -> str:  # pragma: no cover
...
-        return os.path.join(package_root, "include")  # pragma: no cover
+    if INSTALLED:  # pragma: no cover
+        return os.path.join(package_root, "include")
+
     return os.getenv("CATALYST_INCLUDE_DIRS", DEFAULT_INCLUDE_PATHS.get("mlir", ""))
 
 
-def get_cli_path() -> str:  # pragma: nocover
-    """Method to obtain the Catalyst CLI path packaged via the data_files mechanism."""
+def get_cli_path() -> str:
```
kipawaa, line 103: This is so much nicer 😄

kipawaa (approved):
This is such a great QoL upgrade, thanks @dime10 !

frontend/catalyst/utils/runtime_environment.py
dime10: Truly 😅

frontend/catalyst_cli_shim.py
dime10: Hmm, I would consider that an installation instruction, which we usually want to keep in a centralized placed. I'm pretty sure any Python user also knows how to pip install a package 😅 But what we could do is link to the install guide, why not :)

dime10:
```suggestion
            "Could not locate the 'catalyst' package, please make sure it is installed. "
            "Refer to https://docs.pennylane.ai/projects/catalyst/en/stable/dev/installation.html "
            "for additional instructions."
```

dime10: https://github.com/PennyLaneAI/catalyst/pull/2970/commits/42850085305f268db0380e34f085c72a23b62b11

lazypanda10117 (approved):
Thanks @dime10!

sengthai (approved):

PR #2973, opened 2026-06-23
Add support for choosing decomposition rules in `decompose-lowering`

kipawaa:
**Context:**
`decompose-lowering` always applies all (applicable) decomposition rules in the IR, and removes all decomposition rules on cleanup.

Introduces a `target-rules` option to apply only a subset of the rules available in the IR. Rules are now applied by aligning the `target_op`'s operands with the decomposition functions body and cloning the function body inline, to avoid the inconsistencies/heuristics of the built-in inliner. Rules are no longer removed after the `decompose-lowering` pass to allow use throughout the pipeline in `graph-decomposition` passes. Decomposition functions will be removed by the `symbol-dce` pass at the end of the default pipeline.

**Benefits:**
Provides a more versatile `decompose-lowering` pass, and will support a more efficient `graph-decomposition` pass.

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-122013]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -5,6 +5,13 @@
 
 <h3>Improvements 🛠</h3>
 
+* The `decompose-lowering` pass now supports applying a selection of the available decomposition rules via the `target_rules` parameter.
+  The pass also no longer applies the `inline`, `cse` and `canonicalize` passes to avoid unnecessary IR mutations.
+  Instead, decomposition rules are deterministically inlined by a custom function (`inline` is non-deterministic, using an estimated benefit and threshold as criteria for inlining).
+  Decomposition rules are no longer removed after the `decompose-lowering` pass, which allows them to be used by subsequent passes, namely `graph-decomposition`.
+  Instead, rules are removed by the `symbol-dce` pass at the end of the `QuantumCompilationStage`.
+  [(#2973)](https://github.com/PennyLaneAI/catalyst/pull/2973)
+
 * The `ResourceAnalysis` pass now reports each loop body and each subroutine as its own entry
   instead of folding their gate counts into the caller. Loops with constant bounds appear as `for_loop_<N>`
   with their trip count. Loops with dynamic bounds appear as `dyn_for_loop_<N>` with a stable
--- a/mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
+++ b/mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
@@ -57,6 +57,7 @@ namespace quantum {
 static SmallVector<Value> inlineRuleBody(PatternRewriter &rewriter, func::FuncOp rule,
                                          ValueRange operands)
 {
+    assert(rule.getBlocks().size() == 1)
     Block &body = rule.front();
     auto returnOp = cast<func::ReturnOp>(body.getTerminator());
 
--- a/mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
+++ b/mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
@@ -40,16 +40,16 @@
 
 #include "DecomposeLoweringImpl.hpp"
 
+#define DEBUG_TYPE "decompose-lowering"
+
 using namespace mlir;
 using namespace catalyst::quantum;
-
-#define DEBUG_TYPE "decompose-lowering"
 
 namespace catalyst {
 namespace quantum {
 
-SmallVector<Value> inlineDecompositionRule(func::FuncOp rule, ValueRange operands,
-                                           PatternRewriter &rewriter)
+SmallVector<Value> getDecompRuleResults(func::FuncOp rule, ValueRange operands,
+                                        PatternRewriter &rewriter)
 {
     Block &body = rule.front();
     auto returnOp = cast<func::ReturnOp>(body.getTerminator());
@@ -103,6 +103,8 @@ struct DLCustomOpPattern : public OpRewritePattern<CustomOp> {
             return failure();
         }
 
+        // do not nest decomposition rules, they're applied greedily and this can lead to
+        // cycles/identity rules
         if (isInDecompRule(op)) {
             return failure();
         }
@@ -144,17 +146,17 @@ struct DLCustomOpPattern : public OpRewritePattern<CustomOp> {
         assert(analyzer && "Analyzer should be valid");
 
         auto operands = analyzer.prepareOperands(rule, rewriter, op.getLoc());
-        SmallVector<Value> inlinedFunction = inlineDecompositionRule(rule, operands, rewriter);
+        SmallVector<Value> inlinedFunctionResults = getDecompRuleResults(rule, operands, rewriter);
 
         // Replace the op with the inlined function and adjust the insert ops for the qreg mode
-        if (inlinedFunction.size() == 1 &&
-            isa<quantum::QuregType>(inlinedFunction.front().getType())) {
-            auto results =
-                analyzer.prepareResultsForQreg(inlinedFunction.front(), op.getLoc(), rewriter);
+        if (inlinedFunctionResults.size() == 1 &&
+            isa<quantum::QuregType>(inlinedFunctionResults.front().getType())) {
+            auto results = analyzer.prepareResultsForQreg(inlinedFunctionResults.front(),
+                                                          op.getLoc(), rewriter);
             rewriter.replaceOp(op, results);
         }
         else {
-            rewriter.replaceOp(op, inlinedFunction);
+            rewriter.replaceOp(op, inlinedFunctionResults);
         }
 
         return success();
@@ -183,6 +185,8 @@ struct DLMultiRZOpPattern : public OpRewritePattern<MultiRZOp> {
             return failure();
         }
 
+        // do not nest decomposition rules, they're applied greedily and this can lead to
+        // cycles/identity rules
         if (isInDecompRule(op)) {
             return failure();
         }
@@ -223,18 +227,18 @@ struct DLMultiRZOpPattern : public OpRewritePattern<MultiRZOp> {
         assert(analyzer && "Analyzer should be valid");
 
         auto operands = analyzer.prepareOperands(decompFunc, rewriter, op.getLoc());
-        SmallVector<Value> inlinedFunction =
-            inlineDecompositionRule(decompFunc, operands, rewriter);
+        SmallVector<Value> inlinedFunctionResults =
+            getDecompRuleResults(decompFunc, operands, rewriter);
 
         // Replace the op with the inlined function and adjust the insert ops for the qreg mode
-        if (inlinedFunction.size() == 1 &&
-            isa<quantum::QuregType>(inlinedFunction.front().getType())) {
-            auto results =
-                analyzer.prepareResultsForQreg(inlinedFunction.front(), op.getLoc(), rewriter);
+        if (inlinedFunctionResults.size() == 1 &&
+            isa<quantum::QuregType>(inlinedFunctionResults.front().getType())) {
+            auto results = analyzer.prepareResultsForQreg(inlinedFunctionResults.front(),
+                                                          op.getLoc(), rewriter);
             rewriter.replaceOp(op, results);
         }
         else {
-            rewriter.replaceOp(op, inlinedFunction);
+            rewriter.replaceOp(op, inlinedFunctionResults);
         }
 
         return success();
@@ -263,6 +267,8 @@ struct DLPauliRotOpPattern : public OpRewritePattern<PauliRotOp> {
             return failure();
         }
 
+        // do not nest decomposition rules, they're applied greedily and this can lead to
+        // cycles/identity rules
         if (isInDecompRule(op)) {
             return failure();
         }
@@ -289,18 +295,18 @@ struct DLPauliRotOpPattern : public OpRewritePattern<PauliRotOp> {
         assert(analyzer && "Analyzer should be valid");
 
         auto operands = analyzer.prepareOperands(decompFunc, rewriter, op.getLoc());
-        SmallVector<Value> inlinedFunction =
-            inlineDecompositionRule(decompFunc, operands, rewriter);
+        SmallVector<Value> inlinedFunctionResults =
+            getDecompRuleResults(decompFunc, operands, rewriter);
 
         // Replace the op with the inlined results and adjust the insert ops for the qreg mode
-        if (inlinedFunction.size() == 1 &&
-            isa<quantum::QuregType>(inlinedFunction.front().getType())) {
-            auto results =
-                analyzer.prepareResultsForQreg(inlinedFunction.front(), op.getLoc(), rewriter);
+        if (inlinedFunctionResults.size() == 1 &&
+            isa<quantum::QuregType>(inlinedFunctionResults.front().getType())) {
+            auto results = analyzer.prepareResultsForQreg(inlinedFunctionResults.front(),
+                                                          op.getLoc(), rewriter);
             rewriter.replaceOp(op, results);
         }
         else {
-            rewriter.replaceOp(op, inlinedFunction);
+            rewriter.replaceOp(op, inlinedFunctionResults);
         }
 
         return success();
--- a/mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
+++ b/mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
@@ -54,7 +54,7 @@ namespace quantum {
  * replace the parameters of `rule` and returning the results. `rewriter`'s insertion point will be
  * moved to the end of the inlined function body.
  */
-SmallVector<Value> inlineRuleBody(PatternRewriter &rewriter, func::FuncOp rule, ValueRange operands)
+static SmallVector<Value> inlineRuleBody(PatternRewriter &rewriter, func::FuncOp rule, ValueRange operands)
 {
     Block &body = rule.front();
     auto returnOp = cast<func::ReturnOp>(body.getTerminator());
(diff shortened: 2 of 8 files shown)


mlir/lib/Quantum/Transforms/decompose_lowering.cpp
```diff
@@ -12,34 +12,47 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#define DEBUG_TYPE "decompose-lowering"
+#include <cstdint>
+#include <string>
+#include <utility>
 
-// When we read the decomposition rules module from file,
```
kipawaa, line 17: moved to the `stablehlo` include where this is relevant

mlir/lib/Quantum/Transforms/decompose_lowering.cpp
```diff
@@ -150,82 +169,33 @@ struct DecomposeLoweringPass : impl::DecomposeLoweringPassBase<DecomposeLowering
                                             f.front().getTerminator()->getOperandTypes()));
     }
 
-    // Remove unused decomposition functions:
-    // Since the decomposition functions are marked as public from the frontend,
-    // there is no way to remove them with any DCE pass automatically.
-    // So we need to manually remove them from the module
```
paul0403, line 156: Are the rule functions private now? If so, where are they marked private?

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
```diff
@@ -135,9 +157,12 @@ struct GraphDecompositionPass : public impl::GraphDecompositionPassBase<GraphDec
...
 
         ///////////////////////////
-        // Step 5: Re-introduce (all) user rules for future decompositions
+        // Step 5: Re-introduce any missing user rules for future decompositions
+        SymbolTable symbolTable(module);
         for (auto &rule : allUserRules) {
-            module.getBody()->push_back(rule.release());
+            if (!symbolTable.lookup<func::FuncOp>(rule->getName())) {
+                module.getBody()->push_back(rule.release());
+            }
```
paul0403, line 165: This change is independent of the main change?

mlir/lib/Quantum/Transforms/decompose_lowering.cpp
```diff
@@ -150,82 +169,33 @@ struct DecomposeLoweringPass : impl::DecomposeLoweringPassBase<DecomposeLowering
...
             return;
         }
 
-        // Step 1.1: Find the target gate set
+        // Step 2: Find the target gate set
         findTargetGateSet(module, targetGateSet);
 
-        // Step 2: Canonicalize the module
-        RewritePatternSet patternsCanonicalization(&getContext());
-        catalyst::quantum::CustomOp::getCanonicalizationPatterns(patternsCanonicalization,
```
paul0403, line 196: Why is the custom op canonicalization removed?

mlir/lib/Quantum/Transforms/decompose_lowering.cpp
```diff
@@ -96,9 +109,15 @@ struct DecomposeLoweringPass : impl::DecomposeLoweringPassBase<DecomposeLowering
...
     void discoverAndRegisterDecompositions(ModuleOp module,
-                                           llvm::StringMap<func::FuncOp> &decompositionRegistry)
+                                           llvm::StringMap<func::FuncOp> &decompositionRegistry,
+                                           llvm::StringSet<> targetRules)
     {
         module.walk([&](func::FuncOp func) {
+            // if targetRules is provided, only add requested rules
+            if (!targetRules.empty() && !targetRules.contains(func.getName())) {
+                return WalkResult::skip();
+            }
```
paul0403, line 119: Just confirming, this is essentially skipping all rule functions that are not explicitly requested in the `targetRules` command line option right?

mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
```diff
@@ -12,26 +12,75 @@
...
+                                        PatternRewriter &rewriter)
+{
+    Block &body = rule.front();
+    auto returnOp = cast<func::ReturnOp>(body.getTerminator());
+
+    IRMapping mapping;
+    mapping.map(body.getArguments(), operands);
+
+    for (Operation &op : body.without_terminator()) {
+        rewriter.clone(op, mapping);
```
paul0403, line 68: We should add a insertion guard in this scope (i.e. the scope of the `getDecompRuleResults` function) and set a insertion point, if there's any new ops being written to the IR

paul0403, line 68: Also, I assume this is part of the new "manual inlining" the rule func into the gate "callsite", instead of generating a call op and delagating the actual inline to the mlir inliner? In that case we should give this helper function a name that reflects this action 🤔 The current name sounds like a getter, and doesn't imply any IR edits

mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
```diff
@@ -205,6 +267,12 @@ struct DLPauliRotOpPattern : public OpRewritePattern<PauliRotOp> {
             return failure();
         }
 
+        // do not nest decomposition rules, they're applied greedily and this can lead to
+        // cycles/identity rules
+        if (isInDecompRule(op)) {
+            return failure();
+        }
```
paul0403, line 280: It's a bit unfortunate that there's 3 pattern classes for custom. multirz and paulirot, but oh well 😅

mlir/test/Quantum/DecomposeLoweringTest.mlir
```diff
@@ -219,92 +227,60 @@ module @param_rxry {
     return %out_qubits_1 : !quantum.bit
   }
 }
-// -----
-
-// Test parametric gates and wires
-module @param_rxry_2 {
```
paul0403, line 225: Why are parametrized gates and wires not supported anymore?

mlir/test/Quantum/DecomposeLoweringTest.mlir
```diff
@@ -219,92 +227,60 @@ module @param_rxry {
...
-      // CHECK:   [[EXTRACT2:%.+]] = quantum.extract [[INSERT1]][[[EXTRACTED_3]]] : !quantum.reg -> !quantum.bit
-      // CHECK:   [[INSERT2:%.+]] = quantum.insert [[REG1]][[[EXTRACTED_3]]], [[EXTRACT2]] : !quantum.reg, !quantum.bit
-      // CHECK:   [[EXTRACT3:%.+]] = quantum.extract [[INSERT2]][[[EXTRACTED_4]]] : !quantum.reg -> !quantum.bit
-      // CHECK:   [[RZ2:%.+]] = quantum.custom "RZ"([[CST]]) [[EXTRACT3]] : !quantum.bit
-      // CHECK:   [[INSERT3:%.+]] = quantum.insert [[INSERT2]][[[EXTRACTED_4]]], [[RZ2]] : !quantum.reg, !quantum.bit
-      // CHECK:   [[EXTRACT4:%.+]] = quantum.extract [[INSERT3]][[[EXTRACTED_3]]] : !quantum.reg -> !quantum.bit
-      // CHECK:   [[INSERT4:%.+]] = quantum.insert [[INSERT2]][[[EXTRACTED_3]]], [[EXTRACT4]] : !quantum.reg, !quantum.bit
-      // CHECK:   scf.yield [[INSERT4]] : !quantum.reg
+
+      // CHECK: [[q0:%.+]] = quantum.extract [[reg0]][ 0] : !quantum.reg -> !quantum.bit
```
paul0403, line 247: The extra tensor.extract ops and stablehlo tensor manipulation ops here in this test are because we no longer run cse?

mlir/lib/Quantum/Transforms/decompose_lowering.cpp
kipawaa: They are marked private by the frontend as of https://github.com/PennyLaneAI/catalyst/pull/2658

mlir/lib/Quantum/Transforms/decompose_lowering.cpp
kipawaa: Exactly

mlir/test/Quantum/DecomposeLoweringTest.mlir
kipawaa: This was a duplicate test

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
kipawaa: This is required because `decompose-lowering` no longer removes all of the decomposition rules. This means that the current `graph-decomposition` pass will introduce duplicate symbols after `decompose-lowering` runs.

mlir/lib/Quantum/Transforms/decompose_lowering.cpp
kipawaa: I don't think this was necessary, and in general we probably shouldn't run it if it's not needed. I could re-introduce it if you see cause. I removed it initially because my decomposition rules were disappearing so I removed all of the extra passes/patterns.

paul0403: > I removed it initially because my decomposition rules were disappearing

Hmm, sounds a bit concerning, we don't have any canonicalization patterns on gates that are just plain removals. 

Or is it because the extract/insert canonicalizations are removing them and the gate is left with no users, and then they are removed since they are `NoMemoryEffect`? This would make sense since the insert ops at the end of the decomp rule functions wouldn;t have any users for their result registers?

kipawaa: That's a very plausible explanation. I just bulk removed all of the extra passes/patterns, so I can't say for sure which was causing me issues 😅

mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
kipawaa: I wasn't feeling inspired enough to try to add that to this PR. I think once the `DecomposableGate` interface is implemented, we can maybe rewrite `decompose-lowering` to work specifically on that interface, rather than any particular op types!

mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
kipawaa: The insertion point needs to be after whichever ops the `analyzer.prepareOperands` generates. I don't think it makes sense to guard/change insertion point in this helper since it's _intended_ to move the insertion point. I've documented this and updated the name per your other comment in [0c3b7c7](https://github.com/PennyLaneAI/catalyst/pull/2973/commits/0c3b7c760dc59f984196b85030f3e919b6afe564).

kipawaa: Good point, thanks! Updated in [0c3b7c7](https://github.com/PennyLaneAI/catalyst/pull/2973/commits/0c3b7c760dc59f984196b85030f3e919b6afe564)

mlir/test/Quantum/DecomposeLoweringTest.mlir
kipawaa: I believe so. It adds some unnecessary complications to the `decompose-lowering` output, but these will be removed later by the `cse` run as part of the default pipeline anyways.

mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
paul0403: Ah, if it's intended then all good.

paul0403 (approved):
Nice 💯

doc/releases/changelog-dev.md
```diff
@@ -5,6 +5,13 @@
 
 <h3>Improvements 🛠</h3>
 
+* The `decompose-lowering` pass now supports applying a selection of the available decomposition rules via the `target_rules` parameter.
+  The pass also no longer applies the `inline`, `cse` and `canonicalize` passes to avoid unnecessary IR mutations.
+  Instead, decomposition rules are deterministically inlined by a custom function (`inline` is non-deterministic, using an estimated benefit and threshold as criteria for inlining).
+  Decomposition rules are no longer removed after the `decompose-lowering` pass, which allows them to be used by subsequent passes, namely `graph-decomposition`.
+  Instead, rules are removed by the `symbol-dce` pass at the end of the `QuantumCompilationStage`.
```
maliasadi, line 12: The visibility of rules is still `public` (we set it in the frontend) and `symbol-dce` doesn't remove public symbols. Now, `cse` is removed from the pipeline, any plans to change the visibility to `private`?

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
```diff
@@ -135,9 +157,12 @@ struct GraphDecompositionPass : public impl::GraphDecompositionPassBase<GraphDec
         }
 
         ///////////////////////////
-        // Step 5: Re-introduce (all) user rules for future decompositions
+        // Step 5: Re-introduce any missing user rules for future decompositions
+        SymbolTable symbolTable(module);
```
maliasadi, line 161: Nice use of `SymbolTable`!

mlir/test/Quantum/DecomposeLoweringTargetRulesTest.mlir
maliasadi, line 1: Could you (update the frontend and) test that `symbol-dce` actually works end-to-end without leaking rules?

mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
```diff
@@ -12,26 +12,80 @@
...
+    }
+
+    SmallVector<Value> results;
+    for (Value operand : returnOp.getOperands()) {
+        results.push_back(mapping.lookupOrDefault(operand));
+    }
+    return results;
+}
+
+bool isInDecompRule(Operation *op)
```
maliasadi: as this is only used in this file
```suggestion
static bool isInDecompRule(Operation *op)
```

mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
```diff
@@ -12,26 +12,80 @@
...
 namespace catalyst {
 namespace quantum {
 
+/**
+ * @brief
+ * Inline the body of `rule` at `rewriter`'s current insertion point, using `operands` to
+ * replace the parameters of `rule` and returning the results. `rewriter`'s insertion point will be
+ * moved to the end of the inlined function body.
+ */
+SmallVector<Value> inlineRuleBody(PatternRewriter &rewriter, func::FuncOp rule, ValueRange operands)
```
maliasadi: it's only used in this file -- let's keep it static! 
```suggestion
static SmallVector<Value> inlineRuleBody(PatternRewriter &rewriter, func::FuncOp rule, ValueRange operands)
```

mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
```diff
@@ -12,26 +12,80 @@
...
+{
+    Block &body = rule.front();
```
maliasadi, line 61: Any concerns with using `front()` for structured multi-block rules? Does it drop the IR after the first block? 🤔

maliasadi: Asking from Gemini, we can add an assert here to fail ^:

```suggestion
    assert(rule.getBlocks().size() == 1)
    Block &body = rule.front();
```

maliasadi (commented):
Nice refactor @kipawaa!

dime10:
Maybe a general design question, but is there a use case to having a separate decompose-lowering pass that is distinct from a graph solver pass (assuming that is the current design)?

Naively I would have thought there is one "decompose" pass, with the target gate set, and the pass computes whatever is necessary and replaces your gates until all are in the target.
Of course this doesn't mean the graph and the rewrite can't be separate components within the pass.

Alternatively, if the pass separation is in fact useful, one could perhaps leverage an analysis result from the first pass in the second, but it depends on how they are intended to be used.
A target-rules pass option doesn't really make too much sense to me though, as that is a relatively static configuration.

kipawaa:
> is there a use case to having a separate decompose-lowering pass that is distinct from a graph solver pass (assuming that is the current design)?

I think that these two should be combined in the near future, but until the graph is able to accommodate _all_ decomposition requirements we'll need the separation. I've been working towards this goal of unifying the decomposition passes, and hopefully this can be complete in the next couple of months.

dime10:
> > is there a use case to having a separate decompose-lowering pass that is distinct from a graph solver pass (assuming that is the current design)?
> 
> I think that these two should be combined in the near future, but until the graph is able to accommodate _all_ decomposition requirements we'll need the separation. I've been working towards this goal of unifying the decomposition passes, and hopefully this can be complete in the next couple of months.

What does the separation do for us while we don't have all requirements?

doc/releases/changelog-dev.md
kipawaa: @maliasadi #2658 moved decomposition rules to private, so we should be okay to use this :)

mlir/test/Quantum/DecomposeLoweringTargetRulesTest.mlir
kipawaa: A test for this was added [here](https://github.com/PennyLaneAI/catalyst/pull/2658/changes#diff-0f5507ff6a68593e9e594c12dea5ccfaa47299edb0533c6df60a7e8ee637fcd5R298) in #2658

mlir/lib/Quantum/Transforms/DecomposeLoweringPatterns.cpp
kipawaa: Good point, I hadn't considered that we might find a function with multiple blocks! We'll use the assert for now, and we can implement better handling if the need arises.

kipawaa:
> > > is there a use case to having a separate decompose-lowering pass that is distinct from a graph solver pass (assuming that is the current design)?
> > 
> > 
> > I think that these two should be combined in the near future, but until the graph is able to accommodate _all_ decomposition requirements we'll need the separation. I've been working towards this goal of unifying the decomposition passes, and hopefully this can be complete in the next couple of months.
> 
> What does the separation do for us while we don't have all requirements?

Chatted offline, seems likely I was mistaken and that we can move up the timeline on combining these passes. I'll evaluate and make necessary changes in a follow-up PR.

maliasadi (approved):
🦅

PR #2974, opened 2026-06-24
Drop Support for Python 3.11

JerryChen97:
**Context:**
Support for Python 3.11 has been gradually dropped across the entire ecosys.

For more detailed context why we drop, see https://github.com/PennyLaneAI/pennylane/pull/9700 description

**Related:**
https://github.com/PennyLaneAI/pennylane-lightning/pull/1395

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-124857]


JerryChen97:
Why `cuda-quantum==0.6.0`?

.github/workflows/build-wheel-linux-arm64.yaml
```diff
@@ -292,7 +292,7 @@ jobs:
       uses: actions/cache/restore@v4
       with:
         path: ${{ github.workspace }}/llvm-build
-        key: ${{ matrix.container_img }}-llvm-${{ needs.constants.outputs.llvm_version }}-3.11-wheel-build
+        key: ${{ matrix.container_img }}-llvm-${{ needs.constants.outputs.llvm_version }}-3.12-wheel-build
```
dime10: How is the generation of the cache updated accordingly?

JerryChen97:
Let's figure out the impact over cache first

.github/workflows/build-wheel-linux-arm64.yaml
mehrdad2m: The cache generation is already using the constant `matrix.python_version`. Does it make sense to use the same constant here as well? https://github.com/PennyLaneAI/catalyst/pull/2974/changes#diff-1d070300e8a4097dfeb320744e6a01cbd8baba474d7424e57aa85a2f3b5dd013R205

dime10: seems reasonable

JerryChen97: https://github.com/PennyLaneAI/catalyst/pull/2974/commits/1bb36aad28de404f5b0b5dc58bd5534e12aa4359

mehrdad2m (approved):
Thanks @JerryChen97, Let me know once you ready to merge and I can delete the old cache entries right after you merge.

JerryChen97:
> Thanks @JerryChen97, Let me know once you ready to merge and I can delete the old cache entries right after you merge.

Nice. I see people are still using CI. Is the deletion gonna affect a lot? How long typically does it take to recover to normal status after we merge

mehrdad2m:
> Nice. I see people are still using CI. Is the deletion gonna affect a lot? How long typically does it take to recover to normal status after we merge

I don't think so. those cache entries are for wheel builds. The reason not deleting them could cause problems is not that they are used by CI, it is that hitting the cache limit may evict other artifacts that are used by CI. So hopefully as long as no one is building wheels on their PRs, clearing them should not affect PR CI.

dime10 (approved):

PR #2975, opened 2026-06-24
improve mlir & mlir_opt property documentation

kipawaa:
**Context:**
The `QJIT.mlir` property is relatively poorly documented. The return type is unclear and unspecified, and the docstring provides little clarification.

Adds type hints and updates docstring.

**Benefits:**
clearer `mlir` and `mlir_opt` properties.

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-123089]


isaacdevlugt (changes requested):
Just needs a changelog entry :heh: 

Otherwise, nice 🚀

doc/releases/changelog-dev.md
```diff
@@ -351,6 +351,8 @@
 * A broken link was removed in the [Compiler Core](https://docs.pennylane.ai/projects/catalyst/en/stable/modules/mlir.html) documentation page. The link referred to where precompiled decomposition rules were implemented, which has since been refactored.
   [(#2913)](https://github.com/PennyLaneAI/catalyst/pull/2913)
 
+* The documentation for `QJIT.mlir` and `QJIT.mlir_opt` was updated with type hints and docstrings that better reflect the compilation-dependent nature of the properties.
```
isaacdevlugt, line 426:
```suggestion
* The documentation for `QJIT.mlir` and `QJIT.mlir_opt` was updated with type hints and docstrings that better reflect the compilation-dependent nature of the properties.
  [(#2975)](https://github.com/PennyLaneAI/catalyst/pull/2975)
```

isaacdevlugt (approved):
🚀

frontend/catalyst/jit.py
```diff
@@ -626,8 +631,13 @@ def mlir(self):
...
 
     @property
-    def mlir_opt(self):
-        """Obtain the MLIR representation after optimization"""
+    def mlir_opt(self) -> str | None:
+        """The MLIR representation of the QJIT object after optimization, if available.
+
+        Returns:
+            str: The textual MLIR module after applying user passes and the default pipeline.
+            None: If compilation has not yet occurred.
```
dime10: This statement is a bit tricky. Technically, accessing this property triggers the "compilation" (at least compilation in the MLIR / core Catalyst compiler sense), but what needed to have happened before is that the `capture` and `generate_ir` stages of the QJIT object needed to have been run, whether via AOT compilation, JIT compilation, or some other way.

kipawaa: Good point! It's kind of difficult to convey all of that concisely, what about something like `None: If MLIR cannot be generated for the circuit.` and we can explain further in the core of the docstring?

dime10: Yeah that makes sense! Maybe a slight modification, since the MLIR having been generated is the criteria, and our programs are more general than circuits:
`None: If the MLIR has not yet been generated for this program`

kipawaa: How does [f145021](https://github.com/PennyLaneAI/catalyst/pull/2975/commits/f14502167bf6e1077e74494b31c3d1345e9aaae4) sound?

dime10: > If optimized MLIR has not yet been generated for the program.

The condition for both properties is the same actually, because they take the jax-generated mlir (`.mlir_module`) as input to produce their output.

kipawaa: I guess I was trying to get at the required success of running the requested pipeline to generate `.mlir_opt`, whereas `.mlir_module` is sufficient for `.mlir`.

dime10: If there is a failure in the pipeline an error should be raised. The `None` result is produced precisely when the input IR isn't available.

kipawaa: Makes sense, added in [a98893b](https://github.com/PennyLaneAI/catalyst/pull/2975/commits/a98893b3357d0e79433362b84b49c6709b8a3f2b)

comp-phys-marc (approved):
Thanks so much!

PR #2977, opened 2026-06-25
Graph decomp rule improvements

kipawaa:
**Context:**
The `graph-decomposition` pass manipulates the decomposition rules in the IR in a number of ways, often duplicating work.

- Uses the improved `decompose-lowering` pass (#2973) to apply exclusively the solver-chosen rules.
- No longer removes user-rules from the IR
- No longer requires re-adding user rules to the IR after `decompose-lowering`
- No longer requires re-adding the builtin rules for each `graph-decomposition` pass

**Benefits:**
A much more efficient and simpler `graph-decomposition` pass.

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-122013]


mlir/lib/Quantum/Transforms/graph_decomposition.cpp
```diff
@@ -139,31 +138,24 @@ struct GraphDecompositionPass : public impl::GraphDecompositionPassBase<GraphDec
                                  std::move(altDecomps));
         DecompositionSolver solver(graph);
         auto solution = solver.solve();
-        ///////////////////////////
-        // Step 3: Insert decomposition rules picked by the graph solver (solution) into the
-        // module
-        insertChosenRules(solution, ruleNameToFuncOp);
```
maliasadi, line 145: 👋

doc/releases/changelog-dev.md
```diff
@@ -205,6 +205,9 @@
 
 <h3>Internal changes ⚙️</h3>
 
+* The `graph-decomposition` pass now performs far less IR manipulation.
```
```suggestion
* The `graph-decomposition` pass eliminates three redundant IR manipulations:
  the cloning, removal, and re-insertion of user rules. This optimization is particularly
  beneficial when the pass is executed multiple times within the compilation pipeline.
```

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
```diff
@@ -455,84 +516,22 @@ struct GraphDecompositionPass : public impl::GraphDecompositionPassBase<GraphDec
...
-     * @param solution The chosen decomposition rules from the graph solver.
-     * @param ruleNameToFuncOp A mapping from rule names to their corresponding function
-     * operations.
-     */
-    void insertChosenRules(GraphResult &solution,
-                           llvm::StringMap<mlir::OwningOpRef<func::FuncOp>> &ruleNameToFuncOp)
-    {
-        mlir::ModuleOp module = getOperation();
-        for (const auto &[_, chosenRule] : solution) {
-            if (chosenRule.isBasis) {
```
maliasadi, line 522: Hm.. Are you dropping the `isBasis` guard?

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
```diff
@@ -360,20 +407,34 @@ struct GraphDecompositionPass : public impl::GraphDecompositionPassBase<GraphDec
...
-
-            ruleRegistry.push_back(std::move(outOp));
+            LDBG() << "adding rule " << funcOp.getName();
+            module.push_back(std::move(outOp.release()));
         }
-
         return success();
     }
 
+    bool isInDecompRule(Operation *op)
```
maliasadi, line 430: duplicated with https://github.com/PennyLaneAI/catalyst/pull/2973/changes#diff-b330fef4c89c48fcc2802d9ec894c6866269829d616879a2547a5fca6eeb6229R76 (PR #2973)

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
```diff
@@ -236,66 +228,111 @@ struct GraphDecompositionPass : public impl::GraphDecompositionPassBase<GraphDec
...
+        if (!resourcesAttr) {
+            ResourceAnalysis analysis(rule);
+            if (const ResourceResult *flat = analysis.getFlattenedResource(rule.getName())) {
+                rule->setAttr("resources", buildResourceDict(&getContext(), *flat));
+            }
+            resourcesAttr = rule->getAttrOfType<DictionaryAttr>("resources");
+        }
+
+        // 2. Extract 'operations' dictionary from resources
+        auto operations = mlir::dyn_cast_or_null<DictionaryAttr>(resourcesAttr.get("operations"));
```
maliasadi, line 263: if the prev step fails, I think we still need an explicit failure here:

```suggestion
        if (!resourcesAttr) {
            return failure();
        }
        auto operations = mlir::dyn_cast_or_null<DictionaryAttr>(resourcesAttr.get("operations"));
```

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
```diff
@@ -455,84 +516,22 @@ struct GraphDecompositionPass : public impl::GraphDecompositionPassBase<GraphDec
...
-        }
-        if (failed(loadUserDecompositionRules(userRuleNames, graphRules, userRules))) {
-            return signalPassFailure();
-        }
-
-        for (auto &ruleOpRef : graphRules) {
-            mlir::func::FuncOp func = ruleOpRef.get();
-            llvm::StringRef ruleName = func.getName();
+        // Load pre-compiled rules (ignore failure, we can try to solve without)
+        std::ignore = loadBuiltInDecompositionRules(filename, rules);
```
maliasadi, line 534: could we change the failures from `loadBuiltInDecompositionRules` to make them non-fatal instead of treating all decomp failures as fatal `failure()`?

maliasadi (commented):
Thanks @kipawaa!

mlir/test/Quantum/GraphDecomposition/TestAltDecomps.mlir
maliasadi, line 1: Will you add tests for rule retention and basis-rule filtering?

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
```diff
@@ -360,20 +407,34 @@ struct GraphDecompositionPass : public impl::GraphDecompositionPassBase<GraphDec
...
-            auto analysis = ResourceAnalysis(funcOp);
-            if (const ResourceResult *flat = analysis.getFlattenedResource(funcOp.getName())) {
-                funcOp->setAttr("resources", buildResourceDict(context, *flat));
+            if (failed(addRuleNode(funcOp, ruleNodes))) {
+                return failure();
             }
-
-            ruleRegistry.push_back(std::move(outOp));
+            LDBG() << "adding rule " << funcOp.getName();
+            module.push_back(std::move(outOp.release()));
```
maliasadi: it's not needed here, `release` is already a raw ptr:
```suggestion
            module.push_back(outOp.release());
```

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
```diff
@@ -305,19 +342,29 @@ struct GraphDecompositionPass : public impl::GraphDecompositionPassBase<GraphDec
...
         mlir::ModuleOp module = getOperation();
         MLIRContext *context = &getContext();
 
         llvm::StringSet<> addedWords;
+        // Add words from existing paulirot rules
+        module.walk([&](mlir::func::FuncOp func) {
+            if (func->hasAttr("target_gate")) {
+                if (func.getName().starts_with("paulirot_decomp_rule_")) {
+                    addedWords.insert(func.getName().drop_front(21));
+                }
```
maliasadi, line 367: TODO: generalize the system and remove these magic str & number 🙂

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
kipawaa: I didn't really want to include that just to use it here, do you have any suggestions for cleaning this up?

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
kipawaa: `LogicalResult` is non-fatal, it's effectively a fancy `bool` with `[[nodiscard]]`. I think that the `std::ignore` here nicely demonstrates that we are aware that the result of `loadBuiltInDecompositionRules` is important, and that we are consciously choosing to continue regardless of the outcome. Not to be confused with `signalPassFailure()`, which _can be_ non-fatal, but generally will stop the pass manager.

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
kipawaa: Can you explain? I think since we are no longer inserting rules we don't need the basis rule logic, but I'm not sure I understand your meaning.

mlir/test/Quantum/GraphDecomposition/TestAltDecomps.mlir
kipawaa: I add the `CHECK-LABEL` directives to the decomposition rules to assert they exist after the pass is run, i.e. that the rules are successfully retained despite use in decomposition. If you think there should be a more dedicated test for this I'm happy to add one.

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
kipawaa: added in [73afab2](https://github.com/PennyLaneAI/catalyst/pull/2977/commits/73afab23c6b44e2f928f1a399a700539f26f71ec), this will be addressed in #2983 😄

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
kipawaa: Thanks! Added in [73afab2](https://github.com/PennyLaneAI/catalyst/pull/2977/commits/73afab23c6b44e2f928f1a399a700539f26f71ec) to include an error message.

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
```diff
@@ -139,31 +138,24 @@ struct GraphDecompositionPass : public impl::GraphDecompositionPassBase<GraphDec
                                  std::move(altDecomps));
         DecompositionSolver solver(graph);
         auto solution = solver.solve();
-        ///////////////////////////
-        // Step 3: Insert decomposition rules picked by the graph solver (solution) into the
```
paul0403, line 143: I trust you guys on this, but why do we no longer need to insert the rules picked by the graph solver into the module?

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
```diff
@@ -139,31 +138,24 @@ struct GraphDecompositionPass : public impl::GraphDecompositionPassBase<GraphDec
...
-        // module
-        insertChosenRules(solution, ruleNameToFuncOp);
 
         ///////////////////////////
         // Step 4: Convert python-decompositions from reference to value semantics and run
```
paul0403: I guess this is step 3 now

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
```diff
@@ -139,31 +138,24 @@ struct GraphDecompositionPass : public impl::GraphDecompositionPassBase<GraphDec
...
         // decompose-lowering to apply the chosen decomposition rules
         ModuleOp module = getOperation();
         OpPassManager pm("builtin.module");
+
+        DecomposeLoweringPassOptions dlOptions;
+        for (auto &[op, chosenRule] : solution) {
+            dlOptions.targetRulesOption.push_back(chosenRule.ruleName);
+        }
```
paul0403, line 151: Oh! We don't need to add the rules above because now we are adding them in through the new `target-rules` option you added to decomp lowering pass?

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
```diff
@@ -236,88 +228,153 @@ struct GraphDecompositionPass : public impl::GraphDecompositionPassBase<GraphDec
...
+            resourcesAttr = rule->getAttrOfType<DictionaryAttr>("resources");
+        }
+
+        // Fail if resources are missing
+        if (!resourcesAttr) {
+            llvm::errs() << "Decomposition rule " << ruleName
+                         << " was provided without resources, and resources could not be generated "
+                            "for it.\n";
+            return failure();
+        }
```
paul0403, line 260: I think it would make more sense to put this into the previous `if(!resourceAttr)` scope since they are logically together

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
```diff
@@ -236,88 +228,153 @@ struct GraphDecompositionPass : public impl::GraphDecompositionPassBase<GraphDec
...
 
         for (auto rule : llvm::make_early_inc_range(moduleOp.get().getOps<mlir::func::FuncOp>())) {
-            rule->remove();
-            ruleRegistry.push_back(std::move(rule));
+            if (failed(addRuleNode(rule, ruleNodes))) {
+                return failure();
+            }
+            // avoid double-insertion
+            if (!symbolTable.lookup<mlir::func::FuncOp>(rule.getName())) {
+                rule->remove();
```
paul0403, line 312: Why the `remove()` here? The purpose is to add a rule to the module only if it doesn't exist in the module already right?

Is it to detach the rule function from its original module?

paul0403: The naming is a bit confusing, want to rename `moduleOp` to `libModule` and `module` to `payloadModule`? 😅

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
paul0403: We can put all the graph passes (decomp lowering, graph decomp, graph solver) in its own folder under `lib/Quantum/Transforms`, just like with the adjoint lowering. This way the graph decomp utils can live in its own utils file in that folder. 

We can also do this in another PR if it's too big, but I suspect it's ok.

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
kipawaa: They should already be there! https://github.com/PennyLaneAI/catalyst/pull/2973 allowed us to choose which rules are applied by `decompose-lowering`, which used to apply all of the rules. Because we only wanted to apply the chosen rules, we had to remove all of them, add the ones we want applied, then re-add the rest. Now we can leave them in the IR 😄

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
maliasadi: Yeah, we can add all the graph work in one folder. Do you want me to create a PR while you're working on the interface?

mlir/test/Quantum/GraphDecomposition/TestAltDecomps.mlir
maliasadi: Nice! I must have missed that in my first review 👍

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
kipawaa: That would be great @maliasadi, thanks!

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
kipawaa: my feeling was that having it as a nested else in the previous if statement was a bit harder to read, but I don't mind switching it back

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
kipawaa: yep! The remove is to release the op from the builtin rules module, so that we can add it to the user module (it can only belong to one).

I also renamed the rules module to `builtinModule` in [a5e03da](https://github.com/PennyLaneAI/catalyst/pull/2977/commits/a5e03dadd1f02f97f2990441b68bc799a127a83f)

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
kipawaa: [a5e03da](https://github.com/PennyLaneAI/catalyst/pull/2977/commits/a5e03dadd1f02f97f2990441b68bc799a127a83f)

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
paul0403: It's fine, up to you here, doesn't really matter tbh

paul0403 (approved):

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
paul0403: So for this PR let's keep it as is, just to get it in and make the git-ing easier. In your PR @maliasadi remember to de-duplicate this `isInDecompRuleFunc` helper with a util!

mlir/lib/Quantum/Transforms/graph_decomposition.cpp
maliasadi: Oh yes! we're not inserting rules. we forward `ruleName`s into `targetRulesOption` and decomp-lowering uses them purely as a filter set :+1:

maliasadi (approved):

PR #2978, opened 2026-06-26
Fix `catalyst_decompose` with `capabilities=None` and state prep

dwierichs:
**Context:**
We use `catalyst_decompose` with `capabilities=None` for our workflows, currently.
If the leading op of the tape (or sub-tape/region within a program) begins with a `BasisState` or `StatePrep` operation, `catalyst_decompose` tries to access `capabilities.initial_state_prep`, which errors out for `capabilities=None`. We just want `skip_initial_state_prep` to remain `False` in this scenario.

Move the described clause to the part of the setup that already asserts that `capabilities is not None`.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


lazypanda10117 (approved):

dime10 (approved):
Great, thanks David!

PR #2981, opened 2026-06-29
Add support for lowering operators with Python-only data

mudit2812:
**Context:**
Follow up for lowering `Operator2` to MLIR. This PR adds support for lowering operators with Python-only data (hybrid and static non-compilable arguments)

* Add `uid.py` containing function `generate_uid` for creating a UID for operators with hybrid/static arguments
* Add `_process_params` and `_process_qubits` helper functions to partition operands into numerical parameters and qubits, as well as create their respective maps
  * `_process_params` also partitions numerical data into operator parameters and feed-forward data. To do this, I had to implement a function for walking over the hybrid pytrees. This was necessary because unflattening operators wasn't an option as the leaves would be MLIR SSA values. This can be improved by handling some of this processing during capture time, when values are tracers, not MLIR SSA values or abstract types.
* Update the MLIR operations to have UID be an attribute instead of a property
* Update MLIR verifiers and printers to correctly handle the case where there are no qubits; previously we were raising an error if there were no qubits and no input qreg.

**Benefits:**

**Possible Drawbacks:**
No drawbacks, but the implementation is quite complex.

**Related GitHub Issues:**
Currently blocked by https://github.com/PennyLaneAI/pennylane/pull/9808

[sc-123232]


frontend/catalyst/from_plxpr/uid.py
```diff
@@ -0,0 +1,144 @@
...
+):
+    """Generate a unique identifier that allows us to distinguish between
+    operators with unique non-compilable arguments."""
+    reduced = []
+
+    # Flat dynamic arguments
+    dynamic_args = avals_in[: len(op_cls.dynamic_argnames)]
+    dynamic_avals = []
+    for val in dynamic_args:
+        dynamic_avals.append((val.shape, val.dtype.name))
```
albi3ro:
```suggestion
    dynamic_avals = [(val.shape, val.dtype.name) for val in dynamic_args]
```
purely stylistic.

frontend/catalyst/from_plxpr/uid.py
```diff
@@ -0,0 +1,142 @@
...
+    wire_lens: tuple[int, ...],
+    hybrid_lens: tuple[int, ...],
+    hybrid_trees: tuple[PyTreeStructure, ...],
+    adjoint: bool,
+    n_ctrls: int,
+    static_args: dict[str, Any],
+):
+    """Generate a unique identifier that allows us to distinguish between
+    operators with unique non-compilable arguments."""
+    reduced = []
```
albi3ro: Just to make it a little easier to read, could we initialize `reduced` right before we start putting things into it? Then I don't have to keep it in "working memory" when reading the first few setup steps.

*nitpick*

mudit2812: Good idea. I'm also in the process of adding some more dev comments. Hopefully, that will make the review a bit easier

frontend/test/lit/test_operator.py
```diff
@@ -38,14 +38,14 @@ def c_no_params():
     # CHECK: [[q0:%.+]] = qref.get {{%.+}}[ 0]
     # CHECK: qref.operator "NoParams"() qubits([[q0]])
     # CHECK: static_data = {}
-    # CHECK: param_map = {} qubit_map = {reg = [0]}
+    # CHECK: qubit_map = {reg = [0]}
```
albi3ro, line 41: Where did param_map go?

mudit2812: I updated the implementation so that if there are no parameters, we don't add `param_map`. The MLIR attribute is a default-valued optional attribute (default value is `{}`), so it doesn't create any problems. I did the same thing for `qubit_map` as well

frontend/catalyst/from_plxpr/qref_operator2_primitives.py
```diff
@@ -89,27 +91,125 @@ def _is_custom_op(op_cls, avals_in):
...
-
+    assert all(_is_qref_qubit(w) for w in wires)
+
+
+def _process_params(
+    *args, op_cls, wire_lens, hybrid_lens, forward_mask
+) -> tuple[list, list, dict[str, list[int]]]:
+    """Process non-qubit operands of an operator. This function returns the flattened sequence
+    of qubit operands of the operator, feed-through arguments of any operator arguments, and a
+    dictionary mapping argument names to the indices of their respective qubits.
```
albi3ro: how do non-qubit operands have qubits? Did you mean to write "respective qubits" there?

mudit2812: Good catch, silly mistake on my end

frontend/test/lit/test_operator.py
albi3ro: So what change did this?

I'm especially confused now since why was it displayed to begin with if it was a DefaultValuedOptionalAttr? Aren't those supposed to be skipped in the display?

mlir/lib/QRef/IR/QRefOps.cpp
```diff
@@ -230,8 +230,8 @@ LogicalResult OperatorOp::verify()
     const bool hasQregMode = static_cast<bool>(getQreg());
     const bool hasQubitMode = !getQubits().empty();
 
-    // Exactly one mode must be used: explicit qubits or qreg-based addressing.
-    if (hasQregMode == hasQubitMode) {
+    // At most one mode must be used: explicit qubits or qreg-based addressing.
+    if (hasQregMode && hasQubitMode) {
```
albi3ro, line 234: 😕 What happens if we have neither mode? What's that?  Would that be like an identity or global phase?

mlir/lib/QRef/IR/QRefOps.cpp
```diff
@@ -477,13 +481,14 @@ void OperatorOp::print(OpAsmPrinter &p)
     }
 
     // 4. Qubits
-    if (!getQubits().empty()) {
+    if (!getQreg()) {
```
albi3ro, line 484: Is this for improved printing of operations on no qubits?

frontend/test/lit/test_operator.py
mudit2812: We were explicitly setting them in the lowering rule before. Now, I'm just setting them to `None` if they're empty.

frontend/catalyst/from_plxpr/qfunc_interpreter.py
```diff
@@ -292,20 +293,44 @@ def __call__(self, jaxpr, *args):
         return self.eval(jaxpr.jaxpr, jaxpr.consts, *args)
 
 
-# pylint: disable=too-many-arguments
-@PLxPRToQuantumJaxprInterpreter.register_primitive(operator_p)
-def _handle_operator(self, *args, op_cls, hybrid_lens, hybrid_trees, adjoint, n_ctrls, **kwargs):
+def _new_hybrid_arg(interp: PLxPRToQuantumJaxprInterpreter, arg) -> list:
+    """Create new flattened hybrid arguments by apping wire values to abstract qubits."""
```
paul0403:
```suggestion
    """Create new flattened hybrid arguments by appending wire values to abstract qubits."""
```

mlir/lib/QRef/IR/QRefOps.cpp
mudit2812: That is one example. In the previous iteration of this PR where the qubits of an operator argument were ending up in `forward_args`, this was causing errors, because if the only argument was an operator, there would be nothing in `in_qubits`, causing the verifier to fail. There are other identity-like operators that don't have any wires either, such as `Snapshot` and `Barrier`. Neither of them are technically Catalyst-friendly, but I don't think we should strictly disallow operators on no wires as a general strategy.

mlir/lib/QRef/IR/QRefOps.cpp
mudit2812: Yes. Before, the printing for an operator with no wires would be:
```
qref.operator "Dummy"(%p1 : f64)
    ...
```

Now, it would be
```
qref.operator "Dummy"(%p1 : f64) qubits()
    ...
```

If the former is preferred, I can revert the change

frontend/catalyst/from_plxpr/qfunc_interpreter.py
```diff
@@ -292,20 +293,44 @@ def __call__(self, jaxpr, *args):
         return self.eval(jaxpr.jaxpr, jaxpr.consts, *args)
 
 
-# pylint: disable=too-many-arguments
-@PLxPRToQuantumJaxprInterpreter.register_primitive(operator_p)
-def _handle_operator(self, *args, op_cls, hybrid_lens, hybrid_trees, adjoint, n_ctrls, **kwargs):
+def _new_hybrid_arg(interp: PLxPRToQuantumJaxprInterpreter, arg) -> list:
```
paul0403, line 296:
```suggestion
def _new_hybrid_arg(interp: PLxPRToQuantumJaxprInterpreter, args) -> list:
```
(and uses in the function)

frontend/test/lit/test_operator.py
```diff
@@ -522,3 +531,159 @@ def c_pcphase(x: float, dim: int):
...
+
+@qp.qjit(capture=True, target="mlir")
+@qp.qnode(qp.device("null.qubit", wires=1))
+def c_static_data():
+    # CHECK-LABEL: func.func public @c_static_data
+
+    # CHECK: [[q0:%.+]] = qref.get {{%.+}}[ 0]
+    # CHECK: qref.operator "StaticData"() qubits([[q0]])
+    # CHECK: UID({{[0-9]+}})
+    # CHECK-NOT: static_data
```
paul0403, line 553: Kinda unfortunate naming that `static_argnames` on the python operator2 class lowers to UID in mlir, and `compilable_argnames` lowers to `static_data`, i.e. `static_argnames` not lowering to `static_data` 😅 

No action needed since I think these are locked down now, just a somewhat unfortunate naming in my opinion

frontend/test/lit/test_operator.py
```diff
@@ -522,3 +531,159 @@ def c_pcphase(x: float, dim: int):
...
+@qp.qnode(qp.device("null.qubit", wires=2))
+def c_hybrid_wires():
+    # CHECK-LABEL: func.func public @c_hybrid_wires
+
+    # CHECK: [[q0:%.+]] = qref.get {{%.+}}[ 0]
+    # CHECK: [[q1:%.+]] = qref.get {{%.+}}[ 1]
+    # CHECK: qref.operator "HybridWires"() qubits([[q0]], [[q1]])
+    # CHECK: UID({{[0-9]+}})
+    # CHECK: qubit_map = {cwires = [0, 1]}
+    HybridWires(cwires=qp.wires.Wires([0, 1]))
```
paul0403: Where is this UID coming from? `HybridWires` class doesn't have a `static_argnames` right? Or is my understanding of what lowers to UID incorrect?

paul0403: Oh! Is the uncompilable static data the pytree structure (which is static, as opposed to the values in the pytree structure which is dynamic)?

frontend/test/lit/test_operator.py
```diff
@@ -522,3 +531,159 @@ def c_pcphase(x: float, dim: int):
...
+    op1 = MultiRZ(x, [0, 1, 2])
+    op2 = MultiRZ(x, [0, 1, 2])
+    op3 = MultiRZ(x, [0, 1])
+
+    # CHECK: [[q0:%.+]] = qref.get {{%.+}}[ 3]
+    # CHECK: [[q1:%.+]] = qref.get {{%.+}}[ 0]
+    # CHECK: [[q2:%.+]] = qref.get {{%.+}}[ 1]
+    # CHECK: [[q3:%.+]] = qref.get {{%.+}}[ 2]
+    # CHECK: qref.operator "HybridOpArg"({{%.+}}: tensor<f64>) qubits([[q0]], [[q1]], [[q2]], [[q3]])
+    # CHECK-NEXT: UID([[UID_A:.+]]) forward({{%.+}}: tensor<f64>)
```
paul0403:
```suggestion
    # CHECK: qref.operator "HybridOpArg"(%arg1: tensor<f64>) qubits([[q0]], [[q1]], [[q2]], [[q3]])
    # CHECK-NEXT: UID([[UID_A:.+]]) forward(%arg0: tensor<f64>)
```

I think explicitly checking the params are valuable here to make sure the distinction between the forward args and the non-forward args are correctly lowered. 

If there are `tensor.extract` ops, it's ok to include them in the lit test checks

frontend/test/lit/test_operator.py
```diff
@@ -522,3 +531,159 @@ def c_pcphase(x: float, dim: int):
...
+    # CHECK-NEXT: param_map = {angle = [0]} qubit_map = {cwires = [0], op = [1, 2, 3]}
+    HybridOpArg(y, op1, cwires=[3])
```
paul0403, line 658: Question: where is the following two pieces of info?
- The `n_iters` arg. This is `static_argnames`, so is it hashed into the UID?
- The fact that `op1` is a `MultiRZ`.

paul0403 (commented):
I want to review the tests first to make sure my understanding is correct before looking at the implementation. This is my questions/reviews for the test part. Great job!

frontend/test/lit/test_operator.py
mudit2812: Yes, exactly :)

frontend/test/lit/test_operator.py
mudit2812: Anything in `static_argnames` is assumed to be uncompilable, even if they actually are, so `n_iters` is part of the UID. The fact that `op1` is a `MultiRZ` is also part of the UID, since the pytree structure of hybrid arguments is used in the UID (which the `MultiRZ` should be a part of)

frontend/catalyst/from_plxpr/qfunc_interpreter.py
mudit2812:
```suggestion
    """Create new flattened hybrid arguments by mapping wire values to abstract qubits."""
```

frontend/catalyst/from_plxpr/qfunc_interpreter.py
mudit2812: The function only takes a single hybrid argument as input though

frontend/test/lit/test_operator.py
mudit2812: Totally fair!

mlir/test/Quantum/VerifierTest.mlir
```diff
@@ -777,14 +785,6 @@ func.func @operator_invalid_both_modes(%q : !quantum.bit, %r : !quantum.reg, %id
...
 // -----
 
-func.func @operator_invalid_no_mode() {
-    // expected-error@+1 {{must use either qubits or registers, but not both}}
-    "quantum.operator"() <{op_name = "bad_none", operandSegmentSizes = array<i32: 0, 0, 0, 0, 0, 0, 0, 0, 0>, resultSegmentSizes = array<i32: 0, 0, 0>}> : () -> ()
-    return
-}
-
-// -----
-
```
albi3ro, line 787: Should there be version of this test that actually does check for both?

frontend/catalyst/from_plxpr/qfunc_interpreter.py
paul0403: Ah yes, never mind, sorry I read the helper before its use

frontend/catalyst/from_plxpr/qref_operator2_primitives.py
```diff
@@ -89,28 +91,125 @@ def _is_custom_op(op_cls, avals_in):
...
+    and a dictionary mapping argument names to the indices of their respective values.
+    """
+    params = []
+    forward_params = []
+    param_map = {}
+
+    # Flat dynamic arguments
+    for i, dname in enumerate(op_cls.dynamic_argnames):
+        params.append(args[i])
+        param_map[dname] = ir.DenseI64ArrayAttr.get([i])
```
paul0403, line 123: What if we have a case like 
```python
class MultiParams(qp.core.Operator2):

    dynamic_argnames = ("c", "b", "a")
    wire_argnames = ("reg",)

    def __init__(self, reg, a, b, c):
        super().__init__(reg, a, b, c)
```

i.e. `op_cls.dynamic_argnames` does not go in the same order as the arguments? Is there a check on the operator2 class to guarantee that this will not happen?

frontend/catalyst/from_plxpr/qref_operator2_primitives.py
```diff
@@ -89,28 +91,125 @@ def _is_custom_op(op_cls, avals_in):
...
+                params += cur_params
+                param_map[hname] = ir.DenseI64ArrayAttr.get(
+                    list(range(map_idx, map_idx + len(cur_params)))
+                )
+                map_idx += len(cur_params)
+
+        mask_idx += hsize
+        args_idx += hsize
+
+    param_map = get_mlir_attribute_from_pyval(param_map) if param_map else None
```
paul0403, line 158: This is the `None` that made all the lit test updates (i.e. removal of empty `param_map`s) right?

frontend/test/pytest/from_plxpr/test_uid.py
```diff
@@ -0,0 +1,285 @@
...
+    wire_argnames = ()
+
+    # pylint: disable=useless-parent-delegation
+    def __init__(self, op, label=""):
+        super().__init__(op, label=label)
+
+
+class _Opaque:
+    """Opaque type for testing."""
+
```
paul0403, line 81:
```suggestion
class _Opaque:
    """Opaque type for testing."""
    pass
```

I would think pylint would bark here, but feel free to disregard if pylint doesn't flag these empty bodies

mlir/test/Quantum/VerifierTest.mlir
paul0403: Didn't we add the empty mode in this PR, for things like identity and global phase?

albi3ro (approved):
Fantastic work 🚀 !!!

Did struggle a little holding all the pieces together in my mind, but it all works and I can't see any other suggestions, so it has my full support and I very much look forward to being able to use this logic!

mlir/test/Quantum/VerifierTest.mlir
mudit2812: There is already a test for both modes

frontend/catalyst/from_plxpr/qref_operator2_primitives.py
mudit2812: Why do we want to disallow this? It's not typical, but it is valid. And we have tests to check that the lowering works correctly for such cases as well

frontend/catalyst/from_plxpr/qref_operator2_primitives.py
mudit2812: Yes

frontend/test/pytest/from_plxpr/test_uid.py
mudit2812: Once you add a docstring you don't need to add a `pass`. Same reason why you can have empty function bodies without a `return` without pylint complaining

frontend/catalyst/from_plxpr/qref_operator2_primitives.py
paul0403: Ah ok, if it's allowed and checked then all good :+1:

paul0403 (approved):

PR #2983, opened 2026-06-30
Generalize python-decompositions

kipawaa:
**Context:**
The [initial implementation of python-decompositions](https://github.com/PennyLaneAI/catalyst/pull/2769) was specialized to `quantum.paulirot`. This was done partly for simplicity, and partly due to a lack of consistency in op interfaces, which would have required bespoke solutions to parse dynamic data, static data etc for each quantum operation.

Introduces a generic `DecomposableGate` interface for interacting with ops in the context of decomposition. This provides consistent methods for accessing dynamic data, static data, op graph IDs and more. This PR introduces the interface as a PoC for `quantum.paulirot` and `quantum.operator`, providing tests as a demonstration of the intended functionality. The intended result of this implementation will be to remove any op-specific logic from the `graph-decomposition` pass, instead using exclusively methods from the interface to improve generality and improve extensibility to other ops without altering core logic in the graph.

**Benefits:**
Enables consistent interfaces across all op types relevant to decomposition, reduces type-checking boilerplate and code duplication. This allows any op implementing the interface to use python-decompositions, and further simplifies the general graph infrastructure.

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-122013]
[sc-122018]


mlir/include/Quantum/IR/QuantumInterfaces.td
```diff
@@ -179,6 +179,34 @@ def DifferentiableGate : OpInterface<"DifferentiableGate", [ParametrizedGate]> {
...
+            "mlir::DictionaryAttr", "getStaticData"
+        >,
+        InterfaceMethod<
+            "Return the operations ID for use with the graph.", "std::string", "getDecompId"
+        >,
+        InterfaceMethod<
+            "Return the name of the operation.", "std::string", "getOpName"
+        >,
+        InterfaceMethod<
+            "Return the number of wires the operation acts on.", "size_t", "getNumWires"
```
dime10: I guess this won't generalize to multiple wire args 🤔

mlir/include/Quantum/IR/QuantumInterfaces.td
```diff
@@ -179,6 +179,34 @@ def DifferentiableGate : OpInterface<"DifferentiableGate", [ParametrizedGate]> {
...
+        InterfaceMethod<
+            "Return the shape and dtype of dynamic data.",
+            "mlir::TypeRange", "getDynamicShape"
+        >,
+        InterfaceMethod<
+            "Return the values of static data.",
+            "mlir::DictionaryAttr", "getStaticData"
+        >,
+        InterfaceMethod<
+            "Return the operations ID for use with the graph.", "std::string", "getDecompId"
```
dime10: What is the ID? Is it the same as UID or some other concept?

dime10: Oh I see, for paulirot it's currently `"paulirot" + getPauliWord()`. Yeah I think that's the format that we should standardize on, but as a concept makes sense 👍 

Essentially we want a standard recipe to build a unique identifier for every op that enters the graph as a unique node.

dime10 (commented):
This is great thanks! Moving in a good direction :)

mlir/include/Quantum/IR/QuantumInterfaces.td
kipawaa: Can you explain this? I think that even when an op has multiple wire args, from a decomposition perspective we only need to know the total number of wires to lower the rule/properly assess resources.

mlir/include/Quantum/IR/QuantumInterfaces.td
kipawaa: Exactly, it's effectively a generalization of UID that includes static data, since the graph needs to distinguish all instantiations of an operator. I think that this is only relevant to the compiler graph, so standardization should be as simple as requiring the use of the interface when using/checking the decomp IDs!

mlir/include/Quantum/IR/QuantumInterfaces.td
dime10: I don't think that's true. If there was an operator that had multiple wire args that could be varied independently, the decomp rule would likely be specific the number of wires for each arg, not the total number (i.e. `Op(Wires[3], Wires[2])` and `Op(Wires[2], Wires[3])` are most likely different operators for decomposition purposes, or at least we cannot assert that they would always be the same). 
The best approach is to simply treat the wires as part of the signature (which they are). The full spec for an operator ID for decomp purposes should look something like this `(dyn arg sig, wire sig, static data OR uid)`. But this is why I think it would be good to align with the Python decomp folks on the exact form of the ID.

mlir/include/Quantum/IR/QuantumInterfaces.td
dime10: > Exactly, it's effectively a generalization of UID that includes static data, since the graph needs to distinguish all instantiations of an operator.

Yeah I think this is a good idea, one could just hash this tuple `(dyn arg sig, wire sig, static data OR uid)` to generate a single distinguishable operator ID to attach to the graph nodes.

frontend/catalyst/device/python_decompositions.py
maliasadi, line 1: Could we return a more structured format (to be used/parsed later at MLIR) rather than a pretty-printed IR string? Can we use Bytecode? it is versioned and far more robust than textual round-tripping here. It also avoids the name-prefix conventions in the string-version of rules.

mlir/lib/Quantum/Transforms/GraphDecomposition/graph_decomposition.cpp
maliasadi, line 1: We probably need to raise an error to guarantee the biggest constraint with the system which is if any rule for an op is present, all its rules need to be present. in other words, no support for user defined rules via `add_decomps` with python decomps.

frontend/catalyst/device/python_decompositions.py
```diff
@@ -42,27 +42,34 @@
...
 
     @qp.qjit(
         target="mlir",
         capture=True,
     )
     @qp.qnode(device=device)
     def circuit():
-        paulirot_subroutine(theta, wires)
+        for subroutine in subroutines:
+            subroutine(*[0.5 for _ in dynamic_shape], wires=wires)
```
maliasadi, line 73: Pls add a TODO for `0.5`. It's probably safe for all present ops/rules, but let's bring this up in the Op2 sync meeting to double-check with the rest of the team.

mlir/include/Quantum/IR/QuantumInterfaces.td
```diff
@@ -179,6 +179,39 @@ def DifferentiableGate : OpInterface<"DifferentiableGate", [ParametrizedGate]> {
     ];
 }
 
+def DecomposableGate : OpInterface<"DecomposableGate", [QuantumGate]> {
```
maliasadi, line 182: Is it also meant to eventually replace the special-casing in `getRuleNodes` or basically name-handling (`gphase` -> GlobalPhase, which already has a "TODO: replace this with `DecomposableGate` interface")? If so, sketching that end state now would clarify whether the current method set is sufficient for use cases.

maliasadi (commented):
Happy to see this PR come through 🙌

mlir/include/Quantum/IR/QuantumInterfaces.td
kipawaa: After chatting with @mudit2812 and @albi3ro it seems that any operator with a wire signature that is not a flat array of qubits will be lowered with a UID, so this shouldn't present any issues for graph-decomp

frontend/catalyst/device/python_decompositions.py
kipawaa: Discussed offline, the textual IR is reliable enough and saves the overhead of spawning a compiler process since we can use `QJIT.mlir_module`, so we'll keep as-is.

mlir/include/Quantum/IR/QuantumInterfaces.td
dime10: I don't think this is quite right in this context. I was talking about an operator with two wire arguments, both flat. Such an operator will not generate a UID (there is no python-specific data). Where the UID would come in is if there was *one* wire argument with a nested structure (tuple of two wire sequences).

mlir/include/Quantum/IR/QuantumInterfaces.td
dime10: Update from offline discussion. We settled on a consistent format across frontend/compiler for a graph node identifier. It's a string representation of the following pieces of information:
```
Op name
Dynamic argument signature (shapes & dtypes, in canonical form across frontend/compiler)
Qubit signature (shapes, i.e. length for each wire arg)
Static compilable data (in canonical string form), if present
UID for Python-specific data, if present
```

frontend/catalyst/device/python_decompositions.py
kipawaa: This is a placeholder for now, but I definitely agree that we should check in about this and what exactly might come through as dynamic data. I'll add the TODO and be sure to discuss this

mlir/include/Quantum/IR/QuantumInterfaces.td
kipawaa: Definitely! My end goal is that any op that wants to participate in graph decomp should implement this interface, so that we can standardize how we identify which ops need decomposition and how to gather data from them

mlir/include/Quantum/IR/QuantumInterfaces.td
kipawaa: Got it! added support for multiple flat wire args in [c4d2091](https://github.com/PennyLaneAI/catalyst/pull/2983/commits/c4d2091e054b0a1178288ffdbb3aa8639a5c83ad) and tests in [2e3518d](https://github.com/PennyLaneAI/catalyst/pull/2983/commits/2e3518df6abf50f23e2e41443d9441a360babc21)

mlir/include/Quantum/IR/QuantumInterfaces.td
kipawaa: Just to have all of the information in one place, we'll implement this interface for other ops in follow-up PRs. Once all of the core operations have implemented the interface, we can update the `graph-decomposition` pass to support the interface (and _only_ the interface). I've improved the explanation of this in [e3e6736](https://github.com/PennyLaneAI/catalyst/pull/2983/commits/e3e67367fbaa065c38d6434ce36279f5d8468726)

frontend/catalyst/device/python_decompositions.py
kipawaa: added TODOs here: [0437262](https://github.com/PennyLaneAI/catalyst/pull/2983/commits/0437262f2ecfd8fe19785ed6a9e159ad7dd7cc41)

mlir/lib/Quantum/Transforms/GraphDecomposition/graph_decomposition.cpp
kipawaa: I think that it will be more appropriate to try to assert this once we have implemented the auto-tracing of decomposition rules with operators which will be future work for Operator2. Does that seem reasonable?

mlir/include/Quantum/IR/QuantumInterfaces.td
kipawaa: For completeness of information in this thread, the format we landed on is
`name + {dtype tree of args} + {number of wires per wires arg} + {static data} + {uid}`
as documented in the interface description. The interface implements this as a default method for any op without a UID (UID must override since the interface does not provide a UID getter).

dime10: Perfect, thanks for the update! 

Would it make sense to add the UID to the interface? Probably too specific, but perhaps an "extra_data" field to complement the static_data field, which could return any string and doesn't prescribe its format (unlike static_data)?

That way, the default method for the graph node id does not need to be overridden by pretty much anything.

frontend/catalyst/device/python_decompositions.py
```diff
@@ -42,27 +42,36 @@
...
-        )
+        # TODO remove this once we have unified lowering, we should be able to set target_gate and
+        # stop relying on function names
+        decomp_rule.__name__ = op_id + "_" + rule.name
 
-    paulirot_subroutine = qp.capture.subroutine(paulirot_decomp_rule)
+        return qp.capture.subroutine(decomp_rule)
+
+    # let this fail with the standard error message if the op is not found
+    op_class = getattr(qp, op_name)
```
dime10: Is there any chance we could process operators that are not in the top-level pennylane namespace?

dime10: Perhaps it's possible to query the decomp rule registry with the name directly instead of the class, that would circumvent this problem (cc @astralcai).

frontend/catalyst/device/python_decompositions.py
```diff
@@ -42,27 +42,36 @@
...
         capture=True,
     )
     @qp.qnode(device=device)
     def circuit():
-        paulirot_subroutine(theta, wires)
+        for subroutine in subroutines:
+            # TODO I know this is dynamic, but we should probably have a better way of handling this
+            # than hard-coded dummy values. Revisit this when unifying the decomp-rule lowering
+            # pipeline
+            subroutine(*[0.5 for _ in dynamic_shape], wires=wires)
```
dime10, line 73: Yeah especially cause the tracer type might be completely wrong here (it always uses `tensor<f64>`). Are we not getting this information (shape & dtype per argument) passed down from the context invoking this function?

frontend/catalyst/device/python_decompositions.py
```diff
@@ -42,27 +42,36 @@
 import pennylane as qp
 
 
-def paulirot_decomposition_wrapper(theta, pauli_word, wires):
-    """Wraps the paulirot decomp rule for compile-time lowering with a static pauli word.
+def python_decomposition_wrapper(op_name, op_id, dynamic_shape, wire_lens, static_data) -> str:
+    """Generic decomposition wrapper."""
+    device = qp.device("null.qubit", wires=sum(wire_lens))
+    wires = tuple(jnp.array(range(length)) for length in wire_lens)
```
dime10: no dtype for this array? I guess int is the default or something? 😅

frontend/test/pytest/test_QPD.py
```diff
@@ -14,22 +14,45 @@
...
 
     def test_paulirot_wrapper(self):
         """Test that the paulirot QPD wrapper correctly returns the IR as a string."""
-        result = paulirot_decomposition_wrapper(0.4, "XZZ", [0, 1, 2])
+        result = python_decomposition_wrapper(
+            "PauliRot", "paulirot[f64][3]{pauli_word:XZZ}", [0.4], [3], {"pauli_word": "XZZ"}
+        )
         assert isinstance(result, str)
-        assert "paulirot_decomp_rule" in result
+        assert "paulirot[f64][3]{pauli_word:XZZ}__pauli_rot_decomposition" in result
```
dime10: how come "pauli rot" appears twice in this string?

mlir/include/Quantum/IR/QuantumInterfaces.td
```diff
@@ -179,6 +179,41 @@ def DifferentiableGate : OpInterface<"DifferentiableGate", [ParametrizedGate]> {
...
+    let description = [{
+        This interface provides a generic way to query decomposition data from a quantum operation. An operation must implement this interface in order to compatible with the `graph-decomposition` pass. Note that implementing this interface does not guarantee that an operation will decomposed successfully, only that it will be registered with the graph.
+    }];
+
+    let cppNamespace = "::catalyst::quantum";
+
+    let methods = [
+        InterfaceMethod<
+            "Return the name of the corresponding PennyLane operator.", "std::string", "getPlName"
```
dime10: (minor) If we want to be a bit more frontend agnostic 😄 (after all catalyst doesn't care where they come from, at least in the past it's been fully frontend agnostic)

```suggestion
            "Return the name of the corresponding frontend operator (i.e. the rule supplier).", "std::string", "getOperatorName"
```

mlir/include/Quantum/IR/QuantumInterfaces.td
kipawaa: Nice! It definitely seems better to override an extra data field than to override the ID generator, added in [d22d836](https://github.com/PennyLaneAI/catalyst/pull/2983/commits/d22d83636aa13c62727c16bc6627f10025829b5e)

mlir/test/Quantum/DecomposeLoweringTest.mlir
```diff
@@ -669,7 +669,7 @@ module @test_paulirot {
     }
 
     // CHECK: my_paulirot_decomp
-    func.func private @my_paulirot_decomp(%inreg : !quantum.reg, %angle_tensor : tensor<f64>, %q_tensor : tensor<3xi64>) -> !quantum.reg attributes {target_gate = "paulirotZXY"} {
+    func.func private @my_paulirot_decomp(%inreg : !quantum.reg, %angle_tensor : tensor<f64>, %q_tensor : tensor<3xi64>) -> !quantum.reg attributes {target_gate = "PauliRot[f64][3]{pauli_word:ZXY}"} {
```
dime10, line 672: very nice 💯

dime10 (approved):
Nice work, this is coming together nicely 💯

frontend/test/pytest/test_QPD.py
kipawaa: the start of the string is the `graphOpId` which starts with the op name, the end of the string is the name of the decomposition rule which is `_pauli_rot_decomposition` in PL. The plan is to lower rules with the `target_gate` attribute set to the associated `graphOpId` once the rule capture/lowering is generalized for Operator2, but right now it's set by the compiler.

frontend/catalyst/device/python_decompositions.py
kipawaa: woops! [9b33de7](https://github.com/PennyLaneAI/catalyst/pull/2983/commits/9b33de77d4fa42c3297f510d6600b4f1f0030c40)

mlir/include/Quantum/IR/QuantumInterfaces.td
kipawaa: [fde3c2e](https://github.com/PennyLaneAI/catalyst/pull/2983/commits/fde3c2e4546f94d975e3f5bf4f129aee74db79ec)

frontend/catalyst/device/python_decompositions.py
kipawaa: My main concern is finding a way to do this that avoids heuristics. Since we may be receiving shape + dtype from python or MLIR, the handling isn't entirely simple. For now, this correctly handles the only supported case (PauliRot), and I'll explore this in a follow-up that generalizes the lowering pathway to handle pre-compiled, trace-time and on-demand decomposition rules 😄

frontend/catalyst/device/python_decompositions.py
kipawaa: Good idea, we could probably remove this and just rely on `list_decomps(op_name)`? @astralcai does this seem like a  reasonable approach? Since we intend to let this fail anyways if the op isn't found, that failure can be generated from the frontend graph, which allows a wider set of operators

frontend/catalyst/device/python_decompositions.py
dime10: Agreed, there shouldn't be any heuristics. We simply need a format to transmit shape and dtype from mlir to python, but a follow up PR is fine for this (make sure to record a story on shortcut).

kipawaa: Added as a point in [sc-122018]

frontend/test/pytest/test_QPD.py
```diff
@@ -14,22 +14,45 @@
...
 
+    def test_multiple_rules(self):
+        """Test that the python decomposition wrapper supports multiple rules."""
+        with qp.decomposition.local_decomps():
+
+            def test_resources():
+                return {qp.X: 1}
+
+            @qp.register_resources(test_resources)
+            def test_decomp(angle, wires, pauli_word):
```
maliasadi: unused `pauli_word`:

```suggestion
            def test_decomp(angle, wires):
```

mlir/lib/Quantum/Transforms/QuantumPythonDecompositions/PythonFunction.cpp
```diff
@@ -12,37 +12,82 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <exception>
+#include <iostream>
```
maliasadi:
```suggestion
```

mlir/lib/Quantum/Transforms/QuantumPythonDecompositions/PythonFunction.cpp
```diff
@@ -12,37 +12,82 @@
...
+#include "Quantum/IR/QuantumInterfaces.h"
+
 #include "PythonDriverUtils.hpp"
 
 #define DEBUG_TYPE "[QPD] "
 
 namespace nb = nanobind;
 
-std::string pythonLowerPauliRot(double theta, const std::string &pauliWord, std::vector<int> wires)
+nb::object getPyvalFromTypeRange(mlir::TypeRange typerange)
```
maliasadi: could we wrap this and `getPyvalFromMlirAttribute` in a namespace?

maliasadi:
```suggestion
static nb::object getPyvalFromTypeRange(mlir::TypeRange typerange)
```

mlir/lib/Quantum/Transforms/QuantumPythonDecompositions/PythonFunction.cpp
```diff
@@ -12,37 +12,82 @@
...
+    for (auto type : typerange) {
+        std::string typestr;
+        llvm::raw_string_ostream ss(typestr);
+        type.print(ss);
+        pyTypes.append(typestr);
+    }
+    return pyTypes;
+}
+
+nb::object getPyvalFromMlirAttribute(mlir::Attribute attr)
```
maliasadi:
```suggestion
static nb::object getPyvalFromMlirAttribute(mlir::Attribute attr)
```

maliasadi (commented):
Thanks @kipawaa! Happy to approve after resolving ^ comments and adding a changelog entry.

frontend/test/pytest/test_QPD.py
kipawaa: The signature of the decomposition rule has to match the signature of the operator so this is required now, but I added a pylint ignore in [77ccac8](https://github.com/PennyLaneAI/catalyst/pull/2983/commits/77ccac8a7513af511e174fd4bbc366d39669bc7b)

mlir/lib/Quantum/Transforms/QuantumPythonDecompositions/PythonFunction.cpp
kipawaa: [77ccac8](https://github.com/PennyLaneAI/catalyst/pull/2983/commits/77ccac8a7513af511e174fd4bbc366d39669bc7b)

mlir/lib/Quantum/Transforms/QuantumPythonDecompositions/PythonFunction.cpp
kipawaa: [77ccac8](https://github.com/PennyLaneAI/catalyst/pull/2983/commits/77ccac8a7513af511e174fd4bbc366d39669bc7b)

mlir/lib/Quantum/Transforms/QuantumPythonDecompositions/PythonFunction.cpp
kipawaa: [77ccac8](https://github.com/PennyLaneAI/catalyst/pull/2983/commits/77ccac8a7513af511e174fd4bbc366d39669bc7b)

frontend/catalyst/device/python_decompositions.py
astralcai: Yeah, `list_decomps` naturally works with operator names.

kipawaa: Great! Switched to using `op_name` in [18b7971](https://github.com/PennyLaneAI/catalyst/pull/2983/commits/18b79712e0a030cd3ef599f76b5fcbca79180121)

doc/releases/changelog-dev.md
```diff
@@ -238,9 +238,14 @@
 
 <h3>Internal changes ⚙️</h3>
 
+<<<<<<< Updated upstream
```
```suggestion
```

doc/releases/changelog-dev.md
```diff
@@ -238,9 +238,14 @@
...
 * The `cond` PLxPR primitive's lowering rule no longer expects a `True` Literal for the predicate
   of the default else branch.
   [(#3018)](https://github.com/PennyLaneAI/catalyst/pull/3018)
+=======
```
```suggestion

```

doc/releases/changelog-dev.md
```diff
@@ -238,9 +238,14 @@
...
+* Add the `DecomposableGate` op interface to allow generic handling of operations in the `graph-decomposition` pass. This allows arbitrary operations implementing the interface to be registered to and decomposed by the graph. This also allows the use of python-decompositions for any operator pre-registered in the frontend graph.
+  [(#2983)](https://github.com/PennyLaneAI/catalyst/pull/2983)
+>>>>>>> Stashed changes
```
```suggestion
```

maliasadi (approved):
🙌

PR #2985, opened 2026-07-01
Add Readout Error Mitigation (REM)

volodymyr-hq:
**Context:**

Readout Error Mitigation (REM) is a probabilistic QEM technique aimed at mitigating the readout/measurement noise (not to be confused with quantum noise). This is achieved by running two additional calibration circuits: all-zeroes (an empty circuit) and all-ones (a circuit with a Pauli-X gate on each qubit), then using the resulting bitstring distribution after measurement to calculate per-qubit confusion matrices. Finally, these confusion matrices and the resulting bitstring distribution after measurement for the mitigatee circuit are used to solve a linear system and obtain mitigated bitstring distribution. Haiqu's aim as part of the CODE project is to add REM support to the Catalyst compiler.

1. Frontend
We introduced the public `mitigate_with_rem` wrapper. This function hooks into the Catalyst compilation pipeline by wrapping target QNodes, binding a new `rem_p` JAX primitive, and routing the resulting user and calibration samples through JAX-compiled post-processing functions. We tried to follow the format/naming conventions established the existing ZNE Mitigation implementation where possible.

2. JAX-Compilable Post-Processing
Top-level JAX-compilable post-processing functions were added to calculate per-qubit confusion matrices and to solve the linear system to correct noisy samples. A trace-time path dispatch switches from full histograms (bounded by `2^n_qubits x 2^n_qubits`) to a sort-based encoding proposed by [Nation et al.](https://doi.org/10.1103/PRXQuantum.2.040326), which is bounded by `num_shots x num_shots`, when `2^n_qubits` exceeds shot count. This optimization allows the program to scale past 25 qubits. Currently, the `SampleMP`, `CountsMP` and `ProbsMP` are supported.

3. Mitigation Dialect and Lowering Pass
New `RemOp` (`mitigation.rem`) operation was added to the `Mitigation` dialect. The lowering pass rewrites a single calling operation into the original callee alongside two inlined calibration circuits (all-zeros and all-ones states). The `quantum.device` attributes are forwarded to the calibration circuits, automatically propagating the noise probability to both calibration circuits.

4. Calibration Caching Support
A boolean `runCalibration` attribute in the `mitigation.rem` operation that turns on or off the emission of operations for calibration circuits. When set to False, the lowering pass skips generating calibration circuits, which is useful for per-device confusion matrix caching.

**Benefits:**
- Extends catalysts support for QEM technique (only ZNE is supported as of now)
- Is scalable for large qubits counts (not bound by exponential growth of `2^num_qubits` for `SampleMP`), which is beneficial for on-hardware execution and large scale simulation.
- Is backend-agnostic
- Research such as [Quantum Error Correction on Error-mitigated Physical Qubits](https://arxiv.org/abs/2601.18384) by Jeon et al. shows that it is possible to combine QEM techniques with QEC. Since Catalyst already offers QEC dialects, it would make sense to support such pipelines in Catalyst.

**Possible Drawbacks:**
- No support for observables, since Catalyst does not support injecting any classical post-processing between shot sampling and observable calculation. Instead, each backend is expected implement support for each observable type. In the future, It is possible to circumvent this by implementing each observable in JAX and adding a pass that replaces every observable MP by `SampleMP` + classical post-processing. This was however out of scope for this project.
- Currently, no supported device in Catalyst has native support for readout noise channels. Therefore, the tests that we added can only test REM on an identity noise channel, as well as separately testing the JAX post-processing functions. We updated the `qrack.simulator` C++ backend to support simulated readout noise, work is being done to get it merged into `pennylane-qrack` upstream. For now it is possible to test with our forked version available here: https://github.com/volodymyr-hq/pennylane-qrack/tree/readout_noise_channel. Adding `pennylane-qrack` as a dependency for the tests seems unreasonable.
- The JAX post-processing functions to calibrate confusion matrices and solve for mitigated results are added at JAX trace-time, not as part of the `lower-mitigation` pass that replaces `mitigation.rem`. This breaks the abstraction to some extent, and it would be better if these functions were inserted into the program as part of the lowering pass. However, we did not find a reasonably simple way to implement this, since the post-processing functions rely on linear algebra functions (best represented in StableHLO), which is lowered at an earlier point in the pipeline. We are open to suggestions if there is a better way to do this.
- The returned type after mitigation for `SampleMP` is not a tensor of all sampled bitstrings, but a tuple of two tensors: one with a sorted list of unique sampled bitstrings, and another with their respective counts. This differs from what is usually expected from a kernel with a `SampleMP` measurement process, but makes sense for mitigation purposes. Unrolling the counts into samples is not optimal, since that would require probabilistic re-sampling of shots.
- Confusion matrix caching has to be done explicitly by the user.
- This PR modifies ‎`mlir/lib/Mitigation/Transforms/CMakeLists.txt` to set `"-Wno-deprecated-declarations"` since `Rem.cpp` still uses the legacy `rewriter.create<OpTy>(...)` builder, while upstream MLIR has deprecated it in favour of `OpTy::create(rewriter, ...)`. This can be corrected and `Rem.cpp` can be ported to the new interface if needed.

### DIFF
--- a/frontend/catalyst/api_extensions/error_mitigation.py
+++ b/frontend/catalyst/api_extensions/error_mitigation.py
@@ -319,23 +319,12 @@ class RemCallable(CatalystCallable):
             callable_fn
         ), "expected callable set as param on the first operation in rem target"
 
-        # Identify the measurement process by walking the inner jaxpr, matching
-        # the logic in `_rem_abstract_eval`.
-        mp_kind = None
-        for eqn in jaxpr.eqns:
-            inner = eqn.params.get("call_jaxpr", None)
-            if inner is None:
-                continue
-            for op_eq in inner.eqns:
-                pname = str(op_eq.primitive)
-                if pname in ("probs", "sample", "counts"):
-                    mp_kind = pname
-                    break
-            if mp_kind is not None:
-                break
-        assert (
-            mp_kind != None
-        ), "measurement process must be one of CountsMP, ProbsMP or SampleMP. Other measurement processes such as observables are not supported yet."
+        mp_kind = _detect_measurement_process(jaxpr)
+        assert mp_kind is not None, (
+            "measurement process must be one of CountsMP, ProbsMP or SampleMP. "
+            "Other measurement processes such as observables are not supported yet."
+        )
+
         run_calibration = self.calibration_matrices is None
         rem_results = rem_p.bind(
             *args_data,
@@ -347,55 +336,62 @@ class RemCallable(CatalystCallable):
         qnode_obj = jaxpr.eqns[0].params.get("qnode", None)
         assert qnode_obj is not None, "REM post-processing requires a QNode target"
         n_qubits = len(qnode_obj.device.wires)
-        measured_qubits = jnp.array([x for x in qnode_obj.device.wires])
-        if mp_kind == "sample":
+        measured_qubits = jnp.array(list(qnode_obj.device.wires))
 
-            # quantum.sample returns f64 in catalyst; the REM helpers index
-            # confusion matrices by bit value and need an integer dtype.
-            mitigatee_samples = rem_results[0].astype(jnp.int32)
-            if run_calibration:
-                zeros_samples = rem_results[1].astype(jnp.int32)
-                ones_samples = rem_results[2].astype(jnp.int32)
-                confusion_matrices = rem_calibrate_samples(zeros_samples, ones_samples)
-            else:
-                confusion_matrices = jnp.asarray(self.calibration_matrices)
-            unique_bitstrings, mitigated_counts = rem_apply_to_samples(
-                mitigatee_samples, confusion_matrices, measured_qubits, n_qubits
-            )
-            mitigated_result = (unique_bitstrings, mitigated_counts)
-
-        elif mp_kind == "counts":
-
-            mitigatee_eigvals = rem_results[0]
-            mitigatee_counts = rem_results[1]
-            if run_calibration:
-                zeros_counts = rem_results[2]
-                ones_counts = rem_results[3]
-                confusion_matrices = rem_calibrate_counts(zeros_counts, ones_counts)
-            else:
-                confusion_matrices = jnp.asarray(self.calibration_matrices)
-            mitigated_counts = rem_apply_to_counts(
-                mitigatee_counts, confusion_matrices, measured_qubits, n_qubits
-            )
-            mitigated_result = (mitigatee_eigvals, mitigated_counts)
-
-        elif mp_kind == "probs":
-
-            mitigatee_probs = rem_results[0]
-            if run_calibration:
-                zeros_probs = rem_results[1]
-                ones_probs = rem_results[2]
-                confusion_matrices = rem_calibrate_probs(zeros_probs, ones_probs)
-            else:
-                confusion_matrices = jnp.asarray(self.calibration_matrices)
-            mitigated_probs = rem_apply_to_probs(
-                mitigatee_probs, confusion_matrices, measured_qubits, n_qubits
-            )
-            mitigated_result = mitigated_probs
+        handler = {
+            "sample": self._apply_sample,
+            "counts": self._apply_counts,
+            "probs": self._apply_probs,
+        }[mp_kind]
+        mitigated_result, confusion_matrices = handler(
+            rem_results, run_calibration, measured_qubits, n_qubits
+        )
 
         if self.return_calibration_matrices:
             return mitigated_result, confusion_matrices
         return mitigated_result
+
+    def _apply_sample(self, rem_results, run_calibration, measured_qubits, n_qubits):
+        # quantum.sample returns f64 in catalyst; the REM helpers index
+        # confusion matrices by bit value and need an integer dtype.
+        mitigatee_samples = rem_results[0].astype(jnp.int32)
+        if run_calibration:
+            zeros_samples = rem_results[1].astype(jnp.int32)
+            ones_samples = rem_results[2].astype(jnp.int32)
+            confusion_matrices = rem_calibrate_samples(zeros_samples, ones_samples)
+        else:
+            confusion_matrices = jnp.asarray(self.calibration_matrices)
+        unique_bitstrings, mitigated_counts = rem_apply_to_samples(
+            mitigatee_samples, confusion_matrices, measured_qubits, n_qubits
+        )
+        return (unique_bitstrings, mitigated_counts), confusion_matrices
+
+    def _apply_counts(self, rem_results, run_calibration, measured_qubits, n_qubits):
+        mitigatee_eigvals = rem_results[0]
+        mitigatee_counts = rem_results[1]
+        if run_calibration:
+            zeros_counts = rem_results[2]
+            ones_counts = rem_results[3]
+            confusion_matrices = rem_calibrate_counts(zeros_counts, ones_counts)
+        else:
+            confusion_matrices = jnp.asarray(self.calibration_matrices)
+        mitigated_counts = rem_apply_to_counts(
+            mitigatee_counts, confusion_matrices, measured_qubits, n_qubits
+        )
+        return (mitigatee_eigvals, mitigated_counts), confusion_matrices
+
+    def _apply_probs(self, rem_results, run_calibration, measured_qubits, n_qubits):
+        mitigatee_probs = rem_results[0]
+        if run_calibration:
+            zeros_probs = rem_results[1]
+            ones_probs = rem_results[2]
+            confusion_matrices = rem_calibrate_probs(zeros_probs, ones_probs)
+        else:
+            confusion_matrices = jnp.asarray(self.calibration_matrices)
+        mitigated_probs = rem_apply_to_probs(
+            mitigatee_probs, confusion_matrices, measured_qubits, n_qubits
+        )
+        return mitigated_probs, confusion_matrices
 
 
 def polynomial_extrapolation(degree):
@@ -410,3 +406,19 @@ def _wrap_callable(fn):
     elif isinstance(fn, Callable):  # Keep at the bottom
         return Function(fn)
     raise TypeError(f"Target must be callable, got: {type(fn)}")
+
+
+def _detect_measurement_process(jaxpr):
+    """Walk the inner jaxpr for a probs/sample/counts primitive; return its name or None.
+
+    Mirrors the detection logic in :func:`_rem_abstract_eval`.
+    """
+    for eqn in jaxpr.eqns:
+        inner = eqn.params.get("call_jaxpr", None)
+        if inner is None:
+            continue
+        for op_eq in inner.eqns:
+            pname = str(op_eq.primitive)
+            if pname in ("probs", "sample", "counts"):
+                return pname
+    return None
--- a/frontend/catalyst/api_extensions/error_mitigation.py
+++ b/frontend/catalyst/api_extensions/error_mitigation.py
@@ -27,7 +27,8 @@ import jax.numpy as jnp
 import pennylane as qp
 from jax._src.tree_util import tree_flatten
 
-from catalyst.api_extensions.rem_postprocessing import rem_apply_to_samples, rem_calibrate
+from catalyst.api_extensions.rem_postprocessing import rem_apply_to_samples, rem_apply_to_counts, rem_apply_to_probs
+from catalyst.api_extensions.rem_postprocessing import rem_calibrate_samples, rem_calibrate_counts, rem_calibrate_probs
 from catalyst.jax_primitives import Folding, func_p, quantum_kernel_p, rem_p, zne_p
 from catalyst.jax_tracer import Function
 from catalyst.utils.callables import CatalystCallable
@@ -308,7 +309,9 @@ class RemCallable(CatalystCallable):
                     break
             if mp_kind is not None:
                 break
-
+        assert mp_kind != None, (
+            "measurement process must be one of CountsMP, ProbsMP or SampleMP. Other measurement processes such as observables are not supported yet."
+        )
         rem_results = rem_p.bind(
             *args_data,
             compute_all_zeroes_ones=self.compute_all_zeroes_ones,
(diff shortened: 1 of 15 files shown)


volodymyr-hq:
See also https://github.com/unitaryfoundation/pennylane-qrack/pull/40 for a PR into `pennylane-qrack` repo with a readout noise channel support. It was used to test REM on simulated noise.

mlir/lib/Mitigation/Transforms/CMakeLists.txt
```diff
@@ -18,6 +18,16 @@ set(DEPENDS
...
 
+# MitigationMethods/Rem.cpp still uses the legacy `rewriter.create<OpTy>(...)`
+# builder; upstream MLIR has deprecated it in favour of `OpTy::create(rewriter, ...)`.
+# COMPILE_FLAGS rather than target_compile_options so the suppression also
+# applies to the OBJECT library that `add_mlir_library` creates internally.
+# Migration to the new builder is tracked separately.
+set_source_files_properties(
+    MitigationMethods/Rem.cpp
+    PROPERTIES COMPILE_FLAGS "-Wno-deprecated-declarations"
+)
```
rniczh, line 29: Could you use `OpTy::create(...)` instead of `rewriter.create<OpTy>(...)` to avoid introducing this option?

mlir/lib/Mitigation/Transforms/MitigationMethods/Rem.hpp
```diff
@@ -0,0 +1,23 @@
...
+    using OpRewritePattern<mitigation::RemOp>::OpRewritePattern;
+
+    LogicalResult matchAndRewrite(mitigation::RemOp op, PatternRewriter &rewriter) const override;
+};
+
+// Add patterns from this file into a pattern set.
+void populateRemLoweringPatterns(RewritePatternSet &patterns);
+
+} // namespace mitigation
+} // namespace catalyst
```
rniczh, line 23: Always has one newline at the end

```suggestion
} // namespace catalyst

```

frontend/catalyst/api_extensions/error_mitigation.py
```diff
@@ -160,6 +168,33 @@ def workflow(weights, s):
...
+    ``return_calibration_matrices=True`` to get the matrices back alongside
+    the mitigated result so they can be cached for subsequent calls.
+    """
+
+    kwargs = copy.copy(locals())
+    kwargs.pop("fn")
+
+    if fn is None:
+        return functools.partial(mitigate_with_rem, **kwargs)
+
```
rniczh, line 190: I think you could add validation for the calibration matrices here, for example, at least checking that the dimensions, shapes, etc. are correct.

e.g.

```
_check_calibration_matrices(...)
```

frontend/catalyst/api_extensions/error_mitigation.py
```diff
@@ -160,6 +168,33 @@ def workflow(weights, s):
...
+):
+    """A qjit-compatible frontend that emits a `mitigation.rem` operation.
+
+    When ``calibration_matrices`` is ``None`` the lowering pass emits the
+    all-zeros and all-ones calibration circuits and this callable computes
+    the confusion matrices itself. When a precomputed ``(n_qubits, 2, 2)``
+    matrix stack is supplied, ``rem_calibrate_*`` is skipped and the matrices
+    are passed straight into ``rem_apply_to_*``. Pass
+    ``return_calibration_matrices=True`` to get the matrices back alongside
+    the mitigated result so they can be cached for subsequent calls.
```
rniczh, line 182: Because this comment will end up in the API docs, you may need to check how `mitigate_with_zne` is written and make this section a bit clearer.

frontend/catalyst/api_extensions/error_mitigation.py
```diff
@@ -235,6 +270,130 @@ def __call__(self, *args, **kwargs):
...
+            *args_data,
+            run_calibration=run_calibration,
+            jaxpr=jaxpr,
+            fn=callable_fn,
+        )
+
+        qnode_obj = jaxpr.eqns[0].params.get("qnode", None)
+        assert qnode_obj is not None, "REM post-processing requires a QNode target"
+        n_qubits = len(qnode_obj.device.wires)
+        measured_qubits = jnp.array(list(qnode_obj.device.wires))
```
rniczh, line 339: `device.wires` returns wire labels, for example, `qp.device("lightning.qubit", wires=[3, 4])` will still be 2 qubits, the indices are `[3, 4]`,  I think you should use `jnp.arange(n_qubits)`? And you should also change the related uses: (`_apply_sample`, `_apply_counts`, and `_apply_probs`).

e.g.

The `selected` would be identical to the `confusion_matrices`
```python
selected = confusion_matrices[measured_qubits]
stretched = jax.vmap(_stretch_confusion_matrix, in_axes=(1, 0))(
    unique_bitstrings, selected
)
```
You can rewrite it to
```python
stretched = jax.vmap(_stretch_confusion_matrix, in_axes=(1, 0))(
    unique_bitstrings, confusion_matrices
)
```

For others:
```python
selected = [
    jnp.linalg.inv(confusion_matrices[measured_qubits[i]]) for i in range(n_qubits)
]
```
to 
```python
selected = [jnp.linalg.inv(confusion_matrices[i]) for i in range(n_qubits)]
```

mlir/lib/Mitigation/Transforms/MitigationMethods/Rem.cpp
```diff
@@ -0,0 +1,280 @@
...
+    // pre-compute MP and attribute-dependent values
+    RankedTensorType tensorTyI64;
+    RankedTensorType tensorTyF64;
+    if (MPName == "quantum.probs"s || MPName == "quantum.counts"s) {
+        // avoid undefined behaviour due to integer overflow
+        if (qubitCount >= 63) {
+            // in practice, due to exponential memory growth, users will
+            // OOM much earlier, at < 30 qubits
+            llvm::errs() << "[mitigation.rem] Qubit count" << qubitCount
+                         << "exceeds maximum simulated register size (62 qubits).\n";
```
rniczh, line 218: It just prints the error message, would still run into UB, you should emit error?

```suggestion
        if (qubitCount >= 63) {
            // in practice, due to exponential memory growth, users will
            // OOM much earlier, at < 30 qubits
            return op.emitOpError("Qubit count ") << qubitCount << " exceeds maximum simulated register size (62 qubits)";
```

mlir/lib/Mitigation/Transforms/MitigationMethods/Rem.cpp
```diff
@@ -0,0 +1,280 @@
...
+        tensorTyI64 = RankedTensorType::get({calibrationTensorShape}, rewriter.getI64Type());
+    }
+    else if (MPName == "quantum.sample"s) {
+        // shots must be a compile-time arith.constant with a non-zero value.
+        if (shots->getName().getIdentifier() != arith::ConstantOp::getOperationName()) {
+            llvm::errs() << "[mitigation.rem] Dynamic number of shots not supported currently.\n";
+            return failure();
+        }
+        auto shotsAttr = shots->getAttrOfType<IntegerAttr>("value");
+        if (!shotsAttr || shotsAttr == 0) {
```
rniczh, line 231: `!shotsAttr` is same as `shotsAttr == 0`

I think what you want is check the value:

```suggestion
        auto shotsAttr = shots->getAttrOfType<IntegerAttr>("value");
        if (!shotsAttr || shotsAttr.getInt() == 0) {
```

mlir/lib/Mitigation/Transforms/MitigationMethods/Rem.cpp
```diff
@@ -0,0 +1,280 @@
...
+
+    Operation *measurementProcess =
+        (*calleeOp.getOps<quantum::MeasurementProcess>().begin()).getOperation();
+    const std::string MPName = measurementProcess->getName().getIdentifier().str();
+
+    Operation *shotsLocal = shots->clone();
+    rewriter.insert(shotsLocal);
+    auto devInit =
+        rewriter.create<quantum::DeviceInitOp>(loc, shotsLocal->getResult(0), lib, name, kwargs);
+    llvm::dbgs() << "[mitigation.rem] quantum.device initialization " << tag
```
rniczh, line 55: All the `llvm:dbgs()` should be wrapped by `LLVM_DEBUG(...)`

mlir/lib/Mitigation/Transforms/MitigationMethods/Rem.cpp
```diff
@@ -0,0 +1,280 @@
...
+// compbasis to produce the all-ones basis state. Returns the calibration result
+// values (drops the eigvals half of CountsOp).
+ValueRange emitCalibrationCircuit(PatternRewriter &rewriter, Location loc, func::FuncOp calleeOp,
+                                  RankedTensorType tensorTyF64, RankedTensorType tensorTyI64,
+                                  bool applyPauliX)
+{
+    const char *tag = applyPauliX ? "(ones)" : "(zeros)";
+
+    const int64_t qubitCount =
+        (*calleeOp.getOps<quantum::AllocOp>().begin()).getNqubitsAttr().value_or(0);
```
rniczh, line 40: The calibration tensor would become 1 << 0 == 1 if it get the 0 here, I think you should check it and emit error if getNqbuitsAttr is null?

mlir/lib/Mitigation/Transforms/MitigationMethods/Rem.cpp
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    Type obsType = ::catalyst::quantum::ObservableType::get(rewriter.getContext());
+    auto compbasis = rewriter.create<quantum::ComputationalBasisOp>(loc, TypeRange{obsType},
+                                                                    ValueRange{}, currentQreg);
+    llvm::dbgs() << "[mitigation.rem] quantum.compbasis " << tag << " addition OK, op=" << compbasis
+                 << "\n";
+
+    ValueRange calibratedResult;
+    if (MPName == "quantum.probs"s) {
```
rniczh, line 87: I think it's better to use `dyn_cast` or `isa<>` rather than comparing with string

e.g.

```python
    if (auto probsOp = dyn_cast<quantum::ProbsOp>(measurementProcess))
```

mlir/lib/Mitigation/Transforms/MitigationMethods/Rem.cpp
```diff
@@ -0,0 +1,280 @@
...
+            llvm::dbgs() << "[mitigation.rem] looking for next op: "
+                         << postprocOp->getName().getIdentifier() << "\n";
+            Operation *insertedOp = rewriter.clone(*postprocOp, postprocMapping);
+            if (!insertedOp->getResults().empty())
+                calibratedResult = insertedOp->getResults();
+            rewriter.setInsertionPointAfter(insertedOp);
+            llvm::dbgs() << "[mitigation.rem] insertedOp: " << insertedOp->getName().getIdentifier()
+                         << "\n";
+        }
+    }
```
rniczh, line 127: unreachable?

```suggestion
    } else {
        llvm_unreachable();
    }
```

mlir/lib/Mitigation/Transforms/MitigationMethods/Rem.cpp
```diff
@@ -0,0 +1,280 @@
+#include "Rem.hpp"
```
rniczh, line 1: Don't forget the license header

rniczh (commented):
Thanks for putting this together! Left some comments below

frontend/catalyst/api_extensions/rem_postprocessing.py
```diff
@@ -0,0 +1,324 @@
...
+
+@jax.jit
+def rem_calibrate_samples(zeros_samples: jax.Array, ones_samples: jax.Array) -> jax.Array:
+    """Build per-qubit (n_qubits, 2, 2) confusion matrices from calibration samples.
+
+    Step 1 of the REM pipeline: consume the all-zeros and all-ones
+    calibration samples emitted by the lowered ``mitigation.rem`` op and
+    return the normalized confusion matrices.
+    """
+    rev_zeros = zeros_samples[:, ::-1].astype(jnp.float64)
```
mehrdad2m, line 128: Could you explain why we are reversing the order here? AI is telling me that there are inconsistencies in the way are treating wire order and orientation of the confusion matrix within rem_calibrate_* and rem_apply_*, but it is bit hard for me to confirm. 
I also think it would be helpful to add more explanation on the confusion matrix shape (what the row and columns represent)

mlir/lib/Mitigation/Transforms/MitigationMethods/Rem.cpp
```diff
@@ -0,0 +1,280 @@
...
+            Operation *insertedOp = rewriter.clone(*postprocOp, postprocMapping);
+            if (!insertedOp->getResults().empty())
+                calibratedResult = insertedOp->getResults();
+            rewriter.setInsertionPointAfter(insertedOp);
+            llvm::dbgs() << "[mitigation.rem] insertedOp: " << insertedOp->getName().getIdentifier()
+                         << "\n";
+        }
+    }
+
+    auto dealloc = rewriter.create<quantum::DeallocOp>(loc, qreg);
```
mehrdad2m, line 129: should we deallocate `currentQreg` instead? in general, we should only use the final register in the insert chain as an insert modifies the original register and returns a new value.

mlir/lib/Mitigation/Transforms/MitigationMethods/Rem.cpp
```diff
@@ -0,0 +1,280 @@
...
+    StringAttr kwargs = deviceInitOp.getKwargsAttr();
+
+    Operation *measurementProcess =
+        (*calleeOp.getOps<quantum::MeasurementProcess>().begin()).getOperation();
+    const std::string MPName = measurementProcess->getName().getIdentifier().str();
+
+    Operation *shotsLocal = shots->clone();
+    rewriter.insert(shotsLocal);
+    auto devInit =
+        rewriter.create<quantum::DeviceInitOp>(loc, shotsLocal->getResult(0), lib, name, kwargs);
```
mehrdad2m, line 54: Note that the rewriter.create<Op>() method has been deprecated. We should use `OpTy::create(builder, ...)` instead. (see https://github.com/llvm/llvm-project/pull/164649).

```suggestion
        quantum::DeviceInitOp::create(rewriter, loc, shotsLocal->getResult(0), lib, name, kwargs);
```

mlir/lib/Mitigation/Transforms/MitigationMethods/Rem.cpp
```diff
@@ -0,0 +1,280 @@
...
+    llvm::dbgs() << "[mitigation.rem] qubitCount: " << qubitCount << ", shots: " << *shots
+                 << ", lib: " << deviceInitOp.getLibAttr()
+                 << ", name: " << deviceInitOp.getDeviceNameAttr() << ", kwargs"
+                 << deviceInitOp.getKwargsAttr() << "\n";
+
+    // check for MP type: Sample, Probs or Counts
+    auto measurementProcesses = calleeOp.getOps<quantum::MeasurementProcess>();
+    if (measurementProcesses.empty()) {
+        llvm::errs() << "[mitigation.rem] No valid MP found.\n";
+        return failure();
```
mehrdad2m, line 203:
```suggestion
       return op.emitOpError("[mitigation.rem] No valid MP found.\n");
```
We can use the diagnostics infrastructure in MLIR for better error handling. https://mlir.llvm.org/docs/Diagnostics/

mlir/lib/Mitigation/Transforms/MitigationMethods/Rem.cpp
```diff
@@ -0,0 +1,280 @@
...
+            completeResults.push_back(v);
+        rewriter.replaceOp(op, completeResults);
+        return success();
+    }
+
+    // Determine quantum device information, specifically number of qubits and shots.
+    // Get the number of qubits from quantum.alloc operation inside callee function (FuncOp).
+    const int64_t qubitCount =
+        (*calleeOp.getOps<quantum::AllocOp>().begin()).getNqubitsAttr().value_or(0);
+    // Get the device using quantom.device opeartion  inside callee function (FuncOp)
```
mehrdad2m, line 189:
```suggestion
    // Get the device using quantom.device operation inside callee function (FuncOp)
```

frontend/catalyst/api_extensions/error_mitigation.py
```diff
@@ -235,6 +270,130 @@ def __call__(self, *args, **kwargs):
...
+            func_p,
+            quantum_kernel_p,
+        }, "expected func_p or quantum_kernel_p as first operation in rem target"
+        callable_fn = jaxpr.eqns[0].params.get("fn", callable_fn)
+        assert callable(
+            callable_fn
+        ), "expected callable set as param on the first operation in rem target"
+
+        mp_kind = _detect_measurement_process(jaxpr)
+        assert mp_kind is not None, (
```
mehrdad2m, line 323: we should probably raise a `NotImplementedError` here instead since this is a user facing failure and not a programmer assumption.

frontend/catalyst/jax_primitives.py
```diff
@@ -1083,6 +1085,114 @@ def _zne_lowering(ctx, *args, folding, jaxpr, fn):
...
+    an all-ones calibration result. Calibration shapes/dtypes are derived
+    from the QNode's device size and the measurement-process kind on the
+    inner jaxpr (probs, counts, sample).
+    """
+    qnode = jaxpr.eqns[0].params["qnode"]
+    device_qubit_count = len(qnode.device.wires)
+    device_shot_count = qnode.shots.total_shots
+
+    mp = None
+    for eqn in subjaxprs(jaxpr):
```
mehrdad2m, line 1108: what does subjaxpr does? could we perhaps reuse the `_detect_measurement_process` method here as well?

mlir/test/Mitigation/RemLoweringTest.mlir
```diff
@@ -0,0 +1,185 @@
...
+// -----
+
+// ===================================================================
+// A missing callee leaves the mitigation.rem op untouched: the lowering
+// pattern bails out via notifyMatchFailure, which keeps the IR in a state
+// that downstream passes can either re-try or surface as a hard error.
+// ===================================================================
+
+func.func @remMissingCallee() -> (tensor<4xf64>, tensor<4xf64>, tensor<4xf64>) {
+    %out:3 = mitigation.rem @doesNotExist() runCalibration(true) {resultSegmentSizes = array<i32: 1, 1, 1>} : () -> (tensor<4xf64>, tensor<4xf64>, tensor<4xf64>)
```
mehrdad2m, line 180: I am wondering if there is a way to raise an error in the verifier of this op instead of leaving it for downstream passes to crash.

mehrdad2m (commented):
Thanks @volodymyr-hq for the amazing work! 

The overall implementation look good to me. I put some mostly minor comments. Otherwise my main concern for merging this work is that the testing currently does not act on any noise and therefore the mathematics of the error mitigation is effectively untested. This could be mitigated to some level by adding more testing to specifically test the error mitigation math without any noisy device. 
 We of course have the option of holding off on merging and keeping the branch, to use alongside your qrack branch in the meantime.

PR #2986, opened 2026-07-02
Prevent issues with duplicate function names in `measurements_from_samples`

lillian542:
**Context:**
In certain contexts, the helper functions for `measurements_from_samples` end up being added in the MLIR where there is already a function with that name, causing an error. This came up in an algorithm in the benchmarks.

Instead of generic names like `_where_0`, `_where_1`, etc, we name the functions we are adding for post-processing for `measurements_from_samples` things like `_where_meas_from_samples_0`. 

Note that no test is added because its not entirely obvious to me what about the relevant benchmark is causing this to happen, and I haven't been able to replicate the problem on a small toy circuit for a test.

**Benefits:**
Now the pass won't clash with things other passes have done when it comes to adding helper functions.


paul0403 (approved):

paul0403:
In mlir api, symbol tables have a `renameToUnique` method for precisely this scenario. But I searched on xdsl github and their corresponding `rename_to_unique` is not implemented yet :melting_face:

joeycarter (approved):

PR #2987, opened 2026-07-02
Omnistaging is globally off by default

paul0403:
**Context:**
In tandem PL PR https://github.com/PennyLaneAI/pennylane/pull/9755

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-124076]


dime10:
I was wondering if this was already working on the whole test suite, but I guess not 😅 

Once it does, we should run the benchmarks on it, although I'm not sure if those are set up for program capture yet?

paul0403:
> I was wondering if this was already working on the whole test suite, but I guess not 😅
> 
> Once it does, we should run the benchmarks on it, although I'm not sure if those are set up for program capture yet?

I think there's more quantum primitives that I need to "turn off" the jax const eval context. That's probably why the precompiled rules are failing. 

But I think this is on the right track. For a small example, I can already get it to work correctly.

paul0403:
Also it needs the PL PR but I didn't do the "run with this PL PR" in CI :laughing:

paul0403:
PL PR has been merged, I'll check tests tomorrow after tonight's nightly upload

PR #2980, opened 2026-06-26
Add support for lowering adjoint and control with Operator2

mudit2812:
**Context:**
Follow up for supporting lowering of `Operator2`

**Description of the Change:**
* Add support for translating adjoint and control to Catalyst jaxpr and MLIR.
* Change the `adjoint` input of `qref/quantum.operator` from a property to an attribute.
  * Without this change, adjoint was not able to be lowered. There weren't any errors; the value of adjoint was just not being propagated by the Python binding.

**Note to reviewers:**
There is currently a bug with `qp.ctrl` + `qp.qjit(capture=True)` + `Operator2`, which is being fixed in this PR: https://github.com/PennyLaneAI/pennylane/pull/9754. Until that is merged, the CI for this PR will keep failing. But, I've confirmed locally that all tests pass when using the PL-side bug fix branch, so I don't think test failures need to block reviews.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-123426]

PR #2990, opened 2026-07-03
Add `adjoint` and `n_ctrls` args to Operator2 primitive

mudit2812:
Fixing LLL failures due to disparity between operator primitive in pennylane and the implementation in catalyst.


albi3ro (approved):

frontend/catalyst/from_plxpr/qref_operator2_primitives.py
paul0403, line 141: So I think this line is no longer needed, since we have the actual adjoint arg?

mudit2812: Addressing in #2980 . This PR is just to fix LLL failures.

paul0403 (approved):

PR #2992, opened 2026-07-03
[Experimental Backline] Define runtime transport layer for backline backends

josephleekl:
### Before submitting

**Context:**
Adds an interface that backline backend devices must adhere to.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-123779]


runtime/include/Transport.hpp
```diff
@@ -0,0 +1,187 @@
...
+};
+
+/**
+ * @brief Memory kind: selects the allocation and registration path.
+ */
+enum class MemKind {
+    CpuRam,  // Plain host RAM.
+    GpuHbm,  // GPU HBM, registered via dma-buf.
+    FpgaDdr, // FPGA DDR, allocated via the Xilinx UMM allocator.
+    Bram,    // FPGA on-chip block RAM.
```
multiphaseCFD: what about `FpgaBram`?

runtime/include/Transport.hpp
```diff
@@ -0,0 +1,190 @@
...
+    virtual ~Channel() = default;
+
+    /**
+     * @brief Start the engine driving this channel's transfers.
+     *
+     * Non-blocking: after this call the engine runs autonomously and the CPU is out of the
+     * per-round loop. A bounded channel runs `ChannelDesc::n_rounds` rounds; a persistent channel
+     * (`n_rounds == 0`) runs until `close`.
+     */
+    virtual void start() = 0;
```
josephleekl, line 134: this is possibly too high level, we may need more granular control to support more runtime calls

runtime/include/Transport.hpp
```diff
@@ -0,0 +1,190 @@
...
+#include <memory>
+#include <string>
+#include <vector>
+
+namespace catalyst::transport {
+
+/**
+ * @brief Data-plane strategy: which engine issues the transfer.
+ */
+enum class DataPath {
```
multiphaseCFD: How could we fit the mix path `GPU polling/process + signal CPU for doorbell ring` to the current design?

multiphaseCFD (commented):
Thanks @josephleekl . Looks good and just a couple of Qs

runtime/include/Transport.hpp
```diff
@@ -0,0 +1,190 @@
...
+};
+
+/**
+ * @brief Memory kind: selects the allocation and registration path.
+ */
+enum class MemKind {
+    CpuRam,  // Plain host RAM.
+    GpuHbm,  // GPU HBM, registered via dma-buf.
+    FpgaDdr, // FPGA DDR, allocated via the Xilinx UMM allocator.
+    Bram,    // FPGA on-chip block RAM.
```
josephleekl:
```suggestion
    FpgaBram,    // FPGA on-chip block RAM.
```

runtime/include/Transport.hpp
josephleekl: So in that case we would use CpuVerbs, and have the device called GPU which uses just transfers the data to the CPU and use ibverbs. We can have another device (or the same device that can have a choice) that uses uses bf-inline, but we can ignore that for now

multiphaseCFD: Would it be safer to specify the underlying data type for `DataPath` here, say `enum class DataPath : std::uint8_t`? as it's possible that same definition would be cross compiled to different target devices

runtime/include/Transport.hpp
```diff
@@ -0,0 +1,190 @@
...
+
+/**
+ * @brief A memory region, allocated and registered by the device at runtime.
+ *
+ * The address and keys are produced by the device; the host
+ * holds the region as an opaque handle.
+ */
+struct MemRegion {
+    void *addr = nullptr;           // Device-allocated address.
+    std::size_t size = 0;           // Region size in bytes.
```
multiphaseCFD: Can we set it as `std::uint64_t` to make the contract to be identical across devices?

runtime/include/Transport.hpp
```diff
@@ -0,0 +1,190 @@
...
+    std::uint32_t lkey = 0;         // Local key.
+    std::uint32_t rkey = 0;         // Remote key.
+    MemKind kind = MemKind::CpuRam; // Memory kind / registration path.
+    int device_id =
```
multiphaseCFD: Same thing for `device_id`. I guess `std::int32_t` would be sufficient?

runtime/include/Transport.hpp
```diff
@@ -0,0 +1,190 @@
...
+    virtual int collect(void *const *outputs, std::size_t n) = 0;
+
+    /**
+     * @brief Tear down the channel, stopping a persistent engine if one is running.
+     */
+    virtual void close() = 0;
+};
+
+/**
+ * @brief Control-plane transport session. Verbs-backed for every backend.
```
multiphaseCFD: Nice! Since this is an abstract interface with state dependencies, let's document the lifecycle at the top of the class. Maybe 1. alloc_memory; 2 connect; 3. exchange_keys; 4 establish_channel?

runtime/include/Transport.hpp
josephleekl: good idea i'll update it

runtime/include/Transport.hpp
josephleekl: I've removed device_id for now, there doesn't seem to be an immediate need for it. But I've made all data types explicit now.

runtime/include/Transport.hpp
josephleekl: good idea, i'll add it.

multiphaseCFD (approved):
Thanks @josephleekl for putting them together! Nice work! Happy to approve.

rniczh (approved):

rniczh:
Don't forget to update the PR's description

runtime/include/Transport.hpp
```diff
@@ -0,0 +1,125 @@
...
+    virtual int collect(void *const *outputs, std::size_t n) = 0;
+
+    // Stop the engine and join. Idempotent.
+    virtual void stop() = 0;
+
+    // Register the per-shot decoder (coprocessor role). No-op default so
+    // controller-only / non-decoding devices need not implement it. A
+    // DataPath::KernelFused device also leaves this unused: its fused poll+decode
+    // kernel is provided by the user/compiler, not registered here.
+    virtual void set_decoder(DecodeFn /*fn*/, void * /*ctx*/) {}
```
josephleekl: should decoder be in this network interface?

rniczh: It might be too specific 🤔

josephleekl: hmm, do we think it would make sense to have a decoder class, which inherits the transport class and adds these decoder abilities, and we have the coprocessors inherit this decoder class?

runtime/include/Transport.hpp
```diff
@@ -0,0 +1,125 @@
...
+
+namespace catalyst::transport {
+
+/**
+ * @brief Data-plane strategy: which engine issues the transfer.
+ */
+enum class DataPath : std::uint8_t {
+    CpuVerbs,    // Plain ibverbs on CPU.
+    NicEngine,   // Hardware-embedded RNIC (e.g. ERNIC hardware handshake).
+    KernelFused, // Fused with a kernel (e.g. on GPU). The decode runs
```
josephleekl:
```suggestion
    GpuEngine, // Fused with a kernel (e.g. on GPU). The decode runs
```

runtime/include/Transport.hpp
```diff
@@ -0,0 +1,120 @@
...
+    // Wait for a result and write it out.
+    virtual int collect(void *const *outputs, std::size_t n) = 0;
+
+    // Stop the engine and join. Idempotent.
+    virtual void stop() = 0;
+};
+
+// The decode I/O contract: a guarantee pinned by the frontend/compiler and
+// carried down to the coprocessor.
+struct DecoderSchema {
```
josephleekl: we probably don't need this, or rename to CoprocessorFunctionSchema

runtime/include/Transport.hpp
rniczh: While working on the impl, I've found that this `start()` design has a lot of issues that don't make sense. We need more discussion. For the follow-up work on the overall interface for the backend to consume, I've opened a PR: https://github.com/PennyLaneAI/catalyst/pull/3042

runtime/include/Transport.hpp
```diff
@@ -0,0 +1,120 @@
...
+#include <string>
+
+namespace catalyst::transport {
+
+/**
+ * @brief Data-plane strategy: which engine issues the transfer.
+ */
+enum class DataPath : std::uint8_t {
+    CpuVerbs,    // Plain ibverbs on CPU.
+    NicEngine,   // Hardware-embedded RNIC (e.g. ERNIC hardware handshake).
```
josephleekl:
```suggestion
    NicEngine,
```

runtime/include/Transport.hpp
```diff
@@ -0,0 +1,120 @@
...
+    GpuEngine,   // Gpu-initiated comms.
+};
+
+/**
+ * @brief Memory kind: selects the allocation and registration path.
+ */
+enum class MemKind : std::uint8_t {
+    CpuRam,   // Plain host RAM.
+    GpuHbm,   // GPU HBM, registered via dma-buf.
+    FpgaDdr,  // FPGA DDR, allocated via the Xilinx UMM allocator.
```
josephleekl:
```suggestion
    FpgaDdr,  // FPGA DDR
```

runtime/include/Transport.hpp
```diff
@@ -0,0 +1,190 @@
...
+
+namespace catalyst::transport {
+
+/**
+ * @brief Data-plane strategy: which engine issues the transfer.
+ */
+enum class DataPath : std::uint8_t {
+    CpuVerbs,  // Plain ibverbs on CPU.
+    GpuEngine, // Gpu-initiated comms.
+    NicEngine,
```
josephleekl:
```suggestion
    Other,
```

PR #2993, opened 2026-07-03
add support for OperatorOp with ResourceAnalysisPass

albi3ro:
**Context:**

Adds support for qref::OperatorOp and quantum::OperatorOp to qp.specs and the ResourceAnalysisPass.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-123822]

### DIFF
--- a/mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
+++ b/mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
@@ -77,6 +77,8 @@ static std::string getGateOpName(Operation *op, bool isAdjoint)
         llvm::TypeSwitch<Operation *, std::string>(op)
             .Case<quantum::CustomOp, qref::CustomOp>(
                 [](auto customOp) { return customOp.getGateName().str(); })
+            .Case<quantum::OperatorOp, qref::OperatorOp>(
+                [](auto operatorOp) { return operatorOp.getOpName().str(); })
             .Case<quantum::PauliRotOp, qref::PauliRotOp>([](auto) { return "PauliRot"; })
             .Case<quantum::GlobalPhaseOp, qref::GlobalPhaseOp>([](auto) { return "GlobalPhase"; })
             .Case<quantum::MultiRZOp, qref::MultiRZOp>([](auto) { return "MultiRZ"; })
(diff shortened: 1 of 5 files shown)


mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -482,11 +484,11 @@ void ResourceAnalysis::analyzeRegion(Region &region, ResourceResult &result, boo
...
-            quantum::SetStateOp, qref::SetStateOp, quantum::SetBasisStateOp, qref::SetBasisStateOp>(
-            op)) {
+    if (isa<quantum::CustomOp, qref::CustomOp, quantum::OperatorOp, qref::OperatorOp,
+            quantum::PauliRotOp, qref::PauliRotOp, quantum::GlobalPhaseOp, qref::GlobalPhaseOp,
+            quantum::MultiRZOp, qref::MultiRZOp, quantum::PCPhaseOp, qref::PCPhaseOp,
+            quantum::QubitUnitaryOp, qref::QubitUnitaryOp, quantum::SetStateOp, qref::SetStateOp,
+            quantum::SetBasisStateOp, qref::SetBasisStateOp>(op)) {
         std::string name = getGateOpName(op, isAdjoint);
         int nQubits = getGateQubitCount(op);
         int nParams = getGateParamCount(op);
```
jzaia18, line 494: Does `getGateParamCount` need to be updated? I don't think it would be returning the correct value here. @sengthai do you have more context on why we store different parameter counts as different entries in the `operators` dict?

jzaia18 (approved):
Looks good, nice work, just 1 comment for now. I've tagged @sengthai to review as well since he was the original author of this pass.

And thanks for removing our old `pytest.skip`s that we left in the code 🫣

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: I don't think we need to update it. Also, storing parameter is the requirement for constructing graph decomp Rule, example  `Rot(4,2)` and `Rot(4,3)` stay separate internally. However, later on in Resource Analysis, it will be merge so `Rot(4,2) and Rot(4,3)` both become `Rot(4)`.

sengthai (approved):
Thanks @albi3ro! 💯

PR #2994, opened 2026-07-03
Rename mac wheels away from `universal2` when uploading the nightly

paul0403:
**Context:**
After building mac wheels, we need to rename from universal2 to just arm64
When updating the wheels building runner to macos 15, we didn't update the default name to search for when renaming.


JerryChen97 (approved):

dime10 (approved):

dime10:
Actually, could you remove the internal changelog entry about the mac runner since we don't document document CI setup in the release notes? 

Also, does the same change need to apply to the regular release as well?

paul0403:
> Also, does the same change need to apply to the regular release as well?

For regular releases, the release manager would download the files from CI and rename manually :sweat_smile: kinda strange but this is how it's done, I guess to force a human in the loop to check wheels using `twine`, check sha256 hashes, upload manually so pypi can check account credentials and api keys, etc

PR #2998, opened 2026-07-07
Fix `getAllParams` interface method on operator op

paul0403:
**Context:**
A while ago in #2969 , the string property for the gate name of `operator` op got changed to an attribute. 
This broke the implementation of `getAllParams()` interface method, since the method `getODSOperandIndexAndLength(unsigned index)` expects the index of ODS operands only, completely ignoring attributes. 

The addition of an attribute pushes every index down by one. We fix the index.
Also added an interface test for mlir capi access.

**Benefits:**
Correctness.


dime10 (approved):
Good catch, thanks Paul!

kipawaa (approved):

PR #2999, opened 2026-07-07
Represent Operators' static data as a dict in graph-decomposition

maliasadi:
**Context:**

Operators specialized by compile-time static data (e.g.. `pauli_word` or `quantum.operator` static data dict) are currently handled by mingling that data as metadata into `operator` and `target_gate` names (`"paulirotYXZ"`) and is utilized in the decomposition graph. The graph already supports static data [here](https://github.com/PennyLaneAI/catalyst/blob/c30fe9a9ed35f83e9f83d65cdbf5f2bb2b928215/mlir/lib/Quantum/Transforms/DecompGraphSolver/DGTypes.hpp#L60). This PR **prototypes** a single canonical repr. of `static_data` for decomposition rules and the graph-decomposition pass pipeline. 

**Benefits:**
- Frontend and rule capture/lowering are unchanged. no updates needed there   
- The graph-solver, registry system for decomp rules, and the matcher logic in graph-decomposition are all synced on the name of ops with static_data (`PauliRot` :heavy_check_mark: ) 
- No more name mangling in the graph-decomp system and parsing substrings of rules names anymore.
 
**Possible Drawbacks:**
- `MultiRZ` still is a special case. I think this is a right call for now but we can revisit it later. Note that `MultiRZ` key is `num_wires`, so it's not really a static data, so it keeps its own branch in the pass.
- Integration with `quantum.operator` will be done in a separate PR (@kipawaa's workng on it)

**Related GitHub Issues:**

### DIFF
--- a/mlir/include/Quantum/Transforms/DecompStaticData.h
+++ b/mlir/include/Quantum/Transforms/DecompStaticData.h
@@ -1,0 +1,106 @@
+// Copyright 2026 Xanadu Quantum Technologies Inc.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+
+//     http://www.apache.org/licenses/LICENSE-2.0
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * @file DecompStaticData.h
+ *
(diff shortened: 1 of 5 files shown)


dime10:
Hi @maliasadi, I'm not sure if you're up to date on this we just discussed a canonical way to generate graph decomposition keys from operators after our meeting on Tuesday. I posted a brief update here: https://github.com/PennyLaneAI/catalyst/pull/2983#discussion_r3539642702, but @kipawaa can also chime in since he was taking charge of that.
I just to make sure this is inline with that and we are not duplicating efforts or implementing conflicting strategies.

maliasadi:
@dime10 Thanks David! I've discussed this with River the other day. Compared to his work, this PR provides a more generic fix for the operator's static data, which is also aligned with the serialization required by the graph solver.

PR #3001, opened 2026-07-08
Support decomposition to/from Adjoint Ops to the GraphSolver

maliasadi:
**Context:**
The decomposition graph solver could not decompose Adjoint ops. This PR adds this support to the graph solver to handle them, mirroring the two decomposition families used by PL:

1. Decompose `Adjoint(Op)` directly: Rules registered against the adjoint operator itself: `self_adjoint` and `adjoint_rotation` funcs in PL. and other custom rules like `_adjoint_rot`, etc. These were already supported by the graph. A `RuleNode` whose output has `adjoint = true` is registered and solved like any other rule; so no new logic was needed to the solver; they just need to be provided as rules. I'll take care of this in a follow-up PR to integrate the support with the graph-decomposition pass.

2. Decompose Op, then adjoint each produced gate: From each base rule `Op`, the graph builder should synthesize `Adjoint(Op)` with the same multiplicities. I assumed that gate-order reversal doesn't affect resources, so the graph only needs the `Adjoint(Op)` counts; the actual reversal happens in the follow-up PR when integrating with the pass.

Both pathways coexist in the graph, and the solver picks the cheapest decomp pathway! :tada: 

**Benefits:**
- Supporting Adjoint(Op) decomposition and resource estimation

**Possible Drawbacks:**
- For circuits with many Adjoints, the size of the graph will be doubled as we need to add mirrored rules for each `Adjoint(Op)` and track them all throughout multiple decomp pathways.
- I'm using explicit rules for self-adjoint gates! This can be further optimized later after integration with the graph-decomposition pass.

**Related GitHub Issues:**
[sc-125400]

### DIFF
--- a/mlir/lib/Quantum/Transforms/GraphDecomposition/DecompGraphSolver/DGTypes.hpp
+++ b/mlir/lib/Quantum/Transforms/GraphDecomposition/DecompGraphSolver/DGTypes.hpp
@@ -53,39 +53,16 @@ namespace DecompGraph::Core {
  * when adding support for operators with dynamic numbers of wires/params.
  */
 struct OperatorNode {
-    std::string name;
-    int numWires{-1};
-    int numParams{-1};
+    std::string id;
     bool adjoint{false};
 
-    // Optional static arguments for operators that require additional data.
+    // optional params, primarily for debug use
+    std::string name{""};
+    int numWires{-1};
+    int numParams{-1};
     std::unordered_map<std::string, std::string> staticNamedArgs{};
-    std::string id{""};
 
-    bool operator==(const OperatorNode &other) const
-    {
-        // id match
-        if (!id.empty() && !other.id.empty() && id == other.id) {
-            return true;
-        }
-        // legacy fallback if either op is missing ID
-
-        // For equality, we consider numWires and numParams conditionally equal
-        // if they are not set to -1 (which indicates a wildcard that can match any value).
-        const bool default_wires =
-            (numWires == -1 || other.numWires == -1 || numWires == other.numWires);
-        const bool default_params =
-            (numParams == -1 || other.numParams == -1 || numParams == other.numParams);
-
-        // Static arguments are optional: if either side has no static args, they
-        // are treated as matching (wildcard). When both sides provide entries, the maps must
-        // be equal element-wise for the operators to be considered equivalent.
-        const bool static_args_match = staticNamedArgs.empty() || other.staticNamedArgs.empty() ||
-                                       staticNamedArgs == other.staticNamedArgs;
-
-        return name == other.name && default_wires && default_params && adjoint == other.adjoint &&
-               static_args_match;
-    }
+    bool operator==(const OperatorNode &other) const { return id == other.id; }
     bool operator!=(const OperatorNode &other) const { return !(*this == other); }
 };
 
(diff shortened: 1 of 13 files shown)


mlir/lib/Quantum/Transforms/GraphDecomposition/DecompGraphSolver/DGTypes.hpp
```diff
@@ -140,8 +140,9 @@ struct RuleTerm {
  * - Default: The default rule for decomposing an operator as defined in the decomposition graph.
  * - Fixed: A fixed rule that cannot be changed or overridden by the solver.
  * - Alternative: An alternative rule that can be used in place of the default rule.
+ * - AdjointGenerated: A rule synthesized by adjointing a base decomposition rule.
  */
-enum class RuleOrigin : uint8_t { Default = 0, Fixed = 1, Alternative = 2 };
+enum class RuleOrigin : uint8_t { Default = 0, Fixed = 1, Alternative = 2, AdjointGenerated = 3 };
```
albi3ro, line 131: Should we differentiate between adjoint genereated from default, adjoint generated from fixed and adjoint generated from alternative?

mlir/lib/Quantum/Transforms/GraphDecomposition/DecompGraphSolver/DGTypes.hpp
```diff
@@ -185,6 +186,38 @@ using FixedDecomps = std::unordered_map<OperatorNode, RuleNode, OperatorNodeHash
...
+ *
+ * Given a rule `output -> {inputs}`, produces `Adjoint(output) -> {Adjoint(input), ...}`
+ * with the same multiplicities: the adjoint of a decomposition is obtained by adjointing
+ * every produced gate (and reversing their order, which does not affect resource/cost counting).
+ */
+inline RuleNode makeAdjointRule(const RuleNode &base)
+{
+    RuleNode adj;
+    adj.name = base.name + "_adjoint";
+    adj.output = makeAdjoint(base.output);
```
albi3ro, line 215: Do we have to worry about this in-place mutating base.output?

maliasadi:
Rebased to include #3046 changes in the solver

mlir/lib/Quantum/Transforms/GraphDecomposition/DecompGraphSolver/DGTypes.hpp
maliasadi: No we don't need to consider those cases. Fixed/Alternative are consumed by the builder to enforced which rules exist for an op -- the solver works without knowing about the origin of these rule! `AdjointGenerated` is also consumed by the builder but it's needed in the solver as the solver needs to consider both pathways (in the ADR) and to propagate rules that are built by the solver using `makeAdjointRule` (Pathway 2 in the ADR).

mlir/lib/Quantum/Transforms/GraphDecomposition/DecompGraphSolver/DGTypes.hpp
maliasadi: No. `makeAdjoint` takes operators by "value"; it operates on a copy and returns it so the original op remains untouched.

kipawaa:
@maliasadi can we close the PRs in this stack now that #3127 is merged and #3129 is in review?

PR #3008, opened 2026-07-09
feat: add pre-commit hook 🎉 

andrijapau:
**Context:**

Developers have been wanting it, but we're all too busy (or lazy) to add it ourselves 😆 

Copy the important parts of PL's pre-commit hook over to Catalyst.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,41 @@
+repos:
+-   repo: https://github.com/gitleaks/gitleaks
+    rev: v8.28.0
+    hooks:
+      - id: gitleaks
+        name: "🔒 gitleaks · Detect hardcoded secrets"
+repos:
+-   repo: https://github.com/psf/black
+    rev: 26.3.1
+    hooks:
+    - id: black
+      name: "🐍 black · Format code"
+      args: [
+              "--config",
+              "./pyproject.toml",
+            ]
+-   repo: https://github.com/PyCQA/isort
+    rev: 7.0.0
+    hooks:
+    - id: isort
+      name: "🐍 isort · Sort imports"
+      args:
+        [
+          "--settings-path",
+          "./pyproject.toml",
+        ]
+- repo: local
+  hooks:
+    - id: pylint
+      name: "🐍 pylint · Lint frontend"
+      entry: pylint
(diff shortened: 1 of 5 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,41 @@
+repos:
+-   repo: https://github.com/gitleaks/gitleaks
+    rev: v8.28.0
+    hooks:
+      - id: gitleaks
+        name: "🔒 gitleaks · Detect hardcoded secrets"
+repos:
+-   repo: https://github.com/psf/black
+    rev: 26.3.1
```
andrijapau, line 9: Instead of hard coding this (must match requirements.txt), we can optionally set,
```suggestion
    language: system
```
which picks up whatever version the user locally has installed. Up to you!

.pre-commit-config.yaml
andrijapau, line 1: Bulk is copied from PL

dime10 (approved):
Perfect thanks :)

.pre-commit-config.yaml
```diff
@@ -0,0 +1,41 @@
...
+    hooks:
+    - id: black
+      name: "🐍 black · Format code"
+      args: [
+              "--config",
+              "./pyproject.toml",
```
dime10, line 15: Same question I asked in a prior PR: 

Will this work if git commit is run from a subdirectory of the repo root?

.pre-commit-config.yaml
dime10: yeah I guess auto detection would be nice, since the dev environment would (should) have been set up with requirements.txt

.pre-commit-config.yaml
andrijapau: Yes, precommit should run from top level of the directory. No concern for relative paths like that!

kipawaa (approved):
Oh this is going to be so nice 😌 thank you!

.pre-commit-config.yaml
andrijapau: I'm going to just leave it as is. If we find that keeping them in sync is annoying, we can revisit this. We've been doing it like this in PL and haven't had any issues!

PR #3012, opened 2026-07-10
Get a proper path to the python shared library for the QPD plugin 

maliasadi:
**Context:**
The current `get_libpython_path()` trusts `LDLIBRARY`, and if it finds nothing, then it returns an empty string. This causes the QPD plugin fails on envs when `LDLIBRARY` sets to the static archived python library instead of the shared one (e.g., on conda envs: `LDLIBRARY = libpython3.14.a`). 

This PR fixes this with a minimal patch to `get_libpython_path()` that it now only accepts `LDLIBRARY` when it's actually a shared object, and otherwise falls back to the conventional `libpython{LDVERSION}.so` in `LIBDIR`. This fixes the issues on conda envs while preserving the current behaviour on other envs.

**Benefits:**
- Fix the QPD plugin on conda envs.

**Possible Drawbacks:**
n/a

**Related GitHub Issues:**
[sc-124573]

### DIFF
--- a/frontend/catalyst/utils/runtime_environment.py
+++ b/frontend/catalyst/utils/runtime_environment.py
@@ -14,7 +14,6 @@
 
 """Utility code for keeping paths."""
 
-import glob
 import importlib.util
 import os
 import os.path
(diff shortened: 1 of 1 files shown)


frontend/catalyst/utils/runtime_environment.py
```diff
@@ -63,23 +63,45 @@ def _load_config(filename, attr):
...
+    """Return the path to the python *shared* library, or empty string if failed to find.
+
+    The QPD plugin uses dlopen by the standalone ``catalyst`` compiler process, which does
+    not embed Python, so it needs ``libpython`` loaded with ``RTLD_GLOBAL`` to resolve CPython
+    symbols. So, this path needs to return a shared object instead of the static python library.
+
+    ``LDLIBRARY`` in ``sysconfig`` is not reliably the shared library: some builds (notably
+    conda envs) report the static archive ``libpythonX.Y.a`` there even though a ``.so``
+    is present alongside it. We accept ``LDLIBRARY`` only when it is a shared object
+    that exists, and otherwise fall back to locating the shared library in ``LIBDIR``.
```
dime10: better suited as a code comment than part of the function interface documentation

dime10 (approved):

frontend/catalyst/utils/runtime_environment.py
maliasadi: Updated [73304ef](https://github.com/PennyLaneAI/catalyst/pull/3012/commits/73304ef2b879996e6f798d2936b73dff8226dd11)

kipawaa (approved):
Thanks for the fix @maliasadi !

Issue #1464, opened 2025-01-15
[BUG] parameterizing coefficients of `qml.dot` fails with qjit


bnishanth16:
Experienced `TracerBoolConversionError` when trying to parameterize coeffients of qml.dot with qjit. Minimal example code below.

**Expected behavior:** (without qjit)
```
import pennylane as qml
import jax.numpy as jnp
from catalyst import qjit

@qml.qnode(qml.device('lightning.qubit', wires=1))
def circuit(x):
    coeffs = [x]
    ops = [qml.PauliX(0)]
    H = qml.dot(coeffs, ops) # H = qml.Hamiltonian(coeffs, ops)
    qml.evolve(H)
    return qml.expval(qml.Z(0))

print(circuit(jnp.array([0.5])))
```

**Output:**
```
>>>  [0.54030231]
```

**Actual Behaviour:** (with qjit)
```
import pennylane as qml
import jax.numpy as jnp
from catalyst import qjit

@qjit(autograph=True)
@qml.qnode(qml.device('lightning.qubit', wires=1))
def circuit(x):
    coeffs = [x]
    ops = [qml.PauliX(0)]
    H = qml.dot(coeffs, ops) # H = qml.Hamiltonian(coeffs, ops)
    qml.evolve(H)
    return qml.expval(qml.Z(0))

print(circuit(jnp.array([0.5])))
```

**Output:**
```
---------------------------------------------------------------------------
TracerBoolConversionError                 Traceback (most recent call last)
Cell In[7], line 14
     11     qml.evolve(H)
     12     return qml.expval(qml.Z(0))
---> 14 print(circuit(jnp.array([0.5])))

File ~/anaconda3/envs/pennylane40/lib/python3.12/site-packages/pennylane/logging/decorators.py:61, in log_string_debug_func.<locals>.wrapper_entry(*args, **kwargs)
     54     s_caller = "::L".join(
     55         [str(i) for i in inspect.getouterframes(inspect.currentframe(), 2)[1][1:3]]
     56     )
     57     lgr.debug(
     58         f"Calling {f_string} from {s_caller}",
     59         **_debug_log_kwargs,
     60     )
---> 61 return func(*args, **kwargs)

File ~/anaconda3/envs/pennylane40/lib/python3.12/site-packages/catalyst/jit.py:521, in QJIT.__call__(self, *args, **kwargs)
    517         kwargs = {"static_argnums": self.compile_options.static_argnums, **kwargs}
[...]

PR #3013, opened 2026-07-12
Fix BufferizationStage IR blowup for runtime-parameterized Hamiltonians (Scalarize)

### LINKED ISSUE #2759: BufferizationStage IR amplification for runtime-coefficient Hamiltonians under @qjit
## Summary

When using `qml.TrotterProduct` with JAX-traced runtime coefficients inside `@qjit`, the BufferizationStage of the MLIR compilation pipeline produces a large IR size amplification. This is the dominant contributor to extreme compiler memory consumption (~19 GB compiler-subprocess RSS) for even small molecules.

With **fixed** (Python float) coefficients, BufferizationStage achieves ~1.0x amplification (constant-folding eliminates redundant IR). With **runtime** (JAX array) coefficients, the symbolic representation prevents constant-folding; for an H2 reference workload the IR grows from ~3.9 MB to ~60 MB at this single stage. The subsequent `MLIRToLLVMDialectConversion` then amplifies a further ~4.5x, yielding a total IR growth of ~62x (3.9 MB → 241 MB).

This makes runtime-parameterized Hamiltonians impractical for systems beyond the smallest molecules: H2 (15 Pauli terms) already consumes ~21 GB total peak; H3O+ (193 terms) extrapolates well beyond commodity 32 GB hardware.

The runtime-coefficient code path is the workaround community members are pointed to (cf. #1464) and is what allows electrostatic-embedding workflows (Monte Carlo solvation, VQE-style coefficient updates) to avoid full per-step recompilation. The memory cost at this single stage currently caps how large that workaround can scale.

**Related**: #1464 (qml.dot with runtime coefficients), #1507 (Hamiltonian aggregate-type lowering)

## Minimal reproducible example

```python
import numpy as np
import pennylane as qml
from jax import numpy as jnp

# --- Hamiltonian construction (numpy path, gives 15-term H2) ---
# qml.qchem.molecular_hamiltonian expects atomic units (Bohr).
# Note: passing the geometry as a JAX array currently disables term aggregation
# in molecular_hamiltonian and yields 47 SProd operands instead of 15. We use
# numpy here to keep the workload aligned with the measurements below.
ANGSTROM_TO_BOHR = 1.8897259886
symbols = ["H", "H"]
coords_angstrom = np.array([[0.0, 0.0, 0.0], [0.0, 0.0, 0.74]])
coords_bohr = (coords_angstrom * ANGSTROM_TO_BOHR).flatten()

H, n_qubits = qml.qchem.molecular_hamiltonian(
    symbols=symbols,
    coordinates=coords_bohr,
    mapping="jordan_wigner",
    active_electrons=2,
    active_orbitals=2,
)
assert len(H.operands) == 15  # H2 STO-3G / JW with active space (2e, 2o)

# coeffs is a JAX array, so it becomes a runtime tracer at compile time —
# this is the configuration that triggers the BufferizationStage amplification.
coeffs = jnp.array([float(op.scalar) for op in H.operands if isinstance(op, qml.ops.SProd)])
ops = [op.base for op in H.operands if isinstance(op, qml.ops.SProd)]

n_est = 4
n_trotter = 10
total_wires = n_qubits + n_est
dev = qml.device("lightning.qubit", wires=total_wires)

@qml.qjit(keep_intermediate=True)  # dump per-stage IR for diagnosis
@qml.qnode(dev)
def qpe_circuit(runtime_coeffs):
    H_rt = qml.dot(runtime_coeffs, ops)  # `ops` captured as compile-time closure
    for k in range(n_est):
        t = 2 ** (n_est - 1 - k)
        qml.ctrl(
            qml.adjoint(
                qml.TrotterProduct(
                    H_rt, time=t, n=n_trotter, order=2,
                    check_hermitian=False,  # required to admit JAX-tracer coeffs
                )
            ),
            control=n_qubits + k,
        )
    return qml.probs(wires=range(n_qubits, total_wires))

# Compilation consumes ~19 GB compiler-subprocess RSS (~21 GB total peak)
# for just 15 Pauli terms. Per-stage IR is written under the keep_intermediate
# directory; BufferizationStage shows the dominant ~9.9x growth.
result = qpe_circuit(coeffs)
```

## Measured data

**Environment**: PennyLane 0.44.0, Catalyst 0.14.0, JAX 0.7.1, Linux x86_64, 30.4 GB RAM.
**Measurement note**: numbers below were collected with `keep_intermediate=True` to obtain per-stage IR sizes; production runs without `keep_intermediate` show similar end-to-end memory and time.

### Per-stage IR size (H2, n_est=4, n_trotter=10, 15 Pauli terms)

| Compilation Stage | H_fixed (KB) | H_dynamic (KB) | Fixed amp. | Dynamic amp. |
|---|---|---|---|---|
| mlir (input) | 706 | 3,860 | — | — |
| QuantumCompilationStage | 1,629 | 3,782 | 2.3x | 1.0x |
| HLOLoweringStage | 696 | 6,101 | 0.4x | 1.6x |
| **BufferizationStage** | **696** | **60,320** | **1.0x** | **9.9x** |
| MLIRToLLVMDialectConversion | 2,487 | 273,442 | 3.6x | 4.5x |
| LLVMIRTranslation | 1,820 | 240,654 | 0.7x | 0.9x |
| **Total** | **706 → 1,820** | **3,860 → 240,654** | **2.6x** | **62.4x** |

### End-to-end compilation metrics

| Metric | H_fixed | H_dynamic | Ratio |
|---|---|---|---|
| Compile time | 27 s | 306 s | 11.2x |
| Python RSS | 1.2 GB | 2.2 GB | 1.9x |
| **Compiler subprocess RSS** | **0.9 GB** | **19.1 GB** | **21.4x** |
| Total peak (parent+child) | 2.0 GB | **21.3 GB** | 10.3x |

### System scaling (dynamic mode)

| System | Pauli Terms | IR ops (lower bound) | Memory | Status |
|---|---|---|---|---|
| H2 (4-bit) | 15 | 600 | 21.3 GB | Verified |
| H2 (8-bit) | 15 | 1,200 | OOM @ n_trotter ≥ 7 | Verified |
| H3O+ | 193 | ~7,720 | OOM (extrapolated) | Not attempted |

## Impact

This is currently a blocking issue for any quantum-chemistry workflow that requires runtime-parameterized Hamiltonians under Catalyst:

1. **QM/MM workflows**: Monte Carlo solvation dynamics requires updating Hamiltonian coefficients at each step (electrostatic embedding). Without runtime parameterization the alternative is full per-step recompilation (~67 s/step in our setup).
2. **VQE / variational methods**: any workflow where Hamiltonian coefficients depend on classical parameters.
3. **System-size ceiling**: even H2 (15 Pauli terms) consumes ~21 GB. H3O+ (193 terms) is infeasible on commodity hardware.

## Suggested directions

1. **Sparse / structured IR representation**: detect that runtime coefficients multiply structurally-identical operator blocks and factor them at BufferizationStage rather than expanding each symbolically.
2. **Lazy lowering**: defer constant-folding-dependent optimizations to JIT time when actual coefficient values are available.
3. **Pre-compilation IR-size estimate**: expose an API that estimates IR size before compilation so users can guard against OOM.
4. **Operator fusion**: combine repeated TrotterProduct blocks that share structure but differ only in the time parameter.

jk20342:
**Context:**
`qml.TrotterProduct` with runtime coefficients (via `qml.dot`) causes severe IR
amplification: tracing unrolls all Trotter steps, and every gate angle drags a chain of
single-element tensor ops through bufferization, each becoming an alloc plus copies.
For an H2 QPE workload this meant ~62x total IR growth and ~19 GB of compiler memory.

* New `scalarize-tensor-extracts` pass: folds `tensor.extract` through
  `extract_slice`/`collapse_shape`/small elementwise `linalg.generic`, turning gate-angle
  dataflow into scalar arithmetic.
* New `reroll-loops` pass: detects repeated op sequences via structural hashing and
  rebuilds them as `scf.for` loops.
* Default pipeline (HLO stage only): scalarize + `linalg-fuse-elementwise-ops` before
  `reroll-loops`, and a second scalarize after detensorization (which materializes the
  extract chains).

**Benefits:**
On the H2 QPE benchmark, before/after measured on the same machine, same build, same
input IR, differing only in the pipeline: compile time 1h41m -> ~25s (~240x), peak
memory 15.4 GB -> 1.2 GB (12.8x), final IR 18.4 MB -> 5.4 MB (3.4x), post-bufferization
IR 7.45 MB -> 2.64 MB (2.8x). Output probabilities are bit-identical to the old pipeline.

these baseline numbers differ from the issue report (306 s, 21.3 GB, 241 MB final
IR) because the issue predates current main, which already reduced the
baseline's bufferization blowup (60 MB -> 7.45 MB post-bufferization; input IR drifted
3.86 -> 4.69 MB), and these measurements use an assertions-enabled dev build, ~20x
slower in wall time than the release wheel the issue used. Ratios within this PR are
apples-to-apples; absolute times are not comparable to the issue's.

**Possible Drawbacks:**
Both passes are conservative (size caps, semantic verification, high `min-savings`
default) but rerolling may miss non-contiguous repeats; overhead on workloads without
the pattern is negligible.

**Related GitHub Issues:**
Fixes #2759. Related: #1464, #1507.


jk20342:
here is some benchmark code 

https://gist.github.com/jk20342/d6cc77c374522db0d59e646dfc4a04ae

and results through my docker container :]
```
config: mode=dyn n_est=4 n_trotter=10

compile time: 35.2 s
peak RSS: self=542 MB children=1206 MB

per-stage IR size:
  0_qpe_circuit.mlir                                          4.69 MB
  1_AfterQuantumCompilationStage.mlir                         4.59 MB  (1.0x)
  2_AfterHLOLoweringStage.mlir                                2.82 MB  (0.6x)
  3_AfterGradientLoweringStage.mlir                           2.82 MB  (1.0x)
  4_AfterBufferizationStage.mlir                              2.66 MB  (0.9x)
  5_AfterMLIRToLLVMDialectConversion.mlir                     7.32 MB  (2.8x)
  6_AfterLLVMIRTranslation.ll                                 5.37 MB  (0.7x)

jk20342:
This could be amplified lower I think but there is upstream issue on LLVM that I bumped to try to get more traction on

dime10:
Thanks for this contribution @jk20342, very nice results 💯 I love the scalarization pass, which is beneficial in more ways than performance!

I'd be curious to know the effect size of the scalarization vs the rerolling, as using loops is known to have massive impact on the compile time. Automatically detecting unrolled loops is definitely nice for programs that were suboptimally written, although we are continuously updating PL library code to make sure programs use loops to begin with as well.

jk20342:
> Thanks for this contribution @jk20342, very nice results 💯 I love the scalarization pass, which is beneficial in more ways than performance!
> 
> I'd be curious to know the effect size of the scalarization vs the rerolling, as using loops is known to have massive impact on the compile time. Automatically detecting unrolled loops is definitely nice for programs that were suboptimally written, although we are continuously updating PL library code to make sure programs use loops to begin with as well.

@dime10 when i tested it the reroll improved ~1.2x wall time ~1.35x memory and ~1.27x final IR and a 5x reduction in gate-op volume over just scalarize could remove it, if you think this will be fixed on PL directly though

mehrdad2m:
Thanks @jk20342, this is a very impressive work! It seems to me that the `reroll-loops` is an independent work but still requires the cleanup `scalarize-tensor-extracts` provides to get the performance improvement, is that correct? If so, I would appreciate if we could split the work into two PRs (one for each pass). This would really help us in the review process to be focused on one pass and also be able to benchmark the effects of each independently.

jk20342:
> Thanks @jk20342, this is a very impressive work! It seems to me that the `reroll-loops` is an independent work but still requires the cleanup `scalarize-tensor-extracts` provides to get the performance improvement, is that correct? If so, I would appreciate if we could split the work into two PRs (one for each pass). This would really help us in the review process to be focused on one pass and also be able to benchmark the effects of each independently. 
> 
> 

Yep I can split it up

jk20342:
@mehrdad2m made this PR the scalarize https://github.com/PennyLaneAI/catalyst/pull/3036 is reroll now :)

jk20342:
bump

jk20342:
bump

mehrdad2m:
Hi @jk20342, apologies for the slow turnaround here. The team has been heads-down on a few other things which limited our bandwidth for review. I just want to assure you that your PRs are definitely on our radar and we'll give them a proper review as soon as we can. Thanks for your patience!

jk20342:
all good :)

PR #3017, opened 2026-07-13
Add state and restored to AllocOp

albi3ro:
**Context:**

In `qp.allocate`, users can specify whether or not they need the new qubits to be in a zero state with `state="zero"` or `state="any"`.  They can also specify whether or not the qubits are returned to their initial state.  This allows dynamically allocated qubits to be reused for another allocation after deallocation.

To be able to replace `pennylane.estimator` functionality, we need to be able to track the different types of wire allocations.
 
Just adds optional `state` and `restored` properties to `AllocOp` that default to `state="zero"` and `restored=True`.

**Benefits:**

We can start tracking the different types of wire allocations.

**Possible Drawbacks:**

**Related GitHub Issues:**


dime10:
Adding these might require additional validation in other places (any current users of the alloc op essentially), because in general those mechanisms are not going to be aware of these attributes and won't know what to do with it (likely just ignoring it, otherwise failing). Neither is great, but especially a silent ignore would be bad since it will lead to incorrect results.

PR #3018, opened 2026-07-13
Cond primitive's default "else" branch no longer has a literal `True` predicate 

paul0403:
**Context:**
We recently removed the literal `True` predicate on the default else branch of the cond plxpr primitive. https://github.com/PennyLaneAI/pennylane/pull/9815

We update the lowering rule correspondingly.

**Benefits:**
Things are correct.


kipawaa (approved):
Nice!

albi3ro (approved):

PR #3020, opened 2026-07-14
fix: reject quantum.adjoint input in ppr-to-ppm with a clear error

### LINKED ISSUE #2934: `adjoint-lowering` pass is unable to handle non-`Quantum` dialect ops
Following on from #2932, the same test case reveals a different error due to the algorithm being written to only handle quantum dialect ops. If the user converts to the PBC dialect, the pass will fail.

```py
@qp.qjit
@qp.transforms.ppr_to_ppm
@qp.transforms.to_ppr
def circuit():
   qp.adjoint(lambda: qp.S(0))()
   return qp.probs()
```

```
catalyst.utils.exceptions.CompileError: catalyst failed with error code -6: Assertion failed: (result && "expected 'from' to be contained within the map"), function lookup, file IRMapping.h, line 74.
```

mvanhorn:
### Before submitting

**Context:**

As reported in #2934, applying the `to_ppr` and then `ppr-to-ppm` transforms to a circuit that
contains a functional adjoint (e.g. `qp.adjoint(lambda: qp.S(0))()`) crashes the compiler with a
raw MLIR assertion:

```
Assertion failed: (result && "expected 'from' to be contained within the map"),
function lookup, file IRMapping.h, line 74
```

surfaced to the user as `CompileError: catalyst failed with error code -6`.

The root cause is a pass-ordering interaction. The default `quantum-compilation-pipeline` runs
`adjoint-lowering` automatically after the user transforms, and that pass reverses a `quantum.adjoint`
region by walking its ops. `adjoint-lowering` can reverse `pbc.ppr` ops (see
`mlir/test/PBC/AdjointTest.mlir`), but once `ppr-to-ppm` has turned the region's PPRs into Pauli
product measurement (`pbc.ppm`) ops, those measurements are not reversible, so the reversal hits the
`IRMapping` assertion instead of reporting anything actionable.

Following the maintainer's direction on the issue to tighten the pass's input requirements at the
start of the PBC conversion rather than failing later, this adds an input-validation guard at the
very start of `PPRToPPMPass::runOnOperation`, before any patterns are applied. The pass now walks
its operation for a `quantum.adjoint` op and, if one is found, emits a clear error explaining that
`ppr-to-ppm` cannot be applied inside a `quantum.adjoint` region because the resulting Pauli product
measurements cannot be adjoint-lowered, then signals pass failure.

The guard is scoped to `ppr-to-ppm` only; `to_ppr` produces PPRs, which `adjoint-lowering` handles
correctly, and is left unchanged.

A lit test is added to `mlir/test/PBC/PPRToPPM.mlir` that wraps a `pbc.ppr` in a `quantum.adjoint`
region and asserts, via `-verify-diagnostics`, that the pass emits the new diagnostic and fails
cleanly instead of asserting.

**Benefits:**

A malformed input that previously aborted the compiler with an internal assertion (`error code -6`)
now produces a clear, actionable diagnostic that names the unsupported construct and the location of
the offending op.

**Possible Drawbacks:**

The guard is deliberately conservative: any `quantum.adjoint` op reaching `ppr-to-ppm` is rejected,
matching the maintainer's "no unexpected ops through the pass" framing, rather than attempting to
detect whether a given adjoint region would in fact receive generated measurements.

**Related GitHub Issues:**

Fixes #2934

---

**AI tool disclosure** (per the PennyLane AI Tool Use Policy): this contribution was developed with AI assistance (Claude Code, and Codex for follow-up revisions). I reviewed all generated code and text, wrote this description myself, and am accountable for the work as its author. Happy to answer questions about the implementation during review.

### DIFF
--- a/doc/code/dialects/PBC/PBCPasses.md
+++ b/doc/code/dialects/PBC/PBCPasses.md
@@ -59,6 +59,19 @@ _Convert CliffordT operations to Pauli Product Measurement operations._
 
 _Count specs in Pauli Product Measurement operations._
 
+### `-ppr-to-ppm`
+
+_Convert PPRotation operations to Pauli Product Measurement operations._
+
+This pass first invokes `adjoint-lowering` as a nested pipeline so that PPRotation operations inside `quantum.adjoint` regions are reversed before they are converted to Pauli Product Measurements. Quantum allocation, deallocation, register access, and global phase operations may remain in the resulting PBC representation; other Quantum dialect operations must be converted before this pass.
+
+#### Options
+
+```
+-decompose-method : Decomposition method to use
+-avoid-y-measure  : Avoid Pauli-Y measurements for Clifford rotations. Rather than performing a Pauli-Y measurement for Clifford rotations (sometimes more costly), a Y state is used instead (requires Y state preparation).
+```
+
 ### `-to-ppr`
 
 _Convert quantum dialects to the PBC dialect._
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -178,6 +178,12 @@
 
 <h3>Bug fixes 🐛</h3>
 
+* The `ppr-to-ppm` pass now raises a clear error when its input contains a `quantum.adjoint`
+  region, instead of failing later with a raw compiler assertion. The generated Pauli product
+  measurements cannot be reversed by the `adjoint-lowering` pass, so this input is now rejected
+  up front.
+  [(#2934)](https://github.com/PennyLaneAI/catalyst/pull/2934)
+
 * Fixed a bug where the `ResourceAnalysis` pass only analyzed functions directly contained in
   the top-level module. Functions inside nested modules, such as kernels called through
   `catalyst.launch_kernel`, are now included in the output.
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -178,6 +178,11 @@
 
 <h3>Bug fixes 🐛</h3>
 
+* The `adjoint-lowering` pass now raises a clear error when a `quantum.adjoint` region contains an
+  unsupported operation that produces or consumes quantum values, including operations from
+  non-Quantum dialects, instead of failing with a raw compiler assertion.
+  [(#2934)](https://github.com/PennyLaneAI/catalyst/issues/2934)
+
 * The `ppr-to-ppm` pass now raises a clear error when its input contains a `quantum.adjoint`
   region, instead of failing later with a raw compiler assertion. The generated Pauli product
   measurements cannot be reversed by the `adjoint-lowering` pass, so this input is now rejected
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -178,15 +178,9 @@
 
 <h3>Bug fixes 🐛</h3>
 
-* The `adjoint-lowering` pass now raises a clear error when a `quantum.adjoint` region contains an
-  unsupported operation that produces or consumes quantum values, including operations from
-  non-Quantum dialects, instead of failing with a raw compiler assertion.
-  [(#2934)](https://github.com/PennyLaneAI/catalyst/issues/2934)
-
-* The `ppr-to-ppm` pass now raises a clear error when its input contains a `quantum.adjoint`
-  region, instead of failing later with a raw compiler assertion. The generated Pauli product
-  measurements cannot be reversed by the `adjoint-lowering` pass, so this input is now rejected
-  up front.
+* The `ppr-to-ppm` pass now supports PPRotation operations inside `quantum.adjoint` regions by
+  invoking `adjoint-lowering` before converting the reversed rotations to Pauli product
+  measurements.
   [(#2934)](https://github.com/PennyLaneAI/catalyst/pull/2934)
 
 * Fixed a bug where the `ResourceAnalysis` pass only analyzed functions directly contained in
--- a/mlir/include/PBC/Transforms/Passes.td
+++ b/mlir/include/PBC/Transforms/Passes.td
@@ -120,7 +120,25 @@ def DecomposeCliffordPPRPass : Pass<"decompose-clifford-ppr"> {
 def PPRToPPMPass : Pass<"ppr-to-ppm"> {
     let summary = "Convert PPRotation operations to Pauli Product Measurement operations.";
 
-    let dependentDialects = [ "catalyst::pbc::PBCDialect", "scf::SCFDialect" ];
+    let description = [{
+        This pass first invokes `adjoint-lowering` as a nested pipeline so that
+        PPRotation operations inside `quantum.adjoint` regions are reversed before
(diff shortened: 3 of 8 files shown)


mehrdad2m:
Hi @mvanhorn, welcome to Catalyst, and thanks for the PR! 🎉 

The team will review shortly. The changes looks good overall, but just one thing to flag for discussion. While I think the input rejection at the `ppr-to-ppm` is nice to have, I am wondering if we could also raise a better error at the adjoint lowering pass as well since this pattern could happen with other user passes that would introduce incompatible ops for adjoint lowering. tagging @dime10 as well since he suggested the direction. 

Also note that to be compliant with our [policy](https://github.com/PennyLaneAI/catalyst/blob/main/AI_Policy.md) you must label the AI agent used to assist you in the PR description.

mvanhorn:
Good call, and agreed the adjoint-lowering pass is the right place for the general fix. Pushed in 1010942: the reverse-pass dispatch now emits a clear error for any op it leaves unhandled that produces or consumes quantum values, including non-Quantum-dialect ops like pbc.ppm, instead of falling through to the raw IRMapping assertion. Purely classical ops are still skipped, so it only fires on ops that actually participate in the quantum data flow. That covers the case @dime10 raised where another user pass introduces incompatible ops, not just ppr-to-ppm. I kept the ppr-to-ppm guard too since it gives an earlier, more specific message. Added a lit test with a pbc.ppm inside a quantum.adjoint region asserting the diagnostic, plus a changelog entry. clang-format is clean; I could not run the MLIR lit/build locally (no LLVM build in my environment), so I am relying on CI for that. Also added the AI-tool disclosure to the PR description per your policy.

dime10:
Thank you for working on this issue! I think where we landed on is that the PBC conversion needs to be better built out as a change in _representation_: from the generic gate model (quantum dialect) to the pauli-based representation (pbc dialect). Right it now, it is a basic pass that targets PPR gates from certain quantum ops. 

Adding a narrow check for adjoint op is helpful, but doesn't exactly fix the broader issue. Roughly, I had the following in mind:
- reframe the pass (name, docs, etc) as a full semantic conversion from the quantum to the pbc dialect
  - it doesn't necessarily have to use the dialect conversion infrastructure, although it can (it's know to be slower than "manual" passes, but we still use it because it is easy to work with)
- use an allowlist for operations we know how to convert (e.g. `quantum.custom`), or can safely remain in the IR (e.g. `quantum.alloc`)
  - this is counter to a blocklist approach where we explicitly mark certain ops as not allowed (e.g. `quantum.adjoint`)
  - the reason the allowlist is better is because a blocklist needs to be kept up to date regularly (whenever the quantum dialect has new operations), even when those ops are unrelated to PBC (which means the author will not think of the PBC conversion blocklist)
- this part may require some discussion, but we could invoke any necessary steps required as part of the conversion automatically, for example the adjoint lowering pass
  - can be done as a nested pass invocation

mvanhorn:
That framing makes sense to me, and I'd rather build it the right way than bolt adjoint on as a special case. Reworking it as a full quantum -> pbc semantic conversion (renaming the pass and updating the docs to match) sounds like the right direction.

On allowlist vs blocklist - agreed, and for exactly the reason you gave: a blocklist silently rots every time the quantum dialect gains an op, since nobody adding a new op is thinking about the PBC conversion. I'll switch to an allowlist of ops we know how to convert (`quantum.custom`) or can safely leave in place (`quantum.alloc`), and reject anything else by default.

The part I'd like your steer on is the automatic conversion steps:
1. For adjoint specifically, do you want the pass to invoke adjoint lowering itself as a nested pass, or to assume adjoints are already lowered upstream and just error on a `quantum.adjoint` it sees? A nested invocation is nicer for callers but couples this pass to the adjoint-lowering pipeline.
2. Build it on the dialect conversion framework, or a manual pass? You mentioned the framework is slower but easier to work with - I'd lean manual for a conversion this targeted, but happy to use the framework if you'd rather keep it consistent with the rest of the codebase.

Once I know your preference on those two I'll rework the pass accordingly.

dime10:
@mehrdad2m Any thoughts here?

mehrdad2m:
> @mehrdad2m Any thoughts here?
Thanks both,
I agree with the overall framework discussed here. on the open questions:

- On adjoint, I think the nested approach is nice to have as it makes the transforms self-contained. The only downside is that nesting passes in this pipeline has the potential to confuse users who are modifying the pipeline, but as long as it is well documented, I think it should be fine.

- On the pass framework, I don't have a strong preference here. However, I slightly lean toward the dialect conversion framework as it will make the maintenance easier and natively support the allowlist discussed. We could also consider potentially using the no rollback mode in future if the performance becomes an issue, though that has its own cons. That being said, I am fine with the manual pass as well as long as the pass implementation is simple enough to maintain.

mvanhorn:
Thanks @mehrdad2m and @dime10. Plan based on that: this PR stays scoped to the clear error on quantum.adjoint input, and I'll follow up with the nested-approach implementation as a separate PR - using the dialect conversion framework per your lean, with pipeline docs covering the nesting so modifiers aren't surprised. I'll link the follow-up here when it's up.

dime10:
> Thanks @mehrdad2m and @dime10. Plan based on that: this PR stays scoped to the clear error on quantum.adjoint input, and I'll follow up with the nested-approach implementation as a separate PR - using the dialect conversion framework per your lean, with pipeline docs covering the nesting so modifiers aren't surprised. I'll link the follow-up here when it's up.

I don't think we need this PR in the current state since we're moving to the proper conversion pass, saves you some trouble :) Feel free to reuse this PR for the latter or open a new one

mvanhorn:
Reworked this from the clear-error approach into the nested dialect conversion you asked for. Pushed ac6866b.

`ppr-to-ppm` now runs `adjoint-lowering` as a nested pipeline before conversion, so PPRotation operations inside a `quantum.adjoint` region are reversed rather than rejected. The conversion target keeps `quantum.alloc`, `alloc_qb`, `dealloc`, `dealloc_qb`, `extract`, `insert`, and `global_phase` legal and requires everything else in the Quantum dialect to be converted; PPRotation itself is dynamically legal when it is neither a pi/4 rotation nor an unconditioned non-Clifford one.

The changelog entry is rewritten accordingly: the two old "now raises a clear error" bullets are replaced by one describing the actual behaviour, and `PBCPasses.md` documents the nested-pipeline dependency and the remaining legal Quantum ops.

Built and tested locally against the pinned LLVM (8f26458): all 18 `mlir/test/PBC` lit tests pass, including the two this PR changes, `AdjointTest.mlir` and `PPRToPPM.mlir`.

Two caveats on my local run, so you know what CI still has to cover. I built `quantum-opt` only, with `QUANTUM_ENABLE_BINDINGS_PYTHON=OFF`, so the frontend Python tests did not run, and neither did the `cli/*` tests (no `catalyst-cli` binary) or the `Quantum/GraphDecomposition/*` tests (no `test_rules.mlirbc`). Those 12 failures are missing-artifact failures from my partial build, not behaviour. Everything that `quantum-opt` alone can exercise passes.

PR #3021, opened 2026-07-14
Generic decomp rule lowering

kipawaa:
**Context:**
Currently, multiple pathways exist for lowering a decomposition rule to MLIR. This increases maintenance burden and increases the surface for error and inconsistencies.
The current python-decomposition wrapper provides limited functionality, in particular it requires significant parsing after lowering to provide resources and target ID. 

Improves the generic `python_decomposition_wrapper` to support precompiled, trace-time and on-demand decomposition rule lowering.

**Benefits:**
Consistent, unified decomposition rule lowering across all available timings and reduced maintenance burden.

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-122018]
[sc-122043]
[sc-122041]


frontend/catalyst/device/python_decompositions.py
```diff
@@ -42,25 +22,134 @@
...
+    decomp_rules = list(qp.decomposition.list_decomps(op_name))
+
+    # map rules to resource resources, in a more generic format
+
+    name_to_resources = {}
+    for rule in decomp_rules:
+        # TODO: not all PL ops have been migrated to the operator 2 format expected by mlir graph
+        # decomp This means some rules will fail the python callback compilation. When migration is
+        # complete, remove the try-except.
+        try:
```
paul0403, line 135: This inner try-except around the op id generation means we can remove the original try-except around the qjit right?

kipawaa: It's more-so around the `compute_resources`, but yes! This way we can skip incompatible rules without skipping the op entirely

PR #3022, opened 2026-07-14
Implement decomposable gate interface methods for other gate ops

paul0403:
**Context:**
Following up on #2983 , we implement the decomposable op interface for all other gate ops in the quantum dialect.

Also, move all graph decomp lit tests from `mlir` folder to `frontend` folder, so they are properly connected with python decomp rule callbacks. 

**Benefits:**
All gates representable in the quantum dialect can participate in python-callback-style decompositions.

**Possible Drawbacks:**
Added a try-except around the python callback for the rules. This is because some python decomp rules in PL might still be broken as we are in the process of migrating towards operator2.

[sc-124571]


mlir/include/Quantum/IR/QuantumOps.td
```diff
@@ -410,7 +410,8 @@ def SetBasisStateOp : Gate_Op<"set_basis_state"> {
 }
 
 def CustomOp : UnitaryGate_Op<"custom", [DifferentiableGate, NoMemoryEffect,
-                                         AttrSizedOperandSegments, AttrSizedResultSegments]> {
+                                         AttrSizedOperandSegments, AttrSizedResultSegments,
```
paul0403, line 412: Just realized I forgot to do unitary... doing it now

paul0403: https://github.com/PennyLaneAI/catalyst/pull/3022/commits/d8462be4a98a696d826bc8a7d44f9ef7dffc17f2

frontend/test/lit/GraphDecomposition/TestFailedDecomp.mlir
```diff
@@ -12,13 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// RUN: not --crash quantum-opt --split-input-file --pass-pipeline='builtin.module( graph-decomposition{gate-set=PauliX=1.0 bytecode-rules="%BYTECODE_PATH"})' %s 2>&1 | FileCheck %s
+// RUN: not --crash catalyst --tool=opt --split-input-file --pass-pipeline='builtin.module( graph-decomposition{gate-set=PauliX=1.0 bytecode-rules="%BYTECODE_PATH"})' %s 2>&1 | FileCheck %s
```
mehrdad2m, line 15: is there a specific reason we want to use catalyst cli here?  Note that as far as I know, the cli does not support `split-input-file` option. Here it is fine since we only have one test, but `quantum-opt` should still be the preferable choice for lit tests until we support `split-input-file`

paul0403: @kipawaa can explain better than me, but I think the python callback decomp rules do not work with quantum-opt, but only the cli. I forgot the specific reason why

kipawaa: Yep, the python-decompositions rely on infrastructure that's integrated with the driver, so `quantum-opt` bypasses it. This means that for anything using callbacks (which is now everything implementing the interface) we need to use `catalyst`.

mlir/include/Quantum/IR/QuantumOps.td
```diff
@@ -857,8 +861,8 @@ def PCPhaseOp : UnitaryGate_Op<"pcphase", [DifferentiableGate, NoMemoryEffect,
     }];
 
     let extraClassDeclaration = extraBaseClassDeclaration # [{
-        mlir::ValueRange getAllParams() {
-            return getODSOperands(getParamOperandIdx());
+        mlir::OperandRange getAllParams() {
```
kipawaa: Do we actually want `dim` to be part of the dynamic data? To my understanding, it's required to be static for decomposition rules.

mlir/lib/Quantum/IR/QuantumOps.cpp
```diff
@@ -1187,6 +1213,45 @@ mlir::DictionaryAttr PauliRotOp::getStaticData()
...
+
+std::string PCPhaseOp::getOperatorName() { return "PCPhase"; }
+
+mlir::TypeRange PCPhaseOp::getDynamicShape() { return getAllParams().getTypes(); }
+
+std::vector<size_t> PCPhaseOp::getWireLens() { return {getNonCtrlQubitOperands().size()}; }
+
+mlir::DictionaryAttr PCPhaseOp::getStaticData()
+{
+    return mlir::DictionaryAttr::get(getContext(), {});
```
kipawaa: `dim` should be static data for `PCPhase`, since the decomp rule must be instantiated by this data.

mlir/unittests/Quantum/Interfaces/TestDecomposableGateInterface.cpp
```diff
@@ -82,6 +159,117 @@ module {
...
+    llvm::SmallVector<mlir::Type, 1> backing({tensorType});
+    mlir::TypeRange expectedDynamicShape(backing);
+    ASSERT_EQ(llvm::SmallVector<mlir::Type>(unitary.getDynamicShape()),
+              llvm::SmallVector<mlir::Type>(expectedDynamicShape));
+
+    ASSERT_EQ(unitary.getWireLens(), std::vector<size_t>({2}));
+
+    ASSERT_EQ(unitary.getStaticData().size(), 0);
+
+    ASSERT_EQ(unitary.getGraphOpId(), "QubitUnitary[tensor<4x4xcomplex<f64>>][2]{}");
```
kipawaa: This should be akin to a python representation of the shape & dtype of the dynamic data, i.e. `[[complex<f64>, complex<f64>, ...], [...], ...]`.

kipawaa (commented):
Thanks @paul0403 , I appreciate your help with this!

mlir/unittests/Quantum/Interfaces/TestDecomposableGateInterface.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/3022/commits/553e504fc81c3ed6d3714fff538017e34e6cac21

frontend/test/lit/GraphDecomposition/TestFailedDecomp.mlir
mehrdad2m: Sounds good thanks!

mlir/lib/Quantum/IR/QuantumOps.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/3022/commits/d946c1157239b92c365db779fef43b4dc6051d2f

mlir/include/Quantum/IR/QuantumOps.td
paul0403: https://github.com/PennyLaneAI/catalyst/pull/3022/commits/d946c1157239b92c365db779fef43b4dc6051d2f

mlir/include/Quantum/IR/QuantumInterfaces.td
```diff
@@ -195,7 +195,7 @@ def DecomposableGate : OpInterface<"DecomposableGate", [QuantumGate]> {
             "mlir::TypeRange", "getDynamicShape"
         >,
         InterfaceMethod<
-            "Return the number of wires the operation acts on for each wire argument.",
+            "Return the number of wires the operation acts on for each wire argument (excluding control wires).",
```
kipawaa, line 198: Thanks!

kipawaa (approved):
Wonderful, thanks @paul0403!

frontend/test/lit/GraphDecomposition/TestFailedDecomp.mlir
```diff
@@ -12,12 +12,13 @@
...
 
 func.func @circuit(%q0: !quantum.bit) {
-    %pi = arith.constant 3.14 : f64
-    %out = quantum.pcphase (%pi, dim : 3) %q0 : !quantum.bit
```
maliasadi, line 19: What's the reason behind replacing this with `H`?

frontend/catalyst/device/python_decompositions.py
```diff
@@ -60,16 +60,23 @@ def decomp_rule(*params, wires):
...
+        def circuit():
+            for subroutine in subroutines:
+                # TODO: I know this is dynamic, but we should probably have a better way of handling
+                # this than hard-coded dummy values. Revisit this when unifying the decomp-rule
+                # lowering pipeline
+                subroutine(*[0.5 for _ in dynamic_shape], wires=wires)
+
+        return str(circuit.mlir_module)
+    except:
+        return "builtin.module{}"
```
maliasadi, line 90: We probably can add a user-warning as when it fails the graph will treat the op as it doesn't have any valid rules which might not be the case. Addin a user warning helps to surface callback failures from python

frontend/test/lit/GraphDecomposition/TestFailedDecomp.mlir
paul0403: pcphase decomp no longer fails now that the python callback decomp is available 

Since the python rule lowering is now implemented on the `DecomposableGate` interface, adding pcphase to this interface would mean the decomp no longer fails. This is why I changed this test to a decomp that would actually fail mathematically.

paul0403: Actually, just realized this test has a gateset of just `X`, so it wasn't a mathematically possible decomposition anyway. Reverted!

frontend/catalyst/device/python_decompositions.py
paul0403: https://github.com/PennyLaneAI/catalyst/pull/3022/commits/7824585b34434404de331cc4c3a17e32001b5119

frontend/catalyst/device/python_decompositions.py
```diff
@@ -60,16 +62,24 @@ def decomp_rule(*params, wires):
...
+        def circuit():
+            for subroutine in subroutines:
+                # TODO: I know this is dynamic, but we should probably have a better way of handling
+                # this than hard-coded dummy values. Revisit this when unifying the decomp-rule
+                # lowering pipeline
+                subroutine(*[0.5 for _ in dynamic_shape], wires=wires)
+
+        return str(circuit.mlir_module)
+    except:
+        warnings.warn(f"Failed to compile python decomposition rules for {op_name}", UserWarning)
```
maliasadi:
```suggestion
        warnings.warn(
            f"Python decomposition rule compilation failed for operator "
            f"'{op_name}' (id: {op_id}); it will be treated as non-decomposable "
            f"by the graph solver.", UserWarning,
        )
```

maliasadi (approved):

PR #3024, opened 2026-07-14
Add bufferization to launch_kernel

mehrdad2m:
**Context:**
Currently `catalyst.launch_kernel` has no BufferizableOpInterface implementation, so it can not pass through one-shot bufferization. This is needed for the remote-execution path, where a qnode is launched via launch_kernel and its tensor operands/results must be lowered to memrefs.

Adds a LaunchKernelOpInterface external model and registers it.

**Benefits:**
The `catalyst.launch_kernel` operation is now bufferizable.

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-124786]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -12,6 +12,9 @@
 
 <h3>Improvements 🛠</h3>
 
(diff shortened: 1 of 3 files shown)


paul0403 (approved):

rniczh (approved):
Thanks! @mehrdad2m

PR #3025, opened 2026-07-14
Add Resource Interfaces for extensible resource counting

sengthai:
**Context:**
The current `ResourceAnalysis` in Catalyst only counts the instructions on specific dialects. Any
new dialect needs to be explicitly added to `ResourceAnalysis` to be supported.
This becomes an issue as we add more program forms (PBC, vISA, and so on).

This PR aims to make the ResourceAnalysis counting more dynamic that count the gates based on its interface. 
Doing so, we removes the hard-coded dialect/ops list from the base analysis so new dialects can opt into those counts without modifying `ResourceAnalysis` itself.

Also, this also handles control gates properly that fixes the problem from the previous one. [sc-124928]

- Introduces dialect-agnostic MLIR OpInterfaces for resource counting:
  - `ResourceQuantumOpInterface` → `operations`
  - `ResourceMeasurementOpInterface` → `measurements`
  - `ResourceAllocQubitOpInterface` → `num_alloc_qubits`
  
  
  [sc-124301]
  [sc-124302]


mlir/include/Catalyst/Analysis/ResourceInterfaces.td
```diff
@@ -0,0 +1,95 @@
...
+  }];
+
+  let cppNamespace = "::catalyst";
+  let methods = [
+    InterfaceMethod<
+        "Return the resource display name for this measurement (including observable info when applicable).",
+        "llvm::StringRef", "getResourceMeasurementName", (ins), /*methodBody*/[{}]
+    >,
+  ];
+}
```
jzaia18, line 93: I think we should change this to be explicitly measurement processes, and treat `qp.measure` as an operation (as per the discussion in the recent ADR)

mlir/include/Catalyst/Analysis/ResourceInterfaces.td
```diff
@@ -0,0 +1,95 @@
...
+  let cppNamespace = "::catalyst";
+  let methods = [
+    InterfaceMethod<
+        "Return how many qubits this op allocates.",
+        "uint64_t", "getResourceNumAllocQubits", (ins), /*methodBody*/[{}]
+    >,
+  ];
+}
+
+def ResourceMeasurementOpInterface : OpInterface<"ResourceMeasurementOpInterface"> {
```
jzaia18, line 78: No strong opinions on this, but would it be better to separate out the measurement type and measurement args as 2 separate methods? e.g. `expval` is the measurement type, `PauliZ` is the arg, or `probs` and `3 wires`. Just a thought, feel free to resolve this with no changes if returning a single string is easier and doesn't lose us anything.

jzaia18 (commented):
I don't have too much knowledge of MLIR internals so I'll let the others comment more on the actual implementation. Some high-level questions:

- With these new interfaces, what happens to quantum ops that are not declared to implement these interfaces? Would they just be reported as `classical_instructions`?
- Is there significant maintenance burden to adding new fields that need to be tracked for a resource now, since each op would have to define its own implementation? If so, is there anything we can do to lighten this burden?
- Is it possible to have some sort of "inheritance" for these interfaces? For example in the future, could you easily define something like `QECResourceQuantumOpInterface` which stores all the same things as `ResourceQuantumOpInterface` but with added QEC fields? (or would you just have to create an entirely separate interface that has similar methods)?

mlir/include/PBC/IR/PBCOps.td
jzaia18, line 1: Jukebox helped me find this one: Don't forget `PBCRefOps.td` for the reference semantics version of PPM. Might be worth double checking that all the other ref ops are handled as well

mlir/lib/PBC/IR/PBCOps.cpp
```diff
@@ -238,3 +241,59 @@ void LayerOp::print(OpAsmPrinter &p)
...
+    case 0:
+        return "PPR-identity";
+    case 2:
+        return "PPR-pi/2";
+    case 4:
+        return "PPR-pi/4";
+    case 8:
+        return "PPR-pi/8";
+    }
+    assert(false && "PPRotationOp::getResourceName: invalid rotation kind");
```
jzaia18, line 261: Will this assert exist even when building for release? We might need another failure mode here (unsure if this is typical for our code, if so ignore me)

dime10:
Great idea, this is exactly what the interfaces are for 💯

kipawaa (commented):
I can't really speak to the specs side of things, but from a technical perspective this looks great! Nice work @sengthai!

mlir/include/Catalyst/Analysis/ResourceInterfaces.td
sengthai: Good point!

sengthai: do we want to change treat qp.measure as an operation now or you think we should wait untill we implement with new json output?

mlir/lib/PBC/IR/PBCOps.cpp
sengthai: Yes, this assert also works when releases.

sengthai:
@jzaia18 Thanks for these questions! 

> With these new interfaces, what happens to quantum ops that are not declared to implement these interfaces? Would they just be reported as classical_instructions?
 
Yes, those ops that are unknown will be reported as `classical_instructions`.  The caveat is that It can be a. false report if that operations don't implement the interface and not classical instructions. This happens if the catalyst developer miss to implement those interface.

> Is there significant maintenance burden to adding new fields that need to be tracked for a resource now, since each op would have to define its own implementation? If so, is there anything we can do to lighten this burden?

Yes, it could be maintenance burden depend on the implementation of ops themself. As shown in this PR, it the dialect has well defined interface (e.g `QuantumOperation` ), we can easy at once implementation in that interface rather in each op.

> Is it possible to have some sort of "inheritance" for these interfaces? For example in the future, could you easily define something like QECResourceQuantumOpInterface which stores all the same things as ResourceQuantumOpInterface but with added QEC fields? (or would you just have to create an entirely separate interface that has similar methods)?

That's good question. it is possible to have something like `QECResourceQuantumOpInterface` -- this can be counted by this PR as well. Or another way around, I plan to create `ResourceExtension` which freely obtains the analysis output and then are mapped into `extended_fields` in json.

mlir/include/Catalyst/Analysis/ResourceInterfaces.td
jzaia18: I think we should start treating it as one now, no harm in doing this

jzaia18 (approved):
Thanks, Sengthai! Looks good to me at a high level, in terms of what we'd expect in `specs`. I'll defer to others for the technical implementation part of the review

mlir/include/PBC/IR/PBCOps.td
sengthai: Good spot! I fixed in [8af8d38](https://github.com/PennyLaneAI/catalyst/pull/3025/commits/8af8d38a59a70a1ee07769a2d88bab71a8597697)

mlir/include/Catalyst/Analysis/ResourceInterfaces.td
sengthai: right! i fixed in [ca59b05](https://github.com/PennyLaneAI/catalyst/pull/3025/commits/ca59b051bbe1b4a70e1074693e2e103c851186f3)

mlir/include/Catalyst/Analysis/ResourceInterfaces.td
sengthai: Agree it could be cleaner structurally, but since we only key a flat string map matching the existing specs format (expval(PauliZ), etc.), keeping one method avoids extra interface complexity without losing info. If we later need aggregation by measurement kind alone, we can split then.

frontend/test/pytest/test_specs.py
```diff
@@ -1064,7 +1064,7 @@ def circ():
             level="decompose-arbitrary-ppr",
             resources=SpecsResources(
                 gate_types={
-                    "pbc.prepare": 1,
```
maliasadi, line 1030: how's `pbc.prepare` now reported in the specs results?

maliasadi (commented):
Nice work!

frontend/test/pytest/test_specs.py
sengthai: the report in specs is the same. We just renamed it to "PrepareState" instead of "pbc.prepare"

maliasadi: has this already discussed with product?

sengthai: I'm discussing with product team and still learn about it. I think we can prefer the previous name report first (e.g pbc.prepare) then we can update them later in next PR since this PR is not for renaming those op.

sengthai: [503cbfa](https://github.com/PennyLaneAI/catalyst/pull/3025/commits/503cbfa06c856662360756cc7c8621621ce9d80e)

mlir/test/Catalyst/ResourceAnalysisTest.mlir
```diff
@@ -1154,3 +1156,134 @@ module @nested_resource_module {
...
+
+    %o4 = quantum.operator "MyOpD"(%theta : f64) adj qubits(%o3)
+      static_data = {pauli_string = "XYZ"}
+
+    %r0 = quantum.insert %0[ 0], %o4 : !quantum.reg, !quantum.bit
+    %r1 = quantum.insert %r0[ 1], %oc1 : !quantum.reg, !quantum.bit
+    %r2 = quantum.insert %r1[ 2], %oc2 : !quantum.reg, !quantum.bit
+    quantum.dealloc %r2 : !quantum.reg
+    return
+}
```
rniczh:
```suggestion
}

```

rniczh (approved):
💯

PR #3026, opened 2026-07-15
Fix:Python headers defined after system headers cause aliasing of _POSIX_C_SOURCE macro definition

### LINKED ISSUE #2962: Python headers defined after system headers cause aliasing of `_POSIX_C_SOURCE` macro definition
When building with recent version of gcc/clang (15/22), the position of the Python headers in the source files can interfere with previously defined macros, such as `_POSIX_C_SOURCE`. The Python [official documentation](https://docs.python.org/3/c-api/intro.html#include-files) states that for extensions Python headers should come before any system headers, with the following guideline (copied verbatim)

```
Note

Since Python may define some pre-processor definitions which affect the standard headers on some systems, you must include Python.h before any standard headers are included.

It is recommended to always define PY_SSIZE_T_CLEAN before including Python.h. See [Parsing arguments and building values](https://docs.python.org/3/c-api/arg.html#arg-parsing) for a description of this macro.
```

Through the codebase, we treat external package headers (like Python) as an expected follow behind system headers, which raises warnings such as
```
/usr/include/python3.14/pyconfig.h:2007:9: error: '_POSIX_C_SOURCE' macro redefined [-Werror,-Wmacro-redefined]
 2007 | #define _POSIX_C_SOURCE 200809L
      |         ^
/usr/include/features.h:319:10: note: previous definition is here
  319 | # define _POSIX_C_SOURCE        202405L
      |          ^
```

on newer systems. With warnings-as-errors enabled by default, this yields a build failure for any modules making use of such headers for extensions (such as anything nanobind related).

 A quick swap of header positions shows this warning as resolved, but will be reintroduced following the clang-format rules applied to headers in the code-base. This will require an update to the rules as an exception for Python headers (both CPython and nanobind) to appear before any system headers on the given platform.

JJRBEST:
**Context:**
Python headers defined after system headers cause aliasing of _POSIX_C_SOURCE macro definition
Reordered includes: moved Python and nanobind headers to the top of the include list.
**Benefits:**
building with recent version of gcc/clang (15/22)
**Possible Drawbacks:**

**Related GitHub Issues:**
Closes #2962

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -15,7 +15,7 @@ IncludeBlocks: Regroup
 IncludeCategories:
   # Python & nanobind & pybind11
   - Regex: '^(")(pybind11|Python|nanobind)(\/.*)?\.([hp]+)(")$'
-    Priority: 0  
+    Priority: -1  
 
   # System/STL
   - Regex:    '^(<)(([^.]+)|((std.*|complex|sys.*|dlfcn)\.h))(>)$'
--- a/.clang-format
+++ b/.clang-format
@@ -13,12 +13,16 @@ IncludeBlocks: Regroup
 # NOTE: Ensure that lists are always sorted alphabetically.
 
 IncludeCategories:
+  # Python & nanobind & pybind11
+  - Regex: '^(")(Python|nanobind)(\/.*)?\.([hp]+)(")$'
+    Priority: 0  
+
   # System/STL
   - Regex:    '^(<)(([^.]+)|((std.*|complex|sys.*|dlfcn)\.h))(>)$'
     Priority: 1
 
   # Dependencies
-  - Regex:    '^(")(llvm|Enzyme|absl|boost|catch2|gmock|gtest|jax_cpu_lapack_kernels|mlir|nanobind|nlohmann|numpy|pybind11|Python|stablehlo|stim|toml\+\+|xla)(\/.*)?\.([hcp]+)(")$'
+  - Regex:    '^(")(llvm|Enzyme|absl|boost|catch2|gmock|gtest|jax_cpu_lapack_kernels|mlir|nlohmann|numpy|pybind11|stablehlo|stim|toml\+\+|xla)(\/.*)?\.([hcp]+)(")$'
     Priority: 2
 
   # Local headers
--- a/.clang-format
+++ b/.clang-format
@@ -14,7 +14,7 @@ IncludeBlocks: Regroup
 
 IncludeCategories:
   # Python & nanobind & pybind11
-  - Regex: '^(")(Python|nanobind)(\/.*)?\.([hp]+)(")$'
+  - Regex: '^(")(pybind11|Python|nanobind)(\/.*)?\.([hp]+)(")$'
     Priority: 0  
 
   # System/STL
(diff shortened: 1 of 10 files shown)


JJRBEST:
Step 1: git status
image
Step 2: make format
image
Step 3: git status
image
Step 4: git diff frontend/catalyst/utils/wrapper.cpp
image

Concludes:
#include "nanobind/nanobind.h" is already sorted as the first line above all the other #include statements.

.clang-format
```diff
@@ -13,12 +13,16 @@ IncludeBlocks: Regroup
...
+  - Regex: '^(")(Python|nanobind)(\/.*)?\.([hp]+)(")$'
+    Priority: 0  
+
   # System/STL
   - Regex:    '^(<)(([^.]+)|((std.*|complex|sys.*|dlfcn)\.h))(>)$'
     Priority: 1
 
   # Dependencies
-  - Regex:    '^(")(llvm|Enzyme|absl|boost|catch2|gmock|gtest|jax_cpu_lapack_kernels|mlir|nanobind|nlohmann|numpy|pybind11|Python|stablehlo|stim|toml\+\+|xla)(\/.*)?\.([hcp]+)(")$'
+  - Regex:    '^(")(llvm|Enzyme|absl|boost|catch2|gmock|gtest|jax_cpu_lapack_kernels|mlir|nlohmann|numpy|pybind11|stablehlo|stim|toml\+\+|xla)(\/.*)?\.([hcp]+)(")$'
```
kipawaa:
```suggestion
  - Regex:    '^(")(llvm|Enzyme|absl|boost|catch2|gmock|gtest|jax_cpu_lapack_kernels|mlir|nlohmann|numpy|stablehlo|stim|toml\+\+|xla)(\/.*)?\.([hcp]+)(")$'
```

.clang-format
```diff
@@ -13,12 +13,16 @@ IncludeBlocks: Regroup
 # NOTE: Ensure that lists are always sorted alphabetically.
 
 IncludeCategories:
+  # Python & nanobind & pybind11
+  - Regex: '^(")(Python|nanobind)(\/.*)?\.([hp]+)(")$'
```
kipawaa:
```suggestion
  - Regex: '^(")(pybind11|Python|nanobind)(\/.*)?\.([hp]+)(")$'
```

kipawaa (commented):
Thank you @JJRBEST! 🦸 I think we should also include `pybind11` headers with these changes, I've suggested the appropriate changes below. I appreciate you making the changes in the `.clang-format` to simplify review, though we'll need to format the codebase using the new format rules in order to pass CI. Could you commit the formatting changes as well please?

mlir/lib/Driver/DefaultPipelines/DefaultPipelines.cpp
```diff
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 #include "Driver/DefaultPipelines/DefaultPipelines.h"
-
 #include "nanobind/nanobind.h"
```
kipawaa: It seems that priority 0 is assigned to the current file header ([clang-format docs](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#includecategories)). I think that since these _always_ need to be first (and should be sorted into a separate category from the main header), we should assign negative priority.

.clang-format
```diff
@@ -13,12 +13,16 @@ IncludeBlocks: Regroup
 # NOTE: Ensure that lists are always sorted alphabetically.
 
 IncludeCategories:
+  # Python & nanobind & pybind11
+  - Regex: '^(")(pybind11|Python|nanobind)(\/.*)?\.([hp]+)(")$'
+    Priority: 0
```
kipawaa:
```suggestion
    Priority: -1  
```

kipawaa: Per my comment above, priority 0 is for the main (current) header.

kipawaa (commented):
Thanks for the quick update @JJRBEST! 🚀 it seems we missed a small detail in the header-sorting, could you update and push formatting changes again to confirm the fix?

kipawaa (approved):
Nice work @JJRBEST! 🚀

mehrdad2m (approved):
Thank you @JJRBEST LGTM

PR #3028, opened 2026-07-15
Improve libpython path detection

kipawaa:
**Context:**
The libpython path detection assumes a default suffix and asserts it as part of the search.

Uses a platform-agnostic suffix and assertion that the found library is in fact static.

**Benefits:**
More reliable libpython path resolution across varied environments.

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/frontend/catalyst/utils/runtime_environment.py
+++ b/frontend/catalyst/utils/runtime_environment.py
@@ -73,7 +73,6 @@ def get_libpython_path() -> str:  # pragma: no cover
     """
     ldlibrary = sysconfig.get_config_var("LDLIBRARY")
     framework_prefix = sysconfig.get_config_var("PYTHONFRAMEWORKPREFIX")
-    shlib_suffix = sysconfig.get_config_var("SHLIB_SUFFIX") or ".so"
 
     # macOS framework-style installations
     if framework_prefix and ldlibrary:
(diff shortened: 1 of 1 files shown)


maliasadi (approved):

frontend/catalyst/utils/runtime_environment.py
dime10: I think the `or ".so"` doesn't really make much sense to me?

mehrdad2m (approved):

frontend/catalyst/utils/runtime_environment.py
```diff
@@ -85,7 +86,7 @@ def get_libpython_path() -> str:   # pragma: no cover
         return ""
 
     # Default: the library sysconfig reports, but only if it is a shared object.
-    if ldlibrary and shlib_suffix in ldlibrary:
+    if ldlibrary and not ldlibrary.endswith(".a"):
```
dime10, line 88: 👍

frontend/catalyst/utils/runtime_environment.py
dime10: Can delete this line now after the last commit I think

kipawaa: [4bcbb32](https://github.com/PennyLaneAI/catalyst/pull/3028/commits/4bcbb3266ef5fd3699b805d5ec325ec44bede731)

maliasadi (approved):

PR #3029, opened 2026-07-16
Support magic-T allocation via qp.allocate (MLIR + capture path)

Qottmann:
## Summary

Adds **capture-path** (`qjit(capture=True)`) support for `qp.allocate(..., state="magic-T"|"magic-T-adj")`, pairing with the merged [PennyLaneAI/pennylane#9846](https://github.com/PennyLaneAI/pennylane/pull/9846). Magic states lower through qref/PBC to `pbc.fabricate` (H+T / H+T†).

This is **PR 1 of 2**. [#3027](https://github.com/PennyLaneAI/catalyst/pull/3027) adds all **legacy (`capture=False`)** tracing on top of this PR.

**User-facing state strings:** `"magic-T"` and `"magic-T-adj"`. MLIR/PBC enums remain `magic` / `magic_conj`; Catalyst maps at lowering time.

### MLIR / PBC lowering
- Add `pbc.ref.fabricate` for magic and magic-conjugate preparation
- Lower `pbc.ref.fabricate` → `pbc.fabricate` in value-semantics conversion, with dead-anchor cleanup
- Resource analysis + MLIR lit tests

### Unified allocate/deallocate jaxpr
- Capture frontend emits PennyLane `allocate_prim` / `deallocate_prim`
- MLIR lowering in `qref_jax_primitives.py` selects `qref.alloc` vs `pbc.ref.fabricate` from `state`
- Map `"magic-T"` / `"magic-T-adj"` → MLIR `magic` / `magic_conj`

### Capture path only
- `qfunc_interpreter.py` handles allocate/deallocate via unified jaxpr primitives
- Device verification accepts new states

### CI / deps
- PennyLane `0.46.0.dev62` (`.dep-versions`, `doc/requirements.txt`)

### Tests & docs (capture-only)
- MLIR lit + Python lit (all `capture=True`)
- Pytest: magic-T allocation, manual deallocate, MLIR lowering (`capture=True` only)
- `sharp_bits.rst` + changelog

## Review guide

| Area | Files |
|------|-------|
| MLIR | `mlir/include/PBC/IR/PBCRefOps.td`, `mlir/lib/QRef/Transforms/value_semantics_conversion.*`, `mlir/lib/PBC/IR/PBCOps.cpp` |
| JAX → MLIR | `frontend/catalyst/from_plxpr/qref_jax_primitives.py` |
| Capture | `frontend/catalyst/from_plxpr/qfunc_interpreter.py`, `frontend/catalyst/device/verification.py` |

## Test plan
- [x] MLIR lit tests
- [x] Capture lit + pytest magic-T tests
- [x] `make format check=1` (clang-format, black, isort)
- [ ] CI green before merging #3027

## Related
- Follow-up: [#3027](https://github.com/PennyLaneAI/catalyst/pull/3027) — **all** legacy `jax_tracer` work
- PennyLane: [PennyLaneAI/pennylane#9846](https://github.com/PennyLaneAI/pennylane/pull/9846) (merged)
- Shortcut: [sc-124605](https://app.shortcut.com/xanadu/story/124605)

### DIFF
--- a/doc/dev/sharp_bits.rst
+++ b/doc/dev/sharp_bits.rst
@@ -1387,7 +1387,7 @@ Currently, however, this is not the case for the following functionalities.
     Capture and legacy pathways share the same ``allocate`` / ``deallocate`` jaxpr
     primitives; the choice between register allocation and fabrication is made
     during MLIR lowering based on the ``state`` argument. Legacy tracing support
-    is added in [#3027](https://github.com/PennyLaneAI/catalyst/pull/3027).
+    is added in `#3027 <https://github.com/PennyLaneAI/catalyst/pull/3027>`_.
 
   - Related to the above point, in PennyLane, dynamic wire allocations do not
     increase the total number of wires used in the circuit. This is because
--- a/doc/dev/sharp_bits.rst
+++ b/doc/dev/sharp_bits.rst
@@ -1376,9 +1376,14 @@ Currently, however, this is not the case for the following functionalities.
 - **Dynamic wire allocation behaviour**: The ``qp.allocate()`` function currently
   behaves differently when Catalyst is present or not. In particular:
 
-  - The ``state`` and ``restored`` keyword arguments of ``qp.allocate()`` are
-    ignored in Catalyst. The reason is that the only supported mode is to always allocate in the
-    zero state.
+  - The ``state`` and ``restored`` keyword arguments for ``state="zero"`` and
+    ``state="any"`` are ignored in Catalyst. The reason is that the only
+    supported mode for these states is to always allocate in the zero state.
+
+  - Magic state allocation via ``state="magic"`` or ``state="magic_conj"`` is
+    supported in Catalyst and lowers to ``pbc.ref.fabricate`` (and ultimately
+    ``pbc.fabricate``) under the hood. These allocations use single-qubit
+    fabricate/deallocate semantics rather than register allocation.
 
   - Related to the above point, in PennyLane, dynamic wire allocations do not
     increase the total number of wires used in the circuit. This is because
--- a/doc/dev/sharp_bits.rst
+++ b/doc/dev/sharp_bits.rst
@@ -1380,7 +1380,7 @@ Currently, however, this is not the case for the following functionalities.
     ``state="any"`` are ignored in Catalyst. The reason is that the only
     supported mode for these states is to always allocate in the zero state.
 
-  - Magic state allocation via ``state="magic"`` or ``state="magic_conj"`` is
+  - Magic state allocation via ``state="magic-T"`` or ``state="magic-T-adj"`` is
     supported in Catalyst and lowers to ``pbc.ref.fabricate`` (and ultimately
     ``pbc.fabricate``) under the hood. These allocations use single-qubit
     fabricate/deallocate semantics rather than register allocation. Both capture
--- a/doc/dev/sharp_bits.rst
+++ b/doc/dev/sharp_bits.rst
@@ -1383,7 +1383,10 @@ Currently, however, this is not the case for the following functionalities.
   - Magic state allocation via ``state="magic"`` or ``state="magic_conj"`` is
     supported in Catalyst and lowers to ``pbc.ref.fabricate`` (and ultimately
     ``pbc.fabricate``) under the hood. These allocations use single-qubit
-    fabricate/deallocate semantics rather than register allocation.
+    fabricate/deallocate semantics rather than register allocation. Both capture
+    and legacy pathways use the same ``allocate`` / ``deallocate`` jaxpr
+    primitives; the choice between register allocation and fabrication is made
+    during MLIR lowering based on the ``state`` argument.
 
   - Related to the above point, in PennyLane, dynamic wire allocations do not
     increase the total number of wires used in the circuit. This is because
--- a/doc/dev/sharp_bits.rst
+++ b/doc/dev/sharp_bits.rst
@@ -1376,18 +1376,16 @@ Currently, however, this is not the case for the following functionalities.
 - **Dynamic wire allocation behaviour**: The ``qp.allocate()`` function currently
   behaves differently when Catalyst is present or not. In particular:
 
-  - The ``state`` and ``restored`` keyword arguments for ``state="zero"`` and
-    ``state="any"`` are ignored in Catalyst. The reason is that the only
-    supported mode for these states is to always allocate in the zero state.
+  - The ``state="any"`` configuration, as well as the ``restored`` keyword, are not
+    supported by Catalyst and will raise a compilation error. Catalyst supports
+    ``state="zero"`` initialization (the default), as well as magic state
+    initialization (see below).
 
   - Magic state allocation via ``state="magic-T"`` or ``state="magic-T-adj"`` is
-    supported in Catalyst (``qjit(capture=True)``) and lowers to ``pbc.ref.fabricate``
-    (and ultimately ``pbc.fabricate``) under the hood. These allocations use
-    single-qubit fabricate/deallocate semantics rather than register allocation.
-    Capture and legacy pathways share the same ``allocate`` / ``deallocate`` jaxpr
-    primitives; the choice between register allocation and fabrication is made
-    during MLIR lowering based on the ``state`` argument. Legacy tracing support
-    is added in `#3027 <https://github.com/PennyLaneAI/catalyst/pull/3027>`_.
+    supported in Catalyst provided program capture is used (``qjit(capture=True)``).
+    PennyLane also supports these allocation modes; the Catalyst-specific behaviour
+    is that dynamically allocated wires are tracked separately from device wires
+    (see below).
 
   - Related to the above point, in PennyLane, dynamic wire allocations do not
     increase the total number of wires used in the circuit. This is because
--- a/doc/dev/sharp_bits.rst
+++ b/doc/dev/sharp_bits.rst
@@ -1382,8 +1382,8 @@ Currently, however, this is not the case for the following functionalities.
     initialization (see below).
 
   - Magic state allocation via ``state="magic-T"`` or ``state="magic-T-adj"`` is
-    supported in Catalyst provided program capture is used (``qjit(capture=True)``).
-    PennyLane also supports these allocation modes; the Catalyst-specific behaviour
+    supported in Catalyst, provided program capture is used (``qjit(capture=True)``).
+    PennyLane currently does not support allocating magic states; the Catalyst-specific behaviour
     is that dynamically allocated wires are tracked separately from device wires
     (see below).
 
--- a/doc/dev/sharp_bits.rst
+++ b/doc/dev/sharp_bits.rst
@@ -1381,12 +1381,13 @@ Currently, however, this is not the case for the following functionalities.
     supported mode for these states is to always allocate in the zero state.
 
   - Magic state allocation via ``state="magic-T"`` or ``state="magic-T-adj"`` is
-    supported in Catalyst and lowers to ``pbc.ref.fabricate`` (and ultimately
-    ``pbc.fabricate``) under the hood. These allocations use single-qubit
-    fabricate/deallocate semantics rather than register allocation. Both capture
-    and legacy pathways use the same ``allocate`` / ``deallocate`` jaxpr
+    supported in Catalyst (``qjit(capture=True)``) and lowers to ``pbc.ref.fabricate``
+    (and ultimately ``pbc.fabricate``) under the hood. These allocations use
+    single-qubit fabricate/deallocate semantics rather than register allocation.
+    Capture and legacy pathways share the same ``allocate`` / ``deallocate`` jaxpr
     primitives; the choice between register allocation and fabrication is made
-    during MLIR lowering based on the ``state`` argument.
+    during MLIR lowering based on the ``state`` argument. Legacy tracing support
+    is added in [#3027](https://github.com/PennyLaneAI/catalyst/pull/3027).
 
   - Related to the above point, in PennyLane, dynamic wire allocations do not
     increase the total number of wires used in the circuit. This is because
--- a/doc/dev/sharp_bits.rst
+++ b/doc/dev/sharp_bits.rst
@@ -1376,8 +1376,8 @@ Currently, however, this is not the case for the following functionalities.
 - **Dynamic wire allocation behaviour**: The ``qp.allocate()`` function currently
   behaves differently when Catalyst is present or not. In particular:
 
-  - The ``state="any"`` configuration, as well as the ``restored`` keyword, are not
-    supported by Catalyst and will raise a compilation error. Catalyst supports
+  - The ``state="any"`` configuration is not supported by Catalyst and will raise a
+    compilation error. The ``restored`` keyword is ignored. Catalyst supports
     ``state="zero"`` initialization (the default), as well as magic state
     initialization (see below).
 
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -2,7 +2,7 @@
 
 <h3>New features since last release</h3>
 
-* ``qp.allocate`` now supports ``state="magic"`` and ``state="magic_conj"``. Both
+* ``qp.allocate`` now supports ``state="magic-T"`` and ``state="magic-T-adj"``. Both
   capture and legacy frontends emit the same ``allocate`` / ``deallocate`` jaxpr
   primitives; MLIR lowering selects ``qref.alloc`` or ``pbc.ref.fabricate`` based
   on the ``state`` argument.
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -35,11 +35,6 @@
   Instead, rules are removed by the `symbol-dce` pass at the end of the `QuantumCompilationStage`.
   [(#2973)](https://github.com/PennyLaneAI/catalyst/pull/2973)
 
-* Catalyst CI and specs tests are updated for PennyLane ``0.46.0.dev62``, aligning with the
-  resource hierarchy from [PennyLane #9841](https://github.com/PennyLaneAI/pennylane/pull/9841)
-  and Operator2 changes from [PennyLane #9846](https://github.com/PennyLaneAI/pennylane/pull/9846).
-  [(#3050)](https://github.com/PennyLaneAI/catalyst/pull/3050)
-
 * The new `pennylane.core.Operator2` can now be lowered to MLIR with program capture for operators
   without non-lowerable arguments. `Operator2` classes are now lowered to specialized operations
   where applicable, unlocking compilation and execution for these cases. `qp.specs` and the
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -35,6 +35,11 @@
   Instead, rules are removed by the `symbol-dce` pass at the end of the `QuantumCompilationStage`.
   [(#2973)](https://github.com/PennyLaneAI/catalyst/pull/2973)
 
+* Catalyst CI and specs tests are updated for PennyLane ``0.46.0.dev62``, aligning with the
+  resource hierarchy from [PennyLane #9841](https://github.com/PennyLaneAI/pennylane/pull/9841)
+  and Operator2 changes from [PennyLane #9846](https://github.com/PennyLaneAI/pennylane/pull/9846).
+  [(#3050)](https://github.com/PennyLaneAI/catalyst/pull/3050)
+
 * The new `pennylane.core.Operator2` can now be lowered to MLIR with program capture for operators
   without non-lowerable arguments. `Operator2` classes are now lowered to specialized operations
   where applicable, unlocking compilation and execution for these cases. `qp.specs` and the
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -2,10 +2,12 @@
 
 <h3>New features since last release</h3>
 
-* ``qp.allocate`` now supports ``state="magic"`` and ``state="magic_conj"``, lowered via
-  ``pbc.ref.fabricate`` to ``pbc.fabricate`` (H+T / H+T†) with ``qref.dealloc_qb`` deallocation.
+* ``qp.allocate`` now supports ``state="magic"`` and ``state="magic_conj"``. Both
+  capture and legacy frontends emit the same ``allocate`` / ``deallocate`` jaxpr
+  primitives; MLIR lowering selects ``qref.alloc`` or ``pbc.ref.fabricate`` based
+  on the ``state`` argument.
   Requires the corresponding PennyLane release for the new ``AllocateState`` values.
-  [(#3027)](https://github.com/PennyLaneAI/catalyst/pull/3027)
+  [(#3027)](https://github.com/PennyLaneAI/catalyst/pull/3027), [(#3029)](https://github.com/PennyLaneAI/catalyst/pull/3029)
 
 * The `local-random` unitary folding option for :func:`~.mitigate_with_zne` is now implemented,
   reproducing Mitiq's ``fold_gates_at_random``: every gate is folded ``floor((scale_factor-1)/2)``
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -2,6 +2,11 @@
 
 <h3>New features since last release</h3>
 
+* ``qp.allocate`` now supports ``state="magic"`` and ``state="magic_conj"``, lowered via
+  ``pbc.ref.fabricate`` to ``pbc.fabricate`` (H+T / H+T†) with ``qref.dealloc_qb`` deallocation.
+  Requires the corresponding PennyLane release for the new ``AllocateState`` values.
+  [(#3027)](https://github.com/PennyLaneAI/catalyst/pull/3027)
+
 * The `local-random` unitary folding option for :func:`~.mitigate_with_zne` is now implemented,
   reproducing Mitiq's ``fold_gates_at_random``: every gate is folded ``floor((scale_factor-1)/2)``
   times, then a random subset is folded once more (without replacement) to reach ``scale_factor * n``
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -2,16 +2,23 @@
 
 <h3>New features since last release</h3>
 
-* ``qp.allocate`` now supports ``state="magic-T"`` and ``state="magic-T-adj"`` with
-  ``qjit(capture=True)``. MLIR lowering selects ``qref.alloc`` or ``pbc.ref.fabricate``
-  based on the ``state`` argument via unified ``allocate`` / ``deallocate`` jaxpr
-  primitives.
-  Requires the corresponding PennyLane release for the new ``AllocateState`` values.
-  [(#3029)](https://github.com/PennyLaneAI/catalyst/pull/3029)
+* You can now dynamically prepare magic T states inside captured Catalyst workflows using
+  ``qp.allocate(state="magic-T")`` and ``qp.allocate(state="magic-T-adj")``, which makes it
+  easier to compile FTQC-style routines that need T-state ancillas on the fly (for example
+  TemporaryAND) with ``qjit(capture=True)``.
 
-* Legacy (``capture=False``) tracing for magic-T dynamic wires, including mid-circuit
-  measurements and control flow, is added in
-  [(#3027)](https://github.com/PennyLaneAI/catalyst/pull/3027).
+  ```python
+  @qjit(capture=True)
+  @qnode(dev)
+  def circuit():
+      qb = qp.allocate(state="magic-T")
+      # ... use qb in your circuit ...
+      qp.deallocate(qb)
+  ```
+
+  Requires a PennyLane release that supports the new ``AllocateState`` values
+  (see [PennyLane #9846](https://github.com/PennyLaneAI/pennylane/pull/9846)).
+  [(#3029)](https://github.com/PennyLaneAI/catalyst/pull/3029)
 
 * The `local-random` unitary folding option for :func:`~.mitigate_with_zne` is now implemented,
   reproducing Mitiq's ``fold_gates_at_random``: every gate is folded ``floor((scale_factor-1)/2)``
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -2,12 +2,16 @@
 
 <h3>New features since last release</h3>
 
-* ``qp.allocate`` now supports ``state="magic-T"`` and ``state="magic-T-adj"``. Both
-  capture and legacy frontends emit the same ``allocate`` / ``deallocate`` jaxpr
-  primitives; MLIR lowering selects ``qref.alloc`` or ``pbc.ref.fabricate`` based
-  on the ``state`` argument.
+* ``qp.allocate`` now supports ``state="magic-T"`` and ``state="magic-T-adj"`` with
+  ``qjit(capture=True)``. MLIR lowering selects ``qref.alloc`` or ``pbc.ref.fabricate``
+  based on the ``state`` argument via unified ``allocate`` / ``deallocate`` jaxpr
+  primitives.
   Requires the corresponding PennyLane release for the new ``AllocateState`` values.
-  [(#3027)](https://github.com/PennyLaneAI/catalyst/pull/3027), [(#3029)](https://github.com/PennyLaneAI/catalyst/pull/3029)
+  [(#3029)](https://github.com/PennyLaneAI/catalyst/pull/3029)
+
+* Legacy (``capture=False``) tracing for magic-T dynamic wires, including mid-circuit
+  measurements and control flow, is added in
+  [(#3027)](https://github.com/PennyLaneAI/catalyst/pull/3027).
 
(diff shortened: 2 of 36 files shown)


doc/dev/sharp_bits.rst
```diff
@@ -1376,9 +1376,18 @@ Currently, however, this is not the case for the following functionalities.
 - **Dynamic wire allocation behaviour**: The ``qp.allocate()`` function currently
   behaves differently when Catalyst is present or not. In particular:
 
-  - The ``state`` and ``restored`` keyword arguments of ``qp.allocate()`` are
-    ignored in Catalyst. The reason is that the only supported mode is to always allocate in the
-    zero state.
+  - The ``state`` and ``restored`` keyword arguments for ``state="zero"`` and
+    ``state="any"`` are ignored in Catalyst. The reason is that the only
+    supported mode for these states is to always allocate in the zero state.
```
dime10: I would reword this slightly since "zero" is supported (and the default behaviour), and there is an alternative now. So the only parameter value that's not supported is "any".
```suggestion
  - The ``state="any"`` configuration, as well as the ``restored`` keyword, are ignored by catalyst.
    The reason is that Catalyst only supports "zero" initialization (default), as well as magic
    state initialization (see below).
```

dime10: Come to think of it, it would be a good time to add an error now when an unsupported config is selected by the user, rather than just ignoring it.

doc/dev/sharp_bits.rst
```diff
@@ -1376,9 +1376,18 @@ Currently, however, this is not the case for the following functionalities.
...
+
+  - Magic state allocation via ``state="magic-T"`` or ``state="magic-T-adj"`` is
+    supported in Catalyst (``qjit(capture=True)``) and lowers to ``pbc.ref.fabricate``
+    (and ultimately ``pbc.fabricate``) under the hood. These allocations use
+    single-qubit fabricate/deallocate semantics rather than register allocation.
+    Capture and legacy pathways share the same ``allocate`` / ``deallocate`` jaxpr
+    primitives; the choice between register allocation and fabrication is made
+    during MLIR lowering based on the ``state`` argument. Legacy tracing support
+    is added in `#3027 <https://github.com/PennyLaneAI/catalyst/pull/3027>`_.
```
dime10: Since this is a user facing guide, we shouldn't include technical implementation details in it.
```suggestion
  - Magic state allocation via ``state="magic-T"`` or ``state="magic-T-adj"`` is
    supported in Catalyst provided program capture is used (``qjit(capture=True)``).
```

dime10: In fact, we should confirm that this is in fact differing behaviour from pennylane, which it might be if PennyLane doesn't support magic state allocation for example.

doc/releases/changelog-dev.md
```diff
@@ -2,6 +2,17 @@
 
 <h3>New features since last release</h3>
 
+* ``qp.allocate`` now supports ``state="magic-T"`` and ``state="magic-T-adj"`` with
+  ``qjit(capture=True)``. MLIR lowering selects ``qref.alloc`` or ``pbc.ref.fabricate``
+  based on the ``state`` argument via unified ``allocate`` / ``deallocate`` jaxpr
+  primitives.
+  Requires the corresponding PennyLane release for the new ``AllocateState`` values.
```
dime10: In a changelog (really the release notes we publish) we wouldn't say things like "this PR depends on that PR" or anything of the sort. That is better left to the PR description.
Here, we are informing any user of our package about what changed for _them_ (not under the hood, unless the change is intended for "under the hood" users).
```suggestion
```

doc/releases/changelog-dev.md
```diff
@@ -2,6 +2,17 @@
...
+  [(#3029)](https://github.com/PennyLaneAI/catalyst/pull/3029)
+
+* Legacy (``capture=False``) tracing for magic-T dynamic wires, including mid-circuit
+  measurements and control flow, is added in
+  [(#3027)](https://github.com/PennyLaneAI/catalyst/pull/3027).
```
```suggestion
```

frontend/catalyst/device/verification.py
```diff
@@ -201,6 +202,8 @@ def _op_checker(op, state):
         # we can't use isinstance(op, Controlled).
         if type(op) in (Controlled, ControlledOp) or isinstance(op, (Adjoint)):
             pass
+        elif isinstance(op, (qp.allocation.Allocate, qp.allocation.Deallocate)):
+            pass
```
dime10: There is a `dynamic_qubit_management` field in the device toml that we should probably check during device verification, I think that's exactly what is was designed for.

frontend/catalyst/from_plxpr/qfunc_interpreter.py
```diff
@@ -455,25 +454,26 @@ def handle_allocate(self, *, num_wires, state=None, restored=False):
     ), "number of dynamically allocated qubits must be statically known"
 
     self.has_dynamic_allocation = True
-    new_qreg = qref_alloc_p.bind(static_num_qubits=num_wires)
-    return [qref_get_p.bind(new_qreg, i) for i in range(num_wires)]
+
+    state = (
```
dime10, line 463: error checking here (mentioned in the changelog comment) probably makes the most sense

frontend/catalyst/from_plxpr/qfunc_interpreter.py
```diff
@@ -455,25 +454,26 @@ def handle_allocate(self, *, num_wires, state=None, restored=False):
...
-        assert (
-            get_op.primitive is qref_get_p
-        ), "Manual deallocation is only supported for manually allocated wires"
-        qreg = get_op.in_tracers[0]
-        qregs.add(qreg)
-    assert (
-        len(qregs) == 1
-    ), "Expected all wires to deallocate to come from the same allocation instruction"
-    qref_dealloc_p.bind(list(qregs)[0])
+    qp.allocation.deallocate_prim.bind(*wires)
```
dime10, line 490: Actually, if these functions just rebind the primitive, do we even need them? @paul0403 what do you think?

frontend/catalyst/from_plxpr/qref_jax_primitives.py
```diff
@@ -147,6 +150,11 @@ class MeasurementPlane(Enum):
 qref_alloc_p = Primitive("qref_alloc")
 qref_dealloc_p = Primitive("qref_dealloc")
 qref_dealloc_p.multiple_results = True
+qref_fabricate_p = Primitive("qref_fabricate")
```
dime10: We can remove the fabricate primitive now, right?

dime10: I'm a little confused by the setup of these four primitives actually:
- qref_alloc_p
  - used to be called during qnode setup, and any dynamic allocs
  - now only qnode setup
- qref_dealloc_p
  - same as qref_alloc_p
- qref_fabricate_p
  - not called at all
- qref_dealloc_qb_p
  - not called at all

Additionally we have:
- allocate_prim
- deallocate_prim

which have direct lowering rules.

So there are two issues here:
1. the qref_fabricate_p/qref_dealloc_qb_p serving no purpose
2. inconsistent use of either qref_alloc_p or allocate_prim to handle allocations

@paul0403 Do you know if can get away with using PL allocation primitives directly, lowering them straight to MLIR, or is there a good reason to keep using the qref alloc primitives?

frontend/catalyst/from_plxpr/qref_jax_primitives.py
```diff
@@ -215,6 +223,112 @@ def _qref_dealloc_lowering(jax_ctx: mlir.LoweringRuleContext, qreg):
...
+def _deallocate_prim_lowering(jax_ctx: mlir.LoweringRuleContext, *qubits):
+    del jax_ctx
+    fabricated_qubits = []
+    qregs = set()
+    for qubit in qubits:
+        defining_op = qubit.owner
+        if defining_op.name == "pbc.ref.fabricate":
+            fabricated_qubits.append(qubit)
+        elif defining_op.name == "qref.get":
+            qregs.add(defining_op.operands[0])
```
dime10, line 288: Looking at the owner here is a bit unorthodox 🤔 But there might not be a way around it since PL doesn't distinguish between register and qubit allocations..

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -532,7 +532,7 @@ void ResourceAnalysis::collectOperation(Operation *op, ResourceResult &result, b
     }
 
     // Metadata: qubit allocation
-    if (isa<quantum::AllocQubitOp, qref::AllocQubitOp>(op)) {
+    if (isa<quantum::AllocQubitOp, qref::AllocQubitOp, pbc::RefFabricateOp>(op)) {
```
dime10: We're adding the ref version here but not the value version of fabricate 🤔 @sengthai Should we be counting the fabricate ops here? 

If so, we probably want to include both of them.

mlir/lib/PBC/IR/PBCOps.cpp
```diff
@@ -91,6 +91,15 @@ LogicalResult RefPPMeasurementOp::verify()
...
 }
 
+LogicalResult RefFabricateOp::verify()
+{
+    auto initState = getInitState();
+    if (initState != LogicalInitKind::magic && initState != LogicalInitKind::magic_conj) {
+        return emitOpError("Only magic states (magic, magic_conj) are supported by ref.fabricate.");
+    }
+    return success();
+}
```
dime10, line 102: We should probably match the verify method of the existing fabricate op (see below)

dime10 (changes requested):
Nice work! I think we just need to sort out the jax primitive situation

frontend/catalyst/from_plxpr/qfunc_interpreter.py
dime10: Chatted with Paul, you can leave them in for now, but please add a TODO comment for us to remove once this is merged :)

frontend/catalyst/device/verification.py
dime10: You can also leave a TODO comment here

frontend/catalyst/from_plxpr/qref_jax_primitives.py
dime10: Alright so we went over this and now that you added MLIR lowering rules for the PL primitives, you can just delete all the qref primitives and associated functions, keeping just the PL primitive lowerings :)

frontend/catalyst/from_plxpr/qref_jax_primitives.py
dime10: I'm going to resolve this since there is no clean solution right now until the PL frontend supports register values.

doc/dev/sharp_bits.rst
Qottmann: so in case it is supported in pennylane (which it is now) we dont add a changelog to catalyst at all?

frontend/catalyst/device/verification.py
Qottmann: yeah I'd prefer the TODO because leaving this to the device TOML would require changes to pennylane-lightning for CI to pass

frontend/catalyst/from_plxpr/qfunc_interpreter.py
Qottmann: This required me to update some tests that currently use "any" (why where they in there in the first place if that is not supported?)

see commit https://github.com/PennyLaneAI/catalyst/pull/3029/changes/9e179cb5b0f159d0b2042ab25ef61ca1eb162980

frontend/test/pytest/from_plxpr/test_decompose_transform.py
```diff
@@ -436,15 +436,15 @@ def my_decomp(angle, wires, **_):
             def true_func():
                 qp.CNOT(wires)
 
-                with qp.allocate(2, state="any", restored=True) as w:
+                with qp.allocate(2, state="zero") as w:
```
Qottmann, line 439: we now explicitly check the input state, and `any` is not allowed, Not sure why this was used in tests in the first place 🤔

frontend/catalyst/from_plxpr/qfunc_interpreter.py
Qottmann: ✅

doc/dev/sharp_bits.rst
Qottmann: ah wait it's not actually supported in pennylane, the literals only exist to allow to allocate in catalyst

frontend/catalyst/from_plxpr/qref_jax_primitives.py
Qottmann: ✅

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
Qottmann: added `pbc::FabricateOp` ✅

mlir/lib/PBC/IR/PBCOps.cpp
Qottmann: ✅

doc/dev/sharp_bits.rst
dime10: >so in case it is supported in pennylane (which it is now) we dont add a changelog to catalyst at all?

This is a sharp bits entry which documents differences between catalyst and pennylane. You can see the section starts with:
> Functionality differences from PennyLane
> ----------------------------------------
>
> The ultimate aim with Catalyst will be the ability to prototype quantum algorithms
> in Python with PennyLane, and easily scale up prototypes by simply adding ``@qjit``.
> This will require that all PennyLane functionality behaves identically whether or not
> the ``@qjit`` decorator is applied.
>
> Currently, however, this is not the case for the following functionalities.
> ...
>
> - **Dynamic wire allocation behaviour**: The ``qp.allocate()`` function currently
>   behaves differently when Catalyst is present or not. In particular:

frontend/catalyst/from_plxpr/qfunc_interpreter.py
dime10: Yeah weird, thanks for updating them!

doc/releases/changelog-dev.md
```diff
@@ -2,6 +2,24 @@
 
 <h3>New features since last release</h3>
 
+* You can now dynamically prepare magic T states inside captured Catalyst workflows using
+  ``qp.allocate(state="magic-T")`` and ``qp.allocate(state="magic-T-adj")``, which makes it
+  easier to compile FTQC-style routines that need T-state ancillas on the fly (for example
+  TemporaryAND) with ``qjit(capture=True)``.
```
dime10, line 8: Nice, thanks for updating the changelog entry :) just some minor fixes:
- pr link goes at the top
- linking to the pennylane pr is not required, the next release will have that pr so it's not something the user will need to pay attention to
```suggestion
  TemporaryAND) with ``qjit(capture=True)``.
  [(#3029)](https://github.com/PennyLaneAI/catalyst/pull/3029)
```

doc/releases/changelog-dev.md
```diff
@@ -2,6 +2,24 @@
...
+  @qnode(dev)
+  def circuit():
+      qb = qp.allocate(state="magic-T")
+      # ... use qb in your circuit ...
+      qp.deallocate(qb)
+  ```
+
+  Requires a PennyLane release that supports the new ``AllocateState`` values
+  (see [PennyLane #9846](https://github.com/PennyLaneAI/pennylane/pull/9846)).
+  [(#3029)](https://github.com/PennyLaneAI/catalyst/pull/3029)
```
```suggestion
```

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: We shouldn't count them here. The fabricate op already be taken care by the above condition. those will count as operation rather then memory op. That was a bug that we don't count allocation from pbc.fabricate and pbc.prepare. This will be fixed in https://github.com/PennyLaneAI/catalyst/pull/3025

frontend/catalyst/device/qjit_device.py
```diff
@@ -289,6 +289,14 @@ def get_qjit_device_capabilities(target_capabilities: DeviceCapabilities) -> Dev
         {"Snapshot": OperatorProperties(invertible=False, controllable=False, differentiable=False)}
     )
 
+    # Catalyst supports dynamic qubit allocation on qjit-compatible devices. Target device
+    # TOMLs (e.g. lightning) still default this flag to false; expose Catalyst support via
+    # qjit capabilities so verification can check the field rather than bypassing Allocate.
+    # TODO: remove this override once all qjit-compatible backends declare
+    # dynamic_qubit_management=true in their device TOMLs.
+    if target_capabilities.qjit_compatible:
+        qjit_capabilities.dynamic_qubit_management = True
```
dime10, line 298: Commenting here to open an issue:

We should update devices to properly declare whether they support dynamic allocation or not. Both null.qubit and lightning.qubit should support it. Then the TODO and capability override can be removed.

frontend/catalyst/from_plxpr/qfunc_interpreter.py
```diff
@@ -453,31 +452,42 @@ def _pauli_rot_bind_call(*invals, op, qubits_len, params_len, ctrl_len, adjoint,
 @PLxPRToQuantumJaxprInterpreter.register_primitive(qp.allocation.allocate_prim)
 def handle_allocate(self, *, num_wires, state=None, restored=False):
     """Handle the conversion from plxpr to Catalyst jaxpr for the qp.allocate primitive"""
+    # TODO: remove this wrapper once PLxPRToQuantumJaxprInterpreter can emit Catalyst
+    # jaxpr primitives directly without rebinding PennyLane allocate/deallocate.
```
dime10, line 455: Commenting for new issue:

Eliminate the rebinding of pl allocation primitives during the from_plxpr lowering, since these primitives now have a direct to mlir lowering. Then the TODO can be removed.

Also remove the qref_alloc_p and qref_dealloc_p primitives, and replace their use in the handle_qnode/calling_convention function with the equivalent pennylane primitives.

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
Qottmann: removed `pbc::FabricateOp` ✅ 😁

dime10: That should equally apply to ref version too no?

mlir/lib/QRef/Transforms/value_semantics_conversion.cpp
```diff
@@ -218,6 +218,16 @@ void eraseAllRemainingAnchorRValues(func::FuncOp f)
                "qref.reg Values must have no uses after the semantic conversion");
         graphStatePrepOp->erase();
     });
+    f.walk([&](qref::AllocQubitOp allocQbOp) {
+        assert(allocQbOp.use_empty() &&
+               "qref.bit Values must have no uses after the semantic conversion");
+        allocQbOp->erase();
+    });
```
dime10, line 225: Leaving this here for someone else to confirm, since the op already existed but wasn't listed here. (@paul0403)

dime10 (approved):
Excellent work 💯 

I'm happy with it, leaving some final comments. @paul0403 Could you provide a second review? :)

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: Yes, that same to ref version.

Qottmann: ✅

doc/releases/changelog-dev.md
Qottmann: ah sorry havent finalized the changelog yet, you're too quick! :D

doc/dev/sharp_bits.rst
Qottmann: I think I was just confused what was meant by being supported in pennylane. It is supported on the frontend to put a magic-T literal in qp.allocate, but the actual allocation is not supported 🤦 updating this

mlir/lib/QRef/Transforms/value_semantics_conversion.cpp
paul0403: Yeah it should have been listed. 

No bugs were surfaced because: 
1. Nothing in frontend used single qubit alloc op
2. The default pipeline runs canonicalization right after the semantcis conversion, which triggers the "remove if no user" canonicalization patterns on the allocation ops themselves

But because this pass should still be correct as a standalone (without canonicalization), it would be correct to list alloc_qb op here too.

Good catch!

mlir/lib/QRef/Transforms/value_semantics_conversion.cpp
```diff
@@ -1205,6 +1215,17 @@ void handlePPM(IRRewriter &builder, pbc::RefPPMeasurementOp rPPMOp, QubitValueTr
...
 
+void handleRefFabricate(IRRewriter &builder, pbc::RefFabricateOp rFabricateOp,
+                        QubitValueTracker &tracker)
+{
+    OpBuilder::InsertionGuard guard(builder);
+    builder.setInsertionPoint(rFabricateOp);
+    Location loc = rFabricateOp.getLoc();
+
+    auto fabricateOp = pbc::FabricateOp::create(builder, loc, rFabricateOp.getInitState());
+    tracker.setCurrentVQubit(rFabricateOp.getQubit(), fabricateOp.getOutQubits().front());
```
paul0403, line 1226: Why is there a discrepency between the two semantics? Why does the value semantics one return a list of qubits, but the reference semantics one only return one qubit?

mlir/test/QRef/SemanticConversion/TestPBC.mlir
```diff
@@ -48,3 +48,27 @@ func.func @test_PPM_op(%angle: f64) -> (i1, i1, i1) attributes {quantum.node} {
...
+    return
+}
+
+// -----
+
+// CHECK-LABEL: test_ref_fabricate_magic_conj
+func.func @test_ref_fabricate_magic_conj() attributes {quantum.node} {
+    // CHECK: [[magic_conj:%.+]] = pbc.fabricate magic_conj : !quantum.bit
+    %q = pbc.ref.fabricate magic_conj : !qref.bit
+    qref.custom "PauliX"() %q : !qref.bit
```
paul0403, line 71: Can you check the X gate is on `magic_conj`?

frontend/catalyst/from_plxpr/qfunc_interpreter.py
```diff
@@ -453,31 +452,42 @@ def _pauli_rot_bind_call(*invals, op, qubits_len, params_len, ctrl_len, adjoint,
...
 
     assert isinstance(
         num_wires, int
     ), "number of dynamically allocated qubits must be statically known"
 
     self.has_dynamic_allocation = True
-    new_qreg = qref_alloc_p.bind(static_num_qubits=num_wires)
-    return [qref_get_p.bind(new_qreg, i) for i in range(num_wires)]
```
paul0403, line 463: Nice! Moving the `GetOp` into the lowering is definitely a bit cleaner.

frontend/catalyst/from_plxpr/qref_jax_primitives.py
```diff
@@ -225,6 +228,81 @@ def _qref_get_abstract_eval(qreg, qubit_idx):
...
+    if state_val in ("magic", "magic_conj"):
+        qubit_type = ir.OpaqueType.get("qref", "bit", ctx)
+        init_state_attr = _pbc_logical_init_attr(ctx, state_val)
+        return [
+            RefFabricateOp(qubit=qubit_type, init_state=init_state_attr).results[0]
+            for _ in range(num_wires)
+        ]
+
+    size_attr = ir.IntegerAttr.get(ir.IntegerType.get_signless(64, ctx), num_wires)
+    qreg_type = ir.OpaqueType.get("qref", f"reg<{num_wires}>", ctx)
```
paul0403, line 266: What happens if the number of wires is dynamic? In the IR the type would be `qref.reg<?>`, which means `num_wires` here would be the string `"?"`. 

I don't think this is what's happening on the PL alloc primitive's `num_wires` param?

paul0403: It also exposes the fact that we have no tests for an allocation of a dynamic size. But I think it's because the PL alloc primitive cannot allocate a dynamic size of qubits yet (since it returns a list of abstract qubit values, and lists need a static size)

frontend/test/pytest/test_dynamic_qubit_allocation.py
```diff
@@ -549,6 +574,72 @@ def circuit():
...
+    @qp.qnode(qp.device(backend, wires=1))
+    def circuit():
+        with qp.allocate(1, state="magic-T") as q:
+            qp.X(q[0])
+        return qp.probs(wires=[0])
+
+    mlir_str = circuit.mlir
+    assert "pbc.ref.fabricate" in mlir_str
+    assert "magic" in mlir_str
+    assert "qref.dealloc_qb" in mlir_str
```
paul0403: We don't need this test since the lit test already checks the compiled mlir

mlir/lib/QRef/Transforms/value_semantics_conversion.h
```diff
@@ -57,6 +57,8 @@ void handleMeasure(IRRewriter &builder, qref::MeasureOp rMeasureOp, QubitValueTr
 void handleMeasureInBasis(IRRewriter &builder, mbqc::RefMeasureInBasisOp rMeasureInBasisOp,
                           QubitValueTracker &tracker);
 void handlePPM(IRRewriter &builder, pbc::RefPPMeasurementOp rPPMOp, QubitValueTracker &tracker);
+void handleRefFabricate(IRRewriter &builder, pbc::RefFabricateOp rFabricateOp,
```
paul0403, line 60: What about the conversion in the other direction? i.e. the conversion from value to ref

frontend/catalyst/from_plxpr/qref_jax_primitives.py
Qottmann: yes I think this is fine for now, added a comment

frontend/test/pytest/test_dynamic_qubit_allocation.py
Qottmann: removed ✅

mlir/test/QRef/SemanticConversion/TestPBC.mlir
Qottmann: good point! ✅

mlir/lib/QRef/Transforms/value_semantics_conversion.cpp
Qottmann: :cozy-blob:

mlir/lib/QRef/Transforms/value_semantics_conversion.cpp
Qottmann: `qp.allocate(k, state="magic-T")` lowers to k separate `pbc.ref.fabricate` ops (loop in _allocate_prim_lowering), not one ref op with k results.

`pbc.fabricate` has variadic `out_qubits` (same pattern as `pbc.prepare`) so we can build one or many !quantum.bit results in a single op. The ref-side `pbc.ref.fabricate` op is intentionally one `!qref.bit` per op (consistent with other ref memory ops like `alloc_qb`)

In this PR we never emit a multi-result ref fabricate: allocate_prim lowering creates one ref.fabricate per wire, and handleRefFabricate maps each to a value FabricateOp using the single-qubit builder, then tracks getOutQubits().front(). So the conversion is 1 ref qubit → 1 value qubit. The variadic type is just how the value op is defined, not an N-for-1 mismatch in what we generate today.

mlir/lib/QRef/Transforms/value_semantics_conversion.h
Qottmann: I only need Ref -> value for `pbc.ref.fabricate` for this PR. Do you want me to add the value -> ref path here? I think it'd be out of scope 🤔 Note that pbc.prepare doesn’t have a ref counterpart in ODS, so full PBC prep/fabricate symmetry would definitely be a follow-up 😬

paul0403: It would be nice to have the value->ref since it's also quite straightforward, and one less thing to keep in the back of our minds and worry about in the future.

It's in `mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp`, it's the same handling as the other direction. You can just copy paste the handler function.

Qottmann: Ah nice! Okay with this I can add handleFabricate mirroring handleRefFabricate!

mlir/lib/QRef/Transforms/value_semantics_conversion.cpp
paul0403: > `pbc.fabricate` has variadic `out_qubits` (same pattern as `pbc.prepare`) so we can build one or many !quantum.bit results in a single op. The ref-side `pbc.ref.fabricate` op is intentionally one `!qref.bit` per op

Sorry, this is the discrepency I was wondering about. Why does `pbc.ref.fabricate` only return one qubit value, but `pbc.fabricate` returns a variadic list of qubit values? The frontend lowering should serve the mlir, and the mlir definition should be the source of truth. 

Is it because in plxpr the allocation primitive returns a list of individual qubit values? So when interpreting/lowering the plxpr, the allocate primitive also has to return a list of individual qubit values?

paul0403: In any case, the fact that plxpr alloc prim only returns a static number of qubits is probably what constrained the lowering and therefore the `pbc.ref.fabricate` definition here. We can revisit when the plxpr alloc prim can return register types :+1:

mlir/test/Quantum/SemanticsConversion/TestPBC.mlir
```diff
@@ -46,3 +46,29 @@ func.func @test_PPM_op(%angle: f64) -> (i1, i1, i1) attributes {quantum.node} {
...
+// CHECK-LABEL: test_fabricate_magic
+func.func @test_fabricate_magic() attributes {quantum.node} {
+    // CHECK: [[magic:%.+]] = pbc.ref.fabricate magic : !qref.bit
+    %q = pbc.fabricate magic : !quantum.bit
+    quantum.custom "PauliX"() %q : !quantum.bit
+    // CHECK: qref.custom "PauliX"() [[magic]] : !qref.bit
+    // CHECK: qref.dealloc_qb [[magic]] : !qref.bit
+    quantum.dealloc_qb %q : !quantum.bit
+    return
+}
```
paul0403, line 61: Quantum dialect is value semantics, so it needs an output.

```mlir
func.func @test_fabricate_magic() attributes {quantum.node} {
    %q = pbc.fabricate magic : !quantum.bit
    %out_qubit = quantum.custom "PauliX"() %q : !quantum.bit
    quantum.dealloc_qb %out_qubit : !quantum.bit
    return
}
```

paul0403: Same for the other test

paul0403 (commented):
Thanks! Almost there

mlir/test/QRef/SemanticConversion/TestPBC.mlir
```diff
@@ -57,7 +57,7 @@ func.func @test_ref_fabricate_magic() attributes {quantum.node} {
     %q = pbc.ref.fabricate magic : !qref.bit
     qref.custom "PauliX"() %q : !qref.bit
     // CHECK: quantum.custom "PauliX"() [[magic]] : !quantum.bit
-    // CHECK: quantum.dealloc_qb {{.*}} : !quantum.bit
+    // CHECK: quantum.dealloc_qb [[magic]] : !quantum.bit
```
paul0403: Same here, the converted result in value semantics would have dealloc take in the output of the X gate

Qottmann: ✅

mlir/test/Quantum/SemanticsConversion/TestPBC.mlir
Qottmann: ✅

paul0403 (approved):
Awesome! 🥳

doc/dev/sharp_bits.rst
```diff
@@ -1376,9 +1376,16 @@ Currently, however, this is not the case for the following functionalities.
 - **Dynamic wire allocation behaviour**: The ``qp.allocate()`` function currently
   behaves differently when Catalyst is present or not. In particular:
 
-  - The ``state`` and ``restored`` keyword arguments of ``qp.allocate()`` are
-    ignored in Catalyst. The reason is that the only supported mode is to always allocate in the
-    zero state.
+  - The ``state="any"`` configuration is not supported by Catalyst and will raise a
+    compilation error. The ``restored`` keyword is ignored. Catalyst supports
+    ``state="zero"`` initialization (the default), as well as magic state
+    initialization (see below).
```
albi3ro, line 1382: Why can't we just continue treating `state="any"` as `state="zero"`? A little less efficient, but should still work?

Qottmann: I'm not sure I understand, why should we treat "any" state as "zero"? 🤔

PR #3030, opened 2026-07-16
Isolate executor kernels in per-object JITDylibs

mehrdad2m:
**Context:**
ross-compiled/external kernels reuse fixed entry-wrapper symbol names (_catalyst_pyface_<entry>). Loading two such objects into one session would collide due to duplicate-definition.

Give each cross-compiled kernel object shipped to an executor its own JITDylib, keyed by the object's file path, instead of loading everything into one shared namespace.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


runtime/lib/executor/ExecutorSession.cpp
```diff
@@ -251,11 +257,23 @@ struct ExecutorSession {
         return std::make_unique<ExecutorSession>(std::move(ES), std::move(*DL));
     }
 
-    Error addObjectFile(std::unique_ptr<MemoryBuffer> Buf)
+    Error addObjectFile(StringRef path, std::unique_ptr<MemoryBuffer> Buf)
     {
-        return ObjectLayer.add(MainJD, std::move(Buf));
+        JITDylib &jd = ES->createBareJITDylib(("kernel:" + path).str());
```
rniczh, line 266: Is an object path really required, or would any symbol name work? And about the `catalyst_pyface`, wouldn't it has a suffix to represent the function? If I remembered: `_catalyst_pyface_jit_<func_name>`

mehrdad2m: Here it could be anything but we also keep track of the path in the `KernelJDs[path] = &jd;` line.

> wouldn't it has a suffix to represent the function? If I remembered: _catalyst_pyface_jit_<func_name>

As far as I remember we add the `jit` prefix in the frontend to the name of the main entry point function. otherwise it shouldn't exist.

rniczh (approved):
💯

maliasadi (approved):
🦘

runtime/lib/executor/ExecutorSession.cpp
maliasadi, line 266: do we need to check `KernelJDs`?
```suggestion
        if (KernelJDs.count(path)) {
            return make_error<StringError>("object already loaded: " + path,
                                           inconvertibleErrorCode());
        }
        JITDylib &jd = ES->createBareJITDylib(("kernel:" + path).str());
```

PR #3031, opened 2026-07-16
Imorove executor session lifetime

mehrdad2m:
**Context:**
Adding two robustness fixes to the executor runtime's ORC session 

- Added a watchdog that aborts if the peer isn't a live executor, instead of blocking forever.
- Closing a session could hang forever, because a background thread kept waiting on the connection that never fully closed. We now shut the connection down properly so sessions always exit cleanly. (This is mainly assisted by Claude)
-  Also wires -DLLVM_DIR into the runtime Makefile so the executor lib builds 

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


rniczh (approved):
💯

maliasadi (approved):
🙌

PR #3033, opened 2026-07-16
Cross-compile target modules and dispatch to executor

mehrdad2m:
**Context:**
This PR adds a compile path that turns a catalyst.target-tagged QNode into a standalone object and either statically links it (local) or ships it to a remote executor. 

- Adds `cross-compile-targets ` pass which lowers each `catalyst.target` nested module and emits a standalone .o. Then it records the path as `catalyst.object_file`, and reduce the module to external declarations. Local targets are also recorded for static linking.
- Adds `dispatch-executor-targets` pass which ships the remote binaries and rewrites their host calls into `executor.launch`.
- Modifies the compiler driver to run both passes after the bufferization stage and writes the local objects to a <module>.objects manifest for the linker.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


mlir/Makefile
```diff
@@ -76,7 +76,7 @@ llvm:
 	cmake -G Ninja -S llvm-project/llvm -B $(LLVM_BUILD_DIR) \
 		-DCMAKE_BUILD_TYPE=$(BUILD_TYPE) \
 		-DLLVM_BUILD_EXAMPLES=OFF \
-		-DLLVM_TARGETS_TO_BUILD=${LLVM_TARGETS_TO_BUILD} \
+		-DLLVM_TARGETS_TO_BUILD="${LLVM_TARGETS_TO_BUILD}" \
 		-DLLVM_ENABLE_PROJECTS="$(LLVM_PROJECTS)" \
 		-DLLVM_ENABLE_ASSERTIONS=ON \
```
maliasadi, line 89: this is consistent with other places: 
```suggestion
		-DLLVM_TARGETS_TO_BUILD=${LLVM_TARGETS_TO_BUILD} \
		-DLLVM_ENABLE_PROJECTS=$(LLVM_PROJECTS) \
		-DLLVM_ENABLE_ASSERTIONS=ON \
```

mlir/test/Executor/DispatchExecutorTargetModules.mlir
```diff
@@ -0,0 +1,70 @@
...
+// RUN: quantum-opt %s --dispatch-executor-targets | FileCheck %s
+
+// This pass consumes the input IR left by cross-compile-targets which carries catalyst.object_file,
+// plus a catalyst.dispatch attributes.
+
+// setup() opens the session once and ships the module's object.
+// CHECK-LABEL: func.func @setup()
+// CHECK: executor.open("ADDR:PORT")
+// CHECK: executor.send_binary("ADDR:PORT", "/tmp/target_compute.o")
+// CHECK-NOT: executor.send_binary
```
maliasadi: What do CHECK and CHECK-NOT `executor.send_binary` check?

mlir/include/Executor/Transforms/Passes.td
```diff
@@ -36,4 +36,29 @@ def ConvertExecutorToLLVMPass : Pass<"convert-executor-to-llvm", "mlir::ModuleOp
     ];
 }
 
+def DispatchExecutorTargetsPass : Pass<"dispatch-executor-targets", "mlir::ModuleOp"> {
+    let summary = "Ship cross-compiled catalyst.target modules to a executor executor.";
```
maliasadi:
```suggestion
    let summary = "Ship cross-compiled `catalyst.target` modules to an executor.";
```

mlir/include/Executor/Transforms/Passes.td
```diff
@@ -36,4 +36,29 @@ def ConvertExecutorToLLVMPass : Pass<"convert-executor-to-llvm", "mlir::ModuleOp
...
+    let description = [{
+        For every `builtin.module` carrying a `catalyst.dispatch` attribute and the
+        `catalyst.object_file` path, this pass will perform the following:
+
+        1. Injects `executor.open` into `setup()` (once per unique address) and `executor.send_binary`
+           into `setup()` (once per module).
+        2. Replaces every host-side `func.call` to the module's entry functions with a
```
maliasadi: What's the strategy to replace host `func.call` with `executor.launch`? What does filter the host-side `func.call`s  which need to run on the host from those need to be handled with `executor`?

mlir/lib/Executor/Transforms/DispatchExecutorTargets.cpp
```diff
@@ -0,0 +1,267 @@
...
+            nested.erase();
+        }
+
+        const size_t numQnodeExecutors = openedAddresses.size();
+
+        if (!libCalls.empty()) {
+            for (catalyst::CustomCallOp call : libCalls) {
+                StringAttr dispatch = executorDispatchOf(call);
+                if ((!dispatch || dispatch.getValue().empty()) && numQnodeExecutors > 1) {
+                    call.emitOpError("ambiguous executor executor");
```
maliasadi:
```suggestion
                    call.emitOpError("ambiguous executor");
```

mlir/test/Executor/DispatchExecutorLibCalls.mlir
```diff
@@ -0,0 +1,95 @@
...
+// When the program targets more than one executor, the custom_call should explicitly bind the 
+// dispatch otherwise it would be ambiguous.
+module @jit_ambiguous {
+  func.func @setup() {
+    return
+  }
+  func.func @main() {
+    catalyst.launch_kernel @t1::@c1() : () -> ()
+    catalyst.launch_kernel @t2::@c2() : () -> ()
+    // expected-error @below {{ambiguous executor executor}}
```
maliasadi:
```suggestion
    // expected-error @below {{ambiguous executor}}
```

mlir/lib/Executor/Transforms/DispatchExecutorTargets.cpp
```diff
@@ -0,0 +1,267 @@
...
+
+namespace catalyst {
+namespace executor {
+
+#define GEN_PASS_DEF_DISPATCHEXECUTORTARGETSPASS
+#include "Executor/Transforms/Passes.h.inc"
+
+namespace {
+
+// Ships cross-compiled `catalyst.target` modules to a executor executor using the `executor` dialect.
```
maliasadi:
```suggestion
// Ships cross-compiled `catalyst.target` modules to a executor using the `executor` dialect.
```

mlir/lib/Executor/Transforms/DispatchExecutorTargets.cpp
```diff
@@ -0,0 +1,267 @@
...
+        if (!libCalls.empty()) {
+            for (catalyst::CustomCallOp call : libCalls) {
+                StringAttr dispatch = executorDispatchOf(call);
+                if ((!dispatch || dispatch.getValue().empty()) && numQnodeExecutors > 1) {
+                    call.emitOpError("ambiguous executor executor");
+                    return signalPassFailure();
+                }
+                StringAttr addrAttr = libCallAddress(call, executorAddress);
+                if (!addrAttr) {
+                    call.emitOpError("executor runtime_call has no executor address");
```
maliasadi: what about this?
```suggestion
                    call.emitOpError("custom_call dispatch has no executor address");
```

mlir/lib/Executor/Transforms/DispatchExecutorTargets.cpp
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    LogicalResult rewriteExecutorLibCalls(ArrayRef<catalyst::CustomCallOp> libCalls,
+                                        StringAttr fallbackAddress)
+    {
+        MLIRContext *ctx = &getContext();
+        for (catalyst::CustomCallOp call : libCalls) {
+            StringAttr addressAttr = libCallAddress(call, fallbackAddress);
+            if (!addressAttr) {
+                call.emitOpError("executor runtime_call has no executor address");
```
maliasadi:
```suggestion
                call.emitOpError("custom_call dispatch has no executor address");
```

mlir/lib/Executor/Transforms/DispatchExecutorTargets.cpp
```diff
@@ -0,0 +1,267 @@
...
+        if (!terminator) {
+            return;
+        }
+        OpBuilder b(terminator);
+        executor::SendBinaryOp::create(b, setupFn.getLoc(), addressAttr, pathAttr);
+    }
+
+    // Open a session (once per address), ship the object recorded in
+    // `catalyst.object_file`, and rewrite each host-side func.call into a `executor.launch`.
+    LogicalResult dispatchViaOrcExecutor(ModuleOp host, ModuleOp nested,
```
maliasadi: What does this function do? 🤔

mlir/lib/Catalyst/Transforms/CrossCompileTargets.cpp
```diff
@@ -0,0 +1,429 @@
...
+        codegenPM.run(*llvmModule);
+        dest.flush();
+        return objPath;
+    }
+
+    // Cross-compile a catalyst.target module to a `.o` for its triple and return the path.
+    FailureOr<std::string> compileTargetModule(ModuleOp nested)
+    {
+        MLIRContext *ctx = &getContext();
+        StringRef name = nested.getName().value_or("unnamed");
```
maliasadi: hmm shouldn't be `SynName`? looks like an inconsistency `getName` vs. `getSymName` 🤔 

```suggestion
        StringRef name = nested.getSymName().value_or("unnamed");
```

mlir/lib/Catalyst/Transforms/CrossCompileTargets.cpp
```diff
@@ -0,0 +1,429 @@
...
+            translateModuleToLLVMIR(*standalone, llvmCtx, name);
+        if (!llvmModule) {
+            nested.emitError("failed to translate target module to LLVM IR: " + name.str());
+            return failure();
+        }
+        if (dumpIntermediate) {
+            dumpLLVMIR(*llvmModule, kernelDir, name.str() + ".ll");
+        }
+        std::string objPath =
+            emitObjectFile(std::move(llvmModule), name, kernelDir, *targetMachine);
```
maliasadi, line 419: Could you add a guard before the compile loop to fix the object-file name collision issue? as it generates `{workspace}/{name}/{name}.o` and for two targets with the same sym name, this will silently fail.

mlir/lib/Driver/CompilerDriver.cpp
```diff
@@ -483,6 +483,39 @@ llvm::LogicalResult catalyst::driver::runPipeline(PassManager &pm, const Compile
...
+            catalyst::utils::LinesCount::call(moduleOp);
+
+            // cross-compile-targets records the objects of local (statically-linked) targets on the
+            // root module. Write them to a manifest the frontend hands to the linker.
+            if (auto objFiles = moduleOp->getAttrOfType<mlir::ArrayAttr>("catalyst.object_files")) {
+                std::string manifestPath =
+                    options.workspace.str() + "/" + options.moduleName.str() + ".objects";
+                std::error_code ec;
+                llvm::raw_fd_ostream manifest(manifestPath, ec);
+                if (!ec) {
```
maliasadi: I'd replace this with a failure:

```suggestion
                if (ec) {
                    moduleOp->emitError("failed to write object manifest '")
                        << manifestPath << "': " << ec.message();
                    return failure();
                } else {
```

mlir/include/Catalyst/Transforms/Passes.td
```diff
@@ -143,6 +143,43 @@ def ApplyTransformSequencePass : Pass<"apply-transform-sequence"> {
     let summary = "Apply the passes scheduled with the transform dialect.";
 }
 
+def CrossCompileTargetsPass : Pass<"cross-compile-targets", "mlir::ModuleOp"> {
```
maliasadi, line 146: 🤩

maliasadi (commented):
Thanks @mehrdad2m! Will you fix the build issues on this PR or on the base branch/PR?

mlir/lib/Catalyst/Transforms/CrossCompileTargets.cpp
mehrdad2m: I think they are the same as long as `nested` is a `moduleOp`, but you are right about inconsistency. Fixed

mlir/lib/Catalyst/Transforms/CrossCompileTargets.cpp
mehrdad2m: Good catch! Although having two targets with the same sym name would not be a valid IR to begin with, we default to `"unnamed"` for modules with no name and that could trigger this issue. I added a guard https://github.com/PennyLaneAI/catalyst/pull/3033/commits/e22b237eebe260b546a77171c130dd817070291d

mlir/Makefile
mehrdad2m: This was added to allow multiple targets/project to be parsed correctly. e.g. without the quotation, cmake can not parse `DLLVM_TARGETS_TO_BUILD=X86;AArch64` correctly. Let me know if you know a cleaner way to achieve this.

mlir/test/Executor/DispatchExecutorTargetModules.mlir
mehrdad2m: This was added to ensure we only send the binary once even if we have multiple launch_kernels on the same object file. I'll add a comment to clarify that

mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/3033/commits/948d5d3be0854e92c4e267fb221614c1c14634fa

mlir/include/Executor/Transforms/Passes.td
mehrdad2m: Ah good catch. This docstring is outdated. I updated it in https://github.com/PennyLaneAI/catalyst/pull/3033/commits/2aafda2df04f90673cb9836bf81454e0beb662e1. the `func.call`s are unaffected by this pass. It only acts on nested modules with `catalyst.dispatch` attribute.

mlir/lib/Executor/Transforms/DispatchExecutorTargets.cpp
mehrdad2m: It handles one dispatch module which involves opening the session, shipping the object, and rewriting that module's `launch_kernels` into `executor.launch`. I just noticed though that the name carries an implementation detail it shouldn't, I renamed it to `dispatchTargetModule`. https://github.com/PennyLaneAI/catalyst/pull/3033/commits/350665b300ae10f11e826a043309a97a57629406

maliasadi (approved):
Thanks @mehrdad2m! Happy to approve! But let me surface the remaining concerns for the future reference. These are not blockers! They can be discussed offline and fixed as followups.

1.  The `cross-compile-targets` pass orchestrates a full compilation pipeline (MLIR -> LLVMIR -> object file). All these steps are implemented inside this MLIR pass. To me, this is the responsibility of the driver! I'm not sure about the difficulties this design will force to the E2E pipeline. I reckon this may not be the cleanest solution either. As an alternative that's worth considering here is to keep this pass IR-to-IR and let the driver to take its part. I think we can isolate each target into a module and annotate it via device triple and let the driver own the lowering part to LLVMIR + emit object. I'd be interested in learning about the concerns with my suggestion.

2. Right now,  there is a strong implicit ordering relation between `cross-compile-targets` and `dispatch-executor-targets`. `dispatch-executor-targets` depends on `cross-compile-targets` having populated `object_file`, but there's nothing enforces or verifies the ordering! It'd be good to orchestrate them to ensure one's always called before the other in a more automated way.

3. In `CrossCompilerTargets.cpp`, `CompileTargetModule` walks the host for `LaunchKernelOps` to build the kernel launch entries, and `lowerLocalTargetCalls` walks it again to rewrite them. This is not an issue now but if this can be a scalability problem when there are many kernels and many targets. I wonder if we can use a single pass to create a `moduleName -> {entries, launches}` map and use this map to reduce the double walk in the IR.

4. [TESTING] There are a couple of missing tests based on previous review comments:
    - There's no test for when `workspace.empty()` in `CrossCompileTargets.cpp`.
    - The kernel/obj-file naming collision (e.g., two "noname" kernels) is not tested.

maliasadi:
Don't forget to add a changelog entry :smile:

mehrdad2m:
Thanks @maliasadi for the comments. 

> The cross-compile-targets pass orchestrates a full compilation pipeline (MLIR -> LLVMIR -> object file). All these steps are implemented inside this MLIR pass. To me, this is the responsibility of the driver! I'm not sure about the difficulties this design will force to the E2E pipeline. I reckon this may not be the cleanest solution either. As an alternative that's worth considering here is to keep this pass IR-to-IR and let the driver to take its part. I think we can isolate each target into a module and annotate it via device triple and let the driver own the lowering part to LLVMIR + emit object. I'd be interested in learning about the concerns with my suggestion.

Yeah this is certainly an alternative we could consider. I think this certainly makes the architecture of the driver cleaner compared to the current one-pass-handle-everything approach, but it may come with a performance cost. In fact that was my first instinct as well, but I ended up not implementing it because I couldn't figure out a way to do this without introduction of a new print/parse cost for the cross-compiled module. I would be nice to spend some time to analyse the tradeoffs. Happy to discuss more.

> It'd be good to orchestrate them to ensure one's always called before the other in a more automated way. 

Yeah for sure, that we can definitely improve in a follow up. For more context These two passes were implemented in a one-shot pass but I tried to detach the distribution from compilation since one might consider different distribution routes in future.

> In CrossCompilerTargets.cpp, CompileTargetModule walks the host for LaunchKernelOps to build the kernel launch entries, and lowerLocalTargetCalls walks it again to rewrite them.

Thanks agreed. this should be a fairly easy fix. I'll leave that to the follow up if that's OK for you.

> [TESTING] There are a couple of missing tests based on previous review comments:

Thanks I'll add those tests shortly

mehrdad2m:
> The kernel/obj-file naming collision (e.g., two "noname" kernels) is not tested.

mlir/test/Catalyst/CrossCompileTargetNameCollision.mlir is already added for this. 

> There's no test for when workspace.empty() in CrossCompileTargets.cpp.

Added https://github.com/PennyLaneAI/catalyst/pull/3033/commits/eed8ee5bf4069afc2f9c6198fae4fa4305cffee6

> Don't forget to add a changelog entry 😄

https://github.com/PennyLaneAI/catalyst/pull/3033/commits/9fd52fe2ce5b2857c5d7d3c56b3e84abef2c4306

mlir/lib/Catalyst/Transforms/CrossCompileTargets.cpp
```diff
@@ -0,0 +1,437 @@
...
+    auto llvmPasses = driver::getLLVMDialectLoweringStage();
+    std::vector<std::string> passes;
+    passes.reserve(buf.size() + llvmPasses.size());
+    passes.insert(passes.end(), buf.begin(), buf.end());
+    for (const auto &passName : llvmPasses) {
+        if (passName != "convert-executor-to-llvm") {
+            passes.push_back(passName);
+        }
+    }
+    return passes;
```
rniczh, line 86: would it more reasonable if we just checking the module attribute in `convert-executor-to-llvm` (to skip it) rather than removing the pass from pass list here?

mlir/lib/Executor/Transforms/DispatchExecutorTargets.cpp
```diff
@@ -0,0 +1,271 @@
...
+        ModuleOp host = getOperation();
+
+        SmallVector<ModuleOp> targetMods;
+        for (auto &op : host.getBody()->getOperations()) {
+            if (auto mod = dyn_cast<ModuleOp>(&op)) {
+                if (mod->hasAttr("catalyst.dispatch")) {
+                    targetMods.push_back(mod);
+                }
+            }
+        }
```
rniczh: I think you can simplify it by using `getOps`
```suggestion
        for (auto mod : host.getBody()->getOps<ModuleOp>()) {
            if (mod->hasAttr("catalst.dispatch")) {
                targetMods.push_back(mod);
            }
        }
```

rniczh (commented):
Thanks @mehrdad2m, just some small comments

mlir/lib/Catalyst/Transforms/CrossCompileTargets.cpp
mehrdad2m: I think I added that becasue of a weird error in dialect registration otherwise the `convert-executor-to-llvm` should be noop unless we have nested executors. I just realized though we can remove it by simply registering the dialect, which make the code here simpler.  https://github.com/PennyLaneAI/catalyst/pull/3033/commits/179fdd7a5b49ea1e249813c4e7675a233adb54f4

mlir/lib/Catalyst/Transforms/CrossCompileTargets.cpp
```diff
@@ -0,0 +1,432 @@
...
+        ModuleOp host = getOperation();
+
+        SmallVector<ModuleOp> targetMods;
+        for (auto &op : host.getBody()->getOperations()) {
+            if (auto mod = dyn_cast<ModuleOp>(&op)) {
+                if (mod->hasAttr("catalyst.target")) {
+                    targetMods.push_back(mod);
+                }
+            }
+        }
```
rniczh: I think you can simplify this by using `getOps<ModuleOp>` as well.

```suggestion
        for (auto mod : host.getBody()->getOps<ModuleOp>()) {
            if (mod->hasAttr("catalyst.target")) {
                targetMods.push_back(mod);
            }
        }
```

rniczh (approved):
Awesome! The last comment on my end.

PR #3034, opened 2026-07-16
PCPhase op's `dim` parameter is static

paul0403:
**Context:**
The `dim` parameter on `PCPhase` op should be static. Like the Pauli word on `PauliRot` ops, different static instances of `dim` lead to different nodes in the decomposition graph. 

However, currently the `dim` parameter is taken in by `PCPhase` ops as a dynamic operand, not a static attribute.

Change `quantum/qref.pcphase` ops' `dim` operand into an attribute.
Primitives, lowerings, runtime CAPI and tests are updated accordingly. 

**Benefits:**
PCPhase op can better participate in decomp graph. This is because its decomp rule in python uses the dim as a piece of static data, and different values of `dim` give different decomp rules.

[sc-124926]

### DIFF
--- a/frontend/catalyst/from_plxpr/qref_jax_primitives.py
+++ b/frontend/catalyst/from_plxpr/qref_jax_primitives.py
@@ -307,7 +307,7 @@ def _qref_qinst_abstract_eval(
     # does not need to carry the PCPhase around
     # Or as an alternative, expand the kwarg so that it can be an arbitrary kwarg list for a general
     # primitive's attributes
-    assert not (pcphase_dim is not None) ^ (op == "PCPhase")
+    assert (pcphase_dim is not None) == (op == "PCPhase"), "pcphase_dim must be provided exactly for PCPhase ops"
     return ()
 
 
--- a/frontend/catalyst/from_plxpr/qref_jax_primitives.py
+++ b/frontend/catalyst/from_plxpr/qref_jax_primitives.py
@@ -285,6 +285,7 @@ def _qref_set_basis_state_lowering(jax_ctx: mlir.LoweringRuleContext, *qubits_or
 #
 # qref_qinst_p
 #
+# pylint: disable=too-many-arguments
 @qref_qinst_p.def_abstract_eval
 def _qref_qinst_abstract_eval(
     *qubits_or_params,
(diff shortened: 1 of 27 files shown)


kipawaa (approved):
Lovely, thanks @paul0403 !

frontend/catalyst/from_plxpr/qref_jax_primitives.py
```diff
@@ -285,16 +285,29 @@ def _qref_set_basis_state_lowering(jax_ctx: mlir.LoweringRuleContext, *qubits_or
...
     qubits = qubits_or_params[:qubits_len]
     ctrl_qubits = qubits_or_params[-2 * ctrl_len : -ctrl_len]
     all_qubits = qubits + ctrl_qubits
     assert all(isinstance(qubit, AbstractQubit) for qubit in all_qubits[: qubits_len + ctrl_len])
+
+    # TODO: make a specialized primitive for PCPhase so that the main qinst primitive's kwarg
+    # does not need to carry the PCPhase around
+    # Or as an alternative, expand the kwarg so that it can be an arbitrary kwarg list for a general
+    # primitive's attributes
+    assert not (pcphase_dim is not None) ^ (op == "PCPhase")
```
maliasadi: could you add a msg to these asserts?
```suggestion
    assert (pcphase_dim is not None) == (op == "PCPhase"), "pcphase_dim must be provided exactly for PCPhase ops"
```

frontend/catalyst/jax_primitives.py
```diff
@@ -1362,6 +1368,7 @@ def _qinst_abstract_eval(
     for idx in range(qubits_len + ctrl_len):
         qubit = all_qubits[idx]
         assert isinstance(qubit, AbstractQbit)
+    assert not (pcphase_dim is not None) ^ (op == "PCPhase")
```
maliasadi: and here:
```suggestion
    assert (pcphase_dim is not None) == (op == "PCPhase"), "pcphase_dim must be provided exactly for PCPhase ops"
```

maliasadi (approved):
🦕

frontend/catalyst/from_plxpr/qref_jax_primitives.py
paul0403: Makes sense!

Also of course the conditions could just be compared directly, my brain just reflexively likes xor for no reason hahaha

PR #3039, opened 2026-07-17
Add support for `graphOpId` to the Graph Solver

kipawaa:
**Context:**
With the introduction of the `DecomposableGate` interface we have the opportunity to simplify the graph solver by operating on `graphOpId`s. This adds introductory support to the solver for ops with `graphOpId`s while continuing to support the legacy pathway.

Adds the `id` field to `OperatorNode` and updates comparison functions to use this, with the legacy pathway as a fallback for compatibility.

**Benefits:**
Steps towards a simplified graph solver and generic op support in the graph.

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-124975]

### DIFF
--- a/mlir/lib/Quantum/Transforms/GraphDecomposition/DecompGraphSolver/DGTypes.hpp
+++ b/mlir/lib/Quantum/Transforms/GraphDecomposition/DecompGraphSolver/DGTypes.hpp
@@ -20,7 +20,9 @@
  */
 #pragma once
 
+#include <cstddef>
 #include <cstdint>
+#include <functional>
 #include <limits>
 #include <string>
 #include <unordered_map>
@@ -58,9 +60,16 @@ struct OperatorNode {
 
     // Optional static arguments for operators that require additional data.
     std::unordered_map<std::string, std::string> staticNamedArgs{};
+    std::string id{""};
 
     bool operator==(const OperatorNode &other) const
     {
+        // id match
+        if (!id.empty() && !other.id.empty() && id == other.id) {
+            return true;
+        }
+        // legacy fallback if either op is missing ID
+
         // For equality, we consider numWires and numParams conditionally equal
         // if they are not set to -1 (which indicates a wildcard that can match any value).
         const bool default_wires =
@@ -98,6 +107,10 @@ struct OperatorNode {
 struct OperatorNodeHash {
     std::size_t operator()(const OperatorNode &node) const
     {
+        // prefer id if available
+        if (!node.id.empty()) {
+            return std::hash<std::string>{}(node.id);
+        }
         return std::hash<std::string>{}(node.name);
     }
 };
@@ -108,7 +121,23 @@ struct OperatorNodeHash {
 struct WeightedGateset {
     std::unordered_map<OperatorNode, double, OperatorNodeHash> ops;
 
-    [[nodiscard]] bool contains(const OperatorNode &op) const { return ops.find(op) != ops.end(); }
+    [[nodiscard]] bool contains(const OperatorNode &op) const
+    {
+        // hash match
+        if (ops.find(op) != ops.end()) {
+            return true;
+        }
+
+        // use op-matching if hashes failed (could be id vs name)
+        for (auto [gatesetOp, cost] : ops) {
+            if (gatesetOp == op) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
     [[nodiscard]] double getCost(const OperatorNode &op) const
     {
         auto it = ops.find(op);
@@ -137,7 +166,8 @@ struct RuleTerm {
  * @brief This represents the origin of a decomposition rule.
  *
  * This enum is used to categorize decomposition rules based on their source or type:
- * - Default: The default rule for decomposing an operator as defined in the decomposition graph.
+ * - Default: The default rule for decomposing an operator as defined in the decomposition
+ * graph.
  * - Fixed: A fixed rule that cannot be changed or overridden by the solver.
  * - Alternative: An alternative rule that can be used in place of the default rule.
  */
@@ -154,7 +184,8 @@ enum class RuleOrigin : uint8_t { Default = 0, Fixed = 1, Alternative = 2 };
  *
  * TODO:
  * - We can add a field for work_wires_required if we want to consider the number of ancillary
- * wires needed for the decomposition, which can be an important factor in resource optimization.
+ * wires needed for the decomposition, which can be an important factor in resource
+ * optimization.
  * - We can also consider adding a field for the decomposition function or a pointer to it,
  * which can be used to actually perform the decomposition after the graph solver selects
  * the rules.
--- a/mlir/unittests/DecompGraphSolver/Test_DecompGraphCore.cpp
+++ b/mlir/unittests/DecompGraphSolver/Test_DecompGraphCore.cpp
@@ -146,3 +146,45 @@ TEST_CASE("Test ChosenDecompRule construction", "[DecompGraph::Core]")
     REQUIRE(chosenRule.basisCounts[rz] == 2);
     REQUIRE(chosenRule.basisCounts[rx] == 1);
 }
+
+TEST_CASE("Test graphOpId Support", "[DecompGraph::Core]")
+{
+    // comparing an op without an ID should fallback to legacy match
+    const OperatorNode h{"H"};
+    const OperatorNode hId{"H", -1, -1, false, {}, "H[][1]{}"};
+
+    REQUIRE(h == hId);
+
+    // id + legacy match with non-wildcard params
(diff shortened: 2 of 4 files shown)


mlir/unittests/DecompGraphSolver/Test_DecompGraphCore.cpp
```diff
@@ -146,3 +146,45 @@ TEST_CASE("Test ChosenDecompRule construction", "[DecompGraph::Core]")
     REQUIRE(chosenRule.basisCounts[rz] == 2);
     REQUIRE(chosenRule.basisCounts[rx] == 1);
 }
+
+TEST_CASE("Test graphOpId Support", "[DecompGraph::Core]")
```
paul0403, line 150: Just want to say thanks for setting up the mlir unit tests, we probably have many pass modules that don't have cpp unit tests (for example methods on the ops themselves, implemented in places like `QuantumOps.td/cpp`), but just tested by mlir lit tests. These are definitely a good start!

mlir/lib/Quantum/Transforms/GraphDecomposition/DecompGraphSolver/DGTypes.hpp
paul0403, line 59: Hmm, this makes me wonder whether the interface needs an adjoint method

mlir/lib/Quantum/Transforms/GraphDecomposition/DecompGraphSolver/DGTypes.hpp
```diff
@@ -98,6 +107,10 @@ struct OperatorNode {
 struct OperatorNodeHash {
     std::size_t operator()(const OperatorNode &node) const
     {
+        // prefer id if available
+        if (!node.id.empty()) {
+            return std::hash<std::string>{}(node.id);
+        }
         return std::hash<std::string>{}(node.name);
```
paul0403, line 114: Beautiful, now each node can differentiate between different static data versions of the same gate 🥳

mlir/lib/Quantum/Transforms/GraphDecomposition/DecompGraphSolver/DGTypes.hpp
```diff
@@ -20,7 +20,9 @@
  */
 #pragma once
 
+#include <cstddef>
```
paul0403, line 23: What is this header for?

mlir/lib/Quantum/Transforms/GraphDecomposition/DecompGraphSolver/DGTypes.hpp
```diff
@@ -108,7 +121,23 @@ struct OperatorNodeHash {
 struct WeightedGateset {
     std::unordered_map<OperatorNode, double, OperatorNodeHash> ops;
```
paul0403, line 122: This is a map between 3 things? 🤔

paul0403: Oh nvm, `OperatorNodeHash` is the hash function of the map 👍

mlir/lib/Quantum/Transforms/GraphDecomposition/DecompGraphSolver/DGTypes.hpp
```diff
@@ -108,7 +121,23 @@ struct OperatorNodeHash {
...
+        if (ops.find(op) != ops.end()) {
+            return true;
+        }
+
+        // use op-matching if hashes failed (could be id vs name)
+        for (auto [gatesetOp, cost] : ops) {
+            if (gatesetOp == op) {
+                return true;
+            }
+        }
```
paul0403, line 136: Just confirming: the preferred case is when both the ops in the map and the input op to `contains` are hashed from the new interface. But for backward compatibility, for all the other 3 cases we just fallback to the legacy checking (by `OperatorNode::Operator==`)?

Which means when the legacy is dropped eventually, this will just be `return ops.find(op) != ops.end()`?

mlir/unittests/DecompGraphSolver/Test_DecompGraphCore.cpp
```diff
@@ -146,3 +146,45 @@ TEST_CASE("Test ChosenDecompRule construction", "[DecompGraph::Core]")
...
+
+    // id + legacy match with non-wildcard params
+    const OperatorNode x{"X", 1, 0};
+    const OperatorNode xId{"X", -1, -1, false, {}, "X[][1]{}"};
+
+    REQUIRE(x == xId);
+
+    // id nodes should fail legacy match if params differ
+    const OperatorNode op1{"op", 1, 1};
+    const OperatorNode op2{"op", 2, 2, false, {}, "op[f64,f64][2]{}"};
```
paul0403, line 166: While we are still in the transition period, should we add some asserts in the constructor of `OperatorNode` that if both legacy and id init args are received, they agree with each other?

mlir/unittests/DecompGraphSolver/Test_DecompGraphCore.cpp
```diff
@@ -146,3 +146,45 @@ TEST_CASE("Test ChosenDecompRule construction", "[DecompGraph::Core]")
...
+
+    // nodes with same ids should match
+    const OperatorNode pr1{"PauliRot", -1, -1, false, {}, "PauliRot[f64][2]{pauli_word:XX}"};
+    const OperatorNode pr2{"PauliRot", -1, -1, false, {}, "PauliRot[f64][2]{pauli_word:XX}"};
+
+    REQUIRE(pr1 == pr2);
+
+    // nodes with matching ids should ignore other parameters (id is source of truth)
+    const OperatorNode id1{"name1", 1, 1, false, {}, "sameID"};
+    const OperatorNode id2{"name2", 2, 2, true, {}, "sameID"};
```
paul0403, line 178: Oh, we allow id and legacy info to mismatch? 🤔

mlir/unittests/DecompGraphSolver/Test_DecompGraphCore.cpp
```diff
@@ -146,3 +146,45 @@ TEST_CASE("Test ChosenDecompRule construction", "[DecompGraph::Core]")
...
+
+    REQUIRE(id1 == id2);
+
+    // check ID hashes match for same IDs, doesn't match name:
```
paul0403:
```suggestion
    // Unit test for `OperatorNodeHash`. Check that the hash function prefers ID over name
```

paul0403 (commented):
Beautiful!

mlir/lib/Quantum/Transforms/GraphDecomposition/DecompGraphSolver/DGTypes.hpp
kipawaa: I've been wondering that as well, but will leave it up to @maliasadi as part of the adjoint support work

mlir/lib/Quantum/Transforms/GraphDecomposition/DecompGraphSolver/DGTypes.hpp
kipawaa: Almost. If both are hashed with names then that also hits the first case here. This extra bit is only needed for cross-compatibility between ID and legacy hashes. We can return to the simple search once ID is the only pathway.

mlir/unittests/DecompGraphSolver/Test_DecompGraphCore.cpp
kipawaa: We shouldn't assume that we can extract any information from IDs (we should think of them as hashes, we may want to replace them with true hashes in the future) so I think it would be best to avoid this.

mlir/unittests/DecompGraphSolver/Test_DecompGraphCore.cpp
kipawaa: Inline with my comment above, we should consider them as hashes. If they match, they should be considered source of truth on which data is important for considering nodes to be identical.

mlir/lib/Quantum/Transforms/GraphDecomposition/DecompGraphSolver/DGTypes.hpp
kipawaa: Needed for `size_t` definition

paul0403 (approved):

mlir/lib/Quantum/Transforms/GraphDecomposition/DecompGraphSolver/DGTypes.hpp
maliasadi: let's get back to this in PR #3001

maliasadi (approved):

PR #3040, opened 2026-07-17
:partying_face: Decompose lowering pass uses DecomposableGate interface

paul0403:
**Context:**
Currently, the `decompose-lowering` pass only supports custom op, pauli rot and multirz. The three ops have almost identical patterns, only varying in very specific places.

With the new `DecomposableOp` interface, and all quantum dialect gate ops implemented under this interface (#3022), we can just implement the decompose lowering rewrite pattern on the interface, instead of on the individual ops!

**Benefits:**
De-duplicate code: easier to maintain, read, ship, etc.
More importantly, this is future-facing and easily extensible: any new op that wants to participate in graph decomposition (at least the `decompose-lowering` part) only needs to implement the interface, instead of adding their own pattern on the pass! 

**Related GitHub Issues:**
[sc-124932]


mlir/lib/Quantum/Transforms/GraphDecomposition/decompose_lowering.cpp
```diff
@@ -92,18 +92,7 @@ struct DecomposeLoweringPass : impl::DecomposeLoweringPassBase<DecomposeLowering
...
-                    // for MultiRZ, since it has multiple decomposition functions
-                    // based on the number of target qubits
-                    std::string newTargetOpStr =
-                        targetOp.str() + "_" + std::to_string(DecompUtils::getNumWires(func));
-
-                    decompositionRegistry[newTargetOpStr] = func;
-                }
-                else {
-                    decompositionRegistry[targetOp] = func;
-                }
```
paul0403, line 106: I don't know why multirz is special here, is something in the frontend appending the number of wires to the multirz rules?

kipawaa: I think this is because `numWires` has to be considered static data when it comes to `MultiRZ`'s decomposition rule

mlir/lib/Quantum/Transforms/GraphDecomposition/decompose_lowering.cpp
```diff
@@ -92,18 +92,7 @@ struct DecomposeLoweringPass : impl::DecomposeLoweringPassBase<DecomposeLowering
...
+                decompositionRegistry[targetOp] = func;
```
kipawaa: 🥳

mlir/lib/Quantum/Transforms/GraphDecomposition/DecomposeLoweringPatterns.cpp
```diff
@@ -158,161 +178,12 @@ struct DLCustomOpPattern : public OpRewritePattern<CustomOp> {
     }
 };
 
-struct DLMultiRZOpPattern : public OpRewritePattern<MultiRZOp> {
```
kipawaa, line 161: Yessssssss!!!

mlir/test/Quantum/DecomposeLoweringTest.mlir
kipawaa, line 1: Could you also add some tests using IDs? I don't think we current have any - I know it's the same mechanic but it would be good to check that it still works as expected and that the fallback mechanism works as well.

kipawaa (commented):
So happy to see this, thanks @paul0403! Just some test requests and we're good to go

mlir/lib/Quantum/Transforms/GraphDecomposition/decompose_lowering.cpp
paul0403: Makes sense, let me try to remove this special handling

mlir/test/Quantum/DecomposeLoweringTest.mlir
paul0403: https://github.com/PennyLaneAI/catalyst/pull/3040/commits/9ed7b4ba55444be8f354c342a9bae1b1e1ef5a75

kipawaa (approved):
So great! Thanks @paul0403!

mlir/test/Quantum/DecomposeLoweringTest.mlir
```diff
@@ -598,6 +598,37 @@ module @circuit_with_multirz {
...
+    %0 = stablehlo.slice %arg2 [0:1] : (tensor<1xi64>) -> tensor<1xi64>
+    %1 = stablehlo.reshape %0 : (tensor<1xi64>) -> tensor<i64>
+    %extracted = tensor.extract %1[] : tensor<i64>
+    %2 = quantum.extract %arg0[%extracted] : !quantum.reg -> !quantum.bit
+    %c0 = arith.constant 0 : index
+    %extracted_0 = tensor.extract %arg1[%c0] : tensor<1xf64>
+    %out_qubits = quantum.custom "RZ"(%extracted_0) %2 : !quantum.bit
+    %extracted_1 = tensor.extract %1[] : tensor<i64>
+    %3 = quantum.insert %arg0[%extracted_1], %out_qubits : !quantum.reg, !quantum.bit
+    return %3 : !quantum.reg
```
maliasadi, line 626:
```suggestion
    return %arg0 : !quantum.reg
```

maliasadi (approved):
Nice 🤩

mlir/test/Quantum/DecomposeLoweringTest.mlir
paul0403: Hmm, I wanted to check RZ is indeed in the new circuit, but I guess just check-not-ing the original operator op is fine?

PR #3043, opened 2026-07-20
[Experimental Backline] Define runtime transport layer interface

josephleekl:
**Context:**
Add a runtime transport layer interface for remote executors. This will be implemented by backend devices for a backline/heterogeneous compilation device.

**Description of the Change:**
A new transport interface `Transport.hpp` is added into runtime include directory.

**Benefits:**
A uniform interface for backend transport executor device implementation.

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-123779]

PR #3045, opened 2026-07-20
[Experimental Backline] Transport Layer Loader is added

rniczh:
**Context:**

Adds a backend-agnostic transport loader to the runtime:

`TransportCAPI.h/.cpp`: The ABI the compiler/codegen can emit against.
`TransportBackend.h`: the plugin factory contract; backends are separate `.so` the runtime dlopens at session create, selected via backend_lib.

`make all -DENABLE_TRANSPORT=ON` (default OFF) to build

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-123779]


runtime/include/Transport.hpp
```diff
@@ -136,22 +136,42 @@ class TransportSession {
      * @brief Wait for a result and write it out.
      *
      * @param outputs Array of output buffers to write into.
+     * @param output_bytes Array of output buffer sizes.
      * @param n Number of output buffers.
      *
      * @return `int`
      */
-    virtual int collect(void *const *outputs, std::size_t n) = 0;
+    virtual int collect(void *const *outputs, const std::size_t *output_bytes, std::size_t n) = 0;
```
josephleekl: should we use std::uint64_t maybe? for uniformity it doesn't really matter

runtime/include/Transport.hpp
```diff
@@ -136,22 +136,42 @@ class TransportSession {
...
+    // Build the work item in slot `work_item_idx` from `schema`.
+    virtual void commit_work_item(std::uint32_t work_item_idx, std::uint64_t in_bytes,
+                                  std::uint64_t out_bytes) = 0;
+
+    // Fire one round using work item `work_item_idx` and whatever payload is currently in
+    // data_slot(). Pairs with a subsequent collect(). Returns 0 on success.
+    virtual int kick(std::uint32_t work_item_idx = 0) = 0;
+
+    // Current round's outbound slot in the transport-owned ring.
+    virtual void *data_slot() = 0;
```
josephleekl, line 164: should this return something?

runtime/include/TransportBackend.h
```diff
@@ -0,0 +1,55 @@
...
+#pragma once
+#ifndef TRANSPORTBACKEND_H
+#define TRANSPORTBACKEND_H
+
+#include <cstdio>
+#include <exception>
+
+#include "Transport.hpp"
+
+#define CATALYST_TRANSPORT_CONTROLLER_FACTORY_SYMBOL "CatalystTransportControllerFactory"
```
josephleekl, line 29: do we need one for coprocessor too? or can they be united?

runtime/include/TransportCAPI.h
```diff
@@ -0,0 +1,107 @@
...
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// Opaque transport session handle
+typedef struct CatalystTransportSession CatalystTransportSession;
+
+// Return codes: 0 == success; negative == error
```
josephleekl, line 34: `negative == error`
is this typical? should we not just have errors as positive?

runtime/include/TransportCAPI.h
```diff
@@ -0,0 +1,107 @@
...
+
+// Create a controller session from a named backend plugin `.so` (dlopen'd by the runtime).
+// `config` is the backend's "key=value;..." string. Returns NULL on failure.
+CatalystTransportSession *__catalyst__transport__controller_create(const char *backend_lib,
+                                                                   const char *config);
+
+void __catalyst__transport__close(CatalystTransportSession *s);
+int __catalyst__transport__connect(CatalystTransportSession *s, const char *peer,
+                                   uint16_t oob_port);
+int __catalyst__transport__alloc_reply(CatalystTransportSession *s, uint64_t size, int32_t mem_kind,
```
josephleekl: what is alloc_reply? should we add that to transport.hpp?

runtime/include/TransportCAPI.h
```diff
@@ -0,0 +1,107 @@
...
+                                       uint32_t access, CatalystTransportMemRegion *out);
+
+int __catalyst__transport__exchange_keys(CatalystTransportSession *s,
+                                         CatalystTransportPeerRef *out);
+int __catalyst__transport__establish_channel(CatalystTransportSession *s, int32_t data_path,
+                                             const CatalystTransportPeerRef *peer);
+int __catalyst__transport__commit_work_item(CatalystTransportSession *s, uint32_t work_item_idx,
+                                            uint64_t in_bytes, uint64_t out_bytes);
+void *__catalyst__transport__data_slot(CatalystTransportSession *s);
```
josephleekl, line 78: what is this?

josephleekl (commented):
just some quick comments, looks good! tag me again when addressed. thanks @rniczh !

runtime/include/Transport.hpp
rniczh: Yes, `(void *)` and where it points to is based on the implementation. For example, in ring mode, you need to make sure the data_slot always pointing to the location where you are going to place.

runtime/include/TransportBackend.h
rniczh: Actually I don't know how the coprocessor works in this design, do you have any idea of this?

runtime/include/TransportCAPI.h
rniczh: I think it's standard for linux abi?

runtime/include/Transport.hpp
josephleekl: oh i see, i thought this would be an index. that's ok

runtime/include/TransportBackend.h
josephleekl: the coprocessor is just the same as the controller, it will also have a .so

runtime/include/TransportCAPI.h
rniczh: oh, I forgot to remove this redundancy

runtime/include/TransportCAPI.h
rniczh: It's for return the data slot for cpu to place the data

runtime/include/TransportBackend.h
rniczh: okay, let me add it

rniczh: Done https://github.com/PennyLaneAI/catalyst/pull/3045/commits/cdee60bc9d1099eeb026c0b27b1eb199b65bcf62

runtime/include/Transport.hpp
rniczh: I updated it in your PR

runtime/include/TransportCAPI.h
rniczh: Done https://github.com/PennyLaneAI/catalyst/pull/3045/commits/6e942a9fc3c257fafb5214c24efbc82067e56575 https://github.com/PennyLaneAI/catalyst/pull/3045/commits/4d211cefd98b20d3eb441c6211d41074f4cea909

runtime/include/TransportCAPI.h
mehrdad2m, line 1: what about `start()`?  see https://github.com/PennyLaneAI/catalyst/pull/3043/changes#diff-01e7d52b33c37ab016a8280744ff88e997d75b4aa2bcaa3d051c921c95742e3dR133

rniczh: Good catch, I forgot this [`4408322` (this PR)](https://github.com/PennyLaneAI/catalyst/pull/3045/commits/4408322703086dfd95173ec82f4e960281ef86f9)

rniczh: But on my side, the implementation of start is empty now, would be useful in the future if we need something to setup like warmup

runtime/include/TransportCAPI.h
```diff
@@ -0,0 +1,107 @@
...
+enum {
+    CATALYST_TRANSPORT_MEM_CPU_RAM = 0,
+    CATALYST_TRANSPORT_MEM_GPU_HBM = 1,
+    CATALYST_TRANSPORT_MEM_DDR = 2,
+    CATALYST_TRANSPORT_MEM_OTHER = 3,
+};
+
+// ibverbs access flags for the advertised reply region
+enum {
+    CATALYST_TRANSPORT_ACCESS_REPLY = 7,
```
mehrdad2m: what is this? :)

runtime/lib/transport/TransportCAPI.cpp
```diff
@@ -0,0 +1,281 @@
...
+
+int __catalyst__transport__kick(CatalystTransportSession *s, std::uint32_t work_item_idx)
+{
+    if (!s || !s->sess) {
+        return CATALYST_TRANSPORT_ERR;
+    }
+    return guard([&] { return s->sess->kick(work_item_idx); });
+}
+
+int __catalyst__transport__collect(CatalystTransportSession *s, void *correction,
```
mehrdad2m: the signature here seems a bit different than the frontend UI, is that intentional? see https://github.com/PennyLaneAI/catalyst/pull/3043/changes#diff-01e7d52b33c37ab016a8280744ff88e997d75b4aa2bcaa3d051c921c95742e3dR144

rniczh: Yes, the current API only supports 1 correction data

```
int __catalyst__transport__collect(CatalystTransportSession *s, void *correction,
                                   std::uint64_t bytes)
{
    if (!s || !s->sess) {
        return CATALYST_TRANSPORT_ERR;
    }
    return guard([&] {
        void *outputs[1] = {correction};
        std::size_t caps[1] = {static_cast<std::size_t>(bytes)};
        return s->sess->collect(outputs, caps, 1);
    });
}
```

runtime/include/TransportCAPI.h
rniczh: Redundant code, now the allocation was handled by the implementation: https://github.com/PennyLaneAI/catalyst/pull/3045/commits/060458b4319033542fdea882a3bb3ec07177732b

runtime/lib/transport/TransportCAPI.cpp
rniczh: I think it's also a design should be changed (as in offline discussion with @josephleekl ), the number of collection should be always 1, I'll update this later

rniczh: https://github.com/PennyLaneAI/catalyst/pull/3045/commits/246a4b7b2e3df32296f85c87c72f1c5a7f0e2ae1

runtime/lib/transport/TransportCAPI.cpp
```diff
@@ -0,0 +1,251 @@
...
+    }
+    return guard([&] { return s->sess->collect(correction, bytes); });
+}
+
+std::uint64_t __catalyst__transport__last_rtt_ns(CatalystTransportSession *s)
+{
+    if (!s || !s->sess) {
+        return 0;
+    }
+    return s->sess->last_rtt_ns();
```
josephleekl: do we need to wrap this in a try catch?

runtime/include/TransportCAPI.h
```diff
@@ -0,0 +1,91 @@
...
+                                             const CatalystTransportPeerRef *peer);
+int __catalyst__transport__commit_work_item(CatalystTransportSession *s, uint32_t work_item_idx,
+                                            uint64_t in_bytes, uint64_t out_bytes);
+void *__catalyst__transport__data_slot(CatalystTransportSession *s);
+void __catalyst__transport__start(CatalystTransportSession *s);
+int __catalyst__transport__kick(CatalystTransportSession *s, uint32_t work_item_idx);
+int __catalyst__transport__collect(CatalystTransportSession *s, void *correction, uint64_t bytes);
+uint64_t __catalyst__transport__last_rtt_ns(CatalystTransportSession *s);
+void __catalyst__transport__stop(CatalystTransportSession *s);
+void __catalyst__transport__destroy(CatalystTransportSession *s);
```
josephleekl, line 86: do we need to worry about people calling `close` and `destroy` which may double free?

runtime/include/TransportCAPI.h
```diff
@@ -0,0 +1,91 @@
...
+// Remote peer region descriptor
+typedef struct {
+    uint32_t rkey;
+    uint64_t remote_addr;
+    uint64_t size;
+} CatalystTransportPeerRef;
+
+// Create a controller session from a named backend plugin `.so` (dlopen'd by the runtime).
+// `config` is the backend's "key=value;..." string. Returns NULL on failure.
+CatalystTransportSession *__catalyst__transport__controller_create(const char *backend_lib,
```
josephleekl: can we add the coprocessor one too?

josephleekl: or i can add the coprocessor ones in a follow-up PR with the CPU migration

mehrdad2m: I am working on this in my local branch. there would be some changes pushed to this branch. Will let you know once it's done

runtime/include/TransportCAPI.h
mehrdad2m: I'd like to remove close as it is simply a combination of destroy and stop

runtime/include/Transport.hpp
```diff
@@ -133,14 +123,16 @@ class TransportSession {
     virtual void start() = 0;
 
     /**
-     * @brief Wait for a result and write it out.
+     * @brief Wait for a result and scatter it into the reply buffers.
```
josephleekl, line 126: do we actually do any scattering?

just wondering why do we need this expanded collect method? is this how we do the new result collection/has this been used in the device controller that you're implementing @rniczh ?

runtime/lib/transport/TransportCAPI.cpp
```diff
@@ -0,0 +1,372 @@
...
+// Provision the local reply region on first use (idempotent).
+void ensure_reply(CatalystTransportSession *s)
+{
+    if (!s->reply_ready) {
+        s->reply = s->sess->alloc_memory(kReplyBytes, MemKind::CpuRam);
+        s->reply_ready = true;
+    }
+}
+
+ControllerSession *as_controller(CatalystTransportSession *s)
```
josephleekl: can we call them `async_` instead of `as_`?

runtime/tests/Test_Transport.cpp
```diff
@@ -0,0 +1,140 @@
...
+#include <cstdint>
+
+#include "catch2/catch_test_macros.hpp"
+
+#include "TransportCAPI.h"
+
+namespace {
+constexpr std::int32_t kController = CATALYST_TRANSPORT_ROLE_CONTROLLER;
+constexpr std::int32_t kCoprocessor = CATALYST_TRANSPORT_ROLE_COPROCESSOR;
+constexpr const char *kStub = STUB_BACKEND_PATH;
```
josephleekl: can we just remove these?

runtime/lib/transport/TransportCAPI.cpp
```diff
@@ -0,0 +1,372 @@
...
+    auto it = g_registry.find(registry_key(role, key));
+    if (it == g_registry.end()) {
+        std::cerr << "[transport] get_session: no session registered for role " << role << " key '"
+                  << (key ? key : "") << "'\n";
+        return nullptr;
+    }
+    return it->second;
+}
+
+void __catalyst__transport__destroy(CatalystTransportSession *s)
```
josephleekl, line 417: should we drain the pending tokens here too?

runtime/lib/transport/TransportCAPI.cpp
```diff
@@ -0,0 +1,372 @@
...
+    return guard([&] {
+        auto *co = dynamic_cast<CoprocessorSession *>(s->sess);
+        if (!co) {
+            std::cerr << "[transport] set_coprocessor_fn on a non-coprocessor session\n";
+            return CATALYST_TRANSPORT_ERR;
+        }
+        // Empty symbol selects the built-in echo; a named-but-unresolved symbol is a hard error
+        CoprocessorFn fn = &echo_fn;
+        if (symbol && *symbol) {
+            fn = reinterpret_cast<CoprocessorFn>(dlsym(RTLD_DEFAULT, symbol));
```
josephleekl: JB: Coprocessor symbol resolution scope. set_coprocessor_fn resolves the user symbol via dlsym(RTLD_DEFAULT, symbol). The backend plugin is loaded with RTLD_LAZY | RTLD_NODELETE (no RTLD_GLOBAL), so symbols defined inside the plugin .so are not visible to RTLD_DEFAULT. This works only if the coprocessor symbol lives in the main compiled image. If a plugin is ever expected to provide it, resolution will silently fail. Worth confirming the intended location and documenting it.

runtime/lib/transport/TransportCAPI.cpp
```diff
@@ -0,0 +1,372 @@
...
+// brought up in one function can be resolved in another.
+std::unordered_map<std::string, CatalystTransportSession *> g_registry;
+
+std::string registry_key(std::int32_t role, const char *key)
+{
+    return std::to_string(role) + "/" + (key ? key : "");
+}
+
+// Local reply region size, provisioned automatically at exchange_keys.
+constexpr std::uint64_t kReplyBytes = 16 * 1024;
```
josephleekl, line 73: does catalyst use this style `k....` for constants, or all caps?
also is this size always sufficient? should it be configurable?

runtime/lib/transport/TransportCAPI.cpp
```diff
@@ -0,0 +1,372 @@
...
+                CATALYST_TRANSPORT_CONTROLLER_FACTORY_SYMBOL);
+            h->sess = factory(cfg);
+        }
+        if (!h->sess) {
+            std::cerr << "[transport] backend factory returned null for config: " << cfg << "\n";
+            return nullptr;
+        }
+        auto *raw = h.release();
+        if (key && *key) {
+            g_registry[registry_key(role, key)] = raw; // resolved later via get_session
```
josephleekl: JB: Overwrite-on-create can leak. In create, when key already exists, `g_registry[...] = raw` overwrites the prior pointer without destroying it. The test documents that the caller must still destroy the old handle — but if the caller has lost that handle, the session leaks. A warning log on overwrite would help catch misuse.

runtime/include/TransportBackend.h
```diff
@@ -0,0 +1,76 @@
...
+// e.g. GENERATE_TRANSPORT_CONTROLLER_FACTORY(CatalystTransportController, make_controller)
+// where `make_controller(const std::string &config) -> ControllerSession*`.
+#define GENERATE_TRANSPORT_CONTROLLER_FACTORY(IDENTIFIER, BUILDER)                                 \
+    extern "C" catalyst::transport::ControllerSession *IDENTIFIER##Factory(const char *config)     \
+    {                                                                                              \
+        try {                                                                                      \
+            return (BUILDER)(config ? std::string(config) : std::string());                        \
+        }                                                                                          \
+        catch (const std::exception &e) {                                                          \
+            std::fprintf(stderr, "[transport] controller factory failed: %s\n", e.what());         \
```
josephleekl: other places use `std::cerr`? should we be consistent?

josephleekl (approved):
thanks @rniczh ! Nothing major, just some minor comments

runtime/include/Transport.hpp
rniczh: I think this was updated by @mehrdad2m

runtime/lib/transport/TransportCAPI.cpp
rniczh: Yes. https://github.com/PennyLaneAI/catalyst/pull/3045/commits/68149c3093239cfaf9b4df82dbfea11b098ba470

runtime/include/TransportCAPI.h
rniczh: I think we have removed `close`, not sure where we did this. There is no `close` here.

runtime/include/Transport.hpp
rniczh: Just checked the code, I think the behaviour is exactly the same as the scatter? you need to collect the data in bytes.

runtime/lib/transport/TransportCAPI.cpp
rniczh: I think there is no `async` meaning in this function? @mehrdad2m

runtime/tests/Test_Transport.cpp
rniczh: @mehrdad2m

mehrdad2m: Thanks, will get to these later today

runtime/lib/transport/TransportCAPI.cpp
rniczh: Yes, we should this! Fixed: https://github.com/PennyLaneAI/catalyst/pull/3045/commits/bbebcd2ee277e60d2250fcdb14ed3319ee398dde
https://github.com/PennyLaneAI/catalyst/pull/3045/commits/f59cd15c5cd2d8a8726166e12e6d9e8a687b167e

runtime/lib/transport/TransportCAPI.cpp
rniczh: Fixed. dlysm to resolve the symbol in plugin first. https://github.com/PennyLaneAI/catalyst/pull/3045/commits/7d5c4248f40d22a86c25122bd2bedeedae05cb47

And it wouldn't sliently fail actually, it will be captured by CATALYST_TRANSPORT_ERR

runtime/lib/transport/TransportCAPI.cpp
rniczh: `g_registry[key] = raw` just replace the lookup pointer, it wouldn't destroy the old session, just need to make ensure this get destroyed by the caller. I added a warning here. https://github.com/PennyLaneAI/catalyst/pull/3045/commits/793d8bfa50e2d65f25049fb25ce418b84a588b00 (cc. @mehrdad2m)

e.g. TestTransport
```cpp
TEST_CASE("re-create under the same key overwrites", "[transport]")
{
    auto *s1 = make(kController, "ovr_key");
    auto *s2 = make(kController, "ovr_key");
    REQUIRE(s1 != s2);
    CHECK(__catalyst__transport__get_session(kController, "ovr_key") == s2);
    __catalyst__transport__destroy(s1);
    // s1 was already overwritten, so s2 remains resolvable until its own destroy.
    CHECK(__catalyst__transport__get_session(kController, "ovr_key") == s2);
    __catalyst__transport__destroy(s2);
    CHECK(__catalyst__transport__get_session(kController, "ovr_key") == nullptr);
}
```

runtime/lib/transport/TransportCAPI.cpp
rniczh: I think catalyst isn't strictly one-style (e.g. `runtime/tests/Test_SharedLibraryManager.cpp`) you'll see the kCamelStyle keywords in this file as well. But I agree we should make it configurable, we could expose a api to support this? how do you think? `__catalyst__transport__set_reply_bytes(s, n)` (and 16KB as default if we don't set the region size)

runtime/include/TransportBackend.h
rniczh: Yes, this make sense. updated: https://github.com/PennyLaneAI/catalyst/pull/3045/commits/5063573b7b3069441d0c3adc47e315514b530db2

runtime/lib/transport/TransportCAPI.cpp
mehrdad2m: no this is not meant to be short for `async` this is a cast from `CatalystTransportSession->sess` to a `ControllerSession`. Let me know id you want to rename it.

runtime/lib/transport/TransportCAPI.cpp
josephleekl: maybe let's add a `TODO` to make it configurable later

runtime/lib/transport/TransportCAPI.cpp
josephleekl: yeah let's rename this to cast or sth, i was confused 😓

runtime/include/TransportCAPI.h
josephleekl: @mehrdad2m you'll deal with this?

runtime/lib/transport/TransportCAPI.cpp
rniczh: added: https://github.com/PennyLaneAI/catalyst/pull/3045/commits/392cbddca3f3a99f15657aec318e174fe0216492

runtime/lib/transport/TransportCAPI.cpp
mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/3045/commits/4c635d7d63aea7e271bfac9cc17d8bbfdc3de2b3

runtime/tests/Test_Transport.cpp
mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/3045/commits/b0e8b2284f891dff7295295c837e4b8dd79d61c9

runtime/include/TransportCAPI.h
mehrdad2m: This is already fixed

mehrdad2m (approved):

PR #3047, opened 2026-07-21
Add transport dialect definitions

mehrdad2m:
**Context:**
The runtime transport layer exposes a CAPI for driving a remote session over the data plane. However in order to have fine grained control over the CAPI through frontend, we need to be able to express them in MLIR.

Adds the transport MLIR:
- Two opaque handle types: `!transport.session<role>` (controller/coprocessor) and `!transport.token`
- Bring-up ops: `create`, `connect`, `exchange_keys`, `establish_channel`, `commit_work_item`, `start` (async connect_async/exchange_keys_async joined by barrier; `set_coprocessor_fn` for the coprocessor).
- Per-round ops: `kick` (buffer-in), `collect` (buffer-out), `last_rtt_ns`.
- Teardown ops: `stop`, `destroy`.
- Cross-function resolution: `get_session`.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


mlir/include/Transport/IR/TransportDialect.td
```diff
@@ -0,0 +1,104 @@
...
+include "mlir/IR/AttrTypeBase.td"
+include "mlir/IR/EnumAttr.td"
+include "mlir/Interfaces/SideEffectInterfaces.td"
+
+//===----------------------------------------------------------------------===//
+// Transport dialect definition.
+//===----------------------------------------------------------------------===//
+
+def Transport_Dialect : Dialect {
+    let summary = "Typed ops for setting up and driving a transport session.";
```
josephleekl, line 29: love this summary

mlir/tools/quantum-opt/quantum-opt.cpp
```diff
@@ -77,6 +79,7 @@ int main(int argc, char **argv)
     registry.insert<catalyst::pauli_frame::PauliFrameDialect>();
     registry.insert<catalyst::ion::IonDialect>();
     registry.insert<catalyst::rtio::RTIODialect>();
+    registry.insert<catalyst::transport::TransportDialect>();
```
josephleekl, line 83: should we update this as well? https://github.com/PennyLaneAI/catalyst/blob/main/mlir/tools/quantum-lsp-server/quantum-lsp-server.cpp

mlir/include/Transport/IR/TransportDialect.td
```diff
@@ -0,0 +1,104 @@
...
+    CPred<"::llvm::isa<::catalyst::transport::SessionType>($_self) && "
+          "::llvm::cast<::catalyst::transport::SessionType>($_self).getRole() == "
+          "::catalyst::transport::Role::Coprocessor">,
+    "coprocessor transport session">;
+
+//===----------------------------------------------------------------------===//
+// Operation base.
+//===----------------------------------------------------------------------===//
+
+// All transport ops perform side-effecting I/O
```
josephleekl, line 93: is this true even for e.g. `last_rtt_ns`? or do we not care about that

mlir/include/Transport/IR/TransportOps.td
```diff
@@ -0,0 +1,156 @@
...
+def Transport_StartOp : Transport_Op<"start"> {
+    let summary = "Start the session (non-blocking; runs until stop()).";
+    let arguments = (ins Transport_SessionType:$session);
+    let assemblyFormat = "$session attr-dict `:` type($session)";
+}
+
+def Transport_CollectOp : Transport_Op<"collect"> {
+    let summary = "Wait for this round's reply and return it as a value.";
+    let arguments = (ins Transport_SessionType:$session, I64Attr:$bytes);
+    let results = (outs I64:$result);
```
josephleekl: not really an issue, but is I64 good enough for what we need?

josephleekl (approved):
thanks @mehrdad2m. Nothing major to add. Just:
- do we need to update the changelog?
- Is `!transport.peer` actually defined here? (mentioned in the changelog)

mlir/include/Transport/IR/TransportDialect.td
mehrdad2m: Yeah I think `last_rtt_ns` could be a read op but I am trying to be conservative here. We can probably update operations once the design is set in place

mlir/include/Transport/IR/TransportOps.td
mehrdad2m: no will update that. This was just for testing. I think this should be a memref to handle real cases.

mlir/tools/quantum-opt/quantum-opt.cpp
mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/3047/commits/e4a779b1ed1253a3402fb0f193000e6a78ab92e4

mlir/include/Transport/IR/TransportOps.td
rniczh: @josephleekl  in which case that i64 would not enough?

mehrdad2m:
> is !transport.peer actually defined here? (mentioned in the changelog)

sorry that is removed. I'll update the PR description.

> do we need to update the changelog?

yes, will add shortly

mlir/include/Transport/IR/TransportDialect.td
```diff
@@ -0,0 +1,104 @@
...
+    let cppNamespace = "::catalyst::transport";
+}
+
+def Transport_DataPath : I32EnumAttr<"DataPath", "transport data-movement path", [
+        I32EnumAttrCase<"CpuVerbs", 0, "cpu_verbs">,
+        I32EnumAttrCase<"GpuEngine", 1, "gpu_engine">,
+        I32EnumAttrCase<"Other", 2, "other">
+    ]> {
+    let cppNamespace = "::catalyst::transport";
+}
```
rniczh: no action needed here: just out of curious, would string attribute (StringAttr) make it more easier to maintain compared to the closed set enum attr here?

mlir/include/Transport/IR/TransportDialect.h
```diff
@@ -0,0 +1,25 @@
+// Copyright 2026 Xanadu Quantum Technologies Inc.
```
rniczh, line 1: Could you add a mlir lit test for verifying the IR?

rniczh (commented):
Thanks @mehrdad2m !

mlir/include/Transport/IR/TransportDialect.td
mehrdad2m: Yeah that's fair. I think it's a tradeoff, but extending a StringAttr would be easier later on. I'll try to change this now rather than later as would ripple down some changes down to runtime.

mlir/include/Transport/IR/TransportOps.td
josephleekl: we haven't really decided what `collect` actually collects, but e.g. could it be a floating point?

rniczh (approved):
Awesome! @mehrdad2m

mlir/include/Transport/IR/TransportOps.td
mehrdad2m: I have changed this to be a tensor type to be general, we could later modify if needed.

PR #3048, opened 2026-07-21
Add the convert-transport-to-llvm pass

mehrdad2m:
**Context:**
Follow-up to ["Add the transport dialect"](https://github.com/PennyLaneAI/catalyst/pull/3047) which added the transport dialect. This PR connects it to the runtime by lowering each op to its __catalyst__transport__* CAPI call.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


mlir/lib/Transport/Transforms/TransportToLLVM.cpp
```diff
@@ -0,0 +1,345 @@
...
+IntegerType i64Ty(MLIRContext *ctx) { return IntegerType::get(ctx, 64); }
+
+ModuleOp moduleOf(Operation *op) { return op->getParentOfType<ModuleOp>(); }
+
+Value emitCall(ConversionPatternRewriter &rewriter, Location loc, ModuleOp mod, StringRef name,
+               ArrayRef<Type> paramTys, Type resultTy, ValueRange args)
+{
+    Type rty = resultTy ? resultTy : LLVM::LLVMVoidType::get(rewriter.getContext());
+    auto fn = LLVM::lookupOrCreateFn(rewriter, mod, name, paramTys, rty);
+    assert(succeeded(fn) && "failed to declare transport CAPI function");
```
josephleekl, line 51: will this `assert` be compiled in release build? would it be better to do e.g. `if (failed(fn)) return {};`

mlir/lib/Transport/Transforms/TransportToLLVM.cpp
```diff
@@ -0,0 +1,345 @@
...
+    auto call = LLVM::CallOp::create(rewriter, loc, *fn, args);
+    return call.getNumResults() ? call.getResult() : Value();
+}
+
+Value globalStr(ConversionPatternRewriter &rewriter, Location loc, ModuleOp mod, StringRef prefix,
+                StringRef value)
+{
+    static int counter = 0;
```
josephleekl: Jukebox is not happy with this 😆 : 

Convention divergence / thread-safety 

`globalStr` uses a function-local `static int counter` to name globals. Three problems: The sibling `ExecutorToLLVM.cpp` already solves this correctly: it derives a content-based key (e.g. `"executor_path_" + tag, addrGlobalKey(...)`) and calls `mod.lookupSymbol<LLVM::GlobalOp>(key)` to reuse an existing global. Please follow that established pattern rather than a static counter.
- It's a data race: MLIR runs module-level passes concurrently across a context's thread pool, so `counter++` is unsynchronized shared mutable state (UB).
- Symbol names become non-deterministic across the process (invocation-order dependent), which hurts reproducible IR/test stability.
- Every op emits a brand-new global even for identical strings — no de-duplication.

josephleekl (approved):
Nice one, thanks @mehrdad2m . Just some small comments

mlir/lib/Transport/Transforms/TransportToLLVM.cpp
mehrdad2m: Not by default but there is an option called `LLVM_ENABLE_ASSERTIONS` which enables them even if you build in realese mode, and we in fact use that option when building wheels, so no concern here.

mlir/lib/Transport/Transforms/TransportToLLVM.cpp
mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/3048/commits/55ec769962fe1f90755a44b289c7e331990dc73c

mlir/lib/Transport/Transforms/TransportToLLVM.cpp
```diff
@@ -0,0 +1,374 @@
...
+    auto call = LLVM::CallOp::create(rewriter, loc, *fn, args);
+    return call.getNumResults() ? call.getResult() : Value();
+}
+
+std::string globalStrKey(StringRef prefix, StringRef value)
+{
+    std::string key = prefix.str();
+    for (char c : value) {
+        bool ok =
+            (c >= '0' && c <= '9') || (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || c == '_';
```
multiphaseCFD:
```suggestion
            llvm::isAsciiIdentifierContinue(c);
```

rniczh (approved):
Thanks @mehrdad2m

multiphaseCFD (approved):
Thank you @mehrdad2m ! Nice one

mlir/lib/Transport/Transforms/TransportToLLVM.cpp
mehrdad2m: seems like `isAsciiIdentifierContinue` is only accessable from clang, I replaced it with `llvm::isAlnum(c) || c == '_'`

PR #3049, opened 2026-07-21
Fetch all reacheable decomp rules during frontend lowering

paul0403:
**Context:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-122036]

### DIFF
--- a/frontend/catalyst/device/python_decompositions.py
+++ b/frontend/catalyst/device/python_decompositions.py
@@ -65,7 +65,7 @@ def get_dummy_values_for_container(container):
     return tuple(dummy_args)
 
 
-def get_graph_op_id(op: qp.decomposition.CompressedResourceOp | qp.core.Operator2):
+class GraphOpID:
     """
     Return the graph operator id for the operator2 instance `op`.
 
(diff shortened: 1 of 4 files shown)


frontend/catalyst/device/python_decompositions.py
```diff
@@ -85,36 +88,88 @@ def get_graph_op_id(op: qp.decomposition.CompressedResourceOp | qp.core.Operator
...
+        resources = rule.compute_resources(**static_data)
+        name_to_resources[rule.name] = resources.gate_counts
+        name_to_resource_ids[rule.name] = {
+            GraphOpID(op).getID(): count for op, count in resources.gate_counts.items()
+        }
+
+    return name_to_resources, name_to_resource_ids, decomp_rules
+
+
+def BFS_decomp_rules(op_name, static_data):
```
albi3ro, line 158: What does BFS mean? 

I know it's in draft mode still, but just trying to get an idea of how things fit together.

frontend/catalyst/device/type_stringify_utils.py
```diff
@@ -0,0 +1,43 @@
...
+# limitations under the License.
+
+import pennylane as qp
+
+from catalyst.utils.exceptions import CompileError
+
+
+def _py_dtype_to_mlir_type_string(python_dtype: type):
+    match python_dtype:
+        case qp.numpy.float64:
```
albi3ro, line 22: Any idea how pennylane.numpy is creaping in? We've been trying to be careful to not mix up autograd and catalyst, so would be great if we could fix this upstream.

frontend/catalyst/device/python_decompositions.py
kipawaa: Breadth First Search for graph traversal. The idea is to see which operators might appear in the known decompositions of the ops in the circuit, and to lower decomposition rules for them to maximize the chances that the rules we need are already in the circuit

frontend/catalyst/device/type_stringify_utils.py
paul0403: I need to convert python data types to a ID string used in the IR. 

I just thought it's better to use `qp.*` instead of `np.*` (or `jnp.*`). Would it be better to just use the numpy types?

paul0403: For an op like 
```python
from pennylane.typing import Wire, Float

class Bob(qp.core.Operator2):

    arg_specs = {'wires': Wire[1], "other_wires": Wire[2], "phi": Float, "thetas": Float[2]}
    dynamic_argnames = ("phi","thetas")
    wire_argnames = ("wires", "other_wires")
    compilable_argnames = ("bob_word",)

    def __init__(self, wires, other_wires, phi, thetas, bob_word):
        super().__init__(wires=wires, other_wires=other_wires, phi=phi, thetas=thetas, bob_word=bob_word)
```

The dynamic param types (for `phi` and `thetas`) are 
```
[AbstractArray((), float64, weak_type=True), AbstractArray((2,), float64, weak_type=True)]
```
where the individual types are `<class 'pennylane.typing.AbstractArray'>`

PR #3035, opened 2026-07-16
Fix LLL failures 17 July

JerryChen97:
**Context:**
Time traveler from the future...

https://github.com/PennyLaneAI/catalyst/actions/runs/29522362925

**Description of the Change:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

PR #3050, opened 2026-07-22
Migrate specs tests to PennyLane #9841 resource API

Qottmann:
## Summary

PennyLane **dev62** is the minimum pin that includes both [PennyLaneAI/pennylane#9846](https://github.com/PennyLaneAI/pennylane/pull/9846) (`magic-T`) and [PennyLaneAI/pennylane#9841](https://github.com/PennyLaneAI/pennylane/pull/9841) (resource API). This PR bumps Catalyst to that pin and fixes the resulting fallout.

**Resource API migration (#9841)**
- Bump PennyLane to `0.46.0.dev62` in `.dep-versions` and `doc/requirements.txt`
- Replace `SymbolicSpecsResources` checks with `SpecsResources.is_symbolic`
- Rename resource fields in tests: `gate_types`/`gate_counts` → `quantum_operations`, `measurements` → `measurement_processes`, `num_gates` → `total_quantum_operations`
- Drop `gate_sizes` assertions (no longer tracked)
- Handle `PBCSpecsResources` in PPR specs tests (depth fields)

**dev62 fallout (Operator2 / magic-T)**
- `decompose.py`: safe `.params` access for Operator2 ops; skip abstract `Adjoint(...)` / `Controlled(...)` decomposition-graph nodes that lack `op_type`
- PLXPR tests: single-qubit ops like `S`/`T` now lower to `qref_operator` instead of `qref_qinst`
- `test_adjoint.py`: replace removed `qp.Rotation` with `qp.Rot`
- `test_quantum_control.py`: compare `ControlledOp2` decompositions with `qp.equal`
- `test_specs.py`: normalize doubly-controlled `C(S)` counts in `test_complex`

This is the base branch for the magic-state allocation stack (`#3029` → `#3027`).

### DIFF
--- a/frontend/catalyst/from_plxpr/decompose.py
+++ b/frontend/catalyst/from_plxpr/decompose.py
@@ -195,7 +195,8 @@ class DecompRuleInterpreter(qp.capture.PlxprInterpreter):
         # and compile them to Catalyst JAXPR decomposition rules
         for op, rule in self._decomp_graph_solution.items():
             # Get number of wires if exists
-            op_num_wires = op.op.params.get("num_wires", None)
+            op_params = getattr(op.op, "params", {}) or {}
+            op_num_wires = op_params.get("num_wires", None)
             if (
                 o := next(
                     (
@@ -222,7 +223,7 @@ class DecompRuleInterpreter(qp.capture.PlxprInterpreter):
                 # In this case, we fall back to using the COMPILER_OPS_FOR_DECOMPOSITION
                 # dictionary to get the number of wires.
                 num_wires, num_params = COMPILER_OPS_FOR_DECOMPOSITION[op.op.name]
-                pauli_word = op.op.params.get("pauli_word", None)
+                pauli_word = op_params.get("pauli_word", None)
                 requires_copy = num_wires == -1
 
                 if op.op.name in ("PauliRot", "PauliMeasure"):
@@ -237,24 +238,14 @@ class DecompRuleInterpreter(qp.capture.PlxprInterpreter):
                     requires_copy=requires_copy,
                     pauli_word=pauli_word,
                 )
-            elif not (
-                (op_type := getattr(op.op, "op_type", None)) is not None
-                and issubclass(
-                    op_type,
-                    (
-                        qp.ops.Adjoint,
-                        qp.ops.Controlled,
-                        qp.ops.ChangeOpBasis,
-                        qp.ops.Prod,
-                        qp.TemporaryAND,
-                    ),
-                )
-            ):
+            elif _should_skip_decomp_rule_capture(op.op):
                 # Note that the graph-decomposition returns abstracted rules
                 # for Adjoint and Controlled operations, so we skip them here.
                 # These abstracted rules cannot be captured and lowered.
                 # We use MLIR AdjointOp and ControlledOp primitives
                 # to deal with decomposition of symbolic operations at PLxPR.
+                continue
+            else:
                 raise ValueError(f"Could not capture {op.op} without the number of wires.")
 
 
@@ -353,6 +344,26 @@ def _create_decomposition_rule(
     # Note that we shouldn't pass args as kwargs to decomposition_rule
     # JAX doesn't like it and it may fail to preserve the order of args.
     return decomposition_rule(func_cp, pauli_word=pauli_word)(*args)
+
+
+def _should_skip_decomp_rule_capture(op_rep) -> bool:
+    """Return True if a decomposition-graph op should not be captured as a JAX rule."""
+    op_type = getattr(op_rep, "op_type", None)
+    if op_type is not None and issubclass(
+        op_type,
+        (
+            qp.ops.Adjoint,
+            qp.ops.Controlled,
+            qp.ops.ChangeOpBasis,
+            qp.ops.Prod,
+            qp.TemporaryAND,
+        ),
+    ):
+        return True
+
+    # Operator2 resource reps may omit op_type but still encode symbolic ops in the name.
+    name = op_rep.name
+    return name.startswith(("Adjoint(", "Controlled("))
 
 
 # pylint: disable=protected-access
--- a/frontend/catalyst/from_plxpr/decompose.py
+++ b/frontend/catalyst/from_plxpr/decompose.py
@@ -362,8 +362,9 @@ def _should_skip_decomp_rule_capture(op_rep) -> bool:
         return True
 
     # Operator2 resource reps may omit op_type but still encode symbolic ops in the name.
+    # ControlledOp2 uses the ``C(...)`` naming convention rather than ``Controlled(...)``.
     name = op_rep.name
-    return name.startswith(("Adjoint(", "Controlled("))
+    return name.startswith(("Adjoint(", "Controlled(", "C("))
 
 
 # pylint: disable=protected-access
--- a/frontend/catalyst/from_plxpr/qfunc_interpreter.py
+++ b/frontend/catalyst/from_plxpr/qfunc_interpreter.py
@@ -333,6 +333,10 @@ def handle_operator(
     else:
         control_wire_inputs = control_values = ()
 
+    control_wire_inputs += self.control_wires
+    control_values += self.control_values
+    n_ctrls += len(self.control_wires)
+
     new_wires = [
         w if is_abstract_qubit(w) else qref_get_p.bind(self.init_qreg, w) for w in wire_inputs
     ]
--- a/frontend/test/lit/test_qref/test_flat_circuits.py
+++ b/frontend/test/lit/test_qref/test_flat_circuits.py
@@ -70,23 +70,6 @@ def test_custom_op(i: int):
 print(test_custom_op.mlir)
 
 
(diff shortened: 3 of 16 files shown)


frontend/catalyst/from_plxpr/decompose.py
```diff
@@ -355,6 +346,26 @@ def _create_decomposition_rule(
...
+            qp.ops.ChangeOpBasis,
+            qp.ops.Prod,
+            qp.TemporaryAND,
+        ),
+    ):
+        return True
+
+    # Operator2 resource reps may omit op_type but still encode symbolic ops in the name.
+    name = op_rep.name
+    return name.startswith(("Adjoint(", "Controlled("))
```
JerryChen97: But I thought ControlledOp2's name has been overriden as `C(...)` ? Would be helpful if @andrijapau can also double check if this will miss things e.g. `C(S)` etc.

frontend/test/pytest/test_specs.py
JerryChen97, line 1: `check_specs_resources_same()` now accepts either `SpecsResources` or `PBCSpecsResources` on both sides, but only compares the PBC depth fields when both objects are PBC instances.

Because `PBCSpecsResources` subclasses `SpecsResources`, these mismatches pass:

 - actual `SpecsResources`, expected `PBCSpecsResources`;
 - actual `PBCSpecsResources`, expected `SpecsResources`.

The first case is especially problematic: Catalyst could regress to returning ordinary resources and lose `any_commuting_depth` and` qubit_disjoint_depth`, while both new PBC tests still pass. Those fields are the defining additions of the subclass. If this case is truly important we might also need to add exact type equality

JerryChen97 (changes requested):
Also I'm investigating why [previously observed necessary changes](https://github.com/PennyLaneAI/catalyst/pull/3035/changes#diff-9a21c91a4ebeae0a86592c0f5931b932d7882c1a40007c68619a8055030bee4a) are not actually needed for CI...

frontend/catalyst/from_plxpr/decompose.py
Qottmann: Yes you are right! Operator2 controlled resource reps use C(S), not Controlled(S). The skip logic now also matches names starting with C(, with a comment explaining why. Verified it skips C(S) but not CNOT/CRX.

Qottmann: ✅

frontend/test/pytest/test_specs.py
Qottmann: I turned it to now `assert type(actual_res) is type(expected_res)` so a plain `SpecsResources` actual cant pass against a `PBCSpecsResources`

Qottmann:
> Also I'm investigating why [previously observed necessary changes](https://github.com/PennyLaneAI/catalyst/pull/3035/changes#diff-9a21c91a4ebeae0a86592c0f5931b932d7882c1a40007c68619a8055030bee4a) are not actually needed for CI...

I think those changes are actually necessary! This PR actually mostly avoid (or accept) the qfunc+Operator2 control bug. I pulled in the changes from your PR so the bug gets actually fixed

JerryChen97 (approved):
Thanks! Let's have a changelog for the source code change (no need for those tests adjustment) and find a Catalyst reviewer

Qottmann:
> Let's have a changelog for the source code change (no need for those tests adjustment)

what would that entail? I dont think there is any user-facing upgrade no?

frontend/catalyst/from_plxpr/qfunc_interpreter.py
```diff
@@ -333,6 +333,10 @@ def handle_operator(
     else:
         control_wire_inputs = control_values = ()
 
+    control_wire_inputs += self.control_wires
+    control_values += self.control_values
+    n_ctrls += len(self.control_wires)
```
paul0403, line 338: `self` is a jaxpr interpreter. Why would it have control wires?

frontend/test/lit/test_qref/test_flat_circuits.py
```diff
@@ -70,6 +70,23 @@ def test_custom_op(i: int):
 print(test_custom_op.mlir)
 
 
+# CHECK-LABEL: func.func public @test_qfunc_ctrl_operator2_preserves_control_context
+@qp.qjit(capture=True, target="mlir")
+@qp.qnode(qp.device("null.qubit", wires=2))
+def test_qfunc_ctrl_operator2_preserves_control_context():
```
paul0403: How is this test different than the other controlled gate tests in this file?

frontend/test/pytest/from_plxpr/test_from_plxpr.py
```diff
@@ -437,13 +438,15 @@ def c():
         catalyst_xpr = from_plxpr(plxpr)()
         qfunc_xpr = catalyst_xpr.eqns[0].params["call_jaxpr"]
 
-        assert qfunc_xpr.eqns[-5].primitive == qref_qinst_p
+        assert qfunc_xpr.eqns[-5].primitive == qref_operator_p
```
paul0403, line 441: This is because migration of `S` to operator2 has been completed?

frontend/test/pytest/from_plxpr/test_from_plxpr.py
```diff
@@ -484,6 +487,33 @@ def c(x, wire3):
...
+
+        plxpr = jax.make_jaxpr(circuit)()
+        catalyst_xpr = from_plxpr(plxpr)()
+        qfunc_xpr = catalyst_xpr.eqns[0].params["call_jaxpr"]
+
+        operator_eqn = next(
+            eqn
+            for eqn in qfunc_xpr.eqns
+            if eqn.primitive == qref_operator_p and eqn.params["op_cls"] is qp.S
+        )
```
paul0403: Like the other tests in this file, let's just grab the gate primitive from a direct indexing.

I don't like it either, but in my mind all the tests in this file should be lit tests instead of pytests and should be rewritten. For now, let's just be consistent.

frontend/catalyst/from_plxpr/qfunc_interpreter.py
JerryChen97: `self` is the interpreter, but the interpreter instance also carries context while evaluating
nested JAXPRs. `handle_ctrl_transform()` creates `unroller = copy(self)`, appends the transform's
control wires and values to that copy, and evaluates the nested JAXPR with `unroller.eval(...)`.

Therefore, when an `operator_p` inside that nested JAXPR reaches `handle_operator()`, `self` is the
copied interpreter and `self.control_wires` and `self.control_values` contain the enclosing
quantum-function controls. The inner `operator_p` itself has `n_ctrls=0`, so the interpreter context
is the only carrier for those controls.

Constructed controlled operators encode their controls directly in the `operator_p` inputs. The
three lines in `handle_operator()` merge those controls with any enclosing quantum-function
controls before binding `qref_operator_p`. The legacy `interpret_operation()` path already performs
the same merge.

frontend/test/lit/test_qref/test_flat_circuits.py
JerryChen97: S is op2 now?

frontend/test/pytest/from_plxpr/test_from_plxpr.py
JerryChen97: Yes

frontend/test/pytest/from_plxpr/test_from_plxpr.py
JerryChen97: I agree. We should have another story to audit tests for post-Op2 era

frontend/test/lit/test_qref/test_flat_circuits.py
paul0403: But we already have dedicated operator 2 lowering lit tests when we added the lowering.

JerryChen97: Ah... true, `test_doubly_ctrl(as_qfunc=True)` already covered this part. This lit test is indeed redundant

paul0403: No I meant the lit tests in `test_operator.py` with the mocked out `Operator2` test classes

frontend/catalyst/from_plxpr/qfunc_interpreter.py
paul0403: Ah, forgot contorl is a transform in plxpr

frontend/test/lit/test_qref/test_flat_circuits.py
paul0403: But anyway, feel free to leave this test in or delete it, up to you guys

frontend/test/pytest/from_plxpr/test_from_plxpr.py
Qottmann: ✅

frontend/test/lit/test_qref/test_flat_circuits.py
Qottmann: deleted it ✅

paul0403 (approved):

PR #3053, opened 2026-07-23
Decomp/generic rule lowering

kipawaa:
**Context:**

This PR provides the core infrastructure changes for `Operator2` support in Catalyst. This is centred on the `python_decomposition` function, which lowers all of the decomposition for a given `Operator2` instance. Precompiled, trace-time and on-demand decomposition rules will all be lowered through this utility, providing a unified and universal decomposition rule lowering system. We provide the generic utilities needed by each of the lowering pathways, as well as a unified testing infrastructure with mock `Operator2` instances.

**Benefits:**

**Possible Drawbacks:**
This PR deprecates the legacy interface for decomposition.
Note also that tests using any PennyLane operators will fail until the ops are migrated to PL Operator2, since callbacks are only supported for `Operator2` instances (required for graph IDs).

**Related GitHub Issues:**

[sc-122043]
[sc-122018]


frontend/catalyst/decomposition/precompile_decomposition_rules.py
```diff
@@ -0,0 +1,195 @@
...
+    """
+    Path(decomp_file_path).parent.mkdir(parents=True, exist_ok=True)
+
+    # newline to ensure emptystring is never passed
+    bytecode_lib = "\n"
+
+    with ir.Context():
+        # TODO: update this for Operator2, PL will implement a precompilation registry
+        for op in COMPILER_OPS_FOR_DECOMPOSITION:
+            # TODO: this fails because GraphOpID only supports Operator2. Remove try/except once
```
paul0403:
```suggestion
```

frontend/catalyst/decomposition/precompile_decomposition_rules.py
```diff
@@ -0,0 +1,195 @@
...
+            # Operator2 migration is complete
```
paul0403:
```suggestion
```

frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -0,0 +1,291 @@
...
+    """
+    A parser object to compute the graph operator id for the operator2 instance `op`.
+
+    The format of the computed graph op ID string is as follows:
+        op_name{param_shaped_type_dictionary}{wire_lens_dictionary}{static_data_dictionary}[UID]
+
+    For example, an Operator2 instance with class name `HybridOpArg`, taking in one float param
+    argument named `angle`, one wire argument named `cwires`, one static data argument
+    `label="hello"`, and UID 10 would be parsed to the following graph op ID:
+        HybridOpArg{angle:[f64]}{cwires:1}{label:hello}[10]
```
paul0403, line 54:
```suggestion
    The format of the computed graph op ID string is as follows:
        op_name{param_shaped_type_dictionary}{wire_lens_dictionary}{static_data_dictionary}[UID]

    The UID is computed from the shapes, dtypes and pytree structures of the `hybrid_args` of the operator2 instance.
    
    For example, an Operator2 instance with class name `HybridOpArg`, taking in one float param
    argument named `angle`, one wire argument named `cwires`, one static data argument
    `label="hello"`, and a computed UID of 10 would be parsed to the following graph op ID:
        HybridOpArg{angle:[f64]}{cwires:1}{label:hello}[10]
```

frontend/test/lit/GraphDecomposition/TestCost.mlir
```diff
@@ -0,0 +1,28 @@
...
+
+func.func @circuit() -> !quantum.bit {
+    %0 = quantum.alloc(1) : !quantum.reg
+    %q = quantum.extract %0[0] : !quantum.reg -> !quantum.bit
+    // CHECK-NOT: testHadamard
+    // CHECK-DAG: testRZ
+    // CHECK-DAG: testRY
+    %qout = quantum.custom "testHadamard"() %q : !quantum.bit
+    return %qout : !quantum.bit
+}
```
paul0403, line 28: Just confirming again, these are reenabling lit tests of precompoiled?

frontend/test/pytest/test_decomposition.py
```diff
@@ -0,0 +1,139 @@
...
+        # tmp_bcfile = None
+        #
+        # if orig_bcfile.exists():
+        #     tmp_bcfile = orig_bcfile.replace(BYTECODE_FILE_PATH + ".tmpbackup")
+        #
+        # try:
+        #     precompile_decomp_rules()
+        #     assert orig_bcfile.exists()
+        #
+        # finally:
```
paul0403: Should we delete these pytest placeholders, now that they are in lit tests?

mlir/lib/Quantum/IR/QuantumOps.cpp
```diff
@@ -1220,9 +1222,17 @@ ParseResult OperatorOp::parse(OpAsmParser &parser, OperationState &result) {
 
 std::string CustomOp::getOperatorName() { return getGateName().str(); }
 
-mlir::TypeRange CustomOp::getDynamicShape() { return getAllParams().getTypes(); }
+llvm::StringMap<llvm::SmallVector<mlir::Type>> CustomOp::getDynamicShape() {
+    llvm::StringMap<llvm::SmallVector<mlir::Type>> ret;
+    for (auto [i, param] : llvm::enumerate(getParams())) {
+        ret[std::to_string(i)] = llvm::SmallVector<mlir::Type>({param.getType()});
```
paul0403, line 1228: Should the special naming scheme be explicitly documented in the frontend `GraphOpID`'s docstring?

mlir/unittests/Quantum/Interfaces/TestDecomposableGateInterface.cpp
```diff
@@ -357,9 +347,9 @@ module {
         mlir::DictionaryAttr::get(&context, {arrAttr, stringAttr, intAttr});
     ASSERT_EQ(op.getStaticData(), expectedStaticData);
 
-    ASSERT_EQ(
-        op.getGraphOpId(),
-        "testInterfaceOp[i1,f64,i64][2]{myStaticArray:[1,2,3],myStaticInt:4,myStaticString:Test}");
+    ASSERT_EQ(op.getGraphOpId(),
+              "testInterfaceOp{angle:[f64],flag:[i1],index:[i64]}{wire1:1,wire2:1}{"
```
paul0403, line 351: Just confirming again, `angle:[f64]` has the extra brackets because in param_map, `angle = [1]` has the "list of arg indices" brakets, and the index-1 param is a plain f64 (`%angle: f64`)?

mlir/lib/Quantum/IR/QuantumOps.cpp
paul0403: I just realized there's still some problems with the custom op scheme.

Say we have a custom op `RX(theta=0.2, wires=1)`. When the frontend sees this (in precompiled or lower time), because it has access to the names on the op class, it will parse the GOID to `RX{theta:[f64]}{wires:1}{}`. But when calling from a mlir op instance by on-demand, the ID will be parsed to `RX{0:[f64]}{wires:1}{}`. This causes a mismatch in ID!

Because we don't need the argnames when calling the resource functions of custom ops in our python wrapper (we're calling through `args` instead of `kwargs`, hence why we said custom ops can only have `wires` as the last argument`), we can just generate a nameless ID in the frontend for custom ops as well.

paul0403: We can probably merge this generic branch PR in and fix this in lower time, where we have all the lit tests.

paul0403 (approved):

mlir/lib/Quantum/IR/QuantumOps.cpp
kipawaa: > we can just generate a nameless ID in the frontend for custom ops

Exactly, this is the plan!

>We can probably merge this generic branch PR in and fix this in lower time

absolutely, no need to introduce the added complexity here

mlir/unittests/Quantum/Interfaces/TestDecomposableGateInterface.cpp
kipawaa: Yep!

frontend/test/lit/GraphDecomposition/TestCost.mlir
kipawaa: These are MLIR lit tests to verify the pass works as intended, not for precompiled.

maliasadi (commented):
Could you add a changelog entry for this PR?

frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -0,0 +1,294 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+This module provides infrastructure for lowering decomposition rules via python.
+"""
+
+# pylint: disable=protected-access,bare-except
```
maliasadi: I think this can be removed now

```suggestion
# pylint: disable=protected-access
```

frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -0,0 +1,294 @@
...
+
+    def get_wire_lens_id_format(self) -> str:
+        """Return the wire lengths formatted for GraphOpId."""
+        return "{" + ",".join(f"{name}:{shape}" for name, shape in self.wire_lens.items()) + "}"
+
+    def get_static_data_id_format(self) -> str:
+        """Return the static data formatted for GraphOpId."""
+        return "{" + ",".join(f"{k}:{v}" for k, v in self.static_data.items()) + "}"
+
+    def getID(self) -> str:
```
maliasadi: Could we rename this to `getGraphOpId` similar to the C++ part?

mlir/lib/Quantum/IR/QuantumInterfaces.cpp
```diff
@@ -126,8 +162,8 @@ std::string defaultGetGraphOpId(Operation *op) {
     DecomposableGate gate = cast<DecomposableGate>(op);
 
     ss << gate.getOperatorName();
-    printTypeRange(gate.getDynamicShape(), ss);
-    printIterable(gate.getWireLens(), ss);
+    printDynamicShape(gate.getDynamicShape(), ss);
+    printWireLens(gate.getWireLens(), ss);
     printAttr(gate.getStaticData(), ss);
```
maliasadi, line 157: hm.. do we need to sort the attrubutes following the impl for the other two prints (DynamicShape and WireLen)?

mlir/lib/Quantum/IR/QuantumInterfaces.cpp
```diff
@@ -126,8 +162,8 @@ std::string defaultGetGraphOpId(Operation *op) {
     DecomposableGate gate = cast<DecomposableGate>(op);
 
     ss << gate.getOperatorName();
-    printTypeRange(gate.getDynamicShape(), ss);
-    printIterable(gate.getWireLens(), ss);
+    printDynamicShape(gate.getDynamicShape(), ss);
+    printWireLens(gate.getWireLens(), ss);
```
maliasadi, line 156: the implementation of these two funcs are near identical. It'd be good to merge the duplicated parts

mlir/lib/Quantum/IR/QuantumInterfaces.cpp
paul0403: Yes it probably needs to be sorted, the frontend ID generation does sort the dictionary based on the argnames.

frontend/catalyst/decomposition/decomposition_rules.py
paul0403: https://github.com/PennyLaneAI/catalyst/pull/3053/commits/673bf4ea7e9c6e397ea28d36cab36e40d7f19f01

frontend/catalyst/decomposition/decomposition_rules.py
paul0403: https://github.com/PennyLaneAI/catalyst/pull/3053/commits/673bf4ea7e9c6e397ea28d36cab36e40d7f19f01

mlir/lib/Quantum/IR/QuantumInterfaces.cpp
paul0403: Ah wait, it's already sorted! Inside the functions there's a `llvm::sort(keys)`

maliasadi: could you share a link to the code?

mlir/lib/Quantum/IR/QuantumInterfaces.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/3053/commits/a7757bf4b16b45193c185f51ea1d343ff5ac37ff

mlir/lib/Quantum/IR/QuantumInterfaces.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/blob/a7757bf4b16b45193c185f51ea1d343ff5ac37ff/mlir/lib/Quantum/IR/QuantumInterfaces.cpp#L108

Just linked to a new commit where I address the other comment (factor out comment logic)

maliasadi (approved):
It only needs a changelog entry! Happy to approve :shipit:

PR #3058, opened 2026-07-23
Remove non-graph decomposition fallback when capture=True

albi3ro:
**Context:**

We are porting all our operators over to the new `Operator2` and solving the decomposition graph in a transform in c++.  We are wanting to do away with solving the graph in python, and we are *really* wanting to do away with the legacy `Operator.compute_decomposition` framework.  We have also deleted all plxpr transforms over than the decomposition one.  

Removes all of the legacy decomposition fallbacks.  It either works, or it fails.

A few tests are currently xfailed, but they should start working again once their respective operators are converted to `Operator2`. 

**Benefits:**

Can delete the decompose plxpr transform.  

Less code to wade through to add our new way of doing things.

**Possible Drawbacks:**

Things temporarily regress.

**Related GitHub Issues:**

[sc-125643]


doc/releases/changelog-dev.md
```diff
@@ -191,6 +191,8 @@
 
 <h3>Breaking changes 💔</h3>
 
+* Removes the non-graph decomposition fallback when `capture=True` is enabled.
```
albi3ro, line 218:
```suggestion
* Removes the non-graph decomposition fallback when `capture=True` is enabled.
  [(#3058)](https://github.com/PennyLaneAI/catalyst/pull/3058/)
```

maliasadi (approved):
🧨

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -28,11 +28,10 @@
 from pennylane.capture import PlxprInterpreter, qnode_prim
 from pennylane.capture.primitives import transform_prim
 from pennylane.transforms import decompose as pl_decompose
-from pennylane.transforms.decompose import DecomposeInterpreter
```
paul0403, line 31: If nothing else is using the `DecomposeInterpreter`, can we just delete the interpreter entirely?

paul0403 (approved):
🚀 Music to my ears

frontend/catalyst/from_plxpr/from_plxpr.py
albi3ro: That is the next PR tomorrow after this goes in.

PR #3059, opened 2026-07-23
Additional compiler hints resource estimation on conditionals

dime10:
The `ResourceAnalysis` pass can now estimate resources for conditionals with dynamic predicates using probability hints, analogous to the `estimated_iterations` hint for loops.

Annotating an `scf.if` with a `catalyst.estimated_probability` float attribute (the probability that the condition is true) makes  the pass report the expected (probability-weighted) resource counts, instead of the worst case. 

Similarly, an `scf.index_switch` can be annotated with a `catalyst.estimated_probabilities` float array (one entry per case, the default case's probability is computed automatically).

These discardable attributes are namespaced to the `catalyst` dialect and validated by a verifier. The existing loop trip-count hint has been updated to this format as well.

As a result of this update, (almost) all resource counts are now tracked as floating-point values. Upon emission of the JSON format, the counts are returned to integers if they are integral, otherwise they are emitted as a floating-point value. Comments on this aspect of the PR are especially welcome, since I don't quite know the impact of this change.

[sc-125662]


mlir/lib/Catalyst/Analysis/ResourceResult.cpp
```diff
@@ -107,7 +108,20 @@ void ResourceResult::multiplyByScalar(int64_t scalar)
     }
 
     numAllocQubits *= scalar;
-    numArgQubits *= scalar;
+    // TODO: does it make sense to scale the arg qubit number?
+    // numArgQubits *= scalar;
```
dime10: It seemed wrong to make numArgQubits a float since I'm assuming it's literally just the number of qubits in a function sig. But I'm also confused why this would be scaled then...

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -385,10 +409,43 @@ void ResourceAnalysis::analyzeIndexSwitchOp(scf::IndexSwitchOp switchOp, Resourc
...
+            double p = cast<FloatAttr>(probsAttr[idx]).getValueAsDouble();
+            sumProb += p;
+
+            caseResult.multiplyByScalar(p);
+            expected.mergeWith(caseResult, ResourceResult::MergeMethod::Sum);
+        }
+
+        // The verifier guarantees the entries sum to at most 1, but a clamp is safer in case
+        // of floating-point error.
+        double pDefault = std::max(0.0, 1.0 - sumProb);
```
dwierichs: Should we clamp against floating point errors for the case where the input probabilities sum to 0.?

mlir/lib/Catalyst/Analysis/ResourceResult.cpp
dwierichs: Yeah, this is confusing. It reads as if we were multiplying the resources both in depth and width, so basically scaling by `scalar**2` 😅

mlir/lib/Catalyst/Analysis/ResourceResult.cpp
```diff
@@ -107,7 +108,20 @@ void ResourceResult::multiplyByScalar(int64_t scalar)
...
+}
+
+// Emit a count as a JSON number. Counts are tracked as doubles to support probabilistic
+// (fractional) count values, although generally they are whole numbers. To keep the output
+// clean only print in floating-point format when the value is non-integral.
+static llvm::json::Value countToJson(double count)
+{
+    double rounded = std::nearbyint(count);
+    if (count == rounded && std::abs(count) < 9.007199254740992e15 /* 2^53 */) {
```
dwierichs: Just curious what is the meaning of 2^53 here? :)

mlir/lib/Catalyst/Analysis/ResourceResult.cpp
```diff
@@ -188,6 +202,7 @@ llvm::json::Object ResourceResult::toJson() const
  */
 DictionaryAttr buildResourceDict(MLIRContext *ctx, const ResourceResult &result)
 {
+    // TODO: maintain int counts here for now, but this whole function is deprecated
```
dwierichs:
```suggestion
    // TODO: we maintain int counts here for now, but this whole function is deprecated
```

dwierichs: It's not entirely clear what is the to do 🙃

mlir/lib/Catalyst/IR/CatalystDialect.cpp
```diff
@@ -48,6 +50,88 @@ void CatalystDialect::initialize()
...
+{
+    llvm::StringRef name = attribute.getName().strref();
+
+    if (name == EstimatedIterationsAttrName) {
+        if (!isa<scf::ForOp, scf::WhileOp>(op)) {
+            return op->emitError() << "'" << name << "' is only valid on 'scf.for' or 'scf.while'";
+        }
+        auto iters = dyn_cast<IntegerAttr>(attribute.getValue());
+        if (!iters) {
+            return op->emitError() << "'" << name << "' must be an integer attribute";
```
dwierichs: Can we actually support float values? Consider this example:

```
@for_loop(1, 3)
def outer(i):
    @for_loop(i)
    def inner(j):
        qp.RZ(0.2 * j, wires=[0])
```
The expected iteration count for the inner loop is $(1+2)/2= 1.5$, so I'd like to decorate the inner loop with `estimated_iterations=1.5` 🙃

mlir/lib/Catalyst/Transforms/ResourceAnalysisPass.cpp
```diff
@@ -132,23 +133,35 @@ struct ResourceAnalysisPass : public impl::ResourceAnalysisPassBase<ResourceAnal
...
     void accumulateStats(const ResourceResult &r)
     {
+        double gates = 0.0;
         for (const auto &opEntry : r.operations) {
             for (const auto &sizeEntry : opEntry.getValue()) {
-                totalGates += sizeEntry.second;
+                gates += sizeEntry.second;
             }
         }
+        totalGates += static_cast<int64_t>(std::llround(gates));
```
dwierichs, line 142: So we always round in this accumulation function, irrespective of whether we got "proper" floats?

dwierichs (changes requested):
Nice, thank you @dime10 💯 Just had some very minor questions :)

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
Qottmann: why even make this restriction? The current max-union default also doesnt sum to one

mlir/lib/Catalyst/IR/CatalystDialect.cpp
```diff
@@ -48,6 +50,88 @@ void CatalystDialect::initialize()
                               CallbackCallOp, CallbackOp>();
 }
 
+//===----------------------------------------------------------------------===//
+// Catalyst attributes.
+//===----------------------------------------------------------------------===//
+
+// Verify a probability value: must be a float attribute in the range [0, 1].
```
Qottmann, line 57: that makes sense

Qottmann (commented):
as a user I'd be mostly interested to see how this is realized in the frontend, do you have (working) example code for that?

dwierichs:
> as a user I'd be mostly interested to see how this is realized in the frontend, do you have (working) example code for that?

I'm working on that :)

mlir/lib/Catalyst/Analysis/ResourceResult.cpp
dime10: @jzaia18 @sengthai

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
dime10: This is to compute the probability of the default branch in a switch case, not counts, so I'm not sure it's comparable to the max-union case. We'd want it to be between 0 and 1, but a tiny error is not that big of a deal I guess so we could also just remove the clamp. The verification will already check that we're not egregiously outside of the expected range.
Then again, this clamp prevents very small negative gate counts from showing up, so that does seem somewhat useful 😅 I have no objection to clamping at the top either.

mlir/lib/Catalyst/Analysis/ResourceResult.cpp
dime10: That is the max integer we can represent with a double without loss of precision. The idea is to convert back to int only if we know that we have an integer in our float value.

But the output format is definitely up for discussion here. We could always round and output ints, round everything to 1 decimal, the current conditional conversion back to int, ...

mlir/lib/Catalyst/Analysis/ResourceResult.cpp
dime10: Remove it I guess 😆 I can remove the comment. I was actually told in a meeting yesterday that we'll keep this method so ...
```suggestion
```

mlir/lib/Catalyst/IR/CatalystDialect.cpp
dime10: Good point, I'll update that hint as well 👍

mlir/lib/Catalyst/Transforms/ResourceAnalysisPass.cpp
dime10: The limitation here is that the MLIR pass statistics we are reporting here only works with ints (or at least the way it's currently set up). This does not affect specs output though, just the builtin mlir statistics reporting.

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
dime10: Actually, clamping the other end is not needed. At worst `sumProb` is 0.0, it can't be negative. So the maximum value we can get here is already capped at 1.0. I don't float rounding errors can change this.

mlir/lib/Catalyst/Analysis/ResourceResult.cpp
sengthai: Ooop this should not be scaled since the numArgQubits is from entry function arg. So, we are safe to remove this. also, we probably want to remove in here as well. 
https://github.com/PennyLaneAI/catalyst/blob/ca59b051bbe1b4a70e1074693e2e103c851186f3/mlir/lib/Catalyst/Analysis/ResourceResult.cpp#L83

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
dime10: Added it anyway: https://github.com/PennyLaneAI/catalyst/pull/3059/commits/7e21caf8c16ba50a68712e8e3e52e4e94d60620c

mlir/lib/Catalyst/Analysis/ResourceResult.cpp
dime10: Anton decided to go for all ints, so this line code is removed now: https://github.com/PennyLaneAI/catalyst/pull/3059/commits/d6069ab0f5f700860044e1d839db096d6df3e8af

mlir/lib/Catalyst/Analysis/ResourceResult.cpp
dime10: Thanks for weighing in, I removed both instances: https://github.com/PennyLaneAI/catalyst/pull/3059/commits/245e51c0c98aa0af1a6bd8327af86d0b76125205

mlir/lib/Catalyst/IR/CatalystDialect.cpp
dime10: https://github.com/PennyLaneAI/catalyst/pull/3059/commits/ba857785ebce248eaf7fea7967fc9a3928596d71

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -372,6 +374,27 @@ void ResourceAnalysis::analyzeIfOp(scf::IfOp ifOp, ResourceResult &result, bool
...
+            analyzeRegion(ifOp.getElseRegion(), elseResult, isAdjoint);
+        }
+
+        thenResult.multiplyByScalar(pThen);
+        elseResult.multiplyByScalar(pElse);
+        thenResult.mergeWith(elseResult, ResourceResult::MergeMethod::Sum);
+
+        result.mergeWith(thenResult);
+        return;
+    }
```
sengthai, line 395: Looks nice! 👍

mlir/test/Catalyst/ResourceAnalysisTest.mlir
sengthai, line 1: These tests look great! As a completely optional suggestion, it would be awesome to add a more complex test that combines a for-loop, a switch, and an if-op.

sengthai (approved):
Thanks @dime10 for the hint! 🫆

mlir/lib/Catalyst/Analysis/ResourceResult.cpp
dwierichs: Thanks for the explainer 🙏 
Sounds good 🤷

dwierichs (approved):
Thanks @dime10 , looking forward to using this!

Qottmann (approved):

mlir/test/Catalyst/ResourceAnalysisTest.mlir
dime10: Not a bad idea, but there is a test combining for and if. For expediency I'm going to skip adding another that combines all three

PR #3060, opened 2026-07-24
Expose resource hints

dwierichs:
**Context:**
In #3059, additional hints are added to the resource analysis of `cond` statements, and the one for `for_loop` is exposed one layer higher up. They are not fully exposed to the frontend yet.

Expose and pass through these hints, in form of arguments.

**Benefits:**
Together with another PR in PennyLane, we will be able to use these hints in user code.

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-125662]


frontend/catalyst/api_extensions/control_flow.py
```diff
@@ -66,13 +66,14 @@
...
 )
+from catalyst.resource_hints import normalize_estimated_probabilities_for_cond
 from catalyst.tracing.contexts import EvaluationContext, EvaluationMode
 from catalyst.utils.exceptions import PlxprCaptureCFCompatibilityError
 from catalyst.utils.patching import Patcher
 
 
 ## API ##
-def cond(pred: DynamicJaxprTracer):
+def cond(pred: DynamicJaxprTracer, *, estimated_probabilities: float | list[float] | None = None):
```
dime10: I'm guessing the list variant is for elif branches? I was thinking we could keep it a single float, and the elif decorators each accept an optional float as well.

dwierichs: Ah, yeah, that makes sense! Would we want the probability in the `elif` to be the conditional or unconditional probability? I.e. if cases a, b c trigger with probs `[0.5, 0.25, 0.25]`, would we want
```
@qp.cond(case==a, estimated_probability=0.5)
def func(): ...

@func.else_if(case==b, estimated_probability=0.25)
def case_b(): ...

@func.otherwise
def case_c(): ...
```
or 
```
@qp.cond(case==a, estimated_probability=0.5)
def func(): ...

@func.else_if(case==b, estimated_probability=0.5)
def case_b(): ...

@func.otherwise
def case_c(): ...
```

dwierichs: Went with the unconditional probability for now, it feels slightly more canonical.

dime10: Yeah I think uncoditional, because the if statements aren't nested yet 👍

frontend/catalyst/jax_tracer.py
```diff
@@ -509,13 +509,17 @@ def __init__(
         apply_reverse_transform=False,
         expansion_strategy=None,
         debug_info=None,
+        estimated_iterations=None,
+        estimated_probabilities=None,
```
dime10: This looks a bit strange (specific kwargs in generic base class), but I'm guessing there is no easy way around that?

frontend/catalyst/resource_hints.py
dime10, line 1: This ones a bit of an odd module, because it contains both frontend API helper functions as well as jaxpr-to-mlir lowering functions. So it covers very disparate stages of the frontend 🤔 

On top of that, we are trying to avoid new top-level modules, in order to move to a more organized code structure, regrouped around stages or core functionality of the frontend. Unfortunately, for the jaxpr primitive / lowering code, we haven't moved to such a structure yet. 

What I would recommend is, move `collect_estimated_probabilities_for_cond` to `api_extensions/control_flow`, since that is where it's used (helper functions go at the bottom of the file).

For the other functions, they all revolve around lowering the cf primitives. They should normally go into the module where they are used, but because the `jax_primitves` module was getting so large, helper functions have been outsourced to the `jax_primitives_utils` (not great, but I guess it'll have to do until the refactor).

frontend/test/pytest/test_loops.py
dime10, line 1: We probably want some postitive tests somewhere of the attributes being set and correctly lowered to jaxpr and mlir.

dime10 (changes requested):
Looks good, I think we just need some tests.

frontend/test/pytest/test_loops.py
dime10: Oh and an integration test?

dwierichs: Sorry, I was doing this in tandem with the PL PR, and overlooked the necessity of tests 🤦

frontend/catalyst/api_extensions/control_flow.py
```diff
@@ -901,6 +931,9 @@ def _call_with_quantum_ctx(self):
             out_classical_tracers,
             regions,
             expansion_strategy=self.expansion_strategy,
+            estimated_probabilities=collect_estimated_probabilities_for_cond(
```
dwierichs: I guess this is a little funny: At the frontend, we have `estimated_probability` in `cond` and `else_if`, they get collected into a single sequence `estimated_probabilities` for the `HybridOp`, but then they get split up via the recursion/nesting at the lower level, so that we get `estimated_probability` attributes on multiple `scf.if`s again. I suppose that also reflects that we move from unconditional to conditional logic, I just thought I mentioned this realization :D

dime10: Yeah 😅 I think it's okay given how we handle elifs

dime10 (approved):
Awesome!

paul0403 (approved):

PR #3061, opened 2026-07-24
Decomp/lower time rules

paul0403:
**Context:**
Automatically compile decomposition rules reachable from a gate when lowering that gate in frontend.

[sc-122036]


frontend/test/lit/test_operator2/test_lower_time_rules.py
```diff
@@ -468,12 +468,16 @@ def c():
...
+# CHECK-SAME:   qubits({{%.}}, {{%.}}, {{%.}}, {{%.}}, {{%.}}, {{%.}}, {{%.}})
+# CHECK:   UID([[uid:[0-9]+]]) forward({{%.+}}: tensor<f64>, {{%.+}}: tensor<i64>)
+# CHECK:   param_map = {angles1 = [0], angles2 = [1]}
+# CHECK-SAME:  qubit_map = {hwires1 = [4, 5], hwires2 = [6], op1 = [2], op2 = [3], reg1 = [0], reg2 = [1]}
+#
+# CHECK: func.func private @"__builtin_rule_MultipleFullArgs{angles1:[[f64]],angles2:[[f64]]}{hwires1:2,hwires2:1,op1:1,op2:1,reg1:1,reg2:1}[[[uid]]]"
 # CHECK-SAME:   resources = {operations = {
 # CHECK-SAME:   "NoParams{}{reg:1}{}" = 2 : i64
-# CHECK-SAME:   target_gate = "MultipleFullArgs{angles1:[f64],angles2:[f64]}{reg1:1,reg2:1}{}[[[uid]]]"
+# CHECK-SAME:   target_gate = "MultipleFullArgs{angles1:[[f64]],angles2:[[f64]]}{hwires1:2,hwires2:1,op1:1,op2:1,reg1:1,reg2:1}{}[[[uid]]]"
```
paul0403: @kipawaa I'm leaving it to you to deal with the extra layer of list

kipawaa: [330d2ac](https://github.com/PennyLaneAI/catalyst/pull/3061/commits/330d2acd80fe7fed19db05e9e925d9956d0ed60a)

frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -12,160 +12,121 @@
...
 )
-from catalyst.from_plxpr.qref_operator2_primitives import _is_custom_op
 from catalyst.from_plxpr.uid import generate_uid
 from catalyst.jax_extras.lowering import get_mlir_attribute_from_pyval
 
 
-class GraphOpID:
-    """
-    A parser object to compute the graph operator id for the operator2 instance `op`.
+def get_rule_strings_from_module(module: ir.Module) -> list[str]:
```
kipawaa, line 36: It seems like we got the names of these helpers backwards - this one returns the funcops but is named `get_rule_strings`, the other returns strings and is called `get_rules`

frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -12,160 +12,121 @@
...
+             `__builtin_` prefix to their names.
+    """
+    funcOps = get_rule_strings_from_module(module)
+    return "\n".join(str(funcOp) for funcOp in funcOps) if funcOps else ""
+
+
+def inject_new_rules_into_module(module: ir.Module, decomp_rules: list[str]):
+    with ir.InsertionPoint(module.body):
+        for decomp_rule in decomp_rules:
+            if not decomp_rule:
```
kipawaa: Is this used?

frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -12,160 +12,121 @@
...
+                continue
+
+            decomp_rule_op = ir.Operation.parse(decomp_rule)
```
kipawaa, line 92: If we have a helper that gets the rules (as `funcOp`s) why don't we design this function to work from those, rather than stringifying and re-parsing?

frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -174,10 +135,8 @@ def collect_resources_for_op(op_name, kwargs, is_custom_op=False):
         try:
             # The `compute_resources` function's signature is the same as the Operator2 signature
             # for the original op of the rule
-            if is_custom_op:
-                args = tuple(val for key, val in kwargs.items() if key != "wires")
-                kwargs = {"wires": kwargs["wires"]}
             resources = rule.compute_resources(*args, **kwargs)
+            print(f"got resources for rule {rule.name}: {resources}")
```
```suggestion
```
😅 my bad

frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -289,3 +257,61 @@ def compile_decomposition_rules_wrapper(
             is_custom_op=is_custom_op,
         )
     )
+
+
+def fetch_all_reachable_decomposition_rules_from_op(
+    op_name, op_id, dynamic_shape, wire_lens, static_data, extra_data=None
+):
+    extra_data = extra_data or {}
```
kipawaa, line 255: Love this pattern, I didn't know this worked!

frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -289,3 +257,61 @@ def compile_decomposition_rules_wrapper(
...
+            op_name, op_id, dynamic_shape, wire_lens, static_data, extra_data=extra_data
+        )
+    )
+
+    while len(queue) != 0:
+        this_name, this_dynamic_shape, this_wire_lens, this_static_data, this_extra_data = (
+            queue.popleft()
+        )
+        this_extra_data = this_extra_data or {}
+        this_args, this_kwargs = prepare_op_args(
```
kipawaa, line 283: I guess `is_custom_op` is only required when calling from the compiler with on-demand rules? Maybe we should rename this for clarity

frontend/catalyst/decomposition/graph_op_id.py
```diff
@@ -0,0 +1,162 @@
...
+        """Return the name of the operator."""
+        return self.operator_name
+
+    def get_dynamic_shape(self) -> dict:
+        """Return a dictionary of names to dynamic shapes."""
+        return self.dynamic_shape
+
+    def get_dynamic_shape_id_format(self) -> str:
+        """Return the dynamic shape formatted for GraphOpId."""
+        return format_dynamic_params_for_id(convert_types_to_mlir_strings(self.dynamic_shape))
```
kipawaa, line 157: If these utilities are only used here then we should move them to this file

frontend/catalyst/from_plxpr/qref_operator2_primitives.py
```diff
@@ -206,6 +219,182 @@ def _process_qubits(*args, op_cls, wire_lens, hybrid_lens) -> tuple[list, dict[s
...
+    repack_static_data=None,
+    uid=None,
+    avals_in=None,
+):
+    """
+    Generate all the decomposition rules registered on the current gate, recursively generating all
+    the rules that are registered on the resource gates of these rules as well.
+    """
+    if is_custom_op:
+        dynamic_shape = {dynamic_argname: ["f64"] for dynamic_argname in op_cls.dynamic_argnames}
```
kipawaa: This should use numbered argnames rather than the `op_cls` argnames, since these are transparent to the compiler and must match.

frontend/catalyst/from_plxpr/qref_operator2_primitives.py
```diff
@@ -206,6 +219,182 @@ def _process_qubits(*args, op_cls, wire_lens, hybrid_lens) -> tuple[list, dict[s
...
+    hybrid_trees=None,
+    repack_static_data=None,
+    uid=None,
+    avals_in=None,
+):
+    """
+    Generate all the decomposition rules registered on the current gate, recursively generating all
+    the rules that are registered on the resource gates of these rules as well.
+    """
+    if is_custom_op:
```
kipawaa, line 244: it might be nice to re-format this as a match-case, maybe with some simplified pre-parsing for the different cases. Maybe we use this as a generic routing function with a match-case that dispatches to the individual parsing/lowering rules

frontend/catalyst/from_plxpr/uid.py
```diff
@@ -67,7 +68,13 @@ def generate_uid(
     reduced.append(("static", reduced_static_args))
     reduced.append(("adjoint", adjoint))
     reduced.append(("n_ctrls", n_ctrls))
-    return hash(tuple(reduced))
+
+    encoded_bytes = str(reduced).encode("utf-8")
+    sha_hash = hashlib.sha256(encoded_bytes).hexdigest()
```
kipawaa, line 73: 🙌

frontend/catalyst/jit.py
```diff
@@ -96,6 +96,7 @@ def qjit(
     dialect_plugins=None,
     capture="global",
     skip_preprocess=False,
+    skip_decomp_rules=False,
```
kipawaa: I feel like naming this `collect_decomp_rules` rather than `skip_decomp_rules` would be more intuitive (avoiding the double-negative in the checks)? Not a crucial change though

frontend/test/lit/test_operator2/test_decomposition_rule_entry_point.py
```diff
@@ -0,0 +1,938 @@
...
+        )
+        print(result)
+
+
+# CHECK: func.func private @"rule_MultipleFullArgs{angles1:{{\[\[f64\]\]}},angles2:{{\[\[f64,f64\]\]}}}{reg1:1,reg2:2}{}[-4321]"
+# CHECK-SAME:   resources = {operations = {
+# CHECK-SAME:   "NoParams{}{reg:1}{}" = 1 : i64
+# CHECK-SAME:   "SingleParam{x:{{\[\[f64\]\]}}}{reg:1}{}" = 1 : i64
+# CHECK-SAME:   "SingleParam{x:{{\[\[i64\]\]}}}{reg:1}{}" = 1 : i64
+# CHECK-SAME:   target_gate = "MultipleFullArgs{angles1:{{\[\[f64\]\]}},angles2:{{\[\[f64,f64\]\]}}}{reg1:1,reg2:2}{}[-4321]"
```
kipawaa: Should this be updated since UID is unsigned?

frontend/test/lit/test_operator2/test_lower_time_rules.py
```diff
@@ -0,0 +1,549 @@
...
+
+            @qp.qjit(capture=True, target="mlir")
+            @qp.qnode(qp.device("null.qubit", wires=3))
+            def c():
+                NoParams(reg=[0, 1])
+                return qp.state()
+
+            print(c.mlir)
+
+        # CHECK: func.func public @c()
```
kipawaa:
```suggestion
        # CHECK-LABEL: func.func public @c()
```
and for the rest of these tests

frontend/test/lit/test_operator2/test_lower_time_rules.py
```diff
@@ -0,0 +1,549 @@
...
+# CHECK-SAME:   "NoParams{}{reg:2}{}" = 1 : i64
+# CHECK-SAME:   target_gate = "CompilableData{}{wires:1}{a:a,b:b,thing:thing}"
+# CHECK: func.func private @"__builtin_ruleAB_NoParams{}{reg:3}{}"
+# CHECK-SAME:   resources = {operations = {
+# CHECK-SAME:   "SingleParam{x:{{\[\[f64\]\]}}}{reg:1}{}" = 1 : i64
+# CHECK-SAME:   target_gate = "NoParams{}{reg:3}{}"
+test_with_cycles()
+
+
+def test_to_multiple_full_args_op():
```
kipawaa, line 359: Does this more belong in the entry point test file? Same with the tests below

kipawaa (commented):
Some clean-up to do, but overall it's looking good!

frontend/test/lit/test_operator2/test_lower_time_rules.py
paul0403: This is not a label tho? This is a full line.

I can however give each test's circuit its own name (instead of using `c` for all of them), so we can use `CHECK-LABEL`, and have an easier time debugging

frontend/test/lit/test_operator2/test_lower_time_rules.py
paul0403: No, they go through different logic. This test is lower time, as it's calling the entry point during lowering of the `NoParam` in the main circuit.

These are the "to" and "from" for an operator 2 with multiple argnames on all fields. I want to test that I called the entry point in the operator_p's lowering rule correctly.

frontend/test/lit/test_operator2/test_lower_time_rules.py
kipawaa: What distinguishes a line from a label? Either way, I think that more descriptive names and using `CHECK-LABEL` here to indicate that we're transitioning to a new test would improve readability and simplify parsing.

frontend/test/lit/test_operator2/test_lower_time_rules.py
kipawaa: Got it, sounds good!

frontend/catalyst/decomposition/decomposition_rules.py
kipawaa: I was mistaken, this doesn't actually return the funcops - but I think it should? Why are we choosing to return strings? I'll look into this

paul0403: Discussed offline, we'll keep it as strings for now. 

The rule funcop is created in the rule qjit context, so its function name (which is a string ref) is referencing a string in the rule qjit context. That context is destroyed when the rule qjit process finishes. When moving the rule funcop to the user qjit context, the name of the rule is now referring to a destroyed string, causing a segfault. Calling `rule_func_op.detach_from_parent()` is apparantly insufficient to resolve this.

There's probably a way around this, but we don't have time for it now.

frontend/catalyst/decomposition/decomposition_rules.py
paul0403: Same comment as above.

frontend/catalyst/decomposition/graph_op_id.py
paul0403: Are you talking about `format_dynamic_params_for_id` and `convert_types_to_mlir_strings`? They are used in other places too.

frontend/catalyst/from_plxpr/qref_operator2_primitives.py
paul0403: This is a bit tricky because there's no case you would match for custom op. For the special ops we can match-case on `op_cls`, but then the default can either be custom op or operator op.

frontend/catalyst/jit.py
paul0403: https://github.com/PennyLaneAI/catalyst/pull/3061/commits/052976a4610584121696e27ba0c52b84eda96182

frontend/catalyst/decomposition/decomposition_rules.py
paul0403: Good catch, removed https://github.com/PennyLaneAI/catalyst/pull/3061/commits/052976a4610584121696e27ba0c52b84eda96182

frontend/catalyst/decomposition/decomposition_rules.py
paul0403: It's also required for lower time. When lowering an op, the `_is_custom_op` check from the primitive lowering rule is passed through into here.

frontend/catalyst/from_plxpr/qref_operator2_primitives.py
kipawaa: you can use `case _ if (condition)` to match a conditional wildcard. Not a requirement though, I looked into this yesterday and it's more work than it's worth right now

frontend/catalyst/decomposition/decomposition_rules.py
kipawaa: We should probably update this so it's not just `False` then. I think I set this when I was updating the custom op handling, but since there wasn't any logic here for custom op I just set it to false. Now I'm thinking there's likely a bug here if we decompose into custom ops during lower-time. Could you add a test for this and if needed update the logic here?

kipawaa: We should probably update this so it's not just `False` then. I think I set this when I was updating the custom op handling, but since there wasn't any logic here for custom op I just set it to false. Now I'm thinking there's likely a bug here if we decompose into custom ops during lower-time. Could you add a test for this and if needed update the logic here?

paul0403: Isn't it already not just false now? `is_custom_op=this_is_custom_op`

paul0403: The test is `test_lower_time_rules.py/test_to_custom_op()`

doc/releases/changelog-dev.md
```diff
@@ -81,6 +81,12 @@
     2. When lowering a gate operation from JAXPR to MLIR, all rules reachable from that gate are injected into the IR.
     [(#3061)](https://github.com/PennyLaneAI/catalyst/pull/3061)
 
+    This pathway of rule injection can be opted-out via a new keyword argument on `qp.qjit` named `collect_decomp_rules`.
+    This kwarg controls whether or not to compile the decomposition rules during lower-time. Default value is `True`.
+    If ``False``, only the circuit itself will be compiled (aka standard legacy behavior).
+    If `True`, all the decomposition rules reachable from all gates in the circuit will be compiled.
+    If `capture=False`, or `capture="global"` and `qp.capture.enabled() == False`, this argument will be ignored.
```
maliasadi, line 88:
```suggestion
    If the program capture mechanism is disabled, this argument will be ignored.
```

kipawaa (approved):
LGTM! 🥇

doc/releases/changelog-dev.md
paul0403: capture=global means follow whatever the global capture flag is when qjit is called.

I copy pasted this docstring from Mudit's `skip_device_preprocessing` above

frontend/catalyst/decomposition/graph_op_id.py
```diff
@@ -0,0 +1,187 @@
...
+            for arg in self.op.dynamic_args.values()
+        ) and not issubclass(type(self.op), tuple(_SPECIAL_LOWERINGS.keys()))
+
+    def get_operator_name(self) -> str:
+        """Return the name of the operator."""
+        return self.operator_name
+
+    def get_dynamic_shape(self) -> dict:
+        """Return a dictionary of names to dynamic shapes."""
+        return self.dynamic_shape
```
maliasadi: we could remove this now. right?

```suggestion
```

maliasadi (approved):
🥳

doc/releases/changelog-dev.md
maliasadi: we could use a more generic statement for capture disabled as the "global" flag seems an ad-hoc

maliasadi: no strong opinion here. feel free to resolve it

frontend/catalyst/decomposition/graph_op_id.py
paul0403: https://github.com/PennyLaneAI/catalyst/pull/3061/commits/0f99929a1bbcb9f719693dec40fae3240cd9492f

doc/releases/changelog-dev.md
paul0403: I think the spirit was just giving the behavior for each possible value of the kwarg. But if there's no strong opinion I'll resolve

PR #3063, opened 2026-07-24
Add inject-transport-session pass to emit transport session IR

mehrdad2m:
**Context:**
Backline places a QNode's controller and its coprocessors on separate executors that talk over a transport. The placement is serialized onto the module as the catalyst.backline attribute (controller + coprocessors + per-node backend/peer/triple/data-path info). This pass turns that declarative attribute into the concrete transport-dialect session lifecycle, wired into the setup/teardown functions the runtime already calls around every execution.

Reading `catalyst.backline,` the pass injects, per session:

bring-up into `@setup`: `transport.create` → `connect` → `exchange_keys` → `establish_channel` → `commit_work_item` (controller) / `set_coprocessor_fn` (coprocessor) → start
tear-down into `@teardown`: `get_session` → `stop` → `destroy`

Topologies handled:
- Controller-only: single controller session, straight-line bring-up/teardown.
- Local coprocessors: controller + coprocessor sessions created together in the host @setup; the coprocessor's `connect/exchange_keys` run async so the two ends handshake concurrently.
- Remote coprocessors: each coprocessor's ops go into a generated `module_coproc` tagged with its catalyst.target triple and catalyst.dispatch address; the host launches `coproc_serve` (nonblocking) / `coproc_stop`.
- Remote controller:  the controller's ops go into its role-tagged target module as `setup_transport/teardown_transport` and the host `@setup/@teardown` become the single orchestration point that launches every dispatched role in dependency order (serve → controller setup; coproc stop → controller teardown), and `@setup` is marked catalyst.backline_bringup so the object-shipping pass knows where to ship from.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


mlir/test/Transport/InjectTransportSession.mlir
```diff
@@ -0,0 +1,240 @@
...
+// Multiple co-located coprocessors: one controller session each, keyed by the peer name.
+
+// CHECK-LABEL: func.func public @jit_circuit
+// CHECK-NOT:     {{[^#]}}transport.
+
+// CHECK-LABEL: func.func @setup
+// CHECK:         quantum.init
+// CHECK-DAG:     transport.create {{.*}}key = "cop0"{{.*}} -> !transport.session<controller>
+// CHECK-DAG:     transport.create {{.*}}key = "cop1"{{.*}} -> !transport.session<controller>
+// CHECK-DAG:     transport.create {{.*}}key = "cop0"{{.*}} -> !transport.session<coprocessor>
```
multiphaseCFD, line 274: why ''cop0'' is required to create a `!transport.session<coprocessor>` here? `transport.create {{.*}}key = "cop0"{{.*}} -> !transport.session<coprocessor>`

mlir/lib/Transport/IR/TransportDialect.cpp
```diff
@@ -37,13 +37,69 @@ using namespace catalyst::transport;
...
+bool NodeAttr::isRemote() const
+{
+    BoolAttr r = getRemote();
+    return r && r.getValue();
+}
+
+static int64_t intOr(IntegerAttr field, int64_t dflt) { return field ? field.getInt() : dflt; }
+
+int64_t NodeAttr::oobPort() const { return intOr(getOobPort(), 0); }
+int64_t NodeAttr::inBytes() const { return intOr(getInBytes(), 8); }
```
multiphaseCFD: yes, `8` is used as a default value, which aligns with the default number of bytes for `syndrome` and `correction` in the backend. Could we use a `const` variable instead?

mlir/lib/Transport/Transforms/InjectTransportSession.cpp
```diff
@@ -0,0 +1,336 @@
...
+
+namespace catalyst {
+namespace transport {
+
+#define GEN_PASS_DEF_INJECTTRANSPORTSESSIONPASS
+#include "Transport/Transforms/Passes.h.inc"
+
+namespace {
+
+constexpr llvm::StringRef kBacklineAttr = "catalyst.backline";
```
multiphaseCFD, line 43: Just want to confirm if `k` prefix naming for const variables is a Catalyst convention

mlir/include/Transport/IR/TransportDialect.td
```diff
@@ -86,6 +87,72 @@ def Transport_CoprocessorSession : Type<
...
+    let summary = "A backline participant: a controller or a coprocessor.";
+
+    let parameters = (ins
+        Transport_StrField:$name,
+        Transport_StrField:$peer,
+        OptionalParameter<"mlir::IntegerAttr">:$oob_port,
+        Transport_StrField:$backend_lib,
+        Transport_StrField:$config,
+        Transport_StrField:$data_path,
+        Transport_StrField:$triple,
```
multiphaseCFD, line 113: Would we add a remote inquiry to ensure the remote OS info is correct in the program capture stage?

mlir/lib/Transport/Transforms/InjectTransportSession.cpp
```diff
@@ -0,0 +1,336 @@
...
+            OpBuilder rb(blk, blk->begin());
+            func::ReturnOp::create(rb, loc);
+            return fn;
+        };
+
+        // Remoteness comes from the attribute, not from whether a target module is present.
+        bool remoteController = ctrl.isRemote();
+
+        ModuleOp ctrlMod;
+        if (remoteController) {
```
multiphaseCFD, line 425: Just want to confirm if it's a requirement that there would be a module with `catalyst.backline_role` attribute if `remote = true`? Is that module injected in a pass before the current one? Could you point me to the pass?

mlir/lib/Transport/Transforms/InjectTransportSession.cpp
```diff
@@ -0,0 +1,336 @@
...
+        };
+        func::FuncOp hostSetup = findOrCreate("setup");
+        func::FuncOp hostTeardown = findOrCreate("teardown");
+
+        // A remote controller's ops go into its target module, which @setup/@teardown launch.
+        func::FuncOp setupFn =
+            remoteController ? makeVoidFunc("setup_transport", ctrlMod) : hostSetup;
+        func::FuncOp teardownFn =
+            remoteController ? makeVoidFunc("teardown_transport", ctrlMod) : hostTeardown;
+
```
multiphaseCFD: Per L110-L113: if controller is remote the whole program is driven by host and we need to add setup/teardown for the remote?

mlir/test/Transport/InjectTransportSession.mlir
```diff
@@ -0,0 +1,240 @@
...
+// CHECK:         transport.start %[[S]] : !transport.session<controller>
+
+// CHECK-LABEL: func.func @teardown
+// CHECK:         quantum.finalize
+// CHECK:         %[[T:.*]] = transport.get_session {{.*}} : !transport.session<controller>
+// CHECK:         transport.stop %[[T]] : !transport.session<controller>
+// CHECK:         transport.destroy %[[T]] : !transport.session<controller>
+module attributes {catalyst.backline = #transport.backline<transport = "net", controller = #transport.node<backend_lib = "x", config = "c",
+    peer = "127.0.0.1", oob_port = 18590 : i16, data_path = "cpu_verbs",
+    in_bytes = 8 : i64, out_bytes = 8 : i64>>} {
```
multiphaseCFD, line 39: Can we add a test for `remote = true` (no coprocessor specified)?

mlir/lib/Transport/IR/TransportDialect.cpp
```diff
@@ -37,13 +37,69 @@ using namespace catalyst::transport;
...
+int64_t NodeAttr::outBytes() const { return intOr(getOutBytes(), 8); }
+int64_t NodeAttr::workItemIdx() const { return intOr(getWorkItemIdx(), 0); }
+
+LogicalResult BacklineAttr::verify(function_ref<InFlightDiagnostic()> emitError,
+                                   StringAttr transport, NodeAttr controller,
+                                   llvm::ArrayRef<NodeAttr> coprocessors)
+{
+    if (!controller)
+        return emitError() << "backline requires a controller";
+    for (NodeAttr c : coprocessors) {
```
multiphaseCFD, line 86: So there is no need to specify `coprocessor`?

mlir/lib/Transport/Transforms/InjectTransportSession.cpp
```diff
@@ -0,0 +1,336 @@
...
+constexpr llvm::StringRef kControllerRole = "controller";
+constexpr llvm::StringRef kCoprocessorRole = "coprocessor";
+
+struct InjectTransportSessionPass
+    : public impl::InjectTransportSessionPassBase<InjectTransportSessionPass> {
+    using InjectTransportSessionPassBase::InjectTransportSessionPassBase;
+
+    void runOnOperation() override
+    {
+        ModuleOp mod = getOperation();
```
multiphaseCFD, line 404: I found it a bit hard to follow for me. Each configuration spans several branches. If you're open to it, I'd suggest refactoring it into one function per config.

multiphaseCFD (commented):
Thank you @mehrdad2m ! A few comments.

mlir/lib/Transport/IR/TransportDialect.cpp
```diff
@@ -37,13 +37,69 @@ using namespace catalyst::transport;
...
+        if (!c)
+            return emitError() << "null coprocessor";
```
rniczh: You can add this into your .clang-format

```
InsertBraces: true
```

mlir/lib/Transport/Transforms/InjectTransportSession.cpp
```diff
@@ -0,0 +1,336 @@
...
+                    .getSession();
+            Value t1 =
+                ConnectAsyncOp::create(b, loc, tokTy, co, coproc.getPeer(), i16A(coproc.oobPort()))
+                    .getToken();
+            ConnectOp::create(b, loc, ct, coproc.getPeer(), i16A(coproc.oobPort()));
+            BarrierOp::create(b, loc, t1);
+            Value t2 = ExchangeKeysAsyncOp::create(b, loc, tokTy, co).getToken();
+            ExchangeKeysOp::create(b, loc, ct);
+            BarrierOp::create(b, loc, t2);
+            EstablishChannelOp::create(b, loc, co, coproc.dataPathOr("cpu_verbs"));
```
rniczh: Use the keyword here (e.g. `kCpuVerbs`) would be much more easier for maintaining?

rniczh (approved):
Thanks @mehrdad2m

mlir/include/Transport/IR/TransportDialect.td
mehrdad2m: yes, the logic should be at the https://github.com/PennyLaneAI/catalyst/pull/3082 where two executors are launched on the remote devices once the backline object is created, and if that fails, we should raise an error.

mlir/lib/Transport/IR/TransportDialect.cpp
mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/3063/commits/809817c3b134925a6e6ca0100c9ea91a869a8920

mlir/lib/Transport/Transforms/InjectTransportSession.cpp
mehrdad2m: no I don't think there is any convention in Catalyst. This is just a coding style I have seen some people use. k is for "constant" so you can tell something is constant by looking at the name. Let me know If you prefer to remove it; I don't have a strong preference :)

mlir/lib/Transport/Transforms/InjectTransportSession.cpp
mehrdad2m: yeah so that would be injected in the frontend while capturing the program. you can see in https://github.com/PennyLaneAI/catalyst/pull/3068 where it is added.

mlir/lib/Transport/Transforms/InjectTransportSession.cpp
mehrdad2m: Yes. Basically, the only way to interact with a remote device is to launch a function similar to how we currently do launch_kernel for calling qnode main function. so if the controller is remote, instead of calling the transport ops individually from host, we have to wrap them inside a function and call it over the remote device

mlir/lib/Transport/IR/TransportDialect.cpp
mehrdad2m: Correct, coprocessors are optional

mlir/lib/Transport/IR/TransportDialect.cpp
mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/3063/commits/68469ffe7bd63f9db1e74a6face32b2ad553650d

mlir/lib/Transport/Transforms/InjectTransportSession.cpp
mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/3063/commits/d495ccf7f0fa9a6b0a4beb0b29c719fa746ba5cf

mlir/test/Transport/InjectTransportSession.mlir
mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/3063/commits/d495ccf7f0fa9a6b0a4beb0b29c719fa746ba5cf

mlir/lib/Transport/Transforms/InjectTransportSession.cpp
mehrdad2m: refactored the code to make the logic clearer per config. Let me know what you think. https://github.com/PennyLaneAI/catalyst/pull/3063/commits/5c4395bca662544760a4089d724df8984c8f4f71

multiphaseCFD (approved):
Thanks @mehrdad2m ! Nice work!

mlir/test/Transport/InjectTransportSession.mlir
mehrdad2m: The key here is identifying a link and since we are only supporting a single controller connecting to multiple copocessors in a star shaped graph, the link is uniquely identifiable by the name of the coprocessor as long as the name of the coprocessors are unique. ofc this need to change if we make the topology more general but it should work for now.

PR #3065, opened 2026-07-24
`compile_decomposition_rules` entry point takes in dictionary to keep track of arg names

paul0403:
[sc-125811]


frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -87,11 +91,11 @@ def get_operator_name(self) -> str:
 
     def get_dynamic_shape_id_format(self) -> str:
         """Return the dynamic shape formatted for GraphOpId."""
-        return f"[{','.join(map(mlir_stringify_type, self.dynamic_shape))}]"
+        return f"[{','.join(self.dynamic_shape.values())}]"
```
kipawaa: We should probably update the formatting of the ID to look like a dictionary (i.e. `{` and `}`)

frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -144,33 +151,42 @@ def compile_decomposition_rules(op_name, op_id, dynamic_shape, wire_lens, static
...
+    device = qp.device("null.qubit", wires=sum(wire_lens.values()))
+    for wire_name, wire_len in wire_lens.items():
+        kwargs[wire_name] = jnp.array(range(wire_len), dtype=int)
+    for arg_name, arg_shape in dynamic_shape.items():
+        kwargs[arg_name] = get_dummy_values_for_container(arg_shape)
+    kwargs.update(static_data)
+
+    _, name_to_resource_ids, decomp_rules = collect_resources_for_op(op_name, kwargs)
+
+    # The static_data was only needed to query the correct decomp rule
```
kipawaa: It's more-so about instantiating the decomposition rules, not finding the correct ones.

frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -144,33 +151,42 @@ def compile_decomposition_rules(op_name, op_id, dynamic_shape, wire_lens, static
...
-        op_name, dummy_dynamic_args, dummy_wires, static_data
-    )
-
+    kwargs = {}
+    device = qp.device("null.qubit", wires=sum(wire_lens.values()))
+    for wire_name, wire_len in wire_lens.items():
+        kwargs[wire_name] = jnp.array(range(wire_len), dtype=int)
+    for arg_name, arg_shape in dynamic_shape.items():
+        kwargs[arg_name] = get_dummy_values_for_container(arg_shape)
+    kwargs.update(static_data)
```
kipawaa: Rather than `kwargs.update(static_data)` here, and deleting after, we can call `collect_resources_for_op(op_name, kwargs | static_data)` below

frontend/catalyst/decomposition/decomposition_rules.py
paul0403: outdated by new code

frontend/catalyst/decomposition/decomposition_rules.py
paul0403: True, I don't know what I was thinking :sweat_smile:

frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -216,7 +234,11 @@ def update_funcop_attributes(op):
 
 
 def compile_decomposition_rules_wrapper(
-    op_name, op_id, dynamic_shape, wire_lens, static_data
+    op_name, op_id, dynamic_shape, wire_lens, static_data, extra_data=None
```
kipawaa, line 259: If we assume that this will only ever be used by QPD, then there isn't any need to accept `extra_data`, though it's not strictly harmful either.

frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -64,34 +70,72 @@ def __init__(self, op: qp.core.Operator2, uid=None):
...
+                *hybrid_args,
+                op_cls=type(self.op),
+                wire_lens=tuple(self.wire_lens.values()),
+                hybrid_lens=tuple(hybrid_lens),
+                hybrid_trees=tuple(hybrid_trees),
+                adjoint=False,
+                n_ctrls=0,
+                static_args=self.op.static_args,
+            )
+            return uid
```
```suggestion
            return generate_uid(
                *tuple(self.op.dynamic_args.values()),  # dynamic args
                *(None,)
                * sum(
                    self.wire_lens.values()
                ),  # non hybrid wires, unused during uid generation, so just give empty values
                *hybrid_args,
                op_cls=type(self.op),
                wire_lens=tuple(self.wire_lens.values()),
                hybrid_lens=tuple(hybrid_lens),
                hybrid_trees=tuple(hybrid_trees),
                adjoint=False,
                n_ctrls=0,
                static_args=self.op.static_args,
            )
```

frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -64,34 +70,72 @@ def __init__(self, op: qp.core.Operator2, uid=None):
...
+            return {}
+
     def get_operator_name(self) -> str:
         """Return the name of the operator."""
         return self.operator_name
 
     def get_dynamic_shape_id_format(self) -> str:
         """Return the dynamic shape formatted for GraphOpId."""
-        return f"[{','.join(map(mlir_stringify_type, self.dynamic_shape))}]"
+        return f"{{{','.join(f"{name}:{shape}" for name, shape in self.dynamic_shape.items())}}}"
```
```suggestion
        return '{' + ','.join(f"{name}:{shape}" for name, shape in self.dynamic_shape.items()) + '}'
```

kipawaa: I think something like this would be a bit nicer to read

kipawaa (commented):
A couple nitpicks but overall looks great, thanks @paul0403 !

frontend/catalyst/decomposition/decomposition_rules.py
paul0403: https://github.com/PennyLaneAI/catalyst/pull/3065/commits/81329446102ff6bd31a9e12dda8cd478a25f05b6

frontend/catalyst/decomposition/decomposition_rules.py
paul0403: https://github.com/PennyLaneAI/catalyst/pull/3065/commits/1b62ff00a7c89e22dc9e08d5e1a76ae2ad278d06

PR #3068, opened 2026-07-24
Add backline frontend

mehrdad2m:
**Context:**
A PennyLane Backline device says where a program's pieces run: a controller and any coprocessors, each with a transport and hardware, and optionally an executor to deploy it. This PR makes sure a PennyLane Backline device is serialized to the catalyst.backline module attribute captured in the IR and compiled through the transport passes.

Because the work was reviewed in a stack, it also carries four PRs that are already merged into it, so the diff here is the union of all five:

[(#3098)](https://github.com/PennyLaneAI/catalyst/pull/3098) Refine the transport dialect, lowering and runtime.
[(#3101)](https://github.com/PennyLaneAI/catalyst/pull/3101) qp.runtime_call lowering and the transport wrappers.
[(#3113)](https://github.com/PennyLaneAI/catalyst/pull/3113) memcpy transport backends and the rdma/ reorganisation.
[(#3117)](https://github.com/PennyLaneAI/catalyst/pull/3117) Frontend sugar: backend selection from hardware, and automatic executor plugins.
Related PennyLane PRs: [9772](https://github.com/PennyLaneAI/pennylane/pull/9772) (the Backline device and Placement) and [9970](https://github.com/PennyLaneAI/pennylane/pull/9970) (runtime_call).

The backline placement frontend: (#3068, #3113, #3117)

- New `frontend/catalyst/backline.py` serializes a Placement into the `catalyst.backline` module attribute with its role, address, port, out_of_process flag, the backend library it should load, and any backend_lib/config hints forwarded from the node's init_args.
- A node's hardware and the placement's transport select the concrete Catalyst backend. PennyLane describes only protocol and hardware, so the backend naming stays a Catalyst detail.
- A placement adds `inject-transport-session`, `lower-decode-to-transport` and `convert-transport-to-llvm` to the compiler's stages. A placement naming a `qec_code` also registers the encoding transforms and adds `convert-qecp-to-llvm`, with the per-QNode encoding chain appended through `device_pass_pipeline`.
- QJIT add the hooks that drive this where the placement is found once per trace and cached, `_configure_backline` picks the pipeline and settles executor addresses before IR generation, `attach_backline_attr` writes the attribute onto the top module, and `launch_executors` deploys at call time.
- Catalyst-launched executors automatically receive `librt_transport.so` and `librt_capi.so`, and the controller's device runtime library is resolved from its PennyLane device. Any user-provided libs can be provided by plugins.

qp.runtime_call (#3101)

- `qp.runtime_call` lowers to an ordinary `catalyst.custom_call`. The shared library exporting the symbol is given through `qp.runtime_declare(..., library=...)`, recorded on the module, collected by `collect_runtime_artifacts` and passed to the linker with the matching `-L/-rpath` flags.
- Opaque buffer arguments and side-effect-only calls with no results are supported for a local call. A dispatched call allows neither.
- `runtime/lib/transport/TransportWrappers.cpp` exposes the transport C API to both call paths: a `__wrapper` entry point for the dispatched case, where the executor resolves the symbol through LLVM ORC and passes one flat argument buffer, and a `__call` entry point for the in-process case, where each argument arrives as its own encoded memref. `runtime/include/TransportABI.h` pins the two layouts, which have to agree with `pennylane/runtime/operands.py`.
- `__catalyst__transport__start_benchmark` runs `iters` rounds back to back and reports each round's round-trip time in nanoseconds.

Transport dialect and lowering (#3098, #3113)

- Ops are renamed for consistency: `transport.barrier` to `transport.await`, `transport.commit_work_item` to `transport.set_message_sizes`, `transport.kick` to `transport.post`, and `data_path` to `transport`. `transport.reply_slot` and `transport.stage_payload` are new, replacing `transport.data_slot` and `transport.write_data_slot`.
- The node attribute's remote flag is renamed `out_of_process` across the dialect, the runtime C API and the frontend. The flag distinguishes a node whose code runs in a separate process from one in the compiling process, which is not the same as running on a different machine.
- `lower-decode-to-transport` no longer heap-allocates for corrections, and the remaining heap allocations on the round path are promoted. `inject-transport-session` and `convert-transport-to-llvm` follow the new ops, and the verifiers gained the cases the new invalid-IR tests cover. The port is a typed `UI16Attr`, and memcpy is accepted as a transport mode.
- Backline now lowers for cross-compiled targets: nested target modules are inlined, and keep_intermediate handles more than one level.

Runtime (#3098, #3101, #3113)

- New memcpy transport: `cpu_controller`, `cpu_coprocessor` and `gpu_coprocessor` sessions over `MemcpyLink`, moving data through host or GPU memory copies instead of RDMA. This lets a user prototype a controller/coprocessor split without `libibverbs` or the matching hardware.
- The RDMA sources move to `runtime/lib/transport/rdma/`, and the code the backends share is hoisted out of `gpu_verbs/` into `common/coproc/` and `common/interface/`: the Steane lookup table, the coprocessor entry points, the wire protocol, error handling, and a `ConfigParser`.
- A backend now chooses where its reply buffer lives, rather than the common layer assuming it. The backends ship with the runtime so a node on another machine can dlopen one by filename from its deployed bundle.
- Coprocessor polling replaces scheduler yields with architecture-specific spin hints, plus best-effort CPU pinning and real-time scheduling on Linux, applied to both the CPU and GPU sessions.
- SSH control-path robustness and clearer errors on the deploy path, including falling back to scp -O on hosts with no SFTP subsystem and reporting that on stderr, where scp's own failure goes.

Build, CI and tests

- CMake builds only the libraries actually used, and `transport_all` names the real per-backend targets. make test runs `every runner_tests_transport*` binary the toolchain produced.
- New `check-transport-gpu.yaml` runs the runtime transport tests on the single-gpu-x64 runner, gated by the gpu label.
- `frontend/test/pytest/test_backline.py` covers serialization, backend and library resolution, pipeline construction, placement discovery and executor realization. `Test_TransportMemcpyCpu.cpp` and `Test_TransportMemcpyGpu.cpp` cover the new backends, and the MLIR tests are updated for the renamed ops with new invalid-IR cases.
PennyLane is bumped to `0.46.0.dev77`.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


frontend/test/pytest/test_backline.py
```diff
@@ -0,0 +1,175 @@
...
+from catalyst.backline import add_transport_passes, backline_pipeline, serialize_backline
+
+pytestmark = pytest.mark.skipif(
+    not hasattr(qp, "backline"), reason="pennylane.backline UI not available"
+)
+
+if hasattr(qp, "backline"):
+    from pennylane.backline import Transport, register_transport
+
+    register_transport("net")(lambda: Transport("net"))
```
josephleekl, line 49: can we put this in a fixture so it doesn't pollute globally?

frontend/catalyst/backline.py
```diff
@@ -0,0 +1,142 @@
...
+def _node_dict(node) -> dict:
+    """Map one backline node (controller/coprocessor) to a ``catalyst.backline`` node dict."""
+    d = {"remote": bool(node.remote)}
+    if node.name is not None:
+        d["name"] = node.name
+    if node.addr is not None:
+        d["peer"] = node.addr
+    if node.port is not None:
+        d["oob_port"] = int(node.port)
+    executor = _realize_executor(node)
```
josephleekl: I think this _node_dict does too much currently - it also does _realize_executor, so the serialize step also launches the realization execution. We should probably split out the realize_executors, so `jit.py` can call it directly, isntead of implicitly through `backline_attr_text`?

josephleekl: so people can do in `jit.py`:
```
if backline is not None:
    from catalyst.backline import backline_attr_text, realize_executors
    realize_executors(backline)
    attr = backline_attr_text(backline)
```

and in here we do 
```
def _node_dict(node) -> dict:
    d = {"remote": bool(node.remote)}
    ...
    executor = getattr(node, "executor", None)
    if isinstance(executor, ExecutorSpec):
        # raise error here if not already realized
        raise RuntimeError(
            f"executor for node {node.name!r} not realized; call realize_executors() first"
        )
    triple = (getattr(executor, "triple", None) if executor is not None else None) or node.triple
    if executor is not None and getattr(executor, "address", None):
        d["address"] = executor.address
    ...
```

frontend/catalyst/jit.py
```diff
@@ -886,6 +886,19 @@ def generate_ir(self):
...
+        device = getattr(self.user_function, "device", None)
+        backline = getattr(device, "backline", None)
+        if backline is not None:
+            from jax.interpreters.mlir import ir
+
+            from catalyst.backline import backline_attr_text
+
+            with mlir_module.context:
+                mlir_module.operation.attributes["catalyst.backline"] = ir.Attribute.parse(
+                    backline_attr_text(backline)
```
josephleekl: as mentioned in another comment, let's separate out the realization step from this single func call

josephleekl (commented):
just some small comments, tag me again after you take care of them

rniczh (approved):

frontend/test/pytest/test_backline.py
```diff
@@ -0,0 +1,186 @@
...
+        name=name,
+        addr="127.0.0.1",
+        port=port,
+        remote=False,
+        coprocessor_fn=fn,
+        init_args={"backend_lib": "backend.so", "config": "cfg", "data_path": "cpu_verbs"},
+    )
+
+
+def test_executor_node_mapping():
```
mehrdad2m: Should we mark these xfail for now since the pennylane PR is not merged yet?

multiphaseCFD: Done here [e79e248](https://github.com/PennyLaneAI/catalyst/pull/3068/commits/e79e248585430239d79b9edcdf8c72696e6ae1a1). and a small fix to apply `use_capture` fixture

frontend/catalyst/backline.py
```diff
@@ -0,0 +1,148 @@
...
+def serialize_backline(backline) -> dict:
+    """Serialize a ``Backline`` into the ``catalyst.backline`` attribute dict."""
+    result = {
+        "transport": backline.transport.name,
+        "controller": _node_dict(backline.controller),
+    }
+    nodes = []
+    for coproc in backline.coprocessors:
+        node = _node_dict(coproc)
+        fn = coproc.coprocessor_fn
```
multiphaseCFD: @mehrdad2m @josephleekl   Is `coprocessor_fn` optional? The runtime binds a default when symbol is empty, so if it's optional, I would suggest:
```
if symbol := getattr(fn, "symbol_name",""):
            node["symbol"] = symbol
```

frontend/catalyst/jit.py
```diff
@@ -886,6 +886,22 @@ def generate_ir(self):
...
+        # If the device declares a backline placement, serialize it onto the top module.
+        device = getattr(self.user_function, "device", None)
+        placement = getattr(device, "placement", None)
+        if placement is not None:
+            from jax.interpreters.mlir import ir
+
+            from catalyst.backline import backline_attr_text, realize_executors
+
+            # Launch any requested-but-unlaunched executors first, so serialization is a pure read.
+            realize_executors(placement)
```
josephleekl: this should probably be somewhere else? probably after the compilation is done

josephleekl: add a TODO here to fix it later

frontend/catalyst/backline.py
```diff
@@ -0,0 +1,244 @@
...
+                passes, ref_pass="convert-catalyst-to-llvm", new_pass="convert-transport-to-llvm"
+            )
+    return stages
+
+
+def backline_pipeline():
+    """Default Catalyst pipeline plus the transport passes."""
+    from catalyst.pipelines import default_pipeline
+
+    return add_transport_passes(default_pipeline())
```
josephleekl: is this the right way to do this in catalyst? @mehrdad2m i don't know well enough. we should also document better at the top of the file what passes we add

frontend/catalyst/backline.py
```diff
@@ -0,0 +1,244 @@
...
+        str: The library path for a local node, or its bare filename for a remote one.
+
+    Raises:
+        ValueError: If no library matches on a local node, naming every directory searched. A
+            backend shipping only one role (``gpu_verbs`` has no controller library) fails here
+            rather than later at ``dlopen``.
+    """
+    names = [f"libcatalyst_transport_{backend}_{role}.{ext}" for ext in _BACKEND_LIB_EXTS]
+    if remote:
+        return names[0]
```
josephleekl, line 120: add a note this only works for linux .so

frontend/catalyst/jit.py
```diff
@@ -886,6 +886,22 @@ def generate_ir(self):
         # Inject Runtime Library-specific functions (e.g. setup/teardown).
         inject_functions(mlir_module, ctx, self.compile_options.seed)
 
+        # If the device declares a backline placement, serialize it onto the top module.
+        device = getattr(self.user_function, "device", None)
+        placement = getattr(device, "placement", None)
```
josephleekl: JB:
this line reads the placement off `self.user_function.device,` which only exists when `@qjit` decorates the QNode directly — with using `@qjit` on a wrapper function that calls the QNode, `.device` is absent so catalyst.backline is never emitted and the placement is silently ignored; all the integration tests here decorate the QNode directly, so it's untested. Can we discover placements from the QNodes seen during tracing (in lower_qnode_to_funcop) instead?

frontend/test/pytest/test_backline.py
```diff
@@ -0,0 +1,478 @@
...
+from catalyst.backline import (
+    _realize_executor,
+    _resolve_backend_lib,
+    add_transport_passes,
+    backline_pipeline,
+    realize_executors,
+    serialize_backline,
+)
+
+pytestmark = pytest.mark.skipif(
```
josephleekl: FYI this might cause failure when the pennylane frontend PR is merged

frontend/catalyst/api_extensions/target.py
```diff
@@ -0,0 +1,214 @@
...
+        str: ``"controller"`` for the controller of a remote backline placement.
+        None: If ``device`` has no backline role.
+    """
+    ctrl = _get_backline_controller(device)
+    if _attr(ctrl, "remote", False):
+        return "controller"
+    return None
+
+
+def get_dispatch(device: Device) -> Optional[RemoteDispatch]:
```
josephleekl: JB:
`_attr` is documented "None-safe getattr" but 3-arg `getattr` only swallows `AttributeError`, and `Executor.address` raises `RuntimeError` when unlaunched — so `get_dispatch()` on a device whose node has an unlaunched executor blows up with a raw `RuntimeError` from inside lowering, while `get_target()` on the same device happily returns (it reads `triple`, which works unlaunched). Worth guarding the address read and giving an error that names the device.

Not sure if we care about this, but just thought worth mentioning

josephleekl (approved):
Thanks @mehrdad2m , just some comments, feel free to merge after resolving them :D

frontend/catalyst/api_extensions/target.py
mehrdad2m: this module is deleted now

frontend/test/pytest/test_backline.py
mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/3068/commits/fe51299d617fdd49122e632499db7a1975625913

frontend/catalyst/jit.py
mehrdad2m: Fixed. The placement is no longer read off the decorated callable. there is anow a `find_placement(callable_, jaxpr, name)` function which falls back to searching the traced jaxpr when the callable carries no device of its own.

frontend/catalyst/backline.py
mehrdad2m: added to the docstring

frontend/catalyst/backline.py
mehrdad2m: Added to the top file docstring. also `add_transport_passes` and `backline_pipeline` no longer exist. The passes are a declarative table instead of an if/elif chain, and there's one function that applies it:

```
_TRANSPORT_PASSES = (
    ("QuantumCompilationStage", "inject-transport-session", None),
    ("BufferizationStage", "lower-decode-to-transport", None),
    ("MLIRToLLVMDialectConversion", "convert-transport-to-llvm", "convert-catalyst-to-llvm"),
) 
```

multiphaseCFD (approved):
LGTM! Thanks @mehrdad2m

frontend/catalyst/backline.py
```diff
@@ -0,0 +1,575 @@
...
+    ``comm_host``/``oob_port`` are coprocessor-only; a controller
+    carries no connection endpoint of its own.
+
+    Args:
+        node: A backline node.
+        role: The node's role, used to resolve its ``backend`` to a library. An explicit
+            ``init_args["backend_lib"]`` path takes precedence over ``backend``.
+    """
+    d: dict = {"remote": bool(_out_of_process(node))}
+    if node.label is not None:
```
josephleekl: @kszenes you probably have to make a catalyst PR to update these? https://github.com/PennyLaneAI/pennylane/pull/10017

josephleekl: also for in_bytes, out_bytes

josephleekl: and backend -> hardware, and for memcpy

mlir/include/Transport/IR/TransportOps.td
```diff
@@ -50,13 +50,13 @@ def Transport_CreateOp : Transport_Op<"create"> {
 
 def Transport_ConnectOp : Transport_Op<"connect"> {
     let summary = "Connect to the peer (blocking).";
-    let arguments = (ins Transport_SessionType:$session, StrAttr:$peer, I16Attr:$oob_port);
```
josephleekl, line 53: could you help update this line https://github.com/PennyLaneAI/catalyst/blob/fe51299d617fdd49122e632499db7a1975625913/mlir/lib/Transport/Transforms/TransportToLLVM.cpp#L139 to i32?

frontend/catalyst/backline.py
```diff
@@ -0,0 +1,575 @@
...
+    searched = ", ".join(str(d) for d in search)
+    raise ValueError(
+        f"no transport backend library for backend={backend!r} role={role!r}: looked for "
+        f"{names[0]} in {searched}. Transport backends are built with -DENABLE_TRANSPORT=ON "
+        f"and are not shipped in the wheel; set {_BACKEND_PATH_ENV} to point at an "
+        "out-of-tree build."
+    )
+
+
+def _out_of_process(node: Node) -> bool:
```
josephleekl, line 136: this doesn't technically return a bool?

frontend/catalyst/backline.py
```diff
@@ -0,0 +1,575 @@
...
+                f"{e}. Pass executor_options= and let the compiler settle the address, or launch "
+                f"the executor before compiling."
+            ) from e
+        if address:  # also rejects "", which would serialize as an empty, unusable address
+            d["address"] = address
+    init = node.init_args or {}
+    if unknown := sorted(set(init) - set(_INIT_KEYS)):
+        raise CompileError(
+            f"backline node has unrecognized init_args {unknown}; the recognized keys are "
+            f"{list(_INIT_KEYS)}. Settings the backend itself interprets go in 'config', as a "
```
josephleekl:
```suggestion
            f"{list(_INIT_KEYS)}. Setting the backend itself interprets go in 'config', as a "
```

frontend/catalyst/from_plxpr/from_plxpr.py
```diff
@@ -307,7 +308,10 @@ def calling_convention(*args):
         gateset = [to_name(op) for op in self.decompose_tkwargs.get("gate_set", [])]
         gateset = list(sorted(gateset))  # consistent ordering for testing
         setattr(qnode, "decompose_gatesets", [gateset])
-    pipelines = (("main", tuple(self._pass_pipeline)),)
+    # The device may require passes of its own, e.g. a backline placement naming a QEC code implies
+    # implicit encoding applied to it. Therefore we append the pass pipeline with the qec lowering
+    # psses.
```
josephleekl:
```suggestion
    # passes.
```

frontend/catalyst/backline.py
```diff
@@ -0,0 +1,575 @@
...
+            f"which puts it on another machine -- so the node's libraries would be looked for in "
+            f"this installation and loaded from that one. Set remote=True on the node, or {fix} to "
+            f"run it here as a subprocess."
+        )
+    if node.remote and not (host or address or preset):
+        raise CompileError(
+            "backline node is remote but its executor_options name neither a 'host' to deploy it "
+            "to nor an 'address' to attach to, which asks for a subprocess of this process on this "
+            "machine. Pass a 'host', or leave remote unset to run the node here."
+        )
```
josephleekl, line 300: cool!

frontend/catalyst/backline.py
```diff
@@ -0,0 +1,575 @@
...
+    Values are plain Python, ready for ``get_mlir_attribute_from_pyval``.
+
+    Args:
+        device: A PennyLane device that may carry a backline placement.
+
+    Returns:
+        dict: Attribute name to value, empty for a device that implies no attributes.
+    """
+    # Only a controller on another machine plays a role: one running in this process needs no
+    # module of its own, so it is not tagged.
```
josephleekl: wait why only controller needs a role?

frontend/catalyst/executor/manager.py
```diff
@@ -240,15 +232,14 @@ def __init__(
     ):
         self.host = host
         self.name = name
-        self._local = local
         self._address = address
+        self._local = not (host or address)
```
josephleekl: can we change to self._remote?

josephleekl (approved):
amazing stuff, thanks @mehrdad2m , just some small comments. please merge when you get a 2nd approval and the comments are addressed

frontend/catalyst/backline.py
mehrdad2m: Because we are not currently capturing any coprocessor code (qnodes are controllers). The coprocessor module is created in the InjectTransportSession pass where it calls the symbols from the precompiled object and we set the role there manually as we create the module.

josephleekl: I see - would be nice if this is written somewhere :D

frontend/catalyst/executor/manager.py
mehrdad2m: Hmm that woulds still be not clear as we actually have three modes. I changed it mode with better documentation 
https://github.com/PennyLaneAI/catalyst/pull/3068/commits/d9f40390c30a6effcbd2c63d0260e434271cffe0

frontend/catalyst/backline.py
mehrdad2m: Thanks fixed it https://github.com/PennyLaneAI/catalyst/pull/3068/commits/54a5a43e76574d841f5fc5a31100cce74faa14b2

mlir/include/Transport/IR/TransportOps.td
mehrdad2m: hmm I don't think we should change it to i32. The main problem is that port is a unsigned i16 ( 0 to 65535). I think this was a hack to get around that but I think the proper fix is to use `UI16Attr` 
fixed it in https://github.com/PennyLaneAI/catalyst/pull/3068/commits/f06c339d02c72d76afa1e7d8d2250e6eda2b112c

mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/3068/commits/10b751e16695b8b9d20c39b1b1e7bfc6008c249a

frontend/catalyst/backline.py
mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/3068/commits/0c51966a10f9380006d2ef5c8fa953378b01f131

PR #3070, opened 2026-07-27
Make resource-analysis extensible via ResourceExtension

sengthai:
**Context:**
Previously, the resource analysis covers standard metric such as gate/qubit count and device information. Also, it included the PBC depth analysis as well.  The resource analysis also required editing `ResourceAnalysis`/`ResourceResult`
directly to add any new metric, and such PBC depth field was hardcoded onto the shared
structure. This made it impossible for new analysis contribute their own resource metrics without touching the core analysis.

In this PR, we introduce the ResourceExtension (data) and ResourceExtensionAnalysis (behaviour) interface. So then the new analysis could create a new classes then inherit from these interfaces.  

To let extensions plug in without `ResourceAnalysis` knowing about them at compile time, we add a `ResourceExtensionRegistry` which is a global registry of extension factories that dialects and plugins register themselves into. `ResourceAnalysisPass` builds its analysis from whatever is currently registered, so PBC's depth extension and any future one is picked up automatically with no change to the pass.

[sc-124299]
[sc-124300]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -23,6 +23,13 @@
   [(#3043)](https://github.com/PennyLaneAI/catalyst/pull/3043)
   [(#3045)](https://github.com/PennyLaneAI/catalyst/pull/3045)
 
+* A CPU transport backend built on libibverbs is added, which implements the controller and coprocessor
+  session roles over RDMA.
+  [(#3062)](https://github.com/PennyLaneAI/catalyst/pull/3062)
+
+* A GPU transport backend is added, which implements the coprocessor session role.
+  [(#3069)](https://github.com/PennyLaneAI/catalyst/pull/3069)
+
 * Catalyst can now cross-compile target nested modules to standalone object files and
   either statically link them into the host program or ship them to an executor for dispatch.
  [(#3033)](https://github.com/PennyLaneAI/catalyst/pull/3033)
@@ -64,7 +71,7 @@
     [(#2915)](https://github.com/PennyLaneAI/catalyst/pull/2915)
 
   - Added `executor.launch_async` and `executor.await` ops, paired by a new `!executor.token` type
-    to the executor dialect along with the necessary lowerings. This allows one to start an async 
+    to the executor dialect along with the necessary lowerings. This allows one to start an async
     kernel on a background host thread and join it later.
     [(#3073)](https://github.com/PennyLaneAI/catalyst/pull/3073)
     [(#3031)](https://github.com/PennyLaneAI/catalyst/pull/3031)
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -116,7 +116,13 @@
   (`any_commuting_depth` / `qubit_disjoint_depth`) per function and lifted loop entry.
   [(#2967)](https://github.com/PennyLaneAI/catalyst/pull/2967)
 
-* The `ResourceAnalysis` now uses a single JSON serializer owned by `ResourceResult`, removing
+* The `resource-analysis` pass now supports pluggable resource metrics through a
+  `ResourceExtension`/`ResourceExtensionAnalysis` interface and a self-registering global
+  registry. Dialects and plugins can contribute additional per-function resource data
+  (such as PBC circuit depth) without modifying the core analysis.
+  [(#3070)](https://github.com/PennyLaneAI/catalyst/pull/3070)
+
+* `ResourceAnalysis` now uses a single JSON serializer owned by `ResourceResult`, removing
   duplicate serialization logic and keeping its output consistent.
   [(#3007)](https://github.com/PennyLaneAI/catalyst/issues/3007)
 
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -116,7 +116,7 @@
   (`any_commuting_depth` / `qubit_disjoint_depth`) per function and lifted loop entry.
   [(#2967)](https://github.com/PennyLaneAI/catalyst/pull/2967)
 
-* `ResourceAnalysis` now uses a single JSON serializer owned by `ResourceResult`, removing
+* The `ResourceAnalysis` now uses a single JSON serializer owned by `ResourceResult`, removing
   duplicate serialization logic and keeping its output consistent.
   [(#3007)](https://github.com/PennyLaneAI/catalyst/issues/3007)
 
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -116,10 +116,43 @@
   (`any_commuting_depth` / `qubit_disjoint_depth`) per function and lifted loop entry.
   [(#2967)](https://github.com/PennyLaneAI/catalyst/pull/2967)
 
-* The `resource-analysis` pass now supports pluggable resource metrics through a
-  `ResourceExtension`/`ResourceExtensionAnalysis` interface and a self-registering global
+* The `resource-analysis` pass now supports pluggable resource metrics through the
+  `ResourceResultExtension`/`ResourceAnalysisExtension` interface and a self-registering global
   registry. Dialects and plugins can contribute additional per-function resource data
   (such as PBC circuit depth) without modifying the core analysis.
+
+  A metric is added by defining a value object, an analysis that fills it, and
+  registering that analysis with the global registry:
+  ```cpp
+  // 1. Per-function value object; shows up in the JSON under name().
+  class TCountExtension : public ResourceResultExtension {
+   public:
+    llvm::StringRef name() const override { return "t_count"; }
+    llvm::json::Value toJson() const override { return tCount; }
+    // Accumulated in collect(), so define how it combines and scales.
+    void mergeWith(const ResourceResultExtension &other, MergeMethod method) override {
+      tCount += static_cast<const TCountExtension &>(other).tCount;  // use `method` for max/min
+    }
+    void multiplyBy(double factor) override { tCount *= factor; }
+    double tCount = 0;
+  };
+
+  // 2. Analysis that fills it, one operation at a time.
+  class TCountAnalysis : public ResourceAnalysisExtensionOf<TCountExtension> {
+   public:
+    llvm::StringRef name() const override { return "t_count"; }
+   protected:
+    void collect(mlir::Operation *op, TCountExtension &ext, bool isAdjoint) override {
+      if (isTGate(op)) ext.tCount += 1;
+    }
+  };
+
+  // 3. Self-register from your dialect or plugin (no core changes needed).
+  REGISTER_RESOURCE_ANALYSIS_EXTENSION(std::make_unique<TCountAnalysis>());
+  ```
+  The metric then appears under its `name()` key (`"t_count"`) in each function's JSON
+  output. Override `analyze(Region&, Ext&, bool)` instead of, or alongside, `collect` to
+  compute a metric per region rather than per operation (as `PBCDepthExtension` does).
   [(#3070)](https://github.com/PennyLaneAI/catalyst/pull/3070)
 
 * `ResourceAnalysis` now uses a single JSON serializer owned by `ResourceResult`, removing
--- a/mlir/cmake/modules/CMakeLists.txt
+++ b/mlir/cmake/modules/CMakeLists.txt
@@ -26,7 +26,7 @@ set(llvm_cmake_builddir "${LLVM_BINARY_DIR}/${LLVM_INSTALL_PACKAGE_DIR}")
 
 # Targets with third party dependencies are not exported.
 get_property(MLIR_EXPORTS GLOBAL PROPERTY MLIR_EXPORTS)
-set(TARGETS_TO_REMOVE nlohmann_json tomlplusplus_tomlplusplus ion-transforms CatalystCompilerDriver PBCUtils QuantumCAPI pbc-transforms)
+set(TARGETS_TO_REMOVE nlohmann_json tomlplusplus_tomlplusplus ion-transforms CatalystCompilerDriver PBCUtils PBCAnalysis QuantumCAPI pbc-transforms)
 list(REMOVE_ITEM MLIR_EXPORTS ${TARGETS_TO_REMOVE})
 export(TARGETS ${MLIR_EXPORTS} FILE ${catalyst_cmake_builddir}/CatalystTargets.cmake)
 
--- a/mlir/include/Catalyst/Analysis/ResourceAnalysisRegistry.h
+++ b/mlir/include/Catalyst/Analysis/ResourceAnalysisRegistry.h
@@ -64,7 +64,7 @@ class ResourceAnalysisRegistry {
 #define RES_EXT_CONCAT_(a, b) a##b
 #define RES_EXT_CONCAT(a, b) RES_EXT_CONCAT_(a, b)
 #define REGISTER_RESOURCE_ANALYSIS_EXTENSION(CTOR_EXPR)                                            \
-    static const int LLVM_ATTRIBUTE_UNUSED _resExtReg_##__COUNTER__ = []() {                       \
+    static const int LLVM_ATTRIBUTE_UNUSED RES_EXT_CONCAT(_resExtReg_, __COUNTER__) = []() {       \
         ::catalyst::ResourceAnalysisRegistry::get().add([] { return CTOR_EXPR; });                 \
         return 0;                                                                                  \
     }()
--- a/mlir/include/Catalyst/Analysis/ResourceAnalysisRegistry.h
+++ b/mlir/include/Catalyst/Analysis/ResourceAnalysisRegistry.h
@@ -64,7 +64,7 @@ class ResourceAnalysisRegistry {
 #define RES_EXT_CONCAT_(a, b) a##b
 #define RES_EXT_CONCAT(a, b) RES_EXT_CONCAT_(a, b)
 #define REGISTER_RESOURCE_ANALYSIS_EXTENSION(CTOR_EXPR)                                            \
-    static const int LLVM_ATTRIBUTE_UNUSED RES_EXT_CONCAT(_resExtReg_, __COUNTER__) = []() {       \
+    static const int LLVM_ATTRIBUTE_UNUSED _resExtReg_##__COUNTER__ = []() {       \
         ::catalyst::ResourceAnalysisRegistry::get().add([] { return CTOR_EXPR; });                 \
         return 0;                                                                                  \
     }()
--- a/mlir/include/Catalyst/Analysis/ResourceAnalysisRegistry.h
+++ b/mlir/include/Catalyst/Analysis/ResourceAnalysisRegistry.h
@@ -24,25 +24,25 @@
 #include "llvm/Support/Compiler.h"
 #include "llvm/Support/WithColor.h"
 
-#include "Catalyst/Analysis/ResourceExtension.h"
+#include "Catalyst/Analysis/ResourceResultExtension.h"
 
 namespace catalyst {
 
-// Global registry of ResourceExtensionAnalysis providers.
-// Dialects / plugins self-register via the ResourceExtensionRegistry::add method,
-// or the REGISTER_RESOURCE_EXTENSION macro.
-class ResourceExtensionRegistry {
+// Global registry of ResourceAnalysisExtension providers.
+// Dialects / plugins self-register via the ResourceAnalysisRegistry::add method,
+// or the REGISTER_RESOURCE_ANALYSIS_EXTENSION macro.
+class ResourceAnalysisRegistry {
   public:
-    using ExtensionProvider = std::function<std::unique_ptr<ResourceExtensionAnalysis>()>;
+    using ExtensionProvider = std::function<std::unique_ptr<ResourceAnalysisExtension>()>;
 
-    static ResourceExtensionRegistry &get();
+    static ResourceAnalysisRegistry &get();
 
     void add(ExtensionProvider extensionProvider)
     {
         std::string name = extensionProvider()->name().str();
         if (llvm::is_contained(names, name)) {
-            llvm::WithColor::warning()
-                << "ResourceExtension '" << name << "' is already registered; ignoring duplicate\n";
+            llvm::WithColor::warning() << "ResourceAnalysisExtension '" << name
+                                       << "' is already registered; ignoring duplicate\n";
             return;
         }
         names.push_back(std::move(name));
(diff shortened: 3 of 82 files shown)


mlir/cmake/modules/CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ set(llvm_cmake_builddir "${LLVM_BINARY_DIR}/${LLVM_INSTALL_PACKAGE_DIR}")
 
 # Targets with third party dependencies are not exported.
 get_property(MLIR_EXPORTS GLOBAL PROPERTY MLIR_EXPORTS)
-set(TARGETS_TO_REMOVE nlohmann_json tomlplusplus_tomlplusplus ion-transforms CatalystCompilerDriver PBCUtils QuantumCAPI pbc-transforms)
+set(TARGETS_TO_REMOVE nlohmann_json tomlplusplus_tomlplusplus ion-transforms CatalystCompilerDriver PBCUtils PBCAnalysis QuantumCAPI pbc-transforms)
```
dime10, line 29: Is there a third-party dependency that means this shouldn't be exported?

mlir/lib/PBC/Analysis/PBCDepthAnalysis.cpp
```diff
@@ -0,0 +1,98 @@
...
+    ext.setDepth(layerContext.computePBCDepth(block));
+}
+
+void registerPBCResourceExtensions()
+{
+    static const bool once = [] {
+        ResourceExtensionRegistry::get().add([] { return std::make_unique<PBCDepthAnalysis>(); });
+        return true;
+    }();
+    (void)once;
```
dime10: minor: slightly cleaner I think
```suggestion
    [[maybe_unused]] static const bool once = [] {
        ResourceExtensionRegistry::get().add([] { return std::make_unique<PBCDepthAnalysis>(); });
        return true;
    }();
```

mlir/include/Catalyst/Analysis/ResourceExtensionRegistry.h
```diff
@@ -0,0 +1,68 @@
...
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/Support/Compiler.h"
+
+#include "Catalyst/Analysis/ResourceExtension.h"
+
+namespace catalyst {
+
+/// Global registry of ResourceExtensionAnalysis providers.
+/// Dialects / plugins self-register
+// (e.g. pbc::registerPBCResourceExtensions, or REGISTER_RESOURCE_EXTENSION in a shared library).
```
dime10:
```suggestion
/// Dialects / plugins self-register via the ResourceExtensionRegistry::add method,
/// or the REGISTER_RESOURCE_EXTENSION macro.
```

mlir/include/Catalyst/Analysis/ResourceExtensionRegistry.h
```diff
@@ -0,0 +1,68 @@
...
+class ResourceExtensionRegistry {
+  public:
+    using ExtensionProvider = std::function<std::unique_ptr<ResourceExtensionAnalysis>()>;
+
+    static ResourceExtensionRegistry &get();
+
+    void add(ExtensionProvider extensionProvider)
+    {
+        llvm::StringRef name = extensionProvider()->name();
+        assert(!llvm::is_contained(names, name) && "ResourceExtension name already registered");
```
dime10: Should we crash when an extension double registers, or silently skip? Not sure what the better approach is in this instance. Since this is an external interface, we might want to be careful / more lenient when validating input, as long as it doesn't cause issues.

mlir/include/Catalyst/Analysis/ResourceExtensionRegistry.h
```diff
@@ -0,0 +1,68 @@
...
+  private:
+    llvm::SmallVector<ExtensionProvider> extensionProviders;
+
+    // parallel to `extensionProviders`, avoids re-invoking them
+    llvm::SmallVector<std::string> names;
+};
+
+/// Self-register a ResourceExtensionAnalysis factory.
+// Example:
+// REGISTER_RESOURCE_EXTENSION(std::make_unique<ABCAnalysis>());
```
dime10: Is the inconsistent use of 2 vs 3 slashes intentional?

mlir/include/Catalyst/Analysis/ResourceExtensionRegistry.h
```diff
@@ -0,0 +1,68 @@
...
+#define RES_EXT_CONCAT_(a, b) a##b
+#define RES_EXT_CONCAT(a, b) RES_EXT_CONCAT_(a, b)
+#define REGISTER_RESOURCE_EXTENSION(CTOR_EXPR)                                                     \
```
dime10: Does the macro exist for convenience? I'm just wondering because it is fairly easy to call the add function, so I don't know if we needed it just for that purpose 🤔

mlir/include/Catalyst/Analysis/ResourceResult.h
```diff
@@ -23,13 +23,15 @@
...
+#include "Catalyst/Analysis/ResourceExtension.h"
+
 namespace catalyst {
 
 // ResourceResult holds the resource counts for a single function.
 // It mirrors the Python-side ResourcesResult from specs_collector.py.
 struct ResourceResult {
     // method for merging two ResourceResult values
-    enum class MergeMethod { Sum, Max, Min };
+    using MergeMethod = catalyst::MergeMethod;
```
dime10, line 36: The merge method enum seems more core to the analysis or resource result than an extension, since it is used regardless of whether there are any extensions or not. So it's a bit odd to have the definition in the extension header 🤔

mlir/include/Catalyst/Analysis/ResourceExtension.h
```diff
@@ -0,0 +1,88 @@
...
+#include "llvm/ADT/StringRef.h"
+#include "llvm/Support/JSON.h"
+#include "mlir/IR/Operation.h" // Operation, Region
+
+namespace catalyst {
+
+enum class MergeMethod { Sum, Max, Min };
+
+/// Value object for an optional resource metric. Lives in ResourceResult.
+class ResourceExtension {
```
dime10: The naming of these types is a bit confusing, but if I understand correctly
- `ResourceExtension` is an interface for extending `ResourceResult`
- `ResourceExtensionAnalysis` is an interface for extending `ResourceAnalysis`

If this is correct, then I think `ResourceResultExtension` and `ResourceAnalysisExtension` would be much clearer

mlir/include/Catalyst/Analysis/ResourceExtension.h
```diff
@@ -0,0 +1,88 @@
...
+    virtual void mergeWith(const ResourceExtension &other, MergeMethod mergeMethod) {}
+    virtual void multiplyBy(int64_t factor) {}
+};
+
+template <typename Ext> class ResourceExtensionAnalysisOf;
+
+/// Owned by ResourceAnalysis for the duration of a run; writes into per-result data.
+/// Do not inherit directly, use ResourceExtensionAnalysisOf<Ext>.
+class ResourceExtensionAnalysis {
+    template <typename Ext> friend class ResourceExtensionAnalysisOf;
```
dime10: Can you explain the design behind these two classes and how they relate to each other?

mlir/include/PBC/Analysis/CMakeLists.txt
dime10, line 1: Is this file doing anything?

mlir/lib/Catalyst/Transforms/ResourceAnalysisPass.cpp
```diff
@@ -58,7 +56,10 @@ struct ResourceAnalysisPass : public impl::ResourceAnalysisPassBase<ResourceAnal
 
     void runOnOperation() final
     {
-        auto &analysis = getAnalysis<ResourceAnalysis, ModuleOp>();
+        pbc::registerPBCResourceExtensions();
```
dime10: Why are we explicitly registering the PBC extension here, doesn't that defeat the whole point of the extension system? This means in the analysis pass we already need to know about every extension. I would imagine that instead the registration happens once in the PBC dialect somewhere.

mlir/lib/Catalyst/Analysis/ResourceResult.cpp
```diff
@@ -107,6 +118,11 @@ void ResourceResult::multiplyByScalar(double scalar)
     }
 
     numAllocQubits *= scalar;
+    numArgQubits *= scalar;
```
dime10: Why are you adding this back in? Didn't we just remove it?

mlir/include/PBC/Analysis/PBCDepthAnalysis.h
```diff
@@ -0,0 +1,59 @@
...
+
+#include "Catalyst/Analysis/ResourceExtension.h"
+
+namespace catalyst {
+namespace pbc {
+
+/// PBC depth data stored in ResourceResult::extensions.
+class PBCDepthExtension : public ResourceExtension {
+  public:
+    llvm::StringRef name() const override { return "depth"; }
```
dime10, line 32: this is not very unique, there could be many different depths 🤔

dime10 (dismissed):
Great, thanks for putting this together @sengthai 💯

mlir/lib/PBC/Analysis/PBCDepthAnalysis.cpp
sengthai: Thanks! looks better~

mlir/cmake/modules/CMakeLists.txt
sengthai: It is not directly depend, but it links to PBCUtils which uses Stim.

mlir/include/Catalyst/Analysis/ResourceExtensionRegistry.h
sengthai: Not any reason. I updated to use 2 slashes. 😅

[e786412](https://github.com/PennyLaneAI/catalyst/pull/3070/commits/e7864128c6b4f6d6843eae6b49b6d53220e86819)

mlir/include/Catalyst/Analysis/ResourceExtensionRegistry.h
sengthai: The Macro is used for analysis that implement as plugin, so extension's .cpp file registers itself just by being linked it. Doing so, user won't need to added into the ResourceAnalysisPass like in-tree PBC Depth analysis. 
https://github.com/PennyLaneAI/catalyst/blob/e7864128c6b4f6d6843eae6b49b6d53220e86819/mlir/lib/Catalyst/Transforms/ResourceAnalysisPass.cpp#L59-L60

mlir/include/PBC/Analysis/CMakeLists.txt
sengthai: Initially, i wanted to implement the in td, but since we implement in cpp, so we should remove that.
deleted! [8a79927](https://github.com/PennyLaneAI/catalyst/pull/3070/commits/8a79927ea0c17d93cbe7630e5f5fccacbd3a38fd)

mlir/include/Catalyst/Analysis/ResourceExtension.h
sengthai: Actually, the purpose of this design ensures that user won't extend by inheriting directly from `ResourceExtensionAnalysis`. Also, we want to enforce the user to explicitly implement the ResourceExtensions and pass it to the analysis class, such this:
https://github.com/PennyLaneAI/catalyst/blob/e7864128c6b4f6d6843eae6b49b6d53220e86819/mlir/include/PBC/Analysis/PBCDepthAnalysis.h#L47-L48

If we merge these classes, we may need to define it ResourceExtensions in ResourceAnalysis as well, which seems little bit messy. 
https://github.com/PennyLaneAI/catalyst/blob/e7864128c6b4f6d6843eae6b49b6d53220e86819/mlir/include/Catalyst/Analysis/ResourceAnalysis.h#L34-L35

mlir/lib/Catalyst/Transforms/ResourceAnalysisPass.cpp
sengthai: We introduced two ways for extension here. Once, user can added analysis via its function in resource analysis pass and another way, user can add via REGISTER_RESOURCE_EXTENSION. So, for _official extended analysis_ , we would recommend to added in that, so then we can track what analysis passes are used. For _external extended analysis_ (e.g visa), we can use macro which won't needed to edit the mainstream catalyst.

mlir/lib/Catalyst/Analysis/ResourceResult.cpp
sengthai: My bad, I mistakenly bring it back during merge.
Fixed, [a318f62](https://github.com/PennyLaneAI/catalyst/pull/3070/commits/a318f62f1a52f4be467420bd4d40fa56394bc418)

mlir/include/PBC/Analysis/PBCDepthAnalysis.h
sengthai: You are right. The depth is so generic here. This will very soon updated. Here, we just updated to satisfy the PL frontend where it hardcore parse the `depth`. 
https://github.com/PennyLaneAI/pennylane/pull/9841/changes#diff-52ee52034354f266aed81584e9691917e88492d215cb2d8be1e8aa1eb76261d0L117

sengthai: I updated that in this branch, will opened PR shortly. 

https://github.com/PennyLaneAI/catalyst/commit/f4828c2e537b2d9717419acf370a8bfc087d4080

mlir/include/Catalyst/Analysis/ResourceResult.h
sengthai: Make sense!  I moved back to ResultResult file. [6e92dc8](https://github.com/PennyLaneAI/catalyst/pull/3070/commits/6e92dc837e5a3b0a53ba31903acd772ee9ff7552)

mlir/include/Catalyst/Analysis/ResourceExtension.h
sengthai: > * ResourceExtension is an interface for extending ResourceResult
> * ResourceExtensionAnalysis is an interface for extending ResourceAnalysis

Yes right. Actually they are a composition-based interface. The `ResourceExtension` is the property that lives in `ResourceResult`. And, ResourceAnalysis holds lists of ResourceExtensionAnalysis. Initially, I also used ResourceResultExtension but it seemed that sound like the ResourceResultExtension inherit from ResourceResult. 🤔

mlir/include/Catalyst/Analysis/ResourceExtensionRegistry.h
sengthai: I think warning and skipping is preferable here. we can keep the first registered, emit warning with duplicate extension name, and ignore the later registration. This keeps the external interface lenient without siliendly hiding accidental name collisions. what do you think?

sengthai: [1063395](https://github.com/PennyLaneAI/catalyst/pull/3070/commits/10633953066da81e7f0edb9e5bd62c025e22683b)

mlir/include/Catalyst/Analysis/ResourceAnalysisRegistry.h
```diff
@@ -0,0 +1,72 @@
...
+    llvm::SmallVector<std::string> names;
+};
+
+// Self-register a ResourceExtensionAnalysis factory.
+// Example:
+// REGISTER_RESOURCE_EXTENSION(std::make_unique<ABCAnalysis>());
+#define RES_EXT_CONCAT_(a, b) a##b
+#define RES_EXT_CONCAT(a, b) RES_EXT_CONCAT_(a, b)
+#define REGISTER_RESOURCE_EXTENSION(CTOR_EXPR)                                                     \
+    static const int LLVM_ATTRIBUTE_UNUSED RES_EXT_CONCAT(_resExtReg_, __COUNTER__) = []() {       \
```
rniczh, line 67:
```suggestion
    static const int LLVM_ATTRIBUTE_UNUSED _resExtReg_##__COUNTER__ = []() {       \
```

rniczh (approved):
💯

mlir/include/Catalyst/Analysis/ResourceExtension.h
sengthai: I realized it would be better to rename them. 🙌
[01c38f4](https://github.com/PennyLaneAI/catalyst/pull/3070/commits/01c38f4ad478234fa3863cef2981327cc68f3dd9)

mlir/include/Catalyst/Analysis/ResourceAnalysisRegistry.h
```diff
@@ -0,0 +1,72 @@
...
+#include "llvm/Support/Compiler.h"
+#include "llvm/Support/WithColor.h"
+
+#include "Catalyst/Analysis/ResourceResultExtension.h"
+
+namespace catalyst {
+
+// Global registry of ResourceAnalysisExtension providers.
+// Dialects / plugins self-register via the ResourceAnalysisRegistry::add method,
+// or the REGISTER_RESOURCE_ANALYSIS_EXTENSION macro.
```
lazypanda10117, line 33: Who actually uses this macro? This seems unnecessary?

mlir/include/Catalyst/Analysis/ResourceResultExtension.h
```diff
@@ -0,0 +1,92 @@
...
+
+// Value object for an optional resource metric. Lives in ResourceResult.
+class ResourceResultExtension {
+  public:
+    virtual ~ResourceResultExtension() = default;
+    virtual llvm::StringRef name() const = 0;
+    virtual llvm::json::Value toJson() const = 0;
+
+    virtual void mergeWith(const ResourceResultExtension &other, MergeMethod mergeMethod) {}
+    virtual void multiplyBy(int64_t factor) {}
```
lazypanda10117: Is multiplyBy something that needs to be in ResourceResultExtension?

What is the purpose of this object?

mlir/include/Catalyst/Analysis/ResourceAnalysisRegistry.h
```diff
@@ -0,0 +1,72 @@
...
+
+    // parallel to `extensionProviders`, avoids re-invoking them
+    llvm::SmallVector<std::string> names;
+};
+
+// Self-register a ResourceAnalysisExtension factory.
+// Example:
+// REGISTER_RESOURCE_ANALYSIS_EXTENSION(std::make_unique<ABCAnalysis>());
+#define RES_EXT_CONCAT_(a, b) a##b
+#define RES_EXT_CONCAT(a, b) RES_EXT_CONCAT_(a, b)
```
lazypanda10117, line 65: When and how do we use this macro: RES_EXT_CONCAT?

mlir/include/Catalyst/Analysis/ResourceAnalysisRegistry.h
sengthai: The external analysis pass (e.g visa depth analysis) is required to use this.

mlir/include/Catalyst/Analysis/ResourceResultExtension.h
sengthai: Yes, it is needed for the analysis pass that wants to calculated the counts in static for-loop.

lazypanda10117: I guess my question is whether this function is  foundational enough to be considered inside this class. Is this something that should be implemented by every ResourceResultExtension?

mlir/include/Catalyst/Analysis/ResourceAnalysisRegistry.h
sengthai: Ohh right, this should be called in function. fixed: 
https://github.com/PennyLaneAI/catalyst/blob/380a52bdbb0a08b3bb6b4f901d063e55aaab21eb/mlir/include/Catalyst/Analysis/ResourceAnalysisRegistry.h#L67-L68

this is used to concatenate the name of pass and the increment counter.

sengthai: [380a52b](https://github.com/PennyLaneAI/catalyst/pull/3070/commits/380a52bdbb0a08b3bb6b4f901d063e55aaab21eb)

mlir/include/Catalyst/Analysis/ResourceResultExtension.h
sengthai: Good question! Yes, it belongs on the base class, but it should not be mandatory for every subclass. If an extension carries a count-like metric (e.g PBC depth which is scaled by for-loop) then it must be implemented the `multiplyBy`. However, this function has no meaningful for metric like finding maximum of group ID (e.g vISA group ID). So, i think we keep this function here but should not make it as pure virtual.

sengthai: [d09a212](https://github.com/PennyLaneAI/catalyst/pull/3070/commits/d09a21200f0ea01845f7f39985198558e63085ae)

doc/releases/changelog-dev.md
```diff
@@ -99,6 +99,12 @@
   (`any_commuting_depth` / `qubit_disjoint_depth`) per function and lifted loop entry.
   [(#2967)](https://github.com/PennyLaneAI/catalyst/pull/2967)
 
+* The `resource-analysis` pass now supports pluggable resource metrics through a
```
lazypanda10117: I think it would be really helpful to have an example here on how to register external analysis pass.

lazypanda10117 (approved):

doc/releases/changelog-dev.md
sengthai: Done! [9504a69](https://github.com/PennyLaneAI/catalyst/pull/3070/commits/9504a69ac0f64bd9d787c7ed24c854a2f9a7fa84)

mlir/lib/Catalyst/Analysis/ResourceResult.cpp
```diff
@@ -84,9 +86,18 @@ void ResourceResult::mergeWith(const ResourceResult &other, MergeMethod method)
 
     hasBranches = hasBranches || other.hasBranches;
     hasDynLoop = hasDynLoop || other.hasDynLoop;
+
+    for (auto &m : extensions) {
+        for (const auto &o : other.extensions) {
```
jzaia18: Is it guaranteed that these extensions will be in the same order? (Or if not, can we make it so that they will be?). It seems wasteful to have a double nested loop to search for extensions here rather than just iterate through the extensions pairwise.

Not a huge deal, since we expect the number of extensions to be small anyway

sengthai: Good point! These both should have extensions in the same orders. 

here we go
https://github.com/PennyLaneAI/catalyst/blob/026fad3cfa6041f03fc3fe5e31cb30a7f9196abf/mlir/lib/Catalyst/Analysis/ResourceResult.cpp#L90-L93

mlir/lib/PBC/Analysis/PBCDepthAnalysis.cpp
```diff
@@ -0,0 +1,98 @@
...
+        block->getParentOp()->getContext(), [](Diagnostic &diag) {
+            if (diag.getSeverity() == DiagnosticSeverity::Error &&
+                diag.str().find("worst-case depth") != std::string::npos) {
+                return success();
+            }
+            return failure();
+        });
+
+    PBCLayerContext layerContext;
+    ext.setDepth(layerContext.computePBCDepth(block));
```
jzaia18, line 86: One concern here is that every time `analyze` is called we will call `computePBCDepth` which is effectively recursive and gets the depth of the full region including any nested regions. But it seems like we actually call `analyze` on every region that is encountered (see [Line 382 of ResourceAnalysis.cpp](https://github.com/PennyLaneAI/catalyst/pull/3070/changes#diff-13e92b41ee7a126037399bf869677f8518ea3a498d4360ee9543c6456d5ac2baR382)). I think this will cause us to waste a large amount of computation

sengthai: Good point! we can avoid this redundant computation by ignoring for-loop handling in `computePBCDepth`. However, i want do that in next PR after we remove ppm_spec (https://github.com/PennyLaneAI/catalyst/pull/3081) since ppm_spec test's relied on that for-loop handling.

sengthai: I put in the shortcut there for this one. https://app.shortcut.com/xanaduai/story/126884/optimize-pbcdepthanalysis-avoid-redundant-recursive-depth-recomputation-across-nested-regions

PR #3073, opened 2026-07-28
add launch_async and await to executor dialect

mehrdad2m:
**Context:**
The executor dialect currently only has a blocking launch. But for in some cases some entries cannot complete until a later launch runs. for example one that waits on a connection the next dispatch establishes. Dispatching such an entry with executor.launch deadlocks the host on work it is itself responsible for triggering.

Added `!executor.token`, `executor.launch_async`, and `executor.await` to the executor dialect, their llvm lowering and runtime CAPI. 

**Benefits:**

**Possible Drawbacks:**
The lowering of the code (specifically the placement of the await ops) is currently somewhat hacky. Adding counterpart ops to catalyst dialect would make things easier so the lowering is 1-to-1. 

**Related GitHub Issues:**


runtime/lib/executor/ExecutorRuntime.cpp
maliasadi, line 1: Could you add docstrings?

runtime/lib/executor/ExecutorRuntime.cpp
```diff
@@ -34,15 +36,18 @@ struct ExecutorEntry {
...
-            session = nullptr;
-        }
-    }
+    ~ExecutorEntry();
 };
 
+struct AsyncWorker {
+    int64_t session;
+    std::thread thread;
+};
```
maliasadi, line 46: What I don't fully understand here is the robustness of this impl to handle errors (we can probably discuss it offline). But if a thread fails, the worker currently raises a RT_Error which breaks the whole system as well as calls to `await()`. I think a better way to handle errors here would be by caching error messages here and then let `await()` rethrows on the main thread. `std::thread::join()` establishes the happens-before order, so no extra sync will be needed to read the cached error messages.

maliasadi: add `error` and then propagate it within `await()`
```suggestion
struct AsyncWorker {
    int64_t session;
    std::thread thread;
    std::shared_ptr<std::string> error;
};
```

doc/releases/changelog-dev.md
```diff
@@ -36,6 +36,11 @@
     into the remote JIT.
     [(#2915)](https://github.com/PennyLaneAI/catalyst/pull/2915)
 
+  - Added `executor.launch_async` and `executor.await` ops, paired by a new `!executor.token` type
+    to the executor dialect along with the necessary lowerings. This allows one to start a async
```
```suggestion
    to the executor dialect along with the necessary lowerings. This allows one to start an async 
```

maliasadi (commented):
Nice work @mehrdad2m!

runtime/lib/executor/ExecutorRuntime.cpp
```diff
@@ -309,4 +357,58 @@ void __catalyst__executor__launch(int64_t session, const char *entry_symbol, con
...
+        }
+    };
+
+    std::lock_guard<std::mutex> reg(g_async_mu);
+    int64_t token = g_next_token++;
+    g_async_workers.emplace(token, AsyncWorker{session, std::thread(std::move(work))});
+    return token;
+}
+
+void __catalyst__executor__await(int64_t token)
```
maliasadi, line 463: What'd be the behaviour of `await` on an unknown/already-awaited token? Do we need a RT failure or deal with it silently or with a warning

runtime/lib/executor/ExecutorRuntime.cpp
mehrdad2m: Thanks for the suggestion. I fixed it in https://github.com/PennyLaneAI/catalyst/pull/3073/commits/26da9c69e1a638ecb1ff2145d81ae4a3285d3a56

runtime/lib/executor/ExecutorRuntime.cpp
mehrdad2m: I am not sure what is the best way to handle it. For now I raised a warning but let me know if you prefer to raise a RT failure https://github.com/PennyLaneAI/catalyst/pull/3073/commits/26da9c69e1a638ecb1ff2145d81ae4a3285d3a56

runtime/lib/executor/ExecutorRuntime.cpp
mehrdad2m: Done!

runtime/lib/executor/ExecutorRuntime.cpp
maliasadi: Raising a warning seems good! We can revisit this later if needed

runtime/lib/executor/ExecutorRuntime.cpp
```diff
@@ -89,6 +137,12 @@ std::shared_ptr<ExecutorEntry> find_entry_by_handle(int64_t session)
 
 extern "C" {
 
+/**
+ * @brief Open (or reuse) an executor session for `addr`.
+ *
+ * @param addr The executor endpoint to connect to. Must be non-empty.
+ * @return int64_t A non-zero session handle for use with the other executor CAPI calls.
+ */
 int64_t __catalyst__executor__open(const char *addr)
```
maliasadi, line 146: Could the session handle be neg?

```suggestion
 * @return uint64_t A non-zero session handle for use with the other executor CAPI calls.
 */
uint64_t __catalyst__executor__open(const char *addr)
```

runtime/lib/executor/ExecutorRuntime.cpp
```diff
@@ -140,6 +194,14 @@ int64_t __catalyst__executor__open(const char *addr)
...
 
+/**
+ * @brief Load a binary (object or asset) into the session, once per path.
+ *
+ * @param session The i64 handle of the session, from __catalyst__executor__open.
+ * @param path The filesystem path of the binary to load. Empty is a no-op.
+ * @param format The binary format tag: 0 for an object file, 1 for an asset.
+ * @return int64_t 0 on success.
+ */
 int64_t __catalyst__executor__send_binary(int64_t session, const char *path, uint32_t format)
```
maliasadi, line 205: Does it make sense to drop `binary` from the name of this instruction (`__executor__send`)?

maliasadi (approved):
🤠

runtime/lib/executor/ExecutorRuntime.cpp
mehrdad2m: yeah I think binary is not most accurate. Would `__executor__send` be too generic though? how about `__executor__send__file`? in any case, this is following the executor.sed_binary op from the dialect. So I would prefer if the renaming can be done after the whole stack is merged in a separate PR, so we can easily change it across the stack.

runtime/lib/executor/ExecutorRuntime.cpp
mehrdad2m: Yeah good point! I think this is also something we can modify, but again this would likely disrupt other PRs. Would it be ok for if we make issue pages to track these and address them in a separate PR?

rniczh (approved):
Awesome 👍

runtime/lib/executor/ExecutorRuntime.cpp
maliasadi: Agree -- let's add a ticket to the backlog. also `__executor__send__file` :+1:

runtime/lib/executor/ExecutorRuntime.cpp
maliasadi: these are not blockers -- just a few more suggestions that can be done as follow-ups

PR #3080, opened 2026-07-29
Forward Catalyst subprocess diagnostics to Python

sengthai:
**Context:**
Successful Catalyst subprocess calls capture both output streams but previously returned only `stdout`, silently discarding warnings and other diagnostics written to stderr. This prevented warnings from appearing in PennyLane workflows such as `qp.specs`. 

We also has `verbose` but this will output like system log which are not necessary for end users.

Here, we forward captured Catalyst `stderr` while preserving `stdout` as normal. When Python’s stderr is attached to a terminal, pass `--color` to Catalyst so LLVM diagnostics retain their colors. Note that to make this work PL needs to update to pass the `stderr_return=True`.


mehrdad2m:
Thanks @sengthai. I noticed this is very similar to what https://github.com/PennyLaneAI/catalyst/blob/0d5db5fba2b7509008cf30e0f880860fd0f08e75/frontend/catalyst/python_interface/inspection/xdsl_conversion.py#L89-L99 does for the `inspection` module in `python_interface`. Do you think  it make sense to unify them and follow the same approach of making it optional?

sengthai:
@mehrdad2m Thanks and good to know! I made a change and mirror the xdsl's approach [b7a204f](https://github.com/PennyLaneAI/catalyst/pull/3080/commits/b7a204ffa55bb1eb80d6ae1b61973cc04a6cfe2e).  And I just realized that qjit has `verbose` flag to print those diagnostic from Catalyst. So, here I enable this in qp.specs. 
https://github.com/PennyLaneAI/pennylane/pull/9938

rniczh (approved):

mehrdad2m (approved):
Thanks @sengthai

PR #3088, opened 2026-08-03
Add catalyst executor server side to catalyst

mehrdad2m:
**Context:**
Catalyst already has the host side of the remote-executor connection which is `rt_executor` that exposes the C ABI a compiled program calls, and `catalyst_executor_session` which is the ORC client that holds the connection. 

This PR adds the node end the ORC server that receives objects, maps them and calls them.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


runtime/lib/executor/executor.cpp
```diff
@@ -0,0 +1,452 @@
...
+            ExitOnErr.setBanner("CatalystExecutor[" + GLabel + "]: ");
+            {
+                std::unique_ptr<SimpleRemoteEPCServer> Server =
+                    ExitOnErr(SimpleRemoteEPCServer::Create<FDSimpleRemoteEPCTransport>(
+                        setupCatalystServer, CSock, CSock));
+                ExitOnErr(Server->waitForDisconnect());
+            }
+            std::fprintf(stderr, "[%s] client disconnected, exiting\n", GLabel.c_str());
+            std::fprintf(stderr, "[%s] executor ready, waiting for next connection\n",
+                         Label.c_str());
```
josephleekl: it exits immediately, and doesn't wait for anything?

runtime/lib/executor/executor.cpp
```diff
@@ -0,0 +1,452 @@
...
+            llvm::formatv("kernel entry function must have 2 arguments, got {0}", args.size()));
+    }
+    using pyface_t = void (*)(void *, void *);
+    fn.toPtr<pyface_t>()(args[0].toPtr<void *>(), args[1].toPtr<void *>());
+}
+
+int32_t _catalyst_remote_store_asset(std::vector<char> bytes, std::string name)
+{
+    namespace fs = std::filesystem;
+    fs::path dst = fs::temp_directory_path() / "catalyst-assets" / fs::path(name).filename();
```
josephleekl: should we put it in a subdir with PID so multiple writes won't overwrite/collide?

runtime/lib/executor/executor.cpp
```diff
@@ -0,0 +1,452 @@
...
+    std::error_code ec;
+    fs::create_directories(dst.parent_path(), ec);
+    if (ec) {
+        std::fprintf(stderr, "catalyst-executor: mkdir %s failed: %s\n", dst.parent_path().c_str(),
+                     ec.message().c_str());
+        return -1;
+    }
+
+    // wrirte to file
```
josephleekl:
```suggestion
    // write to file
```

runtime/lib/executor/executor.cpp
```diff
@@ -0,0 +1,452 @@
...
+    FILE *f = std::fopen(dst.c_str(), "wb");
+    if (!f) {
+        std::fprintf(stderr, "catalyst-executor: fopen %s failed: %s\n", dst.c_str(),
+                     std::strerror(errno));
+        return -1;
+    }
+    size_t wrote = std::fwrite(bytes.data(), 1, bytes.size(), f);
```
josephleekl, line 130: JB: Partial-write file left on disk. On a short fwrite, the function returns -1 but leaves the truncated file in place and does not check fclose for flush errors. A later load_asset could pick up a corrupt file. Remove the file on failure and check fclose.

josephleekl (approved):
Thanks @mehrdad2m . Do you think we should add tests for this? up to you

multiphaseCFD (approved):
LGTM! Thanks @mehrdad2m

runtime/lib/executor/executor.cpp
rniczh: Yes, it's the thing that listed in TODO: 
```
// TODO: This is a temporary solution to initialize the catalyst CTX.
// Done per-connection so each circuit owns its own context.
```

So what the issue on that executor:  exit(0) skips destructors and finalize, the child never explicitly tears down CTX or device state for now. Memory and FDs are reclaimed by the OS, state may linger until the driver notices the process died, so the next connection can occasionally need a reset (It's fine I think if we now what we are doing)

I think it's low priority for now, we can revise it later, it need more careful lifetime control over the system.

runtime/lib/executor/executor.cpp
mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/3088/commits/05cd5c11c644275e3c710b98b6ad39fa06edb702

runtime/lib/executor/executor.cpp
mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/3088/commits/8e435b5f7fe9fe70bcfd62650c5b204e229693fc

runtime/lib/executor/executor.cpp
mehrdad2m: I improved things in https://github.com/PennyLaneAI/catalyst/pull/3088/commits/9a5f734009759abdc9b5f2b0f8443f8dfbb3667d and also tried to address the TODO in https://github.com/PennyLaneAI/catalyst/pull/3088/commits/6acf1cd4b5d2497a75566ddbff2bc35cab280a5b

runtime/lib/executor/executor.cpp
```diff
@@ -0,0 +1,522 @@
...
+    if (!finalizeFn) {
+        std::fprintf(stderr,
+                     "[%s] Warning: __catalyst__rt__finalize not found "
+                     "in any loaded plugin\n",
+                     GLabel.c_str());
+        return;
+    }
+    finalizeFn();
+}
+
```
josephleekl: should we factorize this, in case there are any other functions we want to do this too?

mehrdad2m: yeah good idea. I'll fix that shortly

mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/3088/commits/deb6a9236768ac7c0430bf12ab1d6f437bb1c4e9

runtime/lib/executor/executor.cpp
```diff
@@ -0,0 +1,503 @@
...
+void _catalyst_remote_free(ExecutorAddr addr) { std::free(addr.toPtr<void *>()); }
+
+void _catalyst_remote_invoke(ExecutorAddr fn, std::vector<ExecutorAddr> args) {
+    if (!fn) {
+        return;
+    }
+    if (args.size() != 2) {
+        llvm::report_fatal_error(
+            llvm::formatv("kernel entry function must have 2 arguments, got {0}", args.size()));
```
josephleekl, line 91: should we include "FormatVariadic.h"

josephleekl (approved):
thanks @mehrdad2m , approve again

mehrdad2m:
@josephleekl Sorry for the repeated request, I forgot to push the tests. Could you have a look at them too?

runtime/tests/Test_Executor.cpp
```diff
@@ -0,0 +1,296 @@
...
+    REQUIRE(fd >= 0);
+    sockaddr_in addr{};
+    addr.sin_family = AF_INET;
+    addr.sin_addr.s_addr = htonl(INADDR_LOOPBACK);
+    addr.sin_port = 0;
+    REQUIRE(::bind(fd, reinterpret_cast<sockaddr *>(&addr), sizeof(addr)) == 0);
+    socklen_t len = sizeof(addr);
+    REQUIRE(::getsockname(fd, reinterpret_cast<sockaddr *>(&addr), &len) == 0);
+    int port = ntohs(addr.sin_port);
+    ::close(fd);
```
josephleekl: maybe we shouldn't close it here - maybe we should close it after the test is done, so we can guarantee the port is free for the duration of the test?

runtime/tests/Test_Executor.cpp
```diff
@@ -0,0 +1,296 @@
...
+    // The staging subdirectory is keyed on the serving process's PID.
+    CHECK(Dir.filename().string().find_first_not_of("0123456789") == std::string::npos);
+    CHECK(readFile(Dir / Name) == Payload);
+    // Write-then-rename leaves no temporary alongside the asset.
+    CHECK(countEntries(Dir) == 1);
+
+    catalyst::executor::close(S);
+}
+
+TEST_CASE("concurrent connections stage same-named assets without collision", "[executor]") {
```
josephleekl: i think technically it's not testing concurrency, just testing name/directory isolation

runtime/lib/executor/executor.cpp
mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/3088/commits/df90423a80e824fd44b2b2f62159e238098730a1

runtime/tests/Test_Executor.cpp
mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/3088/commits/c8041d274d8d445ad2b744e46c420d0a0a953c22

runtime/tests/Test_Executor.cpp
mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/3088/commits/d8bfe77299c2256f47470bd601c025ef8b31cd8f

PR #3089, opened 2026-08-03
[MLIR] Add quantum.ctrl region op and ctrl-lowering pass

maliasadi:
**Context:**
As part of adding Adjoint/Ctrl region/ops support to the graph-decomposition system, we need to reduce region-level modifiers to op-level nodes before the solver runs (Check the ADR for details). Controlled subcircuits need the same treatment: a `Controlled(Op)` produced by the solver (or a `ctrl` region coming from the frontend) must be reduced to op-level controlled gates before calling into the graph solver.

This PR adds the `ctrl` counterpart to the existing `adjoint` region infra via a `quantum.ctrl` region op and a `ctrl-lowering` pass.

We started simply by just implementing a forward walk pass over the region in the `ctrl-lowering` pass, for each op:

- **gate**:  rebuilt with the region's controls appended to whatever controls it already carries.
- **already-controlled gate**: controls concatenate (via the control-merging method)
- **nested `quantum.ctrl`**: outer controls merged in and lowered on the next iteration
- **structural ops** (`extract`/`insert`/`alloc`/`dealloc`) and classical ops: just cloned

**Related GitHub Issues:**
[sc-125409]


mlir/test/Quantum/CtrlTest.mlir
```diff
@@ -0,0 +1,181 @@
...
+    %h = quantum.custom "Hadamard"() %q : !quantum.bit
+    %r = quantum.insert %arg0[ 0], %h : !quantum.reg, !quantum.bit
+    quantum.yield %r : !quantum.reg
+  }
+  return %outc, %outr : !quantum.bit, !quantum.reg
+}
+
+// -----
+
+// CHECK-LABEL: @ctrl_nested
```
kipawaa, line 136: Could you add one of each: a gate inside the outer ctrl but before the inner ctrl, and a gate inside the outer control but after the inner ctrl? Just to verify that the inlining of the inner region preserves the order of operations within the regions.

mlir/test/Quantum/CtrlTest.mlir
kipawaa, line 1: Could you add a test for duplicate control values? i.e. an op that is already controlled on the control value of the `ctrl` region?

kipawaa (commented):
Nice work @maliasadi 💯 Just requesting a couple of test cases that I think would be good to have

doc/releases/changelog-dev.md
```diff
@@ -2,6 +2,14 @@
 
 <h3>New features since last release</h3>
 
+* To add Adjoint/Ctrl support to the graph-decomposition system, we need to reduce
+  region-level modifiers to op-level nodes before the solver runs.
+  A new `quantum.ctrl` region op and a `ctrl-lowering` pass are added to the Quantum Dialect
+  for controlled subcircuits in Catalyst. A `Controlled(Op)` produced by the graph-solver
+  (or a ctrl region coming from the frontend) can be reduced to op-level controlled gates
+  before calling into the graph solver.
+  [(#3089)](https://github.com/PennyLaneAI/catalyst/pull/3089)
```
paul0403: Usually we describe the change itself, then describe the potential impacts. Since the addition of the `quantum.ctrl` operation is itself a new feature, I think we can reword into something like this:
```suggestion
* A new `quantum.ctrl` region op and a `ctrl-lowering` pass are added to the Quantum Dialect
  for controlled subcircuits in Catalyst. 
  [(#3089)](https://github.com/PennyLaneAI/catalyst/pull/3089)

Programs can now express an entire controlled quantum region, as opposed to individual operations. The `ctrl-lowering` pass distributes the control wires and control values from the `quantum.ctrl` operation onto the individual gate operations inside the region.
```

I think the changelog for `ctrl` op doesn't need to talk about graph decomp?

mlir/include/Quantum/IR/QuantumOps.td
```diff
@@ -966,7 +966,46 @@ def AdjointOp : Region_Op<"adjoint", [SingleBlockImplicitTerminator<"YieldOp">,
...
+        The `quantum.ctrl` operation applies the quantum operations in its single-block region
+        conditioned on a set of control qubits. Each `in_ctrl_values` entry selects whether the
+        matching control qubit activates on the |1> (true) or |0> (false) state.
+
+        The `ctrl-lowering` pass distributes the controls over the region: every gate in the
+        region gains the control qubits/values (appended to any controls it already carries),
+        and the control qubits are threaded through the region. Structural ops (extract, insert,
+        alloc, dealloc) are passed through unchanged, and a nested `quantum.ctrl` region has its
+        controls merged. Measurements, `scf` control flow, and nested `quantum.adjoint` regions
+        inside a `quantum.ctrl` region are rejected.
```
paul0403, line 984: Any reason why `scf` control flow are rejected? Or is it just in a follow-up PR?

I'm also happy to reject it for now if it's not needed for the workflows we care about now.

mlir/lib/Quantum/IR/QuantumOps.cpp
```diff
@@ -762,6 +762,38 @@ LogicalResult AdjointOp::verify()
...
 
+LogicalResult CtrlOp::verify()
+{
+    auto res = this->getRegion().walk([](Operation *op) {
+        return isa<MeasurementProcess, MeasureOp>(op) ? WalkResult::interrupt()
+                                                      : WalkResult::advance();
+    });
+
+    if (res.wasInterrupted()) {
+        return emitOpError("quantum measurements are not allowed in the ctrl regions");
```
paul0403, line 735: Nothing to do for now, since we can always add it later, but just wondering are controlled PPMs a thing?

mlir/lib/Quantum/IR/QuantumOps.cpp
```diff
@@ -762,6 +762,38 @@ LogicalResult AdjointOp::verify()
     return success();
 }
 
+LogicalResult CtrlOp::verify()
+{
+    auto res = this->getRegion().walk([](Operation *op) {
+        return isa<MeasurementProcess, MeasureOp>(op) ? WalkResult::interrupt()
```
paul0403, line 730: According to the op description
```
Measurements, `scf` control flow, and nested `quantum.adjoint` regions
        inside a `quantum.ctrl` region are rejected.
```

do we want to check for the others here in the verifier too?

mlir/test/Quantum/CtrlTest.mlir
```diff
@@ -0,0 +1,181 @@
...
+
+// RUN: quantum-opt --ctrl-lowering --split-input-file -verify-diagnostics %s | FileCheck %s
+
+// CHECK-LABEL: @ctrl_single_gate
+func.func @ctrl_single_gate(%ctrl: !quantum.bit, %q: !quantum.bit) -> (!quantum.bit, !quantum.bit) {
+  // CHECK-NOT: quantum.ctrl
+  // CHECK: %[[TRUE:.*]] = arith.constant true
+  %true = arith.constant true
+  // CHECK: %[[OUT:.*]], %[[OUTC:.*]] = quantum.custom "Hadamard"() %{{.*}} ctrls(%{{.*}}) ctrlvals(%[[TRUE]]) : !quantum.bit ctrls !quantum.bit
+  %outc, %outq = quantum.ctrl(%ctrl) ctrlvals(%true) (%q : !quantum.bit) : !quantum.bit -> !quantum.bit {
```
paul0403: Any way to not duplicate the input and output `!quantum.bit/reg` types?

I think I solved this issue too when updating the adjoint op a while ago, you can probably just take a look at the assembly format for adjoint op.

paul0403: Also, shouldn't there be two `!quantum.bit`s here? One for control and one for target?

mlir/test/Quantum/CtrlTest.mlir
```diff
@@ -0,0 +1,181 @@
...
+// -----
+
+// CHECK-LABEL: @ctrl_merge_existing
+func.func @ctrl_merge_existing(%ctrl: !quantum.bit, %inner: !quantum.bit, %q: !quantum.bit)
+    -> (!quantum.bit, !quantum.bit, !quantum.bit) {
+  %true = arith.constant true
+  // CHECK: quantum.custom "PauliX"() %{{.*}} ctrls(%{{.*}}, %{{.*}}) ctrlvals(%{{.*}}, %{{.*}}) : !quantum.bit ctrls !quantum.bit, !quantum.bit
+  %outc, %outq:2 = quantum.ctrl(%ctrl) ctrlvals(%true) (%inner, %q : !quantum.bit, !quantum.bit) : !quantum.bit -> !quantum.bit, !quantum.bit {
+  ^bb0(%argc: !quantum.bit, %argq: !quantum.bit):
+    %xo, %xc = quantum.custom "PauliX"() %argq ctrls(%argc) ctrlvals(%true) : !quantum.bit ctrls !quantum.bit
```
paul0403, line 62: For this test, can you check the source qubits' values too? I think it's important here to check that the target is `%q` and the controls are `%inner, %ctrl`.

While we're at it, can we use one true and one false for the ctrlvals and check these too? So we can make sure the appending is in the correct order.

Also, probably it would be a good idea to check the out qubits of the new custom op too, and check that they are correctly threaded to the return op.

mlir/test/Quantum/CtrlTest.mlir
```diff
@@ -0,0 +1,181 @@
...
+    quantum.yield %xc, %xo : !quantum.bit, !quantum.bit
+  }
+  return %outq#0, %outq#1, %outc : !quantum.bit, !quantum.bit, !quantum.bit
+}
+
+// -----
+
+// CHECK-LABEL: @ctrl_zero_value
+func.func @ctrl_zero_value(%ctrl: !quantum.bit, %q: !quantum.bit) -> (!quantum.bit, !quantum.bit) {
```
paul0403: Just wondering, how is this test different from the first test (with true ctrlval)?

mlir/lib/Quantum/IR/QuantumOps.cpp
paul0403: Oh, I see you are asserting the others in the pass. 

I assume it's because we'll add it in the future when we have time?

mlir/test/Quantum/CtrlTest.mlir
maliasadi: [cc1bfa5](https://github.com/PennyLaneAI/catalyst/pull/3089/commits/cc1bfa5ad1dec0a6ec8743674872b78385a5c159)

mlir/test/Quantum/CtrlTest.mlir
maliasadi: [8508c06](https://github.com/PennyLaneAI/catalyst/pull/3089/commits/8508c06315a9868579f2c27225f4651933ee9782)

mlir/lib/Quantum/Transforms/CtrlLowering/CtrlLowering.cpp
```diff
@@ -0,0 +1,285 @@
...
+                for (auto [oldResult, newResult] : llvm::zip_equal(
+                         gate.getNonCtrlQubitResults(), newGate.getNonCtrlQubitResults())) {
+                    map.map(oldResult, newResult);
+                }
+                // The new control results are [old controls ..., threaded region controls ...].
+                ResultRange newCtrlResults = newGate.getCtrlQubitResults();
+                for (unsigned i = 0; i < numOldControls; ++i) {
+                    map.map(gate.getCtrlQubitResults()[i], newCtrlResults[i]);
+                }
+                currentCtrlQubits.assign(newCtrlResults.begin() + numOldControls,
```
paul0403, line 197: I am finding this very similar to what I did in the reference/value semantics conversion pass, in particular the need to keep track of the "current qubit values".

Reading the implementation, I want to suggest that we write this pass in reference semantics. Over there, we don't need to worry about any qubit threading at all! The distribution is stupidly easy: just grab the same SSA `qref.bit` control qubit values on the ctrl op, add it to the gate ops, and we're done. 

No need to do it in this PR of course, but we are already thinking of moving the adjoint lowering pass to qref.  

I except to have some time after the lowering-time decomp rules are finished, so I can help out with moving this pass to reference semantics (in a follow-up PR).

mlir/lib/Quantum/Transforms/CtrlLowering/CtrlLowering.cpp
```diff
@@ -0,0 +1,285 @@
...
+
+/// Rebuild a quantum gate with additional control qubits/values appended to whatever controls it
+/// already carries. The new op is inserted at the rewriter's insertion point.
+static Operation *createControlledGate(PatternRewriter &rewriter, QuantumGate gate, IRMapping &map,
+                                       ValueRange addCtrlQubits, ValueRange addCtrlValues)
+{
+    Operation *op = gate.getOperation();
+    ValueRange nonCtrlQubits = gate.getNonCtrlQubitOperands();
+    ValueRange oldCtrlQubits = gate.getCtrlQubitOperands();
+    ValueRange oldCtrlValues = gate.getCtrlValueOperands();
```
paul0403, line 45: Why not just use the `gate.setQubitOperands()` methods? You don't have to keep track of all the other operands if you use that right?

paul0403: This util can probably just update the op in-place, and the cloning can happen in the main pattern (and when merging nested ctrl regions)?

paul0403 (commented):
Nice, thanks @maliasadi !

One thing I'm wondering is do we need a reference semantics version? Since the frontend `qp.ctrl` will be lowered to reference semantics.

I can help out with the qref.ctrl op (and the bidirectional semantics conversions) in another PR. For now, happy to get this one merged and unblock other graph decomp work!

mlir/lib/Quantum/Transforms/CtrlLowering/CtrlLowering.cpp
```diff
@@ -0,0 +1,280 @@
...
+        SmallVector<Value> currentCtrlQubits(ctrl.getInCtrlQubits().begin(),
+                                             ctrl.getInCtrlQubits().end());
+        ValueRange ctrlValues = ctrl.getInCtrlValues();
+
+        rewriter.setInsertionPoint(ctrl);
+
+        for (Operation &op : block.without_terminator()) {
+            if (isa<MeasurementProcess, MeasureOp>(op)) {
+                op.emitError("cannot control a measurement inside a quantum.ctrl region");
+                return failure();
```
paul0403: This was already checked in the verifier right?

mlir/include/Quantum/IR/QuantumOps.td
maliasadi: Added https://github.com/PennyLaneAI/catalyst/pull/3096

mlir/lib/Quantum/Transforms/CtrlLowering/CtrlLowering.cpp
maliasadi: Yeah, this is a much better way to tackle it in CQRs

mlir/lib/Quantum/IR/QuantumOps.cpp
maliasadi: we could apply ctrl to paulis before measuring the wires. But it isn't implemented in PL or Catalyst and I've never seen any applications requiring this support. It's safe to raise an error if it's appearing inside a ctrl region for now and revisit later.

mlir/test/Quantum/CtrlTest.mlir
maliasadi: Good catch! Fixed [42cabf3](https://github.com/PennyLaneAI/catalyst/pull/3089/commits/42cabf3ee2bdbb16b4c7f0e2f6aece41b40da42f)

mlir/test/Quantum/CtrlTest.mlir
maliasadi: Updated [256c3fc](https://github.com/PennyLaneAI/catalyst/pull/3089/commits/256c3fcca51b568460b7c33b1f5e23118deb8cd6)

mlir/lib/Quantum/Transforms/CtrlLowering/CtrlLowering.cpp
maliasadi: Hm.. because the region ctrl also adds a result (`out_ctrl_qubit`) to each op, so it causes the MLIR op grow its results list in-place. I tried to use it but I failed as once we try to use it we must recreate the op with updated results. Asking gemini, it suggested to use `TypeSwitch` to remove `numLeading` for a better implementation but not using the setter method. I can update the code to use a `TypeSwitch` to build the new ops inside a ctrl region if you also think it's a better way to support this.

paul0403: Oh true it has a new result! Never mind then, we probably need to edit op state.

paul0403 (approved):

kipawaa (approved):
Nice, thanks @maliasadi!

PR #3090, opened 2026-08-03
[MLIR] Add nested adjoint/ctrl support to quantum.ctrl and quantum.adjoint regions

maliasadi:
**Context:**

This PR adds the nested `adjoint` and `ctrl` compositin. 

At the op level, `adjoint` and `ctrl` are commuting fields (`Adjoint(Ctrl(g)) == Ctrl(Adjoint(g))`), so the only work is composing the two region lowerings! Both passes now defer silently when their region still contains the other kind of unlowered region:
- `ctrl-lowering` defers on a nested `quantum.adjoint`;
- `adjoint-lowering` defers on a nested `quantum.ctrl`.

^ Nesting will be reduced innermost-out: each pass reduces the layers whose body is already op-level gates, and the modifiers land on the gate in any order. So a controlled gate under adjoint just gets its adjoint flag flipped while preserving control wires & values.

**Benefits:**
- Add support for nested `adjoint` and `ctrl` composition.

**Related GitHub Issues:**
[sc-125409]

### DIFF
--- a/mlir/lib/Quantum/Transforms/AdjointLowering/AdjointLowering.cpp
+++ b/mlir/lib/Quantum/Transforms/AdjointLowering/AdjointLowering.cpp
@@ -44,6 +44,15 @@ struct AdjointSingleOpRewritePattern : public OpRewritePattern<AdjointOp> {
     /// correspondence with a notable exception caused by `insert`/`extract` API asymmetry.
     LogicalResult matchAndRewrite(AdjointOp adjoint, PatternRewriter &rewriter) const override
     {
+        // Defer (not an error) if the region still contains a nested quantum.ctrl region.
+        // ctrl-lowering must reduce it to op-level controlled gates first; reversing and adjointing
+        // those gates is then trivial ((C(g))^dagger = C(g^dagger)). The pipeline runs
+        // (ctrl-lowering, adjoint-lowering) to a fixpoint, so this adjoint op lowers on a later
+        // iteration. Pre-scanning here avoids the ReversePass "Unhandled operation" error path.
+        if (adjoint.getRegion().walk([](CtrlOp) { return WalkResult::interrupt(); }).wasInterrupted()) {
+            return failure();
+        }
+
         QuantumCache cache =
             QuantumCache::initialize(adjoint.getRegion(), rewriter, adjoint.getLoc());
 
--- a/mlir/lib/Quantum/Transforms/AdjointLowering/AdjointLowering.cpp
+++ b/mlir/lib/Quantum/Transforms/AdjointLowering/AdjointLowering.cpp
@@ -42,7 +42,17 @@ struct AdjointSingleOpRewritePattern : public OpRewritePattern<AdjointOp> {
     /// We build a map from values mentioned in the source data flow to the values of
     /// the program where quantum control flow is reversed. Most of the time, there is a 1-to-1
     /// correspondence with a notable exception caused by `insert`/`extract` API asymmetry.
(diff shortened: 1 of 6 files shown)


mlir/lib/Quantum/Transforms/AdjointLowering/AdjointLowering.cpp
```diff
@@ -44,6 +44,17 @@ struct AdjointSingleOpRewritePattern : public OpRewritePattern<AdjointOp> {
     /// correspondence with a notable exception caused by `insert`/`extract` API asymmetry.
     LogicalResult matchAndRewrite(AdjointOp adjoint, PatternRewriter &rewriter) const override
     {
+        // Defer (not an error) if the region still contains a nested quantum.ctrl region.
```
paul0403, line 46: This strategy works because it's a pattern under the greedy rewriter right? So the pattern applicator will repeatedly try the lowering itself from inner regions outwards? 🤩

mlir/test/Quantum/CtrlAdjointNestedTest.mlir
```diff
@@ -0,0 +1,126 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Nested quantum.ctrl / quantum.adjoint regions are reduced innermost-out by running the two
+// lowering passes as a fixpoint. Two rounds of (ctrl-lowering, adjoint-lowering) are enough for the
+// nesting depths exercised here.
+// RUN: quantum-opt --ctrl-lowering --adjoint-lowering --ctrl-lowering --adjoint-lowering \
```
paul0403, line 18: Ah, this is slightly more concerning, as it requires the pipeline itself to loop until success 🤔 

How about in the control lowering pass, if we detected an adjoint nested inside, we just call the adjoint lowering pass? And vice versa. This means the pipeline would stay the same for any case of these two nesting each other.

mlir/lib/Quantum/Transforms/AdjointLowering/AdjointLowering.cpp
maliasadi: Yes, but for cross-nesting, the current design leans on the pipeline looping `(ctrl-lowering, adjoint-lowering)*` and not on the greedy rewriter alone.

mlir/test/Quantum/CtrlAdjointNestedTest.mlir
maliasadi: I wanted to keep these passes independent from each other but compatible! So they can be used separate from each-other without unexpected effects on other regions. But we'll implement the nested region detection and recursive calls to adjoint/ctrl lowering in `graph-decomposition`. the plan is to invoke both in the `graph-decomposition` pass to ensure region-lelve -> op-level prior to the solver.

paul0403: Ah, so `--graph-decomposition` pass will have a pass manager than computes the necessary local pipeline?

maliasadi: Yes, that's the plan. I've already tackled this in our new approach to support adjoint and ctrl

paul0403 (approved):

kipawaa (approved):
Looks great!

PR #3095, opened 2026-08-04
Accelerate PBC layer commutation with symplectic bases

jk20342:
**Context:**

`PBCLayer` checked a candidate against every operation already in the layer
this repeated Pauli normalization and became quadratic for large commuting layers

- store Pauli operators as packed X and Z bit vectors
- maintain a lazy row echelon basis over GF(2)
- test commutation with XOR and popcount against independent basis rows
- invalidate the basis when an operation is removed
- add an MLIR regression test for dependent commuting operators

**Benefits:**

- preserves layer membership and operation order
- bounds basis rank by the number of qubits
- avoids basis construction for disjoint only layers
- reached 26.7x speedup for 16000 dependent terms on 64 qubits

**Possible Drawbacks:**

None off the top of my head

**Related GitHub Issues:**

None


zsb-xqc:
Thanks @jk20342, this looks like an efficient representation for the PBC layer, leading to a significant speedup.  We will review shortly and get back to you.

jk20342:
sounds good :)

PR #3098, opened 2026-08-05
Refine the backline transport dialect, lowering and runtime

mehrdad2m:
**Context:**
This PR adds some refinement to the transport dialect and runtime to be able to integrate backline frontend and the backends.

*Transport dialect and lowering*
- Renames the dialect ops for consistency and adds `transport.reply_slot` and
  `transport.stage_payload`.
- `lower-decode-to-transport` no longer heap-allocates for corrections, and remaining
  heap allocations on the round path are promoted.
- `inject-transport-session` and `convert-transport-to-llvm` are updated to match.
- Backline now lowers for cross-compiled targets: nested target modules are inlined
  and `keep_intermediate` handles multiple levels.
- Renames the node's `remote` flag to `out_of_process` across the dialect, the runtime
  C API and the one frontend line that emits it. The flag distinguishes a node whose
  code runs in a separate process from one in the compiling process, which is not the
  same as running on a different machine.

*Runtime*
- Adds the `catalyst-executor` server side, plus SSH control-path robustness and
  better error reporting on the deploy path.
- A backend now chooses where its reply buffer lives, rather than the common layer
  assuming it.
- The transport backends ship with the runtime, so a node on another machine can
  `dlopen` a backend by filename from its deployed bundle.
- CMake builds only the libraries actually used, and `transport_all` names the real
  per-backend targets.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


runtime/lib/transport/CMakeLists.txt
```diff
@@ -33,26 +23,32 @@ set_property(TARGET rt_transport PROPERTY POSITION_INDEPENDENT_CODE ON)
...
+find_library(IBVERBS_LIB ibverbs)
 
-# GPU coprocessor backend — only when a HIP toolchain is present.
-include(CheckLanguage)
-check_language(HIP)
-if(CMAKE_HIP_COMPILER)
-    message(STATUS "transport: HIP found (${CMAKE_HIP_COMPILER}) -> building gpu_verbs")
-    set(TRANSPORT_HAS_HIP TRUE CACHE INTERNAL "transport: HIP toolchain available")
-    add_subdirectory(gpu_verbs)
+add_custom_target(transport_all)
```
josephleekl, line 39: can you help update [the `runtime/Makefile`](https://github.com/PennyLaneAI/catalyst/blob/d37f75f1869ea5e93c2fc8670feb2f50603df60d/runtime/Makefile#L60) to reflect this?

mlir/lib/Transport/Transforms/LowerDecodeToTransport.cpp
```diff
@@ -74,15 +82,71 @@ struct LowerDecodeToTransportPass
...
+            //   into the program's buffer          into the ring slot
+            //   -------------------------          ---------------------------------
+            //   %buf = memref.alloc()              %slot = transport.reply_slot %s
+            //   transport.collect %s, %buf         transport.collect %s, %slot
+            //   %v   = memref.load %buf            %v    = memref.load %slot
+            //   memref.dealloc %buf
+            //
+            // The right-hand one costs neither the allocation nor the backend's copy out of the
+            // ring, but it hands %slot to code that was written against %buf, so it is valid
+            // only while nothing can tell the two apart. Two conditions establish that.
```
josephleekl: JB: `LowerDecodeToTransport.cpp:110-119` accepts any number of `memref.load` users when deciding to alias the correction buffer onto the ring slot, but the slot is reused every 256 rounds and collect now skips the copy for it (`CpuControllerSession.cpp:104`) — so a load that ends up after a later round on the same session silently reads the wrong reply. Condition (1) needs an ordering/dominance check against the next post on that session, plus a two-decode lit test, since all three new reply_slot tests are single-decode.

josephleekl: Fix: Tighten condition (1) to require every memref.load user to dominate the next transport op on that session — or simplest, require that the only users are loads in the same block as the anchor with no intervening PostOp on the same session. Add a lit test with two decodes on one session where the first correction is read late, asserting it takes the copy-out path. Also add a ConvertTransportToLLVM.mlir case for reply_slot — right now the descriptor construction at TransportToLLVM.cpp:826-835 has no test at all.

mlir/lib/Driver/CompilerDriver.cpp
```diff
@@ -476,10 +476,23 @@ catalyst::driver::runLowering(const CompilerOptions &options, MLIRContext *ctx,
...
+            case SaveTemps::AfterPass:
+                saveIrAfterEach = "pass";
+                break;
+            case SaveTemps::None:
+                break;
+            }
             targetPipeline.setPasses({"cross-compile-targets{workspace=\"" +
                                           options.workspace.str() +
-                                          "\" dump-intermediate=" + dumpIntermediate + "}",
+                                          "\" save-ir-after-each=" + saveIrAfterEach + "}",
```
josephleekl: should we only do this when non-empty? something like

```
std::string opts = "workspace=\"" + options.workspace.str() + "\"";
if (!saveIrAfterEach.empty()) {
    opts += " save-ir-after-each=" + saveIrAfterEach;
}
targetPipeline.setPasses({"cross-compile-targets{" + opts + "}", "dispatch-executor-targets"});
```

mlir/include/Driver/DefaultPipelines/DefaultPipelines.h
```diff
@@ -62,6 +62,8 @@ const PipelineList pipelineList{
      {"canonicalize",
       "func.func(chlo-legalize-to-stablehlo)",
       "func.func(stablehlo-legalize-control-flow)",
+      // builtin.module is added to support nested modules 
+      "builtin.module(func.func(stablehlo-legalize-control-flow))",
```
josephleekl, line 66: just to check - this is meant to be here?

runtime/lib/transport/TransportCAPI.cpp
```diff
@@ -87,10 +87,23 @@ template <typename Fn> auto guard(Fn &&fn) -> decltype(fn()) {
...
 }
 
+const char *collect_error_name(int rc) {
+    switch (rc) {
+    case CATALYST_TRANSPORT_ERR_MEMORY:
+        return "memory";
+    case CATALYST_TRANSPORT_ERR_TIMEOUT:
+        return "timeout";
+    case CATALYST_TRANSPORT_ERR_STUCK:
+        return "stuck — no reply before the deadline";
```
josephleekl:
```suggestion
        return "stuck - no reply before the deadline";
```

that's a non-ascii em dash 😆

mlir/include/Catalyst/Transforms/Passes.td
```diff
@@ -168,14 +168,15 @@ def CrossCompileTargetsPass : Pass<"cross-compile-targets", "mlir::ModuleOp"> {
...
+            /*type=*/"std::string",
+            /*default=*/"\"\"",
             /*description=*/
-            "Also write each target module's extracted MLIR and translated LLVM IR "
-            "(`extracted.mlir`, `<name>.ll`) into its workspace subdirectory. Set by the "
-            "driver from `keep_intermediate`."
+            "How much of each target module's lowering to write into its workspace subdirectory."
+            "Empty writes  nothing; `pipeline` writes the extracted MLIR and translated LLVM IR "
+            "(`extracted.mlir`, `<name>.ll`); `changed` and `pass` add the IR after each pass "
+            "that altered it, or after every pass. Set by the driver from `keep_intermediate`."
```
josephleekl, line 179: is this related? or is this upstream changes

mlir/lib/Transport/Transforms/LowerDecodeToTransport.cpp
```diff
@@ -91,9 +155,8 @@ struct LowerDecodeToTransportPass
                 anchors.push_back(op);
             }
         });
-        // Distribute the decoding tasks across the coprocessors in a round-robin fashion.
-        for (size_t k = 0; k < anchors.size(); ++k) {
-            qecp::DecodeEsmCssOp anchor = anchors[k];
+        // Decodes are handed to the coprocessors round-robin.
```
josephleekl, line 182: do you want to add a TODO to refine this implementation in the future?

josephleekl (approved):
amazing stuff, thanks @mehrdad2m ! Just some small comments, again please merge when you get enough reviews/comments are addressed 🔥

mlir/include/Transport/IR/TransportDialect.td
```diff
@@ -113,7 +113,7 @@ def Transport_NodeAttr : Transport_Attr<"Node", "node"> {
         Transport_StrField:$triple,
         Transport_StrField:$address,
         Transport_StrField:$symbol,
-        OptionalParameter<"mlir::BoolAttr">:$remote,
+        OptionalParameter<"mlir::BoolAttr">:$out_of_process,
```
multiphaseCFD, line 115: Note `data_path` is removed in the PR#3113

runtime/lib/transport/CMakeLists.txt
mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/3098/commits/8002a98c1a2a5cee9d138a9b88005d93d6baf965

mlir/include/Transport/IR/TransportOps.td
```diff
@@ -94,9 +94,14 @@ def Transport_EstablishChannelOp : Transport_Op<"establish_channel"> {
 // Round setup
 //===----------------------------------------------------------------------===//
 
-def Transport_CommitWorkItemOp : Transport_Op<"commit_work_item"> {
```
multiphaseCFD, line 97: we need to update the runtime accordingly. A SC story was added sc-128136

runtime/include/TransportCAPI.h
```diff
@@ -32,13 +32,13 @@ extern "C" {
...
 // Return codes: 0 == success; negative == error
-enum {
+typedef enum {
     CATALYST_TRANSPORT_OK = 0,
     CATALYST_TRANSPORT_ERR = -1,         // Generic exception
     CATALYST_TRANSPORT_ERR_MEMORY = -2,  // Memory error
     CATALYST_TRANSPORT_ERR_TIMEOUT = -3, // Timeout error
     CATALYST_TRANSPORT_ERR_STUCK = -4,   // Something got stuck
-};
+} CatalystTransportStatus;
```
multiphaseCFD, line 41: Nice

runtime/lib/transport/rdma/cpu_verbs/controller/CpuControllerSession.hpp
```diff
@@ -42,6 +42,7 @@ class CpuControllerSession : public CpuSessionBase<ControllerSession> {
     int kick(std::uint32_t work_item_idx = 0) override;
     void *data_slot() override;
     void write_data_slot(const void *src, std::uint64_t bytes, std::uint32_t decoder_id) override;
+    void *reply_slot() override;
```
multiphaseCFD, line 45: Do we want to rename the `commit_work_item` method?

mlir/include/Driver/DefaultPipelines/DefaultPipelines.h
mehrdad2m: Yes, so the when you wrap function passes like `func.func(stablehlo-legalize-control-flow)` it only looks at the function inside your module and never enters nested modules which causes a stablehlo.while to not be lowered if in a nested module. This is essentially telling the pass to look inside one level of nested modules. Ideally we need to analyse the rest of the function passes to potentially do the same thing for them as well, but in the lack of time I am just fixing what I saw broke.

runtime/lib/transport/rdma/cpu_verbs/controller/CpuControllerSession.hpp
mehrdad2m: Any suggestions?

multiphaseCFD (approved):
awesome work @mehrdad2m !

runtime/lib/transport/rdma/cpu_verbs/controller/CpuControllerSession.hpp
mehrdad2m: wait this is already changed though right? to `set_message_sizes`?

mlir/include/Transport/IR/TransportOps.td
mehrdad2m: Ah yeah I think we changed the CAPI to `__catalyst__transport__set_message_sizes` that calls the `commit_work_item` inside it.

mlir/include/Transport/IR/TransportDialect.td
mehrdad2m: I think I have already remove it on https://github.com/PennyLaneAI/catalyst/pull/3113

runtime/lib/transport/rdma/cpu_verbs/controller/CpuControllerSession.hpp
multiphaseCFD: given the timeline we have, i think good for now, i added a SC story for this later.

mlir/lib/Transport/Transforms/LowerDecodeToTransport.cpp
mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/3098/commits/cd909ba7f9c6a228a213b8bbad2aa3a83e98ffe8

mlir/lib/Transport/Transforms/LowerDecodeToTransport.cpp
mehrdad2m: https://github.com/PennyLaneAI/catalyst/commit/8002a98c1a2a5cee9d138a9b88005d93d6baf965

mlir/include/Catalyst/Transforms/Passes.td
mehrdad2m: Yes this is an option for the cross-compile-targets pass which is responsible for cross compiling controller/coprocessor modules. It runs a separate pipeline + llc to create a object file. The problem solved by this PR was this: 
we currently have something called `keep_intermediate` to determine how much intermediate IR we want to save it can be 'True`/'pipeline', 'changed`, and `pass` which prints the IR after each pipeline, pass(if changed the IR) and pass respectively. As this pass runs a separate pipeline, it had a separate instrumentation which only supported keep_intermediate='True'. but now we are supporting the whole verbosity levels, so one can see the whole list of passes run for each module. The point of this option is to pipe the same level of verbosity the outer pass manager operates under to the nested modules pass manager as well.

mlir/lib/Driver/CompilerDriver.cpp
mehrdad2m: https://github.com/PennyLaneAI/catalyst/pull/3098/commits/96724e9b3fce929795553f197f48962cfd283a8e

PR #3100, opened 2026-08-05
Soften AOT failures to warnings

albi3ro:
**Context:**

I get a lot of AOT failures when either writing qjit for things that don't work yet or for resource estimation. Really annoying to have to remember to specify `target="mlir"`.

So this just turns those hard errors into warnings and logs the full traceback.

**Benefits:**

I can start leaving `target="mlir"` out.

**Possible Drawbacks:**

Didn't think abotu the design too deeply. Might not be the best, but it gives me less rage than I had before at least.

**Related GitHub Issues:**


doc/releases/changelog-dev.md
```diff
@@ -12,6 +12,8 @@
 
 <h3>Improvements 🛠</h3>
 
+* A failure during AOT compilation is now downgraded to a warning and logged.
```
albi3ro, line 24:
```suggestion
* A failure during AOT compilation is now downgraded to a warning and logged.
  [(#3100)](https://github.com/PennyLaneAI/catalyst/pull/3100)
```

frontend/catalyst/jit.py
```diff
@@ -694,15 +694,30 @@ def aot_compile(self):
...
         if self.compile_options.target in ("llvmir", "binary"):
-            self.compiled_function, self.llvm_ir = self.compile()
+            try:
+                self.compiled_function, self.llvm_ir = self.compile()
+            except Exception as e:  # pylint: disable=broad-exception-caught
+                logger.exception(e, exc_info=True)
+                warnings.warn("AOT generation of llvmir failed. Error logged at exception level")
+                return
 
         if self.compile_options.target in ("binary",) and self.compile_options.link:
```
jzaia18, line 722: Is an error not possible at the binary level as well? We don't soften a failure here as a warning

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_diagonalize_measurements.py
```diff
@@ -891,8 +891,11 @@ def test_non_commuting_multiple_measurements_only(self, capture_mode, device):
         def circuit():
             return qp.expval(qp.X(0)), qp.var(qp.Z(0))
 
+        with pytest.warns(UserWarning):
```
jzaia18:
```suggestion
        with pytest.warns(UserWarning, match="AOT.*failed"):
```

Just to make sure that this isn't silencing other warnings (same goes for a few other places in the code)

frontend/catalyst/jit.py
albi3ro: It looked to me like this line wasn't actually producing the binary, but just inserting it into the cache.

jzaia18:
I'll ask the dangerous question: do we need to raise a warning at all? In any situation where an AOT compilation has failed we're going to have to JIT-compile when the user runs the circuit anyway, and the failure would happen then (with essentially the same error message and traceback).

If we remove the warning, the downside is that at qjit-time there is no indication of an issue with the defined qnode. But is it common to define a qnode without running it immediately in the next line or 2 (assuming that it _will_ be run)?

Just food for thought, I'm not very confident that being completely silent on AOT failure is the better answer, just want to think through all the options.

frontend/catalyst/jit.py
jzaia18: Inserting into the cache runs a bit of extra code, but looking through it there's still no obvious place that would raise new errors. Okay, I'll resolve this. We can always add a catch here later on if it becomes an issue

frontend/test/pytest/python_interface/transforms/quantum/test_xdsl_diagonalize_measurements.py
albi3ro: all the test spots should be updated now.

albi3ro:
> I'll ask the dangerous question: do we need to raise a warning at all? In any situation where an AOT compilation has failed we're going to have to JIT-compile when the user runs the circuit anyway, and the failure would happen then (with essentially the same error message and traceback).
> 
> If we remove the warning, the downside is that at qjit-time there is no indication of an issue with the defined qnode. But is it common to define a qnode without running it immediately in the next line or 2 (assuming that it _will_ be run)?
> 
> Just food for thought, I'm not very confident that being completely silent on AOT failure is the better answer, just want to think through all the options.

I honestly don't know and am trying to avoid letting the perfect get in the way of the good. I think there is room to improve this, but I'm trying to avoid overthinking it too much. All I know if that I always forget to set `target="mlir"` and I want to be allowed to do so.

jzaia18:
> I honestly don't know and am trying to avoid letting the perfect get in the way of the good. I think there is room to improve this, but I'm trying to avoid overthinking it too much. All I know if that I always forget to set target="mlir" and I want to be allowed to do so.

Fair enough, let's proceed with this option then, it's not a heavy lift to change course later if we need to

jzaia18 (approved):
The changes here all look good to me. 1 CI is failing but it seems unrelated to this PR, just make sure to update from `main` and see if that makes it green

frontend/test/pytest/test_preprocess.py
```diff
@@ -287,7 +287,7 @@ def circuit(x: float, y: float):
 
     @pytest.mark.xfail(
         reason="Legacy preprocessing cannot decompose Operator2 Hadamard subclasses",
-        raises=CompileError,
+        raises=TypeError,  # no mlir since CompileError silenced
```
sengthai: Could we avoid making `TypeError` as the expected failure? it is only an incidental result of evaluating `"RX" in mlir` when `mlir is None`, and could hide an unrealted `TypeError`. 
Since AOT failure is now represnted by missing MLIR, could we remove the decorator and handle that condition explicitly? 

```python
...
mlir = qjit(circuit, target="mlir").mlir
if mlir is None:
    pytest.xfail("Legacy preprocessing cannot decompose Operator2 Hadamard subclasses")
...
```

This could help other failure visible.

albi3ro: Strangely enough locally I'm getting a DeviceError instead of CompileError locally. Will try and figure out what in the world is going on.

albi3ro: OK hopefully just pushed a fix.

sengthai: The latest change also works! If it passes, this looks good to me.

AntonNI8 (approved):
Looks good to me, a warning is fine here for now 👍🏻

sengthai (approved):
Thanks @albi3ro!

PR #3101, opened 2026-08-05
Runtime call is added to catalyst

rniczh:
**Context:**

This PR lets `qp.runtime_call` invoke symbols from a local/remote shared library directly. And a `transport_start_benchmark` api is added to provide the benchmark function, which is supposed to be called by the `runtime_call` from pennylane https://github.com/PennyLaneAI/pennylane/pull/9970.

### For runtime call:

- Lower `qp.runtime_call` operations to `catalyst.custom_call`.
- Track required shared libraries during lowering and pass them to the linker.
- Support opaque buffer arguments and side-effect-only calls with no results.

### Transport API:

- Transport wrapper is added for providing the interface to call to the library via remote runtime_call or local runtime_call.
- A benchmark api is added

### Coprocessor changes:

- Replace scheduler yields with architecture-specific spin hints.
- Add best-effort CPU pinning and real-time scheduling on Linux.
- Apply the polling changes to both CPU and GPU coprocessor sessions.  (reduce the time that system_call involved in the CPU path) 

TODO: 
- [ ] For polling the buffer in coprocessor CPU, it should be spin-loop only, no need to aware the terminate signal, we need to kill the process manually. -> It's quite painful for now during the benchmark, so I use the accumulator to only trigger the detection on the terminate signal in  600'000'000 times.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-127281]


frontend/catalyst/jax_primitives.py
```diff
@@ -546,6 +548,25 @@ def _python_callback_lowering(
...
+    # pylint: disable=unused-argument
+    results_ty = list(convert_shaped_arrays_to_tensors(jax_ctx.avals_out))
+    call_op = CustomCallOp(results_ty, list(operands), symbol, number_original_arg=len(operands))
+    if dispatch is not None:
+        call_op.operation.attributes["backend_config"] = ir.DictAttr.get(
+            {"dispatch": ir.StringAttr.get(dispatch)}
+        )
+    elif library:
+        record_runtime_artifact(jax_ctx.module_context.module.operation, library)
+    return call_op.results
```
josephleekl, line 568: should we move these to pennylane frontend?

runtime/include/TransportCAPI.h
```diff
@@ -90,6 +90,18 @@ uint64_t __catalyst__transport__last_rtt_ns(CatalystTransportSession *s);
...
+// Flags for start_benchmark.
+enum {
+    CATALYST_BENCH_CLEAR_SENTINEL = 1u << 0,
+    CATALYST_BENCH_FORCE_SW_RTT = 1u << 1,
+    CATALYST_BENCH_PROGRESS = 1u << 2,
+};
+
+// Run `iters` rounds back to back and report each round's round-trip time in nanoseconds.
+int __catalyst__transport__start_benchmark(CatalystTransportSession *s, uint32_t iters,
+                                           uint32_t decoder_id, uint32_t flags, uint64_t *samples,
```
josephleekl, line 103: do we really need decoder_id? can we just put a dummy 0 in the data?

runtime/lib/transport/CMakeLists.txt
```diff
@@ -1,25 +1,15 @@
-###############################################
-# transport backends: shared dependencies     #
-###############################################
-
-find_package(Threads REQUIRED)
-find_library(IBVERBS_LIB ibverbs)
-if(NOT IBVERBS_LIB)
-    message(FATAL_ERROR
-        "ENABLE_TRANSPORT requires libibverbs (e.g. libibverbs-dev / rdma-core), but it was not found.")
-endif()
```
josephleekl: i think later on @kszenes will add a local device that doesn't depend on ibverbs, so those one doesn't require this check. We can change it later too, or now.

runtime/lib/transport/TransportCAPI.cpp
```diff
@@ -420,7 +439,98 @@ std::uint64_t __catalyst__transport__last_rtt_ns(CatalystTransportSession *s) {
...
 }
 
-void __catalyst__transport__start(CatalystTransportSession *s) {
+// The benchmark loop
+// Flags:
+//     CATALYST_BENCH_CLEAR_SENTINEL: Clear the sentinel value in the reply slot.
+//     CATALYST_BENCH_FORCE_SW_RTT: Force the use of the software RTT.
+//     CATALYST_BENCH_PROGRESS: Print progress information.
+int __catalyst__transport__start_benchmark(CatalystTransportSession *s, std::uint32_t iters,
+                                           std::uint32_t decoder_id, std::uint32_t flags,
```
josephleekl, line 480: we don't need to expose decoder_id here, let's just hardcode to 0 in the code?

runtime/lib/transport/TransportCAPI.cpp
```diff
@@ -420,7 +439,98 @@ std::uint64_t __catalyst__transport__last_rtt_ns(CatalystTransportSession *s) {
...
+    if (!s->work_item_ready) {
+        std::cerr << "[transport] start_benchmark: no work item committed; call "
+                     "commit_work_item first so the round's sizes are known\n";
+        return CATALYST_TRANSPORT_ERR;
+    }
+
+    // The round's shape is whatever was committed, so it cannot disagree with it.
+    const std::uint32_t work_item_idx = s->work_item;
+    const auto syndrome_bytes = static_cast<std::uint32_t>(s->in_bytes);
+    const auto correction_bytes = static_cast<std::uint32_t>(s->out_bytes);
```
josephleekl: for this benchmark mode, let's just call them outgoing_message_bytes and reply_bytes, these have nothing to do with syndrome/correction, they're just dummy data?

runtime/lib/transport/TransportCAPI.cpp
```diff
@@ -420,7 +439,98 @@ std::uint64_t __catalyst__transport__last_rtt_ns(CatalystTransportSession *s) {
     return 0;
 }
 
-void __catalyst__transport__start(CatalystTransportSession *s) {
+// The benchmark loop
+// Flags:
+//     CATALYST_BENCH_CLEAR_SENTINEL: Clear the sentinel value in the reply slot.
+//     CATALYST_BENCH_FORCE_SW_RTT: Force the use of the software RTT.
+//     CATALYST_BENCH_PROGRESS: Print progress information.
+int __catalyst__transport__start_benchmark(CatalystTransportSession *s, std::uint32_t iters,
```
josephleekl, line 479: we already have iter here, so in the demo we don't need a for loop, right?

runtime/lib/transport/TransportCAPI.cpp
```diff
@@ -420,7 +439,98 @@ std::uint64_t __catalyst__transport__last_rtt_ns(CatalystTransportSession *s) {
...
+    return guard([&] {
+        std::vector<std::uint8_t> syndrome(
+            std::max<std::size_t>(syndrome_bytes, sizeof(catalyst::transport::common::Payload)), 0);
+        std::uint64_t written = 0;
+        int rc = CATALYST_TRANSPORT_OK;
+
+        for (std::uint32_t i = 0; i < iters; ++i) {
+            // Frame the syndrome as a Payload in a local buffer, then hand it to the transport.
+            // write_data_slot copies it into the outbound slot and enforces the slot capacity.
+            std::uint64_t value = (static_cast<std::uint64_t>(i) << 32) | 0xC0DE1515u;
```
josephleekl: can we define the magic number `0xC0DE1515u` in some shared protocol file, instead of here?

runtime/lib/transport/TransportCAPI.cpp
```diff
@@ -420,7 +439,98 @@ std::uint64_t __catalyst__transport__last_rtt_ns(CatalystTransportSession *s) {
...
+
+            std::uint64_t hw_rtt = force_sw_rtt ? 0 : s->sess->last_rtt_ns();
+            samples[written++] = (hw_rtt != 0) ? hw_rtt : sw_rtt;
+
+            if (progress && ((i & 1023) == 0 || i == iters - 1)) {
+                std::uint64_t cval = 0;
+                std::memcpy(&cval, rslot, std::min<std::size_t>(correction_bytes, sizeof(cval)));
+                std::cerr << "[transport] round " << i << " rtt=" << samples[written - 1] << " ns ["
+                          << (hw_rtt ? "hw" : "sw") << "] corr[0:8]=0x" << std::hex << cval
+                          << std::dec << "\n";
```
josephleekl, line 559: will we keep this?

runtime/lib/transport/TransportWrappers.cpp
```diff
@@ -0,0 +1,519 @@
...
+// TransportWrappers.cpp
+//
+//   __wrapper   dispatched. The executor resolves the symbol through LLVM ORC and calls it with one
+//               flat argument buffer; the result goes back as another flat buffer. Neither buffer
+//               is framed, so the layout comes from the signature the call site declared.
+//   __call      in-process, through catalyst.custom_call. Each argument arrives as its own encoded
+//               memref, read through its data pointer.
+//
+// Both layouts have to agree with pennylane/runtime/operands.py, which is what builds the
+// arguments.
```
josephleekl, line 24: interesting, so when do we use _wrapper and when do we use _call? Curious to see what you think @mehrdad2m

runtime/lib/transport/TransportWrappers.cpp
```diff
@@ -0,0 +1,519 @@
...
+    const char *config = in.str();
+    auto role = in.get<std::int32_t>();
+    const char *key = in.str();
+    Out out(U64);
+    if (in.ok() && out) {
+        *out.slot<std::uint64_t>() = reinterpret_cast<std::uintptr_t>(
+            __catalyst__transport__create(library, config, role, key));
+    }
+    return finish(in, out);
+}
```
josephleekl, line 200: is there any way to template these functions? it seems difficult to have to create a wrapper for every single CAPI call...

runtime/include/Transport.hpp
```diff
@@ -23,7 +23,7 @@ namespace catalyst::transport {
 /**
  * @brief Memory kind: selects the allocation and registration path.
  */
-enum class MemKind : std::uint8_t {
+enum class MemKind : int {
```
josephleekl, line 26: why do we change it ?

runtime/lib/transport/common/WireProtocol.hpp
```diff
@@ -19,7 +19,7 @@
 namespace catalyst::transport::common {
 
 // Ring size, identical on both sides; power of two so index is a mask.
-inline constexpr std::size_t K_RING_SLOTS = 256;
+inline constexpr std::size_t K_RING_SLOTS = 8;
```
josephleekl: is this enough? why not keep it bigger?

runtime/lib/transport/common/WireProtocol.hpp
```diff
@@ -31,41 +31,34 @@ inline constexpr std::uint32_t SALT = 0xC0DE1515u;
...
 #pragma pack(push, 1)
 struct Payload {
     std::uint64_t value;
+    std::uint8_t value_ext[PAYLOAD_DATA_BYTES - sizeof(std::uint64_t)];
     std::uint32_t decoder_id; // selects which decoder handles this message
     std::uint32_t seq_num;    // arrival flag
 };
 #pragma pack(pop)
-static_assert(sizeof(Payload) == 16, "Payload must be exactly 16 B");
+static_assert(sizeof(Payload) == PAYLOAD_DATA_BYTES + 8, "Payload is the data area plus an 8 B header");
```
josephleekl: wait this seems very weird. why do we change it? and why do we do it like this? what is value_ext? why are we making this 40B??

josephleekl (commented):
thanks @rniczh ! just some comments :D

frontend/catalyst/jax_primitives.py
josephleekl: oh i see the pennylane PR now

runtime/lib/transport/common/WireProtocol.hpp
rniczh: This is mainly for testing. We'll switch back (when remove the Draft tag) after the current tests are confirmed solid, because the original size testing is prone to corner cases. If payload size may be variable in the future, locking it down early for development is relatively risky. So in my own testing I use other sizes for simple checks until the round-trip values come back correct.

runtime/include/Transport.hpp
rniczh: `MemKind` crosses a `.so` boundary by value, and the two sides are built by different compilers, one is `librt_transport` with g++, and the otherone is `gpu_verbs` with hipcc (ROCm).

```cpp
s->reply = s->sess->alloc_memory(kReplyBytes, s->sess->preferred_mem_kind());
```

**Issue:**
With a `uint8_t` underlying type, `RDMA_CHECK(kind == MemKind::GpuHbm)` inside `GpuCoprocessorSession::alloc_memory` fired with `kind=1` but `GpuHbm=1`, two values that both print as 1, comparing unequal 😅.

**Reason:**
The x86-64 ABI leaves the bits above a sub-word argument unspecified and compilers differ on who extends them, so the callee compares a full register while only the low byte is meaningful, `int` is the ABI's natural width, so there is nothing to disagree about. It does change MemRegion's layout, so both sides must be rebuilt together. cheap next to the 3 bytes the narrow type saved.

runtime/lib/transport/common/WireProtocol.hpp
rniczh: It's also for testing (same as [this rely](https://github.com/PennyLaneAI/catalyst/pull/3101#discussion_r3742542420)), it makes it easier for me to see the values when the index wraps to 0 after more than 8 rounds.

runtime/lib/transport/TransportWrappers.cpp
rniczh: Whichever the call site asked for: `qp.runtime_call(..., address=...)` dispatches and resolves `__wrapper`; without `address` it is a local call and resolves `__call`. So it's a same operation with two adapters.

runtime/lib/transport/TransportWrappers.cpp
rniczh: The layouts differ because the calling conventions do. A local call goes through `catalyst.custom_call`, which hands each argument over as its own encoded memref and with same address space, nothing to serialise. A dispatched call goes through LLVM ORCv2's `callWrapper`, whose only signature is `CWrapperFunctionResult(const char *, size_t)`: one flat buffer in, one out. So the arguments have to be packed.

**And to answer your question**: Mostly yes (except for the start_benchmark, but it's fine I think), and LLVM already does it that way for the builtins  -- `executor.cpp` uses `shared::WrapperFunction<shared::SPSExecutorAddr(uint64_t)>::handle(...)`, which deduces the serialisation from the signature. A variadic template over the parameter pack of `decltype(Fn)` would emit `in.session()`/`in.str()`/`in.get<T>()` for us the same way.

runtime/lib/transport/TransportCAPI.cpp
rniczh: yes

runtime/lib/transport/CMakeLists.txt
```diff
@@ -33,18 +23,33 @@ set_property(TARGET rt_transport PROPERTY POSITION_INDEPENDENT_CODE ON)
...
+    add_subdirectory(common)
+    add_subdirectory(cpu_verbs)
+    add_dependencies(transport_all catalyst_transport_cpu_controller
+                     catalyst_transport_cpu_coprocessor steane_coprocessor)
+
+    # GPU coprocessor backend — only when a HIP toolchain is present.
+    include(CheckLanguage)
+    check_language(HIP)
+    if(CMAKE_HIP_COMPILER)
+        message(STATUS "transport: HIP found (${CMAKE_HIP_COMPILER}) -> building gpu_verbs")
```
kszenes: I think we need to link the CUDA runtime for NVIDIA (see https://github.com/kszenes/catalyst/blob/99aaaf5e7d215822bc85ab22e2710a398acc5cf2/runtime/lib/transport/CMakeLists.txt#L55-L57)

runtime/include/TransportCAPI.h
rniczh: Two reasons I'd prefer to keep it rather than hard-code 0:

1. `CpuCoprocessorSession` hands each message's decoder_id to the `CoprocessorFn` and echoes it in the reply. If we fix it at 0 here, the benchmark becomes the one path that cannot address a second decoder, on the day someone writes one.
2. And it is cheaper to keep than to add back.

runtime/lib/transport/TransportCAPI.cpp
rniczh: Same as https://github.com/PennyLaneAI/catalyst/pull/3101#discussion_r3752665337

runtime/lib/transport/common/WireProtocol.hpp
rniczh: Now the data payload has been updated to the original payload size: https://github.com/PennyLaneAI/catalyst/pull/3101/commits/7d0d2d569614c9d40683b8b942a41e76b8eff068

runtime/lib/transport/common/WireProtocol.hpp
rniczh: updated: https://github.com/PennyLaneAI/catalyst/pull/3101#discussion_r3755307663

runtime/lib/transport/TransportCAPI.cpp
rniczh: It has been removed, for the start_benchmark function, now we just only use iteration as the input:
```
const std::uint64_t value = static_cast<std::uint64_t>(i) + 1;
```

runtime/include/TransportCAPI.h
josephleekl: sure sounds good

runtime/lib/transport/TransportWrappers.cpp
josephleekl: I see I see. I was a little confused by the logic of this file, would be nice to have an explanation at the top :D

runtime/lib/transport/TransportCAPI.cpp
rniczh: Yes, and it wouldn't be triggered if you never pass the flag to it. So I think it's fine to leave it here for debugging purpose?

josephleekl: ah i see, yeah let's keep it then

runtime/lib/transport/CMakeLists.txt
rniczh: could you add this on your end as it's not related this PR

runtime/lib/transport/TransportCAPI.cpp
rniczh: As discussed in the previous meeting, we'll still keep a explicit loop case in the demo/benchmark, so both runtime call to the start_benchmark and explicit loop call the post and collect are presented

runtime/lib/transport/TransportCAPI.cpp
rniczh: The name is updated https://github.com/PennyLaneAI/catalyst/pull/3101/commits/5cdcbc969911e2c2b0a546e2eda5a1a3dea986eb

frontend/catalyst/compiler.py
```diff
@@ -193,6 +193,16 @@ def get_default_flags(options):
         if os.path.isfile(os.path.join(rt_lib_path, "librt_OQD_capi" + file_extension)):
             default_flags.append("-lrt_OQD_capi")
 
+        # Shared libraries exporting symbols reached by a local `runtime_call`, recorded on the
+        # module via `catalyst.runtime_artifacts` and collected into the options.
+        for artifact_path in options.runtime_artifacts:
+            dir_name = os.path.dirname(artifact_path)
```
josephleekl, line 200: JB: When `library=` is a bare filename, `os.path.dirname` yields `""`, so we emit a bare `-L` that swallows the `.so` path as its argument and the library never gets linked — the link still exits 0 and it fails later at dlopen with undefined symbol. Running the path through `os.path.abspath` and skipping the empty-dirname case fixes it.

josephleekl:
```
for artifact_path in options.runtime_artifacts:
    artifact_path = os.path.abspath(artifact_path)
    dir_name = os.path.dirname(artifact_path)
    default_flags += [f"-Wl,-rpath,{dir_name}", f"-L{dir_name}", artifact_path]
```

runtime/lib/transport/TransportCAPI.cpp
```diff
@@ -419,6 +437,91 @@ std::uint64_t __catalyst__transport__last_rtt_ns(CatalystTransportSession *s) {
     return 0;
 }
 
+// The benchmark loop
+// Flags:
+//     CATALYST_BENCH_FORCE_SW_RTT: Force the use of the software RTT.
+//     CATALYST_BENCH_PROGRESS: Print progress information.
+int __catalyst__transport__start_benchmark(CatalystTransportSession *s, std::uint32_t iters,
```
josephleekl, line 479: JB: __catalyst__transport__start_benchmark writes iters 8-byte samples into a caller buffer whose capacity it is never told, and the __call adapter passes a buffer sized only by the user's out_bytes — mismatched values are a straight heap overflow (ASan-confirmed). It needs an explicit capacity argument plus a frontend check that out_bytes >= iters * 8.

runtime/lib/transport/TransportCAPI.cpp
```diff
@@ -55,10 +57,22 @@ struct CatalystTransportSession {
     PeerRef peer; // peer region learned in exchange_keys
     bool peer_ready = false;
     std::vector<std::int64_t> pending_tokens;
+
+    std::uint32_t work_item = 0;
+    std::uint64_t in_bytes = 0;
+    std::uint64_t out_bytes = 0;
```
josephleekl, line 64: so are we removing these? @kszenes @mehrdad2m ?

josephleekl: in_bytes and out_bytes specifically?

josephleekl (approved):
Thanks @rniczh ! Just some tiny comments. Please merge when updated/get a 2nd approval

frontend/catalyst/compiler.py
rniczh: abspath is added: https://github.com/PennyLaneAI/catalyst/pull/3101/commits/8094cd89e7700665d3986841da5f76777a794e6d

runtime/lib/transport/TransportCAPI.cpp
rniczh: sure thing: https://github.com/PennyLaneAI/catalyst/pull/3101/commits/48228e54929c3cc16a649fb2300d2dbe6a78d1a8

runtime/lib/transport/TransportCAPI.cpp
kszenes: @josephleekl We won't be removing this for now but https://github.com/PennyLaneAI/pennylane/pull/10017 forces 8 bytes in/out in the frontend

frontend/catalyst/utils/runtime_artifacts.py
```diff
@@ -0,0 +1,63 @@
...
+local (in-process) external symbol exported by library. The read side is used by the compiler driver
+to add those libraries to the link command.
+"""
+
+from jax._src.lib.mlir import ir
+
+RUNTIME_ARTIFACTS_ATTR = "catalyst.runtime_artifacts"
+
+
+def record_runtime_artifact(module_op, artifact_path):
```
multiphaseCFD, line 28: I was confused by the naming at first glance. I initially wondered what other kind of "artifact" this might record, but turns out that it's for collecting path/to/artifact. Would something `record_artifact_path` be more precise?

frontend/catalyst/utils/runtime_artifacts.py
```diff
@@ -0,0 +1,63 @@
...
+        if RUNTIME_ARTIFACTS_ATTR in attrs
+        else []
+    )
+    if artifact_path in existing:
+        return
+    existing.append(artifact_path)
+    attrs[RUNTIME_ARTIFACTS_ATTR] = ir.ArrayAttr.get([ir.StringAttr.get(p) for p in existing])
+
+
+def collect_runtime_artifacts(mlir_module, compile_options):
```
multiphaseCFD, line 42: A suffix like _path / _paths would help me perceive the function's job at a glance.

multiphaseCFD (approved):
Nice work @rniczh . A couple of minor suggestions.

frontend/catalyst/utils/runtime_artifacts.py
rniczh: The name follows the attribute it writes to: `RUNTIME_ARTIFACTS_ATTR = "catalyst.runtime_artifacts"`, the `runtime_artifacts.py` module, the read-side `collect_runtime_artifacts`, and `CompileOptions.runtime_artifacts`. Renaming this one will make the `record_*`/`collect_*` pair inconsistent. So it's not following the type of its argument

As for “what other kind of artifact,” that is what “runtime” disambiguates. We already use "artifact" to refer to compilation artifacts in the workspace (`.mlir`/`.ll` dumps) and cross-built executor bundles. Dropping the qualifier would therefore make the name less specific.

frontend/catalyst/utils/runtime_artifacts.py
rniczh: I don't think the suffix gets you what you're after? Your goal is to see the function's job at a glance, but the job is "gather what the linker needs," which is what `runtime_artifacts` names. `_paths` names the element type, and type isn't job; after the rename you'd know the tuple holds strings and still not know why the function exists.

Meanwhile, the functions currently read the same as the `catalyst.runtime_artifacts` attribute they write and the `CompileOptions.runtime_artifacts` they populate. I'd rather keep those three aligned.

PR #3102, opened 2026-08-05
Track graph operation IDs in decomposition resource analysis

sengthai:
**Context:**
Previously, `ResourceAnalysis` grouped quantum operations using only their name, wire(qubit) count, and parameter count, producing identifiers. Meanwhile graph-decompositions uses `graphOpId`, which also captures parameter types, wire structure, static data, and operator UIDs (some of these are part of Operator2). 

This PR aligns resource generation with `graphOpId` by adding an optional detailed operation count used specifically by decomposition passes. 

To use that, developer can toggle `collectDetailedOperation` when construct the analysis. 

[sc-125008]


mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -360,6 +365,9 @@ void ResourceAnalysis::analyzeRegion(Region &region, ResourceResult &result, boo
 
             if (needsCollection) {
                 collectOperation(&op, result, isAdjoint);
+                if (collectDetailedOperations) {
+                    collectDetailOperation(&op, result, isAdjoint);
+                }
```
sengthai, line 371: In ADR, we stated the mechanism that allow us to toggle between `quantum_operation` and `quantum_operation_detailed`, but in actual implementation, we still need to perform for `quantum_operation` by the way, because it needs to collect the function name for handling function call. Also, since `buildResourceDict()` serializes only `detailOperations` into MLIR attribute, we here just toggle only `quantum_operation_detailed`.

mlir/include/Catalyst/Analysis/ResourceAnalysis.h
```diff
@@ -89,6 +93,7 @@ class ResourceAnalysis {
 
     // categorize and count a single operation
     void collectOperation(Operation *op, ResourceResult &result, bool isAdjoint) const;
+    void collectDetailOperation(Operation *op, ResourceResult &result, bool isAdjoint) const;
```
jzaia18:
```suggestion
    void collectDetailedOperation(Operation *op, ResourceResult &result, bool isAdjoint) const;
```

To match the name of the flag above

mlir/include/Catalyst/Analysis/ResourceResult.h
```diff
@@ -40,6 +42,8 @@ struct ResourceResult {
     // name -> ((numWires, numParams) -> count)
     llvm::StringMap<llvm::DenseMap<std::pair<int, int>, double>> operations;
 
+    llvm::StringMap<double> detailOperations;
```
jzaia18:
```suggestion
    llvm::StringMap<double> detailedOperations;
```

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
jzaia18: If we needed to we could probably collect the function names using `CollectDetailedOperation`, but I think the current solution is fine, it shouldn't be much more expensive and saves us having 2 implementations.

mlir/lib/Catalyst/Analysis/ResourceResult.cpp
jzaia18, line 252: Doesn't this method only work if we collect detailed operations now? What if the user didn't select this setting?

jzaia18 (commented):
Looks good so far, just a few comments

mlir/test/Catalyst/RegisterDecompRuleResourceTest.mlir
```diff
@@ -106,7 +106,7 @@ func.func @rule_with_loop(%arg0: !quantum.bit) -> !quantum.bit attributes {targe
 
 // Rules with branching (take max per op)
 
-// CHECK: resources = {measurements = {}, num_alloc_qubits = 0 : i64, num_arg_qubits = 1 : i64, num_qubits = 1 : i64, operations = {"Hadamard(1,0)" = 3 : i64, "PauliX(1,0)" = 2 : i64}}
+// CHECK: resources = {measurements = {}, num_alloc_qubits = 0 : i64, num_arg_qubits = 1 : i64, num_qubits = 1 : i64, operations = {}}
```
jzaia18: I'm confused, why are the operations missing here?

mlir/test/Catalyst/RegisterDecompRuleResourceTest.mlir
jzaia18, line 1: This seems like we're replacing the existing unit tests with ones that check the detailed ops. I think we should instead add new unit tests -- we still want to test that the old pathway works as well

mlir/test/Catalyst/RegisterDecompRuleResourceTest.mlir
```diff
@@ -56,7 +56,7 @@ func.func @pbc_operations() attributes {target_gate="pbc"} {
 
 // Rule with measure
 
-// CHECK: resources = {measurements = {}, num_alloc_qubits = 1 : i64, num_arg_qubits = 0 : i64, num_qubits = 1 : i64, operations = {"Hadamard(1,0)" = 1 : i64, "MidCircuitMeasure(1,0)" = 1 : i64}}, target_gate = "gate"
+// CHECK: resources = {measurements = {}, num_alloc_qubits = 1 : i64, num_arg_qubits = 0 : i64, num_qubits = 1 : i64, operations = {"Hadamard[][1]{}" = 1 : i64}}, target_gate = "gate"
```
jzaia18: Why do we no longer expect the MCM to show?

mlir/test/Catalyst/RegisterDecompRuleResourceTest.mlir
```diff
@@ -153,7 +153,7 @@ func.func @rule_with_nested_loop(%arg0: !quantum.bit) -> !quantum.bit attributes
 // -----
 
 // Rules with parametric ops
-// CHECK:  operations = {"2C(Adjoint(Rot))(4,3)" = 1 : i64, "2C(Rot)(4,3)" = 1 : i64}}
+// CHECK:  operations = {"Rot[f64,f64,f64][2]{}" = 2 : i64}}
```
jzaia18: Seems like we lost information about adjoint? I would expect to still see the adjoint represented in the string repr somehow

mlir/test/Catalyst/RegisterDecompRuleResourceTest.mlir
```diff
@@ -35,7 +35,7 @@ func.func @basic_gates() attributes {target_gate="basic"}  {
 
 // Rule with PBC ops
 
-// CHECK: resources = {measurements = {}, num_alloc_qubits = 2 : i64, num_arg_qubits = 0 : i64, num_qubits = 2 : i64, operations = {"PPM(1,0)" = 2 : i64, "PPR-pi/4(1,0)" = 3 : i64, "PPR-pi/8(1,0)" = 1 : i64}}, target_gate = "pbc"
+// CHECK: resources = {measurements = {}, num_alloc_qubits = 2 : i64, num_arg_qubits = 0 : i64, num_qubits = 2 : i64, operations = {}}, target_gate = "pbc"
```
jzaia18: Seems like this doesn't work for PPMs or PPRs? Do GraphOpIDs not exist for these yet? If not, we should fall back to the regular format in these cases

mlir/test/Catalyst/RegisterDecompRuleResourceTest.mlir
sengthai: Fixed: [4b092a7](https://github.com/PennyLaneAI/catalyst/pull/3102/commits/4b092a76172940d496825a9dc548d7208930ec6e)

mlir/test/Catalyst/RegisterDecompRuleResourceTest.mlir
sengthai: Now i separate it by adding `[adj]` to that adjoint Rot. 
https://github.com/PennyLaneAI/catalyst/blob/4b092a76172940d496825a9dc548d7208930ec6e/mlir/test/Catalyst/RegisterDecompRuleResourceTest.mlir#L156-L157

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -470,10 +470,16 @@ void ResourceAnalysis::collectOperation(Operation *op, ResourceResult &result,
...
-void ResourceAnalysis::collectDetailOperation(Operation *op, ResourceResult &result,
-                                              bool isAdjoint) const {
+void ResourceAnalysis::collectDetailedOperation(Operation *op, ResourceResult &result,
+                                                bool isAdjoint) const {
     if (auto inst = dyn_cast<quantum::DecomposableGate>(op)) {
-        result.detailOperations[inst.getGraphOpId()] += 1;
+        std::string opId = inst.getGraphOpId();
+        if (auto resourceOp = dyn_cast<ResourceQuantumOpInterface>(op)) {
+            if (isAdjoint ^ resourceOp.getResourceAdjointFlag()) {
+                opId += "[adj]";
```
jzaia18: Will this break anything for the actual decomposition graph IDs?

CC @kipawaa

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -470,10 +470,16 @@ void ResourceAnalysis::collectOperation(Operation *op, ResourceResult &result,
...
+            }
+        }
+        result.detailedOperations[opId] += 1;
     }
```
jzaia18, line 486: Don't forget to add an else-branch to catch cases when we don't have a graph operator ID

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: FYI, it should be like `Rot[f64,f64,f64][2]{}[adj]`

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: I added another `if-branch` for handling the ResourceAnlaysis op type. [0a277f1](https://github.com/PennyLaneAI/catalyst/pull/3102/commits/0a277f1008f41afd1105c5b3cc58e832c779652c)

I'm not sure how we handle unknown quantum ops here. So, i would leave that alone for unknown op as it can be classical ops as well.

mlir/test/Catalyst/RegisterDecompRuleResourceTest.mlir
sengthai: Fix in [0a277f1](https://github.com/PennyLaneAI/catalyst/pull/3102/commits/0a277f1008f41afd1105c5b3cc58e832c779652c)

mlir/test/Catalyst/RegisterDecompRuleResourceTest.mlir
sengthai: I include it in [0a277f1](https://github.com/PennyLaneAI/catalyst/pull/3102/commits/0a277f1008f41afd1105c5b3cc58e832c779652c)

mlir/test/Catalyst/RegisterDecompRuleResourceTest.mlir
sengthai: what the old pathway you mention here? We actually replace the existing detailed ops fetching, completely.

mlir/lib/Catalyst/Analysis/ResourceResult.cpp
sengthai: If the user didn't enable `collectDetailedOperaitons` here.
https://github.com/PennyLaneAI/catalyst/blob/0a277f1008f41afd1105c5b3cc58e832c779652c/mlir/lib/Catalyst/Transforms/RegisterDecompRuleResourcePass.cpp#L39-L40

then, this return empty operations.

jzaia18: But that's a bug, isn't it? If we leave `collectDetailedOperations` as false then the output JSON will be completely missing all operations data

sengthai: i can see that. Being empty when `collectDetailedOperations` is false seems incorrect interpretation. Okay, i will add an assert if in that function then.

sengthai: [fbea7aa](https://github.com/PennyLaneAI/catalyst/pull/3102/commits/fbea7aa688c3ca16bf47f36c981b4fd89bf85d4c)

mlir/test/Catalyst/RegisterDecompRuleResourceTest.mlir
jzaia18: I mean testing when `collectDetailed` is set to false, so its collecting just the regular gate names

jzaia18 (commented):
Looking good! I'll take 1 more pass to look over the PR later today

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
kipawaa: We'll be adding adjoint and control to graph op IDs in the near future - right now this would break some of the graph system since the ID reported by `ResourceAnalysis` would differ from the ID generated by the op/used by decomposition rules. This isn't the end of the world, since the rest of the graph system isn't ready yet. How would you feel about leaving it out for now, and when we introduce it to graph op IDs we can review what should be changed here? I could update your lit tests etc. in that PR

kipawaa: My recommendation (though maybe this conflicts with other constraints, let me know) would be to just report the graph op IDs. The final layout of the ID is still being updated, but in the end it should be the most detailed summary of an operation in a standardized format.

mlir/lib/Catalyst/Transforms/RegisterDecompRuleResourcePass.cpp
```diff
@@ -35,8 +35,8 @@ struct RegisterDecompRuleResourcePass
     using RegisterDecompRuleResourcePassBase::RegisterDecompRuleResourcePassBase;
 
     void runOnOperation() final {
-        auto &analysis = getAnalysis<ResourceAnalysis, ModuleOp>();
         auto module = llvm::cast<mlir::ModuleOp>(getOperation());
+        ResourceAnalysis analysis(module, {}, /*collectDetailedOperations=*/true);
```
kipawaa, line 39: 🙌

mlir/lib/Quantum/Transforms/GraphDecomposition/graph_decomposition.cpp
```diff
@@ -240,7 +240,7 @@ struct GraphDecompositionPass : public impl::GraphDecompositionPassBase<GraphDec
 
         // Try to generate resources if they're missing
         if (!resourcesAttr) {
-            ResourceAnalysis analysis(rule);
+            ResourceAnalysis analysis(rule, {}, /*collectDetailedOperations=*/true);
```
kipawaa, line 245: Thank you! 🙏

kipawaa (approved):
Thanks @sengthai! Whichever way you decide on that discussion should be workable, so happy to approve now 😄

mlir/test/Catalyst/RegisterDecompRuleResourceTest.mlir
sengthai: I see. The regular pathway is done by `ResourceAnalysisTest.mlir` where ResourceAnalysisPass constructs `ResourceAnalysis` with the default `collectDetailedOperations=false`.  The register decomposition here pass users the detailed mode. I’ll add an explicit `CHECK-NOT: "quantum_operations_detailed"` in ResourceAnalysisTest to make the false-mode coverage clear.

sengthai: ✅
https://github.com/PennyLaneAI/catalyst/blob/9a993669420d55e35e1782bf223209bee049cca6/mlir/test/Catalyst/ResourceAnalysisTest.mlir#L15-L16

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: That makes sense. I'll report `getGraphOpID` as what it is, and remove the local `[adj]` suffix. The current Rot test will report as 2 instead of separate count. Once adjoint and control become part of the canonical graph ID, ResourceAnalysis will pick them up automatically.

https://github.com/PennyLaneAI/catalyst/blob/9a993669420d55e35e1782bf223209bee049cca6/mlir/test/Catalyst/RegisterDecompRuleResourceTest.mlir#L160-L161

sengthai: ✅
[256b3c0](https://github.com/PennyLaneAI/catalyst/pull/3102/commits/256b3c02938804bf80f8c2462d73a95d89f835de)

jzaia18 (approved):
Seems good, feel free to merge once the other specs-related PRs are in

PR #3103, opened 2026-08-07
feat(backline): add local transports 

kszenes:
Add support for local controllers/coprocessors:
- CPU -> CPU
- CPU -> GPU

Currently, the GPU launches a kernel for each transaction. Eventually, a persistent kernel for the local case could be implemented as well.

Additionally, this PR removes the `data_path` attribute and relies on the `transport` to properly dispatch to the communication protocol.

### Before submitting

**Context:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


runtime/lib/transport/local_copy/LocalWireProtocol.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+// Header stored at the front of the controller's local reply region. The reply payload bytes
+// follow immediately after this header.
+struct LocalReplyHeader {
+    std::uint64_t bytes = 0;
+};
+
+inline constexpr std::size_t kLocalRequestHeaderBytes = sizeof(LocalRequestHeader);
+inline constexpr std::size_t kLocalReplyHeaderBytes = sizeof(LocalReplyHeader);
+
```
kszenes, line 37: Do we want to hardcode this?

mlir/test/Transport/InjectTransportSessionInvalid.mlir
```diff
@@ -29,8 +29,37 @@ module attributes {catalyst.backline = #transport.backline<transport = "net", co
...
+// -----
+
+// Nor can it dispatch a coprocessor remotely.
+
+// expected-error @below {{local transport does not support remote coprocessors}}
+module attributes {catalyst.backline = #transport.backline<transport = "local", controller = #transport.node<backend_lib = "x">,
+  coprocessors = [#transport.node<backend_lib = "y", peer = "10.0.0.3", symbol = "foo", remote = true>]>} {
+  func.func @setup() { quantum.init  return }
+  func.func @teardown() { quantum.finalize  return }
+}
```
kszenes, line 65: Maybe instead of `local` for the transport protocol, we could be more specific and name it `memcpy`?

mlir/include/Transport/IR/TransportDialect.td
```diff
@@ -109,7 +109,6 @@ def Transport_NodeAttr : Transport_Attr<"Node", "node"> {
         OptionalParameter<"mlir::IntegerAttr">:$oob_port,
```
josephleekl, line 109: don't forget a changelog entry, and also add your name to the changelog if not already there!

mlir/lib/Transport/IR/TransportDialect.cpp
```diff
@@ -89,6 +88,9 @@ LogicalResult BacklineAttr::verify(function_ref<InFlightDiagnostic()> emitError,
         if (!c.getSymbol() || c.getSymbol().getValue().empty()) {
             return emitError() << "coprocessor requires a 'symbol'";
         }
+        if (transport.getValue() == "local" && c.isRemote()) {
+            return emitError() << "local transport does not support remote coprocessors";
+        }
```
josephleekl, line 93: actually in theory local transport could be on a remote executor? so you run pennylane/catalyst on one machine, and you have a remote executor running the cpu/gpu computation, and they communicate 'locally'

runtime/lib/transport/local_copy/cpu_controller/LocalCpuControllerSession.cpp
```diff
@@ -0,0 +1,197 @@
+// Copyright 2026 Xanadu Quantum Technologies Inc.
```
josephleekl, line 1: JB: The coprocessor writes `pair_->coprocessor = this` in `connect()` but its destructor is defaulted, so the pointer dangles once the coprocessor dies — and `finalize()` tears the coprocessor down before the controller. `kick()` then dereferences freed memory; ASan reports heap-use-after-free at `LocalCpuControllerSession.cpp:167`.

mlir/lib/Transport/IR/TransportDialect.cpp
kszenes: How would that work? I'm confused. They communicate via `memcpy`, doesn't that need to be non-remote?

josephleekl: if the coprocessor and controller are on the same remote machine (so i run pennylane/catalyst on machine A, and there's a remote machine B that has a CPU controller as well as GPU coprocessor), they can memcpy to each other. So for example you compile on your laptop, but run on the dell server for both the cpu controller and gpu coprocessor. I believe this is what remote means. does that make sense? we could also ignore this use case.

multiphaseCFD: > actually in theory local transport could be on a remote executor?

I agree. The remote @josephleekl mentioned above means user dispatch executor from a local host to remote server, whereas the coprocessor/controller are running locally on the same remote server.

mlir/test/Transport/SmokeTest.mlir
```diff
@@ -35,10 +35,10 @@ func.func @transport_smoketest(%payload: memref<?xi8>, %reply: memref<?xi8>) {
...
   transport.exchange_keys %ct : !transport.session<controller>
 
-  // CHECK: transport.establish_channel %{{.*}} "cpu_verbs" : !transport.session<controller>
-  transport.establish_channel %ct "cpu_verbs" : !transport.session<controller>
-  // CHECK: transport.establish_channel %{{.*}} "gpu_engine" : !transport.session<coprocessor>
-  transport.establish_channel %co "gpu_engine" : !transport.session<coprocessor>
+  // CHECK: transport.establish_channel %{{.*}} "rdma" : !transport.session<controller>
+  transport.establish_channel %ct "rdma" : !transport.session<controller>
+  // CHECK: transport.establish_channel %{{.*}} "local" : !transport.session<coprocessor>
+  transport.establish_channel %co "local" : !transport.session<coprocessor>
```
multiphaseCFD, line 41: would the combination of `%ct "rdma"` and `%co "local"`  be allowed?

PR #3106, opened 2026-08-10
pin setuptools version

kipawaa:
**Context:**
Setuptools version 84 was released on the weekend, which breaks catalyst's `libcustom_calls.so` resolution on macOS.

Pin setuptools to version 83, and format the pyproject.toml file.

**Benefits:**
Working dylibs on macos.

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,18 @@
+[build-system]
+requires = [
+  "cmake",
+  "nanobind",
+  "ninja",
+  "numpy>2.0.0",
+  "pybind11>=2.12.0",
+  "setuptools==83",
+  "wheel"
+]
+build-backend = "setuptools.build_meta"
+
+[tool.bandit]
+skips = ["B607"]
+
 [tool.black]
 line-length = 100
 extend-exclude = '''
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -1,3 +1,18 @@
+[build-system]
```
paul0403, line 1: The formatting was just sort the fields alphabetically right?

pyproject.toml
```diff
@@ -1,3 +1,18 @@
...
+requires = [
+  "cmake",
+  "nanobind",
+  "ninja",
+  "numpy>2.0.0",
+  "pybind11>=2.12.0",
+  "setuptools==83",
```
paul0403: Can we do a range instead of an exact version pin? i.e. 62~83

paul0403 (approved):

mlxd:
[sc-127342]

pyproject.toml
kipawaa: [4b10543](https://github.com/PennyLaneAI/catalyst/pull/3106/commits/4b10543aaf13d1e7dc37d25b012e5d459813e9f4)

pyproject.toml
kipawaa: yep!

paul0403 (approved):

mlxd (approved):

PR #3107, opened 2026-08-10
Migrate specs integration tests to PennyLane

jzaia18:
**Context:**
Currently, the integration tests for `qp.specs` live within the Catalyst frontend tests. This has caused headache when adding new features since it increases the total number of PRs needed to implement a new feature in specs. These tests should be migrated into PennyLane, which soon will assume the presence of Catalyst for its tests anyway.

In conjunction with https://github.com/PennyLaneAI/pennylane/pull/9988

- Removes specs integration tests from Catalyst frontend tests.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-127285]
https://github.com/PennyLaneAI/pennylane/pull/9988


jzaia18:
One question I have for the reviewers here: does it maybe make sense to leave a couple (simple) integration tests for specs within Catalyst? My concern is that if we remove all of the integration tests then we don't have any sanity check in the CIs that changes to the resource analysis pass won't break specs.

On the other hand, leaving some integration tests here might necessitate additional PRs to fix them when there are changes to the PL frontend

frontend/test/pytest/test_specs.py
AntonNI8, line 1: Regarding your question, I think it's preferable to go all-in on the migration.
If `pennylane` functionality changes, we'll see it there.
If `catalyst` functionality (at least I/O) changes, we should have some other methodology to verify that it's still `pennylane` compatible. As a rough idea, perhaps `catalyst` CI should include a subset of `pennylane` CI where these integration tests are performed, as an optional check.

AntonNI8 (approved):

sengthai (approved):
Thanks @jzaia18 !

PR #3113, opened 2026-08-12
Refactor transport local copy

multiphaseCFD:
### Before submitting

**Context:**
[sc-127490]
A `memcpy` `transport` backend is added to the `runtime`. The new `memcpy` use direct memory access via C/C++ lib or GPU libs for the data communication between `controller` and `coprocessor` instead of RDMA. This feature would allow users to quick prototype without relying on the specific hardware setups.

- Runtime: memcpy backends (cpu_controller, cpu_coprocessor, gpu_coprocessor, MemcpyLink)
- Refactor: common/{ → rdma/} move + hoist shared coproc/interface code out of gpu_verbs/ into common/coproc/ and common/interface/
- Frontend + MLIR: memcpy mode plumbed through backline.py, transport dialect verifier, typed port attr
- Tests: new C++ memcpy tests + backline pytest
- CI: check-transport-gpu.yaml on single-gpu-x64, gpu-label gated

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


doc/releases/changelog-dev.md
```diff
@@ -78,6 +78,9 @@
   syndrome came from, which `lower-decode-to-transport` maps to a `decoder_id` on `transport.kick`.
   [(#3092)](https://github.com/PennyLaneAI/catalyst/pull/3092)
 
+* A backline transport backend via `memcpy` is available.
```
josephleekl:
```suggestion
* CPU & GPU backline transport backends via `memcpy` are added.
```

runtime/lib/transport/memcpy/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
+###############################################
+# Memcpy transport backend                    #
+###############################################
+
+add_library(memcpy_cpu_impl STATIC
```
josephleekl: JB: The memcpy link registry lives in function-local statics in MemcpyLink.hpp:51, but memcpy_cpu_impl is a STATIC lib linked PRIVATE into both plugin .sos, so each role gets its own registry. It only pairs today because GCC emits those statics as STB_GNU_UNIQUE; rebuild with clang (or on macOS, which backline.py:40 targets via .dylib) and kick() throws "no paired coprocessor". Making memcpy_cpu_impl SHARED fixes it — verified.

runtime/lib/transport/memcpy/cpu_coprocessor/CpuCoprocessorSession.cpp
```diff
@@ -0,0 +1,107 @@
...
+    }
+    return n;
+}
+
+} // namespace
+
+CpuCoprocessorSession::~CpuCoprocessorSession() {
+    if (link_) {
+        // Wait for any in-flight kick, then unbind so no future call reaches a dying `this`.
+        std::lock_guard<std::mutex> lock(link_->mu);
```
josephleekl, line 44: JB: `runtime/lib/transport/memcpy/cpu_coprocessor/CpuCoprocessorSession.cpp:34-40` — the destructor unconditionally clears `link_->process_message`. But `connect()` (line 43) assigns `link_ = acquire_memcpy_link(info)` before the duplicate-binding check at line 45 throws. So the rejected session holds a valid `link_`, and its destructor wipes the incumbent coprocessor's binding. Same bug at `gpu_coprocessor/GpuCoprocessorSession.cpp:39-43`. The controller destructor at `cpu_controller/CpuControllerSession.cpp:34-41` gets this right — it guards with `if (link_->controller == this)`.

Trace: the existing test `runtime/tests/Test_TransportMemcpyCpu.cpp:118` ("rejects a second coprocessor") asserts the throw but never kicks afterward, so it passes while leaving the link broken.


Fix: mirror the controller's guard. Either only assign `link_` after the check passes, or make the destructor conditional by storing an ownership flag / comparing a bound-owner pointer:
```
// in connect(), after the duplicate check succeeds:
bound_ = true;
// in ~CpuCoprocessorSession():
if (link_ && bound_) { std::lock_guard lock(link_->mu); link_->process_message = nullptr; }
```

Summary:
`~CpuCoprocessorSession` (`CpuCoprocessorSession.cpp:38`) clears `link_->process_message` unconditionally, but `connect()` assigns `link_` at line 43 before the duplicate check throws at 45 — so a rejected second coprocessor tears down the valid one's binding on the way out. The controller destructor already guards this with if `(link_->controller == this)`; the coprocessor (and the GPU one at line 42) should do the same.

runtime/lib/transport/memcpy/cpu_controller/CpuControllerSession.cpp
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    kick_ns_ = now_ns();
+
+    // Synthesize a wire-shaped Payload so `in` looks the same as it does over cpu_verbs:
+    // value bytes at offset 0, decoder_id at PAYLOAD_DATA_BYTES, then seq_num.
+    common::Payload frame{};
+    if (staged_bytes_ != 0) {
+        std::memcpy(&frame.value, request_staging_.data(),
+                    std::min<std::size_t>(static_cast<std::size_t>(staged_bytes_),
```
josephleekl, line 125: commit_work_item (`memcpy/.../CpuControllerSession.cpp:110`) accepts any `in_bytes`, but `kick()` at line 143 truncates the copy to `PAYLOAD_DATA_BYTES` (8) with no error — so `in_bytes=32` silently loses 24 bytes. `rdma/cpu_verbs/.../CpuControllerSession.cpp:54` rejects this outright; memcpy should match, otherwise the two backends disagree on the payload contract and a circuit that errors on rdma returns wrong data on memcpy.

Let's add a guard in `commit_work_item`:
```
if (in_bytes > common::PAYLOAD_DATA_BYTES || out_bytes > common::PAYLOAD_DATA_BYTES) {
    throw std::runtime_error("memcpy: in/out_bytes exceeds the 8 B payload data area");
}
```

runtime/lib/transport/CMakeLists.txt
```diff
@@ -30,26 +30,41 @@ add_custom_target(transport_all)
...
 else()
     message(STATUS
-        "transport: libibverbs not found -> building rt_transport only. The bundled backends "
-        "are skipped; install libibverbs-dev / rdma-core to build them.")
+        "transport: libibverbs not found -> skipping the RDMA backends (cpu_verbs, gpu_verbs, "
+        "and the shared transport_common); rt_transport and the memcpy backend still build. "
+        "Install libibverbs-dev / rdma-core to get the RDMA backends.")
 endif()
 
 add_custom_target(transport_backends
```
josephleekl: JB: runtime/lib/transport/CMakeLists.txt:70-75 — the DEPENDS list unconditionally names catalyst_transport_cpu_verbs_controller, ..._coprocessor, and steane_coprocessor_cpu, but lines 32-68 only define those inside if(IBVERBS_LIB). The new else() message at lines 64-67 explicitly promises "rt_transport and the memcpy backend still build" — but cmake --build --target transport_backends fails with ninja: error: 'catalyst_transport_cpu_verbs_controller', needed by 'CMakeFiles/transport_backends', missing and no known rule to make it. I verified that behaviour with a minimal reproduction. Configure still succeeds and make runtime uses transport_all (runtime/Makefile:60), which the PR does guard correctly — so this only bites anyone invoking transport_backends directly. Pre-existing, but the new message now advertises a path that does not work.
Fix: guard the transport_backends deps the same way transport_all is guarded, moving the cpu_verbs entries inside the if(IBVERBS_LIB) block and adding the memcpy targets.

runtime/lib/transport/CMakeLists.txt
```diff
@@ -30,26 +30,41 @@ add_custom_target(transport_all)
...
         message(STATUS "transport: HIP found (${CMAKE_HIP_COMPILER}) -> building gpu_verbs")
+        enable_language(HIP)
+        if(DEFINED CMAKE_HIP_PLATFORM)
+            set(HIP_PLATFORM "${CMAKE_HIP_PLATFORM}" CACHE STRING "HIP platform" FORCE)
+        endif()
+        find_package(hip REQUIRED)
+        if(CMAKE_HIP_COMPILER_ID STREQUAL "NVIDIA")
+            find_package(CUDAToolkit REQUIRED)
+        endif()
         set(TRANSPORT_HAS_HIP TRUE CACHE INTERNAL "transport: HIP toolchain available")
```
josephleekl: TRANSPORT_HAS_HIP being set inside if(IBVERBS_LIB) means the memcpy GPU coprocessor cannot be built without libibverbs, even though it has no RDMA dependency.

We should re-organize this cmakelist to reflect what we intend, so by default, always build cpu memcpy, if has HIP, build gpu memcpy too, if has ibverb, build cpu rdma, and if also has hip, build gpu rdma too.

mlir/lib/Transport/IR/TransportDialect.cpp
```diff
@@ -89,6 +85,10 @@ LogicalResult BacklineAttr::verify(function_ref<InFlightDiagnostic()> emitError,
         if (!c.getSymbol() || c.getSymbol().getValue().empty()) {
             return emitError() << "coprocessor requires a 'symbol'";
         }
+        if (transport.getValue() == "memcpy" && c.isRemote() != controller.isRemote()) {
```
josephleekl: in theory if they're both remote, they should have the same address too?

runtime/lib/transport/memcpy/MemcpyLink.hpp
```diff
@@ -0,0 +1,77 @@
...
+#include <string>
+#include <unordered_map>
+
+#include "Transport.hpp"
+
+namespace catalyst::transport::memcpy {
+
+class CpuControllerSession;
+
+// In-process memcpy link keyed by (peer, oob_port). The coprocessor binds `process_message`;
```
josephleekl: wait, i don't think this makes sense - memcpy shouldn't depend on user supplying peer/oob_port! Can we use some other way to link them?

josephleekl: these two options are optional from the frontend. 

The following is JB's suggestion:
Apparently `InjectTransportSession.cpp:144` already computes one key per processor: 
```
StringAttr key = coproc.keyOr("coprocessor." + std::to_string(index));
```
and emitColocated (line 308-310) passes that same key to createSession for both roles:
```
Value ct = createSession(b, ctrlTy, ctrl, key);   // controller
Value co = createSession(b, coTy, coproc, key);   // coprocessor
```

That is exactly the pairing relation you want, already computed by the compiler, already unique per pair, and already threaded through transport.create's key attribute into __catalyst__transport__create(..., role, key) (TransportCAPI.cpp:223). The runtime even keeps a (role, key) registry at TransportCAPI.cpp:78. The memcpy backend reinvents a worse key while the correct one sits one argument away — the factory signature just drops it, since create passes only config to the backend.

key the link by the session key (recommended). Plumb key into the backend factory (or fold it into the config string as pair=<key>, which needs no ABI change and parse already supports via for_each_config_kv), then key acquire_memcpy_link on it.


Please try and see if this works

runtime/lib/transport/memcpy/cpu_controller/CpuControllerSession.cpp
```diff
@@ -0,0 +1,183 @@
...
+        throw std::runtime_error(
+            "memcpy: another controller is already bound to this endpoint (peer, oob_port)");
+    }
+    link_->controller = this;
+    return 0;
+}
+
+MemRegion CpuControllerSession::alloc_memory(std::size_t size, MemKind kind) {
+    if (kind != MemKind::CpuRam) {
+        throw std::runtime_error("memcpy: CPU-only for now; alloc_memory expects CpuRam");
```
josephleekl:
```suggestion
        throw std::runtime_error("CPU device can only allocate CpuRam");
```

josephleekl (commented):
First part of review

mlir/lib/Transport/IR/TransportDialect.cpp
multiphaseCFD: Good point, done!

mlir/lib/Transport/IR/TransportDialect.cpp
```diff
@@ -89,6 +82,24 @@ LogicalResult BacklineAttr::verify(function_ref<InFlightDiagnostic()> emitError,
...
+                c.getAddress().getValue() != controller.getAddress().getValue()) {
+                return emitError() << "memcpy transport with remote nodes requires controller and "
+                                      "coprocessor on the same executor address (got '"
+                                   << controller.getAddress().getValue() << "' vs '"
+                                   << c.getAddress().getValue() << "')";
+            }
+        }
+    }
+    if (!controller.getInBytes() || !controller.getOutBytes()) {
+        return emitError() << "controller requires 'in_bytes' and 'out_bytes'";
```
josephleekl: actually why are we checking in an MLIR pass? why not just check in the python frontend?

runtime/lib/transport/CMakeLists.txt
```diff
@@ -22,42 +22,87 @@ set_property(TARGET rt_transport PROPERTY POSITION_INDEPENDENT_CODE ON)
 ###############################################
 # transport backends                          #
 ###############################################
+#
+# Build matrix:
+#   * memcpy CPU       — always built (no external deps).
+#   * memcpy GPU       — built when HIP + libibverbs are both present. The dependency on
+#                        libibverbs is transitive via gpu_verbs (memcpy_gpu currently reuses
+#                        GpuRuntime/HipCheck/GpuLaunchers from rdma/gpu_verbs); extracting those
+#                        into a shared GPU lib is a follow-up.
```
josephleekl: this shouldn't depend on libibverbs, we should probably update that now

runtime/lib/transport/memcpy/gpu_coprocessor/GpuCoprocessorSession.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+    CoprocLaunchDesc desc{
+        .ring = request_slot_dev_,
+        .ring_slots = 1,
+        .handoff = handoff_.dev,
+        .stop = handoff_.stop_dev,
+        .total = 1,
+        .stream = gpu_->stream(),
+    };
+    CoprocessorLauncherFn launch = launcher_ ? launcher_ : &gpu_verbs::default_echo_launcher;
```
josephleekl: why do we do 1 slot at a time, and launch the kernel for each message, instead of do the same way as rdma device with a ring buffer to poll that? is it much more difficult to do that instead? (it should be very similar to the rdma device). If that change is very big/difficult, let's put a TODO here to follow up on that.

runtime/lib/transport/memcpy/gpu_coprocessor/GpuCoprocessorSession.cpp
```diff
@@ -0,0 +1,172 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "GpuCoprocessorSession.hpp"
+
+#include <cstring>
+#include <stdexcept>
+
+#include "GpuLaunchers.hpp"
+#include "HipCheck.hpp"
```
josephleekl, line 24: let's move these guys out of rdma and put them somewhere common so this build doesn't depend on rdma?

runtime/lib/transport/rdma/cpu_verbs/cpu_verbs_selftest.cpp
```diff
@@ -79,7 +79,7 @@ int main(int argc, char **argv) {
     MemRegion m = s->alloc_memory(REGION_BYTES, MemKind::CpuRam);
```
josephleekl, line 79: not about this file.

in theory all the coprocessor functions could be pulled out of rdma right? in theory the new cpu and gpu memcpy devices should be able to re-use the cpu/gpu rdma coprocessor functions straight out of the box?
can you test that this is true?

josephleekl: and can you add tests to check that this works end to end for cpu->cpu->cpu and cpu->gpu->cpu?

runtime/tests/Test_TransportMemcpyCpu.cpp
```diff
@@ -0,0 +1,200 @@
...
+using namespace catalyst::transport;
+using namespace catalyst::transport::memcpy;
+
+namespace {
+// Per-test pair keys. Matching keys between the controller and its paired coprocessor is what
+// the memcpy backend uses to rendezvous, mirroring what inject-transport-session emits at
+// compile time as the `key` attribute on each transport.create.
+std::string pair_cfg(std::uint16_t port) { return "pair=p" + std::to_string(port); }
+
+std::size_t invert_fn(const void *in, std::size_t in_len, void *out, std::size_t out_cap,
```
josephleekl, line 40: nice. can you also try and test the existing (e.g. steane) coprocessor fn too?

.github/workflows/check-catalyst-gpu.yaml
```diff
@@ -0,0 +1,151 @@
...
+        with:
+          repository: llvm/llvm-project
+          ref: ${{ needs.constants.outputs.llvm_version }}
+          path: mlir/llvm-project
+
+      - name: Set up ccache for the runtime rebuild
+        uses: hendrikmuhs/ccache-action@v1.2
+        with:
+          key: ${{ github.workflow }}-${{ runner.os }}-runtime
+          max-size: "500M"
```
josephleekl: is this the proper way to do it? can you look reference other workflow files to make sure it aligns with how it's done? i'm worried about introducing this new untested action

.github/workflows/check-transport-gpu.yaml
```diff
@@ -0,0 +1,151 @@
+name: Check Catalyst GPU
```
josephleekl, line 1: can you double check that this workflow aligns with the other workflow files in catalyst? it looks a bit different/questionable.

josephleekl (commented):
Thanks @multiphaseCFD for this! I've left some more comments, please tag me after addressing them / cleaning up (there seems to be some generated comments, please clean them up). Also we may need to restructure some of the files to move them out of rdma.

runtime/lib/transport/rdma/cpu_verbs/cpu_verbs_selftest.cpp
multiphaseCFD: Move the coprocess fn out and let's see

runtime/lib/transport/CMakeLists.txt
multiphaseCFD: Good point! Done

runtime/lib/transport/memcpy/MemcpyLink.hpp
multiphaseCFD: good catch! done!

runtime/lib/transport/memcpy/gpu_coprocessor/GpuCoprocessorSession.cpp
multiphaseCFD: Done!

mlir/lib/Transport/IR/TransportDialect.cpp
multiphaseCFD: this isn't a pass, it's an attribute verifier. It's about well-formed IR: hand-written .mlir, quantum-opt, still hit the verifier, and downstream passes rely on it. A Python-side check would be complementary, not a substitute.

runtime/tests/Test_TransportMemcpyCpu.cpp
multiphaseCFD: https://github.com/PennyLaneAI/catalyst/pull/3113/commits/6cf552d14973de734b67f540a62d6b68192e52a0

.github/workflows/check-catalyst-gpu.yaml
multiphaseCFD: removed

.github/workflows/check-transport-gpu.yaml
multiphaseCFD: update the workflow to follow other CIs in Catalyst

rniczh (approved):
LGTM

frontend/catalyst/backline.py
```diff
@@ -180,6 +209,59 @@ def _load_coprocessor_fn_libs(placement: Placement) -> None:
...
+    return stages
+
+
+def backline_pipeline():
+    """Default Catalyst pipeline plus the transport passes."""
+    from catalyst.pipelines import default_pipeline
+
+    return add_transport_passes(default_pipeline())
+
+
```
multiphaseCFD: Do we really need add those methods in this PR @kszenes ? I thought they were implemented in some other branch? @mehrdad2m

frontend/catalyst/backline.py
```diff
@@ -269,6 +352,14 @@ def _check_machine_agrees(node: Node, host, address=None, preset: bool = False)
         )
 
 
+def _launch_executor(name, options):
```
multiphaseCFD: same question

frontend/catalyst/backline.py
multiphaseCFD: Follow @mehrdad2m 's PR 3068, we should remove them

frontend/catalyst/backline.py
```diff
@@ -111,29 +132,34 @@ def _out_of_process(node: Node) -> bool:
...
+
+    endpoint = getattr(node, "endpoint", None)
+    if endpoint is not None:
+        d["peer"] = endpoint.host
+        if endpoint.port is not None:
+            d["oob_port"] = endpoint.port
+
+    hardware = getattr(node, "hardware", None)
+    if hardware is not None and not (node.init_args or {}).get("backend_lib"):
+        d["backend_lib"] = _resolve_backend_lib(transport, hardware, role, bool(node.remote))
```
multiphaseCFD, line 168: Is the current implementation aligned with the frontend UI? @kszenes

runtime/lib/transport/rdma/cpu_verbs/CMakeLists.txt
```diff
@@ -24,11 +24,8 @@ target_link_libraries(catalyst_transport_cpu_verbs_controller PRIVATE cpu_verbs_
 add_library(catalyst_transport_cpu_verbs_coprocessor SHARED coprocessor/CpuCoprocessorFactory.cpp)
 target_link_libraries(catalyst_transport_cpu_verbs_coprocessor PRIVATE cpu_verbs_impl)
 
-add_library(steane_coprocessor_cpu SHARED coprocessor/coprocessor_functions/steane_decoder_fn.cpp)
-set_property(TARGET steane_coprocessor_cpu PROPERTY POSITION_INDEPENDENT_CODE ON)
-
-add_library(echo_coprocessor_cpu SHARED coprocessor/coprocessor_functions/echo_coprocessor_fn.cpp)
-set_property(TARGET echo_coprocessor_cpu PROPERTY POSITION_INDEPENDENT_CODE ON)
+# The steane / echo CoprocessorFn plugin libraries have moved to transport/common/cpu_coproc_common/
```
multiphaseCFD: Let's move it

runtime/lib/transport/rdma/gpu_verbs/CMakeLists.txt
```diff
@@ -1,46 +1,28 @@
...
-elseif(EXISTS "/opt/rocm")
-    list(APPEND CMAKE_PREFIX_PATH "/opt/rocm")
-endif()
-
-find_package(hip REQUIRED)
-enable_language(HIP)
-if(CMAKE_HIP_COMPILER_ID STREQUAL "NVIDIA")
-    find_package(CUDAToolkit REQUIRED)
-endif()
+# Shared GPU plumbing (GpuRuntime, HipCheck, GpuLaunchers) lives in transport/common/coproc/gpu/
```
multiphaseCFD: Let's move it

.github/workflows/check-transport-gpu.yaml
```diff
@@ -0,0 +1,117 @@
...
```
multiphaseCFD, line 1: We might need to update workflow file name

frontend/catalyst/backline.py
kszenes: @multiphaseCFD, yes it should be! I tried running the demos, and I resolved all the UI inconsistencies (demos still failed due to some, I believe unrelated, SSH) issues

runtime/lib/transport/memcpy/cpu_coprocessor/CpuCoprocessorSession.cpp
```diff
@@ -0,0 +1,198 @@
...
+    // Spin-wait for the worker to publish reply_ring_[idx] with the matching seq.
+    volatile std::uint32_t *sseq = &reply_ring_[idx].p.seq_num;
+    while (*sseq != expect) {
+        if (failed_.load(std::memory_order_acquire)) {
+            std::rethrow_exception(error_);
+        }
+    }
+    std::atomic_thread_fence(std::memory_order_acquire);
+
+    const std::size_t n = std::min(out_cap, common::PAYLOAD_DATA_BYTES);
```
josephleekl, line 185: JB: `process_message` returns `min(out_cap, PAYLOAD_DATA_BYTES)` instead of the nb the coprocessor fn actually returned, so the length `run()` computed at line 151 is discarded. Any decoder that writes fewer than 8 bytes either trips a spurious "caller reply buffer too small" in collect or silently overwrites the caller's buffer with zeros.

Potential fix: record `nb` per slot (`std::array<std::uint32_t, K_RING_SLOTS> reply_len_`), write it before the release fence, return it clamped/checked against `out_cap`.

However I know that @kszenes wants to change the return value of the coprocessor fn (not return `nb`) - please check with him about what to do here

runtime/lib/transport/memcpy/cpu_coprocessor/CpuCoprocessorSession.cpp
```diff
@@ -0,0 +1,198 @@
...
+    if (engine_.joinable()) {
+        engine_.request_stop();
+        engine_.join();
+    }
+}
+
+void CpuCoprocessorSession::set_coprocessor_fn(CoprocessorFn fn, void *ctx) {
+    fn_ = fn;
+    ctx_ = ctx;
+}
```
josephleekl, line 123: JB: `fn_` and `ctx_` are plain members written by `set_coprocessor_fn` while the worker thread launched in start() reads them at lines 150-151 — TSAN flags this as a data race. The new tests themselves bind the fn after `start()`, so the shipped test order is the racy one; on a weak-memory host the worker can run the built-in echo instead of the real decoder.

Potential fix: `fn_`/`ctx_` `std::atomic` (release store / acquire load), or reject `set_coprocessor_fn` once `engine_.joinable()` and fix the tests to bind before `start()`.

frontend/test/pytest/test_backline.py
```diff
@@ -306,7 +374,7 @@ def test_remote_controller_behind_a_wrapper_is_still_tagged(use_capture):
     """A remote controller called through a wrapper still gets its module tagged by role."""
     ctrl = qp.Controller(
         device=qp.device("null.qubit", wires=2),
-        label="ctrl",
+        name="ctrl",
         remote=True,
         executor_options={"address": "ctrl:1"},
         init_args={"backend_lib": "backend.so", "config": "cfg", "data_path": "cpu_verbs"},
```
josephleekl: i think this it out of date, can you update this?

frontend/test/pytest/test_backline.py
```diff
@@ -235,13 +275,41 @@ def circuit():
...
+    assert "catalyst.backline" in ir
+    assert 'transport = "memcpy"' in ir
 
 
 def test_placement_behind_a_wrapper_is_found(use_capture):
-    """A placement reaches the module even when qjit was applied to a wrapper, not the QNode."""
+    """A placement reaches the module even when qjit was applied to a wrapper, not the QNode.
+    The transport passes locate it by role rather than by matching a triple or address, which now
+    come only from the node's executor."""
     dev = qp.Backline(controller=_controller(), coprocessors=[_coproc("cop0")], transport="net")
```
josephleekl: delete this?

frontend/test/pytest/test_backline.py
```diff
@@ -464,75 +564,115 @@ def test_out_of_tree_backend_via_search_path(self, tmp_path, monkeypatch):
...
+        fake_lib_dir("rdma/cpu_verbs/libcatalyst_transport_cpu_verbs_controller.so")
+        ctrl = qp.Controller(device=qp.device("null.qubit", wires=2), name="ctrl")
+        d = serialize_backline(qp.Backline(controller=ctrl, transport="rdma").placement)
+        assert d["controller"]["backend_lib"].endswith("_cpu_verbs_controller.so")
+
     def test_no_backend_leaves_backend_lib_unset(self, no_launch):
         """Omitting ``backend`` leaves the field to ``init_args`` or the compiler default."""
-        ctrl = qp.Controller(device=qp.device("null.qubit", wires=2), label="ctrl")
+        ctrl = qp.Controller(device=qp.device("null.qubit", wires=2), name="ctrl")
         d = serialize_backline(qp.Backline(controller=ctrl, transport="net").placement)
```
josephleekl: let's not use transport="net" which will fail. Can you update across all the tests?

runtime/lib/transport/TransportCAPI.cpp
```diff
@@ -229,15 +229,26 @@ CatalystTransportSession *__catalyst__transport__create(const char *backend_lib,
         }
         auto h = std::make_unique<CatalystTransportSession>();
         h->backend = std::make_unique<DynamicLibraryLoader>(backend_lib);
-        const char *cfg = config ? config : "";
+        // Fold the session key (unique per controller/coprocessor pair, as emitted by
+        // inject-transport-session in MLIR) into the config as `pair=<key>` so backends that
+        // need to pair endpoints in-process (e.g. memcpy) share one identifier for both
+        // sides. This is transparent to backends that don't parse `pair`.
+        std::string cfg = config ? config : "";
```
josephleekl: JB: we may have to guard against names with `;` in it: 

suggestion something like: 
```
{
    std::string cfg = config ? config : "";
    if (!key || !*key) {
        return cfg;
    }
    const std::string_view k(key);
    if (k.find(';') != std::string_view::npos) {
        throw std::runtime_error("session key must not contain ';' (got '" + std::string(k) +
                                 "'); it would split the backend config into two entries");
    }
    // Ours would shadow a caller-supplied `pair=` under last-wins, so refuse instead.
    bool reserved_in_use = false;
    catalyst::transport::common::configparser::for_each_kv(
        cfg, [&](std::string_view entry_key, std::string_view) {
            if (entry_key == "pair") {
                reserved_in_use = true;
            }
        });
    if (reserved_in_use) {
        throw std::runtime_error("backend config must not set 'pair'; it is reserved for the "
                                 "compiler-emitted session key");
    }
    if (!cfg.empty()) {
        cfg += ";";
    }
    cfg += "pair=";
    cfg += k;
    return cfg;
}
```
and
```
const std::string cfg = fold_pair_key(config, key);
```

please see if it's useful

.github/workflows/check-transport-gpu.yaml
```diff
@@ -0,0 +1,122 @@
...
+          cmake --build "$(pwd)/runtime-build" --target runner_tests_transport_memcpy -j
+
+      - name: Run memcpy transport tests
+        run: |
+          "$(pwd)/runtime-build/tests/runner_tests_transport_memcpy"
+
+      - name: Clean up runtime build artifacts
+        if: always()
+        run: |
+          rm -rf "$(pwd)/runtime-build"
```
josephleekl: why is this necessary?

runtime/lib/transport/common/coproc/cpu/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+###############################################
+# CPU coprocessor function libraries          #
+###############################################
+# Pluggable `CoprocessorFn` implementations that any transport backend can dlopen at run
+# time. No RDMA/ibverbs/HIP dependencies — pure C++ decoder implementations against the
+# CoprocessorFn signature in Transport.hpp.
```
josephleekl:
```suggestion
# time. Pure C++ decoder implementations against the
# CoprocessorFn signature in Transport.hpp.
```

runtime/lib/transport/common/coproc/gpu/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
+###############################################
+# Shared GPU coprocessor plumbing             #
+###############################################
+# HIP-only, no RDMA. Used by both memcpy_gpu and gpu_verbs.
```
josephleekl:
```suggestion
# HIP-only. Used by both memcpy_gpu and gpu_verbs.
```

runtime/lib/transport/common/coproc/gpu/GpuLaunchers.hip
```diff
@@ -19,12 +19,12 @@
...
 #include "WireProtocol.hpp"
+#include "echo_decoder.hpp"
+#include "steane_decoder.hpp"
 
-#include "coprocessor_functions/echo_decoder.hpp"
-#include "coprocessor_functions/steane_decoder.hpp"
 #include <hip/hip_runtime.h>
 
-namespace catalyst::transport::gpu_verbs {
+namespace catalyst::transport::coproc {
```
josephleekl, line 27: this is for gpu, right?(it's a `hip` file), can we call this something gpu specific? gpu_coproc or something?

runtime/lib/transport/common/coproc/gpu/GpuLaunchers.hpp
```diff
@@ -15,7 +15,7 @@
 #pragma once
 #include "Transport.hpp" // CoprocLaunchDesc
 
-namespace catalyst::transport::gpu_verbs {
+namespace catalyst::transport::coproc {
```
josephleekl, line 18: same here

runtime/lib/transport/common/coproc/gpu/GpuLaunchers.hip
josephleekl: not important, skip if this is annoying

runtime/lib/transport/common/coproc/gpu/GpuRuntime.hip
```diff
@@ -13,22 +13,29 @@
...
-    RDMA_CHECK(device >= 0 && device < dc, "GpuRuntime: HIP device %d out of range (%d present)",
-               device, dc);
+    if (dc <= 0) {
+        throw std::runtime_error("GpuRuntime: no HIP device");
+    }
+    if (device < 0 || device >= dc) {
+        char buf[128];
+        std::snprintf(buf, sizeof(buf), "GpuRuntime: HIP device %d out of range (%d present)",
+                      device, dc);
+        throw std::runtime_error(buf);
```
josephleekl: why move away from the RDMA_CHECK? could we refactor the RDMA_CHECK to something more common so we could use that?

runtime/lib/transport/memcpy/CMakeLists.txt
```diff
@@ -0,0 +1,76 @@
+###############################################
+# Memcpy transport backend                    #
+###############################################
+
+# memcpy_cpu_impl is SHARED (not STATIC) so both memcpy plugins loaded into the same process
```
josephleekl:
```suggestion
# memcpy_cpu_impl is SHARED so both memcpy plugins loaded into the same process
```

runtime/lib/transport/memcpy/CMakeLists.txt
```diff
@@ -0,0 +1,76 @@
...
+# reference one copy of `acquire_memcpy_link` and its static registry map. Linking a STATIC
+# impl into two SHARED plugins would give each plugin its own private map, and a controller
+# and coprocessor dlopen'd from different .so files could not pair via the session pair key.
```
josephleekl:
```suggestion
# reference one copy of `acquire_memcpy_link` and its static registry map.
```

runtime/lib/transport/memcpy/CMakeLists.txt
```diff
@@ -0,0 +1,76 @@
...
+_memcpy_set_plugin_rpath(catalyst_transport_memcpy_controller)
+
+add_library(catalyst_transport_memcpy_coprocessor SHARED
+    cpu_coprocessor/CpuCoprocessorFactory.cpp
+)
+target_link_libraries(catalyst_transport_memcpy_coprocessor PRIVATE memcpy_cpu_impl)
+_memcpy_set_plugin_rpath(catalyst_transport_memcpy_coprocessor)
+
+# memcpy_gpu depends only on HIP (via the shared catalyst_transport_coproc_gpu). No libibverbs
+# needed — the RDMA-side gpu_verbs plugin is not required.
```
josephleekl:
```suggestion
# memcpy_gpu depends only on HIP (via the shared catalyst_transport_coproc_gpu).
```

runtime/lib/transport/memcpy/MemcpyLink.hpp
```diff
@@ -0,0 +1,71 @@
...
+//
+// The pair key is emitted by the compiler (inject-transport-session's `key` attribute on
+// transport.create) and folded into the backend config as `pair=<key>` by TransportCAPI, so
+// pairing is decided entirely at the MLIR level; memcpy's transport.connect intentionally
+// carries no peer or oob_port and ConnectInfo is ignored by memcpy sessions.
+//
+// The coprocessor binds `process_message` on connect(); the controller's kick() invokes it
+// inline. `mu` guards both fields across connect / kick / teardown.
+struct MemcpyLink {
+    // Invoked once per controller kick(). Reads `in_len` bytes from `in` (a wire-shaped
```
josephleekl: I thought we don't call it kick anymore?

runtime/lib/transport/memcpy/cpu_coprocessor/CpuCoprocessorSession.hpp
```diff
@@ -0,0 +1,89 @@
...
+#include <thread>
+#include <vector>
+
+#include "MemcpyLink.hpp"
+#include "Transport.hpp"
+#include "WireProtocol.hpp"
+
+namespace catalyst::transport::memcpy {
+
+// Structurally mirrors the RDMA cpu_verbs coprocessor: a persistent worker jthread polls a
```
josephleekl:
```suggestion
// A persistent worker jthread polls a
```

runtime/lib/transport/memcpy/gpu_coprocessor/GpuCoprocessorSession.cpp
```diff
@@ -0,0 +1,242 @@
...
+        } catch (...) {
+            error_ = std::current_exception();
+            failed_.store(true, std::memory_order_release);
+        }
+    });
+}
+
+int GpuCoprocessorSession::collect(void *const * /*replies*/,
+                                   const std::uint64_t * /*replies_bytes*/, std::size_t /*n*/) {
+    throw std::logic_error("memcpy: coprocessor collect is not used");
```
josephleekl: is logic_error used anywhere else?

josephleekl (approved):
Thanks @multiphaseCFD ! Got some comments, please go ahead and merge when you've addressed them 🚀

runtime/lib/transport/common/coproc/gpu/GpuLaunchers.hip
multiphaseCFD: agree, let's add a TODO and skip for now.

frontend/catalyst/backline.py
multiphaseCFD: Thanks @kszenes . Let's resolve it when the memcpy demo tests pass, let's wait green lights from @mehrdad2m

frontend/test/pytest/test_backline.py
multiphaseCFD: Nice catch!

runtime/lib/transport/memcpy/cpu_coprocessor/CpuCoprocessorSession.cpp
kszenes: @josephleekl @multiphaseCFD  I have this PR that I am working on which should unify the return type for both CPU and GPU coprocessor_fns: https://github.com/PennyLaneAI/catalyst/pull/3128

runtime/lib/transport/TransportCAPI.cpp
multiphaseCFD: Done!

runtime/lib/transport/memcpy/MemcpyLink.hpp
multiphaseCFD: yep, no `kick` anymore in the compilation pipeline and we use `post` for now. The `kick()` reference here is the C++ `ControllerSession::kick()` in `runtime/include/Transport.hpp:172`, which memcpy's controller still implements. `__catalyst__transport__post` is a thin CAPI wrapper that forwards to it. add ticket SC-128082 for it. We can come back to it later

runtime/lib/transport/memcpy/gpu_coprocessor/GpuCoprocessorSession.cpp
multiphaseCFD: yep, it's used in the Transport.hpp header, we need to comeback to sort out it.

runtime/lib/transport/memcpy/cpu_controller/CpuControllerSession.hpp
```diff
@@ -0,0 +1,83 @@
...
+    std::vector<std::unique_ptr<std::byte[]>> caller_memory_regions_;
+
+    std::vector<std::byte> request_staging_;
+
+    /// Set on the first commit_work_item(). Subsequent commits are rejected so any pointer a
+    /// prior data_slot() handed out cannot dangle behind a request_staging_ reallocation.
+    bool committed_ = false;
+
+    std::uint64_t in_bytes_ = 0;
+    std::uint64_t out_bytes_ = 0;
```
kszenes: @multiphaseCFD Could you follow the other [RDMA controller](https://github.com/PennyLaneAI/catalyst/blob/0f969048915d9e5e2e44146da7a3a608c2b6360e/runtime/lib/transport/rdma/cpu_verbs/controller/CpuControllerSession.hpp#L52) and set it to for consistency:
```
    std::uint64_t in_bytes_ = sizeof(common::Payload::value);
    std::uint64_t out_bytes_ = sizeof(common::Payload::value);
```

.github/workflows/check-transport-gpu.yaml
multiphaseCFD: Good call

multiphaseCFD:
> Thanks @multiphaseCFD ! Got some comments, please go ahead and merge when you've addressed them 🚀

@rniczh  mind if i merge the current branch into yours @rniczh  once CIs turn green CC: @kszenes @josephleekl

PR #3119, opened 2026-08-14
Fix SSH multiplexing silently disabled on macOS

mehrdad2m:
**Context:**
macOS puts `$TMPDIR` under a `/var/folders/...` path long enough to overrun the
104-byte `sun_path` limit, so no control socket was ever created and every
remote op re-authenticated. 

Sockets now go in` ~/.catalyst/cm`

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

Assisted by: Claude Code


multiphaseCFD (approved):
Thanks @mehrdad2m !

josephleekl (approved):
thanks @mehrdad2m

PR #3120, opened 2026-08-14
[Frontend] Support Adjoint ops and rules in the decomposition system

maliasadi:
**Context:**

This PR adds `Adjoint(Op)` E2E support to the graph-decomposition system while keeping the graph solver a plain (operator-id -> rules) engine with no modifier logic! The modifier lives entirely in the operator `graphOpId` and in the compiled rules.

I re-implemented my first work (in PR 3115) and removed the old code from that PR. #3115 focuses on the compiler + graph updates while this one focuses on frontend; extending Paul's work in PR #3061 for Adjoint rules and ops support. 

When lowering a gate Op, the reachable-rules pass now captures rules for three pathways (each best-effort):
1. the base op rules;
2. Adjoint(Op) registered rules (e.g. `add_decomps("Adjoint(RX)", …)`), self-adjoint, ...
3. Adjoint(Op) synthesized rules by distributing each base rule over adjoint (`decompose(Adjoint(Op)) = adjoint(decompose(Op))`) 

For (3), the body is traced through `qp.adjoint` into an adjoint region and the resources are the base resources with each produced op wrapped Adjoint(...). A rule whose body is non-invertible (`qp.measure`) will be skipped.

Added tests for:
- [X] The solver unit tests: `Test_DecompGraphSolver.cpp` 
- [X] MLIR lit (`AdjointDecompositionTest.mlir` and `GraphDecomposition/TestAdjoint*`
- [X] Frontend lit (`test_lower_time_rules_adjoint.py`, `test_adjoint_distribution_rule.py`),
- [X]  Frontend pytest (`TestTraceTime`).

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-125403]


frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -12,7 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""This module provides infrastructure for lowering decomposition rules via python."""
```
kipawaa, line 15: Seems like your comment formatter is set to 80-character line limits, this should fit on one line :)

frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -89,6 +104,9 @@ def get_rules_from_module(module: ir.Module) -> str:
 def inject_new_rules_into_module(module: ir.Module, decomp_rules: list[str]):
     with ir.InsertionPoint(module.body):
         for decomp_rule in decomp_rules:
+            if not decomp_rule:
+                continue
+
```
```suggestion
```
I don't think this is needed here

frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -160,47 +187,65 @@ def compile_decomposition_rules(
...
 
     The decomposition rules will be decorated with appropriate resource and target_gate attributes.
+
+    When ``wrap_adjoint`` is True, the rules registered on the base op ``op_name`` are instead
+    synthesized into rules for ``Adjoint(op_name)`` (aka the "distribution" pathway). Each base
+    rule body is wrapped in a ``qml.adjoint`` region (reduced to op-level modified gates by
+    ``adjoint-lowering`` within the decomposition pass), the ``target_gate`` updates to the adjoint
+    id, and each produced op in the resources is wrapped in ``Adjoint(...)``.
     """
-    rule_args, rule_kwargs = prepare_op_args(dynamic_shape, wire_lens, is_custom_op)
```
kipawaa, line 169: Why do you move the `arg`/`kwarg` splitting logic into each of the calls rather than leaving it here before passing them into each of the helpers?

frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -283,41 +383,52 @@ def fetch_all_reachable_decomposition_rules_from_op(
...
+                except Exception as e:
+                    warnings.warn(
+                        f"Failed to lower the {_rule_name} decomposition rule for {this_name}: {e}"
+                    )
                 continue
     return rules
+
+
+def _op_variants(op_name, op_id):
+    """Yield the (name, id) for an operator, unless it is already adjointed.
```
```suggestion
    """
    Yield the (name, id) for an operator, unless it is already adjointed.
```
Multi-line docstrings should be on their own lines :)

frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -32,6 +34,19 @@
 )
 from catalyst.jax_extras.lowering import get_mlir_attribute_from_pyval
 
+# Ops that make a decomposition body non-invertible:
+_NON_INVERTIBLE_MARKERS = ("qref.measure", "quantum.measure")
+
+
+def name_wrap_adjoint(op_id: str) -> str:
```
kipawaa, line 44: Rather than this type of GOID surgery, could we generate a new GOID via the GOID class with the same params and an updated name, `adjoint(op_name)`? I think this would require a flag (ex. `adjoint_resources`) to `collect_resource_for_op` but not much else. This would help preserve the opacity of GOIDs (I know I sound like a broken record about this, sorry 😅)

frontend/catalyst/decomposition/type_stringify_utils.py
kipawaa, line 1: I don't think these are used anywhere? We've got utilities that perform these functions in `decomposition/type_utils.py`

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -1815,20 +1815,26 @@ def circuit(x: float, y: float):
         "libpython_path": str(libpython_path if libpython_path else get_libpython_path()),
     }
 
+    def rule_ref_name(rule):
```
kipawaa, line 1818: Nice!

frontend/catalyst/passes/builtin_passes.py
```diff
@@ -1815,20 +1815,26 @@ def circuit(x: float, y: float):
...
+        # A rule reference may be a name string, a plain decomposition function, or a PennyLane
+        # ``DecompositionRule`` (returned by ``@register_resources``, which has no ``__name__``).
+        if isinstance(rule, str):
+            return rule
+        if hasattr(rule, "_impl"):
```
kipawaa:
```suggestion
        if isinstance(rule, qp.decomposition.DecompositionRule):
            return rule.name
```
I think pennylane decomposition rules require a `name` attribute ([src](https://github.com/PennyLaneAI/pennylane/blob/27838334cd3c01553088038949b34796d3f3ac9b/pennylane/decomposition/decomposition_rule.py#L374))

mlir/include/Driver/DefaultPipelines/DefaultPipelines.h
kipawaa:
```suggestion
      }},  // to remove user decomposition rules after all graph-decomposition passes
```
If we're removing decomp rules earlier in the pipeline then we don't need to do it here anymore :)

frontend/test/lit/GraphDecomposition/TestAdjointDistributeRegion.mlir
```diff
@@ -0,0 +1,47 @@
...
+func.func @distribute_region(%q: !quantum.bit) -> !quantum.bit {
+  // CHECK-NOT: quantum.adjoint
+  // CHECK: %[[A:.*]] = quantum.custom "Hadamard"() %[[Q]] : !quantum.bit
+  // CHECK: %[[B:.*]] = quantum.custom "Hadamard"() %[[A]] : !quantum.bit
+  // CHECK: return %[[B]]
+  %out = quantum.custom "U"() %q adj : !quantum.bit
+  return %out: !quantum.bit
+}
+
+// Adjoint(U) via distribution:
```
kipawaa, line 28: We've run into a few issues lately where `CHECK` directives in the main circuit accidentally match in the decomposition rules below, I recommend adding a `CHECK-LABEL` that matches the rule name to avoid that situation, since it's so hard to catch and debug!
```suggestion
// Adjoint(U) via distribution:
// CHECK-LABEL: func.func private @adj_u`
```

doc/releases/changelog-dev.md
```diff
@@ -17,6 +17,15 @@
   [(#3116)](https://github.com/PennyLaneAI/catalyst/pull/3116)
   [(#3131)](https://github.com/PennyLaneAI/catalyst/pull/3131)
 
+* The graph-based decomposition system now supports **adjoint operators** for `Operator2`.
```
paul0403, line 20: Should this be moved to be under the big decomposition changelog entry (in improvements section)?

frontend/test/lit/test_operator2/test_adjoint_distribution_rule.py
```diff
@@ -0,0 +1,58 @@
...
+
+        @qp.qjit(capture=True, target="mlir")
+        @qp.qnode(qp.device("null.qubit", wires=3))
+        def c():
+            NoParams(reg=[0, 1])
+            return qp.state()
+
+        print(c.mlir)
+    # CHECK-DAG: func.func private @"__builtin_base_rule_NoParams{}{reg:2}{}"
+    # CHECK-DAG:   target_gate = "NoParams{}{reg:2}{}"
```
kipawaa:
```suggestion
    # CHECK-SAME:   target_gate = "NoParams{}{reg:2}{}"
```

frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -32,6 +34,19 @@
...
+    """Name-wrap the adjoint modifier around a graphOpId."""
+    split = len(op_id)
+    for i, char in enumerate(op_id):
+        if char in "{[":
```
paul0403:
```suggestion
        if char == "{":
```

The first field (dynamic shape) is always curly brackets, and will always be present even if there are no params

frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -89,6 +104,9 @@ def get_rules_from_module(module: ir.Module) -> str:
 def inject_new_rules_into_module(module: ir.Module, decomp_rules: list[str]):
     with ir.InsertionPoint(module.body):
         for decomp_rule in decomp_rules:
+            if not decomp_rule:
+                continue
```
paul0403: I think River and I deleted this `continue` in the upstream lower time branch

frontend/test/lit/test_operator2/test_adjoint_distribution_rule.py
```diff
@@ -0,0 +1,58 @@
...
+    # CHECK-DAG: func.func private @"__builtin_base_rule_Adjoint(NoParams){}{reg:2}{}"
+    # CHECK-DAG:   resources = {operations = {"Adjoint(SingleParam){x:{{\[\[f64\]\]}}}{reg:2}{}" = 2 : i64}
+    # CHECK-DAG:   target_gate = "Adjoint(NoParams){}{reg:2}{}"
```
kipawaa:
```suggestion
    # CHECK-SAME:   resources = {operations = {"Adjoint(SingleParam){x:{{\[\[f64\]\]}}}{reg:2}{}" = 2 : i64}
    # CHECK-SAME:   target_gate = "Adjoint(NoParams){}{reg:2}{}"
```

kipawaa: I believe `CHECK-SAME` is order-independent, but we also know that MLIR dictionaryattrs are sorted by name, so we can be sure that `target_gate` comes after `resources` :)

frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -115,9 +133,24 @@ def find_condition(op):
                 decomp_rule_op.clone()
 
 
-def collect_resources_for_op(op_name, args, kwargs):
+def split_call_args(kwargs, is_custom_op):
```
paul0403, line 147: Any particular reason why you moved the custom op handling out into its own function? Just for cleanup?

No need to change back, I'm just wondering : )

frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -160,47 +187,65 @@ def compile_decomposition_rules(
...
+    wrap_adjoint=False,
 ) -> ModuleOp:
     """
     Return a ModuleOp containing the decomposition rules for an operator instance.
 
     The decomposition rules will be decorated with appropriate resource and target_gate attributes.
+
+    When ``wrap_adjoint`` is True, the rules registered on the base op ``op_name`` are instead
+    synthesized into rules for ``Adjoint(op_name)`` (aka the "distribution" pathway). Each base
+    rule body is wrapped in a ``qml.adjoint`` region (reduced to op-level modified gates by
```
paul0403:
```suggestion
    rule body is wrapped in a ``qp.adjoint`` region (reduced to op-level modified gates by
```

frontend/test/lit/test_operator2/test_lower_time_rules_adjoint.py
```diff
@@ -0,0 +1,95 @@
...
+            NoParams(reg=[0, 1])
+            return qp.state()
+
+        print(c.mlir)
+
+    # CHECK: qref.operator "NoParams"()
+    # CHECK-DAG: func.func private @"__builtin_base_rule_NoParams{}{reg:2}{}"
+    # CHECK-DAG:   target_gate = "NoParams{}{reg:2}{}"
+    # CHECK-DAG: func.func private @"__builtin_adj_rule_Adjoint(NoParams){}{reg:2}{}"
+    # CHECK-DAG:   target_gate = "Adjoint(NoParams){}{reg:2}{}"
```
kipawaa:
```suggestion
    # CHECK: qref.operator "NoParams"()
    # CHECK-DAG: func.func private @"__builtin_base_rule_NoParams{}{reg:2}{}"
    # CHECK-SAME:   target_gate = "NoParams{}{reg:2}{}"
    # CHECK-DAG: func.func private @"__builtin_adj_rule_Adjoint(NoParams){}{reg:2}{}"
    # CHECK-SAME:   target_gate = "Adjoint(NoParams){}{reg:2}{}"
```
This should work as desired 😄

frontend/test/lit/test_operator2/test_lower_time_rules_adjoint.py
```diff
@@ -0,0 +1,95 @@
...
+        @qp.qnode(qp.device("null.qubit", wires=3))
+        def c():
+            qp.adjoint(NoParams(reg=[0, 1]))
+            return qp.state()
+
+        print(c.mlir)
+
+    # CHECK: qref.operator "NoParams"() adj
+    # CHECK-DAG: target_gate = "Adjoint(NoParams){}{reg:2}{}"
+    # CHECK-DAG: target_gate = "NoParams{}{reg:2}{}"
```
kipawaa: We should probably check that these appear as targets on the correct rules as well

frontend/test/pytest/test_decomposition.py
```diff
@@ -153,9 +156,155 @@ def test_wrapper_operator(self, mocker):
...
+        # # NOTE: empty pass is needed to prevent running default pipeline
+        # rules = _quantum_opt("--empty", BYTECODE_FILE_PATH)
+        #
+        # assert "_isingxy_to_h_cy" in rules
+        # assert "_doublexcit" in rules
+        # assert "_pauliz_to_ps" in rules
+        # assert "_cphase_to_ppr" in rules
+        # assert "_crot" in rules
+        pass
+
```
kipawaa: This can be removed (matches main)

frontend/test/pytest/test_decomposition.py
kipawaa, line 1: I think we should leave out these changes - this file is intended for end-to-end integration tests, but when we check `"string" in mlir` we're just writing a lit test using pytest infrastructure that's not as well designed for it. If you want to add some end-to-end tests that verify the output of some measurement process or resulting state, that would be more appropriate for this file

kipawaa (commented):
Nice work @maliasadi! It's great to see this all coming together

frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -261,16 +306,71 @@ def fetch_all_reachable_decomposition_rules_from_op(
...
+                    get_rule_strings_from_module(
+                        compile_decomposition_rules(
+                            adj_name,
+                            name_wrap_adjoint(op_id),
+                            dynamic_shape,
+                            wire_lens,
+                            static_data,
+                            extra_data=extra_data,
+                            is_custom_op=is_custom_op,
+                        )
```
paul0403:
```suggestion
            # Rules registered directly against Adjoint(name):
            try:
                out.extend(
                    get_rule_strings_from_module(
                        compile_decomposition_rules(
                            adj_name,
                            name_wrap_adjoint(op_id),
                            dynamic_shape,
                            wire_lens,
                            static_data,
                            extra_data=extra_data,
                            is_custom_op=is_custom_op,
                            wrap_adjoint=False
                        )
```
I know we gave the entry point's `wrap_adjoint` a default of false, but I think it's better to pass in explicitly here to better communicate the purpose of the three pathways.

frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -32,6 +34,19 @@
 )
 from catalyst.jax_extras.lowering import get_mlir_attribute_from_pyval
 
+# Ops that make a decomposition body non-invertible:
+_NON_INVERTIBLE_MARKERS = ("qref.measure", "quantum.measure")
```
paul0403: This list needs mbqc measure in basis and ppm right?

frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -160,47 +187,65 @@ def compile_decomposition_rules(
...
+            for rule_name, ids in name_to_resource_ids.items()
+        }
+
     # The static_data was only needed to instantiate the correct decomp rule
     # Once we have the correct rules, don't send them into qjit
     def rule_to_subroutine(rule):
         def decomp_rule(*_args, **_kwargs):
-            rule._impl(*_args, **_kwargs)
+            if wrap_adjoint:
+                qp.adjoint(rule._impl)(*_args, **_kwargs)
```
paul0403, line 232: Nice!

frontend/catalyst/decomposition/type_stringify_utils.py
```diff
@@ -0,0 +1,44 @@
...
+def _stringify_shaped_type(shape: tuple, dim: int, element_type):
+    if dim + 1 == len(shape):
+        inner_content = _py_dtype_to_mlir_type_string(element_type)
+    else:
+        inner_content = _stringify_shaped_type(shape, dim + 1, element_type)
+    length = shape[dim]
+    return f"[{','.join([inner_content] * length)}]"
+
+
+def mlir_stringify_type(dtype: qp.typing.AbstractArray):
```
paul0403: Is this used anywhere?

frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -283,41 +383,52 @@ def fetch_all_reachable_decomposition_rules_from_op(
...
+
+
+def _op_variants(op_name, op_id):
+    """Yield the (name, id) for an operator, unless it is already adjointed.
+
+    For a gate `Op` we lower the rules registered against both `Op` and `Adjoint(Op)`.
+    """
+    yield op_name, op_id
+    if not op_name.startswith("Adjoint("):
+        yield f"Adjoint({op_name})", name_wrap_adjoint(op_id)
```
paul0403: The idea is to fetch all rules of all variants during lower time?

So base rules of `op`, base rules of `Adjoint(Op)`, and adjointed versions of the base rules of `op` will all be injected when lowering `op`?

frontend/test/lit/test_operator2/test_adjoint_distribution_rule.py
```diff
@@ -0,0 +1,58 @@
...
+        @qp.qnode(qp.device("null.qubit", wires=3))
+        def c():
+            NoParams(reg=[0, 1])
+            return qp.state()
+
+        print(c.mlir)
+    # CHECK-DAG: func.func private @"__builtin_base_rule_NoParams{}{reg:2}{}"
+    # CHECK-DAG:   target_gate = "NoParams{}{reg:2}{}"
+    # CHECK-DAG: func.func private @"__builtin_base_rule_Adjoint(NoParams){}{reg:2}{}"
+    # CHECK-DAG:   resources = {operations = {"Adjoint(SingleParam){x:{{\[\[f64\]\]}}}{reg:2}{}" = 2 : i64}
```
paul0403: The creation of the adjointed resource counts are handled by PL, as the symbolic rules for adjoint, right?

frontend/test/lit/test_operator2/test_lower_time_rules_adjoint.py
paul0403: Question: shouldn't there be 3 rules collected for this test?
1. The base rule for `NoParams{}{reg:2}{}`, containing one `SingleParam`
2. The base rule for `Adjoint(NoParams){}{reg:2}{}`, containing 2 `SingleParam`
3. The synthesized rule for `Adjoint(NoParams){}{reg:2}{}` (by calling `qp.adjoint` on the rules from point 1), containing one `Adjoint(SingleParam)`

frontend/test/lit/test_operator2/test_lower_time_rules_adjoint.py
paul0403: Same here.

Also a bit confused about the order in which the rules show up. Since the `fetch_all_reachable_decomposition_rules_from_op` function itself still just takes in the raw, non adjointed gate name when lowering a gate, shouldn't it only see `NoParams` here? And then it would go in the "NoParam base rule, Adjoint(NoParam) base rule, synthesized rule" order, according to the implementation of `compile_variants`?

mlir/include/Driver/DefaultPipelines/DefaultPipelines.h
```diff
@@ -53,6 +53,10 @@ const PipelineList pipelineList{
       // this into something else.
       "inline-nested-module",
       "lower-mitigation",
+      // Decomposition rules are only consumed by graph-decomposition (run inside
+      // apply-transform-sequence). Any that survive here are dead; drop them before
+      // adjoint-lowering so their `quantum.adjoint` regions are not lowered as user adjoints.
+      "symbol-dce",
```
paul0403, line 59: Doesn't this violate the requirement that decomp pass can be called from any stage in the pipeline?

We can do this for now, but at some point we should move decomp rules into their own module and stop (start?) scheduling transform sequences on them.

frontend/test/lit/test_operator2/test_adjoint_distribution_rule.py
paul0403: Oh never mind, we are wrapping the name in `name_to_resource_ids` ourselves after collecting the resources for the non-adjoint base rule 👍

frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -261,16 +306,71 @@ def fetch_all_reachable_decomposition_rules_from_op(
...
+                dynamic_shape,
+                wire_lens,
+                static_data,
+                extra_data=extra_data,
+                is_custom_op=is_custom_op,
+            )
         )
+        if not name.startswith("Adjoint("):
+            adj_name = f"Adjoint({name})"
+            # Rules registered directly against Adjoint(name):
```
paul0403: One overarching question I have is how would we link up the non-adjoint rules part of the graph to the adjoint rules part of the graph. Say we are lowering a hadamard. Currently we would lower the following rules:
- 1. Base rules of H
- 2. Base rules of adj(H)
- 3. adj-regioned rules of H, targetting adj(H)

This seems to create two unconnected parts, one for non-adjoint and one for adjoint.

Where do we generate the rule that links between H and adj(H)? Are they in the rules of adj(H), registered on PL directly? But even that would only have a rule from adj(H) to H. How about the reverse direction from H to adj(H)?

frontend/catalyst/decomposition/decomposition_rules.py
maliasadi: omg! this is the result of `make format`!! let me check my local version!! but why is it passing on CI?

frontend/catalyst/decomposition/decomposition_rules.py
maliasadi: but this is consistent with other docstrings :D is this also a problem with my py styler?! omg

frontend/catalyst/decomposition/type_stringify_utils.py
maliasadi: yes. you're right! the result of a partial rebase!

doc/releases/changelog-dev.md
maliasadi: This is not an improvement, but rather the introduction of a new feature (`adjoint`).

frontend/catalyst/decomposition/decomposition_rules.py
maliasadi: It's just cleaner :)

frontend/catalyst/decomposition/decomposition_rules.py
maliasadi: Oops! an old habit :smile:

frontend/catalyst/decomposition/decomposition_rules.py
maliasadi: I haven't run into those yet! but YES!!! 

FIXED!

mlir/include/Driver/DefaultPipelines/DefaultPipelines.h
maliasadi: Could you share the link to the SC ticket? I cannot find the ticket -- but I know we discussed/created one a while ago :grimacing:

frontend/catalyst/decomposition/decomposition_rules.py
maliasadi: `Op -> adj(Op)` needs to be defined and registered as additional rules for the base operator. The current assumption with the first version of adjoint/ctrl is that the compiler doesn't synthesize/create extra rules except the "distribution rules" and the graph sovler must not carry any logic for modifiers. These seem to be aligned with the EOY requirements, but we can definitely revisit and fix/extend as follow-ups. 

  As discussed offline -- we may need to work with core to upstream this logic to PL or implement the rule synthesization as an MLIR pass in the future.

frontend/test/lit/test_operator2/test_lower_time_rules_adjoint.py
maliasadi: Yes, there should be 3! I'll fix this :+1:

frontend/test/lit/test_operator2/test_lower_time_rules_adjoint.py
maliasadi: Good catch @paul0403! Yes, the collector keys off the base name (regardless of the op-level adjoint), so lowreing `adjoint)NoParams)` sees only `NoParams` and produces the same 3 rules.  Let me fix the check

mlir/include/Driver/DefaultPipelines/DefaultPipelines.h
paul0403: I'm notoriously bad with sc, I just remember this was a requirement haha

But I'm not super concerned since this is near the end of the quantum stage. Feel free to resolve.

frontend/catalyst/decomposition/decomposition_rules.py
maliasadi: I did it a while ago as both consumers split different dicts and hoisting the split forced every caller to duplicate it. now as the result each helper owns the split for its own calling convention. still the same functionality but I thought this is cleaner a bit

frontend/catalyst/decomposition/decomposition_rules.py
maliasadi: Agreed it's cleaner for the resource ids! I think the main concern is the target id and the whole on-demand path (which only ever have the GOID string). Sooo a string helper is unavoidable there! Happy to do a class refactor as a follow-up if you want the two halves consistent.

frontend/test/lit/test_operator2/test_lower_time_rules_adjoint.py
maliasadi: Yes, Fixed!

frontend/test/pytest/test_decomposition.py
maliasadi: done [c6faf47](https://github.com/PennyLaneAI/catalyst/pull/3120/commits/c6faf47c7c6ab18cbf9fec5f83d9985096c0a3b5)

frontend/catalyst/decomposition/decomposition_rules.py
maliasadi: Yup, lowering Op means adding 1) base rules of Op, 2) rules registered on `Adjoint(Op)`, and 3) `Adjoint(Op)` rules synthesized by `qp.adjoint`  :heavy_plus_sign:  all these 3 types of rules for every op reachable through their resources.

paul0403 (approved):
🥳

frontend/catalyst/decomposition/decomposition_rules.py
kipawaa: Ah I think it's me that was incorrect @maliasadi ! For some reason my formatter (set to google style, as suggested in PL style guide) complains when the docstring description is on the same line as the quotes, but google's docs say that how you wrote it is correct, and is mirrored in our docs. Sorry for the confusion, I'll fix my formatter!

frontend/catalyst/decomposition/decomposition_rules.py
kipawaa: on-demand rules still provide the operator name via the `DecomposableGate` interface's `getOperatorName`, so we should be able to parse around this? The name is one of the parameters provided to `compile_decomposition_rules`

frontend/catalyst/decomposition/decomposition_rules.py
kipawaa: Ah I see, one uses `kwargs | static_data | extra_data` and one simply `kwargs`. This makes sense, thanks for cleaning that up :)

frontend/catalyst/decomposition/decomposition_rules.py
maliasadi: Agreeeeed! Fixed [5cf5f01](https://github.com/PennyLaneAI/catalyst/pull/3120/commits/5cf5f01fdfc9a3dd2ad0e48bf9b92d5b210dbd77)

kipawaa: Nice! I think we could also replace all uses of `name_wrap_adjoint` with `GraphOpID(op_name, dynamic_data, ..., adjoint=True) in order to completely remove the GOID-surgery. We could put the functionality in a similar function maybe, to simplify it, but this would preserve opaqueness without sacrificing functionality!

maliasadi: Hmm we cannot replace all use cases :thinking: I'm currently using the name wrap id in three places:

1. Target id in `compile_decomposition_rules`: This is the op being lowered. At this point we only have `op_name` + the `op_id` string (+ dynamic/wire/static). Btw, I wouldn't call this GOID-surgery haha! It's a name-prepend that carries the whole `{…}…[uid]` suffix through untouched :) 

2. Direct Adjoint(name) rule id in `adjoint_variant_rule_strings`. same reason as (1) we cnnot really remove this as well, at this stage only the `op_id` string is available.

3. Resource ids in `collect_resources_for_op` in the frontend. this can be replaced. I've already updated the helper method but not fully removed the string id use. Fixed [30ebdda](https://github.com/PennyLaneAI/catalyst/pull/3120/commits/30ebddac3fb38fee5bac2997264b2aa05535b7ee) :broom:

kipawaa: Oh of course, I forgot that we require an `Operator2` instance for the `GraphOpID` class. I think this may have been a mistake, in theory we can construct an ID from just the parameters you stated (i.e. name, dynamic shape, wires, static data), but we can update this down the line if it feels necessary.

kipawaa (approved):
LGTM! 🥳

PR #3122, opened 2026-08-15
Sara/hybrid phase folding

sarababaeii:
**Context:**
the input program should be in Clifford+T

Implements [[sc-122685](https://app.shortcut.com/xanaduai/story/122685)], [[sc-122686](https://app.shortcut.com/xanaduai/story/122686)], and [[sc-122687](https://app.shortcut.com/xanaduai/story/122687)]

**Benefits:**

**Possible Drawbacks:**
Does not support registers as function argument or over control flow
Can't understand state preparation as they become dynamic values in mlir

**Related GitHub Issues:**


main.py
sengthai, line 1: I can see all of the files on the top directory such (`main.py, PhaseFoldingQRefTest.mlir, PhaseFoldingTest.mlir, test_decomp.py, and test_phase_folding.py`) are useful for quick testing and debugging in development mode. Since these will be removed before we merge this PR, I would suggest to convert them as tests in the Catalyst ecosystem instead (such in `frontend/test` and `mlir/test`. And of course, you can still keep these files in the locally for developing.

mlir/lib/QRef/Transforms/phase_folding.cpp
```diff
@@ -0,0 +1,840 @@
...
+        if (params.empty()) {
+            return 0.0;
+        }
+
+        mlir::FloatAttr floatAttr;
+        if (mlir::matchPattern(params.front(), mlir::m_Constant(&floatAttr))) {
+            return floatAttr.getValueAsDouble();
+        }
+        else { // dynamic param
+            return 0.0;
```
sengthai, line 656: When the we got the constant as SSA runtime value (`RZ(a) where a is unknown in compilation time`), it seems we return 0.0 here. If two phase gates example RZ(a) and RZ(b) can be merge, and those params are unknown then this could do Rz(a + b)->Rz(0) then it will be normalized and drop out. I think we should rather support this dynamic param or rise or completely ignore and rise error instead.

mlir/lib/QRef/Transforms/phase_folding.cpp
```diff
@@ -0,0 +1,840 @@
...
+                llvm::outs() << "IfOp:  " << "\n";
+                handleIfOp(ifOp, currentAbst, wireTable);
+            })
+            .Case<mlir::scf::ForOp>([&](mlir::scf::ForOp forOp) {
+                llvm::outs() << "ForOp: " << "\n";
+                handleForOp(forOp, currentAbst, wireTable);
+            })
+            .Case<mlir::func::CallOp>([&](mlir::func::CallOp callOp) {
+                llvm::outs() << "CallOp:    " << callOp << "\n";
+                // handleCallOp(callOp, currentAbst, wireTable);
```
sengthai, line 551: I'm not sure if we could support this right now. If we are not planing to support this, normally, I would rise error for unsupported instruction. This opt-in fashion only avoid the silent warning, but also could eliminate  the vague in the program as well.

sengthai (commented):
Thanks for the effort on this PR! @sarababaeii 
Here is my initialed comments.

mlir/lib/QRef/Transforms/phase_folding.cpp
```diff
@@ -0,0 +1,840 @@
...
+        
+        // RegionSummary summary(RegionType::Procedure, procAbst);
+        // parentAbst.applySummary(std::move(summary));
+
+        // if getting from the stored functions, should copy the summary, since applySummary consumes it!
+    }
+
+    // --- MLIR Analysis ---
+
+    void analyzeOperation(mlir::Operation *op, ProgramAbstraction &currentAbst, WireTable &wireTable)
```
sengthai, line 534: Some of unsupported QRef operations are sliently ignored here. I would recommment to conservatively apply Gate::U to their qubits or fail with a diagnostic.

mlir/lib/QRef/Transforms/phase_folding.cpp
```diff
@@ -0,0 +1,840 @@
...
+
+    Gate extractCliffTGate(qref::CustomOp &op)
+    {
+        Gate gate = gateWithName(op.getGateName());
+        if (!op.getCtrlQubits().empty() || !op.getCtrlValues().empty()) {
+            if (isPhaseGate(gate)) {
+                return Gate::I; // C-Rz gates don't alter state space, but alter phase space
+                                // non-linearly, which I'm not going to track for now, but should be
+                                // trackable using xy = x + y - (x \oplus y).
+            }
```
sengthai, line 393: Here Y also is classify as a phase gate, so any controlled Y return `Gate::I`. However, unlike controlled Z/S/T, controlled Y changes computational state. For example: T(q); CY(c,q); T_ dagger(q).

PR #3125, opened 2026-08-17
Run DCE before to_ppr pass

paul0403:
**Context:**
Run DCE before `to_ppr` pass.

Now that the graph decomp infra is (about to be) ready, IRs will carry the decomposition rule functions. 
The to_ppr pass has its own gateset, and rule functions might contain gates that are outside that set.
We just run DCE at the beginning of the `to_ppr` pass to remove these rules.

**Benefits:**
Unblock PBC work.


sengthai (approved):
Thanks @paul0403 !

mlir/lib/PBC/Transforms/to_ppr.cpp
```diff
@@ -41,6 +42,13 @@ struct ToPPRPass : impl::ToPPRPassBase<ToPPRPass> {
         auto ctx = &getContext();
         ConversionTarget target(*ctx);
 
+        // Remove all decomposition rules, they might involve gates we don't care about
+        PassManager pm(ctx);
+        pm.addPass(mlir::createSymbolDCEPass());
+        if (failed(pm.run(getOperation()))) {
+            return signalPassFailure();
+        }
```
kipawaa: Could you change this to use `OpPassManager`? This is [recommended by mlir](https://mlir.llvm.org/docs/PassManagement/) for scheduling sub-passes on particular operation types (i.e. the module) (ex. in `graph-decomposition` [here](https://github.com/PennyLaneAI/catalyst/blob/ad72db959560384ab52afcc7cf4fdf83582e899c/mlir/lib/Quantum/Transforms/GraphDecomposition/graph_decomposition.cpp#L146)).

kipawaa (commented):
Thanks @paul0403 🚀 just a small request for better pass management

mlir/lib/PBC/Transforms/to_ppr.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/3125/commits/2e3f681dd3f253ed06e6c048204d6b37ef4cded2

kipawaa (approved):
Thanks @paul0403 !

PR #3128, opened 2026-08-18
fix: unify cpu/gpu ABI, both return status

kszenes:
Previously GPU returned status code, CPU number of bytes written. This means that if by accident, we assign a GPU coproc_fn to a CPU coproc, it will fail to launch due to ABI incompatibility between the due and return status code 1. The CPU on the other hand used to return the number of bytes written and, as a result, the CPU coproc interprets this status code as 1 byte was written and tries to read from the existing memory location, causing a silent error.

This PR fixes it by returning also a status code on the CPU side, effectively fixing the in/out_bytes, which is what we wanted to do anyway. The circuit still executes but at least the error is not silent:

```
[transport] memcpy: coprocessor function failed with status 1
...
```

NOTE: It seems like everyone assumes in/out_bytes of 8 while the Memcpy controller: defaults both to 0 until commit_work_item(). This should be fixed

### Before submitting

**Context:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/runtime/include/Transport.hpp
+++ b/runtime/include/Transport.hpp
@@ -188,10 +188,11 @@ class ControllerSession : public TransportSession {
 /**
  * @brief Per-message coprocessor function (CPU-style). Invoked once per received
  * message: decode `in` (`in_len` bytes) into `out` (capacity `out_cap`) and
- * return the number of bytes written.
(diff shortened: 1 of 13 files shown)


kszenes:
@multiphaseCFD

PR #3131, opened 2026-08-18
Add reference semantics ctrl region op to from_plxpr and the python interface

maliasadi:
**Context:**
Add reference semantics ctrl region op to `from_plxpr` and the new python interface.

In this PR, `qp.ctrl(qfunc, control=...)` is now captured as a single `qref.ctrl` region wrapping the whole subcircuit. mirroring how `qp.adjoint(qfunc)` lowers to a `quantum.adjoint` region. In the new pipeline,  `convert-to-value-semantics` turns it into a `quantum.ctrl` region and `ctrl-lowering` later distributes into op-level. 

This makes control of subcircuits containing classical control flow (`for`/`while`/`cond`) work end-to-end, and supports multiple controls, dynamic wires, control values, and composition with `qp.adjoint` (nested `ctrl(adjoint(...))` / `adjoint(ctrl(...))`). 

**Related GitHub Issues:**
[sc-128178]


frontend/catalyst/from_plxpr/qfunc_interpreter.py
```diff
@@ -716,18 +717,39 @@ def handle_measure_in_basis(self, angle, wire, plane, reset, postselect):
     return result
 
 
+def _ctrl_region_kernel(interpreter, jaxpr, n_consts, global_qreg, *consts_and_args):
```
paul0403: I'll take a deeper look tomorrow, but I think you don't even need your own primitive, and you don't need to do any custom handling. The only thing needed should just a be lowering rule on the _plxpr_ control primitive.

maliasadi: Good point! I just copied it from other impls. I'll check and fix

maliasadi: I removed it and simplified the impl using JB [c364bc1](https://github.com/PennyLaneAI/catalyst/pull/3131/commits/c364bc12279802a44a90afa217abfb984a9be2b5)

frontend/catalyst/from_plxpr/qref_jax_primitives.py
```diff
@@ -655,6 +658,48 @@ def _pl_adjoint_lowering(
...
+    n_consts,
+):
+    """Lower a `qp.ctrl` transform to a `qref.ctrl` region op."""
+    body_invals = plxpr_invals[: len(plxpr_invals) - n_control]
+    control_qubits = list(plxpr_invals[len(plxpr_invals) - n_control :])
+
+    i1 = ir.IntegerType.get_signless(1)
+    control_values_i1 = [
+        ConstantOp(i1, ir.IntegerAttr.get(i1, 1 if v else 0)).result for v in control_values
+    ]
```
mudit2812, line 681: I know that CSE would get rid of the excess constants, but would it be possible to only create at most 2 constants (true and false) and just reuse them?

mudit2812: I think that's what we do for "regular" controls as well, unless I've been looking at post-CSE IR 😅

frontend/test/pytest/python_interface/inspection/test_xdsl_conversion.py
mudit2812, line 1: Nice

mudit2812 (approved):
Looks great to me!

frontend/catalyst/from_plxpr/qfunc_interpreter.py
```diff
@@ -719,16 +720,26 @@ def handle_measure_in_basis(self, angle, wire, plane, reset, postselect):
...
 
-    unroller = copy(self)
-    unroller.control_wires += tuple(control_wires)
-    unroller.control_values += tuple(control_values)
-    unroller.eval(jaxpr, consts, *args)
-    return []
+    control_qubits = [
+        w if is_abstract_qubit(w) else qref_get_p.bind(self.init_qreg, w) for w in control_wires
+    ]
+    body = jaxpr_to_jaxpr(copy(self), jaxpr, consts, *args)
```
paul0403, line 727: I kinda don't like the fact that there's a `handle_ctrl_transform` but not a `handle_adjoint_transform`. The fact that we are still processing the PL ctrl primitive, despite directly having a lowering rule for it, feels a bit weird.

But I guess the difference is adjoint region in MLIR takes in no operands but ctrl region takes in ctrl qubits, which means it has qubit operands. And as soon as it has qubit operands, it hits the "plxpr represents qubits by integers" problem, and we need to do this `w if is_abstract_qubit(w) else qref_get_p.bind(self.init_qreg, w) for w in control_wires` mini-processing 🫠

paul0403: Nothing to do for now, since this is the same problem as the general "plxpr represents qubits by integers" described [here](https://docs.google.com/document/d/1s37oRBZyeAcCPoiisocyR2jT1hiRI1k_UrdX-es2kBQ/edit?tab=t.0#heading=h.oqht6bjwhm3)

frontend/test/lit/test_qref/test_flat_circuits.py
```diff
@@ -524,19 +535,25 @@ def test_adjoint_with_ctrl():
...
     # CHECK: qref.adjoint {
-    # CHECK:   [[q1:%.+]] = qref.get [[reg]][ 1] : !qref.reg<4> -> !qref.bit
-    # CHECK:   [[q2:%.+]] = qref.get [[reg]][ 2] : !qref.reg<4> -> !qref.bit
     # CHECK:   [[q0:%.+]] = qref.get [[reg]][ 0] : !qref.reg<4> -> !qref.bit
-    # CHECK:   qref.custom "SWAP"() [[q1]], [[q2]] ctrls([[q0]]) ctrlvals([[true]]) : !qref.bit, !qref.bit ctrls !qref.bit
+    # CHECK:   qref.ctrl([[q0]]) ctrlvals([[true]]) {
+    # CHECK:     [[q1:%.+]] = qref.get [[reg]][ 1] : !qref.reg<4> -> !qref.bit
+    # CHECK:     [[q2:%.+]] = qref.get [[reg]][ 2] : !qref.reg<4> -> !qref.bit
+    # CHECK:     qref.custom "SWAP"() [[q1]], [[q2]] : !qref.bit, !qref.bit
+    # CHECK:   }
```
paul0403, line 554: Nice!

mlir/include/Driver/DefaultPipelines/DefaultPipelines.h
```diff
@@ -53,6 +53,13 @@ const PipelineList pipelineList{
       // this into something else.
       "inline-nested-module",
       "lower-mitigation",
+      // Reduce `quantum.ctrl`/`quantum.adjoint` regions to op-level modifiers.
+      // Nested regions (e.g. `ctrl(adjoint(...))`) require alternating the two
+      // passes: `ctrl-lowering` defers on a nested adjoint region,
+      // `adjoint-lowering` reduces it, then `ctrl-lowering` runs again.
+      "ctrl-lowering",
+      "adjoint-lowering",
+      "ctrl-lowering",
```
paul0403, line 62: This hardcoded pipeline feels weird to me

mlir/lib/QRef/Transforms/value_semantics_conversion.cpp
```diff
@@ -1356,13 +1356,18 @@ void handleCtrl(IRRewriter &builder, qref::CtrlOp rCtrlOp, QubitValueTracker &tr
...
+    // then append the region targets.
     SetVector<Value> rValuesUsedByRegion;
-    // Special: ctrl qubits are passed in as operands explicitly instead of closure
-    size_t numCtrlQubits = rCtrlOp.getCtrlQubits().size();
     for (Value rCtrlQubit : rCtrlOp.getCtrlQubits()) {
         rValuesUsedByRegion.insert(rCtrlQubit);
     }
-    collectNecessaryRegionRValues(rCtrlOp.getRegion(), rValuesUsedByRegion);
+    size_t numCtrlQubits = rValuesUsedByRegion.size();
+    rValuesUsedByRegion.insert(rTargetValues.begin(), rTargetValues.end());
```
paul0403, line 1377: What was this change about? 😅

frontend/catalyst/from_plxpr/qfunc_interpreter.py
```diff
@@ -719,16 +720,26 @@ def handle_measure_in_basis(self, angle, wire, plane, reset, postselect):
...
     args = invals[n_consts:-n_control]
     control_wires = invals[-n_control:]
 
-    unroller = copy(self)
-    unroller.control_wires += tuple(control_wires)
-    unroller.control_values += tuple(control_values)
-    unroller.eval(jaxpr, consts, *args)
-    return []
+    control_qubits = [
+        w if is_abstract_qubit(w) else qref_get_p.bind(self.init_qreg, w) for w in control_wires
```
paul0403, line 725: Although, I feel like this wouldn't work with dynamically allocated qubits.

paul0403: Wait, it works. I'll add tests.

paul0403: I forgot the logic here! The `is_abstract_qubit` will be hit when the control qubits are not from the global register

mlir/include/Driver/DefaultPipelines/DefaultPipelines.h
maliasadi: hm.. this is the result of another decision; two reductions are mutually recursive on nesting right now. `ctrl-lowering` cant reduce a `quantum.ctrl` whose body still contains a `quantum.adjoint` region and vice-versa. So we must reduce inside-out with alternating passes.

mlir/lib/QRef/Transforms/value_semantics_conversion.cpp
maliasadi: haha.. `qref.ctrl` region has two kinds of qubit inputs: 1) control qubits and 2) target qubits (which are captured by the closure and collected via `collectNecessaryRegionRValues` into `rTargetValues`) <- your code in the other pr. Here in value-semantics, we need all of them considered as SSA region operands, but `quantum.ctrl` takes controls and targets as separate operand lists. This code applies the partitions basically.

mlir/include/Driver/DefaultPipelines/DefaultPipelines.h
paul0403: I guess these passes right now don't know how to thread the value semantics qubits across regions, and hence why they can't handle nesting. I guess this would be solved when we migrate these passes to reference semantics!

mlir/lib/QRef/Transforms/value_semantics_conversion.cpp
paul0403: https://github.com/PennyLaneAI/catalyst/pull/3131/commits/d83ad15ac0b83330fa73a1092c3e3acc70c7163f discussed offline, reverted

paul0403 (approved):

PR #3133, opened 2026-08-19
Remove Precompiled Rule Copy

kipawaa:
**Context:**
Precompiled rules were temporarily removed by #3053 to be reintroduced later, but the makefile still performs an unconditional copy of the generated file. This crashes when building wheels since the file does not exist.

Removes the `cp` command from the `make wheel` directive, to be reintroduced once precompiled rules are migrated for PL2.

**Benefits:**
Wheels!

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -268,6 +268,7 @@ wheel:
 	$(PYTHON) -m catalyst.decomposition.precompile_decomposition_rules
 
 	mkdir -p $(MK_DIR)/frontend/catalyst/resources
+	# TODO: re-enable this once pre-compiled rules are migrated for Operator2 compatibility
 	# cp $$($(PYTHON) -c 'from catalyst.utils.runtime_environment import BYTECODE_FILE_PATH; print(BYTECODE_FILE_PATH)') $(MK_DIR)/frontend/catalyst/resources/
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -268,7 +268,7 @@ wheel:
 	$(PYTHON) -m catalyst.decomposition.precompile_decomposition_rules
 
 	mkdir -p $(MK_DIR)/frontend/catalyst/resources
-	cp $$($(PYTHON) -c 'from catalyst.utils.runtime_environment import BYTECODE_FILE_PATH; print(BYTECODE_FILE_PATH)') $(MK_DIR)/frontend/catalyst/resources/
```
mehrdad2m, line 271: Could we add a TODO here as well?

mehrdad2m (approved):

Makefile
kipawaa: [d96817c](https://github.com/PennyLaneAI/catalyst/pull/3133/commits/d96817c88234357e4a08d966ea6f285ce0f04423)

paul0403 (approved):

PR #3135, opened 2026-08-19
Run DCE before --ppm-compilation

paul0403:
**Context:**
Run DCE before `--ppm-compilation`. 

#3125 , but I didn't know that `--ppm-compilation` invokes the `to_ppr` pattern directly and doesn't call the `--to-ppr` pass.


kipawaa (approved):
Thanks @paul0403 !

sengthai (approved):

lazypanda10117 (approved):

PR #3136, opened 2026-08-19
Adding PBC reference semantics

lazypanda10117:
**Context:** Adding reference semantics support for the remaining PBC ops.

**Benefits:** Enable the use of reference semantics in later passes that depend on the PBC dialect.


mlir/lib/PBC/IR/PBCOps.cpp
```diff
@@ -90,6 +90,45 @@ LogicalResult RefPPMeasurementOp::verify() {
...
+    size_t numPauliProduct = getPauliProduct().size();
+
+    if (numPauliProduct == 0) {
+        return emitOpError("Pauli string must be non-empty");
+    }
+
+    if (numPauliProduct != getQubits().size()) {
+        return emitOpError("Number of qubits must match number of pauli operators");
+    }
+    return success();
```
mehrdad2m: Can we use verifyPPROp helper instead?

mlir/lib/PBC/IR/PBCOps.cpp
```diff
@@ -344,6 +403,10 @@ uint64_t PPRotationArbitraryOp::getResourceNumQubits() { return getInQubits().si
 uint64_t PPMeasurementOp::getResourceNumQubits() { return getInQubits().size(); }
 uint64_t RefPPMeasurementOp::getResourceNumQubits() { return getQubits().size(); }
 uint64_t SelectPPMeasurementOp::getResourceNumQubits() { return getInQubits().size(); }
+uint64_t RefPPRotationOp::getResourceNumQubits() { return 0; }
```
mehrdad2m: Why is this zero? Maybe I am not too familiar with the definition but my assumption was that `getResourceNumQubits` is counting the number of qubits the operation act on.

mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp
```diff
@@ -374,6 +375,45 @@ void handlePPM(IRRewriter &builder, pbc::PPMeasurementOp vPPMOp, QubitValueTrack
...
+template <typename ROpTy, typename VOpTy>
+void handleQubitProducer(IRRewriter &builder, VOpTy vProducerOp, QubitValueTracker &tracker,
+                         SmallVector<Operation *> &erasureWorklist) {
+    OpBuilder::InsertionGuard guard(builder);
+    builder.setInsertionPoint(vProducerOp);
+    Location loc = vProducerOp.getLoc();
+    MLIRContext *ctx = vProducerOp.getContext();
+
+    SmallVector<Type> rQubitTypes(vProducerOp.getOutQubits().size(), qref::QubitType::get(ctx));
+    auto rProducerOp = ROpTy::create(builder, loc, rQubitTypes, vProducerOp.getInitState());
```
mehrdad2m: Should we preserve discardable attributes similar to `migrateOpToReferenceSemantics`? https://github.com/PennyLaneAI/catalyst/blob/2a7c8d1a32bc51ca9862d9efb59c68f6ae2fc8cc/mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp#L197-L198

mlir/include/PBC/IR/PBCRefOps.td
```diff
@@ -38,3 +38,99 @@ def RefPPMeasurementOp : PBC_Op<"ref.ppm",
...
+
+    let assemblyFormat = [{
+        ` ` `(` $select_switch `?` $pauli_product_0 (`(``-``)` $negated_0^)? `:` $pauli_product_1 (`(``-``)` $negated_1^)? `)` $qubits attr-dict `:` type(results)
+    }];
+
+    let hasVerifier = 1;
+}
+
+def RefFabricateOp : PBC_Op<"ref.fabricate",
+    [DeclareOpInterfaceMethods<ResourceQuantumOpInterface>, DeclareOpInterfaceMethods<ResourceAllocQubitOpInterface>]> {
```
mehrdad2m: I noticed that `FabricateOp` is declaring `getResourceDetailedName` but not this op.

mehrdad2m: same for `RefSelectPPMeasurementOp` and `RefPrepareStateOp`

mlir/include/PBC/IR/PBCRefOps.td
```diff
@@ -38,3 +38,99 @@ def RefPPMeasurementOp : PBC_Op<"ref.ppm",
...
+        ConfinedAttr<SI8Attr, [RotationKindConstraint]>:$rotation_kind,  // angle in fractions of π
+        Variadic<QubitRefType>:$qubits,  // The qubits to apply the rotation to
+        Optional<I1>:$condition
+    );
+
+    let assemblyFormat = [{
+        $pauli_product `(` $rotation_kind `)` $qubits attr-dict (`cond` `(` $condition^ `)`)?
+    }];
+
+    let hasVerifier = 1;
```
mehrdad2m: What about canonicalize? I noticed the value semantic version has one.

mlir/include/PBC/IR/PBCRefOps.td
```diff
@@ -38,3 +38,99 @@ def RefPPMeasurementOp : PBC_Op<"ref.ppm",
 
     let hasVerifier = 1;
 }
+
+def RefPPRotationOp : PBC_Op<"ref.ppr", [AttrSizedOperandSegments,
```
mehrdad2m: I think it would nice to still be able to use `PBCOpInterface` on these ops, but I guess that is assuming `$in_qubits` and `$out_qubits` where as ref ops are only using `$qubits`

mlir/lib/PBC/IR/PBCOps.cpp
lazypanda10117: I just updated this before your review, this should now be fixed.

mlir/lib/PBC/IR/PBCOps.cpp
```diff
@@ -90,6 +90,45 @@ LogicalResult RefPPMeasurementOp::verify() {
...
+        return emitOpError("Number of qubits must match number of pauli operators");
+    }
+    return success();
+}
+
+LogicalResult RefFabricateOp::verify() {
+    auto initState = getInitState();
+    if (initState == LogicalInitKind::zero || initState == LogicalInitKind::one ||
+        initState == LogicalInitKind::plus || initState == LogicalInitKind::minus) {
+        return emitOpError("Logical state should not be fabricated, use `PrepareStateOp` instead.");
```
mehrdad2m:
```suggestion
        return emitOpError("Logical state should not be fabricated, use `RefPrepareStateOp` instead.");
```

mlir/include/PBC/IR/PBCRefOps.td
lazypanda10117: Yea, we don't really have a proper interface for this right now.

mlir/lib/PBC/IR/PBCOps.cpp
lazypanda10117: Not really, the main issue is again the differing interface. One uses `getInQubits` and the other uses `getQubits`.

mlir/include/PBC/IR/PBCRefOps.td
lazypanda10117: That's added now.

mlir/lib/PBC/IR/PBCOps.cpp
lazypanda10117: I created a new function called `getQubitOperands()` to unify the way we retrieve the qubits, so we dont need to duplicate the verification logic across value and ref PBC ops.

mlir/include/PBC/IR/PBCRefOps.td
lazypanda10117: Added a canonicalizer for the ref PBC ops as well.

mlir/lib/PBC/IR/PBCOps.cpp
lazypanda10117: Done.

mlir/lib/Quantum/Transforms/reference_semantics_conversion.cpp
lazypanda10117: I dont think we are preserving discardable attributes for alloc-type ops, so I am just following the precedents.

PR #3137, opened 2026-08-19
Revert and xfail `test_ctrl_before_custom_op`

JerryChen97:
**Context:**
special lowering for MCX as a solution solved the issue with drawer, however introduced much more potential issues.
Given the fact that drawer module is not prioritized for now, we should instead xfail the corresponding test and explore other solutions

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**


paul0403 (approved):

kipawaa (approved):
Thanks! We'll look into a better fix for the drawer and graph

paul0403:
Logged the issue here https://github.com/PennyLaneAI/catalyst/issues/3139

PR #3138, opened 2026-08-20
Fix dialects build when using libstdc++ 14 from gcc

mehrdad2m:
**Context:**
Fix dialects build with libstdc++ 14 headers

The follwing pattern fails when using clang with the libstdc++ 14 headers:
```
llvm::enumerate(llvm::reverse(mcmobs.getMcms()))
```
with this error:

```
    error: no matching constructor for initialization of
      'indexed_accessor_range_base<mlir::OperandRange, ...>::iterator'
```

This is a know issue of std::reverse_iterator` over libstdc++ 14 . The error shows up on Ubuntu 24.04 whenever the gcc-14 headers are installed.

Copies the operands into a `SmallVector<Value>` before reversing, so the reversed iterator is a plain `Value *`

**Benefits:**
`make dialects` builds on Ubuntu 24.04 with the gcc-14 headers present.

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
+++ b/mlir/lib/Quantum/Transforms/dynamic_one_shot.cpp
@@ -415,7 +415,8 @@ void editKernelMCMProbs(IRRewriter &builder, func::FuncOp oneShotKernel, quantum
     auto totalIndex =
         arith::ConstantOp::create(builder, loc, i64Type, builder.getIntegerAttr(i64Type, 0));
     Operation *loopUpdater = totalIndex;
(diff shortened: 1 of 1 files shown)


JerryChen97:
But does our CI use clang + gcc14 headers

mehrdad2m:
> But does our CI use clang + gcc14 headers

Nope, that's why we don't see this failure in CI, but this is a common enough setup on different machines we use that the same problem has come up many times triggered by a different code pattern each time.

kipawaa (approved):

JerryChen97 (approved):

paul0403 (approved):

PR #3140, opened 2026-08-20
Concretize resource counts for resolvable nested loops

sengthai:
**Context:**
qp.specs returns symbolic resource counts for nested loops even when their total iteration count can be determined statically.

This PR updates ResourceAnalysis to resolve directly nested loop counts and report concrete resources. Here we only resolve for nested loops whose upper bounds are the induction variables of their immediately enclosing loops. Unresolvable dynamic loops remain symbolic. 

Note:
The `call_count` in `pennylane/resource/parsing.py` should support `float`, this is updated in https://github.com/PennyLaneAI/pennylane/pull/10044

[sc-123837]


doc/releases/changelog-dev.md
```diff
@@ -118,6 +118,30 @@
   AOT compilation and as arguments to `pennylane.specs` calculations.
   [(#2953)](https://github.com/PennyLaneAI/catalyst/pull/2953)
 
+* The `ResourceAnalysis` pass can now report concrete resource counts for nested loops whose upper
+  bounds are the induction variables of their immediately enclosing loops.
+  [(#3140)](https://github.com/PennyLaneAI/catalyst/pull/3140)
```
```suggestion
* The `ResourceAnalysis` pass can now report concrete resource counts for nested loops in cases
  where the bounds of an inner loop are directly dependent on the loop variable of a static outer loop.
  [(#3140)](https://github.com/PennyLaneAI/catalyst/pull/3140)
```

doc/releases/changelog-dev.md
```diff
@@ -118,6 +118,30 @@
...
+  def circuit():
+      for i in range(8):
+          for j in range(i):
+              for _ in range(j):
+                  qp.PauliX(0)
+
+      return qp.expval(qp.X(0))
+
+  resources = qp.specs(circuit, level="all-mlir")()
+  print(resources) # 56
```
```suggestion
  For example, this program reports a total of `56` `PauliX` operations, since the number of iterations of the inner loops can be statically determined from the outer loop:

  ```python
  import pennylane as qp

  @qp.qjit(autograph=True)
  def circuit():
      for i in range(8):
          for j in range(i):
              for _ in range(j):
                  qp.PauliX(0)

      return qp.expval(qp.X(0))

  resources = qp.specs(circuit, level="all-mlir")()
  print(resources) # 56
```

jzaia18: The multiplicity isn't something directly user-facing, I think we should avoid mentioning it

mlir/include/Catalyst/Utils/SCFUtils.h
```diff
@@ -50,4 +50,8 @@ std::optional<double> getEstimatedIterationsHint(Operation *op);
 // Returns std::nullopt when the trip count cannot be determined statically.
 std::optional<double> resolveForLoopTripCount(scf::ForOp forOp);
 
+// Resolve the average trip count of a loop whose upper bound is the immediately
+// enclosing loop's induction variable, including direct chains of such loops.
+std::optional<double> resolveDirectNestedForLoopAverageTripCount(scf::ForOp forOp);
```
jzaia18: Can we make this support loops where there isn't an immediate chain as well?

e.g.
```
for i in range(a):
    for j in range(b):  # Unrelated to first loop
        for j in range(i):  # Depends on first loop
```

I added an integration test for this as well

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
```diff
@@ -184,16 +184,18 @@ void ResourceAnalysis::analyzeForLoop(scf::ForOp forOp, ResourceResult &result,
     analyzeRegion(forOp.getBodyRegion(), bodyResult, isAdjoint);
 
     // Try to resolve a static trip count.
-    std::optional<double> tripCount = resolveForLoopTripCount(forOp);
-
-    if (tripCount.has_value()) {
+    auto tripCount = resolveForLoopTripCount(forOp);
```
jzaia18, line 192: Why change this to `auto`?

mlir/lib/Catalyst/Utils/SCFUtils.cpp
```diff
@@ -28,9 +34,12 @@ using namespace mlir;
...
 namespace catalyst {
 
-static int64_t getNumIterations(double lowerBound, double upperBound, double step) {
-    assert(upperBound >= lowerBound && step > 0);
-    return std::ceil((upperBound - lowerBound) / step);
+static int64_t computeTripCount(int64_t lowerBound, int64_t upperBound, int64_t step) {
+    assert(step > 0);
+    if (upperBound <= lowerBound) {
+        return 0;
+    }
```
jzaia18, line 48: What's the purpose of this change? Won't this silently allow loops that iterate in reverse to function but always report them as a no-op?

mlir/lib/Catalyst/Utils/SCFUtils.cpp
```diff
@@ -28,9 +34,12 @@ using namespace mlir;
...
+    return (upperBound - lowerBound + step - 1) / step;
```
jzaia18: Why is this preferred over `ceil`?

mlir/lib/Catalyst/Utils/SCFUtils.cpp
```diff
@@ -97,12 +106,117 @@ std::optional<double> resolveForLoopTripCount(scf::ForOp forOp) {
...
+            break;
+        }
+    }
+    if (isCanonical) {
+        int64_t depth = static_cast<int64_t>(ranges.size());
+        double average = static_cast<double>(outerUpper - depth + 1) / static_cast<double>(depth);
+        return std::max(0.0, average);
+    }
+
+    // Each loop is resolved separately, so nested non-canonical loops may revisit outer loops.
```
jzaia18: What is a "canonical" loop?

mlir/lib/Catalyst/Utils/SCFUtils.cpp
```diff
@@ -97,12 +106,117 @@ std::optional<double> resolveForLoopTripCount(scf::ForOp forOp) {
...
+    int64_t lower;
+    int64_t step;
+};
+
+struct TripCountSummary {
+    int64_t total = 0;
+    uint64_t invocations = 0; // how many times the innermost loop is reached
+};
+
+static TripCountSummary accumulateTripCounts(llvm::ArrayRef<LoopRange> ranges, size_t loopIndex,
```
jzaia18: I think some extra comments explaining what these are for would be helpful, it's a bit hard to decipher atm

mlir/lib/Catalyst/Utils/SCFUtils.cpp
```diff
@@ -97,12 +106,117 @@ std::optional<double> resolveForLoopTripCount(scf::ForOp forOp) {
...
+    for (int64_t inductionValue = range.lower; inductionValue < upperBound;
+         inductionValue += range.step) {
+        TripCountSummary nested = accumulateTripCounts(ranges, loopIndex + 1, inductionValue);
+        summary.total += nested.total;
+        summary.invocations += nested.invocations;
+    }
+    return summary;
+}
+
+static double computeAverageTripCountByEnumeration(llvm::ArrayRef<LoopRange> ranges,
```
jzaia18: Is this effectively just iterating through the loops? It feels like we should be able to compute these values analytically without having to actually iterate... although maybe the formula ends up too complicated

jzaia18: Nevermind, did some math to compute this. We can extend the closed form to deal with lower bounds, but not to deal with step sizes without making a complicated sum. The solution you have here is good as-is

doc/releases/changelog-dev.md
sengthai: ✅
[41e9565](https://github.com/PennyLaneAI/catalyst/pull/3140/commits/41e9565d684f5fabfffbe9c3ad3f5066a478fa58)

mlir/include/Catalyst/Utils/SCFUtils.h
sengthai: Yep! now we support that. 🎉
[9ace010](https://github.com/PennyLaneAI/catalyst/pull/3140/commits/9ace0101c4f33de6a8077b9d6eed4f8c4d61d90c)

sengthai: It works when I tested on mlir test. I will try to test on pytest once i finished the build.

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: There's no technical reason. i think it would be cleaner, but i can restore the explicit type if preferred.

mlir/lib/Catalyst/Utils/SCFUtils.cpp
sengthai: Actually, the purpose is to handle empty positive step forloop. 

https://github.com/PennyLaneAI/catalyst/blob/41e9565d684f5fabfffbe9c3ad3f5066a478fa58/mlir/test/Catalyst/ResourceAnalysisTest.mlir#L982-L989

And my understanding is that revese loops are normalized during lowering.

mlir/lib/Catalyst/Utils/SCFUtils.cpp
```diff
@@ -97,12 +106,123 @@ std::optional<double> resolveForLoopTripCount(scf::ForOp forOp) {
...
+    return summary.total / static_cast<double>(summary.invocations);
+}
+
+std::optional<double> resolveDirectNestedForLoopAverageTripCount(scf::ForOp forOp) {
+    llvm::SmallVector<LoopRange> ranges;
+    int64_t outerUpper = 0;
+    scf::ForOp currentLoop = forOp;
+    while (currentLoop) {
+        auto lower = getConstantIntValue(currentLoop.getLowerBound());
+        auto step = getConstantIntValue(currentLoop.getStep());
```
jzaia18:
```suggestion
        auto lower = resolveConstantInt(currentLoop.getLowerBound());
        auto step = resolveConstantInt(currentLoop.getStep());
```

Better to use these since they can handle more ops (e.g. StableHLO constants)

mlir/lib/Catalyst/Utils/SCFUtils.cpp
```diff
@@ -97,12 +106,123 @@ std::optional<double> resolveForLoopTripCount(scf::ForOp forOp) {
...
+
+            int64_t span;
+            if (__builtin_mul_overflow(*iterationCount, *step, &span) ||
+                __builtin_add_overflow(*lower, span, &outerUpper)) {
+                return std::nullopt;
+            }
+            break;
+        }
+
+        if (auto upper = getConstantIntValue(currentLoop.getUpperBound())) {
```
jzaia18:
```suggestion
        if (auto upper = resolveConstantInt(currentLoop.getUpperBound())) {
```

Same comment as above

mlir/lib/Catalyst/Analysis/ResourceResult.cpp
```diff
@@ -185,7 +185,7 @@ llvm::json::Object ResourceResult::toJson() const {
     llvm::json::Object functionObject;
     llvm::json::Object staticFunctionObject;
     for (const auto &entry : functionCalls) {
-        staticFunctionObject[entry.getKey()] = countToJson(entry.getValue());
+        staticFunctionObject[entry.getKey()] = llvm::json::Value(entry.getValue());
```
jzaia18, line 194: Why this change here?

frontend/test/pytest/test_specs.py
```diff
@@ -1261,6 +1261,211 @@ def loop(i, x):
...
+                for j in range(1, i):
+                    for k in range(0, j, 2):
+                        qp.PauliZ(wires=k % 2)
+
+            return qp.expval(qp.X(0))
+
+        resources = qp.specs(circuit, level=0)().resources
+
+        # Expect a symbolic value: combined complexities are not supported for concretization
+        assert resources.quantum_operations["PauliZ"] == 20
```
jzaia18:
```suggestion
        assert resources.quantum_operations["PauliZ"] == 20
```

Whoops, my bad

mlir/lib/Catalyst/Utils/SCFUtils.cpp
sengthai: Since all inputs are integers, so to use ceil we need to convert to double first, otherwise integer division truncates first then may lose precision. So, doing integer division can keep the exact count.

mlir/lib/Catalyst/Utils/SCFUtils.cpp
sengthai: actually, that's bad name. "Canonical" here is actually when the loops starting at 0 with step 1, and when each innner bound is the parent induction variable. 


```
for i in range(N):
   for j in range(i):
      for k in range(j):
        for w in range(k):
           ....
```

This pattern allows a closed form shortcut for the average. 
https://github.com/PennyLaneAI/catalyst/blob/41e9565d684f5fabfffbe9c3ad3f5066a478fa58/mlir/lib/Catalyst/Utils/SCFUtils.cpp#L202-L203

sengthai: i renamed it to be more clearer.

[8b7e5ce](https://github.com/PennyLaneAI/catalyst/pull/3140/commits/8b7e5ce176c64f5ff74b1533bfe6b00bdb9a97ec)

mlir/lib/Catalyst/Utils/SCFUtils.cpp
sengthai: Sure, I added some comment of that here.

[3d5a0f6](https://github.com/PennyLaneAI/catalyst/pull/3140/commits/3d5a0f678b70a5298357d6e66acba446f3495d61)

https://github.com/PennyLaneAI/catalyst/blob/3d5a0f678b70a5298357d6e66acba446f3495d61/mlir/lib/Catalyst/Utils/SCFUtils.cpp#L120-L129

mlir/lib/Catalyst/Utils/SCFUtils.cpp
sengthai: >  We can extend the closed form to deal with lower bounds, but not to deal with step sizes without making a complicated sum

That's right. Now we can handle lower bounds as well. ✅ 
[06f6750](https://github.com/PennyLaneAI/catalyst/pull/3140/commits/06f6750de104a76ca3d25a1bb66fa5d074dadc04)

mlir/lib/Catalyst/Utils/SCFUtils.cpp
sengthai: Good idea!

mlir/lib/Catalyst/Analysis/ResourceResult.cpp
sengthai: Because now static-function-call value could be a double, and `countToJson` could rounds them as integers.

frontend/test/pytest/test_specs.py
```diff
@@ -1261,6 +1289,250 @@ def loop(i, x):
...
+                    qp.PauliZ(wires=j % 2)
+
+            return qp.expval(qp.X(0))
+
+        resources = qp.specs(circuit, level=0)().resources
+
+        # Expect a symbolic value: indirect dependency is not supported for concretization
+        assert not isinstance(resources.quantum_operations["PauliZ"], (int, float))
+
+    # FIXME: This case is failing
```
jzaia18: This case is failing, we should provide a fix

doc/releases/changelog-dev.md
```diff
@@ -118,6 +118,29 @@
...
+  def circuit():
+      for i in range(8):
+          for j in range(i):
+              for _ in range(j):
+                  qp.PauliX(0)
+
+      return qp.expval(qp.X(0))
+
+  resources = qp.specs(circuit, level="all-mlir")().resources
+  print(resources.quantum_operations["PauliX"])  # 56
```
```suggestion
  resources = qp.specs(circuit, level=0)().resources
  print(resources.quantum_operations["PauliX"])  # 56
```

"all-mlir" is deprecated and returns a list instead of a single entry

mlir/test/Catalyst/ResourceAnalysisTest.mlir
jzaia18, line 1: We should add some tests that make sure the flattened resource method works well with `doubles` instead of `int`s

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
jzaia18: No strong opinion on my end, was just curious if the type changed or something

mlir/lib/Catalyst/Utils/SCFUtils.cpp
jzaia18: > And my understanding is that revese loops are normalized during lowering.

I'm okay with this as long as there's a strict guarantee. Is this an official thing the compiler *must* do, or is it just something we typically do which might change in the future?

frontend/test/pytest/test_specs.py
sengthai: I made quick fix here. [1d8cf19]
(https://github.com/PennyLaneAI/catalyst/pull/3140/commits/1d8cf19069884c8d1260517416fa75f47c8ddc61)

mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp
sengthai: Ahh i see.

mlir/lib/Catalyst/Utils/SCFUtils.cpp
sengthai: It is not official compiler thing, it's normalized in Catalyst's frontend. 
https://github.com/PennyLaneAI/catalyst/blob/25ef73f21c48cdbc2093d4d1ab38673ab270428b/frontend/catalyst/jax_primitives.py#L2725-L2726

mlir/include/Catalyst/Utils/SCFUtils.h
```diff
@@ -50,4 +50,15 @@ std::optional<double> getEstimatedIterationsHint(Operation *op);
...
 
+// Resolve the average trip count of a loop whose upper bound is a fixed constant or the
+// induction variable of any enclosing `scf.for` loop, including chains where intermediate
+// loops are skipped or reused by multiple descendants. An enclosing loop whose induction
+// variable is not used composes as a scalar multiplicity from its own `catalyst.
+// estimated_iterations` hint (integer or fractional) or static trip count. An integer hint on
+// an enclosing loop whose induction variable *is* used supplies the first K induction values;
+// a fractional hint cannot, so other unresolved bounds (arithmetic expressions, unrelated
+// dynamic loops, `scf.if` barriers, a fractional estimate needed as an induction domain) return
+// std::nullopt.
```
jzaia18:
```suggestion
// Resolve the average trip count of a loop whose upper bound is 1) a fixed constant or 2) 
// the loop variable of any enclosing scf.for loop, including chains of nested loops regardless
// of if the loops' variables are reused by subsequent loops. An enclosing 
// loop whose induction variable is not used composes as a scalar multiplicity from its own 
// `catalyst.estimated_iterations` hint (integer or fractional) or static trip count. An integer 
// hint on an enclosing loop whose induction variable *is* used supplies the first K induction 
// values; a fractional hint cannot, so other unresolved bounds (arithmetic expressions, 
// unrelated dynamic loops, `scf.if` barriers, a fractional estimate needed as an induction 
// domain) return std::nullopt.
```

mlir/lib/Catalyst/Utils/SCFUtils.cpp
```diff
@@ -97,12 +108,217 @@ std::optional<double> resolveForLoopTripCount(scf::ForOp forOp) {
...
     auto step = resolveConstantInt(forOp.getStep());
-    if (lb && ub && step && *step != 0 && *ub > *lb) {
-        return static_cast<double>((*ub - *lb + *step - 1) / *step);
+    if (lb && ub && step && *step > 0 && *ub > *lb) {
+        return static_cast<double>(computeTripCount(*lb, *ub, *step));
     }
     return std::nullopt;
 }
 
+// Maps each already-enumerated loop's induction variable to its current value.
```
jzaia18: What does "current value" mean here exactly?

mlir/lib/Catalyst/Utils/SCFUtils.cpp
```diff
@@ -97,12 +108,217 @@ std::optional<double> resolveForLoopTripCount(scf::ForOp forOp) {
...
+    }
+    auto iterationCount = intAttr.getValue().trySExtValue();
+    if (!iterationCount || *iterationCount < 0) {
+        return std::nullopt;
+    }
+    int64_t span;
+    int64_t upper;
+    if (__builtin_mul_overflow(*iterationCount, step, &span) ||
+        __builtin_add_overflow(lower, span, &upper)) {
+        return std::nullopt;
```
jzaia18, line 184: Should probably warn here somehow if an overflow happened

mlir/lib/Catalyst/Utils/SCFUtils.cpp
```diff
@@ -97,12 +108,217 @@ std::optional<double> resolveForLoopTripCount(scf::ForOp forOp) {
...
+        unresolved.erase(currentLoop.getInductionVar());
+        if (!closesOwnUpperBound(currentLoop)) {
+            unresolved.insert(currentLoop.getUpperBound());
+        }
+        if (unresolved.empty()) {
+            std::reverse(chain.begin(), chain.end());
+            return chain;
+        }
+
+        auto parent = dyn_cast_or_null<scf::ForOp>(currentLoop->getParentOp());
```
jzaia18, line 236: What if the loop's parent isn't an `scf::ForOp`?

e.g. something like

```python
for i in range(n):
    if i % 2 == 0:
        for j in range(i):
            ...
```

mlir/lib/Catalyst/Utils/SCFUtils.cpp
```diff
@@ -97,12 +108,217 @@ std::optional<double> resolveForLoopTripCount(scf::ForOp forOp) {
...
+    }
+}
+
+// Fast path for the canonical shape where every loop shares a common lower bound, unit step,
+// and uses only its immediate predecessor's induction variable as its upper bound:
+// for i in 0..8
+//     for j in 0..i
+//         for k in 0..j (average trip count is 2)
+//             ...
+// The average trip count is (8 - 3 + 1) / 3 = 2 for this immediate-parent chain.
```
jzaia18:
```suggestion
// Fast path for the most common case, where every loop shares a common lower bound,
// has a step of 1 and uses only its immediate predecessor's loop variable as its upper bound:
// for i in 0..8
//     for j in 0..i
//         for k in 0..j (average trip count is 2)
//             ...
// The average trip count is (8 - 3 + 1) / 3 = 2 for this immediate-parent chain.
```

mlir/test/Catalyst/ResourceAnalysisTest.mlir
sengthai: The operation-count type is double and as flattened resource method just sum up the operation count, then it works well.

https://github.com/PennyLaneAI/catalyst/blob/25ef73f21c48cdbc2093d4d1ab38673ab270428b/mlir/lib/Catalyst/Analysis/ResourceAnalysis.cpp#L476-L477

mlir/lib/Catalyst/Utils/SCFUtils.cpp
```diff
@@ -97,12 +108,217 @@ std::optional<double> resolveForLoopTripCount(scf::ForOp forOp) {
...
+    }
+    InductionValues noInductionValues;
+    auto outerUpper = resolveEnumerableUpperBound(outer, *commonLower, 1, noInductionValues);
+    if (!outerUpper) {
+        return std::nullopt;
+    }
+    int64_t depth = static_cast<int64_t>(chain.size());
+    double average =
+        static_cast<double>(*outerUpper - *commonLower - depth + 1) / static_cast<double>(depth);
+    return std::max(0.0, average);
```
jzaia18:
```suggestion
    auto outerLowerBound = resolveConstantInt(outer.getLowerBound());
    if (!outerLowerBound) {
        return std::nullopt;
    }
    for (size_t i = 0; i < chain.size(); i++) {
        scf::ForOp loop = chain[i];
        if (resolveConstantInt(loop.getLowerBound()) != outerLowerBound ||
            resolveConstantInt(loop.getStep()) != 1) {
            return std::nullopt;
        }
        if (i > 0 && loop.getUpperBound() != scf::ForOp(chain[i - 1]).getInductionVar()) {
            return std::nullopt;
        }
    }
    InductionValues noInductionValues;
    auto outerUpperBound = resolveEnumerableUpperBound(outer, *outerLowerBound, 1, noInductionValues);
    if (!outerUpperBound) {
        return std::nullopt;
    }
    int64_t depth = static_cast<int64_t>(chain.size());
    double average =
        static_cast<double>(*outerUpperBound - *outerLowerBound - depth + 1) / static_cast<double>(depth);
    return std::max(0.0, average);
```

mlir/lib/Catalyst/Utils/SCFUtils.cpp
sengthai: Unfortunately, it will not work for now. 
https://github.com/PennyLaneAI/catalyst/blob/25ef73f21c48cdbc2093d4d1ab38673ab270428b/mlir/test/Catalyst/ResourceAnalysisTest.mlir#L1121-L1144

mlir/lib/Catalyst/Utils/SCFUtils.cpp
```diff
@@ -97,12 +108,217 @@ std::optional<double> resolveForLoopTripCount(scf::ForOp forOp) {
...
+}
+
+// Store the target loop's total iterations and how many times it is reached across every
+// enumerated context. Dividing the total by the invocation count gives its average trip count.
+struct TripCountSummary {
+    double total = 0.0;
+    double invocations = 0.0; // how many times the loop is reached across every enumerated context
```
jzaia18:
```suggestion
    double totalIterations = 0.0;  // The total number of times the loop body is executed
    double entryCount = 0.0;  // The total number of times this loop is reached by the outer caller
```

mlir/lib/Catalyst/Utils/SCFUtils.cpp
```diff
@@ -97,12 +108,217 @@ std::optional<double> resolveForLoopTripCount(scf::ForOp forOp) {
...
+// Evaluate chain[position..], given the induction values recorded for already-enumerated
+// enclosing loops and the scalar weight contributed by loops folded so far. A loop is enumerated
+// (its induction values tracked) only when a later loop's upper bound directly uses it;
+// otherwise its own trip count is folded into `pathWeight` without enumeration.
+static std::optional<TripCountSummary> evaluateChain(llvm::ArrayRef<scf::ForOp> chain,
+                                                     size_t position, double pathWeight,
+                                                     InductionValues &inductionValues) {
+    scf::ForOp loop = chain[position];
+    llvm::ArrayRef<scf::ForOp> descendants = chain.drop_front(position + 1);
+    bool isReferencedLater =
```
jzaia18, line 297:
```suggestion
    // Whether or not *any* subsequent loop's trip count depends on the current induction variable
    bool isReferencedLater =
```

mlir/lib/Catalyst/Utils/SCFUtils.cpp
```diff
@@ -97,12 +108,217 @@ std::optional<double> resolveForLoopTripCount(scf::ForOp forOp) {
...
+        std::any_of(descendants.begin(), descendants.end(), [&](scf::ForOp descendant) {
+            return descendant.getUpperBound() == loop.getInductionVar(); 
+        });
+
+    // Loop of induction variable is not upperbound of any later loop.
```
jzaia18:
```suggestion
    // Current loop's induction variable does not affect upper bound of any subsequent loop.
    // Can just get this loop's trip count and continue
```

mlir/lib/Catalyst/Utils/SCFUtils.cpp
jzaia18: Will not work meaning that it will just remain dynamic? (seems like yes based on that snippet)

jzaia18: Remaining dynamic is fine, I just want to make sure it doesn't return something incorrect

mlir/lib/Catalyst/Utils/SCFUtils.cpp
sengthai: ✅ 
[1279fa7](https://github.com/PennyLaneAI/catalyst/pull/3140/commits/1279fa716d620750addefdbdec567bea43b8654c)

mlir/lib/Catalyst/Utils/SCFUtils.cpp
sengthai: Yes, it will fallback to dynamic one.

mlir/lib/Catalyst/Utils/SCFUtils.cpp
jzaia18: We should flip the order of these conditions to remain consistent with `resolveEnumerableUpperBound`. If we can get it statically I think we should prefer the known value over the user hint

mlir/lib/Catalyst/Utils/SCFUtils.cpp
sengthai: it is value of concrete induction in that specific iteration, example `for i in range(8)` and current value here meant that in the iterations when `%i is 3`. 

I updated the document to be more clearer. 

https://github.com/PennyLaneAI/catalyst/blob/5c4bdc791f4e848b9a4cb1052417837921e5203a/mlir/lib/Catalyst/Utils/SCFUtils.cpp#L118-L120

mlir/include/Catalyst/Utils/SCFUtils.h
jzaia18, line 14:
```suggestion

#pragma once
```

mlir/lib/Catalyst/Utils/SCFUtils.cpp
sengthai: Well consideration! 
Updated>>>
https://github.com/PennyLaneAI/catalyst/blob/facfd31a48abcc46c517dc767e859014663fe9c1/mlir/lib/Catalyst/Utils/SCFUtils.cpp#L103-L108

mlir/lib/Catalyst/Utils/SCFUtils.cpp
jzaia18: Is it a written guarantee of Catalyst somewhere? If not then I think we should not depend on this behaviour

jzaia18 (approved):
Tentative approval from me, everything seems to be working pretty well. You can override the CIs to pull PL from https://github.com/PennyLaneAI/pennylane/pull/10044/ temporarily to get the CIs green (just make sure that we merge that PR first and remove the CI pin before merging this PR)

mlir/lib/Catalyst/Utils/SCFUtils.cpp
sengthai: Catalyst doesn't guarantee, but it states in MLIR itself. So i think we are safe here.

```
   The step is a value of same type but required to be positive, the lower and
    upper bounds can be also negative or zero.
```
https://github.com/llvm/llvm-project/blob/8f264586d7521b0e305ca7bb78825aa3382ffef7/mlir/include/mlir/Dialect/SCF/IR/SCFOps.td#L171-L172

mlir/lib/Catalyst/Utils/SCFUtils.cpp
```diff
@@ -12,10 +12,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <cmath> // std::ceil()
+#include "Catalyst/Utils/SCFUtils.h"
```
rniczh, line 15: should after L31?

mlir/lib/Catalyst/Utils/SCFUtils.cpp
```diff
@@ -88,21 +105,244 @@ std::optional<double> getEstimatedIterationsHint(Operation *op) {
...
+struct TripCountSummary {
+    double totalIterations = 0.0; // The total number of times the loop body is executed
+    double entryCount = 0.0; // The total number of times this loop is reached by the outer caller
+};
+
+// Evaluate chain[position..], given the induction values recorded for already-enumerated
+// enclosing loops and the scalar weight contributed by loops folded so far. A loop is enumerated
+// (its induction values tracked) only when a later loop's upper bound directly uses it;
+// otherwise its own trip count is folded into `pathWeight` without enumeration.
+static std::optional<TripCountSummary> evaluateChain(llvm::ArrayRef<scf::ForOp> chain,
```
rniczh, line 290: Do you think this is the right approach? What if you have relatively large loop bounds or several levels of dependent loops? Outside the closed-form path, evaluateChain recursively enumerates induction values, so compile time could grow to `O(N^2)` or worse

mlir/lib/Catalyst/Utils/SCFUtils.cpp
```diff
@@ -28,9 +36,12 @@ using namespace mlir;
...
 
-static int64_t getNumIterations(double lowerBound, double upperBound, double step) {
-    assert(upperBound >= lowerBound && step > 0);
-    return std::ceil((upperBound - lowerBound) / step);
+static int64_t computeTripCount(int64_t lowerBound, int64_t upperBound, int64_t step) {
+    assert(step > 0);
+    if (upperBound <= lowerBound) {
+        return 0;
+    }
+    return (upperBound - lowerBound + step - 1) / step;
```
rniczh: may overflow

mlir/lib/Catalyst/Utils/SCFUtils.cpp
```diff
@@ -88,21 +105,244 @@ std::optional<double> getEstimatedIterationsHint(Operation *op) {
...
+    if (!lower || !step || *step <= 0) {
+        return std::nullopt;
+    }
+    auto upper = resolveEnumerableUpperBound(loop, *lower, *step, inductionValues);
+    if (!upper) {
+        return std::nullopt;
+    }
+
+    TripCountSummary summary;
+    for (int64_t iv = *lower; iv < *upper; iv += *step) {
```
rniczh: may overflow, a standard way is using llvm::checkedAdd as we are using llvm infra

mlir/lib/Catalyst/Utils/SCFUtils.cpp
```diff
@@ -88,21 +105,244 @@ std::optional<double> getEstimatedIterationsHint(Operation *op) {
...
+
+    InductionValues inductionValues;
+    auto summary = evaluateChain(*chain, 0, /*pathWeight=*/1.0, inductionValues);
+    if (!summary) {
+        return std::nullopt;
+    }
+    if (summary->entryCount == 0.0) {
+        return 0.0;
+    }
+    return summary->totalIterations / summary->entryCount;
```
rniczh, line 375: Store the float number is the right approach? I think it would introduce some problems? like this: https://github.com/PennyLaneAI/pennylane/pull/10044/changes#r3854347985

mlir/lib/Catalyst/Utils/SCFUtils.cpp
sengthai: Yes, i think so. Unfortunately, it is format automatically, the CI won't be happy.

mlir/lib/Catalyst/Utils/SCFUtils.cpp
sengthai: It is not the best approach, but it would get the concrete result. We can avoid the worse scenario on large circuit  by setting the maximum iteration steps (let say 1M steps). If it exceeds the 1M step, it fallbacks to symbolic result. or do you have any recommendation?

sengthai: Improved.
[dee4859](https://github.com/PennyLaneAI/catalyst/pull/3140/commits/dee4859e5360b903529faf49209477369ab6fbe5)

mlir/lib/Catalyst/Utils/SCFUtils.cpp
sengthai: fix! 
[88d0684](https://github.com/PennyLaneAI/catalyst/pull/3140/commits/88d0684b7b798ca01f0391f84fe457c058db3ccc)

mlir/lib/Catalyst/Utils/SCFUtils.cpp
sengthai: ✅ 
[88d0684](https://github.com/PennyLaneAI/catalyst/pull/3140/commits/88d0684b7b798ca01f0391f84fe457c058db3ccc)

mlir/lib/Catalyst/Utils/SCFUtils.cpp
sengthai: Because of the form of our JSON and we want to preserve the information of each for-loop, the average-approach with resulting float type here is the most fit with JSON output right now.

rniczh: Has been discussed offline, fine for now.

rniczh (approved):

dime10:
Great to see this! One thing I didn't see in the tests is the dynamic iteration index being used as the **start** of the inner iteration, I think that's fairly common. Is that supported (or being added)?

sengthai:
@dime10 Good catch! Actually, using the induction variable as start in the inner iteration is not currently supported, so   the analysis will falls back to a dynamic count.

PR #3141, opened 2026-08-21
fix: restore lll compatibility up to dev80

andrijapau:
**Context:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/frontend/catalyst/decomposition/type_utils.py
+++ b/frontend/catalyst/decomposition/type_utils.py
@@ -169,7 +169,7 @@ def replace_abstract_wires_with_concrete_wires(node):
         return tuple(replace_abstract_wires_with_concrete_wires(item) for item in node)
     else:
         if isinstance(node, qp.typing.AbstractWires):
-            return qp.wires.Wires(range(node.num_wires))
+            return qp.wires.Wires(range(len(node)))
         else:
             return node
 
--- a/frontend/catalyst/decomposition/type_utils.py
+++ b/frontend/catalyst/decomposition/type_utils.py
@@ -182,7 +182,7 @@ def _replace_op_abstract_wires_with_concrete_wires(op2):
     new_op = copy.deepcopy(op2)
     for wire_arg in new_op.wire_argnames:
         if isinstance(new_op.arguments[wire_arg], qp.typing.AbstractWires):
-            num_wires = new_op.arguments[wire_arg].num_wires
+            num_wires = len(new_op.arguments[wire_arg])
             new_op.arguments[wire_arg] = qp.wires.Wires(range(-1, -num_wires - 1, -1))
     for hybrid_arg in new_op.hybrid_argnames:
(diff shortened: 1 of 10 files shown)


frontend/catalyst/decomposition/type_utils.py
```diff
@@ -182,7 +182,7 @@ def _replace_op_abstract_wires_with_concrete_wires(op2):
     new_op = copy.deepcopy(op2)
     for wire_arg in new_op.wire_argnames:
         if isinstance(new_op.arguments[wire_arg], qp.typing.AbstractWires):
-            num_wires = new_op.arguments[wire_arg].num_wires
+            num_wires = len(new_op.arguments[wire_arg])
```
andrijapau, line 185: For context, see https://github.com/PennyLaneAI/pennylane/pull/10032

frontend/catalyst/from_plxpr/qref_operator2_primitives.py
```diff
@@ -364,8 +364,35 @@ def compile_decomp_rules(
...
+        # rules that call other funcops
+        # When the above is implemented, uncomment the BasisState decomp rule collection impl below
+
+        op_id = ""
+        decomp_rules = []
+
+        # num_wires = wire_lens[0]
+        # matrix_size = 2**num_wires
+        # dynamic_shape = {
+        #     qp.QubitUnitary.dynamic_argnames[0]: [["complex<f64>"] * matrix_size] * matrix_size
```
paul0403, line 378: @kipawaa The rules for unitaries involve some sort of dot product and actually care about shape. If I give the extra braket (the extra vector layer in `StringMap<Vector<Type>>`) then dot product fails saying rank doesn't match, because the dummy arg for the matrix would also be rank 3.

I had to give it rank 2 here and manually add the `"["` and `"]"` back when generating the ID string.

Although, this is all future talk, since these sort of specific talks are not meaningful until I add support for rules that call helpers.

frontend/test/pytest/test_preprocess.py
```diff
@@ -578,29 +578,6 @@ def test_error_for_pennylane_midmeasure_decompose_nested(self, request):
...
-            capabilities = replace(capabilities, operations={})
-            kwargs = {"capabilities": capabilities}
-        else:
-            kwargs = {"target_gates": {}, "capabilities": None}
-
-        with pytest.raises(
-            CompileError,
-            match="not supported with catalyst on this device and does not provide a decomposition",
-        ):
-            _ = catalyst_decompose(tape, **kwargs)
```
paul0403, line 602: This failing test uses old tape decomposition, so it is using an _old old_ system (as in, it's not even using the python decomp graph: it's using the non-capture tape based decomp!). 

It's failing, so I'm just deleting it.

andrijapau (commented):
Changes LGTM, I just can't approve as I authored the PR.

paul0403:
duplicated by #3145

PR #3147, opened 2026-08-24
The CDF/CGF Trotter MLIR lowering  with numeric Hamiltonians

maliasadi:
**Context:**
Add the MLIR/lowering tests for numeric hamiltonians base and CDF/CGF types https://github.com/PennyLaneAI/pennylane/pull/10048. 

**Description of the Change:**

**Benefits:**
- Support numeric hamiltonians with both **concrete** and **abstract** tensor data as hybrid/dynamic args of Op2 classes.
- This implementation enables E2E support without requiring a dedicated dialect definition (the PR #3000 approach)

**Related GitHub Issues:**
[sc-128098]
[sc-128104]

PR #3149, opened 2026-08-24
Support decomp rules that call helper functions

paul0403:
**Context:**
Many decomposition rules call helper functions. For example, a rule for `qp.BasisState` calls `qp.math.allclose`.

In the current implementation, only the rule function is moved from the rule qjit to the user qjit, leaving out the helpers.
This produces bad IR with undefined symbols.

Inline helpers into the rule functions after compiling a rule function.

**Benefits:**
Rules with helper functions work.

**Possible Drawbacks:**
1. Inling is not possible when there's recursion. However this is fine: JAX cannot compile recursive functions in the first place, we can safely assume no decomp rules will have recursion.
2. There is now an invocation of the compiler at rule-gathering time.

However, if we don't do that, the alternative would be that we need to manually traverse the call graph of the rule and collect the called helpers ourselves. Doing the callgraph analysis is a pretty expensive process. 

On top of that, multiple rules compiled from different rule qjit instances might call the same helper function. If we are implementing a custom logic to move all called helper functions, then we also need to implement the logic to not move a helper function that's already been moved (otherwise multiple functions would have the same symbol in the IR and lead to a clash). 

For example, I wouldn't be surprised if another gate other than `qp.BasisState` (say `qp.A`) is calling `qp.math.allclose`. If we just move `func.func @allclose` when compiling rules for `BasisState`, what do we do when compiling rules for `qp.A` and seeing a call to `allclose`? Should we move this `allclose` too? Keep in mind the rule compilation qjits for the two gates are different qjit instances, so during the rule compilation for `qp.A`, all we have is the already-moved `func.func @allclose` from `BasisState` in the user qjit's MLIR. How do we judge if these functions are the same or not? Just having the same name is not sufficient for saying two functions are the same. If these are different helpers with the same name, then we would need to rename them, and then rename all the calls to them! Furthermore, JAX's mlir python bindings does not save the users of symbols. To rename calls to a function, we would need to manually walk and find the calls. At that point, the savings from not invoking the compiler is completely outweighed by this extra cost: we are doing call graph traversal and symbol table renaming! 

At that point, probably just invoking the compiler is better.

**Related GitHub Issues:**
[sc-128398]
[sc-128297]
[sc-128450]


frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -273,13 +275,54 @@ def update_funcop_attributes(op):
...
+        module.operation.walk(remove_qnode_func)
+
+    for qnode_func in qnode_func_erasure_worklist:
+        qnode_func.erase()
+
+    inlined = _quantum_opt(
+        "--inline=inlining-threshold=4294967295",  # Use uint max to indicate always inline
+        "--mlir-print-op-generic",
+        stdin=str(module),
+    )
```
dime10, line 436: I don't have all the context here, but are we inlining (invoking the compiler) for every unique operation? It does _sound_ quite expensive (although I don't know how it compares to the cost of tracing these functions in the first place). It could still be worth discussing the options we have.

This function seems to return a module with all the decomp rules and helpers inside it, so they are kept together at this stage. Perhaps we could collect all the rules for a program first and then inline (with parallelization enabled)? 

It would probably help if I looked into how the rules are processed after this step before thinking of additional solutions though.

dime10 (commented):
Just a first comment regarding the high-level structure, although I might need some updates on how the most recent version works here :)

Regarding the problem you describe in the PR info on call graph walking, inlining, and renaming, those steps sound pretty much exactly like what Erick [implemented](https://github.com/PennyLaneAI/catalyst/blob/8916740811254324e27eda30cc93a639e8a76fb0/mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp#L44-L46) for the nested module (quantum kernel) inlining. Not sure if it will be helpful here but just thought I'd mention it.

frontend/catalyst/decomposition/decomposition_rules.py
paul0403: > Perhaps we could collect all the rules for a program first

This was actually me first attempt, but turned out to be a bit difficult since we are using a new qjit instance to collect all the rules _for a specific operator_. 

Where would we know all the operators used in a program? During the lower time of each operator primitive, we just see that primitive itself, instead of the entire program. 

One place I could think of is immediately after receiving the plxpr, walk over the entire plxpr to collect all the operators that are used. But this approach has two difficulties:
1. jaxpr does not have a "walk" method. We would need to recursively inspect into nested jaxprs. It could be done but it's a bit annoying. 
2. The compiled entity we would be moving from the rule qjit to the user qjit is now a jaxpr instead of an MLIR. Unlike MLIR that has proper API to handle instructions, moving jaxpr would require using jaxpr internal methods (e.g. moving equations between the internal `jaxpr.eqns` lists). This might require manual editting of jaxpr contexts? Feels a bit error-prone.  Also for moving MLIR there's always the fallback route of serializing via strings, but I don't think it is possible to parse a jaxpr from strings?

Hmm, actually now that I'm typing this out, I thought of another place: after converting the user program's plxpr to MLIR (without the rules), we can probably do a walk over there and collect the necessary gates. We could also build a wrapper around the current `compile_decomposition_rules` function to work with a batch of gates, so there is only one new qjit instance to compile all the rules for all the gates! This means we just invoke the inliner once. This probably also solves the name-clashing problem since now all the different rules that potentially use the same helper functions are traced in the same qjit context. 

What do you think?

paul0403:
> Regarding the problem you describe in the PR info on call graph walking, inlining, and renaming, those steps sound pretty much exactly like what Erick [implemented](https://github.com/PennyLaneAI/catalyst/blob/8916740811254324e27eda30cc93a639e8a76fb0/mlir/lib/Catalyst/Transforms/InlineNestedModules.cpp#L44-L46) for the nested module (quantum kernel) inlining. Not sure if it will be helpful here but just thought I'd mention it.

@kipawaa and I want to handle the rules more properly in Q4 by putting them in their own modules, instead of just being flat functions that are simply appended to the user program. We can definitely design the rule module abstraction in a way that utilizes Erick's work!

For now though, let's just use the inliner to meet the Q3 deadline, since this is by far the easiest option

frontend/catalyst/decomposition/decomposition_rules.py
kipawaa: > Hmm, actually now that I'm typing this out, I thought of another place: after converting the user program's plxpr to MLIR (without the rules), we can probably do a walk over there and collect the necessary gates. We could also build a wrapper around the current compile_decomposition_rules function to work with a batch of gates, so there is only one new qjit instance to compile all the rules for all the gates! This means we just invoke the inliner once. This probably also solves the name-clashing problem since now all the different rules that potentially use the same helper functions are traced in the same qjit context.

Something like this is what I was trying to suggest yesterday - I think this would be _much_ more efficient. For now, we just need the functionality, but I think there's a lot of room for improvement here once we have the time to return to it

paul0403:
> after converting the user program's plxpr to MLIR (without the rules), we can probably do a walk over there and collect the necessary gates.

Just realized this wouldn't work as we will have lost the pytree structures into the UIDs. In fact anything later than lowering the individual gate primitives will have lost this information, making it impossible to query the resources on these gates. 

@kipawaa how does on-demand rule query the resources of ops with hybrid args? Is it even possible? Does it just pass `extra_data=None` into the entry point always?

kipawaa:
> > after converting the user program's plxpr to MLIR (without the rules), we can probably do a walk over there and collect the necessary gates.
> 
> Just realized this wouldn't work as we will have lost the pytree structures into the UIDs. In fact anything later than lowering the individual gate primitives will have lost this information, making it impossible to query the resources on these gates.
> 
> @kipawaa how does on-demand rule query the resources of ops with hybrid args? Is it even possible? Does it just pass `extra_data=None` into the entry point always?

Ah good point, this isn't possible. I still think there's a benefit to collecting all of the rules into one module, but again this can be a follow-up

paul0403:
> I still think there's a benefit to collecting all of the rules into one module

Agreed, but I don't see how to do this in the current structure tbh. It cannot be after lower time due to the loss of pytrees and `static_args`, but both tracing and lowering are interpreter-based, which means it only sees one gate at a time.

kipawaa:
> This function seems to return a module with all the decomp rules and helpers inside it, so they are kept together at this stage. Perhaps we could collect all the rules for a program first and then inline (with parallelization enabled)?

@paul0403 this approach is still available - we can collect all of the jax-mlir modules and run one compiler invocation on these before retrieving the decomposition rules and moving them to the user module. Up to you if you want to do this now or later

paul0403:
> > This function seems to return a module with all the decomp rules and helpers inside it, so they are kept together at this stage. Perhaps we could collect all the rules for a program first and then inline (with parallelization enabled)?
> 
> @paul0403 this approach is still available - we can collect all of the jax-mlir modules and run one compiler invocation on these before retrieving the decomposition rules and moving them to the user module. Up to you if you want to do this now or later

This involves renaming the helpers (e.g. two rules both calling `allclose`), which involves renaming the calls... I can try but I really don't like any implementation I can think of. All of them involve very ugly walking.

frontend/catalyst/from_plxpr/qref_operator2_primitives.py
```diff
@@ -338,61 +338,48 @@ def compile_decomp_rules(
         )
 
     elif op_cls is qp.BasisState:
```
kipawaa, line 340: 🥳

frontend/catalyst/from_plxpr/qref_operator2_primitives.py
```diff
@@ -338,61 +338,48 @@ def compile_decomp_rules(
...
+
+        decomp_rules = fetch_all_reachable_decomposition_rules_from_op(
+            op_name="BasisState",
+            op_id=op_id,
+            dynamic_shape=dynamic_shape,
+            wire_lens={f"{wire_argname}": num_wires},
+            static_data={},
+        )
 
     elif op_cls is qp.QubitUnitary:
```
kipawaa, line 361: 🏆

frontend/test/lit/test_operator2/test_lower_time_rules.py
```diff
@@ -549,6 +549,53 @@ def from_custom():
...
+# CHECK: stablehlo.constant dense<4.200000e+00> : tensor<f64>
+# CHECK-NOT: call
+# CHECK-NOT: my_helper
+#
+# CHECK: func.func private @"__builtin_rule_NoParams{}{reg:2}{}"
+# CHECK-SAME:   resources = {operations = {
+# CHECK-SAME:   "MultiParams{a:{{\[\[f64\]\]}},b:{{\[\[f64\]\]}},c:{{\[\[f64\]\]}}}{reg:1}{}" = 1 : i64
+# CHECK-SAME:   target_gate = "NoParams{}{reg:2}{}"
+# CHECK: stablehlo.constant dense<4.200000e+00> : tensor<f64>
+# CHECK-NOT: call
```
kipawaa, line 594: Hm, if these are after the other `CHECK` directives then they only ensure it doesn't follow the others. Should we simplify this test by using `CHECK-NOT` before and after the main decomp rule (the one that calls the helper) and keep the core decomp rule check as minimal as possible?

frontend/test/lit/test_operator2/test_special_ops_decomp.py
```diff
@@ -109,6 +110,73 @@ def pcphase():
...
+# CHECK: func.func private @"__builtin_xyx_QubitUnitary
+# CHECK-SAME: {U:{{\[\[}}complex<f64>,complex<f64>],
+# CHECK-SAME:     [complex<f64>,complex<f64>{{\]\]}}}{wires:1}{}"
+# CHECK-SAME:   target_gate = "QubitUnitary
+# CHECK-SAME:   {U:{{\[\[}}complex<f64>,complex<f64>],
+# CHECK-SAME:       [complex<f64>,complex<f64>{{\]\]}}}{wires:1}{}"
+#
+# CHECK: func.func private @"__builtin_two_qubit_decomp_rule_QubitUnitary
+# CHECK-SAME: {U:{{\[\[}}complex<f64>,complex<f64>,complex<f64>,complex<f64>],
+# CHECK-SAME:     [complex<f64>,complex<f64>,complex<f64>,complex<f64>],
```
kipawaa: It'll be so nice to simplify this once #3158 goes in 😅

kipawaa (approved):
Thanks @paul0403 !

maliasadi (approved):
LGTM 🎉

frontend/test/lit/test_operator2/test_lower_time_rules.py
paul0403: Sorry what do you mean? 😅 Isn't the point to check that the core rule body does not have a call to the helper?

frontend/catalyst/from_plxpr/qref_operator2_primitives.py
```diff
@@ -337,32 +337,13 @@ def compile_decomp_rules(
             static_data={},
         )
 
-    elif op_cls is qp.BasisState:
```
kipawaa, line 340: Are we removing support for `BasisState`?

kipawaa (approved):
Beautiful 😍

frontend/test/lit/test_operator2/test_lower_time_rules.py
kipawaa: well for the call yes, but for the helper function definition itself we would want to verify that it doesn't exist in the circuit at all after being inlined. No big deal though, checking that it doesn't call is sufficient

frontend/catalyst/from_plxpr/qref_operator2_primitives.py
paul0403: https://xanaduhq.slack.com/archives/C030UB33QH3/p1788191418049839

paul0403:
Oh, lowering `qp.BasisState` to `OperatorOp` means its no longer executable in runtime :face_exhaling: 

I'll revert the change in this PR back to lowering to `SetStateOp` for CI to pass and follow up with another PR to deal with it. In the follow-up PR it would be:
- `qp.BasisState` lowers to operator op, so it can participate in graph decomp
- Add a pass that converts operator ops of "BasisState" to `SetBasisStateOp`.

frontend/catalyst/from_plxpr/qref_operator2_primitives.py
paul0403: will deal with basis state in a followup PR

frontend/catalyst/from_plxpr/qref_operator2_primitives.py
```diff
@@ -338,61 +338,50 @@ def compile_decomp_rules(
...
-        #     op_id=op_id,
-        #     dynamic_shape=dynamic_shape,
-        #     wire_lens={f"{wire_argname}": wire_lens[0]},
-        #     static_data={},
-        # )
+        num_wires = wire_lens[0]
+        matrix_size = 2**num_wires
+        dynamic_shape = {
+            qp.QubitUnitary.dynamic_argnames[0]: [
+                f"tensor<{matrix_size}x{matrix_size}xcomplex<f64>>"
```
maliasadi, line 366: could we not hardcode `f64` here and get it from the matrix `U`?

frontend/test/lit/test_operator2/test_special_ops_decomp.py
maliasadi, line 1: Could we also test with abstract data types?

frontend/catalyst/decomposition/decomposition_rules.py
```diff
@@ -397,13 +399,54 @@ def update_funcop_attributes(op):
...
+        return ir.WalkResult.ADVANCE
+
+    with module.context, ir.Location.unknown():
+        module.operation.walk(remove_qnode_func)
+
+    for qnode_func in qnode_func_erasure_worklist:
+        qnode_func.erase()
+
+    inlined = _quantum_opt(
+        "--inline=inlining-threshold=4294967295",  # Use uint max to indicate always inline
```
maliasadi, line 433: Does this also canonicalize every rule/helper? So it is technically inlining helper funcs and canonicalize every decomp rule body reached during compilation. I didn't see it mentioned anywhere in this PR.

maliasadi: Could we skip the canonicalization of rules at this step?

frontend/catalyst/decomposition/decomposition_rules.py
maliasadi:
```suggestion
) -> ir.Operation:
```

maliasadi:
```suggestion
    Return the top-level ``builtin.module`` operation containing the decomposition rules for an operator instance.
```

frontend/catalyst/decomposition/decomposition_rules.py
paul0403: Unfortunately the inliner seems to run canonicalization within it.

frontend/catalyst/from_plxpr/qref_operator2_primitives.py
paul0403: I wanted to avoid passing in the actual args into the special decomp rule getter (`compile_decomp_rules`) because the types of all these special ops' params are contracted via the actual MLIR tablegen op def already.

frontend/catalyst/decomposition/decomposition_rules.py
paul0403: https://github.com/PennyLaneAI/catalyst/pull/3149/commits/361ee0dd38b4b47d08ea83c4bb26cdc35aec2f50

paul0403: https://github.com/PennyLaneAI/catalyst/pull/3149/commits/361ee0dd38b4b47d08ea83c4bb26cdc35aec2f50

frontend/catalyst/from_plxpr/qref_operator2_primitives.py
maliasadi: It's so unfortunate! Also `TensorLike` doesn't really guarantee `dtype`. Good for now! we can revisit it later

maliasadi (approved):
:shipit:

frontend/catalyst/from_plxpr/qref_operator2_primitives.py
paul0403: > `TensorLike` doesn't really guarantee `dtype`

I think it is guaranteed, we parse type strings like `tensor<axbxdtype>` via a regex in type_utils.py https://github.com/PennyLaneAI/catalyst/blob/361ee0dd38b4b47d08ea83c4bb26cdc35aec2f50/frontend/catalyst/decomposition/type_utils.py#L110

frontend/test/lit/test_operator2/test_special_ops_decomp.py
paul0403: Do you mean things like `qp.MultiRZ(qp.typing.Float, ...)`?

maliasadi: yup! let's get back to it as a follow-up. I need the updates in this PR for #3147 :)

PR #3152, opened 2026-08-25
Fix usage of LLVM_DIR variable in build system

joeycarter:
**Context:** The Catalyst Runtime build system uses the variable `LLVM_DIR` to store the path to the root of the `llvm-project` source tree, i.e. `catalyst/mlir/llvm-project/`. This variable conflicts with CMake standards and conventions used by the [`find_package`](https://cmake.org/cmake/help/latest/command/find_package.html) command. The variable `<PackageName>_DIR` stores the path to the directory containing the `<PackageName>Config.cmake` file that CMake uses to configure and import a package. This variable should typically point to `<install_location>/lib/cmake/<PackageName>/`. For example, the `LLVMConfig.cmake` config file is located at `.../llvm-project/build/lib/cmake/llvm/LLVMConfig.cmake`.

Importantly, if the current environment sets the variable `LLVM_DIR`, then the Catalyst Runtime build configuration fails because the Makefile picks up this environment variable and retrieves the MLIR include directory from `$(LLVM_DIR)/mlir/include/`.

**Description of the Change:** This PR renames the Makefile variable pointing to the root of the `llvm-project` source tree from `LLVM_DIR` to `LLVM_SRC_DIR`. The GitHub workflow configs have also been updated to use `LLVM_SRC_DIR`.

**Benefits:** This change will make Catalyst plugin development easier, since in some cases they require using the `LLVM_DIR` environment variable to point to the local LLVM installation.


runtime/Makefile
```diff
@@ -85,8 +85,7 @@ configure:
 		-DCMAKE_BUILD_TYPE=$(BUILD_TYPE) \
 		-DCMAKE_LIBRARY_OUTPUT_DIRECTORY=$(RT_BUILD_DIR)/lib \
 		-DCMAKE_C_COMPILER=$(C_COMPILER) \
-		-DMLIR_INCLUDE_DIRS=$(LLVM_DIR)/mlir/include \
-		-DLLVM_DIR=$(LLVM_BUILD_DIR)/lib/cmake/llvm \
+		-DMLIR_INCLUDE_DIRS=$(LLVM_SRC_DIR)/mlir/include \
```
paul0403, line 88: Can you explain a bit why the `LLVM_DIR` here is just removed instead of renamed like in the other places?

joeycarter: Ah, I'm glad you asked this, because removing it is incorrect. In the normal Catalyst build, this variable is unused, so I removed it:

```
CMake Warning:
  Manually-specified variables were not used by the project:

    LLVM_DIR
```

But I see now that it was added for Backline work, and that the `executor` lib contains

```cmake
find_package(LLVM CONFIG REQUIRED)
```

so it needs `-DLLVM_DIR=...`

Let me fix this...

joeycarter: I added them back in: 86346f1be

I'd rather have an unused variable warning right now over complex cmake config flags conditional on the `ENABLE_EXECUTOR` variable.

paul0403 (approved):

paul0403 (changes requested):
Actually, don't we need to change this in the wheels scripts too?

joeycarter:
> Actually, don't we need to change this in the wheels scripts too?

No the build-wheels workflow configs use `LLVM_DIR` correctly already.

paul0403 (approved):

mehrdad2m (approved):

PR #3156, opened 2026-08-26
Support Qubits Crossing SCF Boundaries in Decompose-Lowering

kipawaa:
**Context:**
When inlining register-mode decomposition rules for gates in for-loop regions that use qubits extracted outside the region, decompose-lowering fails to locate the register, preventing the inlining.

Example:
```python
@qp.qjit(capture=True)
@catalyst.passes.graph_decomposition(gate_set=gate_set | {"GlobalPhase": 0})
@qp.qnode(null_qubit)
def circuit():
    """Documentation for a function.

    Extended documentation.
    """

    @qp.for_loop(0, 3)
    def loop(i):
        qp.X(0)
        qp.Y(1)
        qp.Z(2)
        qp.RZ(0.1, wires=0)
        qp.S(1)
        qp.T(2)
        qp.H(0)
        qp.SX(1)

    loop()

    return qp.expval(qp.X(0))


circuit()
```

Adds support for for-loop blocks to the `QubitIndex` walkback, allowing register locating through for-loop regions. Also adds tests for if and while operations, which pass without issue.

**Benefits:**
Improved reliability in `decompose-lowering` and `graph-decomposition`

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-128395]

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -113,6 +113,7 @@
     [(#2973)](https://github.com/PennyLaneAI/catalyst/pull/2973)
     [(#2836)](https://github.com/PennyLaneAI/catalyst/pull/2836)
     [(#2855)](https://github.com/PennyLaneAI/catalyst/pull/2855)
+    [(#3156)](https://github.com/PennyLaneAI/catalyst/pull/3156)
 
     1. The pass now supports applying a selection of the available decomposition rules via the `target_rules` parameter.
 
@@ -126,6 +127,8 @@
 
     5. The pass can now handle null decomposition rules, which are rule functions that do not have any quantum values as arguments or results.
     Gates with null decomposition rules are simply removed.
+
+    6. The pass can now handle register-mode rules that target gates in control flow regions whose qubits were extracted outside the region.
 
 * A failure during AOT compilation is now downgraded to a warning and logged.
   [(#3100)](https://github.com/PennyLaneAI/catalyst/pull/3100)
--- a/mlir/include/Quantum/Utils/QubitIndex.h
+++ b/mlir/include/Quantum/Utils/QubitIndex.h
@@ -84,6 +84,17 @@ class QubitIndex {
     }
 };
 
+template <typename OperandRangeT, typename ResultRangeT>
+static mlir::Value getMappedQubitOperand(mlir::Value qubit, const OperandRangeT qubitOperands,
+                                         const ResultRangeT qubitResults) {
+    auto it = llvm::find_if(qubitResults, [&](mlir::Value result) { return result == qubit; });
+
+    assert(it != qubitResults.end());
+    size_t resultIndex = std::distance(qubitResults.begin(), it);
+    assert(resultIndex < qubitOperands.size());
+    return qubitOperands[resultIndex];
+}
+
 /// Trace a qubit SSA value back to the quantum.extract that produced it and
 /// return the corresponding QubitIndex (with the originating qreg).
 ///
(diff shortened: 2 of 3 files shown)


mlir/test/Quantum/DecomposeLoweringTest.mlir
```diff
@@ -814,3 +814,120 @@ module @different_qreg_values{
...
+      // CHECK: "PhaseShift"
+      %out = quantum.custom "T"() %in: !quantum.bit
+      %increment = arith.constant 1 : index
+      %updated_index = index.add %before_index, %increment
+      %condition = index.cmp ult (%updated_index, %limit) 
+      scf.condition(%condition) %updated_index, %out: index, !quantum.bit
+    } do {
+      ^bb0(%after_index: index, %after_qubit : !quantum.bit):
+        scf.yield %after_index, %after_qubit : index, !quantum.bit
+    }
```
paul0403, line 928: Don't we need to do this for while loops as well?

Here the test is only passing because the gate is using `%in` from outside the region directly. The walk would be similarly obfuscated if the gate were using the block argument `%before_qubit` (and similarly in the second block, if there were a gate that uses `%after_qubit`.

doc/releases/changelog-dev.md
```diff
@@ -127,6 +128,8 @@
     5. The pass can now handle null decomposition rules, which are rule functions that do not have any quantum values as arguments or results.
     Gates with null decomposition rules are simply removed.
 
+    6. The pass can now handle register-mode rules that target gates in control flow regions whose qubits were extracted outside the region.
```
maliasadi, line 143:
```suggestion
    6. The pass can now handle register-mode rules that target gates inside an `scf.for` body when the gate acts on a loop-carried qubit whose `quantum.extract` is outside the loop.
```

mlir/include/Quantum/Utils/QubitIndex.h
```diff
@@ -119,6 +123,14 @@ inline QubitIndex getExtractIndex(mlir::Value qubit) {
                        mlir::dyn_cast_or_null<quantum::MeasureOp>(qubit.getDefiningOp())) {
             qubit = measureOp.getInQubit();
             continue;
+        } else if (auto blockArg = mlir::dyn_cast_or_null<mlir::BlockArgument>(qubit)) {
```
maliasadi, line 138: What's the motivation behind this support? I just want to know if this's resolving any known issues in workflows? What's the plan to add support for `scf.while` and `scf.if`?

maliasadi (commented):
Thanks @kipawaa!

mlir/test/Quantum/DecomposeLoweringTest.mlir
kipawaa: 🤦 yep, my bad! Will fix this shortly

mlir/include/Quantum/Utils/QubitIndex.h
paul0403: This was discovered by Isaac trying to run (a minimal example of) one of the workflows.

I commented above about the need for while loops. 

If statements do not need special handling because `scf.if` has no block arguments and use all values from outside via clousure.

mlir/test/Quantum/DecomposeLoweringTest.mlir
kipawaa: [8faac3d](https://github.com/PennyLaneAI/catalyst/pull/3156/commits/8faac3ddabc830029da04e64a52e585b9824f8ad)

doc/releases/changelog-dev.md
kipawaa: As @paul0403 mentioned, it has to support while loops more explicitly as well (I foolishly wrote a bad test for this, fixed in [8faac3d](https://github.com/PennyLaneAI/catalyst/pull/3156/commits/8faac3ddabc830029da04e64a52e585b9824f8ad)). I think we should leave the changelog more generic about scf

mlir/include/Quantum/Utils/QubitIndex.h
```diff
@@ -119,6 +123,18 @@ inline QubitIndex getExtractIndex(mlir::Value qubit) {
...
+            if (auto forOp =
+                    mlir::dyn_cast_or_null<mlir::scf::ForOp>(blockArg.getOwner()->getParentOp())) {
+                // ForOp Regions (and blocks) have args (iter_arg, *init_args), so the index is
+                // off-by-one
+                qubit = forOp.getInitArgs()[blockArg.getArgNumber() - 1];
+                continue;
+            } else if (auto whileOp = mlir::dyn_cast_or_null<mlir::scf::WhileOp>(
+                           blockArg.getOwner()->getParentOp())) {
+                qubit = whileOp.getInits()[blockArg.getArgNumber()];
+                continue;
```
paul0403, line 154: The gate could be in either the first block or the second block. We need to have separate handling for them.

kipawaa: The handling for this is no different since the indices must match, but I've added another gate to the test to demonstrate this: [fa1f1dc](https://github.com/PennyLaneAI/catalyst/pull/3156/commits/fa1f1dc0b4ac5c66c7324e62ecdcc385fb564942)

mlir/test/Quantum/DecomposeLoweringTest.mlir
```diff
@@ -814,3 +814,121 @@ module @different_qreg_values{
...
+    %end = arith.constant 10 : index
+    %step = arith.constant 1 : index
+
+    %rout = scf.for %iter = %start to %end step %step iter_args(%for_in = %in) -> (!quantum.bit) {
+      // CHECK-NOT: "T"
+      // CHECK: "PhaseShift"
+      %out = quantum.custom "T"() %for_in : !quantum.bit
+      scf.yield %out : !quantum.bit
+    }
+
```
paul0403, line 881: Just realized that we also need to link up the walk from after the scf regions, i.e. 

```mlir
    %in = quamtu.custom "S" %0
    %rout = scf.for %iter = %start to %end step %step iter_args(%for_in = %in) -> (!quantum.bit) {
      ....
      scf.yield %out : !quantum.bit
    }

    %1 = quantum.custom "H" %rout
```

When `qubit` walks to `%rout`, it currently cannot walk to `%in`.

paul0403 (approved):

mlir/test/Quantum/DecomposeLoweringTest.mlir
```diff
@@ -814,3 +814,135 @@ module @different_qreg_values{
     return %7 : !quantum.reg
   }
 }
+
+// -----
+
+// CHECK-LABEL module @test_if
```
rniczh:
```suggestion
// CHECK-LABEL: module @test_if
```

rniczh (approved):
Thanks!

PR #3158, opened 2026-08-27
MLIR-Style GOID Format

kipawaa:
**Context:**
GOIDs currently use python-style type annotations, which cause confusion with nested lists in dynamic args ( ex. `[[[f64,f64],f64]]`. Worse yet, there is a collision with `tensor<f64> -> [[f64]] <- tensor<1xf64>`.

**Description of the Change:**
To fix these issues, we migrate GOID to use mlir-style type annotations (ex. `tensor<f64>` as opposed to `[f64]`).

**Benefits:**
Solves the collision issue, improves readability for debugging (GOIDs should still be considered opaque for all implementation purposes), and simplifies lit tests.

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-129176]

PR #3127, opened 2026-08-17
Add C(Op) to graphOpId

maliasadi:
**Context:**
This PR enabled C(Op) and nested Adjoint & Controlled `graphOpId` in `graph-decomposition` + the solver.

**Description of the Change:**
- `defaultGetGraphOpId` now folds control operands after adjoint. Following the resource analysis pass, I implemented it in a control-outermost way, so `C(Adjoint(Op))` and `Adjoint(C(Op))` use one id.
- Updated `parseOperator` in the graph-decomposition pass to follow the new id convention. 

**Related GitHub Issues:**
[sc-125404]

PR #3161, opened 2026-08-28
Decomp/qref decomposable interface

yde773786:
**Context:**

The `DecomposableGate` Interface was introduced to provide a generic interface for decomposition of quantum gates. We want to migrate the `graph-decomposition` and `decompose-lowering` passes to reference semantics, but this will require the `DecomposableGate` interface be implemented for reference semantics gates

The `DecomposableGate` Interface is added for the `Qref` dialect, and the decomposable gates inherit this interface appropriately. The format used is identical to how the interface is defined and used in the `Quantum` dialect. 

`gtest` for this interface addition added.

Using MLIR types in the GOID as defined [here](https://github.com/PennyLaneAI/catalyst/pull/3158) is used. 
A corresponding test that should fail with the previous (python style types in GOID) is also added to the `Qref` dialect's `DecomposableGate` Interface `gtests`.

**Benefits:**

Unblocks implementation of  `graph-decomposition` and `decompose-lowering` passes in reference semantics.

**Possible Drawbacks:**

Code duplication exists with the `Quantum` dialect's implementation of `DecomposableGate`. Modifications need to be applied in both places.

**Related GitHub Issues:**

[[sc-128397](https://app.shortcut.com/xanaduai/story/128397/add-decomposablegate-interface-to-qref-dialect)]


dime10:
> We want to migrate the graph-decomposition and decompose-lowering passes to reference semantics

@kipawaa Could you provide more info on this?

paul0403:
> > We want to migrate the graph-decomposition and decompose-lowering passes to reference semantics
> 
> @kipawaa Could you provide more info on this?

There's at least two points in the current graph decomp pass that's very inefficient. 

1. It is performing the infamous "value semantics walk back" to determine what qubits have what extract indices when inlining the rules back into the actual program's gates.
2. The rules are littered with stablehlo ops everywhere. It is because in the rules' arguments, we are manually taking in extract indices packed in a tensor argument, and then in the rules' bodies `stablehlo.extract` these qubit extract indices and `quantum.extract`. [Linking the whole file: a bit hard to isolate a single line to link to](https://github.com/PennyLaneAI/catalyst/blob/main/mlir/lib/Quantum/Transforms/GraphDecomposition/DecomposeLoweringImpl.hpp)

Issue 1 I'm very confident is non existent if the pass were in reference semantics. So at least there's that. We have too many separate implementations of the walkback, and each of these must be maintained separately (e.g. bug found in #3156 ). Any pass that is walking back should just be in reference semantics since the "walkback" in the semantics conversion pass is the most tested!

Issue 2 might be more tricky. The rules are generated from the frontend as quantum subroutines, which takes in wires as (a tensor of) integers (aka "extract indices). I would really love it if the frontend can generate rules that act on `qref.bit` directly instead of extract indices. That is its own problem; but in a world where we have support for that in the frontend, the decompose lowering pass would need to be in reference semantics to enjoy the benefit. The inlining of the rules back into the user circuit would not need to worry about any qubit dataflow: just grab the `qref.bit` SSA value on the user program's gate, replace some uses and we're done! This might also be possible in value semantics but it will be more involved.

kipawaa:
> > We want to migrate the graph-decomposition and decompose-lowering passes to reference semantics
> 
> @kipawaa Could you provide more info on this?

@paul0403 Made the major points [above](https://github.com/PennyLaneAI/catalyst/pull/3161#issuecomment-5454865091), but to add to this is that the on-demand rules (and truly all rules) are initially lowered in reference semantics and then have to be converted to value semantics for compatibility with the passes. I think it would be nice to migrate the entire graph-decomps system to reference semantics for all of these reasons. We could then place the rules in their own module separate from the user circuit, and this way they can be isolated from the user program and remain in reference semantics throughout their lifetime.

kipawaa:
Nice work @yde773786! Sorry to spring this on you right after you finished the work, but could you update to accommodate controls that were introduced in #3127?

yde773786:
@kipawaa 

Added the tests. I found no need to make any changes, so ported identical functionality/tests from `quantum` to `qref` for the above.

Did want to bring attention to [this test](https://github.com/PennyLaneAI/catalyst/blob/decomp/qref-decomposable-interface/mlir/unittests/QRef/TestDecomposableGateInterface.cpp#L422C1-L451C2) again, though (since it's a test that is not present in the `quantum` dialect). It's for testing the N types to 1 GOID bug that [3158](https://github.com/PennyLaneAI/catalyst/pull/3158) fixes and could be added to the `Quantum` dialect as well.

mlir/unittests/QRef/TestDecomposableGateInterface.cpp
```diff
@@ -0,0 +1,556 @@
...
+    llvm::StringMap<llvm::SmallVector<mlir::Type>> expectedDynamicShape = {
+        {"phi", {mlir::Float64Type::get(&context)}}};
+    ASSERT_EQ(gphase.getDynamicShape(), expectedDynamicShape);
+
+    llvm::StringMap<size_t> expectedWires = {};
+    ASSERT_EQ(gphase.getWireLens(), expectedWires);
+
+    ASSERT_EQ(gphase.getStaticData().size(), 0);
+
+    // Controlled unitary: the control wire is folded into the id (control-outermost).
```
kipawaa, line 325:
```suggestion
```

mlir/unittests/QRef/TestDecomposableGateInterface.cpp
```diff
@@ -0,0 +1,556 @@
...
+
+    DecomposableGate op;
+    module->walk([&](OperatorOp walkOp) { op = walkOp; });
+
+    ASSERT_EQ(op.getOperatorName(), "testOperatorUID");
+
+    // This is needed to keep the backing array from being deleted
+    // llvm::SmallVector<llvm::SmallVector<mlir::Type>, 1> backing(
+    //     {mlir::IntegerType::get(&context, 1), mlir::Float64Type::get(&context),
+    //      mlir::IntegerType::get(&context, 64)});
```
kipawaa, line 542:
```suggestion
```

kipawaa (approved):
Great work, thanks @yde773786!

mlir/lib/QRef/IR/QRefInterfaces.cpp
```diff
@@ -14,11 +14,153 @@
...
 
+#include "QRef/IR/QRefInterfaces.cpp.inc"
+
+//===----------------------------------------------------------------------===//
+// Helpers
+//===----------------------------------------------------------------------===//
+
+namespace {
+
+void printAttr(mlir::Attribute attr, llvm::raw_string_ostream &ss) {
```
dime10, line 42: Just wanted to point out that this printer should be unnecessary, and I'm removing it in this PR: https://github.com/PennyLaneAI/catalyst/pull/3188/changes/67e6bf869decc4d0ee6d73472f7ee9f551ac5b6a#diff-020c06e34c09643c0f3a337d825e94edfc387efafe3eb64d705a89e50bb941eaL40

I'm not sure if changing this proactively in this PR is possible without breaking anything, but if so we should probably do it so we don't forget to later.

mlir/lib/QRef/IR/QRefInterfaces.cpp
```diff
@@ -14,11 +14,153 @@
...
+            ss << "]";
+        })
+        .Case<mlir::StringAttr>([&](mlir::StringAttr attr) { ss << attr.str(); })
+        .Case<mlir::IntegerAttr>([&](mlir::IntegerAttr attr) { ss << attr.getInt(); })
+        .Case<mlir::FloatAttr>([&](mlir::FloatAttr attr) { ss << attr.getValueAsDouble(); })
+        .Default([&](mlir::Attribute attr) { attr.print(ss); });
+}
+
+template <typename T, typename PrintFunc>
+void printSortedMap(const llvm::StringMap<T> &map, llvm::raw_string_ostream &ss,
```
dime10, line 73: `wrapModifiers`, `printSortedMap`, `printDynamicShape` and `printWireLens` are all copies that are not specific to the either dialect, and should be factored out

doc/releases/changelog-dev.md
```diff
@@ -92,6 +92,7 @@
 
     This allows arbitrary operations implementing the interface to be registered to and decomposed by the graph.
     This also allows the use of python-decompositions for any operator pre-registered in the frontend graph.
+    - Added the interface + implementations for decomposable operators in `QRef`. [(#3161)](https://github.com/PennyLaneAI/catalyst/pull/3161)
```
maliasadi: could you add the PR link to line 78-79?

maliasadi, line 95:
```suggestion
- Added the `DecomposableGate` op interface to allow generic handling of operations in the `graph-decomposition` pass.
[(#2983)](https://github.com/PennyLaneAI/catalyst/pull/2983)
[(#3022)](https://github.com/PennyLaneAI/catalyst/pull/3022)
[(#3161)](https://github.com/PennyLaneAI/catalyst/pull/3161)

This allows arbitrary operations implementing the interface to be registered to and decomposed by the graph.
This also allows the use of python-decompositions for any operator pre-registered in the frontend graph.
```

mlir/lib/QRef/IR/QRefInterfaces.cpp
maliasadi, line 1: `frontend/catalyst/decomposition/graph_op_id.py` documents the sync obligation in two places, both
of which now name or imply a single C++ impl! Could you also update the docstring in the frontend? We need to update all references to name `QRefInterfaces.cpp`.

mlir/lib/QRef/IR/QRefInterfaces.cpp
```diff
@@ -14,11 +14,153 @@
...
+    if (gate.getExtraData() != "") {
+        ss << '[' << gate.getExtraData() << ']';
+    }
+    ss.flush();
+
+    return out;
+}
+
+} // namespace qref
+} // namespace catalyst
```
maliasadi, line 166: I'm thinking if it makes sense to move the impl to `mlir/lib/Catalyst/Analysis/` which both Qref and Quantum dialects already depend on. This could also let us to simplify the C++ unittests

mlir/unittests/QRef/TestDecomposableGateInterface.cpp
```diff
@@ -0,0 +1,552 @@
...
+        {"index", {mlir::IntegerType::get(&context, 64)}}};
+    ASSERT_EQ(op.getDynamicShape(), expectedDynamicShape);
+
+    llvm::StringMap<size_t> expectedWires = {{"reg", 3}};
+    ASSERT_EQ(op.getWireLens(), expectedWires);
+
+    ASSERT_EQ(op.getStaticData(), mlir::DictionaryAttr::get(&context, {}));
+
+    ASSERT_EQ(op.getGraphOpId(),
+              "testOperatorUID{angle:[f64],flag:[i1],index:[i64]}{reg:3}{}[248]");
```
maliasadi, line 551: aren't we expecting `testOperatorUID{angle:[tensor<f64>],index:[tensor<i1>,tensor<i64>]}{reg:3}{}[248]`?

mlir/unittests/QRef/TestDecomposableGateInterface.cpp
maliasadi, line 1: Could you add tests for 
- the `llvm::SmallVector<mlir::Type>` loop in `OperatorOp::getDynamicShape`
- the joining loop in `printDynamicShape` that results `[a,b,...]`.

paul0403:
I am branching off this PR to migrate `--decompose-lowering` to reference semantics.

mlir/unittests/QRef/TestDecomposableGateInterface.cpp
yde773786: [dcac902](https://github.com/PennyLaneAI/catalyst/pull/3161/commits/dcac9028f483f386a8449d9d6f64463da2c2b096)

This test should cover both, the testing of `getDynamicShape` and `printDynamicShape`

mlir/unittests/QRef/TestDecomposableGateInterface.cpp
yde773786: Since the test is written such that the `qref.operator` is applied as such:

`qref.operator "testOperatorUID"(%flag: i1, %angle: f64, %index: i64)...`

the corresponding `printDynamicShape` does seem to be appropriately inserting those types in the `GOID` unless I'm mistaken

mlir/lib/QRef/IR/QRefInterfaces.cpp
yde773786: Discussing with @kipawaa , it would make sense to address this proposal in a new issue.

Let me know if I have captured the ask correctly [here](https://github.com/PennyLaneAI/catalyst/issues/3223).

mlir/lib/QRef/IR/QRefInterfaces.cpp
yde773786: [b588908](https://github.com/PennyLaneAI/catalyst/pull/3161/commits/b58890804ef202ab4e714dc5bd865d2fff8ab339)

mlir/lib/QRef/IR/QRefInterfaces.cpp
yde773786: I think [this](https://github.com/PennyLaneAI/catalyst/issues/3223) applies here as well.

mlir/lib/QRef/IR/QRefInterfaces.cpp
yde773786: [1427346](https://github.com/PennyLaneAI/catalyst/pull/3161/commits/142734655d7c9f74eaa56b00ed295164c34d0aac)

doc/releases/changelog-dev.md
```diff
@@ -148,6 +148,13 @@
 
     This allows arbitrary operations implementing the interface to be registered to and decomposed by the graph.
     This also allows the use of python-decompositions for any operator pre-registered in the frontend graph.
+- Added the `DecomposableGate` op interface to allow generic handling of operations in the `graph-decomposition` pass.
+[(#2983)](https://github.com/PennyLaneAI/catalyst/pull/2983)
+[(#3022)](https://github.com/PennyLaneAI/catalyst/pull/3022)
+[(#3161)](https://github.com/PennyLaneAI/catalyst/pull/3161)
+
+This allows arbitrary operations implementing the interface to be registered to and decomposed by the graph.
+This also allows the use of python-decompositions for any operator pre-registered in the frontend graph.
```
```suggestion
  - Added the `DecomposableGate` op interface to allow generic handling of operations in the `graph-decomposition` pass.
    [(#3161)](https://github.com/PennyLaneAI/catalyst/pull/3161)

    This allows arbitrary operations implementing the interface to be registered to and decomposed by the graph.
    This also allows the use of python-decompositions for any operator pre-registered in the frontend graph.
```

maliasadi (approved):

maliasadi (approved):
Looks good! Thanks @yde773786

doc/releases/changelog-dev.md
maliasadi: Please double-check the rendered version of the changelog after accepting the suggestion and before merging the PR

yde773786: I believe it's fine now. Issue above resulted from me trying to merge changes directly from web but I think I'm doing something wrong there..

mlir/lib/QRef/IR/QRefInterfaces.cpp
paul0403: Since it's tracked, let's tackle it in a follow up PR.

paul0403 (approved):

mlir/lib/QRef/IR/QRefInterfaces.cpp
dime10: I don't think an interface is necessary here, this is just about factoring out duplicate helper functions. But please do tackle this in a follow up!

paul0403: I think the idea is to put the interface itself in the `Catalyst` dialect, which means all these helpers would exist just once in `Catalyst` dialect too.

paul0403: Although I don't necessarily agree with putting quantum things in `Catalyst`, I can't really think of a better place :sweat_smile:

dime10: Yeah Catalyst doesn't sound like the right place for a quantum related interface, it should just go into quantum. Regarding the helpers, if they are moved to the interface that's fine too, I'm just saying all that was necessary here was moving them to a header from which they can be imported into both dialects.

PR #3162, opened 2026-08-28
Update `operator_p` lowering to pop work wire related arguments

mudit2812:
**Context:**
`operator_p` now has two new parameters to provide information related to the work wires of controlled operators. This caused the operator lowering for controlled operators to fail, as they were incorrectly being identified as static arguments to the underlying operator.

* Pop `n_ctrl_work_wires` and `ctrl_work_wires_type` from the keyword arguments in `from_plxpr`. They're not used by the MLIR operation, so they don't need to be used, just ignored. So the lowering rule never receives them.

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-128897]

### DIFF
--- a/.dep-versions
+++ b/.dep-versions
@@ -8,7 +8,7 @@ enzyme=v0.0.238
 
 # For a custom PL version, update the package version here and at
 # 'doc/requirements.txt'
-pennylane=0.46.0.dev84
+pennylane=0.46.0.dev86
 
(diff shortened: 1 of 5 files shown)


paul0403 (approved):

kipawaa (approved):

Qottmann:
how related is this? https://github.com/PennyLaneAI/pennylane/pull/10069

PR #3165, opened 2026-09-01
Categorize `DecompRuleWarnings`

kipawaa:
**Context:**
Recent changes in the graph decomposition system result in attempting to lower all rules for all decomposition-reachable operators for the user circuit. When these rules fail to lower for any reason, we catch the exception and raise a warning, since one rule failure may not prevent compilation. These warnings are often caused by un-migrated PennyLane `Operator`s (need to be upgraded to `Operator2`) or capture/compilation-incompatible decomposition rules or resource functions.

Adds the `RuleLoweringWarning` warning category to allow silencing these warnings with 
```python
from catalyst.decomposition import RuleLoweringWarning
warnings.filterwarnings("ignore", category=RuleLoweringWarning)
```

**Benefits:**
Users can ignore decomposition rule warnings.

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-129420]

### DIFF
--- a/frontend/catalyst/decomposition/rule_lowering_warning.py
+++ b/frontend/catalyst/decomposition/rule_lowering_warning.py
@@ -1,0 +1,21 @@
+# Copyright 2026 Xanadu Quantum Technologies Inc.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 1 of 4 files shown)


jzaia18 (approved):
LGTM, just don't forget the changelog

kipawaa:
> LGTM, just don't forget the changelog

@jzaia18 I'm not convinced this needs a changelog, ideally it will be temporary (just needed until the migration is complete)

paul0403 (approved):

frontend/catalyst/decomposition/rule_lowering_warning.py
```diff
@@ -0,0 +1,21 @@
...
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""This module provides a warning category for grouping decomposition rule lowering warnings."""
+
+import warnings
```
jzaia18, line 17:
```suggestion
```

Unused. You can also do `from warnings import Warning` for if you want to avoid the top-level Python warning

PR #3169, opened 2026-09-01
Support qp.BasisState in graph decomp

paul0403:
**Context:**
Currently, `qp.BasisState` has a special lowering to `quantum.set_basis_state` op.

However, the `quantum::SetBasisStateOp` cannot participate in graph decomposition. This is because the `DecomposableGateInterface` inherits from the `QuantumGate` interface (which is for a unitary gate, i.e. a gate that can be adjointed or controlled), but the `SetBasisStateOp` is a `QuantumOperation`, i.e. not adj/ctrl-able. 

Remove the special lowering. Just lower `qp.BasisState` to a regular `OperatorOp`.

In addition, since `OperatorOp` is not bufferized and not runtime-lowerable, it must not survive the pipeline.
Ideally we just run graph decomp on `quantum.operator "BasisState"` to "implement" it, but we don't have graph decomp in the default pipeline yet as we do not yet have a "standard" device gateset. 

So we just lower `quantum.operator "BasisState"` to `quantum.set_basis_state` in a new pass, completely separate from the graph decomp system, to support its execution. 

**Benefits:**
`qp.BasisState` can participate in graph decomp if a user wants to, while remaining executable in a workflow without any decompositions.

**Related GitHub Issues:**
[sc-128297]

### DIFF
--- a/.dep-versions
+++ b/.dep-versions
@@ -8,7 +8,7 @@ enzyme=v0.0.238
 
 # For a custom PL version, update the package version here and at
 # 'doc/requirements.txt'
-pennylane=0.46.0.dev86
+pennylane=0.46.0.dev88
 
 # For a custom LQ/LK version, update the package version here and at
 # 'doc/requirements.txt'
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -125,6 +125,7 @@
     [(#3061)](https://github.com/PennyLaneAI/catalyst/pull/3061)
     [(#3160)](https://github.com/PennyLaneAI/catalyst/pull/3160)
     [(#3149)](https://github.com/PennyLaneAI/catalyst/pull/3149)
+    [(#3169)](https://github.com/PennyLaneAI/catalyst/pull/3169)
 
     This pathway of rule injection can be opted-out via a new keyword argument on `qp.qjit` named `collect_decomp_rules`.
     This kwarg controls whether or not to compile the decomposition rules during lower-time. Default value is `True`.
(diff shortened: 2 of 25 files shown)


dime10 (approved):
Thanks Paul, looks good! One question, we have two state prep operations right, are you doing the other one as well?

paul0403:
> Thanks Paul, looks good! One question, we have two state prep operations right, are you doing the other one as well?

You mean `quantum.set_state`? I don't know if the frontend one has decomp rules yet but I can add them to the pass. It should be very straightforward. Also without special lowering even if `qp.StatePrep` did have decomp rules it would just go to an usual operator op, so adding the pattern to the pass should be compatible with the current system that's in place for `qp.BasisState`.

dime10:
> > Thanks Paul, looks good! One question, we have two state prep operations right, are you doing the other one as well?
> 
> You mean `quantum.set_state`? I don't know if the frontend one has decomp rules yet but I can add them to the pass. It should be very straightforward. Also without special lowering even if `qp.StatePrep` did have decomp rules it would just go to an usual operator op, so adding the pattern to the pass should be compatible with the current system that's in place for `qp.BasisState`.

Yeah that's the one! It's existed for a long time, so it should have a decomp rule. 

Generally these are two ops from the same family, so it would make sense to process them together here.

paul0403:
> > > Thanks Paul, looks good! One question, we have two state prep operations right, are you doing the other one as well?
> > 
> > 
> > You mean `quantum.set_state`? I don't know if the frontend one has decomp rules yet but I can add them to the pass. It should be very straightforward. Also without special lowering even if `qp.StatePrep` did have decomp rules it would just go to an usual operator op, so adding the pattern to the pass should be compatible with the current system that's in place for `qp.BasisState`.
> 
> Yeah that's the one! It's existed for a long time, so it should have a decomp rule.
> 
> Generally these are two ops from the same family, so it would make sense to process them together here.

Added in https://github.com/PennyLaneAI/catalyst/pull/3169/commits/f7900a00f037cc8a7766d3f41c1c3c2f047768f9
Didn't add frontend test because `qp.StatePrep` has not been migrated yet.

maliasadi (approved):

PR #3170, opened 2026-09-01
Decomp/pbc decomposable interface

hsajjad14:
**Context:**
The DecomposableGate interface was introduced to provide a generic interface for decomposition of [quantum gates](https://github.com/PennyLaneAI/catalyst/blob/main/mlir/include/Quantum/IR/QuantumInterfaces.td). We want the graph-decomposition to be able to decompose to the PBC dialect, but this will require the DecomposableGate interface be implemented for [PBC gates](https://github.com/PennyLaneAI/catalyst/blob/8e196b734ac7b67285913f99c0ae7fcd0839b541/mlir/include/PBC/IR/PBCOpInterfaces.td).

This PR extends the `DecomposableGate` interface to the PBC dialect. It uses the same template for how the interface is extended for Quantum and Qref dialects. 

I added the DecomposableGate interface to `PBCOpInterfaces.td` (this does not inherit `QuantumGate` unlike in `QuantumInterface.td`), then extend the two PPR operations in `PBCOps.td`, `PPRotationOp` and `PPRotationArbitraryOp`. Currently only these two operations in the PBC dialect have decomposition rules, `PPMeasurementOp` is `qp.PauliMeasure`, but it has no `add_decomps` in PennyLane yet.

`PBCOpInterfaces.cpp` is slightly different than `QuantumInterfances.cpp`. For `defaultGetGraphOpId` because the PPR operations don't have adjoint in its `arguments` in the `PBCOps.td`, it is not an attribute. Also the operation will never be a QuantumGate, thus there is no need for a `wrapModifiers` function around `getOperatorName()`.
I used the same helpers from `QuantumInterfances.cpp` in `PBCOpInterfaces.cpp`, but from comparing with the QRef those helpers are slightly different, so I held off on moving the helpers into their own file.

In `PBCOps.cpp` I implement two operations `PPRotationOp` and `PPRotationArbitraryOp`. I use the same implementation from `QuantumOps.cpp` for `PauliRotOp` except because DecomposableGate for PBC does not inherit `QuantumGate`, for `getWireLens()` I used `getInQubits().size()` to get wires instead of `getNonCtrlQubitOperands().size()`.

**Benefits:**
Allows PBC operations to be compatible with the `graph-decomposition` pass.

**Possible Drawbacks:**
Code duplication with Quantum and Qref dialect, especially for the helper functions.

**Related GitHub Issues:**
[sc-129100]

### DIFF
--- a/mlir/unittests/PBC/Interfaces/TestDecomposableGateInterface.cpp
+++ b/mlir/unittests/PBC/Interfaces/TestDecomposableGateInterface.cpp
@@ -1,0 +1,111 @@
+// Copyright 2026 Xanadu Quantum Technologies Inc.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+
+//     http://www.apache.org/licenses/LICENSE-2.0
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <cstddef>
+#include <string>
+
+#include "gtest/gtest.h"
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/ADT/StringMap.h"
+#include "llvm/Support/Format.h" // for gtest printing on failure
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/IR/Attributes.h"
+#include "mlir/IR/BuiltinAttributes.h"
+#include "mlir/IR/BuiltinOps.h"
+#include "mlir/IR/BuiltinTypeInterfaces.h"
+#include "mlir/IR/BuiltinTypes.h"
+#include "mlir/IR/DialectRegistry.h"
+#include "mlir/IR/MLIRContext.h"
+#include "mlir/IR/OwningOpRef.h"
+#include "mlir/IR/Types.h"
+#include "mlir/Parser/Parser.h"
+
+#include "PBC/IR/PBCDialect.h"
+#include "PBC/IR/PBCOpInterfaces.h"
+#include "PBC/IR/PBCOps.h"
+#include "Quantum/IR/QuantumDialect.h"
+
+using namespace mlir;
+using namespace catalyst::pbc;
+
+TEST(DecomposableGateInterfaceTests, PPRotationOp) {
+    std::string moduleStr = R"mlir(
+module {
+  %q0 = quantum.alloc_qb : !quantum.bit
+  %q1 = quantum.alloc_qb : !quantum.bit
+  %q2 = quantum.alloc_qb : !quantum.bit
+  %0:3 = pbc.ppr ["X", "Y", "Z"](4) %q0, %q1, %q2 : !quantum.bit, !quantum.bit, !quantum.bit
+}
+    )mlir";
+
+    DialectRegistry registry;
+    registry.insert<mlir::arith::ArithDialect, catalyst::quantum::QuantumDialect, PBCDialect>();
+    MLIRContext context(registry);
+    ParserConfig config(&context, /*verifyAfterParse=*/false);
+    OwningOpRef<ModuleOp> module = parseSourceString<ModuleOp>(moduleStr, config);
+
+    DecomposableGate ppr = *module->getOps<PPRotationOp>().begin();
+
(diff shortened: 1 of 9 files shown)


kipawaa:
Thanks @hsajjad14! 
> PPMeasurementOp is qp.PauliMeasure, but it has no add_decomps in PennyLane yet.

We still need the interface for PPMs because they can be the _result_ of a decomposition, and so need to be registered in the graph. Could you add the interface & tests for this as well?

mlir/unittests/PBC/Interfaces/TestDecomposableGateInterface.cpp
kipawaa, line 1: Could you add tests for controlled/adjoint gates as well?

hsajjad14: For controlled PPR gates, the mlir generated for this 
```
qp.ctrl(qp.PauliRot, control = (0), control_values=(1))(angle, pauli_word="YZ", wires=[1, 2])`
```
 after `to-ppr` is:
```
%16:2 = pbc.ppr.arbitrary ["Y", "Z"](%cst) %15#1, %15#2 : !quantum.bit, !quantum.bit
```
 there is no control attribute for pbc. So that test would be the same as `PPRotationArbitraryOp`.

For adjoint gates, for PBC dialect the adjoint attribute gets lowered out to be a negative angle. I added tests for that in here [6cca181](https://github.com/PennyLaneAI/catalyst/pull/3170/commits/6cca18194209d2fe2d2d9527ebe832eb72c6f6e7)

hsajjad14:
> We still need the interface for PPMs because they can be the _result_ of a decomposition, and so need to be registered in the graph. Could you add the interface & tests for this as well?

I added ppm op to the interface in [2f9a1cb](https://github.com/PennyLaneAI/catalyst/pull/3170/commits/2f9a1cb80e5f75ad2087ad8ce321f50275c27eec). `meas_uid` and `postselect` are not lowered into PPMeasurementOp, so I set them to None manually.

mlir/unittests/PBC/Interfaces/TestDecomposableGateInterface.cpp
kipawaa: Does this hold for 
```mlir
quantum.ctrl {
  ^bb0:
    %out:2 = pbc.ppr.arbitrary ["Y", "Z"](%cst) %15#1, %15#2 : !quantum.bit, !quantum.bit
}
```
?

hsajjad14: Yes, pbc.ppr ops don't get any ctrl attributes.

I confirmed with this test:
```python
def NestedCtrlPPRcircuit():
    angle = 0.5
    qp.ctrl(
        qp.ctrl(qp.PauliRot, control=(0), control_values=(1)),
        control=(3),
        control_values=(1),
    )(angle, pauli_word="YZ", wires=[1, 2])
    return qp.expval(qp.Z(0))
```
goes to 
```
qref.ctrl(%1) ctrlvals(%true) {
        %5 = qref.get %0[ 0] : !qref.reg<4> -> !qref.bit
        qref.ctrl(%5) ctrlvals(%true) {
          %6 = qref.get %0[ 1] : !qref.reg<4> -> !qref.bit
          %7 = qref.get %0[ 2] : !qref.reg<4> -> !qref.bit
          qref.paulirot ["Y", "Z"](%cst) %6, %7 : !qref.bit, !qref.bit
        }
      }
```
then after `to-ppr`:
```
%out_ctrl_qubits, %results:3 = quantum.ctrl(%1) ctrlvals(%true) (%2, %3, %4) : !quantum.bit -> !quantum.bit, !quantum.bit, !quantum.bit {
    ^bb0(%arg0: !quantum.bit, %arg1: !quantum.bit, %arg2: !quantum.bit):
      %out_ctrl_qubits_0, %results_1:2 = quantum.ctrl(%arg2) ctrlvals(%true) (%arg0, %arg1) : !quantum.bit -> !quantum.bit, !quantum.bit {
      ^bb0(%arg3: !quantum.bit, %arg4: !quantum.bit):
        %11:2 = pbc.ppr.arbitrary ["Y", "Z"](%cst) %arg3, %arg4 : !quantum.bit, !quantum.bit
        quantum.yield %11#0, %11#1 : !quantum.bit, !quantum.bit
      }
      quantum.yield %results_1#0, %results_1#1, %out_ctrl_qubits_0 : !quantum.bit, !quantum.bit, !quantum.bit
    }
```
then after control lowering pass, controls are removed, and it's lowered to:
```
%5:2 = pbc.ppr.arbitrary ["Y", "Z"](%cst) %2, %3 : !quantum.bit, !quantum.bit
```

PR #3168, opened 2026-09-01
Fix AttributeError for abstract-wire resource reps in capture decomposition

Qottmann:
## Summary

`AbstractWires` (`pennylane.typing.Wire[n]`) exposes `__len__` but not `.num_wires`, unlike the concrete `Wires` class. `from_plxpr/decompose.py` accessed `.num_wires` directly in two places, crashing whenever a decomposition rule's resource representation used abstract wires.

Fixes #3163.

## Changes

- `_resource_num_wires`: `op_rep.wires.num_wires` → `len(op_rep.wires)`
- `_create_decomposition_rule`: `wire_spec.num_wires` → `len(wire_spec)`

## Test plan

Note: the repro goes on to hit a separate, pre-existing MLIR segfault unrelated to abstract wires, tracked separately in #3167.

Made with [Cursor](https://cursor.com)

PR #3176, opened 2026-09-02
Capture Operator2 (abstract-wire) solution nodes in graph decomposition

### LINKED ISSUE #3175: Graph decomposition under capture raises ValueError for Operator2 (abstract-wire) solution nodes (e.g. SemiAdder)
### Summary

Under `qml.qjit(capture=True)` with graph-based decomposition, if the decomposition-graph solution contains an `Operator2` whose resource representation is built from abstract wires (`pennylane.typing.Wire[n]`), and that operator is neither one of the `COMPILER_OPS_FOR_DECOMPOSITION` nor a skippable symbolic op, `DecompRuleInterpreter.cleanup` (`frontend/catalyst/from_plxpr/decompose.py`) falls into its final `else: raise` branch:

```
ValueError: Could not capture SemiAdder(x_wires=AbstractWires(2), y_wires=AbstractWires(2), work_wires=AbstractWires(1)) without the number of wires.
```

`SemiAdder` (an `Operator2`) is a concrete example: it appears in the solution with an abstract-wire resource rep, but `cleanup` has no branch that materializes its wire count from that rep, so capture aborts.

### Minimal reproducer

```python
import pennylane as qml

@qml.qjit(capture=True, target="mlir")
@qml.transforms.decompose(gate_set={"CNOT", "Toffoli", "X", "Hadamard", "PhaseShift"})
@qml.qnode(qml.device("null.qubit", wires=5))
def circuit():
    qml.SemiAdder(x_wires=[0, 1], y_wires=[2, 3], work_wires=[4])
    return qml.expval(qml.Z(0))

circuit.mlir  # capture fails; the ValueError above is logged and `.mlir` is None
```

(The failure is raised during frontend capture; `aot_compile` logs it "at exception level" and returns, so `circuit.mlir` is `None`.)

### Cause

`cleanup()` iterates the solution's operators and only knows how to build a decomposition rule for `COMPILER_OPS_FOR_DECOMPOSITION` and skippable symbolic ops. An `Operator2` solution node whose args/wires are fully described by its (abstract) resource rep has no handling, even though the wire count is recoverable from the rep (`_resource_num_wires(op.op)` / `len(op.op.wires)`).

### Environment

- catalyst `0.16.0-dev` (`main` + #3168)
- pennylane `0.46.0-dev88`
- jax `0.7.1`

Qottmann:
**Context:**
Closes #3175. Stacked on top of #3168 (this fix calls `_resource_num_wires`, which relies on the abstract-wire `len()` fix from #3168). The PR is based on the #3168 branch and will retarget to `main` once #3168 merges.

Under `qml.qjit(capture=True)` with graph-based decomposition, when the decomposition-graph solution contained an `Operator2` (such as `SemiAdder`) whose wire and parameter arguments are described by an *abstract* resource representation, and that operator was neither in `COMPILER_OPS_FOR_DECOMPOSITION` nor a skippable symbolic op, `DecompRuleInterpreter.cleanup` fell into its final `else` branch and raised:

```
ValueError: Could not capture SemiAdder(...) without the number of wires.
```

This PR:

- Adds an `Operator2` branch to `cleanup` that materializes the wire count from the abstract resource rep (`_resource_num_wires`) and builds the decomposition rule directly, instead of aborting capture.
- Teaches `_create_decomposition_rule` to fall back to the default value for rule-internal *optional* parameters that are not part of the operator's arguments (e.g. `carry_flip` in the `SemiAdder` rule), so the captured rule matches the standard decomposition.

**Benefits:**
`Operator2` templates like `SemiAdder` can now be captured and decomposed under program capture.

**Possible Drawbacks:**
None anticipated.

**Related GitHub Issues:**
Closes #3175. Depends on #3168.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -505,6 +505,13 @@
 
 <h3>Bug fixes 🐛</h3>
 
+* Fixed a `ValueError` ("Could not capture ... without the number of wires") in the `capture=True`
+  graph-decomposition path when the decomposition-graph solution contains an `Operator2` (such as
+  `SemiAdder`) whose wire and parameter arguments are described by an abstract resource
+  representation. The wire count is now materialized from the abstract rep, and rule-internal
+  optional parameters (e.g. `carry_flip`) fall back to their defaults.
+  [(#3176)](https://github.com/PennyLaneAI/catalyst/pull/3176)
+
 * Fixed an `AttributeError` in the `capture=True` graph-decomposition path when a decomposition
   rule's resource representation uses abstract wires (`pennylane.typing.Wire[n]`), e.g. for
   `register_resources` rules that describe cost without concrete wires.
--- a/frontend/catalyst/from_plxpr/decompose.py
+++ b/frontend/catalyst/from_plxpr/decompose.py
@@ -250,6 +250,21 @@ class DecompRuleInterpreter(qp.capture.PlxprInterpreter):
                 # We use MLIR AdjointOp and ControlledOp primitives
                 # to deal with decomposition of symbolic operations at PLxPR.
                 continue
+            elif isinstance(op.op, qp.core.Operator2):
+                # The operator appears in the decomposition-graph solution with an abstract
+                # resource representation (an ``Operator2`` such as ``SemiAdder``), but it is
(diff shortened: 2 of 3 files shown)


Qottmann:
didnt realize this is fixing old things that are not maintained anymore

PR #3181, opened 2026-09-02
[Jukebox] opt out of decomposition-rule collection in non-decomposition frontend tests

paul0403:
Every qjit trace under program capture collects and lowers all decomposition rules reachable from every operator in the circuit. Those rules are dead code unless the -graph-decomposition pass is applied, but they still cost roughly 0.3 s per operator at trace time, which the frontend test suite pays thousands of times over.

Pass collect_decomp_rules=False in tests that do not exercise rule collection:

- every qjit(..., capture=True) call site in non-decomposition test files
- qjit call sites inside tests that enable capture via a fixture or an explicit qp.capture.enable()

Tests that cover rule collection and lowering are deliberately untouched: frontend/test/lit/test_operator2/**, lit/test_from_plxpr.py, pytest/test_decomposition.py, pytest/test_operator.py, pytest/from_plxpr/test_from_plxpr.py, pytest/test_graph_decomposition_pass.py and pytest/test_builtin_passes.py.

Measured on the touched files, same pass/fail set before and after:
- pytest (29 files, -n 4): 360 s -> 182 s
- lit (12 files, serial):   93 s ->  30 s

-------------------
Disclaimer: this PR and the above description are generated by Jukebox via the following prompt:
```
Can you add as many `collect_decomp_rules=False` as possible to the catalyst frontend test suite? Especially the tests that have nothing to do with decomp. This will alleviate CI pressure significantly.
```

But I'll massage it.

### DIFF
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -748,14 +748,6 @@
   `-stage` naming convention used when invoking them from the command line (e.g. `quantum-compilation-stage`).
   [#3002](https://github.com/PennyLaneAI/catalyst/pull/3002)
 
-* Frontend tests that do not exercise graph-based decomposition now pass
-  `collect_decomp_rules=False` to `qjit`. Collecting and lowering every decomposition rule
-  reachable from every gate in a circuit costs roughly 0.3 s per operator at trace time, and the
-  rules are dead code unless the `graph-decomposition` pass is applied, so this removes a large
-  amount of redundant work from CI. Tests that specifically cover rule collection and lowering
-  are unchanged.
-  [(#XXXX)](https://github.com/PennyLaneAI/catalyst/pull/XXXX)
-
 <h3>Documentation 📝</h3>
 
 * A broken link was removed in the [Compiler Core](https://docs.pennylane.ai/projects/catalyst/en/stable/modules/mlir.html) documentation page. The link referred to where precompiled decomposition rules were implemented, which has since been refactored.
--- a/doc/releases/changelog-dev.md
+++ b/doc/releases/changelog-dev.md
@@ -748,6 +748,14 @@
   `-stage` naming convention used when invoking them from the command line (e.g. `quantum-compilation-stage`).
   [#3002](https://github.com/PennyLaneAI/catalyst/pull/3002)
 
+* Frontend tests that do not exercise graph-based decomposition now pass
+  `collect_decomp_rules=False` to `qjit`. Collecting and lowering every decomposition rule
+  reachable from every gate in a circuit costs roughly 0.3 s per operator at trace time, and the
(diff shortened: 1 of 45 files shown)


frontend/test/pytest/test_capture_kwarg.py
```diff
@@ -255,10 +255,10 @@ class TestCaptureCompileOptionsIdentity:
     """
 
     def test_capture_flag_creates_distinct_compile_options(self):
-        """Test that qjit(capture=True) and qjit(capture=False) have distinct CompileOptions."""
+        """Test that qjit(capture=True, collect_decomp_rules=False) and qjit(capture=False) have distinct CompileOptions."""
```
paul0403:
```suggestion
        """Test that qjit(capture=True) and qjit(capture=False) have distinct CompileOptions."""
```

doc/releases/changelog-dev.md
```diff
@@ -748,6 +748,14 @@
...
   [#3002](https://github.com/PennyLaneAI/catalyst/pull/3002)
 
+* Frontend tests that do not exercise graph-based decomposition now pass
+  `collect_decomp_rules=False` to `qjit`. Collecting and lowering every decomposition rule
+  reachable from every gate in a circuit costs roughly 0.3 s per operator at trace time, and the
+  rules are dead code unless the `graph-decomposition` pass is applied, so this removes a large
+  amount of redundant work from CI. Tests that specifically cover rule collection and lowering
+  are unchanged.
+  [(#XXXX)](https://github.com/PennyLaneAI/catalyst/pull/XXXX)
+
```
```suggestion
```

frontend/test/lit/test_quantum_subroutine.py
```diff
@@ -196,7 +196,7 @@ def false_path(): ...
 
         qp.cond(param != 0.0, true_path, false_path)()
 
-    @qp.qjit(autograph=False)
+    @qp.qjit(autograph=False, collect_decomp_rules=False)
```
kipawaa: Do we want to simplify this further while were here? The same could apply to the rest of the tests in this file I think.
```suggestion
    @qp.qjit(autograph=False, collect_decomp_rules=False, target="mlir")
```

kipawaa (approved):
Thanks @paul0403! 🚀

frontend/test/lit/test_quantum_subroutine.py
paul0403: Since some tests in this file have `target="mlir"` and some don't, I assume the intention was to test execution not crashing.

kipawaa: makes sense!

maliasadi (approved):
:shipit:

mehrdad2m (approved):

PR #3182, opened 2026-09-03
Adding dockerfile

runora95:
### Before submitting

**Context:**

**Benefits:**

**Possible Drawbacks:**

**Related GitHub Issues:**

[sc-130938]

### DIFF
--- a/.github/workflows/build-docker.yaml
+++ b/.github/workflows/build-docker.yaml
@@ -1,0 +1,27 @@
+name: Build Pennylane Lightning Catalyst Docker Image
+
+permissions:
+  contents: read
+  
+on:
+    pull_request:
+        
+jobs:
+  build:
+    name: Build Docker Image
+    runs-on:
+      group: 'PL Additional Runners'
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: Dockerfile
+          platforms: linux/amd64,linux/arm64
+          push: false
--- a/.github/workflows/build-docker.yaml
+++ b/.github/workflows/build-docker.yaml
@@ -23,7 +23,7 @@ jobs:
         with:
           context: .
           file: Dockerfile
-          platforms: linux/amd64,linux/arm64
+          platforms: linux/amd64
           push: false
           build-args: |
             PENNYLANE_VERSION=${{ github.event.inputs.pennylane_version }}
--- a/.github/workflows/build-docker.yaml
+++ b/.github/workflows/build-docker.yaml
@@ -25,3 +25,6 @@ jobs:
           file: Dockerfile
           platforms: linux/amd64,linux/arm64
           push: false
+          build-args: |
+            PENNYLANE_VERSION=${{ github.event.inputs.pennylane_version }}
+            CATALYST_VERSION=${{ github.event.inputs.catalyst_version }}
--- a/.github/workflows/build-docker.yaml
+++ b/.github/workflows/build-docker.yaml
@@ -26,5 +26,5 @@ jobs:
           platforms: linux/amd64
           push: false
           build-args: |
-            PENNYLANE_VERSION=${{ github.event.inputs.pennylane_version }}
-            CATALYST_VERSION=${{ github.event.inputs.catalyst_version }}
+            PENNYLANE_VERSION=main
+            CATALYST_VERSION=main
--- a/.github/workflows/build-docker.yaml
+++ b/.github/workflows/build-docker.yaml
@@ -9,8 +9,7 @@ on:
 jobs:
   build:
     name: Build Docker Image
-    runs-on:
-      group: 'PL Additional Runners'
+    runs-on: blacksmith-8vcpu-ubuntu-2404	
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
--- a/.github/workflows/build-docker.yaml
+++ b/.github/workflows/build-docker.yaml
@@ -20,7 +20,7 @@ jobs:
     
   build:
     name: Build Docker Image
-    runs-on: blacksmith-16vcpu-ubuntu-2404-arm
+    runs-on: blacksmith-8vcpu-ubuntu-2404-arm
     needs: [check_if_wheel_build_required, constants]    
     strategy:
       fail-fast: false
@@ -43,6 +43,4 @@ jobs:
           build-args: |
             PENNYLANE_VERSION=main
             CATALYST_VERSION=main
-            GITHUB_WORKSPACE=${{ github.workspace }}
-            LLVM_CACHE=false
-            STABLEHLO_CACHE=false
+            LIGHTNING_VERSION=main
--- a/.github/workflows/build-docker.yaml
+++ b/.github/workflows/build-docker.yaml
@@ -99,5 +99,5 @@ jobs:
             PENNYLANE_VERSION=main
             CATALYST_VERSION=main
             GITHUB_WORKSPACE=${{ github.workspace }}
-            LLVM_CACHE=true
+            LLVM_CACHE=false
             STABLEHLO_CACHE=false
--- a/.github/workflows/build-docker.yaml
+++ b/.github/workflows/build-docker.yaml
@@ -20,7 +20,7 @@ jobs:
     
   build:
     name: Build Docker Image
-    runs-on: blacksmith-8vcpu-ubuntu-2404	
+    runs-on: blacksmith-16vcpu-ubuntu-2404-arm
     needs: [check_if_wheel_build_required, constants]    
     strategy:
       fail-fast: false
--- a/.github/workflows/build-docker.yaml
+++ b/.github/workflows/build-docker.yaml
@@ -17,32 +17,11 @@ jobs:
     uses: ./.github/workflows/constants.yaml
     with:
       runs_on: ubuntu-24.04
-  
-  check_caches:
-    needs: [check_if_wheel_build_required, constants]
-    name: Check for cached builds
-    runs-on: ubuntu-24.04
-    outputs:
-      llvm_cached: ${{ steps.llvm.outputs.cache-hit == 'true' }}
-      stablehlo_cached: ${{ steps.stablehlo.outputs.cache-hit == 'true' }}
-    steps:
-      - uses: actions/cache/restore@v6
-        id: llvm
-        with:
-          path: llvm-build
-          key: manylinux_2_28_x86_64-llvm-${{ needs.constants.outputs.llvm_version }}-${{ needs.constants.outputs.primary_python_version }}-wheel-build
-          lookup-only: true 
-      - uses: actions/cache/restore@v6
-        id: stablehlo
-        with:
-          path: stablehlo-build
-          key: manylinux_2_28_x86_64-stablehlo-${{ needs.constants.outputs.stablehlo_version }}-wheel-build
-          lookup-only: true
     
   build:
     name: Build Docker Image
     runs-on: blacksmith-8vcpu-ubuntu-2404	
-    needs: [check_if_wheel_build_required, constants, check_caches]    
+    needs: [check_if_wheel_build_required, constants]    
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


.github/workflows/build-docker.yaml
```diff
@@ -0,0 +1,75 @@
...
+  constants:
+    needs: [check_if_wheel_build_required]
+    name: "Set build matrix"
+    uses: ./.github/workflows/constants.yaml
+    with:
+      runs_on: ubuntu-24.04
+    
+  build:
+    name: Build Docker Image
+    runs-on: blacksmith-8vcpu-ubuntu-2404-arm
```
josephleekl, line 23: why do we build on the arm ones? we should probably start with x86

Dockerfile
```diff
@@ -0,0 +1,384 @@
...
+
+# Define global build defaults
+ARG PENNYLANE_VERSION=main
+ARG LIGHTNING_VERSION=main
+ARG CATALYST_VERSION=main
+ARG GCC_VERSION=13
+ARG CUDA_INSTALLER=https://developer.download.nvidia.com/compute/cuda/12.9.1/local_installers/cuda_12.9.1_575.57.08_linux.run
+ARG ROCM_INSTALLER=https://repo.radeon.com/amdgpu-install/7.2.4/ubuntu/noble/amdgpu-install_7.2.4.70204-1_all.deb
+ARG AMD_ARCH=AMD_GFX942
+ARG CUDA_ARCH=AMPERE80
```
josephleekl: would you mind removing all the GPU stuff that are unnecessary for this?

Dockerfile
```diff
@@ -0,0 +1,384 @@
...
+RUN git clone --depth 1 --branch ${CATALYST_VERSION} \
+    --recurse-submodules --shallow-submodules \
+    https://github.com/PennyLaneAI/catalyst.git /tmp/catalyst-src \
+    && cp -a /tmp/catalyst-src/. /opt/catalyst/ \
+    && rm -rf /tmp/catalyst-src
+ENV LLVM_BUILD_DIR=/opt/catalyst/llvm-build
+ENV PATH="${LLVM_BUILD_DIR}/bin:${PATH}"
+
+# ENV LLVM_TARGETS=check-mlir
+RUN cd /opt/catalyst/mlir/llvm-project \
```
josephleekl, line 51: try and use the makefile/make commands in general (over using cmake directly), which i think will apply these patches as well. that makes it more maintainable instead of applying them by hand here. have a look at the other CI actions :)

Dockerfile
```diff
@@ -0,0 +1,384 @@
...
+    && git apply /opt/catalyst/mlir/patches/llvm-bufferization-segfault.patch \
+    && git apply /opt/catalyst/mlir/patches/llvm-python-bindinggen-annotations.patch
+RUN cd /opt/catalyst/mlir/Enzyme \
+    && git apply /opt/catalyst/mlir/patches/enzyme-nvvm-fabs-intrinsics.patch
+
+RUN PYTHON=$PYTHON \
```
josephleekl, line 57: also could we make a separate image just for the deps (llvm, stablehlo, enzyme) - possibly even using a separate dockerfile - so we can re-use that when we rebuild the catalyst docker?

Dockerfile
```diff
@@ -0,0 +1,384 @@
...
+
+FROM base-build AS base-build-python
+ARG LIGHTNING_VERSION
+WORKDIR /opt/pennylane-lightning
+ENV VIRTUAL_ENV=/opt/venv-build
+RUN python3 -m venv $VIRTUAL_ENV
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+RUN rm -rf tmp && git clone --depth 1 --branch ${LIGHTNING_VERSION} https://github.com/PennyLaneAI/pennylane-lightning.git tmp\
+    && mv tmp/* /opt/pennylane-lightning && rm -rf tmp
+RUN pip install --no-cache-dir build cmake ninja toml wheel setuptools>=75.8.1
```
josephleekl: i wonder if instead of copying the lightning dockerfile here verbatim and building lightning here within the dockerfile, instead we:
- pull lightning in the github action
- build the docker/wheel using the lightning dockerfile
- then use the built docker/wheel here in catalyst? 
there's a slight additional complication: catalyst normally specify a specific pennylane and catalyst to use, which i think by defaults it pulls from e.g. testPyPI. so when you install catalyst last, it'll try to install the version specified by the dependency. If we want to override that, we might have to install lightning catalyst. One suggestion is that if the user doesn't specify a lightning version, we don't install lightning separately; only if they specify, then we build that from source. 

not directly copyable, but how the benchmark suite does the installation might be a useful reference: https://github.com/PennyLaneAI/pennylane-benchmarks/blob/main/.github/workflows/scripts/install_PL-Ln-Ct.sh

Dockerfile
runora95: remove all the lightning gpu stuff?

josephleekl: yes - to build with catalyst, we probably just need lightning qubit and lightning kokkos

josephleekl: (just the basic lightning kokkos, no need for all the variants as well)

Dockerfile
runora95: building a separate image is definitely doable, it would also eliminate the need to maintain the lightning docker build.

runora95: > then use the built docker/wheel here in catalyst?
there's a slight additional complication: catalyst normally specify a specific pennylane and catalyst to use, which i think by defaults it pulls from e.g. testPyPI. so when you install catalyst last, it'll try to install the version specified by the dependency. If we want to override that, we might have to install lightning catalyst. One suggestion is that if the user doesn't specify a lightning version, we don't install lightning separately; only if they specify, then we build that from source.

so install them in this order: catalyst -> lightning -> pennylane?

.github/workflows/build-docker.yaml
```diff
@@ -0,0 +1,109 @@
+name: Build Pennylane Lightning Catalyst Docker Image
```
josephleekl, line 1:
```suggestion
name: Build Catalyst Docker Image
```

.github/workflows/build-docker.yaml
```diff
@@ -0,0 +1,109 @@
...
+  contents: read
+  
+on:
+    pull_request:
+    workflow_dispatch:
+        
+jobs:
+  check_if_wheel_build_required:
+    if: github.event.pull_request.draft == false
+    uses: ./.github/workflows/check-for-wheel-build.yml
```
josephleekl, line 13: i don't think we need to build for every PR, maybe just when a label is added or manually triggered?

.github/workflows/build-docker.yaml
```diff
@@ -0,0 +1,109 @@
...
+
+  constants:
+    needs: [check_if_wheel_build_required]
+    name: "Set build matrix"
+    uses: ./.github/workflows/constants.yaml
+    with:
+      runs_on: ubuntu-24.04
+      
+      # - name: Build and push Docker image
```
josephleekl, line 22: can we remove these?

.github/workflows/build-docker.yaml
```diff
@@ -0,0 +1,109 @@
...
+      matrix:
+        python_version: ${{ fromJson(format('["{0}"]', needs.constants.outputs.primary_python_version)) }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      
+      # - name: Build and push Docker image
```
josephleekl, line 48: same

.github/workflows/build-docker.yaml
```diff
@@ -0,0 +1,109 @@
...
+      #     context: .
+      #     file: Dockerfile
+      #     platforms: linux/arm64
+      #     push: false
+      #     build-args: |
+      #       PENNYLANE_VERSION=main
+      #       CATALYST_VERSION=main
+      #       LIGHTNING_VERSION=main
+      # Set up Blacksmith's high-speed NVMe Docker layer caching
+      - name: Set up Docker Builder
```
josephleekl, line 60: why are we doing both this and `docker/setup-buildx-action@v3` above?

.github/workflows/build-docker.yaml
```diff
@@ -0,0 +1,109 @@
...
+      #   with:
+      #     context: .
+      #     file: ./Dockerfile
+      #     tags: plc:latest
+      #     load: true 
+      #     build-contexts: |
+      #       pennylane-lightning:latest=docker-archive://pennylane-lightning.tar
+      - name: Build Docker Image
+        run: |
+          docker build \
```
josephleekl, line 93: we're building docker directly instead of using the actions?

.github/workflows/build-docker.yaml
josephleekl: i think it might fail when we do both

.github/workflows/build-docker.yaml
josephleekl: this will prevent caching

.github/workflows/check-catalyst.yaml
```diff
@@ -173,6 +174,8 @@ jobs:
         CXX_COMPILER=$(which ${{ needs.constants.outputs[format('cxx_compiler.{0}', matrix.compiler)] }}) \
         LLVM_BUILD_DIR="$(pwd)/llvm-build" \
         COMPILER_LAUNCHER="" \
+        PYTHON3_EXECUTABLE=$(which python3) \
+        Python3_NumPy_INCLUDE_DIRS=$$($(PYTHON) -c "import numpy as np; print(np.get_include())") \
```
josephleekl, line 178: the `$$(` is that correct? error might be hidden by the cache hit

Dockerfile
```diff
@@ -0,0 +1,175 @@
...
+    && git apply /opt/catalyst/mlir/patches/llvm-python-bindinggen-annotations.patch
+RUN cd /opt/catalyst/mlir/Enzyme \
+    && git apply /opt/catalyst/mlir/patches/enzyme-nvvm-fabs-intrinsics.patch
+
+RUN PYTHON=$PYTHON \
+    C_COMPILER=$(which gcc)  \
+    CXX_COMPILER=$(which g++)  \
+    LLVM_BUILD_DIR="/opt/catalyst/llvm-build" \
+    LLVM_PROJECTS="lld;mlir" \
+    LLVM_TARGETS="lld check-mlir" \
```
josephleekl, line 62: we may not need `check-mlir` , this runs unnecessary tests

Dockerfile
```diff
@@ -0,0 +1,175 @@
...
+RUN C_COMPILER=$(which gcc) \
+    CXX_COMPILER=$(which g++) \
+    LLVM_BUILD_DIR="$(pwd)/llvm-build" \
+    STABLEHLO_BUILD_DIR="/opt/catalyst/stablehlo-build" \
+    COMPILER_LAUNCHER="" \
+    ENABLE_LLD=OFF \
+    make stablehlo
+
+# Build enzyme
+FROM base-catalyst AS build-enzyme
```
josephleekl, line 81: i think when doing `make llvm` it applies some patches to enzyme as well. when we split like this, those patches are not applied here.

For the steps of building llvm/enzyme/stable-hlo, we don't have to split them into multiple chunks like this

Dockerfile
```diff
@@ -0,0 +1,175 @@
...
+    make plugin-wheel
+RUN PYTHON=$PYTHON \
+    LLVM_BUILD_DIR="/opt/catalyst/llvm-build" \
+    STABLEHLO_BUILD_DIR="/opt/catalyst/stablehlo-build" \
+    DIALECTS_BUILD_DIR="/opt/catalyst/quantum-build" \
+    RT_BUILD_DIR="/opt/catalyst/runtime-build" \
+    OQC_BUILD_DIR="/opt/catalyst/oqc-build" \
+    ENZYME_BUILD_DIR="/opt/catalyst/enzyme-build" \
+    make wheel
+RUN auditwheel repair dist/*.whl -w ./wheel --no-update-tags --exclude libopenblasp-r0-23e5df77.3.21.dev.so
```
josephleekl, line 154: why do we hardcode the library excluded?

.github/workflows/check-catalyst.yaml
runora95: typo 😅

.github/workflows/build-docker.yaml
runora95: yeah, I wasn't able to get the manually built lightning image from the previous step to work inside the catalyst dockerfile

Dockerfile
runora95: it's how it is in the build wheel workflows https://github.com/PennyLaneAI/catalyst/blob/main/.github/workflows/build-wheel-linux-x86_64.yaml#L447

.github/workflows/build-docker.yaml
runora95: nvm I found a way to use the docker action

PR #3183, opened 2026-09-03
Decomp/xas misc fixes

paul0403:
**Context:**
Miscellaneous fixes found out while testing XAS.


frontend/catalyst/decomposition/decomposition_rules.py
JerryChen97, line 830: Seems that the always-inline pass doesn't reach `func.call`s nested inside `qref.ctrl` / adjoint regions, so pjit helpers survive there. whether those ops implement `RegionBranchOpInterface`/the inliner interface, or whether inlining should run *before* ctrl/adjoint wrapping?
In one of my workflow compilation I saw `_right_givens_jax` survives — and **every** failing loc is `ctrl/…` or `adjoint/…` prefixed (`ctrl/jit(angle)`, `adjoint/jit(remainder)`, `ctrl/adjoint/for/body/jit(allclose)`).

frontend/catalyst/decomposition/graph_op_id.py
JerryChen97, line 191:
```suggestion
            if wire_name not in self.op.hybrid_argnames and len(wire_arg) > 0:
```
When compiling EELS, I saw `SumOfSlatersPrep` with an `identification_wires` register that legitimately has zero wires — `required_register_sizes` returns 0 for these index sets. Python's `parse_wire_lens` still writes it into the ID as `identification_wires:0`, but the compiler builds its ID from qubit operands, and an empty register contributes none, so it leaves the entry out entirely. The two IDs never match, the `sos_state_prep` rule ends up registered against a node no operator instance maps to, and the solver reports "Decomposition rule not found for SumOfSlatersPrep"
I tried skipping empty register here and EELS works. But not sure if this is a proper fix without bad side effects @maliasadi

frontend/catalyst/decomposition/decomposition_rules.py
paul0403: https://github.com/PennyLaneAI/catalyst/blob/main/mlir/test/QRef/InlineTest.mlir are you sure? There are tests for calls in adjoint and control regions

frontend/catalyst/decomposition/graph_op_id.py
paul0403: I mean the MLIR op cannot have a `qubit_index` name that is indexing nothing :sweat_smile:

frontend/catalyst/decomposition/decomposition_rules.py
JerryChen97: > https://github.com/PennyLaneAI/catalyst/blob/main/mlir/test/QRef/InlineTest.mlir are you sure? There are tests for calls in adjoint and control regions

I'm not so sure now since my local branch also changed a lot since last time I reported this :laughing:

JerryChen97: Let me take a second look

PR #3184, opened 2026-09-03
Single-source of truth GraphOpID generation

dime10:
**Context:**

Graph operation identifiers were independently assembled as delimiter-based strings in the Python frontend and C++ compiler. This made the encoding sensitive to formatting differences, some of which are identified below:
image

This PR replaces the existing GraphOpID encodings across the frontend and compiler with the canonical MLIR representation of a typed dictionary attribute. It does this by converting Python values to MLIR attributes first, assembling the full GraphOpID attribute structure (the "struct"), and then converting to string via the builtin MLIR attribute printers. The conversion to string could be eliminated entirely, but to limit the scope of the PR I'm keeping the string representation in all the users of the GraphOpID.

The struct schema is roughly based on the following:

- stores the operation name under `op`,
- uses positional arrays for `wires` and `parameter` types instead of dictionaries to keep the ID shorter,
- adjoint and control modifiers are grouped under `traits` (this is mainly done to highlight the op name more, otherwise the adjoint field would be printed first by MLIR),
- omits fields containing default or empty values to keep the ID shorter,
- includes static data and UID as before.

All frontend producers use a shared `build_graph_op_key` helper. The `DecomposableGate` default implementation constructs the equivalent attribute in C++, but will still produce a string as a result.

**Benefits:**

- Single source of truth for generating GraphOpIDs across Python and C++.
- Less custom logic for the ID generation, which mostly reuses MLIR upstream printers.
- Makes parameter and wire names irrelevant to operator identity.
- Modifier handling now updates structured fields instead of adding or removing textual prefixes.

**Possible Drawbacks:**

To limit the scope of the PR, I didn't change any consumer code of the GraphOpID, so it still expects strings. As a result, we still have a conversion of the structured dictionary representation to string, and on the other side a parsing step currently used for two reasons: generating modified IDs (e.g. adding an adjoint), and obtaining a simplified name for gate set matching. Ideally, the ID should be fully opaque, and once generated not touched except to test equality. Hopefully something we can further improve in the future.

### DIFF
--- a/frontend/catalyst/decomposition/graph_op_id.py
+++ b/frontend/catalyst/decomposition/graph_op_id.py
@@ -14,46 +14,178 @@
 
 """Python implementation of Graph Operator ID."""
 
+from contextlib import contextmanager
 from typing import Any
 
-import jax.numpy as jnp
 import pennylane as qp
+from jax._src.lib.mlir import ir
 from pennylane.pytrees import flatten
 
 from catalyst.decomposition.type_utils import (
     convert_item_to_mlir_type,
-    format_dynamic_params_for_id,
     post_process_concretize_leaves,
     replace_wires_with_placeholder_wires,
 )
 from catalyst.from_plxpr.uid import generate_uid
+from catalyst.jax_extras.lowering import get_mlir_attribute_from_pyval
 
 _SPECIAL_LOWERINGS = {}
+
+
+@contextmanager
+def _mlir_context():
+    """Provide an MLIR context and location for attribute construction."""
+    if current := ir.Context.current:
+        with ir.Location.unknown(context=current):
+            yield current
+    else:
+        with ir.Context() as context, ir.Location.unknown(context=context):
+            yield context
+
+
+def build_graph_op_key(
+    base_name: str,
+    dynamic_types: dict[str, list[str]],
+    wire_lengths: dict[str, int],
+    static_data: dict[str, Any],
+    *,
+    adjoint: bool = False,
+    num_controls: int = 0,
+    uid: int | None = None,
+) -> str:
+    """Build the canonical structured GraphOpID string.
+
+    Parameter and wire groups are identified by their position, so both mappings must be in
+    frontend signature order.
+    """
+    if num_controls < 0:
+        raise ValueError("GraphOpID control count cannot be negative")
+
+    with _mlir_context():
+        fields = {"op": ir.StringAttr.get(base_name)}
+        traits = {}
+        if adjoint:
+            traits["adj"] = ir.BoolAttr.get(True)
+        if num_controls:
+            traits["controls"] = get_mlir_attribute_from_pyval(num_controls)
+        if dynamic_types:
+            fields["params"] = ir.ArrayAttr.get(
+                [
+                    ir.ArrayAttr.get(
+                        [ir.TypeAttr.get(ir.Type.parse(type_name)) for type_name in types]
+                    )
+                    for types in dynamic_types.values()
+                ]
+            )
+        if static_data:
+            fields["static"] = get_mlir_attribute_from_pyval(static_data)
+        if traits:
+            fields["traits"] = ir.DictAttr.get(traits)
+        if uid is not None:
+            fields["uid"] = get_mlir_attribute_from_pyval(uid)
+        if wire_lengths:
+            fields["wires"] = get_mlir_attribute_from_pyval(list(wire_lengths.values()))
+        return str(ir.DictAttr.get(fields))
+
+
+def _parse_graph_op_id(op_id: str) -> ir.DictAttr:
+    """Parse and minimally validate a structured GraphOpID."""
+    parsed = ir.Attribute.parse(op_id)
+    if not isinstance(parsed, ir.DictAttr):
+        raise ValueError(f"Malformed GraphOpID, expected a dictionary attribute: {op_id!r}")
+
+    fields = {entry.name for entry in parsed}
+    if "op" not in fields:
+        raise ValueError(f"Malformed GraphOpID, missing field 'op': {op_id!r}")
+    if unknown := fields.difference({"op", "params", "static", "traits", "uid", "wires"}):
+        raise ValueError(f"Malformed GraphOpID, unknown fields {sorted(unknown)}: {op_id!r}")
+    if "traits" in parsed:
+        traits = ir.DictAttr(parsed["traits"])
+        if unknown := {entry.name for entry in traits}.difference({"adj", "controls"}):
+            raise ValueError(
+                f"Malformed GraphOpID, unknown trait fields {sorted(unknown)}: {op_id!r}"
+            )
+    return parsed
+
+
+def graph_op_id_modifiers(op_id: str) -> tuple[str, bool, int]:
+    """Return ``(base_name, adjoint, num_controls)`` from a GraphOpID."""
+    with _mlir_context():
+        parsed = _parse_graph_op_id(op_id)
+        base_name = ir.StringAttr(parsed["op"]).value
+        traits = ir.DictAttr(parsed["traits"]) if "traits" in parsed else None
+        adjoint = ir.BoolAttr(traits["adj"]).value if traits and "adj" in traits else False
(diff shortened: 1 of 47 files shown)


mlir/include/Quantum/IR/QuantumInterfaces.td
```diff
@@ -207,17 +207,14 @@ def DecomposableGate : OpInterface<"DecomposableGate", [QuantumGate]> {
...
         InterfaceMethod<
             "Return the operation's ID for use with the graph. "
-            "This should be computed by the following convention:\n"
-            "```\n"
-            "name + {map params to list of types} + {map wire args to number of wires} + {static data} + [extra data]\n"
-            "```\n"
-            "Note that extra data is currently specific to `OperatorOp`s, and should only appear when present."
-            "Each of these maps should be sorted lexicographically by key."
+            "Dynamic parameter type groups and wire lengths are identified positionally in "
+            "frontend-signature/operand order; frontend argument names are not part of identity. "
```
kipawaa, line 216: This was part of the original design but created a lot of difficulty when trying to call resource functions, decomposition rules etc. Is the inspiration for removing parameter names just to shorten the ID? Similarly, if we want to handle user ops in this way we cannot know the order of their params, but we _can_ know the names (via `param_map`).

paul0403: Haven't looked at the whole PR yet, but I think David is just removing the names from the ID string in this PR. Args are still passed into the entry point func via dictionaries.

kipawaa:
> assembling the full GraphOpID attribute structure

One of my inspirations for generating a single string ID rather than a structure was because I intended for it to be static and opaque, i.e. it was intended to only be used for equality comparison. Do we need greater functionality than this? Not that a structure is worse, in fact it might be better in terms of performance, but I worry that this will encourage more manipulation of IDs (similar to what we see with adjoint/control modifiers in the frontend right now). My goal would be to generate new IDs from operator data rather than manipulate ID data, though maybe at this point I'm holding on to an idea that will only get in our way...

mlir/include/Quantum/IR/QuantumInterfaces.td
kipawaa: Ahhh okay, if it's still part of the dynamic parsing then I don't think this is an issue. Thanks!

dime10:
> > assembling the full GraphOpID attribute structure
> 
> One of my inspirations for generating a single string ID rather than a structure was because I intended for it to be static and opaque, i.e. it was intended to only be used for equality comparison. Do we need greater functionality than this? Not that a structure is worse, in fact it might be better in terms of performance, but I worry that this will encourage more manipulation of IDs (similar to what we see with adjoint/control modifiers in the frontend right now). My goal would be to generate new IDs from operator data rather than manipulate ID data, though maybe at this point I'm holding on to an idea that will only get in our way...

I agree that the ID should be opaque (at least for the code, inspectability aside), however that is already not the case in two instances: generating adjoint/control versions of decomposition reaches in and manipulates strings, and generating the shortened name to compare against the gate set. I would support moving towards never touching the ID once it's generated. 

In this PR, I don't really change that fact. The ID is generated by going through a structured attribute first, but then it is serialized to string. There is still the parsing of the string for the two use-cases I mention above. Perhaps we can fix those first.
I'm also not convinced this particular implementation in this PR is the way to go, but I do think the core change of using the attribute printer in both frontend and compiler to generate identical IDs across the board and minimize custom serialization code is the way to go.

kipawaa:
@dime10 Sounds like we're on the same page, and I agree that sending everything to MLIR attributes then serializing for consistency is ideal. Let's go ahead with this (I'll review shortly) and I can try to address the string manipulation in a separate PR :)

mlir/lib/Quantum/IR/QuantumInterfaces.cpp
```diff
@@ -37,72 +41,151 @@ using namespace catalyst::quantum;
...
-                printAttr(attr, ss);
-            }
-            ss << "]";
-        })
-        .Case<mlir::StringAttr>([&](mlir::StringAttr attr) { ss << attr.str(); })
-        .Case<mlir::IntegerAttr>([&](mlir::IntegerAttr attr) { ss << attr.getInt(); })
-        .Case<mlir::FloatAttr>([&](mlir::FloatAttr attr) { ss << attr.getValueAsDouble(); })
-        .Default([&](mlir::Attribute attr) { attr.print(ss); });
+int64_t groupStartIndex(NamedAttribute entry) {
+    auto indices = cast<DenseI64ArrayAttr>(entry.getValue()).asArrayRef();
```
kipawaa: This seems weird to me, can you explain why we hard-cast to I64Array? Surely we receive non-I64 attrs in dynamic data, but this seems to handle it fine?

dime10:
> @dime10 Sounds like we're on the same page, and I agree that sending everything to MLIR attributes then serializing for consistency is ideal. Let's go ahead with this (I'll review shortly) and I can try to address the string manipulation in a separate PR :)

I don't really like how many changes there are in this PR (although a lot of them are tests), so I am experimenting with an alternative implementation. Not sure I've landed on something I really like yet 🤔

mlir/lib/Quantum/IR/QuantumInterfaces.cpp
```diff
@@ -37,72 +41,151 @@ using namespace catalyst::quantum;
...
+    return ordered;
 }
 
-template <typename T, typename PrintFunc>
-void printSortedMap(const llvm::StringMap<T> &map, llvm::raw_string_ostream &ss,
-                    PrintFunc printValue) {
-    llvm::SmallVector<llvm::StringRef> keys;
-    for (const llvm::StringRef key : map.keys()) {
-        keys.push_back(key);
+SmallVector<SmallVector<Type>> getOrderedDynamicTypes(DecomposableGate gate, Operation *op) {
```
kipawaa: Can we remove the `Operation*` and just use the `DecomposableGate`? This should be an interface on top of an operation that we can access, and would simplify the signature.

mlir/lib/Quantum/IR/QuantumInterfaces.cpp
```diff
@@ -37,72 +41,151 @@ using namespace catalyst::quantum;
...
+    if (auto custom = dyn_cast<CustomOp>(op)) {
```
kipawaa: It might be nice to use an `llvm::TypeSwitch` here

mlir/lib/Quantum/IR/QuantumInterfaces.cpp
```diff
@@ -37,72 +41,151 @@ using namespace catalyst::quantum;
...
+    result.reserve(names.size());
+    for (StringRef name : names) {
+        result.push_back(wireLengths.lookup(name));
+    }
+    return result;
 }
 
-void printWireLens(const llvm::StringMap<size_t> &map, llvm::raw_string_ostream &ss) {
-    printSortedMap(map, ss, [](size_t len, llvm::raw_string_ostream &stream) { stream << len; });
+mlir::DictionaryAttr buildGraphOpKeyAttr(DecomposableGate gate, Operation *op) {
```
kipawaa: Do we want to switch to `key` terminology, or should we continue using `GOID`? I don't feel strongly either way, but I do think we should be consistent

mlir/lib/Quantum/IR/QuantumInterfaces.cpp
kipawaa: This seems much better, thanks for putting this together!

frontend/catalyst/decomposition/graph_op_id.py
```diff
@@ -14,46 +14,178 @@
...
+    static_data: dict[str, Any],
+    *,
+    adjoint: bool = False,
+    num_controls: int = 0,
+    uid: int | None = None,
+) -> str:
+    """Build the canonical structured GraphOpID string.
+
+    Parameter and wire groups are identified by their position, so both mappings must be in
+    frontend signature order.
```
kipawaa: From the definition in the compiler, this seems to be in _alphabetical order_ sorted by frontend name?

PR #3185, opened 2026-09-04
Support new discrete `qp.PPR` in `to_ppr` pass.

dwierichs:
**Context:**
https://github.com/PennyLaneAI/pennylane/pull/10107 adds a new `qp.PPR` class that represents discrete PPRs.
We would like this op to get lowered into a `pbc.ppr` directly when using `to_ppr`.

Register `PPR` in `to_ppr` and lower it to `pbc.ppr`.

**Benefits:**
Cheaper decompositions/compilation.

**Possible Drawbacks:**

**Related GitHub Issues:**


doc/releases/changelog-dev.md
```diff
@@ -479,6 +479,10 @@
 * Added ``CZ`` support to ``to-ppr`` pass.
   [(#3009)](https://github.com/PennyLaneAI/catalyst/pull/3009)
 
+* ``to_ppr`` now directly lowers PennyLane's discrete ``PPR`` operator to ``pbc.ppr``, without
+  first decomposing it through ``PauliRot``.
+  [(#xxxx)](https://github.com/PennyLaneAI/catalyst/pull/xxxx)
```
dwierichs:
```suggestion
  [(#3185)](https://github.com/PennyLaneAI/catalyst/pull/3185)
```

mlir/lib/PBC/Transforms/ToPPR.cpp
```diff
@@ -469,6 +518,10 @@ struct PBCGateLowering : public OpInterfaceConversionPattern<QuantumOperation> {
             return convertMultiRZGate(originOp, rewriter);
         } else if (auto originOp = dyn_cast<PauliRotOp>(op)) {
             return convertPauliRotGate(originOp, rewriter);
+        } else if (auto originOp = dyn_cast<OperatorOp>(op)) {
```
dwierichs, line 520: Would it be better to add a `PPROp` similar to `PauliRotOp`? 🤔

docs/superpowers/plans/2026-09-10-ppr-to-ppm-adjoint.md
```diff
@@ -0,0 +1,180 @@
+# Adjoint PPR-to-PPM Implementation Plan
```
dwierichs: Will have to work on the agent not committing its helper files. Sorry about that.

mlir/lib/PBC/Transforms/ToPPR.cpp
```diff
@@ -411,6 +411,54 @@ LogicalResult convertPauliRotGate(PauliRotOp op, ConversionPatternRewriter &rewr
                                    op.getAdjoint(), rewriter);
 }
 
+LogicalResult convertPPROperator(OperatorOp op, ConversionPatternRewriter &rewriter) {
+    if (!op.getAllParams().empty()) {
+        return op.emitOpError("PPR operator does not support dynamic parameters");
+    }
```
dime10: Is it the responsibility of this pass to check that the PPR op is malformed? 🤔 

I would say no, because if it was does that mean we check such properties over and over again in any pass that interacts with an op? The line is not 100% clear to me, but generally assumptions the pass uses concretely uses could be added as assertions, but general verification should the responsibility of a different piece of code (normally the verifier, but this is where having opaque ops is a disadvantage, as we don't have any op-specific verifiers for them).

mlir/lib/PBC/Transforms/ToPPR.cpp
```diff
@@ -411,6 +411,54 @@ LogicalResult convertPauliRotGate(PauliRotOp op, ConversionPatternRewriter &rewr
...
+    int64_t denominator = denominatorAttr.getInt();
+    if (denominator != 2 && denominator != -2 && denominator != 4 && denominator != -4 &&
+        denominator != 8 && denominator != -8) {
+        return op.emitOpError("unsupported PPR angle denominator: ") << denominator;
+    }
+
+    auto pauliWordAttr = staticData.getAs<StringAttr>("pauli_word");
+    if (!pauliWordAttr) {
+        return op.emitOpError("PPR operator requires a string 'pauli_word' in static_data");
+    }
```
dime10: Minor inconsistency that here we check the existence of the attribute but above we don't (we just use it directly).

mlir/lib/PBC/Transforms/ToPPR.cpp
```diff
@@ -411,6 +411,54 @@ LogicalResult convertPauliRotGate(PauliRotOp op, ConversionPatternRewriter &rewr
...
+    for (char pauli : pauliWord) {
+        if (pauli != 'X' && pauli != 'Y' && pauli != 'Z' && pauli != 'I') {
+            return op.emitOpError("PPR operator Pauli word may contain only X, Y, Z and I");
+        }
+        pauliCharacters.push_back(rewriter.getStringAttr(StringRef(&pauli, 1)));
+    }
+
+    ArrayAttr pauliProduct = rewriter.getArrayAttr(pauliCharacters);
+    int8_t rotationKind = static_cast<int8_t>(denominator);
+    if (op.getAdjoint()) {
```
dime10, line 429: We might want to check the op is uncontrolled? We are currently just dropping control qubits I think.

dime10 (commented):
Thanks for adding this David 🎉

mlir/lib/PBC/Transforms/ToPPR.cpp
dime10: Do you mean create a custom mlir op for it, rather than representing it as an quantum.operator op? 

It's true that currently we're assuming any quantum.operator instance would be resolved by the graph decomposer (before further compilation or execution).

mlir/lib/PBC/Transforms/ToPPR.cpp
dwierichs: Mhm yeah, that is a fair point. I am not so experienced with contributing at this level, so I was not aware that this is a bit out of place.
What would be your preference here? Drop all of the validation? 🤔

mlir/lib/PBC/Transforms/ToPPR.cpp
dwierichs: Ah, very good catch 😅

mlir/lib/PBC/Transforms/ToPPR.cpp
dwierichs: I think I would maybe keep the validation that there is no dynamic parameter, similar to how the other ops are checked for having the right number (usually 1) of dynamic parameters?

mlir/lib/PBC/Transforms/ToPPR.cpp
dwierichs: I think we actually do this already in general during `matchAndRewrite` in `PBCGateLowering`, no?

mlir/lib/PBC/Transforms/ToPPR.cpp
dwierichs: This is gone with the removed validation.

mlir/lib/PBC/Transforms/ToPPR.cpp
dime10: Ah I might have missed that, thanks!

mlir/lib/PBC/Transforms/ToPPR.cpp
dime10: Yeah not so easy. I guess a good location for verification is the PPR operator in the PBC dialect itself. It already checks that rotation kind attribute, so that's superfluous. It doesn't currently check the Pauli letters, which we could add there (it's actually supposed to be a enum attribute which have verified this intrinsically, but right now it's just strings). It does verify the length of the pauli matches the qubits though. 

So I think you're right, the one that couldn't be verified there is the dynamic params since those would be silently dropped at this stage. It still irks me a bit to have that here because it has nothing to do with the pass: a PPR op with dynamic params is intrinsically malformed, no matter where it appears. Compare that to control qubits which are legal but not supported by this pass, so those should be verified.

If we look into the design philosophy in mlir, passes should generally assume that any input IR is legal, and the invariant they must fulfil is that any legal IR is transformed into legal IR at the output, otherwise the pass has a bug. The legality _can_ be enforced by the verifier, but it can also be turned off for performance if desired. I think the goal of this philosophy is to not pay the cost of legality verification repeatedly or excessively, and to have it centralized such that it can be controlled.

Imo this means that checking dynamic args in this pass is not the right place, but I'd be happy with an assertion about it (shows that this is an assumption we make, but we are not error checking it here (with a plan on what to do if didn't hold)).

dwierichs: Okay, that makes a lot of sense to me 👌 From my side, we can even skip the assertion about the dynamic args, unless you'd prefer it be added? :)

mlir/lib/PBC/Transforms/ToPPR.cpp
dwierichs: Yes. It seems to work fine as-is, but I don't want to ruin the separation boundary...

>It's true that currently we're assuming any quantum.operator instance would be resolved by the graph decomposer (before further compilation or execution).

Although this pass _does_ receive quantum operators and it does not leave any in that it didn't leave in before, right? 🤔

mlir/lib/PBC/Transforms/ToPPR.cpp
dime10: I think it would be good to add :)

mlir/lib/PBC/Transforms/ToPPR.cpp
dime10: While I think a better separation would be nice, I think we can live with the current implementation and I don't want to slow you down :)

dwierichs: Sounds good then :)

dime10 (approved):

mlir/lib/PBC/Transforms/ToPPR.cpp
dwierichs: Done :)

mlir/lib/PBC/Transforms/ToPPR.cpp
```diff
@@ -411,6 +411,31 @@ LogicalResult convertPauliRotGate(PauliRotOp op, ConversionPatternRewriter &rewr
                                    op.getAdjoint(), rewriter);
 }
 
+LogicalResult convertPPROperator(OperatorOp op, ConversionPatternRewriter &rewriter) {
+    assert(op.getAllParams().empty() && "PPR operator does not support dynamic parameters");
+
+    DictionaryAttr staticData = op.getStaticData();
```
sengthai, line 417: it would be nice if we guard the keys (`pauli_word`, `angle_denominator` ) as well.

mlir/lib/PBC/Transforms/ToPPR.cpp
sengthai, line 1: This PR looks good! It’s nice that pauli_measure now lowers to PPM in JAX. That said, it would be really convenient to add a PPM alias for pauli_measure in the frontend as well, just for ease of use.

sengthai (approved):
Thanks @dwierichs!

mlir/lib/PBC/Transforms/ToPPR.cpp
dwierichs: Ah, nice suggestion! I'll forward this :)

mlir/lib/PBC/Transforms/ToPPR.cpp
dwierichs: So I used to have more validation of inputs before and @dime10 suggested to reduce it. I might be misunderstanding which validation you are asking for, exactly, do you mean simple assertions for the existence of the keys?

dime10: This is the thread if you want to weigh in: https://github.com/PennyLaneAI/catalyst/pull/3185#discussion_r4019088184
I was trying to think how to best distribute the responsibility of verifying operations, the mlir philosophy is that a pass can assume verified IR as input and shouldn't have to check intrinsic operation correctness, although it can check conditions that are specific to it (but the op should be legal in either case). 
What I suggested tis hat assertions can be added for things that we know should always be true, but aren't structurally guaranteed, in order to help catch bugs earlier.

PR #3188, opened 2026-09-04
[GOID] Use unified printing for the static data portion

dime10:
This PR makes a narrow update to the graph op ID generation mechanism to use a single source of truth for static data portion only. The general approach consists of generating MLIR objects (attributes) first via the Python bindings, and then use the builtin printer to serialize those fields to string for the ID.

This change does not constitute a single source of truth for the entire ID (unlike #3184), as other fields (e.g. parameter printing), and the structure itself, still use duplicated logic across Python & C++ that must be matched precisely. However, the entire could potentially be migrated in stages with this representing the first step.

Additionally, all generation/formatting functions are refactored to be centralized in one place.

### DIFF
--- a/frontend/catalyst/decomposition/graph_op_id.py
+++ b/frontend/catalyst/decomposition/graph_op_id.py
@@ -14,23 +14,62 @@
 
 """Python implementation of Graph Operator ID."""
 
+import contextlib
 from collections.abc import Mapping, Sequence
 from typing import Any
 
 import jax.numpy as jnp
 import pennylane as qp
+from jax._src.lib.mlir import ir
 from pennylane.pytrees import flatten
 
 from catalyst.decomposition.type_utils import (
     convert_item_to_mlir_type,
-    format_dynamic_params_for_id,
-    format_static_data_dict_for_id,
     post_process_concretize_leaves,
     replace_wires_with_placeholder_wires,
 )
 from catalyst.from_plxpr.uid import generate_uid
+from catalyst.jax_extras.lowering import get_mlir_attribute_from_pyval
 
 _SPECIAL_LOWERINGS = {}
+
+
+@contextlib.contextmanager
+def _attribute_context():
+    """Provide an MLIR context and location for attribute construction."""
+    if current := ir.Context.current:
+        with ir.Location.unknown(context=current):
+            yield current
+    else:
+        with ir.Context() as context, ir.Location.unknown(context=context):
+            yield context
+
+
+def format_static_data_dict_for_id(static_data):
+    """Format the static-data group of a GraphOpID with MLIR's attribute printer."""
+    with _attribute_context():
+        attrs = {name: get_mlir_attribute_from_pyval(value) for name, value in static_data.items()}
+        return str(ir.DictAttr.get(attrs))
+
+
+def format_dynamic_params_for_id(dynamic_shape):
+    """Format the dynamic-parameter group of a GraphOpID."""
+
+    def handle_item(item):
+        match item:
+            case str():
+                return item
+            case list() | tuple():
+                return "[" + ",".join(handle_item(i) for i in item) + "]"
+
+    return (
+        "{"
+        + ",".join(
+            name + ":" + "[" + ",".join(handle_item(item) for item in types) + "]"
+            for name, types in dynamic_shape.items()
+        )
+        + "}"
+    )
 
 
 def build_graph_op_id(
--- a/frontend/catalyst/decomposition/graph_op_id.py
+++ b/frontend/catalyst/decomposition/graph_op_id.py
@@ -14,6 +14,7 @@
 
 """Python implementation of Graph Operator ID."""
 
+from collections.abc import Mapping, Sequence
 from typing import Any
 
 import jax.numpy as jnp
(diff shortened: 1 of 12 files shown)


frontend/catalyst/decomposition/graph_op_id.py
```diff
@@ -14,23 +14,90 @@
...
 from catalyst.from_plxpr.uid import generate_uid
+from catalyst.jax_extras.lowering import get_mlir_attribute_from_pyval
 
 _SPECIAL_LOWERINGS = {}
 
 
+@contextlib.contextmanager
+def _attribute_context():
+    """Provide an MLIR context and location for attribute construction."""
+    if current := ir.Context.current:
```
paul0403, line 40: Wow, didn't know they have a way to just get the current context!

frontend/catalyst/decomposition/graph_op_id.py
```diff
@@ -14,23 +14,90 @@
...
+    else:
+        with ir.Context() as context, ir.Location.unknown(context=context):
+            yield context
+
+
+def format_static_data_dict_for_id(static_data):
+    """Format the static-data group of a GraphOpID with MLIR's attribute printer."""
+    with _attribute_context():
+        attrs = {name: get_mlir_attribute_from_pyval(value) for name, value in static_data.items()}
+        return str(ir.DictAttr.get(attrs))
```
paul0403, line 52: Nice!!

frontend/catalyst/from_plxpr/qref_operator2_primitives.py
```diff
@@ -427,19 +398,16 @@ def compile_decomp_rules(
...
-        op_id = (
-            op_cls.__name__
-            + format_dynamic_params_for_id(dict(sorted(with_hybrid_dynamic_shape.items())))
-            + "{"
-            + ",".join(f"{name}:{shape}" for name, shape in sorted(with_hybrid_wire_lens.items()))
-            + "}"
+        identity_static_data = (
+            repack_static_data if not (op_cls.hybrid_argnames or op_cls.static_argnames) else {}
+        )
+        op_id = build_graph_op_id(
```
paul0403, line 404: Having this single `build_graph_op_id` function is much nicer 🤩

frontend/test/lit/test_operator2/test_decomposition_rule_entry_point.py
```diff
@@ -258,8 +258,8 @@ def rule(reg):
 
 # CHECK: func.func private @"rule_NoParams{}{reg:2}{}"
 # CHECK-SAME:   resources = {operations = {
-# CHECK-SAME:     "CompilableData{}{wires:1}{a:a,b:b,thing:thing}" = 1 : i64,
-# CHECK-SAME:     "CompilableData{}{wires:1}{a:aa,b:bb,thing:stuff}" = 2 : i64
+# CHECK-SAME:     "CompilableData{}{wires:1}{a = \22a\22, b = \22b\22, thing = \22thing\22}" = 1 : i64,
```
paul0403, line 261: There's really no other way to check quotation literals? 😅

paul0403 (approved):
🥳

frontend/test/lit/test_operator2/test_decomposition_rule_entry_point.py
dime10: Yeah this is not great, I'm not sure 🤔

paul0403: Apprantly you can just check it regularly?
```
((xas_test_env) ) paul.wang@L7450-H298SW3:~/catalyst_new/catalyst$ cat words.py 
# CHECK: "hi"
print("\"hi\"")
((xas_test_env) ) paul.wang@L7450-H298SW3:~/catalyst_new/catalyst$ python3 words.py 
"hi"
((xas_test_env) ) paul.wang@L7450-H298SW3:~/catalyst_new/catalyst$ python3 words.py | ./mlir/llvm-project/build/bin/FileCheck words.py 
((xas_test_env) ) paul.wang@L7450-H298SW3:~/catalyst_new/catalyst$ 

```

dime10: You only have a simple string in your test though. I think the escape sequence is what mlir prints when there is a quote character in a string attribute, because the attribute itself is surrounded in quotes. If it were to just print `"` that would cut the attribute short when parsing it back.

frontend/catalyst/decomposition/graph_op_id.py
kipawaa:
```suggestion
        return str(get_mlir_attribute_from_pyval(static_data))
```
There's handling in this function for dicts directly :)

frontend/test/lit/test_operator2/test_decomposition_rule_entry_point.py
kipawaa: 🥲

kipawaa (approved):
Thanks @dime10!

frontend/test/lit/test_operator2/test_decomposition_rule_entry_point.py
dime10: The way around this would be to store the dict attribute directly, only it goes against our intent to keep it opaque :(

frontend/catalyst/decomposition/graph_op_id.py
dime10: Nice :D

maliasadi (approved):

mlir/include/Quantum/IR/QuantumInterfaces.td
```diff
@@ -218,6 +218,8 @@ def DecomposableGate : OpInterface<"DecomposableGate", [QuantumGate]> {
             "```\n"
             "Note that extra data is currently specific to `OperatorOp`s, and should only appear when present."
             "Each of these maps should be sorted lexicographically by key."
+            "The static data map is spelled by MLIR's attribute printer, which the frontend also "
```
maliasadi, line 221:
```suggestion
            "The static data map is printed by MLIR's attribute printer, which the frontend also "
```

mlir/include/Quantum/IR/QuantumInterfaces.td
```diff
@@ -218,6 +218,8 @@ def DecomposableGate : OpInterface<"DecomposableGate", [QuantumGate]> {
...
+            "depends on identical id generation."
```
maliasadi, line 222:
```suggestion
            "depends on identical graphOpId generation."
```

frontend/catalyst/decomposition/graph_op_id.py
```diff
@@ -14,23 +14,89 @@
...
+        "{"
+        + ",".join(
+            name + ":" + "[" + ",".join(handle_item(item) for item in types) + "]"
+            for name, types in dynamic_shape.items()
+        )
+        + "}"
+    )
+
+
+def build_graph_op_id(
```
maliasadi, line 74: Nice! It's the new single source of truth for graphOpId generation

mlir/include/Quantum/IR/QuantumInterfaces.td
dime10: These seem inconsequential, so I'm gonna avoid another CI run

